Issue #10914, opened 2026-05-19
PR-C: [Design][Connector-V2] Define source-side document routing for Knowledge Sync #10914


yzeng1618:
## Background

Knowledge Sync requires document-scoped lifecycle decisions. A vector lifecycle sink needs to read the old chunks for one document, compare hashes, delete stale chunks, upsert changed chunks, and skip unchanged chunks. If chunks for the same `document_id` are processed by multiple sink writers, those writers can make conflicting refresh/delete decisions.

The current implementation direction added a sink-declared partition strategy and a Zeta runtime component named `SinkPartitionExchange`. The mechanism is useful as a focused experiment, but the current runtime path introduces a sink-to-sink point-to-point data channel:

```text
upstream flow -> local sink writer -> SinkPartitionExchange -> remote sink writer
```

That path is outside the normal Zeta source/transform/sink data exchange boundary. It uses Hazelcast task operations for cross-task delivery, blocks the sending task while waiting for remote dispatch completion, and executes the receiving write path synchronously inside the Hazelcast operation call stack. It also does not snapshot or restore exchange-local buffers and sequence state.

This design proposes a safer first production direction: keep document routing on the source side through `SourceSplitEnumerator` and normal engine dataflow, and defer a standalone engine exchange operator until its checkpoint and back-pressure semantics are designed as first-class runtime behavior.

## Goals

- Define the complete `document_id` data flow: assignment, routing, delivery, and landing.
- Keep Knowledge Sync routing aligned with existing source split assignment, task execution, checkpoint, and back-pressure mechanisms.
- Ensure one document belongs to one lifecycle decision domain.
- Make the addressing relationship explicit between document routing buckets, source readers, sink writers, and vector database partitions or logical buckets.
- Treat checkpoint alignment and recovery as a follow-up design area with clear state boundaries from the start.

## Non-Goals

- Do not extend the current sink-side point-to-point exchange for production lifecycle sync in this design.
- Do not introduce a standalone keyed exchange operator in the first iteration.
- Do not implement Qdrant or Milvus lifecycle behavior here.
- Do not promise exactly-once vector-store visibility until checkpoint-driven sink lifecycle semantics are designed and verified.
- Do not require every existing source to support document-level routing. Sources that cannot produce complete document splits should fail fast for lifecycle sync.

## Current Sink-Side Exchange Assessment

[...]

Issue #10889, opened 2026-05-15
[Umbrella][RAG] Define and implement production-grade Knowledge Sync for SeaTunnel


yzeng1618:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/seatunnel/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### Describe the proposal

# [Umbrella][RAG] Define and implement production-grade Knowledge Sync for SeaTunnel

## Background

This umbrella issue summarizes and tracks the work proposed by the `seatunnel-knowledge-sync-execution-plan`.

The plan positions SeaTunnel as a **Knowledge Indexing and Sync Layer** for RAG scenarios. The goal is not to turn SeaTunnel into a RAG application framework, query engine, retriever, reranker, chatbot, or agent runtime. Instead, SeaTunnel should focus on distributed, reliable, and verifiable knowledge data synchronization.

In enterprise RAG scenarios, source documents need to be continuously discovered, parsed, chunked, embedded, and synchronized into vector stores such as Qdrant and Milvus. A simple append or upsert pipeline is not enough, because production knowledge sync must also handle refresh, delete, unchanged-skip, stale chunk cleanup, retry, failover, and parallel writer consistency.

## Positioning

SeaTunnel should be positioned as:

> SeaTunnel as Knowledge Indexing and Sync Layer

SeaTunnel is responsible for:

- discovering new or changed documents
- parsing documents into structured content
- generating stable document and chunk metadata
- splitting documents into embedding-ready chunks
- generating embeddings
- writing vectors and payloads into vector stores
- supporting document-level refresh, delete, and skip-unchanged behavior
- keeping vector indexes consistent under distributed execution

SeaTunnel is not responsible for:

- query engine
- retriever
- reranker
- chatbot application
- agent orchestration
- workflow deployment
- LlamaCloud or LlamaParse-like hosted platform capability

### Enterprise knowledge base indexing

Synchronize enterprise documents into vector databases for RAG retrieval.

Typical sources include:

- local Markdown, TXT, or PDF files
- HDFS, S3, OSS, or COS object storage
- product manuals
- FAQ documents
- operation and troubleshooting documents
- internal engineering documents
- future dedicated sources such as Confluence, Google Drive, and SharePoint

### Incremental knowledge synchronization

[...]

PR #10964, opened 2026-05-27
[Improve][Connector-V2] Route Markdown file source splits by document id

yzeng1618:

This PR adds deterministic split routing for bounded Markdown file sources when `markdown_rag_metadata_enabled=true`.

Previously, file source splits were assigned by the default round-robin strategy. For Markdown RAG ingestion, the source already generates stable document metadata such as `document_id`. This PR reuses that document identity to assign each whole-file Markdown split to a deterministic source route bucket.

Main changes:

- Add `FileSourceDocumentRouting` for shared document id, source URI normalization, SHA-256 hashing, and route bucket calculation.
- Reuse the same document id logic in `MarkdownReadStrategy`.
- Route bounded Markdown file source splits by `document_id` when `markdown_rag_metadata_enabled=true`.
- Keep the existing round-robin assignment unchanged for non-Markdown sources and Markdown sources without RAG metadata.
- Apply the same routing behavior to multiple-table file source enumerators.
- Add unit tests for deterministic routing and restored enumerator behavior.
- Update English and Chinese file source documentation.

Yes.

For Markdown file sources with `markdown_rag_metadata_enabled=true`, bounded source split assignment now uses the stable `document_id` hash so all rows derived from the same Markdown document stay in the same source route bucket.

The default behavior is unchanged when:

- `markdown_rag_metadata_enabled=false`
- the file format is not Markdown
- the source is not using the bounded Markdown document routing path

No new user-facing configuration option is introduced.

Added tests

### Check list

* [* ] If you are contributing the connector code, please check that the following files are updated:

### DIFF
--- a/docs/en/connectors/source/CosFile.md
+++ b/docs/en/connectors/source/CosFile.md
@@ -205,6 +205,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/FtpFile.md
+++ b/docs/en/connectors/source/FtpFile.md
@@ -307,6 +307,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/HdfsFile.md
+++ b/docs/en/connectors/source/HdfsFile.md
@@ -132,6 +132,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/LocalFile.md
+++ b/docs/en/connectors/source/LocalFile.md
@@ -219,6 +219,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/ObsFile.md
+++ b/docs/en/connectors/source/ObsFile.md
@@ -243,6 +243,8 @@ schema {
 > - `chunk_index`: One-based chunk order in the parsed document
 > - `content_hash`: SHA-256 hash of the emitted `text` value
 >
+> When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+>
 > The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 >
 > Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/OssFile.md
+++ b/docs/en/connectors/source/OssFile.md
@@ -246,6 +246,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/OssJindoFile.md
+++ b/docs/en/connectors/source/OssJindoFile.md
@@ -207,6 +207,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/S3File.md
+++ b/docs/en/connectors/source/S3File.md
@@ -256,6 +256,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/en/connectors/source/SftpFile.md
+++ b/docs/en/connectors/source/SftpFile.md
@@ -275,6 +275,8 @@ When `markdown_rag_metadata_enabled` is set to `true`, SeaTunnel appends the fol
 - `chunk_index`: One-based chunk order in the parsed document
 - `content_hash`: SHA-256 hash of the emitted `text` value
 
+When this option is enabled for bounded Markdown file sources, the source enumerator assigns each whole-file split by the same `document_id` hash so all rows derived from one document stay in the same source route bucket. The default round-robin split assignment is unchanged when the option is disabled.
+
 The option defaults to `false`, so the original Markdown schema is unchanged unless you enable it.
 
 Note: Markdown format only supports reading, not writing.
--- a/docs/zh/connectors/source/CosFile.md
+++ b/docs/zh/connectors/source/CosFile.md
@@ -202,6 +202,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/FtpFile.md
+++ b/docs/zh/connectors/source/FtpFile.md
@@ -286,6 +286,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/HdfsFile.md
+++ b/docs/zh/connectors/source/HdfsFile.md
@@ -132,6 +132,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/LocalFile.md
+++ b/docs/zh/connectors/source/LocalFile.md
@@ -219,6 +219,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/ObsFile.md
+++ b/docs/zh/connectors/source/ObsFile.md
@@ -109,6 +109,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/OssFile.md
+++ b/docs/zh/connectors/source/OssFile.md
@@ -273,6 +273,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/OssJindoFile.md
+++ b/docs/zh/connectors/source/OssJindoFile.md
@@ -116,6 +116,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/S3File.md
+++ b/docs/zh/connectors/source/S3File.md
@@ -406,6 +406,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/docs/zh/connectors/source/SftpFile.md
+++ b/docs/zh/connectors/source/SftpFile.md
@@ -273,6 +273,8 @@ markdown 解析器提取各种元素，包括标题、段落、列表、代码�
 - `chunk_index`：解析后文档中的一基 chunk 顺序
 - `content_hash`：已输出 `text` 值的 SHA-256 哈希
 
+启用该选项并读取有界 Markdown 文件时，source enumerator 会使用相同的 `document_id` 哈希分配整文件 split，使同一文档派生的所有行留在同一个 source 路由 bucket 中。禁用该选项时，默认的轮询 split 分配行为保持不变。
+
 该选项默认值为 `false`，因此只有显式启用后才会改变原始 Markdown schema。
 
 注意：Markdown 格式仅支持读取，不支持写入。
--- a/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/BaseFileSource.java
+++ b/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/BaseFileSource.java
@@ -54,6 +54,7 @@ public abstract class BaseFileSource
     private final CatalogTable catalogTable;
     private final List<String> filePaths;
     private final ReadStrategy readStrategy;
+    private final boolean documentRoutingEnabled;
 
     /** shouldn't use this construct method. just for testing */
     @VisibleForTesting
@@ -61,13 +62,17 @@ public abstract class BaseFileSource
         this.catalogTable = null;
         this.filePaths = null;
         this.readStrategy = null;
+        this.documentRoutingEnabled = false;
     }
 
     public BaseFileSource(ReadonlyConfig pluginConfig) {
         this.pluginConfig = pluginConfig;
         HadoopConf hadoopConf = initHadoopConf();
-        this.readStrategy =
-                pluginConfig.get(FileBaseSourceOptions.FILE_FORMAT_TYPE).getReadStrategy();
+        FileFormat fileFormat = pluginConfig.get(FileBaseSourceOptions.FILE_FORMAT_TYPE);
+        this.readStrategy = fileFormat.getReadStrategy();
+        this.documentRoutingEnabled =
+                fileFormat == FileFormat.MARKDOWN
+                        && pluginConfig.get(FileBaseSourceOptions.MARKDOWN_RAG_METADATA_ENABLED);
         this.readStrategy.setPluginConfig(pluginConfig.toConfig());
         this.readStrategy.init(hadoopConf);
         String path = pluginConfig.get(FileBaseSourceOptions.FILE_PATH);
@@ -81,7 +86,6 @@ public abstract class BaseFileSource
 
         // support user-defined schema
         CatalogTable userDefinedCatalogTable;
-        FileFormat fileFormat = pluginConfig.get(FileBaseSourceOptions.FILE_FORMAT_TYPE);
         // only json text csv type support user-defined schema now
         if (pluginConfig.getOptional(ConnectorCommonOptions.SCHEMA).isPresent()) {
             switch (fileFormat) {
@@ -156,7 +160,7 @@ public abstract class BaseFileSource
     @Override
     public SourceSplitEnumerator<FileSourceSplit, FileSourceState> createEnumerator(
             SourceSplitEnumerator.Context<FileSourceSplit> enumeratorContext) throws Exception {
-        return new FileSourceSplitEnumerator(enumeratorContext, filePaths);
+        return new FileSourceSplitEnumerator(enumeratorContext, filePaths, documentRoutingEnabled);
     }
 
     @Override
@@ -164,6 +168,7 @@ public abstract class BaseFileSource
             SourceSplitEnumerator.Context<FileSourceSplit> enumeratorContext,
             FileSourceState checkpointState)
             throws Exception {
-        return new FileSourceSplitEnumerator(enumeratorContext, filePaths, checkpointState);
+        return new FileSourceSplitEnumerator(
+                enumeratorContext, filePaths, checkpointState, documentRoutingEnabled);
     }
 }
--- a/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/BaseMultipleTableFileSource.java
+++ b/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/BaseMultipleTableFileSource.java
@@ -30,6 +30,7 @@ import org.apache.seatunnel.connectors.seatunnel.file.config.BaseFileSourceConfi
 import org.apache.seatunnel.connectors.seatunnel.file.config.BaseMultipleTableFileSourceConfig;
 import org.apache.seatunnel.connectors.seatunnel.file.config.FileBaseSourceOptions;
 import org.apache.seatunnel.connectors.seatunnel.file.config.FileDiscoveryMode;
+import org.apache.seatunnel.connectors.seatunnel.file.config.FileFormat;
 import org.apache.seatunnel.connectors.seatunnel.file.config.FilePostSyncAction;
 import org.apache.seatunnel.connectors.seatunnel.file.exception.FileConnectorException;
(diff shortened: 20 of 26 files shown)


DanielLeens (commented):
Thanks for working on this. I re-reviewed the latest head from scratch against the actual Zeta runtime path, and the overall partition-routing direction makes sense. The blocker I still see is in the schema-evolution path: once a sink declares a partition strategy, `SeaTunnelRow` records can be routed to remote sink writers, but `SchemaChangeEvent` is still consumed only by the local writer.

Runtime chain I checked:

```text
Source / transform emits SchemaChangeEvent
  -> SeaTunnelSourceCollector.collect(event)
      -> sendRecordToNext(new Record<>(event))
  -> SinkPartitionExchange.received(record)
      -> if Barrier: broadcast to all sink writers
      -> if SchemaChangeEvent: drain batches, then consumeLocally(record) only

Subsequent SeaTunnelRow records
  -> SinkPartitionRouter.route(record)
      -> hash partition fields to a remote sink writer
  -> dispatcher.dispatch(batch)
  -> remote writer handles the row in SinkFlowLifeCycle.handleRecord()
      -> writer.write(row)
      -> but that writer may never have received applySchemaChange(event)
```

The key paths are:
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/SeaTunnelSourceCollector.java:119-133`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/flow/SinkPartitionExchange.java:104-129`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/flow/SinkPartitionRouter.java:64-80`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/flow/SinkFlowLifeCycle.java:327-343`

Issue 1: `SchemaChangeEvent` is not broadcast to all participating sink writers
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/flow/SinkPartitionExchange.java:112`
- Why this is a real blocker:
  the new exchange layer correctly broadcasts barriers, but schema changes stay local. After that, normal data rows may still hash to a remote writer. That remote writer then writes post-DDL data before it has applied the schema change.
- Risk:
  this can break the main write path for any sink that combines partition routing with schema evolution.
- Better fix:
  either broadcast `SchemaChangeEvent` to every `partitionTargetTask`, or explicitly reject the combination of partition strategy + schema evolution until the runtime support is complete.
- Severity: High

Issue 2: the new tests still miss the schema-change + remote-routing regression case
- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/task/flow/SinkPartitionExchangeTest.java:44`
- Why it matters:
  the current tests cover same-key routing and barrier alignment, but not the failing path above, so Issue 1 can slip through with all tests green.
- Severity: Medium

Compatibility:
- This is backward compatible for sinks that do not declare a partition strategy.
- For sinks that do declare one, the current runtime contract is incomplete once schema evolution is involved.

Performance / side effects:
- The extra hashing and writer-to-writer dispatch cost looks acceptable.
- The bigger concern here is correctness, not CPU or memory.

Test stability:
- The newly added UTs are structurally stable: they are in-memory tests without `Thread.sleep`, fixed ports, or container timing races.
- The gap is coverage, not flakiness.

CI:
- I did not run this locally.
- The current `Build` check was still queued when I reviewed, but Issue 1 is already a source-level merge blocker independent of CI.

1. Blocking items
- Issue 1 must be fixed first, because the current schema-evolution path is incorrect once rows are routed to remote sink writers.

2. Suggested follow-up
- Add the missing regression test from Issue 2 so this runtime contract stays protected.

Overall, the feature direction is good and the planner/runtime split is clean. The one thing we should not merge as-is is the local-only schema-change handling.

nzw921rx:
This seems to conflict somewhat with STIP‑23. I think we should first come up with a design document. The backpressure PR also needs to be taken into consideration

DanielLeens:
Thanks for raising the STIP-23 angle. I agree this needs to line up with the broader partition-routing and backpressure design rather than being patched in isolation. From Daniel's side, the current blocker from the latest review is still the same runtime gap: `SchemaChangeEvent` is still consumed only locally while partitioned `SeaTunnelRow` records can be routed to remote sink writers. Since there is no new commit after that review yet, I am not starting another full re-review on this exact revision in this round. Once there is either a design write-up or a follow-up commit that closes the runtime path, I am happy to take another pass.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch against the real Zeta sink routing path.

# What this PR solves
- User pain: today sink partitioning is mostly local to a writer task, so users cannot declare connector-level partition routing that redistributes rows across sink writers.
- Fix approach: add planner/runtime support for `SinkPartitionStrategy`, remote writer exchange, and translator guards for unsupported engines.
- One-line summary: the previous schema-change blocker is fixed on the latest head, and my remaining merge concern is CI rather than a new source-level correctness issue.

# Runtime path I checked
```text
upstream row / schema event / barrier
  -> SinkFlowLifeCycle.received(...)
      -> SinkPartitionExchange.received(...)
          -> data row: route to target writer
          -> schema change: drain batches, broadcast to all target writers
          -> barrier: drain batches, broadcast, align once per target writer
  -> remote target writer
      -> receivedSinkPartitionExchange(...)
      -> handleRecord(...)
```

# Re-review result
- The earlier blocker is fixed: `SchemaChangeEvent` is now broadcast through `createBroadcastEnvelopes(...)`, and the new `SinkPartitionExchangeTest.schemaChangeIsBroadcastBeforeRemoteData()` covers the exact runtime path that was missing before.
- I did not find a new source-level blocker in the current partition-routing logic.

# CI
- The current GitHub `Build` check is still red for this PR.
- From the accessible metadata here I can confirm the red status, but I could not retrieve a detailed failing job log from the fork run via the available GitHub API surface in this environment, so I do not want to invent a root cause.

1. Blocking items
- Please bring the current `Build` back to green, or provide the failing job details if the red status is from unrelated fork/infra noise.

- None more important than closing the CI loop now that the source-level schema-change gap is fixed.

yzeng1618:
> This seems to conflict somewhat with STIP‑23. I think we should first come up with a design document. The backpressure PR also needs to be taken into consideration

I rechecked this with STIP-23 in mind.For this PR, hash routing is limited to data rows. Barriers and schema changes are treated as control events, so they are flushed/broadcast instead of being routed by hash. I also updated the design notes to make this explicit. The backpressure/timer signal part is not handled here and should remain in the backpressure PR.

DanielLeens:
Thanks for circling back with the STIP-23 clarification and the updated design note.

That explanation lines up with the latest Daniel re-review on head `5a789c128a387726aedc9972bc7aa37b217d3cbb`: treating barriers and schema changes as control events, rather than hashing them like data rows, is exactly the distinction I needed to see called out explicitly.

Since there is still no new commit after that latest Daniel review, I am keeping this as a reply-only follow-up on the unchanged head instead of reopening a full source review. On my side the remaining merge gate is still CI rather than a new runtime-path blocker.

nzw921rx:
Apologies, my earlier comments may not have been clear enough. Let me reorganize my thoughts here.

I've identified a few points in the current approach that I think are worth discussing:

- Architectural consistency

   The `sink —> sink` point-to-point shuffle effectively introduces an additional data channel outside the engine's existing data exchange, checkpoint alignment, and back-pressure paths, which may not fully align with Zeta's current unidirectional DAG architecture. Additionally, this approach may have limited reusability — if future data flows need to introduce scenarios such as primary-key-based routing, it could be challenging to extend from this foundation.

- Cluster stability risk

   The current hash routing lacks a back-pressure feedback loop, which could make it easier for hot keys to overload a single writer. Combined with the synchronous blocking RPC (the dispatch side's `.get()` blocks the task thread), local skew could potentially escalate into a broader stall. On the receiving side, although `SinkPartitionExchangeOperation` obtains the target task through `TaskExecutionService`, the actual write executes synchronously on the HZ operation thread — if downstream responses slow down, it could occupy HZ operation threads. Worth noting is that this overhead scales as a product of the number of jobs and parallelism. When the cluster runs many concurrent jobs (e.g. 100), HZ operation threads can become heavily saturated. Since this thread pool is shared across the entire cluster (heartbeats, checkpoints, cluster management all depend on it), if heavily saturated, there could be a risk of heartbeat timeouts, false node-down detection, and cascading failover.

- Failover data loss

   Because the `sink -> sink` point-to-point channel is outside the engine's checkpoint scope, data that has been sent but not yet persisted could potentially be lost during failover, and the existing recovery mechanism may not be able to replay it. If exactly-once or at-least-once recovery needs to be supported in the future, the room for retrofitting might be quite limited.

- A thought on direction

   I'm personally leaning towards not adding a point-to-point channel on the Sink side, and perhaps holding off on introducing a standalone exchange operator at this stage. A direction worth considering is to have `SourceSplitEnumerator` handle routing and split strategies on the Source side, which would align better with the engine's existing mechanisms.

Based on the above considerations, my suggestion is to produce a detailed design first, covering at least the following:
- The complete `document_id` data flow (assignment, routing, delivery, landing), including the addressing relationship between each parallelism instance and vector database partitions
- Checkpoint alignment can be considered as a follow-up iteration (state boundaries, snapshot timing, recovery consistency)

The above is solely my personal opinion and I welcome any corrections @yzeng1618 @davidzollo @zhangshenghang .

DanielLeens:
Thanks for reorganizing the concerns here, and I appreciate the level of detail. These are thoughtful architecture questions, especially around how a Sink->Sink side channel relates to checkpoint alignment, backpressure propagation, and shared HZ operation threads.

I want to be careful not to pretend a brand-new full Daniel re-review happened on the same unchanged head, though. The current PR head is still `5a789c128a387726aedc9972bc7aa37b217d3cbb`, which is the same revision I already re-reviewed on May 31. So my latest Daniel full-review conclusion on that exact head is still the same one from that review, rather than a newly expanded blocker list on the same code version.

That said, I do think your discussion points are important input for the next revision. If the author decides to address them with either a design write-up or a code update, please push that as a new commit and I will re-review the latest head from scratch again against the full Zeta runtime path. Until then, from Daniel's side this stays in reply-only / awaiting-update state rather than a new full review on the unchanged head.

yzeng1618:
I re-checked the current implementation against these concerns.

The current SinkPartitionExchange path does introduce a sink-side point-to-point channel: remote delivery goes through Hazelcast task operations, the dispatch side waits on the operation future, and the receive side can synchronously enter the target SinkWriter path from the operation call stack. The exchange also has local buffering/alignment state that is not part of checkpoint state. So I agree this should not be treated as the production Knowledge Sync routing direction without a fuller design.

I have drafted a design document first, focusing on the complete document_id flow: assignment, routing, delivery, landing, and the relationship between route buckets, parallel subtasks, and vector-store partitions/logical buckets. The proposed direction is to move the first production routing path toward SourceSplitEnumerator/source-side document assignment, and defer both the sink-side point-to-point channel and a standalone keyed exchange operator until checkpoint, back-pressure, and recovery semantics are designed as first-class engine behavior.

Checkpoint alignment and recovery consistency are called out as a follow-up iteration with explicit state boundaries.

yzeng1618:
## Background

Knowledge Sync requires document-scoped lifecycle decisions. A vector lifecycle sink needs to read the old chunks for one document, compare hashes, delete stale chunks, upsert changed chunks, and skip unchanged chunks. If chunks for the same `document_id` are processed by multiple sink writers, those writers can make conflicting refresh/delete decisions.

The current implementation direction added a sink-declared partition strategy and a Zeta runtime component named `SinkPartitionExchange`. The mechanism is useful as a focused experiment, but the current runtime path introduces a sink-to-sink point-to-point data channel:

```text
upstream flow -> local sink writer -> SinkPartitionExchange -> remote sink writer
```

That path is outside the normal Zeta source/transform/sink data exchange boundary. It uses Hazelcast task operations for cross-task delivery, blocks the sending task while waiting for remote dispatch completion, and executes the receiving write path synchronously inside the Hazelcast operation call stack. It also does not snapshot or restore exchange-local buffers and sequence state.

This design proposes a safer first production direction: keep document routing on the source side through `SourceSplitEnumerator` and normal engine dataflow, and defer a standalone engine exchange operator until its checkpoint and back-pressure semantics are designed as first-class runtime behavior.

## Goals

- Define the complete `document_id` data flow: assignment, routing, delivery, and landing.
- Keep Knowledge Sync routing aligned with existing source split assignment, task execution, checkpoint, and back-pressure mechanisms.
- Ensure one document belongs to one lifecycle decision domain.
- Make the addressing relationship explicit between document routing buckets, source readers, sink writers, and vector database partitions or logical buckets.
- Treat checkpoint alignment and recovery as a follow-up design area with clear state boundaries from the start.

## Non-Goals

- Do not extend the current sink-side point-to-point exchange for production lifecycle sync in this design.
- Do not introduce a standalone keyed exchange operator in the first iteration.
- Do not implement Qdrant or Milvus lifecycle behavior here.
- Do not promise exactly-once vector-store visibility until checkpoint-driven sink lifecycle semantics are designed and verified.
- Do not require every existing source to support document-level routing. Sources that cannot produce complete document splits should fail fast for lifecycle sync.

## Current Sink-Side Exchange Assessment

The current sink-side exchange has already addressed some local ordering issues: checkpoint barriers and schema-change events are flushed and broadcast instead of being hashed as ordinary data records. This is important, but it does not make the exchange equivalent to an engine-native data channel.

Observed risks:

- **Architecture drift**: routing happens inside the sink lifecycle after data has already reached a sink task. This creates an extra data plane that is not represented as a normal Zeta DAG edge.
- **Shared operation pressure**: remote delivery uses Hazelcast operations. The same operation service is also used by worker heartbeats, checkpoint ACKs, task status, and cluster management paths.
- **Blocking dispatch**: remote dispatch waits for the operation future. A slow target writer can block the sending task thread.
- **Synchronous receive path**: the operation handler calls the target task and can reach `SinkWriter.write(...)` synchronously in the operation thread.
- **Recovery gap**: exchange-local pending batches, per-target sequence numbers, and barrier alignment bookkeeping are not part of checkpoint state.
- **Limited reuse**: future keyed routing scenarios, such as primary-key based routing, would need a more general engine data exchange model rather than a sink-specific transport.

Therefore this design treats sink-side point-to-point routing as unsuitable for the first production Knowledge Sync routing path.

## Proposed Direction

Use source-side document routing controlled by `SourceSplitEnumerator`.

The core rule is:

> A `document_id` is assigned to exactly one route bucket, and all rows derived > from that document are emitted by the reader that owns that bucket.

For the first lifecycle implementation, the job should use a pointwise topology where the source reader subtask, transform subtask chain, and sink writer subtask for the same route bucket stay aligned. If the planner cannot guarantee this alignment, the lifecycle sink must fail fast instead of falling back to best-effort routing.

```text
SourceSplitEnumerator
        |
        | document_id -> route bucket
        v
SourceReader[route bucket]
        |
        | normal Zeta dataflow
        v
Transform chain[route bucket]
        |
        | normal Zeta dataflow
        v
Lifecycle SinkWriter[route bucket]
        |
        v
Vector database partition or logical bucket
```

This keeps routing in the same place where SeaTunnel already reasons about parallelism, split assignment, checkpointed source state, and failover.

## Routing Model

### Routing Key

The primary routing key is the physical field `document_id`.

The `document_id` must be stable across job restarts for the same logical document. For file-backed documents, a recommended default is:

```text
document_id = "doc_" + sha256(canonical_source_uri)
```

Connectors may provide stronger source-native identities when available, such as database primary keys, object version IDs, or external document IDs. The selected identity must be included in the source/enumerator checkpoint state when it affects pending work assignment.

### Route Bucket

The route bucket is deterministic:

```text
route_bucket = floorMod(stableHash(document_id), route_parallelism)
```

`route_parallelism` is fixed when the job is planned and must be persisted or reconstructible during recovery. Changing route parallelism across restore is an incompatible state migration unless a separate rebalance protocol is introduced.

### Assignment

The enumerator discovers documents and assigns each document-level split to the reader that owns the route bucket.

```text
document split:
  source_uri
  source_version
  document_id
  document_hash candidate
  route_bucket
  delete marker, if the document disappeared
```

The split must represent a whole document or a whole document lifecycle event. Chunk-level split assignment is not allowed in lifecycle mode because it can split one document across multiple decision domains.

## Complete Data Flow

### 1. Discovery

The source enumerator discovers source objects, files, or source records. It normalizes the source address and resolves the document identity.

Required output from discovery:

- `source_uri`
- `document_id`
- optional `source_version`
- optional source-native modified timestamp
- delete marker for removed documents when the source supports deletion

### 2. Routing Assignment

The enumerator computes `route_bucket` from `document_id` and assigns the document split to the reader mapped to that bucket.

If no reader is registered for a bucket, the split remains pending in enumerator state until the reader is available. On failover, pending and assigned-but-not checkpointed splits are reconstructed from the latest checkpoint and reassigned using the same deterministic rule.

### 3. Source Delivery

The source reader reads the whole document split and emits one or more rows. Every emitted row for that document must contain the physical routing and lifecycle fields required downstream:

- `document_id`
- `document_hash`
- `source_uri`
- `source_version`, if available
- `deleted`
- `chunk_id`
- `chunk_hash`
- `chunk_index`
- content fields used by parse/chunk/embedding

For a delete event, the reader may emit a compact tombstone row with `deleted = true`, `document_id`, and enough target context for the lifecycle sink to delete all existing chunks for that document.

### 4. Transform Delivery

Transforms must preserve the physical routing fields. Document parse, chunking, metadata projection, and embedding transforms may add fields, but they must not drop or rename `document_id` unless the job explicitly maps it to another physical field before lifecycle routing is enabled.

If a transform can expand one document into many rows, all expanded rows remain inside the same route bucket because expansion happens after the source reader has been assigned the document split.

### 5. Sink Landing

The lifecycle sink writer receives all rows for one document through the route bucket assigned to its subtask. The writer performs document-scoped lifecycle logic:

```text
read old chunks for document_id
compare chunk_hash and document_hash
delete stale chunks
upsert changed or new chunks
skip unchanged chunks
```

The sink writer must use idempotent target identifiers:

```text
point_id = chunk_id
payload.document_id = document_id
payload.chunk_hash = chunk_hash
```

For target systems with physical partitions, such as Milvus, the writer may map the route bucket to a configured partition:

```text
vector_partition = partition_prefix + "_" + route_bucket
```

For target systems without a matching physical partition concept, such as a single Qdrant collection, the route bucket is a SeaTunnel ownership boundary and `document_id` remains a payload filter/delete key. In both cases, the addressing contract is the same from SeaTunnel's perspective.

## Addressing Relationship

Example with `route_parallelism = 4`:

| Document | Route bucket | Source reader | Transform chain | Sink writer | Vector target |
| --- | --- | --- | --- | --- | --- |
| `doc_A` | `hash(doc_A) % 4 = 1` | reader-1 | subtask-1 | writer-1 | partition/logical-bucket-1 |
| `doc_B` | `hash(doc_B) % 4 = 3` | reader-3 | subtask-3 | writer-3 | partition/logical-bucket-3 |
| `doc_C` | `hash(doc_C) % 4 = 1` | reader-1 | subtask-1 | writer-1 | partition/logical-bucket-1 |

The invariant is not that each bucket has only one document. The invariant is that each document has only one bucket.

## Back-Pressure and Hot Keys

This design relies on the existing dataflow back-pressure path instead of adding Hazelcast data operations from sink writer to sink writer. A slow lifecycle sink writer slows its normal upstream chain.

Hot keys are still possible. A single large document or a small number of very large documents can overload one bucket. The first production version should handle this as an explicit lifecycle trade-off:

- do not split one document across buckets in lifecycle mode;
- expose per-bucket document count, row count, bytes, and lifecycle latency metrics;
- optionally fail fast or warn when one document exceeds configured size or chunk-count thresholds;
- allow users to increase route parallelism before the job starts.

Dynamic hot-key migration is out of scope because moving a live document between writers would require a checkpointed rebalance protocol.

## Checkpoint and Recovery Boundaries

Checkpoint alignment can be implemented in a follow-up iteration, but the state boundaries should be fixed now.

### Source Enumerator State

The enumerator owns:

- discovered but unassigned document splits;
- assigned splits that have not been checkpoint-confirmed;
- route parallelism and route bucket mapping inputs;
- source version cursors or listing progress;
- delete markers that still need delivery.

### Source Reader State

Each reader owns:

- currently processing document split;
- read offset inside the document, if the source supports resumable reads;
- emitted chunk progress if parsing/chunking happens inside the reader.

If the reader cannot resume inside a document, it may replay the whole document from the last checkpoint. The sink lifecycle path must therefore be idempotent by `document_id` and `chunk_id`.

### Sink Writer State

The sink writer owns:

- prepared but not yet committed lifecycle operations, if the target sink uses a checkpoint-gated commit model;
- target transaction handles or staging metadata, if supported;
- retryable commit/delete/upsert intent records when external operations are delayed until checkpoint completion.

Visible external side effects before checkpoint completion must be either idempotent or staged. A sink that cannot satisfy this must document at-least-once visibility and pass tests for duplicate replay.

### Recovery Scenarios

- **Failure before checkpoint completion**: source state rolls back to the last completed checkpoint; document splits may be replayed; sink writes must be idempotent or staged.
- **Failure after checkpoint completion but before cleanup**: sink commit or cleanup may be retried; commit/delete/upsert operations must tolerate repeated attempts.
- **Reader failure during a large document**: the document may restart from the last checkpointed reader offset, or from the beginning if no offset is available.
- **Parallelism change on restore**: not supported in the first version for lifecycle routing because it changes route bucket ownership.

## Planner Requirements

For a lifecycle sink that requires document routing, the planner should validate:

- `document_id` exists as a physical field before lifecycle sink input.
- Source supports document-level split assignment or declares that it cannot run lifecycle routing.
- Source, transform, and sink route parallelism are aligned for the first implementation.
- No transform between source and lifecycle sink performs repartitioning or drops routing fields.
- Non-Zeta engines either preserve source split assignment to sink writer ownership or fail fast with a clear error.

## Alternatives Considered

### A. Keep Sink-Side Point-to-Point Exchange

This preserves the current implementation direction and can route rows even when source and sink parallelism differ. It is not recommended for production Knowledge Sync because it adds a sink-local data channel outside the normal DAG, uses Hazelcast operations for data transfer, and lacks exchange-level checkpoint state.

### B. Add a Standalone Engine Keyed Exchange

This is the most general long-term engine solution. It could support `document_id`, primary keys, and other keyed flows. It should be designed as a first-class Zeta operator with explicit queueing, checkpoint, back-pressure, metrics, and failover semantics. It is deferred because introducing it now would expand the scope beyond the immediate Knowledge Sync design.

### C. Source-Side Document Assignment

This is the recommended first production path. It reuses the existing `SourceSplitEnumerator` model, keeps routing before row fan-out, avoids sink-to-sink data transfer, and gives recovery a natural source-state boundary. Its main limitation is that it requires source/sink route alignment and whole document splits.

## Implementation Slices

1. **ADR and contract update**
   - Freeze lifecycle routing as source/enumerator-led for the first production path.
   - Mark sink-side exchange as experimental or not used by lifecycle sinks.

2. **Document metadata contract**
   - Finalize `DocumentId`, `DocumentHash`, `ChunkId`, `ChunkHash`, and physical field mappings.
   - Ensure transforms preserve these fields.

3. **Document split assignment**
   - Add or extend source-side document split strategy for file/Markdown first.
   - Add deterministic bucket assignment in the enumerator.

4. **Lifecycle sink validation**
   - Validate required fields and aligned route parallelism.
   - Fail fast for unsupported sources or topologies.

5. **Qdrant lifecycle V1**
   - Use `document_id` and `chunk_id` for read/compare/delete/upsert/skip.
   - Treat route bucket as ownership boundary even if Qdrant uses one collection.

6. **Checkpoint follow-up**
   - Define checkpoint-gated lifecycle side effects.
   - Add failover and replay tests before claiming production recovery semantics.

## Verification Plan

- Unit test deterministic `document_id -> route_bucket` assignment.
- Unit test enumerator restore reassigns pending documents to the same bucket.
- Integration test same document chunks reach one lifecycle sink writer.
- Integration test different documents can use different writers.
- Delete-event test ensures tombstones route to the same bucket as the original document.
- Failover test replays from checkpoint without losing document lifecycle events.
- Duplicate replay test verifies idempotent `chunk_id` upsert/delete behavior.
- Skew test records per-bucket metrics and validates warning/fail-fast behavior for oversized documents.

@nzw921rx @zhangshenghang

DanielLeens:
Thanks for taking the time to write this up. The source-side document assignment direction is much clearer to me than continuing to harden the current sink-side point-to-point exchange as the first production path. In particular, spelling out the route-bucket ownership, checkpoint boundaries, and the reason to defer a standalone keyed exchange directly addresses the architecture concerns that were raised earlier.

Since the PR head is still the same `5a789c128a387726aedc9972bc7aa37b217d3cbb` revision that I already reviewed, I do not want to pretend a brand-new full Daniel re-review happened on unchanged code. So from my side this stays in reply-only / awaiting-update state rather than becoming a new full review on the current head.

If you decide to keep this PR as the implementation vehicle, the next step needs to be a code update that matches the design direction and makes the intended scope explicit in the PR itself. Once there is a new commit, I am happy to re-review the latest head from scratch against the full Zeta runtime path.

nzw921rx:
@yzeng1618 Thank you for your update. I think this is a feasible direction, and I have a small suggestion. It seems more reasonable to include this design in ISSUES.

DanielLeens:
Thanks, I agree that moving the design into an ISSUE is the cleaner way to carry the architecture discussion forward. From my side the PR head is still unchanged, so I am keeping the current Daniel state as reply-only rather than reopening a full review on the same revision. Once there is either an ISSUE or design link plus a code update, or a fresh commit that narrows the implementation scope, I am happy to re-review the latest head from scratch.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head again before commenting, and the current revision still is not reviewable as the feature named in the PR title.

What I rechecked:
```text
PR title / intended scope
  -> "Support sink partition strategy routing"
current Files changed surface
  -> file-source docs + markdown/document routing
  -> file source enumerators / read strategy additions
```

Issue 1: the current head no longer matches the PR's stated feature scope
- Location: the current `Files changed` surface is in file-source routing and markdown/document handling, for example `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/FileSourceDocumentRouting.java`, `.../FileSourceSplitEnumerator.java`, `.../MultipleTableFileSourceSplitEnumerator.java`, plus matching source-doc updates.
- Why this is a blocker:
  this PR is titled and discussed as a Zeta sink partition routing feature, but the current head is now changing an unrelated file-source/document-routing path. That means I cannot give a safe merge decision on the advertised feature because the branch content no longer matches the review target.
- Risk:
  merging a branch whose implementation scope no longer matches its PR title/discussion is very likely to confuse later reviewers and maintainers, and it breaks the "one focused problem per PR" rule.
- Better fix:
  rebuild the branch so it contains only the sink partition routing work described by this PR, or close this PR and reopen the file-source/document-routing work under the correct title/scope.
- Severity: High
- Already raised by others: Related to the earlier architecture-scope concerns; the current mismatch is even more direct.

1. Blocking items
- Issue 1: align the branch content with the PR's stated feature again.

2. Suggested follow-up
- Once the branch surface matches the feature under review, I can re-check the real runtime path for that feature instead of blocking on scope mismatch.

nzw921rx:
It seems that our original design intended to have a unified API on the engine side for hash-based routing. In our first version, we abandoned the sink → sink approach because it had compatibility issues with the existing engine and performance concerns. However, looking at the second version's design approach, it directly adds new configuration to the source connector-file — if enabled, it performs document hash-based routing. So if in the future, beyond document synchronization, other use cases want to reuse this hash routing mechanism, this approach appears to be a dead end.

Here are my questions:

1. Does this align with the scope/boundaries we defined for the Phase 1 MVP?

1. If the routing strategy needs to be extended to other connectors in the future, how should we handle it?

yzeng1618:
> It seems that our original design intended to have a unified API on the engine side for hash-based routing. In our first version, we abandoned the sink → sink approach because it had compatibility issues with the existing engine and performance concerns. However, looking at the second version's design approach, it directly adds new configuration to the source connector-file — if enabled, it performs document hash-based routing. So if in the future, beyond document synchronization, other use cases want to reuse this hash routing mechanism, this approach appears to be a dead end.
> 
> Here are my questions:
> 
> 1. Does this align with the scope/boundaries we defined for the Phase 1 MVP?
> 2. If the routing strategy needs to be extended to other connectors in the future, how should we handle it?

I see the concern. My intention for this PR is not to define the final generic hash-routing API for SeaTunnel. This PR is only the first concrete Phase 1 implementation slice for Knowledge Sync: bounded Markdown file sources derive a stable document_id and let the source enumerator assign whole-document splits to deterministic route buckets.

This matches the current Phase 1 boundary in #10889 / #10914: source-side document routing first, Markdown file source first, and no sink-to-sink exchange or standalone engine keyed exchange in this PR. The implementation is intentionally gated by Markdown + markdown_rag_metadata_enabled, so existing non-Markdown sources and Markdown sources without RAG metadata keep the current round-robin behavior.

For future connectors, I think we should not copy connector-file-specific code blindly. The extension path should be a follow-up common source-side document routing contract, for example:
1. a stable route key/document_id contract from the metadata work,
2. a source/enumerator capability for whole-document split assignment,
3. planner/lifecycle-sink validation that the source supports document-level routing and route alignment,
4. a later engine-level keyed exchange only if we need generic routing beyond source-side document assignment.

So I would treat this PR as the Markdown/file implementation of the Phase 1 source-side routing direction, not as the final reusable API. I can also make that scope explicit in the PR description/design issue if that helps avoid implying this is the long-term generic routing mechanism.

nzw921rx:
@yzeng1618 all right🚀, Thank you very much for your reply. Here's a small suggestion: Could we redefine the MVP boundary?

yzeng1618:
Thanks @nzw921rx 🚀 Agreed. Let me state the minimal Phase 1 boundary explicitly.
In scope
- Source-side document routing via SourceSplitEnumerator only.
- Markdown file source as the single reference implementation.
- Gated by markdown_rag_metadata_enabled: whole-document splits routed to deterministic buckets by stable document_id hash.

Out of scope
- Non-Markdown sources and Markdown without RAG metadata keep the existing round-robin behavior.
- No generic/engine-level keyed exchange, no sink → sink exchange, no other connectors, no checkpoint-gated lifecycle.

I'll record this boundary in #10889 / #10914 and mirror it in the PR description.

DanielLeens:
Thanks, this narrower MVP statement helps.

Framing this explicitly as "source-side document routing for Markdown only" is the right way to avoid over-claiming a generic engine-level hash-routing API in this PR. That also addresses @nzw921rx's scope question from my side: this can fit a Phase 1 slice as long as the PR description and the related design issues clearly say the reusable cross-connector contract is follow-up work, not something this PR is already trying to finalize.

At the same time, there is still no new commit after Daniel's June 13 full review on `1227aa36`, so I am not turning this discussion update into another full source review on the same code version. My current Daniel state stays reply-only on the unchanged head: if the PR framing is updated to match this narrowed MVP boundary, that improves the scope clarity; if the branch itself changes, I am happy to re-review the latest head from scratch again.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `1227aa3637e2` from scratch.

# What this PR fixes
- User pain: With `markdown_rag_metadata_enabled=true`, Markdown RAG rows from one document could be scattered across readers because bounded file splits were assigned round-robin, breaking per-document locality downstream.
- Fix approach: Introduce a shared `FileSourceDocumentRouting` helper (URI normalization + SHA-256 document id + bucket calc), reuse it in `MarkdownReadStrategy`, and have `BaseFileSource`/`BaseMultipleTableFileSource` pass a routing flag (or per-table-id set) into the split enumerators so bounded Markdown whole-file splits hash to a deterministic reader bucket, leaving round-robin untouched otherwise.
- One-line summary: Core routing logic and enumerator wiring look coherent and opt-in, docs are updated across all nine file connectors in EN and ZH, but I need to flag the constructor signature changes on `FileSourceSplitEnumerator`/`MultipleTableFileSourceSplitEnumerator` (public API surface, per the enumerator test churn) and verify bucket stability across parallelism changes on restore before approving.

# Runtime chain I rechecked
```text
BaseFileSource(ReadonlyConfig) computes documentRoutingEnabled = (fileFormat == FileFormat.MARKDOWN && MARKDOWN_RAG_METADATA_ENABLED)  BaseFileSource.java:70-75
  -> BaseFileSource.createEnumerator() -> new FileSourceSplitEnumerator(ctx, filePaths, documentRoutingEnabled)  BaseFileSource.java:163
  -> BaseFileSource.restoreEnumerator() -> new FileSourceSplitEnumerator(ctx, filePaths, checkpointState, documentRoutingEnabled)  BaseFileSource.java:170-171
     -> FileSourceSplitEnumerator split assignment: when routing enabled, derive documentId from split path and pick reader via FileSourceDocumentRouting.routeBucket(documentId, parallelism); otherwise fall back to round-robin  FileSourceSplitEnumerator.java
        -> FileSourceDocumentRouting: normalizeSourceUri() -> sha256(documentId) -> non-negative bucket = hash mod parallelism  FileSourceDocumentRouting.java
BaseMultipleTableFileSource.createEnumerator() -> resolveDocumentRoutingTableIds() (MARKDOWN + flag per table config, keyed by TablePath string)  BaseMultipleTableFileSource.java:118-122, 143-155
  -> new MultipleTableFileSourceSplitEnumerator(ctx, config, fileSplitStrategy, documentRoutingTableIds[, checkpointState])  BaseMultipleTableFileSource.java:118-122, 135-141
MarkdownReadStrategy reuses FileSourceDocumentRouting for document_id so reader-side metadata and enumerator-side bucket derive from the identical hash  MarkdownReadStrategy.java
```

# Findings
**Issue 1: Possible public API change in seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/source/split/FileSourceSplitEnumeratorTest.java**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/source/split/FileSourceSplitEnumeratorTest.java`
- Severity: High

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/source/split/FileSourceSplitEnumeratorTest.java

DanielLeens:
Thanks @SEZ9. I agree the current head still needs follow-up on the enumerator API-surface / restore-stability questions before merge. The routing direction itself still looks right to me, but the constructor-signature exposure and bucket-stability guarantees are exactly the places that need to be made explicit before we call this ready.

nzw921rx (approved):
+1, good job. The current implementation looks solid, and I’m looking forward to the follow-up engine-level contract framework landing.

Non-blocking: the existing bounded file source recovery path may rediscover already completed splits after restore. This predates the current PR, so it is not a blocker, but the new restore test only proves stable routing—not duplicate-free recovery. Please track this separately with a focused recovery test.

DanielLeens:
Thanks @nzw921rx. I want to keep Daniel's current blocker state explicit here, because I still do not think the unchanged head is merge-ready yet.

I’m still aligned with the enumerator API-surface / restore-stability concerns that were already raised on this branch. So from Daniel's side I still need a code update before I can clear the review, even though the routing direction itself continues to look reasonable.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch, including the previous API-surface / restore-stability concerns and the current CI state. This version looks good to me now.

I also noticed @SEZ9's earlier change request and @nzw921rx's later approval. After checking the latest head, I think the main blocker from that earlier round is cleared: the previous public constructors are still retained, the new routing behavior is added through overloads/internal wiring, and the restored-enumerator test now verifies that document-route ownership remains stable for the same parallelism.

## What this PR fixes

With `markdown_rag_metadata_enabled=true`, all rows derived from one Markdown document need to stay in the same source route bucket so downstream Knowledge Sync logic can make document-scoped decisions consistently. The previous bounded file split assignment was round-robin, so different Markdown files/documents could be scattered without considering document identity.

## Runtime path I checked

```text
User enables markdown_rag_metadata_enabled=true
  -> BaseFileSource(ReadonlyConfig)
      -> documentRoutingEnabled = (fileFormat == MARKDOWN && flag)
      -> createEnumerator()/restoreEnumerator()
          -> new FileSourceSplitEnumerator(..., documentRoutingEnabled)

Split assignment
  -> FileSourceSplitEnumerator.open()
      -> filePaths become whole-file FileSourceSplit entries
  -> run()
      -> assignSplit(taskId)
          -> routing disabled: existing round-robin assignment
          -> routing enabled:
              -> verify the split is a whole-file split
              -> FileSourceDocumentRouting.buildDocumentId(split.getFilePath())
              -> FileSourceDocumentRouting.routeBucket(documentId, currentParallelism)
              -> assign the split only to the bucket owner

Markdown reading
  -> MarkdownReadStrategy.read(path, ...)
      -> normalizeSourceUri(path)
      -> buildDocumentId(sourceUri)
      -> append RAG metadata with the same document_id
      -> all rows from the same document carry the same document id used for routing

Multi-table source
  -> BaseMultipleTableFileSource.resolveDocumentRoutingTableIds()
      -> only MARKDOWN tables with markdown_rag_metadata_enabled=true use document routing
      -> other tables keep round-robin behavior
```

## Review notes

No blocking issues from my side.

The important behavior is scoped correctly:

- Default behavior is unchanged because `markdown_rag_metadata_enabled` still defaults to `false`.
- Non-Markdown sources and Markdown sources without RAG metadata keep round-robin split assignment.
- Document routing rejects non-whole-file splits, which is the right guard for this Phase 1 scope.
- The same helper is reused for reader-side `document_id` metadata and enumerator-side route bucket calculation, so metadata identity and routing identity stay aligned.

Compatibility also looks fine to me. The existing enumerator constructors are still present, and this does not change split/state serialization or any default option value.

Test stability rating: **stable**. The new tests are pure enumerator/context checks; I did not see sleeps, hard-coded ports, external service dependency, or order-sensitive async assertions.

## CI / merge gate

The latest Build is passing: https://github.com/apache/seatunnel/runs/90475822213

The branch is currently `diverged` from `dev` with `behind_by=3`, but because the current Build is green, I am only recording that as queue metadata and I am not asking for a sync-only retry from Daniel's side.

## Merge conclusion

### Conclusion: can merge

1. Blocking items
   - None from my side.

2. Non-blocking suggestions
   - +1 to @nzw921rx's note that duplicate-free bounded file recovery should be tracked separately. That predates this PR and should not block this focused Markdown document-routing slice.

Overall, this is now a nicely scoped Phase 1 implementation: source-side Markdown document routing only, with the broader engine-level routing contract left for follow-up work. Good job narrowing this down.

davidzollo (approved):
Good job
+1

PR #10965, opened 2026-05-27
[Bug] [UI] Fix NaN display in job detail Overview table when tablePaths does not match metrics keys

### LINKED ISSUE #10547: [Bug] [UI] The task home page does not display the rate indicators, but they are clearly shown.
### Search before asking

### What happened

The task home page does not display the rate indicators, but they are clearly shown.

### SeaTunnel Version

2.3.13

### SeaTunnel Config

```conf
***
```

### Running Command

```shell
***
```

### Error Exception

```log
The log does not display error messages.
```

_No response_

### Java or Scala Version

_No response_

### Screenshots

Image

### Code of Conduct

77amyfly:

Fix #10547

Fix NaN display in the job detail Overview table. The `sourceCell` and `sinkCell` functions use `tablePaths` to look up values in metrics, but the `tablePaths` format does not match the metrics keys returned by the backend. For example:
- `tablePaths` returns: `["fake"]`
- metrics key is: `"Source[0].fake"`

This mismatch causes `Number(undefined)` to be evaluated as `NaN`. Fixed by using `endsWith` to match `tablePaths` against metrics keys instead of direct lookup.

Yes. The job detail Overview table now correctly displays metrics values (Received Bytes, Write Bytes, Received Count, Write Count, etc.) instead of NaN when `tablePaths` does not match metrics keys.
before fix:
Screenshot 2026-05-27 at 12 35 00 PM
after fix:
Screenshot 2026-05-27 at 12 35 53 PM

1. Added unit test in `seatunnel-engine-ui/src/tests/detail.spec.ts` to verify that no NaN is displayed when `tablePaths` does not match metrics keys.
2. All existing unit tests pass after the fix (6 test files, 12 tests passed).

### Check list
* [ ] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature.

### DIFF
--- a/seatunnel-engine/seatunnel-engine-ui/src/service/job/index.ts
+++ b/seatunnel-engine/seatunnel-engine-ui/src/service/job/index.ts
@@ -16,7 +16,7 @@
  */
 
 import { get } from '@/service/service'
-import type {Job, JobPage} from './types'
+import type { Job, JobPage } from './types'
 
 export const getRunningJobs = (page: number, rows: number) => get<JobPage>('/running-jobs', {page: page, rows: rows})
 export const getFinishedJobs = (page: number, rows: number) => get<JobPage>(`/finished-jobs`, {page: page, rows: rows})
--- a/seatunnel-engine/seatunnel-engine-ui/src/tests/detail-metrics.spec.ts
+++ b/seatunnel-engine/seatunnel-engine-ui/src/tests/detail-metrics.spec.ts
@@ -1,0 +1,137 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { describe, expect, test } from 'vitest'
+import {
+  collectVertexMetrics,
+  extractVertexIdentifier,
+  readVertexMetricValue
+} from '@/views/jobs/detail-metrics'
+import type { Vertex } from '@/service/job/types'
+
+describe('detail metrics helpers', () => {
+  const sourceVertex = {
+    vertexId: 1,
+    type: 'source',
+    vertexName: 'pipeline-1 [Source[0]]',
+    tablePaths: ['fake.user_table']
+  } as Vertex
+
+  const sinkVertex = {
+    vertexId: 2,
+    type: 'sink',
+    vertexName: 'pipeline-1 [Sink[1]]',
+    tablePaths: ['fake.user_table']
+  } as Vertex
+
+  test('extracts the vertex identifier from the vertex name', () => {
+    expect(extractVertexIdentifier(sourceVertex.vertexName)).toBe('Source[0]')
+    expect(extractVertexIdentifier(sinkVertex.vertexName)).toBe('Sink[1]')
+  })
+
+  test('prefers prefixed metric keys over raw table names', () => {
+    const metricMap = {
+      'Source[0].fake.user_table': '10',
+      'Source[1].fake.user_table': '20',
+      'fake.user_table': '999'
+    }
+    expect(readVertexMetricValue(metricMap, sourceVertex, 'fake.user_table')).toBe(10)
+  })
+
+  test('falls back to raw table names when prefixed keys are unavailable', () => {
+    const metricMap = { 'fake.user_table': '15' }
+    expect(readVertexMetricValue(metricMap, sourceVertex, 'fake.user_table')).toBe(15)
+  })
+
+  test('collects only metrics that belong to the focused vertex', () => {
+    const metricMap = {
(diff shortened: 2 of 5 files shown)


DanielLeens (changes requested):
Thanks for working on this. I went through the full current diff and traced the actual metric path from the engine backend into the UI table rendering.

# What this PR fixes
- User pain: the Job Detail Overview table can render `NaN` when the table path shown on a vertex does not exactly match the metric map key produced by the backend.
- Fix approach: the UI now falls back from exact lookup to suffix matching, so `fake` can match a backend key like `Source[0].fake`.
- One-line value: this removes the `NaN` symptom for the simple single-vertex case.

# Runtime path I checked
```text
Engine metric aggregation
  -> BaseService.processMetric(...) [BaseService.java:441-586]
      -> for single vertex: key can become "Source[0].table" / "Sink[0].table" [BaseService.java:463-472]
      -> for tagged array metrics: key becomes "identifier.table" [BaseService.java:494-523]

UI Overview table
  -> detail.tsx sourceCell/sinkCell [detail.tsx:94-128]
      -> iterate row.tablePaths
      -> find first metric key where key.endsWith(path) [detail.tsx:103-106, 121-124]
      -> sum that value into the displayed cell
```

# Review findings
Problem 1: Wrong metric attribution when multiple vertices share the same table path
- Location: `seatunnel-engine/seatunnel-engine-ui/src/views/jobs/detail.tsx:103-106`, `seatunnel-engine/seatunnel-engine-ui/src/views/jobs/detail.tsx:121-124`
- Severity: High
- Raised before by others: No
- Why this is a real bug:
  The backend intentionally emits per-vertex metric keys such as `Source[0].fake` and `Sink[0].fake` instead of the bare table path when there is vertex-level attribution on the same table [BaseService.java:463-472, 494-523]. The new UI code picks the *first* key whose suffix matches `fake`. That works only if there is exactly one matching vertex. As soon as two sources or two sinks share the same `tablePaths` entry, the first suffix match can belong to the wrong vertex, so the Overview table silently shows another vertex's numbers.
- Risk:
  This is not just a cosmetic mismatch. On normal multi-source or fan-out jobs, the per-row metrics in Overview become incorrect while still looking valid, which is worse than `NaN` because it hides the attribution error.
- Better fix:
  Match on the current vertex identifier as well, not only on the suffix. The UI already has `vertexName`, so it can derive `Source[0]` / `Sink[0]` and prefer `identifier + "." + path`. An even cleaner option would be to expose a backend map keyed by vertex id or vertex identifier directly so the UI does not need to reverse-engineer the key shape.

Problem 2: The new regression test only covers the single-match happy path
- Location: `seatunnel-engine/seatunnel-engine-ui/src/tests/detail.spec.ts:49-111`
- Severity: Medium
- Raised before by others: No
- Why this matters:
  The new test uses one source and one sink, both on `fake`, so there is only one possible suffix match for each metric group. That verifies the `NaN` symptom is gone, but it does not protect against the real ambiguity introduced by `endsWith(path)` when multiple vertices share the same table path.
- Risk:
  The current implementation can regress into wrong attribution without any test failure.
- Better fix:
  Add a regression case with at least two source vertices or two sink vertices sharing the same table path and assert each row reads its own metric bucket.

# Test stability
- Rating: Stable
- Why:
  The new frontend test is deterministic, uses mocked service data, and does not depend on timing, ports, or external state.

# Merge conclusion

Blocking items:
1. Problem 1

Non-blocking suggestions:
1. Problem 2

The current direction is good for the original `NaN` symptom, but I would not merge it until the per-vertex attribution stays correct on the real multi-vertex path as well.

nzw921rx:
+1 good job

DanielLeens:
Thanks for taking a look here as well.

On my side, the current blocker is still the same one from the latest Daniel review: on the current head `262278b65888f0da43900e2bdb86d8360f2ae928`, the `endsWith(path)` fallback can still attribute metrics to the wrong vertex when multiple sources or sinks share the same table path.

Since there is still no new commit after that reviewable head, I am keeping this as a reply-only follow-up rather than starting another full re-review in this batch. Once a code update lands for that attribution path, please ping me again and I will recheck the new head from scratch.

davidzollo (dismissed):
Good job.
+1 if CI passes

DanielLeens:
Thanks @davidzollo. I saw the approval, but on my side the blocker from the latest Daniel review still stands on the unchanged head `262278b65888f0da43900e2bdb86d8360f2ae928`: the `endsWith(path)` fallback can still misattribute metrics when multiple vertices share the same table path. Since there is still no new commit after that head, I am keeping this as a reply-only follow-up and waiting for a code update before re-reviewing from scratch.

DanielLeens (approved):
Thanks for working on this. I re-reviewed the latest head from the actual metric payload path into the UI rendering path.

What this PR fixes:
- User pain: the Job Detail Overview table could show `NaN` when the backend metric key and the vertex `tablePaths` were not exact string matches.
- Fix approach: the UI now resolves metric keys in a safer order: exact vertex-prefixed key first, raw table path second, and same-vertex suffix fallback last.
- One-line summary: this turns a brittle string lookup into a vertex-aware lookup, so the detail page shows numbers instead of `NaN`.

Runtime path checked:
`job.metrics.TableSource*/TableSink*` -> `detail.tsx` source/sink cell rendering -> `detail-metrics.ts` key resolution helpers -> focused vertex drawer metric collection.

I do not see a remaining code-path blocker on the current head. The helper now returns `0` instead of propagating `undefined`, and the new tests cover the important branches around prefixed keys, raw-key fallback, same-vertex suffix matching, and non-numeric values.

CI note: the current red `Build` check on the fork run is failing in broad jobs like `Dead links` and several unrelated integration suites, not on the changed UI path itself. From the code path in this PR, I do not see a UI-specific blocker left.

Conclusion: mergeable from Daniel's side.

davidzollo (approved):
+1
Good job

Issue #10969, opened 2026-05-28
[Bug] [Connector-V2] SQL Server CDC DDL history query has inconsistent LSN boundary semantics causing schema events to be lost


CloverDew:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/seatunnel/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened
The SQL Server CDC connector has an inconsistency between the LSN boundary semantics used in the main data reading window versus the DDL history query window, which can cause DDL events to be lost.

#### Background: 
In SqlServerStreamingChangeEventSource.java, the main data reading logic explicitly uses a closed interval (inclusive of fromLsn):

The code comments clearly state: "reading interval is inclusive", But the DDL history query introduced in https://github.com/apache/seatunnel/pull/10890 uses a left-open, right-closed interval with the condition:
```
ddl_lsn > ? AND ddl_lsn <= ?
```
This excludes DDL statements whose LSN is exactly equal to fromLsn, when a DDL statement's LSN equals exactly fromLsn, data rows with the new schema will be read due to closed interval semantics, the corresponding DDL schema event will be lost, downstream systems will process new data using the old schema.

#### Root Cause: 
The DDL query should use ddl_lsn >= ? AND ddl_lsn <= ? to match the closed interval semantics of the main reading window. The incrementLsn() logic can prevent duplicate processing.

#### Related PR: 
https://github.com/apache/seatunnel/pull/10890
#### Fix: 
Change the DDL history query lower bound from ddl_lsn > ? to ddl_lsn >= ? to align with the closed interval semantics of the main data reading window.

### SeaTunnel Version

--

### SeaTunnel Config

```conf
--
```

### Running Command

```shell
--
```

### Error Exception

```log
--
```

### Zeta or Flink or Spark Version

--

### Java or Scala Version

--

### Screenshots

--

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

nzw921rx:
@CloverDew Looking forward to his restoration🎆

DanielLeens:
Thanks for writing up the root cause so precisely.

I checked the current SQL Server CDC path against your description, and this does look like a real bug rather than just a theoretical edge case.

The key issue is the interval mismatch you pointed out:
[...]

PR #10890, opened 2026-05-15
[Feature][Connector-V2][SqlServer CDC] support sqlserver schema evolution

CloverDew:
### Purpose of this pull request

This PR implements schema evolution support for SQL Server CDC, enabling the SQL Server CDC connector to detect and propagate DDL changes.

The demand comes from: https://github.com/apache/seatunnel/issues/10282

### Key Components:
1. SqlServerSchemaChangeResolver, Core resolver that detects schema changes by comparing the previous catalog schema with Debezium's table change payload.
2. SqlServerStreamingChangeEventSource, Extends Debezium's streaming source.
3. Integration with SqlServerIncrementalSource, Implements SupportSchemaEvolution interface with supported types.
 
 Based on: https://github.com/apache/seatunnel/pull/10648 (review should start from commit 1187917e commit message: implements sqlserver cdc), please wait for the merge at https://github.com/apache/seatunnel/pull/10648, Otherwise, the test results of SqlServerCDCIT on the Flink engine will be unstable.

### How was this patch tested?

1. Unit Tests:

- SqlServerSchemaChangeResolverTest, Tests column diff logic, rename detection, table path parsing, and edge cases

- SqlServerSourceConfigFactoryTest, Tests configuration factory

- SqlServerDialectTest, Tests dialect behavior

2. E2E Tests:

- SqlServerCDCIT with dedicated schema change test cases: add column operations, drop column operations, alter column type changes, sp_rename operations

- Test configuration: sqlservercdc_to_sqlserver_with_schema_change.conf

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
[...]

PR #10970, opened 2026-05-28
[Fix][Connector-V2][SqlServer CDC] Fix SQL Server CDC DDL history query LSN boundary causing schema events to be lost

CloverDew:
This PR fixes a critical correctness issue in the SQL Server CDC connector where DDL schema events can be lost due to inconsistent LSN boundary semantics between the main data reading window and the DDL history query.

Related PR: https://github.com/apache/seatunnel/pull/10890
Related Issue: https://github.com/apache/seatunnel/issues/10969

**Problem**

The issue was introduced in PR #10890 which added DDL history querying functionality. The implementation created a semantic mismatch:

- Main data reading window: Uses a closed interval `[fromLsn, toLsn]`
- Explicitly documented in code comments as "reading interval is inclusive"
- `fromLsn` is set to `lastProcessedPosition.getCommitLsn()` for initial/recovery scenarios
- `incrementLsn()` is called in steady-state to advance position
- DDL history query before this fix: Used a left-open interval `(fromLsn, toLsn]`
 ```sql
 WHERE ddl_lsn > ? AND ddl_lsn <= ?
 ```

**Solution**

Change the DDL history query to use a closed interval matching the main reading window:

```diff
- WHERE ddl_lsn > ? AND ddl_lsn <= ?
+ WHERE ddl_lsn >= ? AND ddl_lsn <= ?
```

Yes, this is a minimal patch.

`SqlServerCDCIT`.`testWithSchemaEvolution` tests it.

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
+++ b/seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
@@ -136,7 +136,7 @@ public class SqlServerConnection extends JdbcConnection {
                     + " ddl_lsn,"
                     + " ddl_time"
                     + " FROM [#db].cdc.ddl_history"
-                    + " WHERE ddl_lsn > ? AND ddl_lsn <= ?"
+                    + " WHERE ddl_lsn >= ? AND ddl_lsn <= ?"
                     + " ORDER BY ddl_lsn ASC";
     private static final String OPENING_QUOTING_CHARACTER = "[";
     private static final String CLOSING_QUOTING_CHARACTER = "]";


nzw921rx:
+1 if ci pass

davidzollo (approved):
+1 if CI passes

DanielLeens (commented):
Thanks for working on this. I went through the full diff and traced the SQL Server CDC call chain end to end.

I also noticed @davidzollo already left an approval on this PR. After my own review I largely agree with the code change itself. The only thing I want to call out before merging is the CI state.

What this PR fixes
- User pain point: when a DDL record lands exactly on the lower LSN boundary of one polling window, the old query can skip it, so schema change events may be lost even though the data window is being processed.
- Fix approach: `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java:132-139` changes the lower bound in `GET_DDL_HISTORY` from `ddl_lsn > ?` to `ddl_lsn >= ?`.
- In one sentence: this is a precise boundary fix so the DDL-history query no longer drops schema changes sitting exactly at `fromLsn`.

Full runtime path I checked
```text
Incremental SQL Server CDC polling
  -> SqlServerStreamingChangeEventSource.executeIteration()
    -> compute fromLsn / toLsn
    -> getDdlEntriesToQuery(databaseName, fromLsn, toLsn)
      -> SqlServerConnection.getDdlHistory(databaseName, fromLsn, toLsn)
        -> execute GET_DDL_HISTORY SQL
    -> dispatchPendingDdlHistory(...)
```

Core review notes
- The normal path does hit this change: `executeIteration()` always queries DDL history for the current LSN window.
- The fix matches the surrounding semantics. On later iterations the code already advances the lower bound via `incrementLsn(...)`, so using `>=` here does not create a duplicate-consumption problem in the steady state.
- This is a precise fix, not a workaround.
- I did not find a code-level compatibility, concurrency, or resource-management regression in the modified path.

Compatibility
- Fully backward compatible.
- No API/config/default/protocol/serialization change.

Testing / CI
- `labeler` and `Notify test workflow` are green.
- The PR still shows `Build` as failed, so I would not merge until that is green again.
- I checked the current head SHA and could not correlate that red `Build` state with an active failing check-run on the latest head, so this looks more like a CI state/rerun issue than a problem introduced by the one-line SQL change. I would rerun `Build` first, and only dig deeper if it fails again.

Conclusion: merge after the CI state is cleared
- Blocking item: the `Build` check still needs to go green.
- Non-blocking suggestion: it would still be worth adding a regression test for the `ddl_lsn == fromLsn` boundary case in a follow-up, to keep this from coming back.

Happy to re-check once the CI status is green again.

yzeng1618 (approved):
LGTM

PR #10971, opened 2026-05-28
[Improve][e2e] Improve the stability of Kudu e2e testing e2e

nzw921rx:

Fix Kudu E2E flaky test (`KuduIT.java`) — multiple bugs causing intermittent failures:

- Fix `readData()` row boundary bug: `row` list was created at batch level, causing all rows within a scanner batch to be concatenated into one list; moved `row` creation inside the per-row loop
- Fix `readData()` BINARY column `break` bug: `break` on BINARY column skipped all subsequent columns; changed to `continue`
- Fix `readData()` timezone-sensitive timestamp formatting: replaced `Timestamp.toString()` (uses JVM local timezone) with explicit UTC `SimpleDateFormat` for `UNIXTIME_MICROS` columns, preventing 8-hour offset failures in non-UTC environments (e.g. Asia/Shanghai)
- Fix ordering assertion: Kudu hash-partition scan order is non-deterministic; added sort-by-id in `readData()` return value so `assertIterableEquals` comparisons are stable
- Fix unreliable table cleanup: wrapped all test methods with `try-finally` to guarantee `dropTableIfExists` runs even on mid-test failures; added `dropTableIfExists` helper that swallows "table not found" exceptions
- Fix `batchInsertData()` session resource leak: added `kuduSession.flush()` and `kuduSession.close()` in both overloads
- Fix `tearDown()` missing `toxiProxy.close()`: Docker container resource leak on test completion
- Fix `getKuduClient()` client leak on retry: close previous `kuduClient` before creating a new one during Awaitility retry loop

No.

```bash
./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kudu-e2e -DskipUT -DskipIT=false verify
```

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kudu-e2e/src/test/java/org/apache/seatunnel/e2e/connector/kudu/KuduIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kudu-e2e/src/test/java/org/apache/seatunnel/e2e/connector/kudu/KuduIT.java
@@ -63,10 +63,13 @@ import java.net.Inet4Address;
 import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
+import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


DanielLeens (approved):
Thanks for the follow-up here. I reviewed the full diff and walked through the Kudu E2E lifecycle from setup to teardown.

What this PR fixes
- User pain point: the Kudu E2E suite could still fail intermittently because of leftover tables, unclosed Kudu sessions, unstable row ordering, and timestamp formatting differences that depend on environment details.
- Fix approach: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kudu-e2e/src/test/java/org/apache/seatunnel/e2e/connector/kudu/KuduIT.java` now cleans tables proactively, flushes and closes `KuduSession`, normalizes timestamp output in UTC, sorts result rows before comparing, and closes the remaining external resources during teardown.
- In one sentence: this PR makes the Kudu E2E suite much more repeatable by cleaning state explicitly and removing nondeterministic result formatting/order.

Full runtime path I checked
```text
Kudu E2E
  -> dropTableIfExists(...)
  -> initializeKuduTable(...)
  -> batchInsertData() with flush/close
  -> execute SeaTunnel job
  -> readData(table)
     -> normalize binary/timestamp values
     -> sort rows by the first column
  -> compare source vs sink
  -> tearDown() closes kuduClient / tServers / master / toxiProxy
```

Core review notes
- The normal path definitely hits this change because all major Kudu E2E cases share these helpers.
- The fixes are targeted at real flaky sources: stale tables, incomplete session lifecycle, and nondeterministic row/timestamp rendering.
- I did not find a production-path regression because everything stays inside the E2E test layer.
- The cleanup additions and result normalization are consistent with the goal of making CI behavior repeatable.

Compatibility
- Fully backward compatible.
- No production API/config/protocol/serialization change.

Test stability assessment
- Rating: stable.
- `KuduIT.java:154-203` now flushes/closes sessions explicitly before assertions depend on the data.
- `KuduIT.java:328-345` and the individual test bodies clean tables before/after execution so state does not leak between runs.
- `KuduIT.java:592-624` normalizes timestamp rendering in UTC and sorts rows, which removes environment-specific ordering/timezone noise.
- `KuduIT.java:630-644` closes the remaining external resources more completely than before.

Testing / CI
- The current GitHub checks are green.

Conclusion: can merge
- Blocking items: none.

This is a good stabilization pass for the Kudu E2E suite. Approving from my side.

davidzollo (approved):
+1
Great job for CI

PR #10972, opened 2026-05-28
[Improve][e2e] Improve the stability of sftp e2e testing e2e

nzw921rx:

Fix SFTP E2E flaky test (`SftpFileIT.java`) — multiple issues across resource management, assertions, and timing causing intermittent CI failures.

**Resource issues:**
- Pin SFTP Docker image to `atmoz/sftp:alpine-3.7` (was `:latest`) to prevent cross-run version drift
- Add `Wait.forListeningPort()` WaitStrategy to ensure SSHD is ready before file operations; remove redundant double-start (`sftpContainer.start()` + `Startables.deepStart()`)
- Remove hardcoded host port binding `2222:22` to avoid port conflicts in parallel CI; rely on Testcontainers random port + Docker network alias `sftp:22`
- Replace low-level `dockerClient.execCreateCmd` with `sftpContainer.execInContainer()` in `getFileListFromContainer` and `deleteFileFromContainer`

**Assertion issues:**
- Fix `getFileListFromContainer` returning `ls` stderr as filenames when directory doesn't exist; now checks exit code and returns empty list on failure
- Replace immediate `assertEquals(size, N)` file-count assertions with `Awaitility.await()` polling (30s timeout) to handle file commit delay (rename from tmp)
- Add `awaitFileCount()` helper used in `testMultipleTableAndSaveMode` and `testSftpFileReadAndWrite`

**Timing issues:**
- Increase `Awaitility.atMost` from 60s to 120s in `testSftpBinaryUpdateModeContinuousDiscoveryDistcp` for both file-appearance checks
- Replace `Thread.sleep(2500)` mtime-stability check with `Awaitility.during(6s)` — verifies mtime stays unchanged for 6 consecutive seconds (covering 6+ scan intervals), instead of a single point-in-time check that could pass immediately
- Replace `Thread.sleep(1100)` mtime-gap guarantee with `waitUntilContainerTimeAfter()` — polls container clock until epoch second is strictly after src mtime, then writes dst file, ensuring deterministic mtime ordering for distcp `len_mtime` comparison
- Verify `chown -R` exit code after file copy to catch permission failures early
- Wrap `testFtpToAssertJsonFilter`, `testSftpBinaryUpdateModeDistcp`, and `testSftpBinaryUpdateModeContinuousDiscoveryDistcp` with `try-finally` for guaranteed cleanup

No. Test-only changes.

```bash
./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-file-sftp-e2e -DskipUT -DskipIT=false verify
```

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-file-sftp-e2e/src/test/java/org/apache/seatunnel/e2e/connector/file/fstp/SftpFileIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-file-sftp-e2e/src/test/java/org/apache/seatunnel/e2e/connector/file/fstp/SftpFileIT.java
@@ -36,18 +36,14 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.TestTemplate;
 import org.testcontainers.containers.Container;
 import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.lifecycle.Startables;
-import org.testcontainers.shaded.com.github.dockerjava.core.command.ExecStartResultCallback;
 
-import com.github.dockerjava.api.command.ExecCreateCmdResponse;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
@@ -59,13 +55,11 @@ import java.util.stream.Stream;
 @Slf4j
 public class SftpFileIT extends TestSuiteBase implements TestResource {
 
-    private static final String SFTP_IMAGE = "atmoz/sftp:latest";
+    private static final String SFTP_IMAGE = "atmoz/sftp:alpine-3.7";
 
     private static final String SFTP_CONTAINER_HOST = "sftp";
 
     private static final int SFTP_PORT = 22;
-
-    private static final int SFTP_BIND_PORT = 2222;
 
     private static final String SFTP_CONTAINER_HOME = "/home/seatunnel";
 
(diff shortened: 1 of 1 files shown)


DanielLeens (approved):
Thanks for working on this. I reviewed the full diff and traced the test flow end to end.

What this PR fixes
- User pain point: the SFTP E2E suite still had a few classic flaky-test patterns: fixed host-port assumptions, immediate assertions on asynchronously produced files, incomplete cleanup, and container exec calls whose exit codes were not being checked.
- Fix approach: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-file-sftp-e2e/src/test/java/org/apache/seatunnel/e2e/connector/file/fstp/SftpFileIT.java` now waits for the container readiness explicitly, replaces immediate assertions with Awaitility-based condition waits, validates `execInContainer(...)` exit codes, and moves cleanup into `finally` blocks.
- In one sentence: this turns the SFTP E2E from timing-sensitive checks into condition-driven checks with proper cleanup.

Full runtime path I checked
```text
SFTP E2E
  -> start SFTP container and wait for listening port
  -> submit SeaTunnel job via TestHelper.execute(...)
  -> poll file outputs / mtimes inside the container
  -> assert eventual state with Awaitility
  -> cancel long-running job when needed
  -> clean test directories in finally
```

Core review notes
- The normal path absolutely hits this change because every SFTP E2E in this class depends on the same container setup and verification helpers.
- Replacing fixed-time assumptions with explicit waits is the right fix here.
- I did not find new flaky patterns like `Thread.sleep`, fixed host-port conflicts, or unreset shared state introduced by this PR.
- The helper additions are still small and readable.

Compatibility
- Fully backward compatible.
- No production API/config/protocol/serialization change. This is test-only stabilization.

Test stability assessment
- Rating: stable.
- `SftpFileIT.java:214-223` and `:372-381` now wait for file counts instead of asserting immediately.
- `SftpFileIT.java:285-324` uses condition-based waiting for the continuous-discovery path and validates job cancellation.
- `SftpFileIT.java:449-457` waits on container time progression instead of using a hard-coded sleep.

Testing / CI
- The current GitHub checks are green.

Conclusion: can merge
- Blocking items: none.

This is a solid stabilization pass. Approving from my side.

davidzollo (approved):
Great job for CI

PR #10975, opened 2026-05-29
[Fix][Connector-V2] Fix wrong exception reference in MilvusSourceReader query error handling

CosmosNi:
When queryResultsR fails, the exception was incorrectly referencing loadStateResponse.getException() instead of queryResultsR.getException(),causing misleading error messages.

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-milvus/src/main/java/org/apache/seatunnel/connectors/seatunnel/milvus/source/MilvusSourceReader.java
+++ b/seatunnel-connectors-v2/connector-milvus/src/main/java/org/apache/seatunnel/connectors/seatunnel/milvus/source/MilvusSourceReader.java
@@ -195,8 +195,7 @@ public class MilvusSourceReader implements SourceReader<SeaTunnelRow, MilvusSour
 
         if (queryResultsR.getStatus() != R.Status.Success.getCode()) {
             throw new MilvusConnectorException(
-                    MilvusConnectionErrorCode.SERVER_RESPONSE_FAILED,
-                    queryResultsR.getException());
+                    MilvusConnectionErrorCode.SERVER_RESPONSE_FAILED, queryResultsR.getException());
         }
         QueryResultsWrapper wrapper = new QueryResultsWrapper(queryResultsR.getData());
         List<QueryResultsWrapper.RowRecord> records = wrapper.getRowRecords();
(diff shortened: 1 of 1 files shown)


CosmosNi:
@chl-wxp CC

DanielLeens (approved):
Thanks for the fix here. I pulled the latest head `e015dfb0c7bd53b80bd419a122a53b639eee676b` locally and traced the actual Milvus source read path from `pollNext(...)` into `pollNextData(...)`.

# What this PR fixes
- User pain: when the Milvus source fails on the `count(*)` query step, the current code throws the exception object from the earlier `getLoadState(...)` call instead of the exception from the failed query itself. That makes the stack trace point at the wrong RPC and slows down root-cause analysis.
- Fix approach: replace `loadStateResponse.getException()` with `queryResultsR.getException()` in the query-failure branch.
- One-line summary: this is a small but correct error-attribution fix that makes Milvus source failures point to the real failing RPC.

# Runtime chain I checked
```text
Milvus bounded source poll
  -> MilvusSourceReader.pollNext(...)
      -> pollNextData(split, output)
          -> client.getLoadState(...)
              -> fail early if the collection / partition is not loaded
          -> client.query(count(*))
              -> old code: throws the previous load-state exception object
              -> new code: throws the current query exception object
          -> queryIteratorData(...)
              -> continue reading the real data batches
```

# Findings
- I did not find a blocking issue on the current head.
- The change is on the real runtime path, because every split goes through the `count(*)` query before iterating the batch data.
- The fix is narrowly scoped: it does not change load-state checks, data iteration, retry semantics, or resource handling.
- Compatibility is fully preserved. This only corrects the exception source on the error path.

# Testing / docs / CI
- This PR does not add new UT/E2E code, so there is no new flaky-test risk to call out in this round.
- This is not a user-visible behavior change, so I do not see a docs update requirement here.
- I also checked the current GitHub checks: `Build` is green. The current `mergeStateStatus=BLOCKED` is the review gate, not a merge conflict.

### Conclusion: can merge

1. Blocking items
- None from my source-level review.

2. Suggested follow-ups
- A dedicated unit test for the query-failure branch would be a nice extra guard, but I do not treat that as a merge blocker for this PR.

Overall, this is a focused and correct fix. I do not see a better alternative than the current minimal change.

davidzollo (approved):
+1
Good job.

PR #10789, opened 2026-04-19
 [Feature][CLI] Add SeaTunnel CLI for natural language config generation

SEZ9:
### Purpose of this pull request

  Add `seatunnel-cli`, a Python CLI tool that generates Apache SeaTunnel HOCON pipeline
  configurations from natural language descriptions (English and Chinese).

  Key capabilities:
  - **Multi-agent pipeline**: Planner → Config Generator → Validator → Auto-fix (up to 3 rounds)
  - **100+ connectors**: Auto-generated metadata catalog from Java source (`*Factory.java`, `*Options.java`) with 1200+ option definitions and
  inheritance chain resolution
  - **Multi-provider LLM support**: AWS Bedrock, Anthropic API, OpenAI (and compatible APIs)
  - **Three-tier knowledge base**: Runtime REST API → auto-generated catalog → keyword routing
  - **Dry-run validation**: Local HOCON syntax check + engine `--check` + REST API validation
  - **Auto-save**: Generated configs automatically saved to `~/.seatunnel/last_job.conf`
  - **Auto-fix on failure**: `/check` and `/run` failures trigger LLM-powered diagnosis and config repair
  - **Session & memory**: Persistent conversation sessions and connection detail memory across sessions
  - **Interactive & single-shot modes**: REPL for exploration, one-liner for scripting

  Usage:
  ```bash
  cd seatunnel-cli
  pip install -e ".[bedrock]"
  seatunnel  # interactive mode
  seatunnel "Sync MySQL users table to S3 Parquet" -o job.conf  # single-shot

  Does this PR introduce any user-facing change?

  Yes. This adds a new seatunnel-cli module (Python) as a standalone tool alongside the existing Java codebase. It introduces:

  - seatunnel CLI command for natural language config generation
  - Interactive REPL with /save, /check, /run, /connectors, /memory commands
  - Built-in connector catalog (connector_catalog.json) with 100 connectors
  - --sync-catalog option to regenerate catalog from SeaTunnel Java source

  No changes to existing Java modules. The CLI is fully self-contained under seatunnel-cli/.

  How was this patch tested?

  - Manual testing of interactive mode and single-shot mode with multiple LLM providers (Bedrock, OpenAI)
  - Connector catalog generation tested against current dev branch: 100 connectors, 1273 options, 99.5% option resolution rate (7 unresolved out of
  1273)
  - Dry-run validation tested with local HOCON parsing and engine --check mode
[...]

PR #10977, opened 2026-05-29
[STIP-28][Feature][Zeta] Enhance OptionRule to support rich declarative validation beyond presence checks

### LINKED ISSUE #10976: [STIP-28][Feature][Zeta] Enhance OptionRule to support rich declarative validation beyond presence checks
### Search before asking

## Background

`Factory.optionRule()` is SeaTunnel's declarative configuration validation system. Every connector and transform implements `optionRule()` to declare which options are required, optional, mutually exclusive, or conditionally dependent. `ConfigValidator` enforces these rules at startup before any connector is instantiated.

However, the current `OptionRule` + `Condition` system only supports **presence-based** validation and **equality-only** conditions (`Condition` source code already has `TODO: support another condition`). This forces connector developers to write additional imperative `if/else` validation code scattered across `*Config.java`, `*Factory.java`, and `Source/Sink` constructors.

```
User Config
  → ConfigValidator.validate(factory.optionRule())   ← only checks presence + equality
  → ConnectorConfig.of(config)                        ← hidden if/else validation
  → Source/Sink constructor                           ← more hidden if/else validation
```

## Current Status: Validations That `optionRule()` Cannot Express

After a full audit of all connector and transform modules, we identified **8 categories** of validation that cannot be declared in `optionRule()` and can only be implemented via manual `if/else`.

### 1. Numeric Range

| Connector | Validation File | Validation |
|---|---|---|
| EdgeSocket | `EdgeSocketConfig.java` | `local_queue_capacity > 0` |
| EdgeSocket | `EdgeSocketConfig.java` | `queue_backpressure_watermark_ratio ∈ (0, 1]` (half-open interval) |
| EdgeSocket | `EdgeSocketConfig.java` | `queue_full_retry_after_ms > 0` |
| Fake | `FakeConfig.java` | `tinyint_min >= -128 && tinyint_min <= 127` (same pattern for smallint/int/bigint/float/double/vector, 7 groups total) |
| MongoDB CDC | `MongodbSourceConfigProvider.java` | `batchSize >= 0` |
| MongoDB CDC | `MongodbSourceConfigProvider.java` | `pollAwaitTimeMillis > 0` |
| MongoDB CDC | `MongodbSourceConfigProvider.java` | `splitSizeMB > 0` |
| MongoDB | `MongodbReadOptions.java` | `fetchSize > 0` |
| DynamoDB | `AmazonDynamoDBConfig.java` | `max_retries >= 0` |
| HBase | `HbaseClient.java` | `start_timestamp >= 0`, `end_timestamp >= 0` |
| File | `FileSplitStrategyFactory.java` | `file_split_size > 0` |
| Prometheus | `PrometheusSinkConfig.java` | `flush_interval > 0` |

### 2. Cross-Field Comparison (min <= max)

| Connector | Validation File | Validation |
|---|---|---|
| HBase | `HbaseClient.java` | `start_timestamp < end_timestamp` |
| HBase | `HBaseUtil.java` | `startRowkey <= endRowkey` |
| Fake | `FakeConfig.java` | `tinyint_min <= tinyint_max` (same pattern for smallint/int/bigint/float/double/vector, 6 groups total) |
| JDBC | `JdbcNumericBetweenParametersProvider.java` | `partition_lower_bound <= partition_upper_bound` |

### 3. String Format

| Connector | Validation File | Validation | Type |
|---|---|---|---|
| JDBC (MySQL/Dameng/Kingbase/Iris/Redshift/OceanBase etc.) | `*CatalogFactory.java` | `StringUtils.isNoneBlank(url)` | Non-blank |
| Paimon | `PaimonConfig.java` | `checkArgumentNotBlank(catalog_name/warehouse/catalog_uri)` | Non-blank |
| Redis | `RedisTableConfig.java` | `keys` and `data_type` must be non-blank | Non-blank |
| Pulsar | `PulsarConsumerConfig.java` | `subscriptionName` must be non-blank | Non-blank |
| Databend | `DatabendSourceFactory.java` | `url.startsWith("jdbc:databend://")` | Prefix match |
| JDBC | `JdbcSourceConfig.java` | `where_condition.toLowerCase().startsWith("where")` | Case-insensitive prefix |
| EdgeSocket | `EdgeSocketConfig.java` | `endpoint` must be `host:port` format | Format match |
| Redis | `RedisParameters.java` | cluster nodes must be `host:port` format | Format match |
| DuckDB | `DuckDBURLParser.java` | JDBC URL regex validation | Regex match |
| Oracle CDC | `OracleSourceConfigFactory.java` | table name `split("\\.").length` must be 2 or 3 | Format validation |
| Postgres CDC | `PostgresSourceConfigFactory.java` | table name must be `schema.table` or `db.schema.table` | Format validation |
| FileSink | `FileSinkConfig.java` | `file_name_expression.contains("#{transactionId}")` | Contains substring |
| Hive | `HiveTableTemplateUtils.java` | DDL template must contain `${database}`, `${table}` and `CREATE TABLE` | Contains substring |

### 4. String Length

| Connector | Validation File | Validation |
|---|---|---|
| ClickHouse File | `ClickhouseFileSinkFactory.java` | `file_fields_delimiter` must be exactly 1 character |

### 5. Case Constraints

| Connector | Validation File | Validation |
|---|---|---|
| Oracle CDC | `OracleSourceConfigFactory.java` | Database name must be all uppercase |
| Oracle CDC | `OracleSourceConfigFactory.java` | Table name must be all uppercase |

### 6. Collection Constraints

| Connector | Validation File | Validation | Type |
|---|---|---|---|
| Pulsar | `PulsarMultiTableConfig.java` | `tables_configs` list must not be empty | Non-empty collection |
| RocketMQ | `RocketMqSourceConfig.java` | `topics` in each table config must not be empty | Non-empty collection |
| Doris | `DorisTableConfig.java` | `database.table` pairs must be unique | Unique elements |
| Fake | `MultipleTableFakeSourceConfig.java` | table names must be unique | Unique elements |
| Pulsar | `PulsarMultiTableConfig.java` | `table_path` must be unique | Unique elements |
| JDBC | `JdbcSourceTableConfig.java` | `table_path` must be unique | Unique elements |
| MongoDB CDC | `MongodbIncrementalSourceFactory.java` | `catalogTables.size() == collections.size()` | Collection size equality |

### 7. Conditional Value Checks (if A == X then B must satisfy condition)

| Connector | Validation File | Trigger Condition | Validation |
|---|---|---|---|
| Kafka | `KafkaSourceConfig.java` | `ignore_no_leader_partition == true` | `partition-discovery.interval-millis > 0` |
| FileSink | `FileSinkConfig.java` | `is_enable_transaction == true` | `file_name_expression` contains `#{transactionId}` |
| Iceberg | `IcebergSinkConfig.java` | `upsert_mode_enabled == true` | `primary_keys` must be non-blank |
| Hudi | `HudiTableConfig.java` | `op_type == UPSERT` | `record_key_fields` required |
| Doris | `DorisSinkConfig.java` | `direct_to_be == true` | `benodes` must not be empty |
| Paimon | `PaimonConfig.java` | `catalog_type == HIVE` | `catalog_uri` must be non-blank |
| EdgeSocket | `EdgeSocketConfig.java` | `auth_type == TOKEN` | `token` must be non-blank |
| Paimon | `PaimonSinkConfig.java` | `non_primary_key == true` | `primary_keys` must be empty list |
| MaxCompute | `MaxcomputeUtil.java` | `sts_token` is not empty | `access_id` + `access_key` must both be present |
| FileSink | `FileSinkConfig.java` | `partition_by` is not empty | `partition_by` fields must be a subset of `sink_columns` |

### 8. Mutual Exclusion / At-Least-One Non-Empty (Beyond exclusive)

| Connector/Transform | Validation File | Validation |
|---|---|---|
| ClickHouse | `ClickhouseTableConfig.java` | `table_path` and `sql` cannot both be empty (at least one required) |
| RabbitMQ | `RabbitmqSource.java` | `schema` and `table_configs` are mutually exclusive (already declared in optionRule but manually re-validated) |
| FilterRowKind Transform | `FilterRowKindTransform.java` | `include_kinds` and `exclude_kinds` are mutually exclusive (should be in optionRule but manually implemented) |
| TableFilter Transform | `TableFilterConfig.java` | `database_pattern`/`schema_pattern`/`table_pattern`/`pattern_mode` — at least one must be configured |

### Concrete Examples from the Codebase

**EdgeSocket** — numeric range + half-open interval:
```java
// connector-edge-socket/.../EdgeSocketConfig.java
if (this.queueBackpressureWatermarkRatio <= 0 || this.queueBackpressureWatermarkRatio > 1) {
    throw new IllegalArgumentException(
            "Invalid queue_backpressure_watermark_ratio: " + this.queueBackpressureWatermarkRatio
            + ", it must be in (0, 1]");
}
```

**HBase** — cross-field comparison:
```java
// connector-hbase/.../HbaseClient.java
long min = startTimestamp == null ? 0L : startTimestamp;
long max = endTimestamp == null ? Long.MAX_VALUE : endTimestamp;
if (min >= max) {
    throw new IllegalArgumentException("start_timestamp must be less than end_timestamp");
}
```

**Kafka** — conditional value check:
```java
// connector-kafka/.../KafkaSourceConfig.java
if (this.ignoreNoLeaderPartition && this.discoveryIntervalMillis <= 0) {
    throw new IllegalArgumentException(
            "partition-discovery.interval-millis must be configured when "
            + "ignore_no_leader_partition is set to true.");
}
```

**Oracle CDC** — uppercase constraint:
```java
// connector-cdc-oracle/.../OracleSourceConfigFactory.java
for (int i = 0; i < database.length(); i++) {
    if (Character.isLetter(database.charAt(i))
            && !Character.isUpperCase(database.charAt(i))) {
        throw new IllegalArgumentException(
                "Oracle database name must be in all uppercase, database: " + database);
    }
}
```

**FileSink** — conditional string contains:
```java
// connector-file-base/.../FileSinkConfig.java
if (this.isEnableTransaction
        && !this.fileNameExpression.contains(FileBaseSinkOptions.TRANSACTION_EXPRESSION)) {
    throw new FileConnectorException(ILLEGAL_ARGUMENT,
            "file_name_expression must contains #{transactionId} when is_enable_transaction is true");
}
```

**FilterRowKind Transform** — should be in `optionRule()` but manually duplicated:
```java
// transforms-v2/.../FilterRowKindTransform.java
if ((includeKinds.isEmpty() && excludeKinds.isEmpty())
        || (!includeKinds.isEmpty() && !excludeKinds.isEmpty())) {
    throw new SeaTunnelRuntimeException(ILLEGAL_ARGUMENT,
            "These options(include_kinds, exclude_kinds) are mutually exclusive...");
}
```

## Problems with the Current Approach

1. **Hidden validation** — Validation rules are buried inside connector `*Config.java` constructors and `Source/Sink` implementations, invisible to the framework, REST API, Web UI, and CLI tools. Users can only discover these constraints by hitting runtime errors.

2. **Duplicated boilerplate** — The same patterns (`> 0`, non-blank, conditional required) are reimplemented across dozens of connectors with slightly different error messages and exception types.

3. **Inconsistent error reporting** — Some connectors throw `IllegalArgumentException`, others throw connector-specific exceptions, and some use `Preconditions.checkArgument`. Error messages vary in quality and helpfulness.

4. **Invisible to tooling and AI** — The REST API `/option-rules` and CLI `seatunnel-metadata-export.sh` only expose what `optionRule()` declares. The actual validation rules in `*Config.java` are invisible, making it impossible for Web UI or AI-assisted config generation to know the real constraints.

5. **Bypassed by `--check`** — `SeaTunnelConfValidateCommand` validates `optionRule()` but does **not** call connector `*Config.of()` methods, so manual validations are missed during offline config checking.

## Goal and Scope

The goal is to enhance `optionRule()` so that the 8 categories of validation listed above can be expressed declaratively instead of hidden in imperative code. This would make them visible to `ConfigValidator`, REST API `/option-rules`, CLI tools, Web UI, `--check` offline validation, and AI-assisted config generation.

The following validation types are **out of scope** because they are runtime logic that depends on parsed data structures, external systems, or execution context:

- Nested list element validation (`tables_configs[]` per-element checks)
- Base64 decode then validate byte length
- Timestamp vs `System.currentTimeMillis()`
- External state validation (DB/metastore connection)
- SQL/regex syntax parsing
- Schema/catalog cross-validation (field existence, type compatibility)

These belong in `*Config.java` constructors or `Source/Sink` initialization and are not suitable for a declarative option rule system.

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

nzw921rx:
## Purpose

Closes #10976

## Description

The current `OptionRule` + `Condition` system only supports presence checks and equality conditions. This forces connector developers to write imperative `if/else` validation scattered across `*Config.java` and `Source/Sink` constructors — invisible to REST API, CLI, Web UI, `--check` offline validation, and AI-assisted config generation.

This PR enhances the existing `required()` / `optional()` / `conditional()` methods with `Condition`-accepting overloads so that value constraints can be declared directly in `optionRule()`.

### What's changed

**`ConditionOperator` enum (new)** — 30 operators with self-describing metadata (`category`, `arity`, `source`, `displaySymbol`), covering:
- Numeric: `>`, `>=`, `<`, `<=`
- String: `is not blank`, `starts with`, `ends with`, `contains`, `matches`, `is uppercase`, `is lowercase`
- String length: `length ==`, `length >=`, `length <=`
- Collection: `is not empty`, `has unique elements`, `size ==`, `size >=`, `size <=`
- Cross-field: `< [field]`, `<= [field]`, `> [field]`, `>= [field]`, `== [field]`, `!= [field]`, `size == [field]`

**`Condition` class** — 27 new static factory methods + defensive constructor validation (null operator, missing `compareOption` for FIELD ops, missing `expectValue` for BINARY+LITERAL ops) + circular chain detection in `and()`/`or()`.

**`OptionRule.Builder`** — 6 new overloads:

| Method | Purpose |
|--------|---------|
| `required(Option, Condition, Condition...)` | Single-field required + value constraints |
| `required(Option, Option, Condition, Condition...)` | Two-field required + cross-field constraints |
| `optional(Option, Condition, Condition...)` | Optional + value constraints (validated when present) |
| `optional(Option, Option, Condition, Condition...)` | Two-field optional + constraints |
| `conditional(Option, T, Condition, Condition...)` | Conditionally triggered value constraints |
| `conditional(Option, T, Option, Option, Condition, Condition...)` | Conditional two-field + constraints |

### Usage example

```java
OptionRule.builder()
    .required(PORT,
        Condition.greaterOrEqual(PORT, 1)
            .and(Condition.lessOrEqual(PORT, 65535)))
    .required(HOST, Condition.notBlank(HOST))
    .required(START_TS, END_TS,
        Condition.lessThanField(START_TS, END_TS))
    .optional(MODE)
    .conditional(MODE, StartMode.TIMESTAMP,
        Condition.greaterThan(TIMESTAMP_VALUE, 0))
    .build();
```

## Test Plan

```bash
./mvnw -B -pl seatunnel-api test -Dtest=ConfigValidatorTest
./mvnw -B -pl seatunnel-engine/seatunnel-engine-server test -Dtest=OptionRulesServiceTest
```

### DIFF
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -37,6 +37,45 @@ An `Option` defines a single configuration field:
 
 This is the smallest reusable configuration contract in SeaTunnel.
 
+Typical usage in a connector options class:
+
+```java
+public static final Option<Integer> PORT =
+        Options.key("port")
+                .intType()
+                .defaultValue(3306)
+                .withDescription("Database server port");
+
+public static final Option<String> HOST =
+        Options.key("host")
+                .stringType()
+                .noDefaultValue()
+                .withDescription("Database server hostname");
+
+public static final Option<List<String>> TABLES =
+        Options.key("tables")
+                .listType()
+                .noDefaultValue()
+                .withDescription("List of tables to read");
+```
+
+`Options.key(...)` builder supports the following type methods:
+
+| Method | Java Type |
+|--------|-----------|
+| `stringType()` | `String` |
+| `intType()` | `Integer` |
+| `longType()` | `Long` |
+| `doubleType()` | `Double` |
+| `floatType()` | `Float` |
+| `booleanType()` | `Boolean` |
+| `listType()` | `List<String>` |
+| `listType(Class<T>)` | `List<T>` |
+| `mapType()` | `Map<String, String>` |
+| `enumType(Class<E>)` | `Enum` subclass |
+| `singleChoice(Class<T>, List<T>)` | single-choice with allowed values |
+| `type(new TypeReference<T>() {})` | any custom type |
+
 ### `OptionRule`
 
 An `OptionRule` describes how multiple options behave together. It can express rules such as:
@@ -48,20 +87,220 @@ An `OptionRule` describes how multiple options behave together. It can express r
 
 This is how SeaTunnel moves beyond flat configuration and supports richer connector contracts.
 
+A connector factory exposes its rules through the `optionRule()` method:
+
+```java
+@Override
+public OptionRule optionRule() {
+    return OptionRule.builder()
+            .required(HOST, PORT)                              // absolutely required
+            .exclusive(USERNAME, BEARER_TOKEN)                 // exactly one must be set
+            .bundled(USERNAME, PASSWORD)                       // all or none
+            .conditional(MODE, WriteMode.UPSERT, UPSERT_KEY)  // required when MODE == UPSERT
+            .optional(BATCH_SIZE, RETRY_COUNT)                 // purely optional
+            .build();
+}
+```
+
+### Value Constraints (`Condition`)
+
+Beyond structural rules (required, exclusive, etc.), options can carry **value-level constraints** that the runtime validates before a job starts. The `Condition` API provides a fluent way to attach these constraints inside `OptionRule.builder()`.
+
+**Numeric range constraints:**
+
+```java
+OptionRule.builder()
+        .required(PORT,
+                Condition.greaterOrEqual(PORT, 1)
+                        .and(Condition.lessOrEqual(PORT, 65535)))
+        .build();
+```
+
+**String constraints:**
+
+```java
+OptionRule.builder()
+        .required(HOST, Condition.notBlank(HOST))
+        .optional(DB_NAME, Condition.upperCase(DB_NAME))
+        .build();
+```
+
+**Cross-field comparison:**
+
+```java
+OptionRule.builder()
+        .required(START_TS, END_TS,
+                Condition.lessThanField(START_TS, END_TS))
+        .build();
+```
+
+**Collection constraints:**
+
+```java
+OptionRule.builder()
+        .required(TABLES,
+                Condition.notEmpty(TABLES)
+                        .and(Condition.unique(TABLES)))
+        .build();
+```
+
+Available operators:
+
+| Category | Operator | Description |
+|----------|----------|-------------|
+| Numeric | `greaterThan` | value > threshold |
+| Numeric | `greaterOrEqual` | value >= threshold |
+| Numeric | `lessThan` | value < threshold |
+| Numeric | `lessOrEqual` | value <= threshold |
+| String | `notBlank` | string is not empty or whitespace-only |
+| String | `startsWith` | string starts with a given prefix |
+| String | `endsWith` | string ends with a given suffix |
+| String | `contains` | string contains a given substring |
+| String | `matches` | string matches a regular expression |
+| String | `upperCase` | string is all uppercase |
+| String | `lowerCase` | string is all lowercase |
+| String length | `lengthEqual` | string length == n |
+| String length | `lengthGreaterOrEqual` | string length >= n |
+| String length | `lengthLessOrEqual` | string length <= n |
+| Collection | `notEmpty` | collection is not empty |
+| Collection | `unique` | collection has no duplicate elements |
+| Collection | `sizeEqual` | collection size == n |
+| Collection | `sizeGreaterOrEqual` | collection size >= n |
+| Collection | `sizeLessOrEqual` | collection size <= n |
+| Cross-field | `lessThanField` | value < another option's value |
+| Cross-field | `lessOrEqualField` | value <= another option's value |
+| Cross-field | `greaterThanField` | value > another option's value |
+| Cross-field | `greaterOrEqualField` | value >= another option's value |
+| Cross-field | `fieldEqual` | value == another option's value |
+| Cross-field | `fieldNotEqual` | value != another option's value |
+| Cross-field | `fieldSizeEqual` | collection size == another collection's size |
+
+:::tip
+Multiple conditions can be chained with `.and(...)` or `.or(...)` to form compound constraints.
+:::
+
 ### `ReadonlyConfig`
 
 `ReadonlyConfig` is the runtime container from which connectors and transforms read their resolved values. It gives plugin implementations a stable, typed access pattern after parsing and validation have already happened.
 
-## End-To-End Flow
+```java
+@Override
+public void prepare(Config pluginConfig) {
+    ReadonlyConfig config = ReadonlyConfig.fromConfig(pluginConfig);
+    String host = config.get(HOST);     // typed access, never returns raw Object
+    int port = config.get(PORT);        // default applied automatically if not set
+}
+```
+
+## Validation Flow
 
 At a high level, configuration flows through the system like this:
 
-1. A plugin defines `Option` and `OptionRule` metadata.
+1. A plugin defines `Option` and `OptionRule` metadata (including value constraints).
 2. A user writes HOCON, JSON, or SQL-based job configuration.
 3. SeaTunnel parses the configuration into a runtime representation.
-4. Validation applies the connector rules.
+4. `ConfigValidator` checks structural rules (required, exclusive, bundled, conditional) and then evaluates value constraints by delegating each `Condition` to `ConditionEvaluators`.
 5. The resolved values are exposed to the runtime through `ReadonlyConfig`.
 6. The same metadata can also be exposed through REST for UI rendering and automation.
+
+When validation fails, `OptionValidationException` is thrown with a message that includes the constraint expression and the option key involved.
+
+## OptionRule Pattern Guide
+
+Validation logic declared in `optionRule()` runs at job submission time, produces uniform error messages, and is automatically exposed to the REST API and Web UI. Placing validation in Config constructors or Writer/Reader code delays failure to task startup time and hides constraints from tooling.
+
+The following patterns cover common scenarios. Each one shows the recommended declarative form inside `OptionRule.builder()`.
+
+### Required fields
+
+Some fields must always be present. A job that omits them should be rejected at submission.
+
+```java
+.required(HOST, PORT, DATABASE)
+```
+
+### Mutually exclusive options
+
+When only one of several options should be set at a time, `exclusive` enforces the constraint.
+
+```java
+.exclusive(TOPIC, TOPIC_PATTERN)
+```
+
+### Bundled options
+
+A group of options that only make sense together. Either all of them are set or none.
+
+```java
+.bundled(USERNAME, PASSWORD)
+```
+
+### Conditional required options driven by an enum
+
+When an enum option takes a specific value, additional fields become required.
+
+```java
+.conditional(START_MODE, StartMode.TIMESTAMP, START_MODE_TIMESTAMP)
+.conditional(START_MODE, StartMode.SPECIFIC_OFFSETS, START_MODE_OFFSETS)
+```
+
+### Conditional required options driven by a boolean
+
+A boolean toggle that activates different sets of required fields depending on its value.
+
+```java
+.conditional(IS_EXACTLY_ONCE, true, XA_DATA_SOURCE_CLASS, TRANSACTION_TIMEOUT)
+.conditional(IS_EXACTLY_ONCE, false, MAX_RETRIES)
+```
+
+### Numeric range
+
+Port numbers, batch sizes, ratios, and similar numeric fields often have valid ranges.
+
+```java
+.required(PORT,
+        Condition.greaterOrEqual(PORT, 1)
+                .and(Condition.lessOrEqual(PORT, 65535)))
+```
+
+### String format and content
+
+Host names that must not be blank, identifiers that must be uppercase, or endpoints that must match a pattern.
+
+```java
+.required(HOST, Condition.notBlank(HOST))
+.required(DATABASE, Condition.upperCase(DATABASE))
+.required(ENDPOINT, Condition.matches(ENDPOINT, "^[^:]+:\\d+$"))
+```
+
+### Cross-field comparison
+
+When the value of one option must be smaller or larger than another.
+
+```java
+.required(START_TS, END_TS,
+        Condition.lessThanField(START_TS, END_TS))
+```
+
+### Collection constraints
+
+Lists that must not be empty, or whose elements must be unique.
+
+```java
+.required(TABLES,
+        Condition.notEmpty(TABLES)
+                .and(Condition.unique(TABLES)))
+```
+
+### Compound constraints
+
+Multiple conditions combined with `.and(...)` or `.or(...)`.
+
+```java
+.required(RATIO,
+        Condition.greaterThan(RATIO, 0.0)
+                .and(Condition.lessOrEqual(RATIO, 1.0)))
+```
 
 ## Why It Matters For Operators
 
@@ -69,6 +308,7 @@ This architecture is also what makes the `option-rules` REST endpoint useful. To
 
 - which fields are required
 - which fields are conditional
+- what value constraints apply (numeric ranges, patterns, cross-field rules)
 - which defaults are active on the running server
 
 That is why the option system sits at the boundary of both developer experience and operations.
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -192,6 +192,18 @@ Validation logic declared in `optionRule()` runs at job submission time, produce
 
 The following patterns cover common scenarios. Each one shows the recommended declarative form inside `OptionRule.builder()`.
 
+Quick reference:
+
+| Scenario | Recommended API |
+|----------|------------------|
+| Always required fields | `.required(opt...)` |
+| Exactly one in a set | `.exclusive(opt...)` |
+| All-or-none group | `.bundled(opt...)` |
+| Required only when trigger matches | `.conditional(trigger, value, requiredOpt...)` |
+| Validate value only when trigger matches | `.conditional(trigger, value, condition...)` |
+| Optional field with value check when present | `.optional(opt, condition...)` |
+| Cross-field comparisons | `Conditions.lessThanField/greaterThanField(...)` |
+
 ### Required fields
 
 Some fields must always be present. A job that omits them should be rejected at submission.
@@ -218,19 +230,29 @@ A group of options that only make sense together. Either all of them are set or 
 
 ### Conditional required options driven by an enum
 
-When an enum option takes a specific value, additional fields become required.
+When an enum option takes a specific value, additional fields become required. The method signature is:
+
+```
+.conditional(triggerOption, triggerValue, requiredOption...)
+```
+
+Meaning: when the user sets `triggerOption` to `triggerValue`, all listed `requiredOption` fields become mandatory.
 
 ```java
+// When START_MODE = TIMESTAMP, START_MODE_TIMESTAMP becomes required
 .conditional(START_MODE, StartMode.TIMESTAMP, START_MODE_TIMESTAMP)
+// When START_MODE = SPECIFIC_OFFSETS, START_MODE_OFFSETS becomes required
 .conditional(START_MODE, StartMode.SPECIFIC_OFFSETS, START_MODE_OFFSETS)
 ```
 
 ### Conditional required options driven by a boolean
 
-A boolean toggle that activates different sets of required fields depending on its value.
+Same pattern as enum-driven, but the trigger value is a boolean.
 
 ```java
+// When IS_EXACTLY_ONCE = true, XA_DATA_SOURCE_CLASS and TRANSACTION_TIMEOUT become required
 .conditional(IS_EXACTLY_ONCE, true, XA_DATA_SOURCE_CLASS, TRANSACTION_TIMEOUT)
+// When IS_EXACTLY_ONCE = false, MAX_RETRIES becomes required
 .conditional(IS_EXACTLY_ONCE, false, MAX_RETRIES)
 ```
 
@@ -306,12 +328,27 @@ AND binds tighter than OR, so `A.or(B.and(C))` evaluates as `A || (B && C)`. Thi
 .optional(PORT)
 ```
 
-### Conditional required with value constraint
+### Conditional required vs conditional value constraint
 
-When a conditional option also needs value validation, pass the constraint as the last argument. The constraint is only enforced when the trigger condition is met.
+:::tip
+
+These two forms look similar but mean different things:
+
+- `conditional(trigger, value, option...)` makes options conditionally required.
+- `conditional(trigger, value, condition...)` only validates values when the target option is present; it does not make that option required.
+
+:::
 
 ```java
-// When START_MODE == TIMESTAMP, START_TIMESTAMP is required and must be > 0
+// A) Conditionally required field
+.conditional(START_MODE, StartMode.TIMESTAMP, START_TIMESTAMP)
+
+// B) Optional field with conditional value validation
+.conditional(START_MODE, StartMode.TIMESTAMP,
+        Conditions.greaterThan(START_TIMESTAMP, 0L))
+
+// C) Required + value constraint (combine A and B)
+.conditional(START_MODE, StartMode.TIMESTAMP, START_TIMESTAMP)
 .conditional(START_MODE, StartMode.TIMESTAMP,
         Conditions.greaterThan(START_TIMESTAMP, 0L))
 ```
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -111,8 +111,8 @@ Beyond structural rules (required, exclusive, etc.), options can carry **value-l
 ```java
 OptionRule.builder()
         .required(PORT,
-                Condition.greaterOrEqual(PORT, 1)
-                        .and(Condition.lessOrEqual(PORT, 65535)))
+                Conditions.greaterOrEqual(PORT, 1)
+                        .and(Conditions.lessOrEqual(PORT, 65535)))
         .build();
 ```
 
@@ -120,7 +120,7 @@ OptionRule.builder()
 
 ```java
 OptionRule.builder()
-        .required(HOST, Condition.notBlank(HOST))
+        .required(HOST, Conditions.notBlank(HOST))
         .optional(DB_NAME, Condition.upperCase(DB_NAME))
         .build();
 ```
@@ -139,8 +139,8 @@ OptionRule.builder()
 ```java
 OptionRule.builder()
         .required(TABLES,
-                Condition.notEmpty(TABLES)
-                        .and(Condition.unique(TABLES)))
+                Conditions.notEmpty(TABLES)
+                        .and(Conditions.unique(TABLES)))
         .build();
 ```
 
@@ -259,8 +259,8 @@ Port numbers, batch sizes, ratios, and similar numeric fields often have valid r
 
 ```java
 .required(PORT,
-        Condition.greaterOrEqual(PORT, 1)
-                .and(Condition.lessOrEqual(PORT, 65535)))
+        Conditions.greaterOrEqual(PORT, 1)
+                .and(Conditions.lessOrEqual(PORT, 65535)))
 ```
 
 ### String format and content
@@ -268,9 +268,9 @@ Port numbers, batch sizes, ratios, and similar numeric fields often have valid r
 Host names that must not be blank, identifiers that must be uppercase, or endpoints that must match a pattern.
 
 ```java
-.required(HOST, Condition.notBlank(HOST))
+.required(HOST, Conditions.notBlank(HOST))
 .required(DATABASE, Condition.upperCase(DATABASE))
-.required(ENDPOINT, Condition.matches(ENDPOINT, "^[^:]+:\\d+$"))
+.required(ENDPOINT, Conditions.matches(ENDPOINT, "^[^:]+:\\d+$"))
 ```
 
 ### Cross-field comparison
@@ -288,8 +288,8 @@ Lists that must not be empty, or whose elements must be unique.
 
 ```java
 .required(TABLES,
-        Condition.notEmpty(TABLES)
-                .and(Condition.unique(TABLES)))
+        Conditions.notEmpty(TABLES)
+                .and(Conditions.unique(TABLES)))
 ```
 
 ### Compound constraints
@@ -298,8 +298,8 @@ Multiple conditions combined with `.and(...)` or `.or(...)`.
 
 ```java
 .required(RATIO,
-        Condition.greaterThan(RATIO, 0.0)
-                .and(Condition.lessOrEqual(RATIO, 1.0)))
+        Conditions.greaterThan(RATIO, 0.0)
+                .and(Conditions.lessOrEqual(RATIO, 1.0)))
 ```
 
 ## Why It Matters For Operators
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -104,79 +104,33 @@ public OptionRule optionRule() {
 
 ### Value Constraints (`Condition`)
 
-Beyond structural rules (required, exclusive, etc.), options can carry **value-level constraints** that the runtime validates before a job starts. The `Condition` API provides a fluent way to attach these constraints inside `OptionRule.builder()`.
+Beyond structural rules (required, exclusive, etc.), options can carry **value-level constraints** that the runtime validates before a job starts. The `Condition` API provides a fluent way to attach these constraints inside `OptionRule.builder()`. See the [OptionRule Pattern Guide](#optionrule-pattern-guide) below for usage examples.
 
-**Numeric range constraints:**
+Available operators (all accessed via the `Conditions` factory class):
 
-```java
-OptionRule.builder()
-        .required(PORT,
-                Conditions.greaterOrEqual(PORT, 1)
-                        .and(Conditions.lessOrEqual(PORT, 65535)))
-        .build();
-```
-
-**String constraints:**
-
-```java
-OptionRule.builder()
-        .required(HOST, Conditions.notBlank(HOST))
-        .optional(DB_NAME, Condition.upperCase(DB_NAME))
-        .build();
-```
-
-**Cross-field comparison:**
-
-```java
-OptionRule.builder()
-        .required(START_TS, END_TS,
-                Condition.lessThanField(START_TS, END_TS))
-        .build();
-```
-
-**Collection constraints:**
-
-```java
-OptionRule.builder()
-        .required(TABLES,
-                Conditions.notEmpty(TABLES)
-                        .and(Conditions.unique(TABLES)))
-        .build();
-```
-
-Available operators:
-
-| Category | Operator | Description |
-|----------|----------|-------------|
-| Numeric | `greaterThan` | value > threshold |
-| Numeric | `greaterOrEqual` | value >= threshold |
-| Numeric | `lessThan` | value < threshold |
-| Numeric | `lessOrEqual` | value <= threshold |
-| String | `notBlank` | string is not empty or whitespace-only |
-| String | `startsWith` | string starts with a given prefix |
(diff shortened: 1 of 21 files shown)


nzw921rx:
@davidzollo @zhangshenghang Can you help review it? Any suggestions would be greatly appreciated. If there are any design issues, you can move on to issue: #10976 😄

nzw921rx:
@yzeng1618 Can you help confirm if the support of OptionRulesService to return transform has a negative impact on AI generated configuration? Thank you very much.

DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current head from the real validation path instead of only reading the new utility classes.

# What this PR fixes
- User pain: plugin authors can describe whether an option is required, bundled, or conditional today, but they still cannot declaratively say "this value must be > 0" or "field B must be greater than field A" in `OptionRule` itself.
- Proposed fix: this PR adds `ConditionOperator`, `ConditionEvaluators`, `OptionRule.valueConstraints`, and REST metadata exposure so those validations can run during config validation.
- One-line summary: the direction is good, but the latest head still has two framework-level correctness holes before it is safe to merge.

# Call chain I traced
```text
connector config validate
  -> SeaTunnelConfValidateCommand.execute()
     -> ConfigValidator.of(readonlyConfig).validate(factory.optionRule())
        -> required / optional / conditional checks
        -> valueConstraints loop [ConfigValidator.java:174-176]
           -> ConditionEvaluators.evaluate(...)
     -> ConfigValidator.validateUnknownKeys(readonlyConfig, factory.optionRule(), pluginName)
        -> collectDeclaredKeys(rule) [ConfigValidator.java:117-129]
        -> validatePaths(...) [ConfigValidator.java:96-114]
```

# Findings
1. **Blocker: keys referenced only from `valueConstraints` are still rejected as unknown keys.**
   At `seatunnel-api/.../ConfigValidator.java:78-93,117-129`, `collectDeclaredKeys()` only collects required / optional options and nested conditional option rules. It does **not** collect `rule.getValueConstraints()` references. So a rule like the `OptionRule` example itself (`Condition.greaterThan(TIMESTAMP_VALUE, 0)`) is accepted by `validate()`, but once the user actually provides `timestamp_value`, `validateUnknownKeys()` rejects it as an unknown key. That means part of the new declarative constraint feature cannot be used on the normal config-validation path.

2. **Blocker: numeric comparison loses correctness for large integers.**
   At `seatunnel-api/.../ConditionEvaluators.java:215-223`, every `Number` comparison is normalized through `Double.compare(number.doubleValue(), ...)`. That is fine for ports or small counters, but it is not safe for large `long` values such as timestamps / offsets / IDs beyond the `2^53` precision boundary. In those cases different integers can collapse to the same `double`, so `greaterThan` / `lessThan` / field-to-field comparisons can silently return the wrong answer.

3. **Non-blocking: the public Javadoc example calls a method that does not exist.**
   `seatunnel-api/.../OptionRule.java:67-74` uses `Condition.range(TIMEOUT, 1000, 60000)`, but there is no `Condition.range(...)` factory in the codebase. That will mislead the next caller even though it does not affect runtime behavior.

# Tests and CI
The new tests cover many happy-path constraint cases, which is a good step. What is still missing is:
- a regression test for `validateUnknownKeys()` + `valueConstraints`
- a regression test for large-integer comparison precision

The good news is the current GitHub `Build` is green, so the remaining blockers here are source-level correctness blockers, not CI noise.

# Merge conclusion
**Conclusion: can merge after fixes.**

## Must fix before merge
- Finding 1: include `valueConstraints`-referenced options (and compare-field options when present) in the declared-key set used by `validateUnknownKeys()`.
- Finding 2: stop normalizing every `Number` through `double` for ordering comparisons.

## Nice to clean up
- Finding 3: update the `OptionRule` example to a real API shape.

Happy to re-review once you push the next revision.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch against the full current validation path.

# What this PR solves
- User pain: connector authors can declare required/optional/exclusive relationships today, but cannot describe richer value constraints directly in `OptionRule`.
- Fix approach: add declarative `ConditionOperator` / `ConditionEvaluators` support, wire it into `ConfigValidator`, and expose the metadata through the REST response.
- One-line summary: the two framework-level blockers from my previous round are fixed on the latest head, and I did not find a new blocking issue on the current code.

# Runtime path I checked
```text
connector config validate
      -> validate required/optional/single-choice rules
      -> validate valueConstraints
          -> ConditionEvaluators.evaluate(...)
  -> ConfigValidator.validateUnknownKeys(...)
      -> collectDeclaredKeys(rule)
      -> includes valueConstraints + compareOption keys
```

# Re-review result
- The old unknown-key blocker is fixed: `collectDeclaredKeys()` now includes value-constraint keys and compare-option keys, so constraint-only options are no longer rejected as unknown.
- The old large-integer blocker is fixed for the real timestamp/offset-style path I called out before: `compareNumbers()` now avoids routing `Long` comparisons through `double`.
- The Javadoc example was also corrected to use real APIs.

# Tests / CI
- The new regression coverage around `ConfigValidatorTest` is much better now.
- The current GitHub `Build` is green.
- I did not see a flaky-test pattern in the added UTs.

### Conclusion: can merge
1. Blocking items

- If this condition DSL grows further later, it may be worth adding one focused regression around decimal-style numeric comparison semantics as a future hardening step, but that is not a blocker for this PR.

yzeng1618:
> @yzeng1618 Can you help confirm if the support of OptionRulesService to return transform has a negative impact on AI generated configuration? Thank you very much.

 I took a rough look at the CLI (#10789) consumption side and wanted to double-check two points with you — I may not have the full picture, so please help confirm.
First, on transform: in connectors.py, get_connector_detail() currently defaults query_types to ["source", "sink"] only, so transform doesn't seem to be requested at this stage. So even if OptionRulesService now returns transform, the CLI side likely won't use it yet — supporting it properly would probably need some follow-up adaptation on the CLI side.
Second, on the new valueConstraints field in this PR: from what I can tell it's just returned in the response for now, and the CLI side doesn't yet have parsing/validation logic to consume it. So it shouldn't affect the existing AI config generation, but I'd appreciate it if you could confirm whether I'm missing anything @SEZ9

DanielLeens:
Thanks for checking the downstream CLI side as well.

Your read is consistent with what I see on the current PR head `d6b235ac9364848695b5d1d2587e9ab13c342d37`: `OptionRulesService` returning transform metadata is additive on the service side, while the current CLI path still requests only `source` / `sink`, and the new `valueConstraints` payload is not being consumed there yet.

Since there is still no new commit after the latest Daniel review on this head, I am keeping this as a reply-only follow-up rather than reopening a full review. From Daniel's side I am not adding a new blocker based on this CLI discussion on the current revision.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch again against the full current validation path.

# What this PR solves
- User pain: `OptionRule` can describe presence rules today, but not richer constraints like `> 0` or `field A < field B` in a declarative way.
- Fix approach: add `ConditionOperator` / `ConditionEvaluators`, wire `valueConstraints` into `ConfigValidator`, and expose the constraint metadata through REST.
- One-line summary: the two framework-level blockers from my earlier rounds are fixed on the latest head, and I did not find a new blocking issue.

# Runtime path I checked
```text
connector config validate
      -> required / optional / conditional checks
      -> valueConstraints checks
          -> ConditionEvaluators.evaluate(...)

unknown key check
  -> ConfigValidator.validateUnknownKeys(...)
      -> collectDeclaredKeys(rule)
      -> now includes valueConstraints + compareOption keys

REST metadata exposure
  -> OptionRulesService
  -> OptionRuleResponse
```

# Re-review result
- The old unknown-key blocker is fixed: keys referenced only from `valueConstraints` are now included in the declared-key set.
- The old large-integer precision blocker is fixed: `Long` comparisons no longer route through `double`.
- The new docs/tests line up with the current runtime path from what I checked.

# Tests / stability
- The updated unit coverage is much stronger now.
- I did not see a flaky-test pattern in the added or changed tests.
- The current GitHub `Build` is still running, but I do not have a source-level blocker to add on top of the latest code.

### Conclusion: can merge
1. Blocking items

- If this DSL grows further later, it may be worth documenting the intended behavior for mixed numeric types more explicitly, but that is not a blocker for this PR.

From the code-review side this looks good now.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch again, this time focusing on the actual delta after my June 1 review instead of assuming the earlier LGTM still held.

### What problem this PR solves
- User pain: `OptionRule` can express presence rules today, but not richer value-level constraints like `> 0`, `not blank`, or `start < end` as part of the contract itself.
- Fix approach: add declarative `Condition` / `ConditionOperator` / `Conditions` support, run those constraints through `ConfigValidator`, and expose the metadata through the REST option-rule path.
- One-line summary: this turns value validation into a first-class part of the SeaTunnel config contract instead of leaving it scattered in ad hoc runtime checks.

### Full runtime path I checked
```text
plugin defines OptionRule
  -> required / optional / conditional declarations
  -> valueConstraints attached to the rule

user config validation
     -> required / optional / conditional checks
     -> valueConstraints checks
        -> ConditionEvaluators.evaluate(...)
           -> numeric / string / collection / field-to-field evaluation

unknown-key guard
  -> ConfigValidator.validateUnknownKeys(...)
     -> collectDeclaredKeys(rule)
     -> collectConditionKeys(condition)

REST metadata exposure
  -> OptionRulesService
  -> OptionRuleResponse
```

### Core logic re-review
I rechecked the current implementations in:
- `seatunnel-api/.../Condition.java`
- `seatunnel-api/.../ConditionEvaluators.java`
- `seatunnel-api/.../ConditionOperator.java`
- `seatunnel-api/.../ConfigValidator.java`
- `seatunnel-api/.../OptionRule.java`
- `seatunnel-engine/.../OptionRulesService.java`
- `seatunnel-engine/.../OptionRuleResponse.java`

The two framework-level blockers I raised in earlier rounds still remain fixed on the current head:
- keys referenced only from `valueConstraints` are still collected into the declared-key set, so they are no longer rejected by `validateUnknownKeys()`;
- large integer comparisons still avoid the old “force everything through `double`” precision problem.

The latest delta is mostly API cleanup (`Condition` + `Conditions` usage), broader `ConfigValidatorTest` coverage, and doc refresh. After tracing the real validation path again, I did not find a new source-level correctness blocker in the current code.

### Compatibility / side effects
- Compatibility: fully compatible from my review perspective. This is additive behavior, not a breaking change to existing option definitions.
- Performance / side effects: the extra work stays on the config-validation path, so I do not see a CPU / memory / concurrency blocker.
- Error handling: the wrapped validation messages in `ConditionEvaluators.evaluate()` are now easier to diagnose.

### Tests / stability
- The expanded UT coverage is materially better now, especially around declared-key collection, cross-field comparisons, and applicability rules.
- I did not see a flaky-test pattern in the new unit tests. No `Thread.sleep`, weak async assertions, or environment-sensitive resource usage showed up in the added test code.
- Stability rating: **Stable**.

### Existing review context
I also checked the current community context again. I do not have another maintainer/reviewer blocker to add on top of the latest code, and my current conclusion is still aligned with the earlier “source path looks good” direction.

### Current blocker outside the source path
The remaining gate is CI, not a new code-path issue from my side:
- GitHub still shows `Build` as failed on the current head.
- From the current check metadata I could confirm at least `Run / unit-test (11, ubuntu-latest)` failed, and several sibling lanes were cancelled afterward.
- When I checked, the contributor-side workflow was still not exposing the final failed-step log yet, so I cannot honestly pin that failure to a specific new source regression from this revision.

### Merge conclusion
**Conclusion: can merge after fixes.**

1. Blocking items
- Please get the current `Build` back to green first, starting with the failing `Run / unit-test (11, ubuntu-latest)` lane.

2. Suggested follow-up
- Once that failing lane exposes its final stack trace, we can tell whether this is just CI noise or a real regression from the latest revision. From the code-review side, I do not have a new blocker beyond the red build gate.

Overall, the current head still looks coherent on the real validation path, and the previous Daniel correctness blockers remain closed.

davidzollo:
## Re-review (after commits 728e3ee...894091d)

Apologies for being less thorough on the first pass — re-evaluating the whole PR from scratch as required, and acknowledging that I had under-credited the depth of the rewrite. Local verification this time: **146/146 tests pass** in `seatunnel-api` (`ConfigValidatorTest` 143 + `ConditionTest` 3), `spotless:check` clean.

### What this PR really does (concrete example)

Today, "port must be in [1, 65535]" or "start_ts must be less than end_ts" lives as imperative `if/else` inside `*Config.java`, invisible to REST/CLI/Web UI/AI tooling, and reported one-at-a-time. After this PR:

```java
import static org.apache.seatunnel.api.configuration.util.Conditions.*;

OptionRule.builder()
    .required(PORT, greaterOrEqual(PORT, 1).and(lessOrEqual(PORT, 65535)))
    .required(HOST, notBlank(HOST))
    .required(START_TS, END_TS, lessThanField(START_TS, END_TS))
    .build();
```

`ConfigValidator` runs every constraint at job-submission time, aggregates **all** failures (structural + value) into a single `Option validation failed (N errors): [1] ... [2] ...` message, REST `/option-rules` exposes both the symbolic `compareOperator` and the structured `conditionOperator` / `conditionOperatorCategory` so frontends/AI don't need to string-match.

### Verifying the previous round's blockers were fixed

All 8 items I raised in the previous review are addressed:

| # | Previous finding | Status | How |
|---|---|---|---|
| 1 | `optional(start, end, lessThanField(start, end))` threw `Cannot compare null` when one side missing | **Fixed** | `isConstraintApplicable` now splits the chain at OR boundaries and requires AND-segments to be fully present; tests `testOptionalCrossFieldOnlyStartPresent` / `OnlyEndPresent` / `NoFalsePositive` pin it down |
| 2 | `greaterOrEqual(PORT,1).or(notBlank(HOST))` broke short-circuit when PORT null | **Fixed** | All numeric/FIELD evaluators are now null-safe (`v != null && ...`); covered by `testOrChainNumericNullWithStringFallback` (4 scenarios) |
| 3 | `greaterThan(longOpt, 0)` raised `Numeric type mismatch` because `int` vs `Long` | **Fixed** | `compareNumberValues` now normalizes via `BigDecimal` / `double` / `long` three-tier; `testNumericTypeMismatchIntVsLong` / `DoubleVsInt` cover it |
| 4 | REST `ConditionNode` only had symbol string; AI/frontend forced to string-match | **Fixed** | Added `conditionOperator` (e.g. `GREATER_OR_EQUAL`) + `conditionOperatorCategory` (e.g. `NUMERIC`) |
| 5 | Raw-type `new Condition(...)` factories | **Fixed** | Factories moved to `Conditions.java` (clean generics); narrow `Condition.java` to data + chain logic |
| 6 | Evaluator exceptions dropped option key | **Fixed** | `ConditionEvaluators.evaluate` wraps inner exceptions as `Failed to evaluate constraint '%s' on option '%s': %s` |
| 7 | `Condition.of(opt, null)` API narrowed without `incompatible-changes.md` entry | **Fixed** | Entry added in both `docs/{en,zh}/introduction/concepts/incompatible-changes.md` |
| 8 | Test coverage gaps (optional cross-field single-side, OR with numeric, type mismatch) | **Fixed** | Test count went from 32 → **143** in `ConfigValidatorTest`; nearly every scenario I called out has an explicit test |

Bonus improvement that wasn't in the previous review but is genuinely production-grade: `collectErrors` aggregates structural + value errors into one `OptionValidationException` instead of fail-fast. Big UX win for users debugging large HOCON configs.

### New findings on this pass

#### Issue 1 (medium) — Asymmetric cross-field: `optional(head) + required(compareField)` makes the optional head implicitly required

`isConstraintApplicable` collects `head + compareOption + all chain nodes` and returns `true` as soon as **any** of them is `AbsolutelyRequiredOptions`. Consider:

```java
.required(START_TS)
.optional(MAX, lessThanField(MAX, START_TS))
```

- `allOptions = {MAX, START_TS}`
- `START_TS` is absolutely required → first-stage hit → applicable
- `FIELD_LESS_THAN(null, START_TS)` → returns `false`
- Error reported: `constraint: 'max' < 'start_ts'`

User's clear intent: "MAX is optional; if set, must be less than START_TS." Actual behavior: MAX is silently required. The symmetric case (`required(head) + optional(compareField)`) is already pinned down by `testRequiredPrimaryWithAbsentOptionalCompareField` as "must fail", which is a consistent but easy-to-trip design choice.

**Suggested fix (Option A, recommended):** keep "required head forces applicable" but only inspect the **head** node, not the entire option set. That makes both directions symmetric — the head decides. Adjust `testRequiredPrimaryWithAbsentOptionalCompareField` accordingly (END_TS is optional, so the constraint should skip and let structural validation handle the absence).

**Option B (conservative):** keep both semantics, but add a Javadoc warning on every cross-field factory in `Conditions.java` stating "if used together with `optional(head, ...)`, the compareOption must also be optional".

#### Issue 2 (low) — `MetadataExportCommand` didn't pick up `conditionOperator` / `conditionOperatorCategory`

`OptionRulesService.toConditionNode` now exposes the structured operator metadata for the REST API, but `MetadataExportCommand.exportCondition` still only emits `compareOperator` (symbol string). The PR description specifically calls out AI-assisted config generation, and the CLI export is exactly the surface AI tools consume offline. Quick fix: mirror the same fields via `op.name()` / `op.getCategory().name()`.

#### Issue 3 (low) — Docs hint at length/size constraints that the PR didn't ship

The previous revision had `LENGTH_*`, `COLLECTION_SIZE_*`, `FIELD_EQUAL/NOT_EQUAL`, `FIELD_SIZE_EQUAL`, etc. Those have been (rightly) trimmed — the surface is now 17 operators in 4 categories. But the architecture doc still uses "string length limit" / "fixed collection size" as motivating examples. Suggest either narrowing the docs or adding a "currently supported operators" callout in `Conditions.java`'s class-level Javadoc so users know what's actually available without grepping.

#### Issue 4 (medium) — `OptionValidationException` message format changed; no `incompatible-changes.md` entry

Old format (per error):
```
ErrorCode:[API-02], ErrorDescription:[Option item validate failed] - There are unconfigured options, the options('host') are required.
```

New format (aggregated):
```
ErrorCode:[API-02], ErrorDescription:[Option item validate failed] - Option validation failed (2 errors):
  [1] option: 'host'
      type: required
      constraint: required option is not configured
  [2] option: 'port'
      type: value
      constraint: 'port' >= 1
```

This is strictly a better UX, but any consumer doing `getMessage().contains("are required")` will break. Connector unit tests doing string assertions on the message (the PR itself rewrote a handful of them) will fail. Worth adding a second `incompatible-changes.md` entry next to the `Condition.of(opt, null)` one.

#### Issue 5 (low) — `Condition.*` static factories were removed without a deprecated bridge

`Condition.greaterThan(...)`, `Condition.notBlank(...)`, etc. moved to `Conditions.*`. Anyone using them outside the main repo will get a compile error. Either:
- Keep `@Deprecated public static <T> Condition<T> greaterThan(...) { return Conditions.greaterThan(...); }` bridges, **or**
- Document the rename in `incompatible-changes.md` alongside the others.

#### Issue 6 (low) — `compareNumbers` error message includes raw values

```java
throw new OptionValidationException(
    "Cannot compare null values in numeric comparison: left=%s, right=%s", a, b);
```

For numeric options this is fine, but if a connector ever wires up a sensitive option as `Number`, the raw value leaks into logs. Safer to swap to `leftType=%s, rightType=%s`.

#### Issue 7 (low) — `Conditions.java` class Javadoc has a truncated example

```java
 * <pre>{@code
 * static *
 * OptionRule.builder()
```

Looks like an incomplete `import static org.apache.seatunnel.api.configuration.util.Conditions.*;` line.

#### Issue 8 (low) — `extractInnerMessage` couples to `OptionValidationException` text format

Splitting on `" - "` to strip the `ErrorCode:[API-02] - ` prefix is fine today, but ties the evaluator wrapper to an implementation detail of the exception. A `getRawMessage()` accessor on `OptionValidationException` would be cleaner. Not urgent.

#### Issue 9 (low) — Missing test for the Issue 1 scenario

`testRequiredPrimaryWithAbsentOptionalCompareField` pins down `required(head) + optional(compareField)`. The mirror case (`optional(head) + required(compareField)`, head absent) is not explicitly tested. Worth adding one to make the chosen semantic regression-proof.

### Issue summary

|---|---|---|---|
| 1 | `optional(head) + required(compareField)` implicitly forces head to be required | `ConfigValidator#isConstraintApplicable` | medium |
| 4 | `OptionValidationException` message format change not documented | `ConfigValidator#validate(OptionRule)` | medium |
| 2 | `MetadataExportCommand` doesn't expose `conditionOperator` / `conditionOperatorCategory` | `MetadataExportCommand#exportCondition` | low |
| 3 | Docs imply length/size constraints that aren't shipped | `Conditions.java`, configuration-and-option-system docs | low |
| 5 | `Condition.*` static factories removed without `@Deprecated` bridges or doc | `Condition.java` | low |
| 6 | `compareNumbers` error message leaks raw values | `ConditionEvaluators#compareNumbers` | low |
| 7 | `Conditions.java` class Javadoc has a truncated example | `Conditions.java` | low |
| 8 | `extractInnerMessage` couples to exception text format | `ConditionEvaluators#extractInnerMessage` | low |
| 9 | Missing test for `optional(head) + required(compareField) + head absent` | `ConfigValidatorTest` | low |

### Verification commands

| Command | Result | Note |
|---|---|---|
| `./mvnw -B -pl seatunnel-api spotless:check -nsu -Dmaven.gitcommitid.skip=true` | PASS | clean |
| `./mvnw -B -pl seatunnel-api test -Dtest=ConfigValidatorTest,ConditionTest -nsu -Dmaven.gitcommitid.skip=true` | **PASS (146/146)** | |
| `./mvnw -B -pl seatunnel-engine/seatunnel-engine-server test -Dtest=OptionRulesServiceTest -am ...` | not run | `seatunnel-config-shade` failed to compile (missing `ConfigValue/Entry` symbols, unrelated to this PR — typical when shade-relocate hasn't run) |
| `git grep "Condition\.of\([^,]*, *null"` in main repo | none | no current production callers, backward-compat risk for `Condition.of(opt, null)` narrowing is effectively zero |

### Conclusion: ready to merge after two doc additions; the rest can be follow-ups

No runtime blockers remain. The remaining 9 items are either documentation gaps (4, 5), low-severity polish (2, 3, 6, 7, 8, 9), or one semantic edge case that I think deserves discussion before code change (1).

**Recommended pre-merge (≈30 min):**
- Add a single `incompatible-changes.md` block covering (a) `OptionValidationException` message format change and (b) `Condition.*` static factory relocation. Or alternatively keep `@Deprecated` bridges on `Condition.*`.

**Recommended follow-up PR:**
- Decide on Issue 1's semantic (Option A: head-only `isConstraintApplicable`, or Option B: Javadoc warning on cross-field factories).
- Mirror the structured operator metadata in `MetadataExportCommand` (Issue 2).
- Tighten docs scope to currently-shipped operators (Issue 3).
- Polish 6/7/8 + add the missing test from Issue 9.

### Overall assessment

This is a meaningful step up from the previous revision. The contributor turned every one of the 8 prior findings into a precise fix backed by tests, plus added error aggregation and structured REST metadata that weren't requested but are genuinely production-grade. The 17-operator scope is a deliberate, healthy narrowing relative to the prior 30 — better to ship a tight set with full evaluator + REST + AI parity than 30 operators in inconsistent states.

Solid work. Looking forward to seeing this land.

nzw921rx:
@davidzollo Thank you for your review. 

I have made the following fixes based on the results, and welcome your review at any time

- `isConstraintApplicable` now only checks the **head** option for required-ness — compareField in cross-field operators no longer forces applicability. This makes `optional(head) + required(compareField)` behave symmetrically: head absent → constraint skipped.

- `MetadataExportCommand.exportCondition` now emits `conditionOperator` and `conditionOperatorCategory`, aligned with the REST API.

-  Fixed truncated Javadoc in `Conditions.java`; added supported operator inventory (17 operators, 4 categories).

- `incompatible-changes.md` entry for `OptionValidationException` structured aggregation format.

- Skipped — `Condition.*` factories are new code (never released), no compat entry needed.

- `compareNumbers` no longer leaks raw values in error messages.

- `OptionValidationException` gained `getRawMessage()`; `extractInnerMessage` replaced with direct accessor.

- Added 3 new tests covering `optional(head) + required(compareField)` scenarios (head absent / both present / violation).

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch again against the full current validation path.

# What this PR solves
- User pain: `OptionRule` can describe presence rules today, but not richer constraints like `> 0` or `field A < field B` in a declarative way.
- Fix approach: add `ConditionOperator` / `ConditionEvaluators`, wire `valueConstraints` into `ConfigValidator`, and expose the constraint metadata through REST.
- One-line summary: the two framework-level blockers from my earlier rounds are fixed on the latest head, and I did not find a new blocking issue.

# Runtime path I checked
```text
connector config validate
      -> required / optional / conditional checks
      -> valueConstraints checks
          -> ConditionEvaluators.evaluate(...)

unknown key check
  -> ConfigValidator.validateUnknownKeys(...)
      -> collectDeclaredKeys(rule)
      -> now includes valueConstraints + compareOption keys

REST metadata exposure
  -> OptionRulesService
  -> OptionRuleResponse
```

# Re-review result
- The old unknown-key blocker is fixed: keys referenced only from `valueConstraints` are now included in the declared-key set.
- The old large-integer precision blocker is fixed: `Long` comparisons no longer route through `double`.
- The new docs/tests line up with the current runtime path from what I checked.

# Tests / stability
- The updated unit coverage is much stronger now.
- I did not see a flaky-test pattern in the added or changed tests.
- The current GitHub `Build` is still queued, but I do not have a source-level blocker to add on top of the latest code.

### Conclusion: can merge
1. Blocking items

- If this DSL grows further later, it may be worth documenting the intended behavior for mixed numeric types more explicitly, but that is not a blocker for this PR.

From the code-review side this looks good now.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch again against the full current validation path, including the new test-only follow-up commit.

# What this PR solves
- User pain: `OptionRule` can describe presence rules today, but not richer declarative value constraints like `> 0`, `not blank`, or `field A < field B`.
- Fix approach: add `ConditionOperator` / `ConditionEvaluators`, wire `valueConstraints` into `ConfigValidator`, and expose the metadata through the REST option-rule path.
- One-line summary: the framework-level correctness blockers from my earlier rounds remain fixed on the current head, and the latest commit only tightens the validation test assertion.

# Runtime chain I checked
```text
plugin defines OptionRule
  -> required / optional / conditional declarations
  -> valueConstraints attached to the rule

user config validation
     -> required / optional / conditional checks
     -> valueConstraints checks
        -> ConditionEvaluators.evaluate(...)

REST metadata exposure
  -> OptionRulesService
  -> OptionRuleResponse
```

# Re-review result
- `ConfigValidator.java`, `ConditionEvaluators.java`, `OptionRule.java`, and the REST metadata path still line up with the intended design.
- The previous Daniel blockers around declared-key collection and large-integer comparison remain fixed on the current head.
- The latest commit only changes `SeaTunnelConfValidateCommandTest.java:101-104` to assert the current unified validation message shape, which is consistent with the runtime behavior.

# Tests / stability
- The updated UT coverage remains strong.
- I do not see a flaky-test pattern in the changed tests.
- The current PR checks are green from the metadata I reviewed.

### Conclusion: can merge
1. Blocking items

- None.

From the code-review side this looks good to merge.

davidzollo (approved):
+1
LGTM

DanielLeens:
Thanks for the detailed follow-up. I rechecked the PR metadata before replying: the current head is still the same revision I last reviewed, so I cannot mark the items as resolved yet based on the comment alone.

Your summary makes sense, but Daniel needs the actual new commit on the PR branch before doing another full review. Once the updated head is pushed, please ping me again and I will re-review the latest code from scratch rather than assuming the comment matches the branch state.

yzeng1618 (approved):
LGTM  +1

Issue #4544, opened 2023-04-11
[Feature][Connector-V2][Docs] Unified format for connector-v2 docs


zhilinli123:
# MySQL

> JDBC Mysql Sink Connector

## Support Connector Version

- 5.5/5.6/5.7/8.0

## Support Those Engines

> Spark<br/>
> Flink<br/>
> Seatunnel Zeta<br/>

## Key Features

- [x] [exactly-once](../../concept/connector-v2-features.md)
- [x] [cdc](../../concept/connector-v2-features.md)

> Use `Xa transactions` to ensure `exactly-once`. So only support `exactly-once` for the database which is
> support `Xa transactions`. You can set `is_exactly_once=true` to enable it.

## Description

Write data through jdbc. Support Batch mode and Streaming mode, support concurrent writing, support exactly-once
semantics (using XA transaction guarantee).

## Supported DataSource Info

| Datasource |                    Supported Versions                    |          Driver          |                  Url                  |                                   Maven                                   |
|------------|----------------------------------------------------------|--------------------------|---------------------------------------|---------------------------------------------------------------------------|
| Mysql      | Different dependency version has different driver class. | com.mysql.cj.jdbc.Driver | jdbc:mysql://localhost:3306:3306/test | [Download](https://mvnrepository.com/artifact/mysql/mysql-connector-java) |

## Database Dependency

> Please download the support list corresponding to 'Maven' and copy it to the '$SEATNUNNEL_HOME/plugins/jdbc/lib/' working directory<br/>
> For example Mysql datasource: cp mysql-connector-java-xxx.jar $SEATNUNNEL_HOME/plugins/jdbc/lib/

## Data Type Mapping

|                                                          Mysql Data type                                                          |                                                                 Seatunnel Data type                                                                 |
|-----------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------|
| BIT(1)<br/>INT UNSIGNED                                                                                                           | BOOLEAN                                                                                                                                             |
| TINYINT<br/>TINYINT UNSIGNED<br/>SMALLINT<br/>SMALLINT UNSIGNED<br/>MEDIUMINT<br/>MEDIUMINT UNSIGNED<br/>INT<br/>INTEGER<br/>YEAR | INT                                                                                                                                                 |
| INT UNSIGNED<br/>INTEGER UNSIGNED<br/>BIGINT                                                                                      | BIGINT                                                                                                                                              |
[...]

PR #10988, opened 2026-06-01
[Feature][Connector-V2] Add pgvector type support in PostgreSQL JDBC connector

CosmosNi:
- Add vector(N) type mapping in PostgresTypeConverter (vector <-> FLOAT_VECTOR)
- Add FLOAT_VECTOR read/write in PostgresJdbcRowConverter
- Parse PGobject string format [f1,f2,...] for read, construct PGobject for write
- Add unit tests for bidirectional type mapping

### Check list


DanielLeens (commented):
Thanks for working on this. I reviewed the full current head locally and traced the real PostgreSQL query-schema path instead of only checking the new vector read/write branches.

# What this PR solves
- User pain: the PostgreSQL JDBC connector cannot currently read/write pgvector columns.
- Fix approach: add vector handling in the PostgreSQL row converter and type converter, plus a pgvector E2E.
- One-line summary: the read/write direction is right, but the latest head still does not preserve `vector(n)` dimensions on the real `query source -> auto-create sink` path.

# Runtime path I checked
```text
query source schema inference
  -> JdbcDialectTypeMapper.mappingColumn(ResultSetMetaData)
  -> PostgresTypeMapper.mappingColumn(...) [35-52]
      -> columnType = metadata.getColumnTypeName(index)
  -> PostgresTypeConverter.convert(...) [247-260]
      -> only parses dimension when columnType looks like "vector(3)"

auto-create sink
  -> uses the inferred column/sourceType to generate target DDL

current E2E
  -> JdbcPostgresPgVectorIT.initializeJdbcTable()
      -> pre-creates the sink table with embedding vector(3)
  -> test only verifies read/write against that already-correct sink schema
```

# Findings
1. **Blocker: the query-metadata path can still lose the vector dimension.**
   - `PostgresTypeMapper.java:35-52` only forwards `metadata.getColumnTypeName(colIndex)` into `BasicTypeDefine.columnType`.
   - `PostgresTypeConverter.java:247-260` only preserves the dimension when that string is literally `vector(n)`.
   - So if the normal JDBC metadata path gives you `vector` instead of `vector(3)`, the dimension is already gone before sink DDL generation.

2. **The new E2E does not cover the risky path above.**
   - `JdbcPostgresPgVectorIT.java:70-76,117-135,159-168` pre-creates the sink table.
   - That means CI can go green while the real auto-create-sink path is still incomplete.

# Test stability
- Rating: **Stable**
- Why:
  - I did not see a flaky-test pattern in the added pgvector E2E itself.
  - The gap is coverage realism, not timing instability.

1. Blocking items
- Finding 1: please close the real query-source schema-inference path so `vector(n)` dimensions survive into sink-side schema generation.

- Finding 2: add one E2E that does **not** pre-create the sink table, so the auto-create path is actually exercised.

I like the direction here. The remaining blocker is not the vector value conversion itself, but the schema contract: on the latest head, the most important production path still does not prove that `vector(n)` survives end to end.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch and traced the exact path that I blocked earlier: query-source metadata inference into the auto-create sink DDL.

# What this PR solves
- User pain: the PostgreSQL JDBC connector needs to support pgvector columns without losing the `vector(n)` dimension on the real source-to-sink path.
- Fix approach: enrich the PostgreSQL metadata path with the vector dimension from `pg_attribute.atttypmod`, pass the live JDBC connection into that mapper path, and add an E2E that does **not** pre-create the sink table.
- One-line summary: the latest head fixes the earlier schema-contract blocker for me, and I do not see a new source-level issue on the pgvector path.

# Runtime path I checked
```text
query source schema inference
  -> PostgresTypeMapper.mappingColumn(metadata, colIndex, connection) [75-108]
      -> if JDBC reports pgvector without dimension
      -> query pg_attribute.atttypmod to recover N
      -> return sourceType vector(N)

catalog / schema path
  -> CatalogUtils.getCatalogTable(..., connection)
      -> passes the live connection into the type mapper

auto-create sink validation
  -> JdbcPostgresPgVectorIT.testPgVectorAutoCreateSink() [152-194]
      -> does not pre-create the sink table
      -> verifies both data correctness and final sink column type via format_type(atttypid, atttypmod)
```

# Re-review result
- The earlier blocker is fixed for me now:
  - `PostgresTypeMapper.java:75-108` no longer relies only on plain `getColumnTypeName()` for pgvector dimension.
  - `CatalogUtils.java` now threads the live connection into the mapping path, which is what makes the extra dimension lookup usable on the real query-source route.
  - the new E2E finally exercises the risky path I called out before instead of pre-creating the sink table.
- I did not find a new reopened source-level blocker on the latest head.

# Tests / stability
- Rating: **Stable**
- Why:
  - the new E2E now covers the real auto-create sink path that mattered most.
  - I did not see a flaky-test pattern in the pgvector additions.
- The current GitHub `Build` is still queued, but I do not have a source-level blocker to add on top of the latest code.

### Conclusion: can merge
1. Blocking items

- None from my side on the current pgvector path.

I like where this landed. The latest revision closes the schema-dimension gap that was blocking the previous round.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from scratch again and traced the exact pgvector path that mattered in the previous round. I found one blocker on the current revision.

# What this PR solves
- User pain: the PostgreSQL JDBC connector needs to support pgvector columns without losing the `vector(n)` dimension on the source-to-sink path.
- Fix approach: add vector read/write support, recover vector dimensions for the metadata path, and verify auto-create sink behavior.
- One-line summary: the core direction is still right, but the latest head removes the only end-to-end coverage for the exact query-source metadata path that this PR had to prove.

# Full runtime chain I checked
```text
query source schema inference
  -> JdbcCatalogUtils.getCatalogTable(connection, sqlQuery, jdbcDialect)
     -> CatalogUtils.getCatalogTable(resultSetMetaData, typeMapper, sqlQuery, connection)
        -> PostgresTypeMapper.mappingColumn(metadata, colIndex, connection)
           -> recover vector dimension from pg_attribute

auto-create sink
  -> uses the inferred sourceType / scale to generate target DDL
```

# Findings
Issue 1: the latest E2E no longer exercises the query-source metadata path that carries `vector(n)`
- Location:
  - `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-3/src/test/resources/jdbc_postgres_pgvector_auto_create_sink.conf:29`
  - `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/utils/JdbcCatalogUtils.java:407-420`
  - `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeMapper.java:75-107`
- Problem description:
  - the head commit changed the auto-create sink E2E source from `query = ...` to `table_path = ...`.
  - with `table_path`, the test now goes through `CatalogUtils.getCatalogTable(connection, tablePath, jdbcDialect)` at `JdbcCatalogUtils.java:407-409`.
  - that bypasses the exact query-metadata branch at `JdbcCatalogUtils.java:415-420` which is the branch that threads the live connection into `PostgresTypeMapper.mappingColumn(..., connection)` for vector-dimension recovery.
- Potential risk:
  - the PR is specifically changing the query-source metadata inference contract for pgvector, but the latest E2E no longer proves that branch end to end.
  - that means a regression in the `query -> metadata -> auto-create sink` path can slip through while CI still stays green on the current test.
- Best improvement:
  - please restore a `query`-based auto-create E2E for this scenario, or add a second E2E that keeps the source on `query` and verifies the sink lands as `vector(N)`.
- Severity: High
- Raised by others: No

# Test stability
- Rating: Stable
- I do not see a flaky-test pattern in the added pgvector tests themselves.
- The current issue is coverage realism, not timing instability.

1. Blocking items
- Issue 1: restore end-to-end coverage for the real `query source -> auto-create sink` path before merge.

- None beyond restoring that coverage.

I still like the overall direction here. The blocker is not the vector conversion code path itself; it is that the latest head stopped proving the exact metadata path this PR needed to make safe.

CosmosNi:
@dybyte @chl-wxp CC

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch and traced the exact path that mattered in the previous round: `query source -> schema inference -> auto-create sink`.

# What this PR fixes
- User pain: pgvector support is only really complete if the source-side schema inference can preserve `vector(N)` and carry that exact dimension into sink auto-creation.
- Fix approach: the query metadata path now passes the live JDBC connection into the PostgreSQL type mapper, so it can recover `vector(N)` from `pg_attribute`, and the E2E now verifies the sink table is auto-created with the final `vector(N)` column type.
- In one sentence: the latest head closes the schema-contract gap on the real query-source auto-create path, not just the value conversion path.

Simple example:
- Before this fix, a PostgreSQL query source could still infer `embedding` as plain `vector`, and the auto-created sink schema had no proof that the original `vector(3)` dimension survived.
- On the current head, the mapper can recover the dimension from PostgreSQL metadata, and the E2E checks the final sink column through `format_type(atttypid, atttypmod)`.

# Full runtime chain I checked
```text
query source schema inference
  -> JdbcCatalogUtils.getCatalogTable(tableConfig, connection, jdbcDialect) [380-405]
      -> when both table_path and query are configured
      -> tableOfQuery = getCatalogTable(connection, tableConfig.getQuery(), jdbcDialect) [390-391]
          -> jdbcDialect.getResultSetMetaData(connection, sqlQuery)
          -> CatalogUtils.getCatalogTable(connection, resultSetMetaData, typeMapper, sqlQuery)
              -> PostgresTypeMapper.mappingColumn(metadata, colIndex, connection, sqlQuery) [70-100]
                  -> if nativeType = vector
                  -> resolveTableName(...) [107-129]
                  -> queryVectorTypeFromPgAttribute(...) [136-163]
                  -> rebuild columnType as vector(N)

auto-create sink validation
  -> jdbc_postgres_pgvector_auto_create_sink.conf [29-30]
      -> source keeps both table_path and query
      -> sink enables generate_sink_sql=true
  -> JdbcPostgresPgVectorIT.testPgVectorAutoCreateSink() [152-193]
      -> does NOT pre-create the sink table
      -> runs the job
      -> verifies the final sink column type with pg_attribute.format_type(...)
```

# Re-review result
- The blocker from the previous round is fixed for me now.
- `JdbcCatalogUtils.java:380-405` no longer drops the query-side schema path when `table_path` and `query` are both present; it still builds `tableOfQuery` from the live query metadata path.
- `PostgresTypeMapper.java:70-100` now has the missing recovery step for `vector(N)` when `ResultSetMetaData` only reports `vector`.
- The new E2E is covering the real risky path and checks the final sink schema instead of stopping at a pre-created sink table, see `JdbcPostgresPgVectorIT.java:152-193` and `jdbc_postgres_pgvector_auto_create_sink.conf:29-30`.
- I do not see a new source-level blocker on the latest head.

# Compatibility
- Fully compatible from my review.
- No API, config, default-value, protocol, or serialization contract break was introduced here.

# Performance / side effects
- The extra PostgreSQL metadata lookup is a one-time schema-inference cost, not a hot data path cost.
- Resource handling looks fine: the added metadata query uses local `PreparedStatement` / `ResultSet` scopes with proper cleanup.

# Error handling / logging
- The fallback behavior is reasonable: if the pgvector metadata query fails, the mapper logs at debug and falls back to the original metadata path.
- I did not find a new logging or exception-handling problem on the current head.

# Test stability assessment
- Rating: stable
- Basis:
  - the new E2E validates the final sink schema directly instead of relying on weak log signals;
  - I did not see a flaky-test pattern such as fixed sleeps, shared mutable global state, or hard-coded local ports in the pgvector additions.

### Conclusion: can merge

1. Blocking items

2. Suggested follow-up
- None from my side on the current pgvector path.

This looks ready to me now. The latest revision closes the exact schema-inference gap that was blocking the previous round, and the current upstream `Build` is green as well.

seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeMapper.java
```diff
@@ -51,4 +59,107 @@ public Column mappingColumn(ResultSetMetaData metadata, int colIndex) throws SQL
...
+     * Queries pg_attribute to get the full vector type string (e.g. "vector(3)") for a specific
+     * column.
+     */
+    private String queryVectorTypeFromPgAttribute(
+            Connection connection, String tableName, String columnName) {
+        String sql =
+                "SELECT format_type(a.atttypid, a.atttypmod) "
+                        + "FROM pg_attribute a "
+                        + "JOIN pg_class c ON a.attrelid = c.oid "
+                        + "WHERE c.relname = ? AND a.attname = ? AND a.attnum > 0";
```
chl-wxp: Does this query not use schema as a parameter? Can uniqueness be ensured?

seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeMapper.java
```diff
@@ -51,4 +59,107 @@ public Column mappingColumn(ResultSetMetaData metadata, int colIndex) throws SQL
...
+    @Override
+    public Column mappingColumn(
+            ResultSetMetaData metadata, int colIndex, Connection connection, String sqlQuery)
+            throws SQLException {
+        String nativeType = metadata.getColumnTypeName(colIndex);
+
+        if (PostgresTypeConverter.PG_VECTOR.equalsIgnoreCase(nativeType) && connection != null) {
+            String tableName = resolveTableName(metadata, colIndex, sqlQuery);
+            if (tableName != null && !tableName.isEmpty()) {
+                String columnName = metadata.getColumnLabel(colIndex);
```
chl-wxp: For vector dimension lookup, should we use ResultSetMetaData#getColumnName instead of getColumnLabel?

The output Column name can still use getColumnLabel, but pg_attribute.attname should be matched against the physical source column name. Otherwise a query like `SELECT embedding AS embedding_vector FROM public.items` may fail to recover vector(N), because pg_attribute contains `embedding`, not the alias `embedding_vector`.

seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-3/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/JdbcPostgresPgVectorIT.java
```diff
@@ -0,0 +1,264 @@
...
+        List<List<Object>> vectorInfo =
+                querySql(
+                        "SELECT format_type(atttypid, atttypmod) FROM pg_attribute "
+                                + "WHERE attrelid = 'pgvector_e2e_auto_create_sink_table'::regclass AND attname = 'embedding'");
+        Assertions.assertFalse(
+                vectorInfo.isEmpty(), "embedding column should exist in pg_attribute");
+        String fullType = vectorInfo.get(0).get(0).toString();
+        Assertions.assertTrue(
+                fullType.startsWith("vector(") && fullType.endsWith(")"),
+                "Expected vector(N) type but got: " + fullType);
```
chl-wxp: check vector(N)

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch and traced the full pgvector path again, especially the real branch that mattered in the previous round: `query source -> schema inference -> auto-create sink`.

# What this PR fixes
- User pain: pgvector support is only really complete if the PostgreSQL source side can preserve `vector(N)` during schema inference and carry that exact dimension into sink auto-creation.
- Fix approach: the latest head recovers `vector(N)` from PostgreSQL metadata, threads the live JDBC connection into that mapper path, and verifies the final sink column type in E2E.
- One-line summary: the latest revision closes the schema-contract gap on the real production path, not just the vector value-conversion path.

# Full runtime chain I checked
```text
query source schema inference
  -> JdbcCatalogUtils.getCatalogTable(tableConfig, connection, jdbcDialect) [JdbcCatalogUtils.java:371-420]
      -> when both table_path and query are configured
          -> tableOfQuery = getCatalogTable(connection, query, jdbcDialect)
              -> CatalogUtils.getCatalogTable(connection, resultSetMetaData, typeMapper, sqlQuery) [CatalogUtils.java:328-345,396-420]
                  -> PostgresTypeMapper.mappingColumn(metadata, colIndex, connection, sqlQuery) [PostgresTypeMapper.java:69-105]
                      -> resolve table/schema
                      -> query pg_attribute / pg_namespace
                      -> recover full native type = vector(N)
                      -> PostgresTypeConverter.convert(...) [PostgresTypeConverter.java:247-260]

sink auto-create
  -> FLOAT_VECTOR column with scale = N
  -> PostgresTypeConverter.reconvert(...) [PostgresTypeConverter.java:413-422]
  -> generated sink DDL becomes vector(N)

value path
  -> PostgresJdbcRowConverter.toInternal(...) [PostgresJdbcRowConverter.java:164-177]
  -> PostgresJdbcRowConverter.toExternal(...) [PostgresJdbcRowConverter.java:320-345]
```

# Re-review result
- The blocker from the previous round is fixed for me now.
- `PostgresTypeMapper.java:75-100` now uses `metadata.getColumnName(colIndex)` for the physical column lookup and keeps `metadata.getColumnLabel(colIndex)` only for the output column name.
- `PostgresTypeMapper.java:150-180` also scopes the metadata lookup through `pg_namespace`, so the earlier same-name-table ambiguity is addressed on the latest head.
- `JdbcPostgresPgVectorIT.java:154-198` plus `jdbc_postgres_pgvector_auto_create_sink.conf:22-45` now exercises the real auto-create sink path and checks the final sink type with `format_type(atttypid, atttypmod)`.
- `+1` to @chl-wxp's earlier point about using the physical source column name for `pg_attribute` lookup. I traced the same path and the latest revision now does the right thing there.
- I did not find a new source-level blocker on the current head.

- Fully compatible from my review.
- No API, config, default-value, protocol, or serialization contract break was introduced here.
- The extra PostgreSQL metadata lookup is a one-time schema-inference cost, not a hot data-path cost.

# Test stability
- Rating: **Stable**
- Basis:
  - the new auto-create E2E validates the final sink schema directly instead of relying on weak signals;
  - I did not see a flaky-test pattern such as fixed sleeps, shared mutable globals, or hard-coded local ports in the pgvector additions.

# CI
- The current `Build` failure looks unrelated to this pgvector change from what I traced in the fork run logs.
- The failing jobs are on SQL Server E2E paths (`SqlServerSchemaChangeIT` row-count timeout and `SqlServerCDCIT` CDC deadlock 1205), not on the PostgreSQL pgvector path touched by this PR.

### Conclusion: can merge
1. Blocking items

2. Suggested follow-up
- Non-blocking only: it would still be nice to add docs later for the `FLOAT_VECTOR` / `vector(N)` mapping, but I do not treat that as a merge blocker for this PR.

This looks ready to me now. The latest revision closes the exact schema-inference gap that was blocking the previous round.

DanielLeens:
Hi @CosmosNi, thank you again for the pgvector work here. This PR had already gone through several careful review rounds, and it would be a shame for the useful context and tests to be lost just because the branch became stale.

I updated this original PR branch directly:

- Merged the latest `apache/dev` into `feature-pg-vector`.
- Kept the PR scope focused on the PostgreSQL JDBC pgvector path: schema inference, type mapping, row conversion, and the pgvector E2E coverage.
- Fixed a small compile-level issue in `PostgresTypeMapper`: the new `LOG.debug(...)` path now has the required logger field.
- Ran scoped formatting: `./mvnw -nsu -Dmaven.gitcommitid.skip=true -pl seatunnel-connectors-v2/connector-jdbc,seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-3 spotless:apply`.

The PR now points to head `8dcf4f646440e18fd92580f1225ec833d8557c36`, and GitHub reports it as mergeable again. Current checks have just been queued, so the remaining gate is the normal GitHub CI/review result. Thanks for pushing this feature forward — continuing on the same PR keeps the discussion and review history intact. 🙏

DanielLeens:
Thanks for the update. I rechecked the current head `8dcf4f646440` against the latest `dev`. This branch is currently `diverged` (`ahead_by=12`, `behind_by=5`), and the current `Build` check is `FAILURE`: https://github.com/apache/seatunnel/runs/90634507052

From Daniel's source review side I do not see a new code-side blocker on the unchanged head; this follow-up is about the current CI/upstream-sync gate.

DanielLeens:

DanielLeens (changes requested):
Sorry for missing this in the last review — that's on me. I re-reviewed the latest head (`8dcf4f6`) end to end, including a fresh trace of the full runtime path for the pgvector feature, and I want to flag one item that I should have caught back in my `APPROVED` review on 2026-06-04: it's a carryover from that version, not something newly introduced since then.

Quick note on process: comparing the previously-approved commit directly against the current head is dominated by unrelated `dev`-branch drift that got folded in through the "sync latest `dev`" merge I helped push on this branch on 2026-07-29 (things like the new `NOT a.attisdropped` filter, the `getLocalDateTime` timestamp handling, and the `Types.OTHER` fallback in `PostgresTypeConverter` are all pre-existing in current `dev`, not authored by this PR — `gh pr diff` against the current base confirms none of that shows up as PR-owned content). The only genuine functional change on top of what I approved in June is the `LOG` field/import addition in `PostgresTypeMapper.java`, which I added myself while rescuing the branch to fix a compile break (the file already called `LOG.debug(...)` in the version I approved, but never declared the field — so the code I approved would not actually have compiled). That's now fixed. Everything else below is a fresh full review of the current head.

- **User pain point**: SeaTunnel's PostgreSQL JDBC connector has no support for the `pgvector` extension's `vector` column type. Users building RAG/embedding pipelines that store embeddings in PostgreSQL (`vector(N)`) currently can't read or write that column with the JDBC connector — the column either gets rejected during schema conversion or silently mishandled.
- **Fix approach**: the PR teaches the PostgreSQL dialect to recognize `vector` as a native type, maps it to SeaTunnel's existing generic `FLOAT_VECTOR` (`VectorType.VECTOR_FLOAT_TYPE`), and adds the plumbing needed to read/write `vector(N)` values in both the catalog-based (`table_path`) and query-based (`query = "..."`) schema-discovery paths — including a `pg_attribute` lookup for the case where `ResultSetMetaData` doesn't expose the dimension.
- **One-sentence summary**: this lets SeaTunnel move embedding vectors into and out of PostgreSQL/pgvector columns end to end, including auto-creating the target column with the right dimension.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Files with the core changes:**
- `PostgresTypeConverter.java` — adds `PG_VECTOR` to `convert()` (source→SeaTunnel type) and `reconvert()` (SeaTunnel type→DDL type string).
- `PostgresTypeMapper.java` — new overload `mappingColumn(ResultSetMetaData, int, Connection, String)` that resolves `vector(N)` via `pg_attribute` when `ResultSetMetaData` alone can't supply the dimension.
- `JdbcDialectTypeMapper.java` — new `default` method carrying the `Connection`/SQL-text through, safely no-op for every other dialect.
- `CatalogUtils.java` / `JdbcCatalogUtils.java` — thread the JDBC `Connection` down into the type-mapping call for the query-based schema-discovery path.
- `PostgresCatalog.java` — catalog-based (`table_path`) column-introspection SQL now special-cases `vector` the same way it already does for `varchar`/`numeric` (`atttypmod - 4`).
- `PostgresJdbcRowConverter.java` — `FLOAT_VECTOR` read/write: text `"[0.1,0.2,0.3]"` ⇄ `Float[]` ⇄ `ByteBuffer` (via the pre-existing `VectorUtils`), write side wraps the value in a `PGobject` typed `"vector"`.

**Before/after for the piece that matters most — dimension propagation into `Column`:**

```java
// PostgresTypeConverter.convert(), current head
case PG_VECTOR:
    builder.dataType(VectorType.VECTOR_FLOAT_TYPE);
    builder.sourceType(PG_VECTOR);                 // always the bare literal "vector"
    if (typeDefine.getColumnType() != null
            && typeDefine.getColumnType().startsWith("vector(")
            && typeDefine.getColumnType().endsWith(")")) {
        String dimStr = typeDefine.getColumnType()
                .substring("vector(".length(), typeDefine.getColumnType().length() - 1);
        builder.scale(Integer.parseInt(dimStr));   // dimension only lands in `scale`
    }
    break;
```

Compare with the pattern this file already uses for every other parameterized PostgreSQL type, e.g. `PG_VARCHAR` a few lines above:

```java
case PG_VARCHAR:
case PG_CHARACTER_VARYING:
    if (typeDefine.getLength() == null || typeDefine.getLength() <= 0) {
        builder.sourceType(pgDataType);
    } else {
        builder.sourceType(String.format("%s(%s)", pgDataType, typeDefine.getLength()));
    }
    break;
```

Every existing parameterized type bakes its parameter into `sourceType` (`"varchar(255)"`, `"numeric(10,2)"`, …). `PG_VECTOR` is the only one that doesn't — `sourceType` is always the bare `"vector"` regardless of dimension, even though `typeDefine.getColumnType()` (`"vector(3)"`) is right there and already being parsed on the next line for `scale`.

**Key findings:**
- The normal path does reach this logic: any `vector` column read through either `table_path` (catalog) or `query` (ResultSetMetaData) schema discovery funnels through this exact `convert()` method.
- The actual scenario this PR is built and tested for — PostgreSQL source → PostgreSQL sink, auto-creating the sink table — is precisely the scenario where the `sourceType` bug below bites, because `PostgresCreateTableSqlBuilder` prefers `column.getSourceType()` verbatim whenever source and sink are both PostgreSQL.
- The fix is a precise, additive extension of an existing pattern (same shape as `varchar`/`numeric`/`bpchar`), not a workaround — it's just missing one line of the pattern for `PG_VECTOR`.
- Read/write of the vector *values* themselves (`PostgresJdbcRowConverter`) is solid and reuses the existing `VectorUtils` byte layout used elsewhere (e.g. Milvus), so no new serialization format is introduced.

**In-depth correctness analysis — Issue 1 walked through the real call chain:**

```
Job with source.table_path = "public.pgvector_e2e_auto_create_source_table"
             and sink.generate_sink_sql = true (auto-create sink, same PostgreSQL instance)
  -> JdbcCatalogUtils.getCatalogTable(tableConfig, jdbcDialect, jdbcCatalog, connection)
       -> jdbcCatalog.getTable(tablePath)          // PostgresCatalog, table_path branch
            -> SELECT_COLUMNS_SQL_TEMPLATE          // PostgresCatalog.java:48 (this PR's new "vector" WHEN branch)
                 -> returns full_type = "vector(3)"
            -> PostgresTypeConverter.INSTANCE.convert(typeDefine)
                 -> case PG_VECTOR: builder.sourceType("vector")   // dimension dropped here
                    builder.scale(3)                                // dimension only kept here
       -> CatalogTable for public.pgvector_e2e_auto_create_source_table
            column "embedding": dataType=VECTOR_FLOAT_TYPE, sourceType="vector", scale=3
  -> Sink auto-create-table path
       -> new PostgresCreateTableSqlBuilder(sourceCatalogTable, ...)
            sourceCatalogName = catalogTable.getCatalogName() = "Postgres"   // PostgresCatalogFactory.factoryIdentifier()
       -> buildColumnSql(embeddingColumn)                             // PostgresCreateTableSqlBuilder.java:144-160
            column.getSinkType() == null
            StringUtils.equalsIgnoreCase("Postgres", "Postgres") == true
            StringUtils.isNotBlank(column.getSourceType())  == true ("vector" is non-blank!)
            -> columnType = column.getSourceType() = "vector"        // <-- dimension already lost, scale=3 never consulted
       -> CREATE TABLE ... "embedding" vector ...                     // no dimension emitted
```

`pgvector` happily accepts an *unconstrained* `vector` column (no fixed dimension), so this `CREATE TABLE` does not fail — it just silently creates the wrong column. `format_type(atttypid, atttypmod)` on that column will report literally `"vector"`, not `"vector(3)"`.

This directly contradicts the PR's own `JdbcPostgresPgVectorIT#testPgVectorAutoCreateSink`, which asserts exactly this:

```java
Matcher dimensionMatcher = Pattern.compile("vector\\((\\d+)\\)").matcher(fullType);
Assertions.assertTrue(dimensionMatcher.matches(), "Expected vector(N) type but got: " + fullType);
```

I traced whether `buildColumnType(column)` (the `reconvert()`-based fallback, which *would* correctly reconstruct `"vector(3)"` from `column.getScale()`) is reachable here — it isn't, because the PostgreSQL-to-PostgreSQL branch in `PostgresCreateTableSqlBuilder.buildColumnSql()` (lines ~148-153) short-circuits to `column.getSourceType()` first whenever it's non-blank, and `"vector"` is never blank. `PostgresCreateTableSqlBuilder.java` isn't part of this PR's diff, but it's an existing, unconditional consumer of `Column.sourceType` that this PR's new code path feeds directly.

**Complete runtime path (write side, for completeness):**

```
SeaTunnelRow with FLOAT_VECTOR field (ByteBuffer, 4 bytes/float)
  -> PostgresJdbcRowConverter.toExternal() [case FLOAT_VECTOR]
       -> VectorUtils.toFloatArray(byteBuffer)
       -> build "[f0,f1,...]" text
       -> PGobject{type="vector", value="[...]"}
       -> PreparedStatement.setObject(idx, pgObject)   // relies on stringtype=unspecified for other STRING columns in the same insert, already set in both new .conf files
  -> PostgreSQL server parses text literal against the target column's declared vector type/dimension
```

This part is fine and matches how the pgvector Java community typically binds vector literals without the optional `pgvector-java` companion jar.

## 1.2 Compatibility Impact

**Fully compatible.** This is purely additive:
- `JdbcDialectTypeMapper.mappingColumn(ResultSetMetaData, int, Connection, String)` is a new `default` method that delegates to the existing `mappingColumn(ResultSetMetaData, int)` for every dialect that doesn't override it — no other connector/dialect is affected.
- `CatalogUtils.getCatalogTable(ResultSetMetaData, JdbcDialectTypeMapper, String)` (the pre-existing 3-arg signature) is preserved and simply forwards to the new 4-arg overload with `connection = null`.
- No existing `Option`, default value, or serialization format is renamed, removed, or changed. No prior release ever supported `vector`, so there's no historical-behavior regression to migrate.

- The `pg_attribute` lookup in `PostgresTypeMapper.queryVectorTypeFromPgAttribute()` issues one extra parameterized query per `vector` column encountered during schema discovery (not per row), and properly closes its `PreparedStatement`/`ResultSet` via try-with-resources — no leak, and the cost is bounded by schema-discovery time, not row-processing time.
- `resolveTableInfo()`'s regex fallback (`FROM_TABLE_PATTERN`) only fires when `ResultSetMetaData.getTableName()` is empty; for the common `SELECT col, col2 FROM tbl` shape it's fine, but for queries with joins, subqueries, or CTEs it can match the wrong (or no) table and silently fall back to an undimensioned `vector` (see Issue 3). Not a crash risk, but a real precision gap for anyone using a non-trivial `query`.
- Row-level read/write path allocates a small `Float[]`/`ByteBuffer` per vector value — same cost profile as the existing Milvus/other vector-capable connectors that already use `VectorUtils`, nothing new here.

- `queryVectorTypeFromPgAttribute()` catches `SQLException` and logs at `LOG.debug(...)` with the schema/table/column context and the exception — appropriate level, since failure here is expected-and-handled (falls back to undimensioned vector), not a hard failure.
- No sensitive data (credentials, connection strings) appears in any new log statement.
- `toExternal()`'s `FLOAT_VECTOR` branch throws a clear `JdbcConnectorException` with the offending class name if a non-`ByteBuffer` value shows up — good defensive check, no silent success.

**Formal issues (see Section 4 for the summary table):**

### Issue 1: `PG_VECTOR` conversion drops the dimension from `Column.sourceType`, breaking Postgres→Postgres auto-create-sink DDL for vector columns
- **Location:** `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeConverter.java:249-251`
- **Problem description:** In `convert()`, every other parameterized PostgreSQL type (`varchar`, `bpchar`, `numeric`, …) bakes its size/precision into `sourceType` (e.g. `String.format("%s(%s)", pgDataType, typeDefine.getLength())`). `PG_VECTOR` instead always sets `builder.sourceType(PG_VECTOR)` — the bare literal `"vector"` — and only records the dimension in `builder.scale(...)`. As traced above, `PostgresCreateTableSqlBuilder.buildColumnSql()` prefers `column.getSourceType()` verbatim whenever source and sink share the same PostgreSQL catalog identifier, so for a same-instance PostgreSQL→PostgreSQL auto-create-sink pipeline (exactly what `testPgVectorAutoCreateSink` exercises), the generated DDL is `"embedding" vector` instead of `"embedding" vector(3)`.
- **Potential risk:** Silent data-shape loss — pgvector allows unconstrained `vector` columns, so the `CREATE TABLE` succeeds but the dimension constraint is gone. Any downstream consumer (or pgvector index, which typically requires a fixed dimension) relying on the column being `vector(N)` will be surprised. It also means the E2E assertion this PR itself added (`Expected vector(N) type but got: ...`) should fail in the PostgreSQL→PostgreSQL auto-create-sink scenario. I could not find `JdbcPostgresPgVectorIT` mentioned in the fork CI logs I was able to fetch (4 of the 8 parallel `updated-modules-integration-test-part-N` shards), so I can't confirm from CI evidence alone whether this new test module actually executed on the current head — worth the author/a maintainer double-checking that `connector-jdbc-e2e-part-3` was picked up by the changed-module detector for this run.
- **Best improvement:** Mirror the existing pattern used by `varchar`/`numeric`: `builder.sourceType(dimStr != null ? String.format("%s(%s)", PG_VECTOR, dimStr) : PG_VECTOR)`, set once, right where the dimension is already being parsed from `typeDefine.getColumnType()`.
- **Severity:** High
- **Raised by another reviewer:** No — but this is a carryover I should have caught in my own `APPROVED` review on 2026-06-04, since this exact code was present then and hasn't changed since (only the unrelated `LOG` field fix landed after that).

### Issue 2: New `vector`/`FLOAT_VECTOR` type support isn't reflected in the connector documentation
- **Location:** `docs/en/connectors/source/PostgreSQL.md` (Data Type Mapping table, ~lines 51-72); `docs/zh/connectors/source/PostgreSQL.md` has the same gap. Neither is touched by this PR.
- **Problem description:** The existing "Data Type Mapping" table enumerates every supported PostgreSQL type and its SeaTunnel counterpart (`BYTEA → BYTES`, `VARCHAR → STRING`, etc.), but `vector → FLOAT_VECTOR` is absent even though the connector now maps it.
- **Potential risk:** Users have no documented way to discover that `vector` columns are supported, or what SeaTunnel type/config they map to; this also violates the project's own "documentation is part of the feature, not an afterthought" rule for user-visible changes.
- **Best improvement:** Add a `vector<br/>` row mapping to `FLOAT_VECTOR` in both the English and Chinese PostgreSQL source docs, and ideally a one-line note on the `pg_attribute` fallback behavior for `query`-based sources without table metadata.
- **Severity:** Medium

### Issue 3: No unit test for the new `pg_attribute`/`FROM`-clause resolution logic in `PostgresTypeMapper`
- **Location:** `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeMapper.java:68-186` (`mappingColumn(..., Connection, String)`, `resolveTableInfo`, `queryVectorTypeFromPgAttribute`)
- **Problem description:** `PostgresTypeConverterTest.java` gained good coverage for the pure `convert()`/`reconvert()` mapping logic (including the scale/dimension round-trip, which is exactly why Issue 1 is surprising — the unit tests validate `scale` but nothing exercises `sourceType` end-to-end against `PostgresCreateTableSqlBuilder`). But the new branching logic in `PostgresTypeMapper` — `ResultSetMetaData`-first vs. regex-fallback table resolution, schema-qualified `pg_attribute` lookup, graceful degradation when nothing resolves — has no dedicated unit test (e.g. with a mocked `ResultSetMetaData`/`Connection`) and is only reachable today through the container-based E2E test.
- **Potential risk:** Regressions in the fallback/degradation logic (which is genuinely non-trivial — three branches: driver metadata, regex fallback, and give-up) would only be caught by a full Testcontainers run, making them slow to detect and easy to miss in fast local iteration.
- **Best improvement:** Add a focused unit test using a mocked `ResultSetMetaData` (table name present vs. absent) and a mocked `Connection`/`PreparedStatement` to cover: (a) driver-supplied table name path, (b) `FROM`-clause regex fallback with a simple query, (c) regex miss on a query with a subquery/CTE, (d) `pg_attribute` returning no row.
- **Severity:** Medium

### Issue 4: `FROM`-clause regex fallback is fragile for non-trivial queries
- **Location:** `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeMapper.java` (`FROM_TABLE_PATTERN`, `resolveTableInfo`)
- **Problem description:** When the PostgreSQL driver doesn't populate `ResultSetMetaData.getTableName()` (which does happen for some query shapes), the code falls back to a single regex match of the first `FROM <ident>` in the raw SQL text. For a query with a subquery/CTE/derived table, this can match nothing useful (safe) or, in principle, the wrong table if a subquery's `FROM` happens to parse successfully first.
- **Potential risk:** Silent, hard-to-diagnose dimension loss for `vector` columns read through non-trivial custom `query` sources — the column still comes back as `FLOAT_VECTOR`, just without a dimension, so nothing errors out; a user would only notice if they inspect the auto-created sink schema.
- **Best improvement:** This is a reasonable best-effort fallback and doesn't need to be perfect, but it would help to log at `INFO`/`WARN` (not just swallow) when the regex fails to resolve a table for a `vector` column, so users get a signal instead of silent degradation. Not a blocker given the primary (`ResultSetMetaData`) path covers the common cases and the catalog/`table_path` path (Issue 1 aside) doesn't depend on this at all.
- **Severity:** Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

The change follows the existing style of the surrounding dialect classes closely (same builder patterns, same import grouping, same `@Slf4j`/`Logger` conventions as neighboring classes). All new public/package-visible methods and the new `LOG`/`FROM_TABLE_PATTERN` fields carry explanatory Javadoc or inline comments describing purpose and constraints (e.g. the `mappingColumn(..., Connection, String)` Javadoc explains *why* the connection is needed for pgvector specifically, and `resolveTableInfo`/`queryVectorTypeFromPgAttribute` both document their fallback strategy). I didn't find any new core method or non-trivial field lacking a comment.

**Coverage:** unit coverage for `PostgresTypeConverter.convert()/reconvert()` is solid (dimension present, dimension absent, round-trip). E2E coverage (`JdbcPostgresPgVectorIT`) exercises both the plain source→sink path and the auto-create-sink path, including checking the physical PostgreSQL column type via `pg_attribute` — good instinct, and it's precisely thorough enough to have caught Issue 1 if it actually ran (see Issue 1's note on CI evidence). Missing: a unit test for `PostgresTypeMapper`'s new resolution logic (Issue 3).

**Mandatory test-stability analysis (PR touches both UT and E2E test code):**

*Assertion-risk / flaky-anti-pattern review:*
- `JdbcPostgresPgVectorIT.java` uses `Awaitility` (`given().ignoreExceptions().await()...untilAsserted(this::initializeJdbcTable)`) for container-readiness, not `Thread.sleep` — the correct pattern already used elsewhere in this test suite.
- Testcontainers assigns dynamic ports; no hardcoded ports/hostnames beyond the docker-network alias `"pgvector"`, which is scoped to the test's own network.
- Floating-point comparisons in `assertVectorEquals()` use an explicit delta (`1e-6`), not exact equality — correct practice for float32 round-tripped values.
- Resources are released in `@AfterAll tearDown()` (container stop). One very minor nit: `initializeJdbcTable()` opens a `Statement` via `connection.createStatement()` without its own try-with-resources (only the `Connection` is try-with-resources) — harmless in practice since closing the connection cascades to its statements, but slightly inconsistent with the rest of the method's resource style.
- No shared static/global mutable state between the two `@TestTemplate` methods; they operate on distinct tables.
- New unit tests in `PostgresTypeConverterTest.java` are pure, deterministic, no I/O or timing dependency.

*Stability rating:* **Stable.** No pattern here rises to "risk present" or "high risk" — the one nit noted above is cosmetic, not a source of intermittent failure.

## 2.3 Documentation Updates

User-visible behavior change (new supported PostgreSQL column type) with no `docs/en` or `docs/zh` update — see Issue 2. Both language versions need the same update to stay consistent; right now both are equally silent on `vector`, so there's no en/zh divergence, just a shared gap.

# 3. Architectural Soundness

**Precise fix**, mostly. Reusing the existing generic `VectorType`/`VectorUtils` infrastructure (already used by other vector-capable connectors) rather than inventing PostgreSQL-specific vector plumbing is the right call, and extending `JdbcDialectTypeMapper` with a safely-defaulted new method is a clean way to let one dialect opt into connection-aware type resolution without touching every other dialect. Issue 1 is a one-line gap in an otherwise-consistent pattern, not a design flaw.

## 3.2 Maintainability

The three-tier fallback in `PostgresTypeMapper` (driver metadata → regex on the query text → give up gracefully) is clearly commented and easy to follow. The `pg_attribute` query is schema-qualified and parameterized (no injection risk), addressing @chl-wxp's earlier review comment about schema uniqueness. Boundaries are reasonably explicit: the vector-specific logic lives entirely in the Postgres dialect classes, not leaked into generic JDBC code beyond the one new `default` interface method.

## 3.3 Extensibility

The `JdbcDialectTypeMapper.mappingColumn(ResultSetMetaData, int, Connection, String)` extension point is a genuinely reusable pattern — any other dialect that needs connection-aware metadata resolution (not just PostgreSQL/pgvector) can now opt in the same way, without a breaking interface change. This is a long-term-solution-shaped piece of infrastructure, not a one-off hack.

No historical version ever supported `vector`, so there's nothing to migrate. All new surface area is additive and defaulted; upgrading from any prior SeaTunnel version requires no action from users who don't use pgvector.

# 4. Issue Summary

| 1 | `PG_VECTOR` conversion drops dimension from `Column.sourceType`, breaking Postgres→Postgres auto-create-sink DDL | `PostgresTypeConverter.java:249-251` | High |
| 2 | New `vector`/`FLOAT_VECTOR` mapping not documented in `docs/en`/`docs/zh` | `docs/en/connectors/source/PostgreSQL.md`, `docs/zh/connectors/source/PostgreSQL.md` | Medium |
| 3 | No unit test for `PostgresTypeMapper`'s new pg_attribute/FROM-clause resolution logic | `PostgresTypeMapper.java:68-186` | Medium |
| 4 | `FROM`-clause regex fallback is fragile for non-trivial queries; degrades silently | `PostgresTypeMapper.java` (`FROM_TABLE_PATTERN`) | Low |

# 5. Merge Recommendation

   - **Issue 1 (High):** the `sourceType` dimension loss must be fixed before merge — it defeats the primary scenario this PR is built and tested for (PostgreSQL→PostgreSQL auto-create-sink with a fixed vector dimension), and it directly risks failing the PR's own `testPgVectorAutoCreateSink` assertion once that test actually runs against the real database. This is a one-line, low-risk fix (mirror the existing `varchar`/`numeric` pattern), not a redesign.

   - **Issue 2 (Medium):** documentation gap — doesn't block functional correctness, but should land alongside the fix so users can discover the feature.
   - **Issue 3 (Medium):** missing unit coverage for the resolution fallback logic — the E2E test provides some safety net, but a fast unit test would have caught (or would catch, going forward) regressions much earlier.
   - **Issue 4 (Low):** regex fallback fragility — acceptable as a best-effort degradation for now; a log line on fallback failure would improve debuggability later.

**On the current CI failure:** the current head's `Build` check is red, but I traced both failing jobs and neither is related to this PR's diff: `updated-modules-integration-test-part-8 (11, ubuntu-latest)` fails in `IcebergSinkCDCIT#testMysqlCdcCheckDataE2e` (Iceberg CDC, unrelated module), and `unit-test (8, windows-latest)` fails on a Maven dependency-resolution network error for `connector-milvus`'s `grpc-core` jar (`Connection reset` fetching from Maven Central — transient infra flake, also unrelated). This PR is currently `diverged` from `dev` (`ahead_by=12`, `behind_by=34`), so once the source-side fix above lands, syncing with the latest `dev` and rerunning CI is the lowest-cost way to get a clean signal — I wouldn't expect either of these two failures to be caused by anything in this diff.

**Overall assessment:** the design direction is sound and the PR is close — reusing existing `VectorType`/`VectorUtils` infrastructure and the new dialect extension point are both good calls, and the read/write value-conversion path (`PostgresJdbcRowConverter`) looks correct as-is. The one real gap is narrowly scoped (a single missing dimension-format call in `sourceType`), so I'd expect this to be quick to close out. Thanks again for sticking with this feature through several review rounds — happy to take another look as soon as the fix is in.

SEZ9 (commented):
Thanks for working on this. I reviewed the latest head `8dcf4f646440`.

# What this PR fixes
- User pain: PostgreSQL users with pgvector columns cannot sync `vector(N)` data through the JDBC connector — reads fail on the unknown type and there is no way to write embeddings back, and the vector dimension is lost because ResultSetMetaData does not expose it.
- Fix approach: Map `vector` <-> `FLOAT_VECTOR` in `PostgresTypeConverter` (dimension carried via scale, parsed from `vector(N)` and reconstructed on reconvert), add FLOAT_VECTOR read (parse the `[f1,f2,...]` PGobject string into a `ByteBuffer` via `VectorUtils`) and write (build a `vector`-typed `PGobject`) paths in `PostgresJdbcRowConverter`, extend `PostgresCatalog`'s column query with a `vector` atttypmod branch, and thread an optional `Connection`/SQL query through `CatalogUtils.getCatalogTable` and a new default `JdbcDialectTypeMapper.mappingColumn` overload so `PostgresTypeMapper` can look up the dimension from `pg_attribute` via `format_type(atttypid, atttypmod)`.
- One-line summary: The end-to-end type mapping and row conversion work and are unit-tested for bidirectional mapping, but connector docs lack any pgvector mention, the regex-based FROM-clause fallback in `resolveTableInfo` is fragile for joins/subqueries, dimension parsing uses a bare `Integer.parseInt`, `java.nio.ByteBuffer` is fully qualified instead of imported, and the PR description/checklist are still template placeholders.

# Runtime chain I rechecked
```text
Schema discovery (query-based source):
CatalogUtils.getCatalogTable(Connection, ResultSetMetaData, JdbcDialectTypeMapper, String)  CatalogUtils.java:398-402
  -> CatalogUtils.getCatalogTable(connection, resultSetMetaData, typeMapper, sqlQuery)  CatalogUtils.java:325-343
    -> JdbcDialectTypeMapper.mappingColumn(metadata, index, connection, sqlQuery)  JdbcDialectTypeMapper.java:222-226 (default delegates to 2-arg overload)
      -> PostgresTypeMapper.mappingColumn(metadata, colIndex, connection, sqlQuery)  PostgresTypeMapper.java:76+
        -> PostgresTypeMapper.resolveTableInfo(metadata, colIndex, sqlQuery)  (ResultSetMetaData first, regex FROM-clause fallback)
        -> PostgresTypeMapper.queryVectorTypeFromPgAttribute(connection, schema, table, column)  (format_type(atttypid, atttypmod) -> "vector(N)")
        -> PostgresTypeMapper.mappingColumn(BasicTypeDefine)  PostgresTypeMapper.java:42
          -> PostgresTypeConverter.convert(typeDefine)  PostgresTypeConverter.java:249-262 (PG_VECTOR -> VectorType.VECTOR_FLOAT_TYPE, dimension parsed into scale)

Catalog-based discovery:
PostgresCatalog column query (vector atttypmod branch)  PostgresCatalog.java:51 -> PostgresTypeConverter.convert  PostgresTypeConverter.java:249

Read path:
PostgresJdbcRowConverter.toInternal(rs, tableSchema)  PostgresJdbcRowConverter.java:162-174
  -> JdbcFieldTypeUtils.getString(rs, resultSetIndex) -> strip brackets, split, Float.parseFloat -> VectorUtils.toByteBuffer(floats)

Write path:
PostgresJdbcRowConverter.toExternal(...)  PostgresJdbcRowConverter.java:318-342
  -> VectorUtils.toFloatArray(byteBuffer) -> build "[f1,f2,...]" -> PGobject.setType("vector")/setValue -> statement.setObject

Sink DDL:
PostgresTypeConverter.reconvert(column)  PostgresTypeConverter.java:420-429 (FLOAT_VECTOR -> vector(scale) or bare vector)
```

# Findings
**Issue 1: New user-facing pgvector type support is not documented in the PostgreSQL connector docs (source/sink data type mapping tables, pgvector extension prerequisite).**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeConverter.java:249`
- Why it matters: - This PR adds a new user-facing type mapping (`vector(N)` <-> `FLOAT_VECTOR`) in `PostgresTypeConverter` and auto-create support via `PostgresCatalog`, but the changed-files list contains no updates under `docs/` (neither `docs/en/connector-v2/source/PostgreSQL.md` nor `docs/en/connector-v2/sink/PostgreSQL.md`, and the zh-CN counterparts).
- Users reading the connector docs will not know that pgvector columns are now supported, what SeaTunnel type they map to, or that the `pgvector` extension must be installed on the target PostgreSQL instance for reads/writes and for auto-created `vector(N)` sink tables to succeed.
- Per our contribution guide (issue #4544 referenced in the PR template), connector docs must be updated for any new type mapping. Please add a `vector` row to the Data Type Mapping tables in both source and sink PostgreSQL docs, mention the pgvector extension prerequisite in the Tips section, and ideally include a small example config (the e2e conf files added in this PR are a good starting point).
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeConverter.java:250  builder.dataType(VectorType.VECTOR_FLOAT_TYPE);`; `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalog.java:51  + "        WHEN t.typname = 'vector' THEN t.typname || '(' || (a.atttypmod - 4) || ')'\n"`
- Suggested fix: Since this adds a new user-facing type mapping (`vector(N)` <-> `FLOAT_VECTOR`), please also update `docs/en/connector-v2/source/PostgreSQL.md` and `docs/en/connector-v2/sink/PostgreSQL.md` (and the zh-CN versions) — add a `vector` row to the Data Type Mapping tables and note that the pgvector extension is required. Without this, users have no way to discover the feature from the docs.
- Severity: Medium

**Issue 2: The regex-based FROM-clause parsing in PostgresTypeMapper.resolveTableInfo can resolve the wrong table for subqueries/joins, yielding an incorrect or missing vector dimension.**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeMapper.java:39`
- Why it matters: - `FROM_TABLE_PATTERN = Pattern.compile("(?i)\\bFROM\\s+([\\w.\"`]+)")` matches the first `FROM <identifier>` occurrence. For `SELECT ... FROM (SELECT ... FROM inner_table) t` the outer `FROM (` does not match, so the regex silently latches onto the inner subquery's table; for multi-table joins it always picks the first table, even if the vector column comes from the second one.
- The consequence is that `queryVectorTypeFromPgAttribute` may look up `pg_attribute` for the wrong table and return a wrong dimension (or none), which then propagates into the sink auto-create DDL (`vector(N)` with wrong N) and can fail inserts at runtime on Zeta/Flink/Spark.
- Also note the character class includes a backtick, which is a MySQL quoting convention, not PostgreSQL.
- I'd suggest restricting the regex fallback to simple single-table queries (the ones `Query` mode actually generates), logging a warning and skipping dimension resolution otherwise, and documenting this limitation in the method Javadoc.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeMapper.java:130  Matcher matcher = FROM_TABLE_PATTERN.matcher(sqlQuery);`
- Suggested fix: This regex will match the first `FROM <identifier>` it finds, so for `SELECT ... FROM (SELECT ... FROM inner_table) t` it silently resolves the subquery's table, and for joins it always picks the first table — potentially returning the wrong vector dimension from `pg_attribute` and generating a wrong `vector(N)` in auto-created sink tables. Also, the backtick in the character class is MySQL quoting, not PostgreSQL. Please restrict this fallback to simple single-table queries, log a warning and skip dimension resolution otherwise, and document the limitation in the Javadoc.
- Severity: Medium

**Issue 3: Vector dimension parsing in PostgresTypeConverter.convert uses a bare Integer.parseInt, which can throw an unclassified NumberFormatException instead of a SeaTunnel error.**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeConverter.java:261`
- Why it matters: - In `convert()`, the dimension is extracted by substring + `Integer.parseInt(dimStr)` with no guard; a malformed `columnType` (e.g. `vector( 3)` with whitespace, or an unexpected catalog string) will surface as a raw `NumberFormatException` rather than the classified errors this converter uses elsewhere (`CommonError` is already imported in this file).
- Users hitting this would get an opaque stack trace during schema discovery instead of an actionable connector error, which is inconsistent with the SeaTunnel error-code convention for type converters.
- Please wrap the parsing in a try/catch (or trim + validate with a regex) and throw `CommonError.convertToSeaTunnelTypeError(DatabaseIdentifier.POSTGRESQL, ...)` on failure, and add a negative unit test in `PostgresTypeConverterTest` for a malformed dimension string.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresTypeConverter.java:255  .substring(`
- Suggested fix: A bare `Integer.parseInt(dimStr)` here will throw a raw `NumberFormatException` for any unexpected columnType string (e.g. whitespace inside the parentheses). Since `CommonError` is already imported in this class, please guard the parse and throw `CommonError.convertToSeaTunnelTypeError(...)` on failure so users get a classified, actionable error, and add a negative case to `PostgresTypeConverterTest`.
- Severity: Medium

**Issue 4: PR description and checklist are incomplete: purpose, user-facing change, and testing sections are left as template placeholders and checklist boxes are unchecked.**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/JdbcDialectTypeMapper.java:224  default Column mappingColumn(`
- Why it matters: - The PR body leaves the 'Purpose of this pull request', 'Does this PR introduce any user-facing change?', and 'How was this patch tested?' sections empty and none of the checklist items are ticked, even though this PR clearly introduces a user-facing change (new type support) and adds e2e tests (`JdbcPostgresPgVectorIT` plus two `.conf` resources).
- A complete description helps release managers assemble the changelog and helps future readers understand why the `mappingColumn(ResultSetMetaData, int, Connection, String)` API was extended in `JdbcDialectTypeMapper`.
- Please fill in the sections (yes, user-facing: pgvector `vector(N)` support in the PostgreSQL JDBC connector), tick the docs checklist item once the connector docs are updated, and link the corresponding GitHub issue per the contribution checklist.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/JdbcDialectTypeMapper.java:225  ResultSetMetaData metadata, int colIndex, Connection connection, String sqlQuery)`
- Severity: Low

**Issue 5: Fully-qualified java.nio.ByteBuffer references in PostgresJdbcRowConverter instead of an import.**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresJdbcRowConverter.java:322`
- Why it matters: - The new `FLOAT_VECTOR` write branch uses `java.nio.ByteBuffer` inline twice instead of adding `import java.nio.ByteBuffer;` at the top of the file, which deviates from the import style used throughout the connector-jdbc module.
- This is purely cosmetic but hurts readability and consistency with the rest of the codebase.
- Please add the import and use the short class name in both the `instanceof` check and the cast.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresJdbcRowConverter.java:323  java.nio.ByteBuffer byteBuffer = (java.nio.ByteBuffer) vectorValue;`
- Suggested fix: Nit: please add `import java.nio.ByteBuffer;` and use the short class name here and in the cast on the next line, consistent with the import style elsewhere in this file.
- Severity: Low

# Review conclusion

**2. Remaining follow-ups**
- Issue 1: New user-facing pgvector type support is not documented in the PostgreSQL connector docs (source/sink data type mapping tables, pgvector extension prerequisite).
- Issue 2: The regex-based FROM-clause parsing in PostgresTypeMapper.resolveTableInfo can resolve the wrong table for subqueries/joins, yielding an incorrect or missing vector dimension.
- Issue 3: Vector dimension parsing in PostgresTypeConverter.convert uses a bare Integer.parseInt, which can throw an unclassified NumberFormatException instead of a SeaTunnel error.
- Issue 4: PR description and checklist are incomplete: purpose, user-facing change, and testing sections are left as template placeholders and checklist boxes are unchecked.
- Issue 5: Fully-qualified java.nio.ByteBuffer references in PostgresJdbcRowConverter instead of an import.

DanielLeens:
Thanks @SEZ9 for the independent full pass on the current head. A few notes tying this back to where things stand:

+1 on Issue 1 (docs gap) and Issue 2 (regex FROM-clause fallback fragility for joins/subqueries, plus the MySQL-style backtick in the character class) — these line up with what I already flagged as my own non-blocking Issue 2 and Issue 4 in my 2026-08-06 review, so good to have independent confirmation on both.

Issue 3 (bare `Integer.parseInt` on the dimension string) and Issue 5 (fully-qualified `java.nio.ByteBuffer` instead of an import) are new catches from your pass that I didn't call out — agreed on both, worth folding into the same fix commit. Issue 4 (PR description/checklist placeholders) is a process nit, agreed as well.

One thing I want to make sure doesn't get lost in the combined list: none of Issue 1-5 here cover the actual merge blocker from my 2026-08-06 `CHANGES_REQUESTED` review — `PostgresTypeConverter.convert()` at `PostgresTypeConverter.java:249-251` still sets `builder.sourceType(PG_VECTOR)` as the bare literal `"vector"` instead of baking the dimension in (the pattern every other parameterized type in this file already follows, e.g. `varchar(255)`). Since `PostgresCreateTableSqlBuilder.buildColumnSql()` prefers `column.getSourceType()` verbatim whenever source and sink share the same PostgreSQL catalog identifier, the Postgres→Postgres auto-create-sink path — the exact scenario `JdbcPostgresPgVectorIT#testPgVectorAutoCreateSink` is built to verify — will still emit `vector` instead of `vector(N)`. There's been no new commit on this PR since I raised that, so it remains open and is still the primary blocker for me.

So, from my side: please fold Issues 3 and 5 above into the same pass as my Issue 1 fix (dimension into `sourceType`), plus docs and the regex-fallback hardening we're both flagging, and I'm happy to do another full re-review once that lands.

SEZ9:
Thanks @DanielLeens for reviving the branch and keeping the review history intact — much appreciated.

To keep the remaining work in one place, here's the consolidated list on head `8dcf4f646440`:

**Merge blocker (from my earlier CHANGES_REQUESTED review, still unresolved):**
- `PostgresTypeConverter.convert()` (`PostgresTypeConverter.java:249-251`) sets `builder.sourceType(PG_VECTOR)` as the bare literal `"vector"`. Since `PostgresCreateTableSqlBuilder.buildColumnSql()` prefers `column.getSourceType()` verbatim, auto-created sink tables lose the dimension. Please bake it in as `vector(N)` using the scale, matching the pattern the other parameterized types in this file already follow (e.g. `varchar(255)`).

**Non-blocking, please fold into the same fix commit:**
- Docs: add a pgvector mention to the Postgres JDBC connector docs (type mapping table + a short example).
- Harden the dimension parsing: guard the bare `Integer.parseInt` on the `vector(N)` string.
- Replace the fully-qualified `java.nio.ByteBuffer` with a proper import.
- The regex FROM-clause fallback in `resolveTableInfo` is fragile for joins/subqueries and includes a MySQL-style backtick in the character class — at minimum drop the backtick and add a comment noting the limitation, or fail gracefully to the no-dimension path.
- Fill in the PR description and checklist — they're still template placeholders.

**CI gate:**
- The branch is currently diverged from `dev` (ahead 12 / behind 5) and the latest `Build` check failed: https://github.com/apache/seatunnel/runs/90634507052. Since `dev` has a newer green run, please sync with the latest `dev` first and rerun. If it still fails after the sync, share the new failing check link and I'll help narrow it down.

The core mapping/conversion work looks solid and the bidirectional unit tests are in good shape, so once the sourceType fix lands and CI is green on a synced head, this should be close. Thanks again for pushing it forward! 🙏

DanielLeens:
Thanks @SEZ9, this consolidated list matches where I landed too, so let's use it as the single source of truth for the fix commit:

- **Merge blocker (unresolved):** `PostgresTypeConverter.convert()` at `PostgresTypeConverter.java:249-251` still sets `builder.sourceType(PG_VECTOR)` to the bare literal `"vector"` instead of `vector(N)`. Since `PostgresCreateTableSqlBuilder.buildColumnSql()` uses `column.getSourceType()` verbatim when source and sink share the same catalog identifier, the Postgres→Postgres auto-create-sink path (`JdbcPostgresPgVectorIT#testPgVectorAutoCreateSink`) will keep emitting `vector` without the dimension until this is fixed.
- **Non-blocking, fold into the same commit:** docs for pgvector type mapping, guarding the bare `Integer.parseInt` on the dimension string, replacing the fully-qualified `java.nio.ByteBuffer` with a proper import, hardening/commenting the regex FROM-clause fallback (and dropping the stray backtick from the character class), and filling in the PR description/checklist placeholders.

One small correction on the CI gate note: I just re-checked the divergence against `dev` on head `8dcf4f646440` and it's `ahead_by=12, behind_by=42` (not `behind_by=5`). Doesn't change the recommendation — sync with the latest `dev` first, then rerun `Build` (https://github.com/apache/seatunnel/runs/90634507052 is indeed currently failing) — just flagging the number so nobody chases a stale diff count.

@CosmosNi, once the sourceType-dimension fix plus the items above land in one commit on a `dev`-synced head, ping us both and I'll do a full re-review.

SEZ9:
Thanks @DanielLeens — agreed, let's treat your consolidated list as the single source of truth for the fix commit.

- **Merge blocker (unresolved):** `PostgresTypeConverter.convert()` still sets `builder.sourceType(PG_VECTOR)` to the bare literal instead of baking the dimension in as `vector(N)`. Since `PostgresCreateTableSqlBuilder.buildColumnSql()` uses `column.getSourceType()` verbatim when source and sink share the same catalog identifier, the Postgres→Postgres auto-create-sink path (`JdbcPostgresPgVectorIT#testPgVectorAutoCreateSink`) keeps losing the dimension until this lands.
- **Non-blocking, same commit:** pgvector docs/type mapping, guarding the bare `Integer.parseInt` on the dimension string, importing `ByteBuffer` instead of using the fully-qualified name, hardening/commenting the regex FROM-clause fallback (and dropping the stray backtick from the character class), and filling in the PR description/checklist placeholders.
- **CI note:** thanks for the correction — head `8dcf4f646440` is `ahead_by=12, behind_by=42` against `dev`, not `behind_by=5`. Recommendation unchanged: sync with the latest `dev` first, then rerun `Build` (https://github.com/apache/seatunnel/runs/90634507052 is currently failing).

Once the sourceType-dimension fix plus the items above land in one commit on a `dev`-synced head, ping us and I'll do a full re-review.

DanielLeens:
Thanks @SEZ9, confirmed we're aligned on the consolidated list as the single source of truth. I just re-checked live and the head is still unchanged at `8dcf4f646440e18fd92580f1225ec833d8557c36`, so the `sourceType`-dimension fix in `PostgresTypeConverter.convert()` is still the open blocker — no new commit has landed yet.

@CosmosNi, to recap what's needed in one pass:
1. Bake the dimension into `sourceType` (`vector(N)`) in `PostgresTypeConverter.convert()` around `PostgresTypeConverter.java:249-251`, matching the `varchar(255)`-style pattern already used for other parameterized types in that file.
2. Fold in the non-blocking items from the consolidated list (docs, `Integer.parseInt` guarding, the `ByteBuffer` import, hardening the regex FROM-clause fallback, and filling in the PR description/checklist).
3. Sync with the latest `dev` (last checked as `ahead_by=12, behind_by=42`) and get `Build` green.

Once that lands on a synced head, ping us and I'll do a full re-review of the complete diff.

SEZ9:
Thanks @DanielLeens — confirmed, we're fully aligned. I've verified the same: the head is still `8dcf4f646440e18fd92580f1225ec833d8557c36` with no new commit since our consolidated list, so nothing on that list is done yet.

To answer your recap directly, my status matches yours point by point:

1. **Merge blocker — still open.** `PostgresTypeConverter.convert()` at `PostgresTypeConverter.java:249-251` still sets `sourceType` to the bare literal `vector` instead of `vector(N)`. Until the dimension is baked in following the `varchar(255)`-style pattern, `PostgresCreateTableSqlBuilder.buildColumnSql()` will keep emitting the dimensionless type on the Postgres→Postgres auto-create-sink path that `JdbcPostgresPgVectorIT#testPgVectorAutoCreateSink` covers.
2. **Non-blocking items — none addressed yet.** Docs for the pgvector type mapping, guarding the bare `Integer.parseInt`, replacing the fully-qualified `java.nio.ByteBuffer` with an import, hardening the regex FROM-clause fallback (including dropping the stray backtick), and filling in the PR description/checklist are all still pending, to be folded into the same fix commit.
3. **CI/sync — unchanged.** The divergence you last checked (`ahead_by=12, behind_by=42` on `8dcf4f646440`) still needs a sync with the latest `dev`, and `Build` (https://github.com/apache/seatunnel/runs/90634507052) still needs to go green after that.

So the concrete ask to the author remains exactly your three-step recap: land the `sourceType`-dimension fix plus the non-blocking items in one commit on a `dev`-synced head, get `Build` passing, then ping us. I'll join the full re-review of the complete diff once that lands.

DanielLeens:
Thanks @SEZ9, confirmed on my side too. Head is still `8dcf4f646440e18fd92580f1225ec833d8557c36`, no new commit, so there's nothing to re-review yet. The `sourceType`-dimension fix in `PostgresTypeConverter.convert()` (around `PostgresTypeConverter.java:249-251`) remains the single open blocker, with the non-blocking docs/parsing/import/regex/checklist items to fold into the same commit, and the `dev` sync + green `Build` still outstanding on top of that.

@CosmosNi, whenever that lands in one push on a synced head, ping us both and we'll do the full re-review together.

PR #10990, opened 2026-06-01
[Feature][API] Define Knowledge Sync metadata field contract

### LINKED ISSUE #10915: PR-B: [Feature][API] Define Knowledge Sync metadata field contract
### Description

This issue tracks the Knowledge Sync metadata contract work under #10889.

Knowledge Sync needs a shared set of metadata keys so document sources, transforms, and lifecycle sinks can use consistent document and chunk identity fields.

The first contract should define logical metadata keys such as:

- `DocumentId`
- `DocumentHash`
- `SourceUri`
- `SourceVersion`
- `SourceModifiedAt`
- `MimeType`
- `Deleted`
- `ChunkId`
- `ChunkHash`
- `ChunkIndex`

This issue only covers the metadata contract and Metadata transform projection support. It does not implement source-side document routing or lifecycle sink behavior.

### Scope

- Define standard Knowledge Sync metadata fields in API.
- Register these fields as valid metadata keys.
- Allow `Metadata` transform to project declared Knowledge Sync metadata into physical columns.
- Add unit tests and documentation.

### Out of scope

- Markdown source routing.
- Sink-side partition strategy.
- Qdrant or Milvus lifecycle sink behavior.
- Document parser or chunker implementation.

yzeng1618:

Closes #10915.
Part of #10889.
Related to #10914 and #10964.

This PR defines the Knowledge Sync metadata field contract and makes these metadata keys recognizable by the existing `Metadata` transform.

The goal is to provide a shared contract for future Knowledge Sync sources, transforms, and lifecycle sinks. The PR adds standard logical metadata keys such as `DocumentId`, `DocumentHash`, `SourceUri`, `Deleted`, `ChunkId`, `ChunkHash`, and `ChunkIndex`.

This PR is contract-only. It does not implement source-side document routing, sink-side partition strategy, lifecycle sink behavior, or document parser/chunker behavior.

Yes.

This PR adds documented Knowledge Sync metadata keys and allows users to explicitly project them as physical fields with the `Metadata` transform.

For example:

```hocon
metadata_fields = {
  DocumentId = "document_id"
  ChunkHash = "chunk_hash"
}
```

This transform does not generate Knowledge Sync metadata by itself.

Existing Markdown RAG output currently exposes physical fields such as:

- `source_uri`
- `document_id`
- `chunk_id`
- `chunk_index`
- `content_hash`

This PR does not change or rename those Markdown fields. Built-in Markdown producer alignment is left to follow-up work.

Added unit tests for:

- Knowledge Sync metadata field registration
- logical metadata key to physical field mapping
- `Metadata` transform projection when metadata schema and row options are present
- rejection of unknown metadata keys
- rejection when required metadata schema entries are missing
- rejection when projection target fields already exist
- metadata schema preservation through schema-preserving transforms

### Check list

* [ *] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.

### DIFF
--- a/docs/en/transforms/metadata.md
+++ b/docs/en/transforms/metadata.md
@@ -33,12 +33,31 @@ The Metadata transform plugin is used to extract metadata information from data 
 | Gtid       | string | Global Transaction ID (`server_uuid:transaction_id`). `null` when GTID is disabled or for snapshot rows. | MySQL-CDC only |
 | Partition |  string  |  Partition information of the data, multiple partition fields separated by commas  | Connectors supporting partitions |
 
+## Knowledge Sync Metadata Fields
+
+Knowledge Sync pipelines can use the following metadata keys to carry document and chunk identity before lifecycle sinks or sink partition routing. These keys are logical row metadata keys. They become real physical columns only after they are projected by the `Metadata` transform.
+
+| Metadata Key | Canonical Physical Field | Output Type | Description |
+|:---:|:---:|:---:|:---|
+| DocumentId | `document_id` | string | Stable document identity. |
+| DocumentHash | `document_hash` | string | Stable document version or content hash. |
+| SourceUri | `source_uri` | string | Original source URI or path. |
+| SourceVersion | `source_version` | string | Source-side version, etag, revision, or similar marker. |
+| SourceModifiedAt | `source_modified_at` | long | Source modified time in epoch milliseconds. |
+| MimeType | `mime_type` | string | Source MIME type. |
+| Deleted | `deleted` | boolean | Whether the document is a deletion marker. |
+| ChunkId | `chunk_id` | string | Stable chunk identity. |
+| ChunkHash | `chunk_hash` | string | Stable chunk content hash. |
+| ChunkIndex | `chunk_index` | int | Zero-based chunk index in the document. |
+
 ### Important Notes
 
 1. **Metadata field names are case-sensitive**: Configuration must strictly follow the Key names in the table above (e.g., `Database`, `Table`, `RowKind`, etc.)
 2. **Time fields**: `Delay` and `SourceTimestamp` are only available for CDC connectors. `EventTime` is also provided by the Kafka source via `ConsumerRecord.timestamp` when available.
 3. **Kafka event time**: The Kafka source writes `ConsumerRecord.timestamp` (milliseconds) into `EventTime` when it is non-negative, so you can surface it with the `Metadata` transform.
 4. **Binlog/GTID fields**: `BinlogFile`, `BinlogPos`, `BinlogRow`, and `Gtid` are MySQL-CDC specific. For `startup.mode = initial`, snapshot rows return `null` for all four fields.
+5. **Knowledge Sync projection is explicit**: Knowledge Sync metadata fields are projected only when they are configured in `metadata_fields` and declared in the input table metadata schema. If the target physical field already exists, the transform fails instead of overwriting it.
+6. **Markdown RAG compatibility**: This change does not rename existing Markdown RAG physical fields such as `source_uri`, `document_id`, `chunk_id`, `chunk_index`, and `content_hash`, and it does not change the current `chunk_id` or `content_hash` generation formulas.
 
 ## Options
 
@@ -75,6 +94,28 @@ metadata_fields {
 - The left side must be a supported metadata Key (see table above), and is strictly case-sensitive
 - The right side is a custom output field name, which cannot duplicate existing field names
 - You can select only the metadata fields you need, not all of them must be configured
+
+### Knowledge Sync Projection Example
+
+Project Knowledge Sync logical metadata keys into canonical physical columns before a sink validates routing fields or lifecycle keys.
+
+```hocon
+transform {
+  Metadata {
+    plugin_input = "knowledge_chunks"
+    plugin_output = "knowledge_chunks_with_meta"
+    metadata_fields = {
+      DocumentId = "document_id"
+      DocumentHash = "document_hash"
+      ChunkId = "chunk_id"
+      ChunkHash = "chunk_hash"
+      ChunkIndex = "chunk_index"
+    }
+  }
+}
+```
+
+After this transform, downstream sinks can validate `document_id`, `chunk_id`, and `chunk_hash` as regular physical fields in the input schema.
 
 ## Complete Examples
 
--- a/docs/en/transforms/metadata.md
+++ b/docs/en/transforms/metadata.md
@@ -35,7 +35,9 @@ The Metadata transform plugin is used to extract metadata information from data 
 
 ## Knowledge Sync Metadata Fields
 
-Knowledge Sync pipelines can use the following metadata keys to carry document and chunk identity before lifecycle sinks or sink partition routing. These keys are logical row metadata keys. They become real physical columns only after they are projected by the `Metadata` transform.
+Knowledge Sync pipelines can use the following logical metadata keys to carry document and chunk identity. These keys become physical fields only after they are explicitly projected by the `Metadata` transform.
+
+The `Metadata` transform does not generate Knowledge Sync metadata by itself. The upstream source or transform must declare these fields in `CatalogTable.metadataSchema` and write the corresponding values into `SeaTunnelRow.options`.
 
 | Metadata Key | Canonical Physical Field | Output Type | Description |
 |:---:|:---:|:---:|:---|
@@ -56,8 +58,8 @@ Knowledge Sync pipelines can use the following metadata keys to carry document a
 2. **Time fields**: `Delay` and `SourceTimestamp` are only available for CDC connectors. `EventTime` is also provided by the Kafka source via `ConsumerRecord.timestamp` when available.
 3. **Kafka event time**: The Kafka source writes `ConsumerRecord.timestamp` (milliseconds) into `EventTime` when it is non-negative, so you can surface it with the `Metadata` transform.
 4. **Binlog/GTID fields**: `BinlogFile`, `BinlogPos`, `BinlogRow`, and `Gtid` are MySQL-CDC specific. For `startup.mode = initial`, snapshot rows return `null` for all four fields.
-5. **Knowledge Sync projection is explicit**: Knowledge Sync metadata fields are projected only when they are configured in `metadata_fields` and declared in the input table metadata schema. If the target physical field already exists, the transform fails instead of overwriting it.
-6. **Markdown RAG compatibility**: This change does not rename existing Markdown RAG physical fields such as `source_uri`, `document_id`, `chunk_id`, `chunk_index`, and `content_hash`, and it does not change the current `chunk_id` or `content_hash` generation formulas.
+5. **Knowledge Sync projection is explicit**: Knowledge Sync metadata fields are projected only when they are configured in `metadata_fields`, declared in the input table metadata schema, and present in row options. This transform reads logical row metadata; it does not read existing physical columns with the same names.
+6. **Markdown RAG compatibility**: Existing Markdown RAG output currently exposes physical fields such as `source_uri`, `document_id`, `chunk_id`, `chunk_index`, and `content_hash`. This transform does not migrate those physical fields into logical Knowledge Sync metadata, and this change does not rename them.
 
 ## Options
 
@@ -97,7 +99,7 @@ metadata_fields {
 
 ### Knowledge Sync Projection Example
 
-Project Knowledge Sync logical metadata keys into canonical physical columns before a sink validates routing fields or lifecycle keys.
+Project Knowledge Sync logical metadata keys into canonical physical columns. The upstream producer must already provide the metadata values through row options and declare them in the table metadata schema.
 
 ```hocon
 transform {
@@ -115,7 +117,7 @@ transform {
 }
 ```
 
-After this transform, downstream sinks can validate `document_id`, `chunk_id`, and `chunk_hash` as regular physical fields in the input schema.
+After this transform, downstream components can read `document_id`, `chunk_id`, and `chunk_hash` as regular physical fields in the input schema.
 
 ## Complete Examples
 
--- a/docs/en/transforms/metadata.md
+++ b/docs/en/transforms/metadata.md
@@ -41,16 +41,16 @@ The `Metadata` transform does not generate Knowledge Sync metadata by itself. Th
 
 | Metadata Key | Canonical Physical Field | Output Type | Description |
 |:---:|:---:|:---:|:---|
-| DocumentId | `document_id` | string | Stable document identity. |
+| DocumentId | `document_id` | string | Non-null stable document identity for every document lifecycle event. |
 | DocumentHash | `document_hash` | string | Stable document version or content hash. |
-| SourceUri | `source_uri` | string | Original source URI or path. |
+| SourceUri | `source_uri` | string | Credential-free stable source URI or path. |
 | SourceVersion | `source_version` | string | Source-side version, etag, revision, or similar marker. |
 | SourceModifiedAt | `source_modified_at` | long | Source modified time in epoch milliseconds. |
 | MimeType | `mime_type` | string | Source MIME type. |
-| Deleted | `deleted` | boolean | Whether the document is a deletion marker. |
-| ChunkId | `chunk_id` | string | Stable chunk identity. |
-| ChunkHash | `chunk_hash` | string | Stable chunk content hash. |
-| ChunkIndex | `chunk_index` | int | Zero-based chunk index in the document. |
+| Deleted | `deleted` | boolean | Non-null lifecycle marker: `false` for normal rows and `true` for document tombstones. |
+| ChunkId | `chunk_id` | string | Stable chunk identity. Required for normal chunk rows and nullable for document tombstones. |
+| ChunkHash | `chunk_hash` | string | Stable chunk content hash. Required for normal chunk rows and nullable for document tombstones. |
+| ChunkIndex | `chunk_index` | int | Zero-based chunk index. Required for normal chunk rows and nullable for document tombstones. |
 
 ### Important Notes
 
@@ -60,6 +60,8 @@ The `Metadata` transform does not generate Knowledge Sync metadata by itself. Th
 4. **Binlog/GTID fields**: `BinlogFile`, `BinlogPos`, `BinlogRow`, and `Gtid` are MySQL-CDC specific. For `startup.mode = initial`, snapshot rows return `null` for all four fields.
 5. **Knowledge Sync projection is explicit**: Knowledge Sync metadata fields are projected only when they are configured in `metadata_fields`, declared in the input table metadata schema, and present in row options. This transform reads logical row metadata; it does not read existing physical columns with the same names.
 6. **Markdown RAG compatibility**: Existing Markdown RAG output currently exposes physical fields such as `source_uri`, `document_id`, `chunk_id`, `chunk_index`, and `content_hash`. This transform does not migrate those physical fields into logical Knowledge Sync metadata, and this change does not rename them.
+7. **Source URI security**: Producers must remove URI user info, access tokens, signatures, and other transient authentication material before writing `SourceUri` into row options. Non-sensitive query parameters that are part of the stable resource identity may be retained.
+8. **Knowledge Sync nullability**: `DocumentId` identifies every document lifecycle event. When `Deleted` is declared, producers must write `false` for normal rows and `true` for document tombstones rather than `null`. Normal chunk rows require `ChunkId`, `ChunkHash`, and `ChunkIndex`; compact document tombstones may leave those chunk fields `null`.
 
 ## Options
 
--- a/docs/zh/transforms/metadata.md
+++ b/docs/zh/transforms/metadata.md
@@ -41,16 +41,16 @@ Knowledge Sync 流程可以使用下面的逻辑元数据 Key 携带文档和 ch
 
 | 元数据 Key | 标准物理字段名 | 输出类型 | 说明 |
 |:---:|:---:|:---:|:---|
-| DocumentId | `document_id` | string | 稳定的文档标识。 |
+| DocumentId | `document_id` | string | 每个文档生命周期事件都必须包含的非空稳定文档标识。 |
 | DocumentHash | `document_hash` | string | 稳定的文档版本或内容哈希。 |
-| SourceUri | `source_uri` | string | 原始来源 URI 或路径。 |
+| SourceUri | `source_uri` | string | 不包含凭证的稳定来源 URI 或路径。 |
 | SourceVersion | `source_version` | string | 来源侧版本、etag、revision 等版本标记。 |
 | SourceModifiedAt | `source_modified_at` | long | 来源修改时间，epoch 毫秒。 |
 | MimeType | `mime_type` | string | 来源 MIME 类型。 |
-| Deleted | `deleted` | boolean | 是否为文档删除标记。 |
-| ChunkId | `chunk_id` | string | 稳定的 chunk 标识。 |
-| ChunkHash | `chunk_hash` | string | 稳定的 chunk 内容哈希。 |
-| ChunkIndex | `chunk_index` | int | 文档内从 0 开始的 chunk 序号。 |
+| Deleted | `deleted` | boolean | 非空生命周期标记：普通行为 `false`，文档 tombstone 为 `true`。 |
+| ChunkId | `chunk_id` | string | 稳定的 chunk 标识。普通 chunk 行必填，文档 tombstone 可为 `null`。 |
+| ChunkHash | `chunk_hash` | string | 稳定的 chunk 内容哈希。普通 chunk 行必填，文档 tombstone 可为 `null`。 |
+| ChunkIndex | `chunk_index` | int | 文档内从 0 开始的 chunk 序号。普通 chunk 行必填，文档 tombstone 可为 `null`。 |
 
 ### 重要说明
 
@@ -60,6 +60,8 @@ Knowledge Sync 流程可以使用下面的逻辑元数据 Key 携带文档和 ch
 4. **Binlog/GTID 字段**：`BinlogFile`、`BinlogPos`、`BinlogRow`、`Gtid` 仅适用于 MySQL-CDC。使用 `startup.mode = initial` 时，快照行的这四个字段均为 `null`。
 5. **Knowledge Sync 投影需要显式配置**：Knowledge Sync 元数据字段只有在 `metadata_fields` 中配置、输入表 metadata schema 中声明了对应 Key，并且行 options 中存在对应值时才会被投影。该转换读取的是逻辑行元数据，不会读取同名的已有物理列。
 6. **兼容 Markdown RAG 物理字段**：现有 Markdown RAG 输出当前以物理字段暴露 `source_uri`、`document_id`、`chunk_id`、`chunk_index` 和 `content_hash`。该转换不会把这些物理字段迁移为逻辑 Knowledge Sync 元数据，本次变更也不会重命名这些字段。
+7. **来源 URI 安全**：producer 在将 `SourceUri` 写入行 options 前，必须移除 URI userinfo、访问令牌、签名以及其他临时鉴权信息。属于稳定资源标识且不敏感的查询参数可以保留。
+8. **Knowledge Sync 可空语义**：`DocumentId` 用于标识每个文档生命周期事件。声明 `Deleted` 后，producer 必须为普通行写入 `false`，为文档 tombstone 写入 `true`，不能使用 `null`。普通 chunk 行必须包含 `ChunkId`、`ChunkHash` 和 `ChunkIndex`；紧凑的文档 tombstone 可以将这些 chunk 字段留为 `null`。
 
 ## 配置选项
 
--- a/docs/zh/transforms/metadata.md
+++ b/docs/zh/transforms/metadata.md
@@ -35,7 +35,9 @@ Metadata 转换插件用于将数据行中的元数据信息提取为普通字�
 
 ## Knowledge Sync 元数据字段
 
-Knowledge Sync 流程可以使用下面的元数据 Key 在 lifecycle sink 或 sink 分区路由之前携带文档和 chunk 身份信息。这些 Key 是逻辑行元数据，只有通过 `Metadata` 转换显式投影后，才会成为真实的物理列。
+Knowledge Sync 流程可以使用下面的逻辑元数据 Key 携带文档和 chunk 身份信息。这些 Key 只有通过 `Metadata` 转换显式投影后，才会成为真实的物理列。
+
+`Metadata` 转换本身不会生成 Knowledge Sync 元数据。上游 Source 或 Transform 必须先在 `CatalogTable.metadataSchema` 中声明这些元数据字段，并将对应值写入 `SeaTunnelRow.options`。
 
 | 元数据 Key | 标准物理字段名 | 输出类型 | 说明 |
 |:---:|:---:|:---:|:---|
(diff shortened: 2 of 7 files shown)


DanielLeens (commented):
Thanks for working on this. I reviewed the full current head locally and traced the real `Metadata` transform path instead of only looking at the new enum/tests.

# What this PR solves
- User pain: Knowledge Sync metadata keys do not have one stable logical contract today, and the existing `Metadata` transform cannot explicitly project those keys by name.
- Fix approach: add `KnowledgeSyncMetadataField`, register those names in `MetadataUtil`, and let `MetadataTransform` accept/project them when they exist in the metadata schema.
- One-line summary: the contract direction is good, but the current head still does not connect that contract to a real producer path that users can actually hit.

# Runtime path I checked
```text
Metadata transform config
  -> MetadataTransform.initOutputFields()
      -> MetadataUtil.isMetadataField("DocumentId") now returns true

Projection path
  -> MetadataTransform.getOutputColumns()
      -> requires metadataSchema.contains("DocumentId")
  -> MetadataTransform.getMetadataFieldValue()
      -> reads inputRow.getOptions().get("DocumentId")

Closest real Knowledge Sync path in this repo
  -> MarkdownReadStrategy.read()
      -> still appends physical fields:
         source_uri / document_id / chunk_id / chunk_index / content_hash
      -> does not produce MetadataColumn entries for DocumentId / ChunkHash
```

# Findings
1. **Blocker: the new Knowledge Sync metadata contract is still not wired into a real producer path.**
   - `MetadataTransform` only works when the upstream `CatalogTable` already contains matching metadata columns and the runtime row options already contain `DocumentId` / `ChunkHash` style keys (`MetadataTransform.java:100-120,126-137`).
   - But the closest real path in the repo today, `MarkdownReadStrategy`, still emits plain physical columns like `document_id`, `chunk_id`, and `content_hash` (`MarkdownReadStrategy.java:89-100,118-145`). It does not produce this new metadata-schema contract.
   - So the documented example can be accepted syntactically, but it is not backed by a real built-in producer path on the latest head.

2. **Related test gap: the new tests are synthetic transform-level tests, not a real producer-to-transform validation path.**
   - `MetadataTransformTest.java:200-328` manually constructs a metadata schema and manually fills `SeaTunnelRow.options`.
   - That proves the transform works *if* an upstream producer already follows the contract, but it does not prove that any real producer in the repo currently does.

# Test / stability notes
- The new UTs are structurally stable. I did not see a flaky-test pattern in the added test code itself.
- The main issue here is coverage realism, not timing instability.

1. Blocking items
- Finding 1: wire at least one real producer path to this contract, or narrow the PR/docs so they only claim an API contract definition rather than an already-usable user feature.

- Finding 2: add one producer-level validation so CI proves this works outside a hand-built test fixture.

Overall, I like the contract direction. The remaining gap is the last mile: on the latest head, users still do not have a real built-in runtime path that actually produces these metadata keys in the form that `Metadata` transform now expects.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from `MetadataTransform` config parsing down into the row-options read path, and I also rechecked the closest built-in producer path in this repo to make sure the contract claim now matches the implementation.

What this PR fixes
- User pain: Knowledge Sync metadata still lacks one stable logical field contract, and `Metadata` transform cannot explicitly project those logical keys by name.
- Fix approach: define `KnowledgeSyncMetadataField`, register those names in `MetadataUtil`, and let `MetadataTransform` project them when the upstream metadata schema declares the keys and the row options carry the values.
- One-line summary: the current head correctly scopes this as a metadata contract definition, not as an already-wired producer feature.

Runtime chain I checked
```text
Metadata transform init
  -> MetadataTransform.initOutputFields(...) [MetadataTransform.java:55-72]
      -> MetadataUtil.isMetadataField("DocumentId") [MetadataUtil.java:122-124]
      -> KnowledgeSyncMetadataField values are part of METADATA_FIELDS [39-49]

Output column generation
  -> MetadataTransform.getOutputColumns() [92-123]
      -> requires metadataSchema.contains("DocumentId")

Runtime value projection
  -> MetadataTransform.getMetadataFieldValue() [126-137]
      -> reads SeaTunnelRow.options using the logical key, e.g. "DocumentId"

Closest existing producer path in this repo
  -> MarkdownReadStrategy.read(...) [MarkdownReadStrategy.java:118-145]
      -> still emits physical fields such as source_uri / document_id / chunk_id / chunk_index / content_hash
      -> does not automatically emit the new logical metadata keys
```

Key findings
1. The earlier Daniel blocker on this PR was that the code could be read as an already-usable producer feature even though the repo still had no built-in producer path emitting these logical keys. On the current head, that mismatch is gone.
2. The updated EN/ZH docs now explicitly say this transform does not generate Knowledge Sync metadata by itself and does not auto-migrate existing Markdown RAG physical fields into logical metadata. That is exactly what the code does today.
3. I do not see a new blocker on the current head.

Test / stability note
- The new tests are transform-level and synthetic, but they are structurally stable and free of obvious flaky-test patterns.
- The remaining blind spot is integration realism, not test stability.

Local verification note
- I checked the full diff, the new API contract enum, the `MetadataTransform` read path, the updated EN/ZH docs, and the current `MarkdownReadStrategy` producer behavior in the repo.
- At the time of review, GitHub `Build` was still in progress.

### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- When a real producer starts emitting this contract, I would add one producer-to-transform integration test so CI proves the contract outside a hand-built fixture as well.

Overall, I like the current shape. The important thing is that the latest head now tells the truth about what the feature is: a stable contract definition, not a magically auto-generated metadata path.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch after the new commit, and I also compared it with the last Daniel-reviewed head (`85817dc323ac9d043124769266e4288e6c930de5`). For the seven PR-owned files, there is no new delta beyond the original feature itself; the current head is effectively this PR plus a merge from `apache/dev`.

What this PR solves:
- User pain: SeaTunnel did not have one stable logical metadata contract for Knowledge Sync fields, and `Metadata` transform could not project keys such as `DocumentId` or `ChunkHash` by name.
- Approach: define `KnowledgeSyncMetadataField`, register those keys in `MetadataUtil`, allow `MetadataTransform` to project them, and update the EN/ZH docs so they accurately describe the real runtime boundary.
- In one line: the latest head turns this into a truthful metadata-contract PR rather than over-claiming that a built-in producer path already emits these keys automatically.

Example:
- Before this PR, a config like `metadata_fields { DocumentId = document_id }` was rejected immediately because `DocumentId` was not a recognized metadata key.
- On the current head, that config is accepted, but it only works when the upstream table actually declares `DocumentId` in `CatalogTable.metadataSchema` and writes the value into `SeaTunnelRow.options`.

Runtime path I checked:
```text
Knowledge Sync contract
  -> KnowledgeSyncMetadataField [KnowledgeSyncMetadataField.java:23-59]
      -> defines logical keys such as DocumentId / ChunkHash
      -> maps them to canonical physical names such as document_id / chunk_hash

Metadata registration
  -> MetadataUtil.METADATA_FIELDS [MetadataUtil.java:39-50]
      -> CommonOptions metadata
      -> + KnowledgeSyncMetadataField values
  -> MetadataUtil.isMetadataField(fieldName) [122-123]

Metadata transform config parsing
  -> MetadataTransform.initOutputFields(...) [55-73]
      -> accepts the new Knowledge Sync logical keys
      -> still rejects unknown metadata keys

Output schema generation
  -> MetadataTransform.getOutputColumns() [92-123]
      -> requires metadataSchema.contains("DocumentId")
      -> renames the metadata column to the configured output field

Runtime projection
  -> MetadataTransform.getMetadataFieldValue(...) [126-137]
      -> reads inputRow.getOptions().get("DocumentId")
      -> reads inputRow.getOptions().get("ChunkHash")

Closest existing built-in producer path in this repo
  -> MarkdownReadStrategy.read(...) [118-145]
      -> still exposes physical fields such as source_uri / document_id / chunk_id / chunk_index / content_hash [89-99]
      -> does not automatically emit the new logical metadata keys

Docs contract
  -> docs/en/transforms/metadata.md:38-62
  -> docs/zh/transforms/metadata.md:38-62
      -> explicitly says Metadata transform does not generate Knowledge Sync metadata by itself
      -> explicitly says existing Markdown RAG physical fields are not auto-migrated into logical Knowledge Sync metadata
```

Key findings:
1. The normal path does hit this PR. Once a user configures `metadata_fields` with `DocumentId` / `ChunkHash`, the new registration and projection logic is on the main execution path.
2. The current implementation is intentionally contract-only. It does not pretend that existing producer paths already emit these logical keys.
3. That was the important blocker in my earlier review: the code and docs could previously be read as a ready-to-use producer feature. On the current head, that mismatch is gone.
4. I do not see a new blocker on this revision.

Test / stability note:
- The added unit tests are deterministic and structurally stable. I did not see any flaky-test patterns such as sleeps, timing races, port conflicts, or environment-sensitive assertions.
- `KnowledgeSyncMetadataFieldTest` covers registration, canonical physical names/types, and `toMetadataColumn()`.
- `MetadataTransformTest` covers both the happy path and the schema-missing rejection path.
- The remaining integration blind spot is acceptable for this PR scope, because the docs now clearly state that this is a contract definition and projection capability, not an already-wired built-in producer feature.

CI note:
- The current `Build` failure is concrete, not generic:
  - failing test: `org.apache.seatunnel.engine.client.SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus`
  - file: `seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java:395-494`
  - failing assertion: expected terminal state `CANCELED`, but the job status remained `CANCELING`
- This PR itself touches only `seatunnel-api`, `seatunnel-transforms-v2`, and the EN/ZH metadata docs.
- The current PR head is already `Merge branch 'apache:dev' into dev-Knowledge-Sync`, so this is not a case where “please just rebase onto the latest dev” is likely to solve it automatically.
- Based on the current evidence, this looks like an engine-side test failure brought in with the merged `dev` state, not a regression introduced by these seven PR-owned files.

### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- When a real producer starts emitting this Knowledge Sync metadata contract, I would add one producer-to-transform integration test so CI also covers the end-to-end runtime path.

Overall, I think the current shape is reasonable to merge. The important part is that the latest head now tells the truth about what the feature is: a stable logical metadata contract plus projection support, not an implicit producer-side auto-generation feature.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `74ba70f22665` from scratch.

# What this PR fixes
- User pain: There is no shared, recognized contract for Knowledge Sync document/chunk identity metadata, so future sources, transforms, and lifecycle sinks (issues #10914/#10964, umbrella #10889) cannot interoperate, and the existing `Metadata` transform rejects keys like `DocumentId` or `ChunkIndex` as unknown metadata fields.
- Fix approach: Add a `KnowledgeSyncMetadataField` enum in `seatunnel-api` defining ten logical keys (`DocumentId`, `DocumentHash`, `SourceUri`, `SourceVersion`, `SourceModifiedAt`, `MimeType`, `Deleted`, `ChunkId`, `ChunkHash`, `ChunkIndex`) with canonical physical names and types, register them into `MetadataUtil.METADATA_FIELDS`, extend `MetadataTransform.getOutputFieldValues` to resolve these keys from row options, and document the contract plus a projection example in the EN/ZH `metadata.md` docs.
- One-line summary: The contract-only change is coherent and well-tested for name/type/column mappings, but the review findings above (credential leakage via `SourceUri`, engine-wide `isMetadataField` semantics widening from generic key names like `Deleted`, hard-coded nullability in `toMetadataColumn()`, and the linear `values()` scan) should be addressed or explicitly acknowledged before merge.

# Runtime chain I rechecked
```text
MetadataUtil.<clinit> registers Knowledge Sync keys into METADATA_FIELDS  MetadataUtil.java:44-49
  -> Stream.of(KnowledgeSyncMetadataField.values()).map(getName)  KnowledgeSyncMetadataField.java:24-33
  -> MetadataUtil.isMetadataField(name) now returns true for the ten new keys (verified by KnowledgeSyncMetadataFieldTest.shouldRegisterKnowledgeSyncMetadataFields)

Transform execution path:
MetadataTransform.transformRow(inputRow)
  -> MetadataTransform.getOutputFieldValues(SeaTunnelRowAccessor)  MetadataTransform.java:83
    -> iterates metadataFieldMapping entries; for legacy keys dispatches via CommonOptions.fromName(metadataFieldName) (DATABASE -> MetadataUtil.getDatabase(inputRow), etc.)
    -> for Knowledge Sync keys, falls through to reading the declared metadata schema field from row options (values must be declared in CatalogTable.metadataSchema and written into SeaTunnelRow.options by the upstream producer; the transform does not synthesize them)

Producer-side schema declaration path (contract consumers):
KnowledgeSyncMetadataField.toMetadataColumn()  KnowledgeSyncMetadataField.java:56-58
  -> MetadataColumn.of(name, dataType, (Long) null, true, null, null)
  -> used by future sources/transforms to populate CatalogTable.metadataSchema (verified by shouldCreateMetadataColumnsForProducerSchemas: column.isPhysical() == false)
```

# Findings
**Issue 1: `SourceUri` metadata is promoted to a canonical physical column with no guidance to strip credentials embedded in URIs, risking secret leakage into downstream sinks**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/KnowledgeSyncMetadataField.java:27`
- Why it matters: - Root cause: the new contract defines `SOURCE_URI("SourceUri", "source_uri", BasicType.STRING_TYPE)` as a first-class, projectable metadata field, and the docs instruct users to project it as a regular physical column. Source URIs in file/FTP/SFTP/HTTP pipelines routinely contain userinfo credentials (e.g., `ftp://user:password@host/path`) or pre-signed query tokens (`?X-Amz-Signature=...`).
- Impact: once projected via the `Metadata` transform, these values become ordinary row data that flows into every downstream sink (Doris, Elasticsearch, JDBC, files), sink error logs, and dirty-data/error records — turning a transient connector-internal URI into persisted plaintext credentials with a much larger blast radius than the source config itself.
- Fix: state in the contract (Javadoc on `SOURCE_URI` and both metadata.md docs) that producers MUST write a sanitized URI (userinfo and signed query parameters stripped) into `SeaTunnelRow.options` for this key, and consider providing a shared sanitization helper in `seatunnel-api` so future Knowledge Sync sources implementing this contract (#10889) do it consistently.
- Evidence: `docs/en/transforms/metadata.md:46  | SourceUri | `source_uri` | string | Original source URI or path. |`
- Suggested fix: `SourceUri` will frequently carry credentials embedded in the URI (e.g., `ftp://user:pass@host/path`, pre-signed S3/HTTP URLs). Once a user projects it via the `Metadata` transform, those secrets become persisted physical column data in every downstream sink and can also surface in dirty-data/error logs. Since this enum is the shared contract for future Knowledge Sync producers, please document (here and in `docs/en|zh/transforms/metadata.md`) that producers MUST strip userinfo and signed query parameters before writing this value into `SeaTunnelRow.options`, and ideally add a shared sanitization helper so all implementations behave the same way.
- Severity: Medium

**Issue 2: Registering generic key names like `Deleted` and `MimeType` into the global `MetadataUtil.METADATA_FIELDS` set changes `isMetadataField` semantics engine-wide**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/MetadataUtil.java:47`
- Why it matters: - Root cause: the static initializer in `MetadataUtil` now unconditionally adds all ten Knowledge Sync names (`Deleted`, `MimeType`, `SourceVersion`, etc.) into the shared `METADATA_FIELDS` set, which backs `MetadataUtil.isMetadataField(...)` for the whole framework, not just Knowledge Sync pipelines.
- Impact: any component on any engine (Zeta/Flink/Spark) that consults `isMetadataField` to decide whether a name is reserved metadata will now treat these very common identifiers as metadata. A user column or option key literally named `Deleted` or `MimeType` in an existing pipeline can silently change behavior wherever this check gates schema/option handling — a subtle compatibility break introduced by a contract-only PR.
- Fix: either namespace the logical keys to make collisions implausible (e.g., `KnowledgeSync.DocumentId` / `KsDeleted`), or keep the Knowledge Sync registry separate (`KnowledgeSyncMetadataField.isKnowledgeSyncMetadataField`) and only consult it inside the `Metadata` transform where projection is explicitly configured, instead of widening the global reserved-name set.
- Evidence: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/MetadataUtil.java:49  .forEach(METADATA_FIELDS::add);`; `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/KnowledgeSyncMetadataField.java:31  DELETED("Deleted", "deleted", BasicType.BOOLEAN_TYPE),`
- Suggested fix: Adding all Knowledge Sync names into the global `METADATA_FIELDS` set makes very common identifiers like `Deleted`, `MimeType`, and `SourceVersion` reserved for every pipeline via `MetadataUtil.isMetadataField(...)`, not just Knowledge Sync ones. That risks silent behavior changes for existing jobs that happen to use these names in options/schemas. Could we either namespace the logical keys (e.g., `KnowledgeSync.Deleted`) or scope the check via `KnowledgeSyncMetadataField.isKnowledgeSyncMetadataField(...)` inside the `Metadata` transform only, rather than widening the global reserved set?
- Severity: Medium

**Issue 3: `toMetadataColumn()` hard-codes every Knowledge Sync column as nullable, including identity fields (`DocumentId`, `ChunkId`) and the `Deleted` tombstone flag**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/KnowledgeSyncMetadataField.java:59`
- Why it matters: - Root cause: `MetadataColumn.of(name, dataType, (Long) null, true, null, null)` passes `nullable = true` uniformly for all ten fields, with no way for a producer to declare a field as required.
- Impact: the contract explicitly positions `DocumentId`/`ChunkId` as "stable identity" and `Deleted` as a deletion marker for future lifecycle sinks (#10889). A nullable `deleted` BOOLEAN yields three-valued semantics (null vs. false), and a lifecycle sink that later drives delete/upsert decisions off a null tombstone or null document identity can silently skip deletions or misroute chunks — a data-correctness hazard baked into the contract now, before any implementation exists.
- Fix: make nullability part of the enum contract (e.g., a `nullable` constructor argument) so `DOCUMENT_ID`, `CHUNK_ID`, and `DELETED` are declared non-nullable, or at minimum document the required semantics in Javadoc so producers and the planned lifecycle sink agree on null handling.
- Evidence: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/KnowledgeSyncMetadataField.java:31  DELETED("Deleted", "deleted", BasicType.BOOLEAN_TYPE),`
- Suggested fix: `MetadataColumn.of(name, dataType, (Long) null, true, null, null)` marks every field nullable, including `DocumentId`, `ChunkId`, and `Deleted`. Since this is the contract future lifecycle sinks will rely on, a null tombstone (`Deleted`) or null identity is ambiguous and can cause silent skipped deletions or misrouted chunks. Consider adding a `nullable` flag to the enum so identity/tombstone fields are non-nullable, or at least document the required-vs-optional semantics per field.
- Severity: Medium

**Issue 4: `isKnowledgeSyncMetadataField` does a linear `values()` scan (allocating a new array) on every invocation**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/KnowledgeSyncMetadataField.java:63`
- Why it matters: - Root cause: `isKnowledgeSyncMetadataField` iterates `values()` per call; `Enum.values()` clones the backing array on each invocation, so this is both O(n) and allocating.
- Impact: if this check ends up on the per-row path of the `Metadata` transform or future Knowledge Sync producers (the same hot path that `MetadataUtil.isMetadataField` serves via a pre-built set), it adds avoidable per-row allocation and scan cost under high throughput on Zeta/Flink/Spark.
- Fix: precompute a `private static final Map<String, KnowledgeSyncMetadataField> NAME_INDEX` (or immutable `Set<String>`) in a static initializer and back both `isKnowledgeSyncMetadataField` and a future `fromName` lookup with it, mirroring how `MetadataUtil` builds `METADATA_FIELDS` once.
- Evidence: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/type/KnowledgeSyncMetadataField.java:64  if (field.getName().equals(name)) {`
- Suggested fix: `values()` clones the enum array on every call, so this lookup is O(n) with an allocation each time. If this ends up on the per-row path of the transform or future producers, please back it with a `private static final Map<String, KnowledgeSyncMetadataField>` built once in a static initializer (same pattern as `MetadataUtil.METADATA_FIELDS`), which would also give you a `fromName(...)` for free.
- Severity: Low

# Review conclusion

- Issue 1: `SourceUri` metadata is promoted to a canonical physical column with no guidance to strip credentials embedded in URIs, risking secret leakage into downstream sinks
- Issue 2: Registering generic key names like `Deleted` and `MimeType` into the global `MetadataUtil.METADATA_FIELDS` set changes `isMetadataField` semantics engine-wide
- Issue 3: `toMetadataColumn()` hard-codes every Knowledge Sync column as nullable, including identity fields (`DocumentId`, `ChunkId`) and the `Deleted` tombstone flag
- Issue 4: `isKnowledgeSyncMetadataField` does a linear `values()` scan (allocating a new array) on every invocation

DanielLeens:
Thanks @SEZ9. I agree these follow-up points should be resolved or explicitly narrowed before merge, especially the widened `isMetadataField` semantics and the `SourceUri` / nullability contract details. From Daniel's side the PR direction is still coherent, but I would wait for a code update or a narrower contract statement before another full re-review.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=4`, `behind_by=192`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

DanielLeens:
Rechecked the same unchanged head `74ba70f22665` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
Knowledge Sync contract
  -> KnowledgeSyncMetadataField [KnowledgeSyncMetadataField.java:23-59]
      -> defines logical keys such as DocumentId / ChunkHash
      -> maps them to canonical physical names such as document_id / chunk_hash

Metadata registration
  -> MetadataUtil.METADATA_FIELDS [MetadataUtil.java:39-50]
      -> CommonOptions metadata
      -> + KnowledgeSyncMetadataField values
  -> MetadataUtil.isMetadataField(fieldName) [122-123]

Metadata transform config parsing
  -> MetadataTransform.initOutputFields(...) [55-73]
      -> accepts the new Knowledge Sync logical keys
      -> still rejects unknown metadata keys

Output schema generation
  -> MetadataTransform.getOutputColumns() [92-123]
      -> requires metadataSchema.contains("DocumentId")
      -> renames the metadata column to the configured output field

Runtime projection
  -> MetadataTransform.getMetadataFieldValue(...) [126-137]
      -> reads inputRow.getOptions().get("DocumentId")
      -> reads inputRow.getOptions().get("ChunkHash")

Closest existing built-in producer path in this repo
  -> MarkdownReadStrategy.read(...) [118-145]
      -> still exposes physical fields such as source_uri / document_id / chunk_id / chunk_index / content_hash [89-99]
      -> does not automatically emit the new logical metadata keys

Docs contract
  -> docs/en/transforms/metadata.md:38-62
  -> docs/zh/transforms/metadata.md:38-62
      -> explicitly says Metadata transform does not generate Knowledge Sync metadata by itself
      -> explicitly says existing Markdown RAG physical fields are not auto-migrated into logical Knowledge Sync metadata
```

## Merge conclusion

1. Blocking items

2. Current gate
- The current CI signal is now green: https://github.com/apache/seatunnel/actions/runs/28992837878/job/86036234818

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current diff and the Knowledge Sync metadata path. The intended feature is clear, but one diff-integrity blocker remains.

## Blocking finding

**Issue 1: the branch includes an unrelated Zeta metrics-test change**

- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/metrics/MetricsApiTest.java:54-71`
- Commit `7802f46f8881` changes engine metrics readiness behavior, which has no call-chain relationship to the Knowledge Sync API or Metadata transform contract.
- Please remove this hunk from the branch or split it into its own focused PR. The current unit-test JDK 11 failure also overlaps this unrelated engine-test change, so it cannot be treated as harmless baseline noise here.

Severity: Medium. Merge gate: blocking under the complete-diff integrity gate. The Knowledge Sync credential/nullability/lookup concerns from earlier rounds are resolved. Happy to re-review a focused head.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch. The Knowledge Sync metadata-contract part is much clearer now, and the previous credential/nullability/lookup concerns are addressed in the API contract, transform behavior, tests, and EN/ZH docs.

There are still two blockers in the complete current diff before I can treat this PR as merge-ready.

## Runtime path I checked

```text
User configures the Metadata transform
  -> MetadataTransform.initOutputFields()
      -> MetadataUtil.isMetadataField("DocumentId")
          -> CommonOptions metadata fields
          -> + KnowledgeSyncMetadataField values
      -> validates that the mapped output field does not duplicate an existing physical field

Output schema generation
  -> MetadataTransform.getOutputColumns()
      -> computed common metadata such as Database / Table / RowKind is generated directly
      -> Knowledge Sync metadata requires inputCatalogTable.getMetadataSchema().contains("DocumentId")
      -> MetadataColumn.rename("document_id").toPhysicalColumn()
      -> if the metadata schema does not declare the key, the transform still rejects it

Runtime projection
  -> MetadataTransform.map(row)
      -> getOutputFieldValues()
          -> Database / Table / RowKind are computed from the row accessor
          -> DocumentId / ChunkHash / etc. are read from row.options
      -> projected metadata becomes normal physical output fields for downstream components
```

That main design looks coherent to me: the transform accepts the logical contract, but it does not pretend to generate Knowledge Sync metadata by itself or read same-named physical columns.

## Blocking findings

**Issue 1: the new file-collector test uses a fixed sleep to create the inactive-cursor condition**

- Location: `seatunnel-edge-agent/seatunnel-edge-agent-connector/src/test/java/org/apache/seatunnel/edge/agent/connector/file/FileCollectReaderBehaviorTest.java:133`
- The new `readsDataAppendedAfterCursorBecomesInactive()` test sleeps for `closeInactiveMs + 50L`, appends data, and immediately asserts that a single `poll()` returns the new line.
- This is timing-sensitive: cursor inactivity, glob rediscovery, file-size visibility, GC pauses, and CI scheduling are not controlled by that fixed sleep.
- Please replace this with a condition-based wait, for example an Awaitility loop that keeps polling until the `late` record is observed. If the test needs to prove that the cursor was actually closed first, it would be better to use a deterministic test hook or injectable clock rather than relying on wall-clock sleep.

Severity: High. Merge gate: blocking because this is a new UT flaky-test risk.

**Issue 2: the current head still contains an unrelated edge/Zeta file-collector change**

- Location: `seatunnel-edge-agent/seatunnel-edge-agent-connector/src/main/java/org/apache/seatunnel/edge/agent/connector/file/FileCollectReader.java:279`
- This PR is about the `seatunnel-api` Knowledge Sync metadata contract and `MetadataTransform` projection path, but the current diff also changes `FileCollectReader.closeInactiveCursors()`, adds `FileTailCursor.hasUnreadData()`, and adds file-collector behavior tests.
- I agree this file-collector behavior may be worth fixing, but it has no direct call-chain relationship to the Knowledge Sync metadata contract reviewed here. This is the same PR-scope/integrity concern as the earlier unrelated Zeta test hunk, just in a different edge collector area.
- Please remove this hunk from this branch or split it into its own focused PR. Once this PR only contains the Knowledge Sync metadata contract/transform/docs/tests, that part looks close from my side.

Severity: Medium. Merge gate: blocking under the focused-diff rule.

## Tests and CI

For the Knowledge Sync/API/transform tests, the coverage is good: field registration, canonical physical names/types, nullability, missing metadata-schema rejection, and row-options projection are all covered.

Test stability rating for the current full diff: **high risk**, because of Issue 1 above.

The current Build failure is in the MySQL CDC E2E lane, not in the Knowledge Sync or edge collector unit-test path I reviewed:

- Build check: https://github.com/apache/seatunnel/runs/90500268642
- Failing fork job: https://github.com/yzeng1618/seatunnel-dev/actions/runs/30428234038/job/90598354889
- Failed lane: `all-connectors-it-7 (11, ubuntu-latest)` / `connector-cdc-mysql-e2e`

This does not look directly caused by the current PR diff, and the branch is currently diverged from `dev` with `behind_by=2`. After the two source-side blockers above are cleaned up, the lowest-cost CI step is to sync the latest `dev` and rerun Build first.

## Merge conclusion

1. Blocking items
   - Issue 1: replace the fixed-sleep file-collector test with a deterministic condition wait or test hook.
   - Issue 2: remove or split out the unrelated edge file-collector change from this Knowledge Sync metadata PR.

2. Non-blocking suggestions
   - After the source blockers are fixed, sync the latest `dev` and rerun the failing Build lane, since the visible failure is in MySQL CDC E2E and does not appear tied to this diff.

Overall, the Knowledge Sync metadata contract itself is in good shape now. Thanks for continuing to narrow the behavior and documentation; this is close once the unrelated hunk is taken out cleanly.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `ad4ad6b8cfd5` from scratch.

The previous Daniel blockers are cleared for me on this head: the unrelated edge/file-collector hunk is gone, and the remaining diff is focused on the Knowledge Sync metadata contract, `MetadataTransform`, tests, and EN/ZH docs.

# What this PR fixes
- User pain: Knowledge Sync did not have one stable logical metadata-field contract, and the `Metadata` transform rejected keys such as `DocumentId` or `ChunkHash` as unknown metadata fields.
- Fix approach: add `KnowledgeSyncMetadataField`, register those keys in `MetadataUtil`, and let `MetadataTransform` project them only when the upstream metadata schema declares the keys and row options carry the values.
- In one sentence: this is now a clear contract/projection PR, not a claim that existing producer paths automatically generate Knowledge Sync metadata.

# Runtime path checked
```text
Metadata transform init
  -> MetadataTransform.initOutputFields()
      -> MetadataUtil.isMetadataField("DocumentId")
          -> CommonOptions metadata fields
          -> + KnowledgeSyncMetadataField values

Output schema generation
  -> MetadataTransform.getOutputColumns()
      -> Database / Table / RowKind are computed common metadata
      -> Knowledge Sync keys require metadataSchema.contains("DocumentId")
      -> MetadataColumn.rename("document_id").toPhysicalColumn()

Runtime projection
  -> MetadataTransform.map(row)
      -> getOutputFieldValues()
          -> common metadata from the row accessor
          -> Knowledge Sync metadata from row.options
      -> projected metadata becomes normal physical output fields downstream
```

The normal `Metadata` transform path does hit this change. The implementation still correctly rejects a Knowledge Sync key if the input metadata schema does not declare it, so it does not accidentally read same-named physical columns.

Fully compatible. This adds logical metadata keys; it does not change existing options, defaults, public APIs in an incompatible way, protocols, serialization formats, or existing physical field behavior.

# Test coverage and stability
The tests cover field registration, canonical physical names/types, nullability, metadata-column creation, missing metadata-schema rejection, and row-options projection. They are deterministic unit tests without sleeps, network, containers, fixed ports, or ordering assumptions. Stability rating: stable.

# CI status
The current `Build` check is failing in the fork workflow, specifically `unit-test (8, windows-latest)` with related cancelled unit-test jobs:
https://github.com/yzeng1618/seatunnel-dev/actions/runs/30507237953/job/90759991496

I could not retrieve the failing job log body because the GitHub log API returned a stream cancellation, so I am not going to guess the underlying failure. Please inspect that job output and rerun once addressed.

# Merge conclusion
### Conclusion: can merge after CI recovers

1. Blocking items that must be fixed
- No source-level blocker found in this review.
- The required Build check is still failing and needs to recover before merge.

2. Non-blocking suggestions
- None from this review.

Overall, the Knowledge Sync metadata contract looks focused and ready from the source side now. Thank you for cleaning up the unrelated hunk; that made the review much cleaner.

SEZ9:
Thanks @DanielLeens for the thorough recheck and for confirming the `Build` gate is now green on `74ba70f22665`.

Since the head is unchanged, my earlier findings still stand as the remaining asks before I can approve:

1. **`SourceUri` credential leakage** — the docs contract should explicitly require producers to strip userinfo/credentials from URIs before writing `SourceUri`, or at minimum call out the risk. A one-paragraph addition to both EN/ZH `metadata.md` would be sufficient.
2. **Widened `isMetadataField` semantics** — generic key names like `Deleted` now return `true` engine-wide via `MetadataUtil.isMetadataField`. Please either namespace the keys, scope the check, or add an explicit statement (code comment + docs) acknowledging this widening is intentional so downstream callers are not surprised.
3. **Hard-coded nullability in `toMetadataColumn()`** — some of these fields (e.g. `DocumentId`) are arguably non-nullable by contract; either derive nullability per field or document why everything is nullable.
4. **Linear `values()` scan** — minor, but a static name-keyed map would avoid the repeated scan on hot paths.

Items 1–3 can be resolved either with a code update or a narrower, explicit contract statement in the docs — I'm fine with either, but one of them needs to land on this PR.

Two housekeeping notes:
- The branch is `behind_by=192` against `dev`; a rebase alongside the fixes above would be appreciated so we're validating against current `dev`.
- Once the update lands, I'll do the re-review and refresh/dismiss the stale `changes requested` state — no separate action needed from your side, Daniel.

The overall direction (contract-only, well-tested mappings, clear docs on what the transform does *not* generate) remains good. Looking forward to the next revision.

yzeng1618:
I checked the current PR head again: `ad4ad6b8cfd5`. The four findings quoted above were based on the earlier `74ba70f` head; the current head is not unchanged from that revision.

1. **`SourceUri` sanitization** is now explicitly required in both the `SOURCE_URI` Javadoc and EN/ZH `metadata.md`: producers must remove URI user info, access tokens, signatures, and other transient authentication material before writing `SourceUri` to row options.

2. **`MetadataUtil.isMetadataField` scope**: the added keys are logical metadata keys for explicit `Metadata` transform projection. The transform still requires the input `metadataSchema` to declare the key and reads the value from row options; same-named physical columns are not classified or read as metadata. In the current repository, the direct caller is the `Metadata` transform, so this does not silently change ordinary physical columns or pipelines that do not opt into the mapping.

3. **Nullability** is now part of the enum contract. `DocumentId` and `Deleted` are non-nullable; chunk fields remain nullable specifically for document tombstones. This is covered by `KnowledgeSyncMetadataFieldTest`.

4. **Lookup cost** is now a precomputed immutable `FIELD_NAMES` set, so `isKnowledgeSyncMetadataField` does not scan `values()` per invocation.

The current diff is also focused on the Knowledge Sync API/Metadata transform/docs/tests only; the prior unrelated hunks have been removed. The required Build check is green on this head.

For branch freshness, the current branch is 10 commits behind `dev` (not 192). I will refresh it against the current base before final merge validation.

DanielLeens:
Thanks @SEZ9 and @yzeng1618 for the careful current-head follow-up.

I agree that the review discussion should be anchored to the actual current head, not the earlier `74ba70f` revision. Since the latest response says the source URI sanitization docs/Javadocs and metadata-field scope have changed in the current head, I am not going to restate the older findings as Daniel blockers in this no-new-commit reply. Please let the current reviewer/author discussion converge on the latest head; if a new commit is pushed, I will re-review the complete metadata contract again.

davidzollo (approved):
+1
LGTM

DanielLeens:
Thanks @davidzollo.

From Daniel's side, I do not have an additional unchanged-head blocker to add in this reply. The current reviewer/author discussion is already on the right latest-head context, so I would let that converge on this revision. If a new commit is pushed later, I can re-review the complete metadata contract again on that new head.

PR #10994, opened 2026-06-02
[Fix][Engine] Fix JDK 8 NoSuchMethodError in StainTracePayload#append

davidzollo:
## Purpose

Fix a JDK 8 runtime incompatibility in `StainTracePayload#append`.

## Problem

In JDK 9+, `ByteBuffer` added a covariant override of `Buffer.position(int)` that returns `ByteBuffer` instead of `Buffer`.  
When the project is **compiled on JDK 9+** (which emits bytecode referencing `ByteBuffer.position(I)LByteBuffer;`) and then **run on JDK 8**, the JVM cannot resolve that method descriptor and throws:

```
java.lang.NoSuchMethodError: java.nio.ByteBuffer.position(I)Ljava/nio/ByteBuffer;
```

The affected line is in `append()`:

```java
// before – compiled reference to ByteBuffer.position(I)LByteBuffer; → NoSuchMethodError on JDK 8
buffer.position(payload.length);
```

## Fix

Cast to the `Buffer` superclass so the call always resolves to `Buffer.position(I)LBuffer;`, which exists on every JDK version:

```java
((Buffer) buffer).position(payload.length);
```

This is the standard idiom used throughout the JDK and many libraries to keep `ByteBuffer` code compatible with JDK 8.

## Checklist

- [x] Code compiles and existing unit tests (`StainTracePayloadTest`) pass
- [x] No behaviour change – only the method-resolution path in bytecode is affected

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -1240,7 +1240,8 @@ jobs:
       matrix:
         java: [ '8', '11' ]
         os: [ 'ubuntu-latest' ]
-    timeout-minutes: 60
+    # Kudu E2E expands each @TestTemplate case across several PR test containers.
+    timeout-minutes: 90
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK ${{ matrix.java }}
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-pulsar-e2e/src/test/java/org/apache/seatunnel/e2e/connector/pulsar/PulsarBatchIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-pulsar-e2e/src/test/java/org/apache/seatunnel/e2e/connector/pulsar/PulsarBatchIT.java
@@ -47,11 +47,7 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.TestTemplate;
 import org.testcontainers.containers.Container;
 import org.testcontainers.containers.PulsarContainer;
-import org.testcontainers.containers.output.Slf4jLogConsumer;
-import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.shaded.org.awaitility.Awaitility;
-import org.testcontainers.utility.DockerImageName;
-import org.testcontainers.utility.DockerLoggerFactory;
 
 import lombok.extern.slf4j.Slf4j;
 
@@ -62,7 +58,6 @@ import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Stream;
 
 @Slf4j
 public class PulsarBatchIT extends TestSuiteBase implements TestResource {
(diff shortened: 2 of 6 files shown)


nzw921rx:
+1 good job🚀

DanielLeens (commented):
Thanks for working on this. I reviewed the latest head from scratch and checked the exact binary-payload append path that runs when stain-trace sampling records another stage hop.

# What this PR solves
- User pain: on JDK 8, `StainTracePayload.append(...)` can hit `NoSuchMethodError` because the compiled call targets the JDK 9+ covariant `ByteBuffer.position(int)` signature instead of the JDK 8 `Buffer.position(int)` signature.
- Fix approach: cast the wrapped buffer to `Buffer` before calling `position(int)`, so the call binds to the JDK 8-compatible superclass method.
- One-line summary: this is a precise compatibility fix on the real payload-append hot path, and I did not find a new source-level blocker.

# Runtime path I checked
```text
sampled row trace path
  -> StainTracePayload.append(...) [81-101]
      -> validate payload
      -> expand byte array
      -> move write cursor to old payload length
      -> append stage code / task ID / timestamp
      -> update entry count
```

# Review result
- `StainTracePayload.java:90-100` is exactly the right place to fix this.
- The new cast at `StainTracePayload.java:92-96` is minimal and targeted: it preserves the existing binary layout while avoiding the JDK 8 linkage problem.
- I did not see a reopened compatibility or payload-corruption risk from this change.

# Tests / CI
- The code change is very small and the compatibility rationale is clear from the in-code comment.
- The current apache-side `Build` check is red, but the corresponding fork Actions run I inspected is still in progress and has not yet produced a clean final PR-specific failure diagnosis. So I am not treating that red badge as source evidence against this fix.

1. Blocking items
- No source-level blocker from my side on the latest head.
- Please make sure the branch ends up with one valid CI result before merge, because the current public `Build` badge is not a trustworthy final signal yet.

- None from my side on this code path.

From the code-review side, the JDK 8 compatibility fix itself looks good.

DanielLeens (approved):
Thanks for the update. I re-checked the latest head from scratch, including the original JDK 8 compatibility fix and the new workflow adjustment.

What this PR fixes:
- User pain: `StainTracePayload.append(...)` can hit `NoSuchMethodError` on JDK 8.
- Fix approach: bind the append path to the JDK 8-compatible `Buffer.position(int)` call; the latest head also increases the Kudu E2E workflow timeout to match real template fan-out.
- One-line summary: the code fix still looks correct, and the newest workflow delta is aligned with the CI symptom rather than masking a code problem.

Runtime path re-checked:
```text
StainTracePayload.append(...)
  -> move buffer cursor
  -> append stage/task/timestamp bytes
```

Findings:
- No blocking source-level issue on the current head.
- The production compatibility fix remains correct on `StainTracePayload.java:90-96`.
- The latest delta in `.github/workflows/backend.yml:1240` simply increases the Kudu E2E timeout from 60 to 90 minutes, which matches the expanded template load.

CI:
- `statusCheckRollup` still shows an old failing `Build` check-run, but `head_sha=e6b2ad3...` maps to the latest Apache workflow runs and both are successful.

### Conclusion: can merge
1. Blocking items
- None.
2. Suggested follow-up
- None.

Overall, this is mergeable.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch, including the original JDK 8 compatibility fix and the new Pulsar E2E / workflow follow-up commits.

# What this PR solves
- User pain: once stain-trace sampling hits, `StainTracePayload.append(...)` runs on the real source/transform/sink trace path. On JDK 8, a direct `ByteBuffer.position(int)` call can link to the JDK 9+ covariant signature and fail at runtime with `NoSuchMethodError`.
- Fix approach: the latest head binds that cursor move to `Buffer.position(int)` via `((Buffer) buffer).position(...)`, while keeping the payload binary layout unchanged.
- One-line summary: this is still a precise compatibility fix on the real hot path, and I did not find a new source-level blocker in the latest head.

# Runtime path I checked
```text
sampled row trace path
  -> SeaTunnelSourceCollector.tryStainTrace(...) [SeaTunnelSourceCollector.java:342-369]
      -> StainTracePayload.init(...)
      -> StainTracePayload.append(...) [StainTracePayload.java:81-101]

row continues through later stages
  -> StainTraceUtils.tryAppendPayload(...) [StainTraceUtils.java:72-91]
      -> StainTracePayload.append(...) again
```

# Review result
- `StainTracePayload.java:90-96` is exactly the right place to fix the JDK 8 linkage issue.
- The `Buffer` cast keeps the on-wire payload format unchanged, so I do not see a serialization or compatibility regression here.
- The new Pulsar E2E helper and test updates look structurally stable as well: no `Thread.sleep`, no fixed-port assumptions, and the image pre-pull path has an explicit timeout.

# CI
- `Build` is still running on the latest head at the time of this review, so I am not treating CI as a negative code signal right now.

### Conclusion: can merge
1. Blocking items
- None from my side on the current source changes.

2. Suggested follow-up
- None.

From the code-review side, this latest head looks good to merge once the running CI finishes cleanly.

dybyte (approved):

PR #10998, opened 2026-06-02
[Fix][Connector-V2] Adapt JdbcHiveIT to JdbcCatalogUtils getTables signature

yzeng1618:

image
This patch passes `MultiTableFailurePolicy.FAIL_FAST`, matching the existing behavior used by `AbstractJdbcIT`.

### Check list


dybyte (approved):
+1 if CI passes

DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff locally and traced the only affected path here: `JdbcHiveIT` -> `JdbcCatalogUtils.getSourceTables(...)` -> the updated `MultiTableFailurePolicy`-accepting overload.

# What this PR fixes
- User pain: after the `JdbcCatalogUtils.getTables` signature change, `JdbcHiveIT` no longer matched the runtime factory call shape.
- Fix approach: pass the explicit `MultiTableFailurePolicy.FAIL_FAST` argument so the IT keeps calling the same production API shape as the source path now expects.
- One-line summary: this is a narrow compatibility fix for the Hive JDBC IT, and I did not find a blocker on the latest head.

# Runtime chain I checked
```text
JdbcHiveIT.test...
  -> JdbcCatalogUtils.getSourceTables(...)
     -> source-table discovery with explicit multi-table failure policy
```

# Review result
- The change is surgical and stays on the intended IT path in `JdbcHiveIT.java:232-239`.
- It matches the current production method signature instead of relying on an old overload shape.
- I agree with @dybyte's earlier approval direction here; I do not have an additional blocker on top of that.

# Tests / stability
- This PR only adjusts an existing IT call site.
- I do not see a new flaky-test pattern in the change itself.
- The current PR checks are green.

### Conclusion: can merge
1. Blocking items

- None.

From the code-review side this looks good to merge.

PR #11003, opened 2026-06-03
[Fix][Connector-V2] Fix Paimon option rule declarations

QuakeWang:

Paimon defines and reads several valid options, including `paimon.table.non-primary-key`, `catalog_name`, `user`, and `password`, but some of them were missing from the connector option rules.

This causes static config validation to report legal Paimon configs as unknown keys, because the unknown-key check only accepts options declared by `optionRule()`.

This PR adds the missing Paimon options to the source, sink, and catalog option rules, and adds unit tests to prevent the declarations from regressing.

Yes. Static config validation no longer rejects valid Paimon configs that use the declared Paimon options.

Added `PaimonOptionRuleTest`.

Tested with:

  - `./mvnw -pl seatunnel-connectors-v2/connector-paimon spotless:check`
  - `./mvnw -pl seatunnel-connectors-v2/connector-paimon -Dtest=PaimonOptionRuleTest -Dskip.spotless=true test`

### Check list


davidzollo:
Does the paimon doc need to be updated?

QuakeWang:
> Does the paimon doc need to be updated?

@davidzollo  Thanks for pointing this out. `paimon.table.non-primary-key`, `user`, and `password` were already documented, but `catalog_name` was used in Paimon examples and declared in the option rule without being listed in the option tables.

I added `catalog_name` to the English and Chinese Paimon source/sink docs.

DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff from the real option-validation path instead of only looking at the docs delta.

What this PR fixes:
- User pain: valid Paimon options were implemented but not declared in `OptionRule`, so config validation could reject legitimate keys as unknown.
- Fix approach: add the missing option declarations in source/sink/catalog factories, cover them with focused tests, and sync the docs.
- One-line summary: this is a precise declaration-alignment fix.

Runtime path re-checked:
```text
factory optionRule()
  -> validateUnknownKeys(...)
  -> declared option set must match the real implementation surface
```

Findings:
- No blocking source-level issue found.
- `PaimonCatalogFactory.java:47-60`, `PaimonSinkFactory.java:50-63`, and `PaimonSourceFactory.java:49-61` now declare the real optional keys (`catalog_name`, `user`, `password`, `non_primary_key`) that the connector already supports.
- `PaimonOptionRuleTest.java:31-92` gives this regression a clean unit-test pin.
- Docs were updated in both `docs/en` and `docs/zh`.
- Test stability rating: **stable**.

CI:
- Current Apache `Build` is already successful on this head.

### Conclusion: can merge
1. Blocking items
- None.
2. Suggested follow-up
- None.

Overall, this is mergeable.

chl-wxp (approved):

PR #11004, opened 2026-06-03
[Fix][Connector-V2][CDC-MySQL] Honor int_type_narrowing option on MyS…

ricky2129:
…QL-CDC source

The `int_type_narrowing` option is documented for the MySQL-CDC source (including a `false` example) and declared in MySqlIncrementalSourceFactory's option rule, but it had no effect on the CDC path, for two reasons:

1. MySqlTypeUtils built the column type from Debezium's bare type name ("TINYINT") and never re-appended the length, so MySqlTypeConverter's narrowing check (columnType.equalsIgnoreCase("tinyint(1)")) never matched — tinyint(1) could not be detected on the CDC path.
2. convertToSeaTunnelColumn always used MySqlTypeConverter.DEFAULT_INSTANCE (narrowing = true), ignoring the configured value.

This change:
- re-appends the length for TINYINT so tinyint(1) is detectable; and
- carries int_type_narrowing through the Debezium properties and selects the converter: true (default) -> tinyint(1) narrows to BOOLEAN (matching the JDBC source and the documented default); false -> tinyint(1) stays TINYINT (Byte).

Adds MySqlTypeUtilsIntTypeNarrowingTest covering true / false / default.

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/config/MySqlSourceConfigFactory.java
+++ b/seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/config/MySqlSourceConfigFactory.java
@@ -87,6 +87,12 @@ public class MySqlSourceConfigFactory extends JdbcSourceConfigFactory {
         // but it'll cause lose of precise when the value is larger than 2^63,
         // so use "precise" mode to avoid it.
         props.put("bigint.unsigned.handling.mode", "precise");
+        // default int_type_narrowing (tinyint(1) -> boolean); the Source / user debezium block
+        // overrides this via the dbzProperties merge below so MySqlTypeUtils can honor a false.
+        // Carried as a property, NOT a field, on purpose: adding a field/method to this
+        // Serializable factory drifts its serialVersionUID and breaks rolling upgrades (jobs
+        // submitted on the prior version fail to deserialize on the new one).
+        props.setProperty("int_type_narrowing", String.valueOf(true));
 
         if (serverIdRange != null) {
             props.setProperty("database.server.id.range", String.valueOf(serverIdRange));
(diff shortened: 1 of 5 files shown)


ricky2129:
@davidzollo please help with this.

DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff from the actual MySQL-CDC schema/type-conversion path.

What this PR fixes:
- User pain: `int_type_narrowing` is documented for MySQL-CDC, but the CDC type-mapping path ignored it and always narrowed `tinyint(1)` to BOOLEAN.
- Fix approach: carry the option through `MySqlSourceConfigFactory` into Debezium properties, then let `MySqlTypeUtils` choose the correct converter and reconstruct `TINYINT(1)` on the CDC path.
- One-line summary: this makes the documented option actually work on the CDC path without changing the default behavior.

Runtime path re-checked:
```text
MySqlIncrementalSource.createSourceConfigFactory(...)
  -> MySqlSourceConfigFactory.intTypeNarrowing(...)
  -> Debezium properties
  -> MySqlTypeUtils.convertToSeaTunnelColumn(...)
  -> choose DEFAULT vs NO_INT_NARROWING converter
```

Findings:
- No blocking source-level issue found.
- `MySqlIncrementalSource.java:84-87` now propagates the option.
- `MySqlSourceConfigFactory.java:34-35,51-60,99-102` stores it in the Debezium properties.
- `MySqlTypeUtils.java:142-163` restores `TINYINT(1)` shape on the CDC path and honors the option when selecting the converter.
- The new unit test covers enabled / disabled / absent states.
- Test stability rating: **stable**.

CI:
- `statusCheckRollup` still carries an old failing `Build` check-run, but `head_sha=2bc3d629...` maps to the latest Apache workflow runs and both are successful.

### Conclusion: can merge
1. Blocking items
- None.
2. Suggested follow-up
- None.

Overall, this is mergeable.

davidzollo (approved):
+1 if CI passes.
LGTM

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head and traced the full path from `MySqlIncrementalSource.createSourceConfigFactory(...)` to `MySqlSourceConfigFactory`, then into the Debezium properties consumed by `MySqlTypeUtils.convertToSeaTunnelColumn(...)`.

`int_type_narrowing` is now carried all the way through the CDC path, and the new `MySqlTypeUtilsIntTypeNarrowingTest` covers the three cases that matter here: explicit `true`, explicit `false`, and default behavior. I also checked the `tinyint(1)` handling branch you added in `MySqlTypeUtils`: that change is what makes the option reachable on the CDC code path, so the fix is hitting the real runtime path rather than just the unit test.

I also noticed @davidzollo already approved this PR. After re-checking the latest head, I agree with that direction.

1. Blocking items
- None from my side.

2. Suggested follow-ups
- None that should block this PR.

Overall, this looks like a precise fix for a real behavior gap in the MySQL CDC source, and the latest version is in good shape.

PR #11009, opened 2026-06-04
[Fix][Connector-V2] Respect Paimon alter table ignore flag

QuakeWang:

 The Paimon catalog wrapper ignored the `ignoreIfNotExists` argument in both `alterTable` overloads and always passed `true` to the underlying Paimon catalog. As a result, alter table operations that should fail when the target table does not exist were silently ignored.

This PR passes `ignoreIfNotExists` through correctly and maps Paimon table-not-found errors to SeaTunnel `TableNotExistException`.

Yes. Paimon alter table operations with `ignoreIfNotExists = false` now fail when the target table does not exist, instead of being silently ignored.

Added unit coverage for both single `SchemaChange` and list-based `alterTable` overloads.

Ran:

  - `./mvnw -pl seatunnel-connectors-v2/connector-paimon -Dtest=PaimonCatalogTest test`
  - `./mvnw -pl seatunnel-connectors-v2/connector-paimon test`
  - `./mvnw -pl seatunnel-connectors-v2/connector-paimon spotless:check`

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/catalog/PaimonCatalog.java
+++ b/seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/catalog/PaimonCatalog.java
@@ -358,9 +358,12 @@ public class PaimonCatalog implements Catalog, PaimonTable {
     public void alterTable(
             Identifier identifier, SchemaChange schemaChange, boolean ignoreIfNotExists) {
         try {
-            catalog.alterTable(identifier, schemaChange, true);
+            catalog.alterTable(identifier, schemaChange, ignoreIfNotExists);
         } catch (org.apache.paimon.catalog.Catalog.TableNotExistException e) {
-            throw new CatalogException("TableNotExistException: {}", e);
+            throw new TableNotExistException(
+                    this.catalogName,
+                    TablePath.of(identifier.getDatabaseName(), identifier.getTableName()),
+                    e);
         } catch (org.apache.paimon.catalog.Catalog.ColumnAlreadyExistException e) {
             throw new CatalogException("ColumnAlreadyExistException: {}", e);
         } catch (org.apache.paimon.catalog.Catalog.ColumnNotExistException e) {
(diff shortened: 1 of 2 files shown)


DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff on the latest head (`d5337cca70`) and this looks like a focused fix that correctly restores the intended `ignoreIfNotExists` semantics in the Paimon catalog wrapper.

# What this PR solves
- User pain: callers can pass `ignoreIfNotExists = false` into `PaimonCatalog.alterTable(...)`, but the current wrapper always forwards `true` to the underlying Paimon catalog, so a missing table can be silently ignored instead of failing fast.
- Fix approach: pass the flag through as-is in both `alterTable` overloads, and translate the underlying Paimon `TableNotExistException` into SeaTunnel's own `TableNotExistException`.
- One-line summary: this PR makes the alter-table wrapper behave consistently with its own method contract.

# Simple example
If schema evolution tries to add a column to a Paimon table that has already been deleted, `ignoreIfNotExists = false` should surface an error immediately. Before this patch it could still be ignored; after this patch it fails fast as expected, while the `ignoreIfNotExists = true` path still keeps the best-effort behavior.

# Actual runtime path I re-checked
```text
schema change event
  -> AlterPaimonTableSchemaEventHandler.apply() [80-92]
      -> applySingleSchemaChangeEvent(...) [94-134]
          -> add/modify/change paths call paimonCatalog.alterTable(..., false)
          -> drop-column path calls paimonCatalog.alterTable(..., true)

catalog wrapper
  -> PaimonCatalog.alterTable(...) [358-387]
      -> old behavior: always delegated with `true`
      -> new behavior: delegates with the caller's actual ignore flag
      -> missing table:
           - ignore=false => throw TableNotExistException
           - ignore=true  => keep ignore semantics
```

# Findings
- The bug is real and it is on the normal schema-evolution path, not an unreachable edge branch.
- The fix is complete for both overloads (`SchemaChange` and `List<SchemaChange>`).
- The new regression test covers both forms and directly checks the intended semantic split between `false` and `true`.
- I did not find a blocking correctness issue in the current implementation.

# Test stability
The new test in `PaimonCatalogTest` looks stable to me:
- no sleeps
- no timing-based assertions
- no fixed-port or external-service dependency
- direct `assertThrows` / `assertDoesNotThrow` coverage for the exact bug

Stability rating for the new test change: **stable**.

# CI readout
I checked the failing Build job as well.
- `Dead links` fails on two external Airtable links in `docs/zh/connectors/source/Airtable.md` returning `406`.
- `unit-test (8, ubuntu-latest)` fails in `seatunnel-engine-client`, specifically `SeaTunnelEngineClusterRoleTest.testWorkerIsFirstMemberThenGetJobDetailStatus`, which times out waiting for `CANCELED` and still sees `CANCELING`.

I do not see evidence that either failure is caused by this PR:
- the PR diff only touches the two Paimon files
- the failing unit test is in `seatunnel-engine-client`, outside this diff
- there is no newer `upstream/dev` fix on that failing test path after this PR's merge-base

So from my side this looks like unrelated CI noise / flakiness rather than a blocker caused by the current Paimon change.

# Conclusion
### Conclusion: can merge

1. Blocking items
- None from my side on the current code.

2. Suggested follow-up
- Please rerun or recheck the current CI failures, but I would treat them as unrelated to this PR's source change unless new evidence shows otherwise.

Overall, this is a clean and appropriately scoped fix. The wrapper behavior now matches the caller contract, and the regression test covers the exact scenario that was previously wrong.

dailai (approved):
LGTM

PR #11022, opened 2026-06-08
[Improve][API] Allow exclusive/bundled options to coexist with optional value constraints

nzw921rx:

Allow `exclusive()`/`bundled()` options to coexist with `optional(Option, Condition)` value constraints in `OptionRule.Builder`.

**Problem:**

`OptionRule.Builder` treated `exclusive`/`bundled` and `optional` as mutually exclusive registrations. When a connector (e.g. MongoDB CDC) declared `exclusive(SCHEMA, TABLE_CONFIGS)`, adding `optional(SCHEMA, Conditions.mapNotEmpty(SCHEMA))` threw:
```
OptionsOption 'schema' duplicate in 'ExclusiveRequiredOptions'
```
This forced connectors to keep imperative `if/throw` checks for value validation that should be declarative.

**Fix:**

Modified `verifyOptionOptionsDuplicate` to skip duplicate checks for `ExclusiveRequiredOptions` and `BundledRequiredOptions`. These structural constraints (presence/exclusivity) are orthogonal to value constraints (content validation). The `required` + `optional` conflict check remains unchanged.

No.

Unit tests in `OptionRuleTest.testVerify()` and `ConfigValidatorTest`:
- `exclusive` + `optional(condition)` — builds and validates correctly
- `bundled` + `optional(condition)` — builds and validates correctly
- `required` + `optional` — still fails (unchanged behavior)
- Duplicate `optional` — still fails (unchanged behavior)
- Full runtime validation: exclusive check, value constraint check, combined scenarios

```bash
./mvnw -pl seatunnel-api test
```

### DIFF
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java
@@ -345,7 +345,7 @@ public class OptionRule {
                 @NonNull Option<?> option,
                 @NonNull Condition<?> condition1,
                 @NonNull Condition<?>... conditions) {
-            verifyOptionOptionsDuplicate(option, "OptionsOption");
+            verifyOptionOptionsDuplicate(option, "OptionsOptionWithConstraint");
             this.optionalOptions.add(option);
             this.valueConstraints.add(condition1);
             Collections.addAll(this.valueConstraints, conditions);
@@ -358,8 +358,8 @@ public class OptionRule {
                 @NonNull Option<?> option2,
                 @NonNull Condition<?> condition1,
                 @NonNull Condition<?>... conditions) {
-            verifyOptionOptionsDuplicate(option1, "OptionsOption");
-            verifyOptionOptionsDuplicate(option2, "OptionsOption");
+            verifyOptionOptionsDuplicate(option1, "OptionsOptionWithConstraint");
+            verifyOptionOptionsDuplicate(option2, "OptionsOptionWithConstraint");
             this.optionalOptions.add(option1);
             this.optionalOptions.add(option2);
             this.valueConstraints.add(condition1);
@@ -544,11 +544,15 @@ public class OptionRule {
                 @NonNull Option<?> option, @NonNull String currentOptionType) {
             verifyDuplicateWithOptionOptions(option, currentOptionType);
 
+            boolean hasOptionsOptionWithConstraint =
+                    "OptionsOptionWithConstraint".equals(currentOptionType);
             requiredOptions.forEach(
                     requiredOption -> {
-                        if (requiredOption instanceof RequiredOption.ExclusiveRequiredOptions
-                                || requiredOption
-                                        instanceof RequiredOption.BundledRequiredOptions) {
+                        if (hasOptionsOptionWithConstraint
+                                && (requiredOption
+                                                instanceof RequiredOption.ExclusiveRequiredOptions
+                                        || requiredOption
+                                                instanceof RequiredOption.BundledRequiredOptions)) {
                             return;
                         }
                         if (requiredOption.getOptions().contains(option)) {
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java
@@ -546,6 +546,11 @@ public class OptionRule {
 
             requiredOptions.forEach(
                     requiredOption -> {
+                        if (requiredOption instanceof RequiredOption.ExclusiveRequiredOptions
+                                || requiredOption
+                                        instanceof RequiredOption.BundledRequiredOptions) {
+                            return;
+                        }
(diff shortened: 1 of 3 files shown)


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current diff against `dev` and traced the real `OptionRule` call chain from connector `optionRule()` builders into `ConfigValidator`.

What this PR is trying to fix:
- User pain: connector factories sometimes need both an `exclusive(...)` / `bundled(...)` relation and a value constraint like `optional(option, Conditions.notEmpty(...))`, but the current builder rejects that as a duplicate definition.
- Approach: relax the duplicate check in `OptionRule.Builder` so exclusive/bundled groups can coexist with optional value constraints.
- In one line: the goal makes sense, but the current implementation widens the rule too much and turns `OptionRule` semantics into something that depends on builder call order.

Example:
- Intended valid case: a connector wants `table_names` and `table_pattern` to stay mutually exclusive, while also rejecting empty values when either one is configured.
- Current implementation also allows a plain `optional(table_names)` after `exclusive(table_names, table_pattern)`, which is broader than the PR title and changes the old duplicate contract.

Actual runtime / validation path I checked:
```text
connector factory init
  -> optionRule()
      -> OptionRule.builder()
          -> exclusive(...) / bundled(...)
          -> optional(...)
          -> build()

job submission config validation
  -> ConfigValidator.validate(config, rule)
      -> consumes requiredOptions / optionalOptions / valueConstraints built above
```

Key logic I traced:
- `optional(@NonNull Option<?>... options)` calls `verifyOptionOptionsDuplicate(...)` at `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:182-187`.
- `optional(@NonNull Option<?> option, @NonNull Condition<?> condition1, ...)` calls the same duplicate checker at `OptionRule.java:344-351`.
- The PR changes `verifyOptionOptionsDuplicate(...)` at `OptionRule.java:543-554` to skip all `ExclusiveRequiredOptions` and `BundledRequiredOptions`.

That leads to one blocker:

1. The duplicate-check relaxation is too broad and makes the result order-dependent.
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:182-187,344-351,543-554`
- Why this is a problem:
  - The intended fix is to allow `exclusive/bundled + optional(with condition)`.
  - The current code also allows plain `optional(option)` if `exclusive(...)` or `bundled(...)` was added first, because the duplicate check now returns early for those required-option types.
  - The reverse order still fails through the required-option duplicate path, so the same logical rule now behaves differently depending on declaration order.
- Risk:
  - This is a shared `seatunnel-api` validation-framework regression, not a connector-local issue.
  - Future connector factories can accidentally admit duplicate definitions that used to be rejected, and the behavior becomes much harder to reason about.
- Best fix:
  - Narrow the bypass to optional declarations that actually carry value constraints, or pass explicit context into the duplicate checker so plain optional declarations keep the old behavior.
- Severity: High

Test coverage notes:
- The new UTs cover `optional(with condition) + exclusive/bundled` and `optional(no condition) -> exclusive/bundled`.
- They do not cover the reverse order `exclusive(...).optional(option)` / `bundled(...).optional(option)`, which is exactly where the regression appears.
- Stability rating for the added tests: Stable. These are in-memory UTs and I did not see timing or environment-sensitive patterns.

Compatibility assessment:
- Not fully compatible at the API-contract level. This changes `OptionRule.Builder` behavior in `seatunnel-api`, and the new behavior is broader than the stated feature.

Merge conclusion:

Blocking items:
- Issue 1: narrow the duplicate-check relaxation so only the intended `optional-with-constraint` scenario is allowed.

Non-blocking suggestions:
- None beyond that blocker.

Overall, I agree with the direction, and I think the final shape can stay small. The key thing is to keep the old plain-optional contract intact while enabling the new value-constraint use case.

nzw921rx:
@davidzollo Could you please help review it? Any suggestions would be greatly appreciated.

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java
```diff
@@ -345,7 +345,7 @@ public Builder optional(
                 @NonNull Option<?> option,
                 @NonNull Condition<?> condition1,
                 @NonNull Condition<?>... conditions) {
-            verifyOptionOptionsDuplicate(option, "OptionsOption");
+            verifyOptionOptionsDuplicate(option, "OptionsOptionWithConstraint");
```
yzeng1618: It is recommended to define a private constant in the OptionRule class:private static final String TYPE_OPTION_WITH_CONSTRAINT = "OptionsOptionWithConstraint";

yzeng1618 (commented):
No major issues overall

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java
nzw921rx: fixed

yzeng1618 (approved):
LGTM +1  if ci pass

PR #11023, opened 2026-06-08
[Improve][Connector-V2] cdc migrate option ruleValidation Framework

nzw921rx:

Related #11007  

1. Remove redundant `checkNotNull` in CDC ConfigFactory — already covered by `optionRule()` declarative validation
2. Fix Oracle `checkNotNull(originUrl)` dead code that made `hostname+port` branch unreachable
3. Fix `database-names` from `optional` to `required` in Oracle/Postgres/SqlServer — runtime unconditionally uses `databaseList.get(0)`, not providing it causes NPE

Yes. Users of Oracle/Postgres/SqlServer CDC connectors who omit `database-names` will now receive a clear configuration validation error instead of a runtime NPE. This is a bug fix — there was no valid use case without `database-names` for these connectors.

## How was this patch tested?

Unit tests in each connector's FactoryTest:

- `OracleIncrementalSourceFactoryTest.testDatabaseNamesRequired` — verifies missing/empty rejected, valid accepted
- `PostgresIncrementalSourceFactoryTest.testDatabaseNamesRequired` — same
- `SqlServerIncrementalSourceFactoryTest.testDatabaseNamesRequired` — same

Existing tests (numeric constraints, table-names notEmpty) continue to pass.

```bash
./mvnw test -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql,seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle,seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres,seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver,seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb -Dtest="*FactoryTest" -DfailIfNoTests=false
```


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current diff against `dev`, not only the connector-side changes but also the shared `OptionRule` change that this migration depends on.

What this PR is trying to fix:
- User pain: several CDC connectors still keep value checks such as non-empty lists, positive numeric ranges, and regex validation in scattered factory/config-builder code.
- Approach: move those checks into `OptionRule + Conditions` for MongoDB/MySQL/Oracle/Postgres/SQLServer CDC factories, and add matching UTs.
- In one line: I agree with the migration direction, but the current head depends on a shared `seatunnel-api` change that is still too broad and introduces order-sensitive validation semantics.

Example:
- Intended outcome: a factory can express `exclusive(table_names, table_pattern)` and also require `table_names` to be non-empty when it is present.
- Current outcome: because the shared duplicate-check relaxation is too broad, the same API also starts allowing plain `optional(table_names)` after `exclusive(...)`, which is outside the stated scope.

Runtime / validation path I checked:
```text
CDC connector factory
  -> optionRule()
      -> exclusive(...) / bundled(...)
      -> optional(..., Conditions.*)
      -> build()

job submission
      -> consumes the shared OptionRule contract
```

Key path:
- Connector factories like `MySqlIncrementalSourceFactory` now rely on `exclusive(...)` plus `optional(..., Conditions.notEmpty(...))`.
- Both optional overloads still go through the shared duplicate checker in `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java`.
- The PR also changes that shared checker to skip all `ExclusiveRequiredOptions` and `BundledRequiredOptions`, not just the intended optional-with-condition case.

That leaves one blocker:

1. The CDC migration now inherits the same order-sensitive `OptionRule` regression from the shared API layer.
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:182-187,344-351,543-554`
- Why this is a problem:
  - These connector migrations are no longer using only local checks; they now trust the shared builder contract.
  - The current `OptionRule` change relaxes duplicate detection too broadly, so the migrated CDC factories inherit the same “builder order changes the result” behavior.
  - That makes this more than a local connector issue: it becomes a framework-level regression amplified across multiple CDC connectors.
- Risk:
  - As more factories migrate to `Conditions`, they will all inherit the same hidden semantic edge.
  - That makes config-rule behavior harder to reason about and harder to verify with connector-local tests alone.
- Best fix:
  - First narrow the shared `OptionRule` relaxation to the intended `optional-with-condition` scenario, then keep the connector migrations on top of that corrected base.
  - If you want to reduce risk further, another option is to split the shared API change and the connector migrations into separate PRs.
- Severity: High

Test coverage notes:
- It is good that the PR adds factory-level UTs for the migrated connectors.
- However, those tests mainly prove the target happy paths; they do not cover the shared reverse-order regression in the underlying `OptionRule` contract.
- Stability rating for the added tests: Stable. I did not see timing, port, or environment-sensitive patterns in the new UTs.

Compatibility assessment:
- Not fully compatible at the shared API-contract level, because this still changes `OptionRule.Builder` behavior more broadly than intended.

Merge conclusion:

Blocking items:
- Issue 1: fix the shared `OptionRule` duplicate-check boundary first, otherwise all migrated CDC factories inherit the framework regression.

Non-blocking suggestions:
- None beyond that blocker.

Overall, I do think moving these connector checks into `Conditions` is the right long-term direction. The important part is making sure that the shared validation layer is correct first, so the migration lands on a stable contract.

DanielLeens (changes requested):
Thanks for working on this. I re-reviewed the latest head from scratch again, and this time I specifically checked the current branch as it exists now, not the older revision that still experimented with the shared `OptionRule` change.

# What This PR Fixes
- User pain: several connectors want to express richer validation such as “these options are mutually exclusive, but the chosen one must also be non-empty or satisfy numeric constraints”.
- Fix approach: the PR migrates multiple connector factories to the new `exclusive(...)` + constrained `optional(...)` style and adds factory tests around those rules.
- One-line summary: the connector-side direction is reasonable, but the current head still depends on a shared `OptionRule` contract that is not actually part of this branch anymore.

# Runtime Chain Rechecked
```text
connector factory startup
  -> XxxIncrementalSourceFactory.optionRule()
      -> OptionRule.builder()
      -> exclusive(optionA, optionB)
      -> optional(optionA, Conditions...)
```

# Key Findings
1. The normal runtime path really does hit `optionRule()` during connector factory initialization.
2. The current PR head no longer carries the shared `OptionRule` change that would relax duplicate registration across `exclusive(...)` and constrained `optional(...)`.
3. Because of that, the connector migrations still rely on a contract that `upstream/dev` does not provide.

# Findings
Issue 1: the migrated factories still build rules that the current shared `OptionRule` contract rejects
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/MySqlIncrementalSourceFactory.java:63-69`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mongodb/MongodbIncrementalSourceFactory.java:65-73`, plus the same pattern in the Oracle/Postgres/SQL Server factory updates.
- Why this is a blocker:
  the branch now does `exclusive(TABLE_NAMES, TABLE_PATTERN)` and then `optional(TABLE_NAMES, Conditions.notEmpty(...))` (same for the other connectors), but the shared `OptionRule` on `dev` still treats that as a duplicate option registration. So the latest connector code is still depending on a builder behavior that is not present on the current base branch.
- Better fix:
  either bring back the shared `OptionRule` support change in the same PR, or express the connector-side validation in a way that works with the current `OptionRule` contract.
- Severity: High

Issue 2: the new tests instantiate the broken rule eagerly, so they do not protect the intended migration contract
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/MySqlIncrementalSourceFactoryTest.java:35-38`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb/src/test/java/mongodb/source/MongodbIncrementalSourceFactoryTest.java:41-45`
- Why this is a blocker:
  both tests initialize `private final OptionRule rule = new ...Factory().optionRule();`. If the shared builder contract still rejects the duplicate registration pattern above, the tests are not validating a working migration path; they are just tightly coupled to a branch-local assumption.
- Better fix:
  fix the shared contract mismatch first, then keep these tests as focused validation for the intended option-rule behavior.
- Severity: Medium

1. Blocking items
- Issue 1: the connector migration cannot land until the shared `OptionRule` contract it depends on is present on this branch as well.
- Issue 2: keep the factory tests, but only after the builder contract mismatch is resolved.

- None beyond the two items above.

Overall, I like the validation direction, but the current head is still split between connector-side usage and base-contract support, so I would not merge it as-is.

nzw921rx:
> Thanks for working on this. I re-reviewed the latest head from scratch again, and this time I specifically checked the current branch as it exists now, not the older revision that still experimented with the shared `OptionRule` change.
> 
> # What This PR Fixes
> * User pain: several connectors want to express richer validation such as “these options are mutually exclusive, but the chosen one must also be non-empty or satisfy numeric constraints”.
> * Fix approach: the PR migrates multiple connector factories to the new `exclusive(...)` + constrained `optional(...)` style and adds factory tests around those rules.
> * One-line summary: the connector-side direction is reasonable, but the current head still depends on a shared `OptionRule` contract that is not actually part of this branch anymore.
> 
> # Runtime Chain Rechecked
> ```
> connector factory startup
>   -> XxxIncrementalSourceFactory.optionRule()
>       -> OptionRule.builder()
>       -> exclusive(optionA, optionB)
>       -> optional(optionA, Conditions...)
>   -> ConfigValidator.validate(rule)
> ```
> 
> # Key Findings
> 1. The normal runtime path really does hit `optionRule()` during connector factory initialization.
> 2. The current PR head no longer carries the shared `OptionRule` change that would relax duplicate registration across `exclusive(...)` and constrained `optional(...)`.
> 3. Because of that, the connector migrations still rely on a contract that `upstream/dev` does not provide.
> 
> # Findings
> Issue 1: the migrated factories still build rules that the current shared `OptionRule` contract rejects
> 
> * Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/MySqlIncrementalSourceFactory.java:63-69`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mongodb/MongodbIncrementalSourceFactory.java:65-73`, plus the same pattern in the Oracle/Postgres/SQL Server factory updates.
> * Why this is a blocker:
>   the branch now does `exclusive(TABLE_NAMES, TABLE_PATTERN)` and then `optional(TABLE_NAMES, Conditions.notEmpty(...))` (same for the other connectors), but the shared `OptionRule` on `dev` still treats that as a duplicate option registration. So the latest connector code is still depending on a builder behavior that is not present on the current base branch.
> * Better fix:
>   either bring back the shared `OptionRule` support change in the same PR, or express the connector-side validation in a way that works with the current `OptionRule` contract.
> * Severity: High
> * Already raised by others: No
> 
> Issue 2: the new tests instantiate the broken rule eagerly, so they do not protect the intended migration contract
> 
> * Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/MySqlIncrementalSourceFactoryTest.java:35-38`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb/src/test/java/mongodb/source/MongodbIncrementalSourceFactoryTest.java:41-45`
> * Why this is a blocker:
>   both tests initialize `private final OptionRule rule = new ...Factory().optionRule();`. If the shared builder contract still rejects the duplicate registration pattern above, the tests are not validating a working migration path; they are just tightly coupled to a branch-local assumption.
> * Better fix:
>   fix the shared contract mismatch first, then keep these tests as focused validation for the intended option-rule behavior.
> * Severity: Medium
> * Already raised by others: No
> 
> ### Conclusion: can merge after fixes
> 1. Blocking items
> 
> * Issue 1: the connector migration cannot land until the shared `OptionRule` contract it depends on is present on this branch as well.
> * Issue 2: keep the factory tests, but only after the builder contract mismatch is resolved.
> 
> 2. Suggested non-blocking follow-up
> 
> * None beyond the two items above.
> 
> Overall, I like the validation direction, but the current head is still split between connector-side usage and base-contract support, so I would not merge it as-is.

@DanielLeens Thank you for your review. The dev branch already supports it. Please use the latest code.

DanielLeens:
Thanks for calling this out, and sorry for the confusion on my side. I rechecked this specific point against the current `dev` base rather than my older local checkout, and you are right: `dev` already contains `Conditions` plus the newer `OptionRule` overloads that allow constrained `optional(...)` entries alongside `exclusive(...)`. So this particular blocker from my last review does not stand on the current base.

Since there is still no new commit after Daniel's latest full review on this PR, I am not starting another full re-review in this round. Please treat this comment only as a correction to that earlier point. If you push another code update, I will re-review the latest head from scratch.

DanielLeens (changes requested):
Sorry for missing this in the last review — that's on me. I re-reviewed the latest head from the PR commit list down to the real connector startup path, and I found two blockers that mean this still should not merge as-is.

# What This PR Solves
- User pain: several CDC connectors still keep non-empty checks, numeric bounds, and table-name validation as scattered imperative checks.
- Fix approach: move those validations into `OptionRule + Conditions` across MongoDB, MySQL, Oracle, Postgres, SQL Server, and TiDB, with matching factory-level tests.
- One-line summary: the migration direction is fine, but the current PR is still mixing unrelated review scope and has not actually closed a critical Oracle validation prerequisite.

# Runtime Chain Rechecked
```text
Submit-time validation
  -> connector factory.optionRule()
          -> exclusive / bundled / required / optional + Conditions

Oracle runtime config assembly
  -> OracleIncrementalSource.createSourceConfigFactory(config) [OracleIncrementalSource.java:84-93]
      -> configFactory.fromReadonlyConfig(readonlyConfig)
      -> configFactory.originUrl(config.get(URL))
  -> later source startup / schema discovery
      -> OracleSourceConfigFactory.create(subtask) [OracleSourceConfigFactory.java:78-114]
          -> if originUrl != null
               setProperty("database.url", originUrl)
             else
               setProperty("database.hostname", hostname)
               setProperty("database.port", String.valueOf(port))
```

# Findings
Issue 1: this PR still contains the full `#11048` EXTENSION API/docs stack, so it is no longer a focused CDC-only change
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionExtension.java:1`, `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java:420-434`, `docs/en/engines/zeta/rest-api-v2.md:143-177`
- Why this is a blocker:
  I rechecked the current PR with `gh pr view 11023 --json commits,files`. The latest branch still contains 7 commits, and the first 5 of them (`a0b9a49777` through `54159a3571`) are the same EXTENSION API / CLI / REST / docs commits that belong to `#11048`. So merging `#11023` right now would not just land CDC validation migration. It would also merge another still-open PR's API/docs changes through the side door.
- Risk:
  this breaks the one-problem-per-PR boundary, makes the review scope much harder to reason about, and complicates revert/blame history later.
- Better fix:
  rebase or rebuild this branch so it contains only the CDC validation migration commits. If `#11048` merges first, then rebase this PR onto the updated `dev` and make sure `Files changed` shrinks back to CDC-only files.
- Severity: High

Issue 2: the Oracle connection-endpoint prerequisite still is not actually enforced in `optionRule()`, so invalid configs still escape to runtime
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactory.java:92-100`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/config/OracleSourceConfigFactory.java:109-113`
- Why this is a blocker:
  the Oracle factory now declares `HOSTNAME` and `PORT` as bundled and `URL` as optional, but it still does not express the real contract: the user must provide either `url` or `hostname + port`. That means a config with `username/password/database-names/table-names` but neither endpoint form still passes submit-time validation and only fails later in `OracleSourceConfigFactory.create()` when runtime config assembly tries to use a null endpoint.
- Risk:
  this is exactly the kind of main-path validation gap this PR is supposed to remove. The user still gets a runtime startup failure instead of a clear submit-time validation error.
- Better fix:
  express the endpoint requirement explicitly at the validation layer. If the current `OptionRule` combinations are not enough to model it cleanly, keep a small imperative fallback, but do not leave this prerequisite unchecked.
- Severity: High

Issue 3: the new Oracle tests still miss the most important negative endpoint case
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactoryTest.java:41-125`
- Why this matters:
  the new tests cover valid host/port input, numeric constraints, table-name validation, and the schema-change/log-mining rule, but they do not cover the invalid “no url and no hostname/port” path that maps directly to Issue 2.
- Better fix:
  add one focused negative test that keeps the other required fields present, removes both endpoint forms, and asserts that validation fails before runtime.
- Severity: Medium

# Test Coverage
- Good factory-level UT coverage was added across the migrated connectors.
- The remaining gap is the Oracle endpoint negative path above.
- Stability rating: Stable. The added tests are unit-level and I did not see timing, port, or environment-sensitive patterns.

# Merge Conclusion

1. Blocking items
- Issue 1: strip the unrelated `#11048` commits/files back out of this PR so the review scope is clean again.
- Issue 2: finish the Oracle endpoint validation migration so invalid configs fail at submit time rather than later at runtime.

2. Suggested follow-up
- Issue 3: add the missing Oracle negative-path test once Issue 2 is fixed.

Overall, I agree with the direction of moving these connector checks into `OptionRule`. But the branch still needs a scope cleanup plus one more real Oracle validation fix before I would be comfortable merging it.

DanielLeens (changes requested):
Sorry for missing this in the last review — that's on me. I re-reviewed the latest head from the PR commit list down to the real connector startup path, and I found two blockers that mean this still should not merge as-is.

# What This PR Solves
- User pain: several CDC connectors still keep non-empty checks, numeric bounds, and table-name validation as scattered imperative checks.
- Fix approach: move those validations into `OptionRule + Conditions` across MongoDB, MySQL, Oracle, Postgres, SQL Server, and TiDB, with matching factory-level tests.
- One-line summary: the migration direction is fine, but the current PR is still mixing unrelated review scope and has not actually closed a critical Oracle validation prerequisite.

# Runtime Chain Rechecked
```text
Submit-time validation
  -> connector factory.optionRule()
          -> exclusive / bundled / required / optional + Conditions

Oracle runtime config assembly
  -> OracleIncrementalSource.createSourceConfigFactory(config) [OracleIncrementalSource.java:84-93]
      -> configFactory.fromReadonlyConfig(readonlyConfig)
      -> configFactory.originUrl(config.get(URL))
  -> later source startup / schema discovery
      -> OracleSourceConfigFactory.create(subtask) [OracleSourceConfigFactory.java:78-114]
          -> if originUrl != null
               setProperty("database.url", originUrl)
             else
               setProperty("database.hostname", hostname)
               setProperty("database.port", String.valueOf(port))
```

# Findings
Issue 1: this PR still contains the full `#11048` EXTENSION API/docs stack, so it is no longer a focused CDC-only change
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionExtension.java:1`, `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java:420-434`, `docs/en/engines/zeta/rest-api-v2.md:143-177`
- Why this is a blocker:
  I rechecked the current PR with `gh pr view 11023 --json commits,files`. The latest branch still contains 8 commits, and the first 5 of them (`a0b9a49777` through `54159a3571`) are the same EXTENSION API / CLI / REST / docs commits that belong to `#11048`. So merging `#11023` right now would not just land CDC validation migration. It would also merge another still-open PR's API/docs changes through the side door.
- Risk:
  this breaks the one-problem-per-PR boundary, makes the review scope much harder to reason about, and complicates revert/blame history later.
- Better fix:
  rebase or rebuild this branch so it contains only the CDC validation migration commits. If `#11048` merges first, then rebase this PR onto the updated `dev` and make sure `Files changed` shrinks back to CDC-only files.
- Severity: High

Issue 2: the Oracle connection-endpoint prerequisite still is not actually enforced in `optionRule()`, so invalid configs still escape to runtime
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactory.java:92-100`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/config/OracleSourceConfigFactory.java:109-113`
- Why this is a blocker:
  the Oracle factory now declares `HOSTNAME` and `PORT` as bundled and `URL` as optional, but it still does not express the real contract: the user must provide either `url` or `hostname + port`. That means a config with `username/password/database-names/table-names` but neither endpoint form still passes submit-time validation and only fails later in `OracleSourceConfigFactory.create()` when runtime config assembly tries to use a null endpoint.
- Risk:
  this is exactly the kind of main-path validation gap this PR is supposed to remove. The user still gets a runtime startup failure instead of a clear submit-time validation error.
- Better fix:
  express the endpoint requirement explicitly at the validation layer. If the current `OptionRule` combinations are not enough to model it cleanly, keep a small imperative fallback, but do not leave this prerequisite unchecked.
- Severity: High

Issue 3: the new Oracle tests still miss the most important negative endpoint case
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactoryTest.java:41-125`
- Why this matters:
  the new tests cover valid host/port input, numeric constraints, table-name validation, and the schema-change/log-mining rule, but they do not cover the invalid “no url and no hostname/port” path that maps directly to Issue 2.
- Better fix:
  add one focused negative test that keeps the other required fields present, removes both endpoint forms, and asserts that validation fails before runtime.
- Severity: Medium

# Test Coverage
- Good factory-level UT coverage was added across the migrated connectors.
- The remaining gap is the Oracle endpoint negative path above.
- Stability rating: Stable. The added tests are unit-level and I did not see timing, port, or environment-sensitive patterns.

# Merge Conclusion

1. Blocking items
- Issue 1: strip the unrelated `#11048` commits/files back out of this PR so the review scope is clean again.
- Issue 2: finish the Oracle endpoint validation migration so invalid configs fail at submit time rather than later at runtime.

2. Suggested follow-up
- Issue 3: add the missing Oracle negative-path test once Issue 2 is fixed.
- Separately, the branch is currently `CONFLICTING` and the GitHub `Build` is red, so please rebase/sync `dev` and rerun the gate after the source fixes above.

Overall, I agree with the direction of moving these connector checks into `OptionRule`. But the branch still needs a scope cleanup plus one more real Oracle validation fix before I would be comfortable merging it.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the real connector validation path again.

What I rechecked this round:
```text
CDC connector factory.optionRule()
  -> submit-time ConfigValidator.validate(...)
Oracle endpoint validation
  -> optionRule() must reject missing endpoint before runtime
current branch scope
  -> commit/file surface against dev
```

The Oracle endpoint validation gap I called out in the previous round is moving in the right direction on the latest head. But one merge blocker still remains on the current branch:

Issue 1: this PR is still carrying unrelated API/docs history instead of being a focused CDC validation migration
- Location: current branch history still includes the `#11048` EXTENSION stack and unrelated non-CDC changes; representative paths already visible from the branch surface include `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionExtension.java`, `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java`, and REST/docs files outside the CDC connector migration.
- Why this is a blocker:
  the advertised purpose of this PR is migrating CDC connector validation. But the current branch still drags in another feature line plus unrelated history, which makes the review scope larger than the problem statement and breaks the repo rule of one focused problem per PR.
- Risk:
  this makes the branch much harder to review safely, complicates blame/revert later, and can merge another still-separate feature through the side door.
- Better fix:
  rebuild or rebase the branch so that `Files changed` only contains the CDC validation migration and its directly related tests/docs.
- Severity: High
- Already raised by others: Partially related to my earlier scope concern; still true on the current head.

1. Blocking items
- Issue 1: strip the unrelated API/docs history back out so this PR is reviewable as a CDC validation migration again.

2. Suggested follow-up
- Once the branch surface is clean, the remaining connector-level validation changes will be much easier to verify in one pass.

seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mongodb/MongodbIncrementalSourceFactory.java
```diff
@@ -114,42 +164,21 @@ private List<CatalogTable> updateAndValidateCatalogTableId(
                                     catalogTable.getCatalogName(), TablePath.of(collectionName));
                     return Collections.singletonList(
                             CatalogTable.of(updatedIdentifier, catalogTable));
-                } else if (!fullName.equals(collectionName)) {
```
CosmosNi, line 117: Why were these validation logic removed? This results in failures to detect and handle inconsistencies between schema table names and their associated collection names.

seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/config/OracleSourceConfigFactory.java
```diff
@@ -129,29 +124,16 @@ public OracleSourceConfig create(int subtask) {
...
-                                        throw new IllegalArgumentException(
-                                                "Invalid table name: " + tableStr);
+                                        return tableStr;
                                     })
                             .collect(Collectors.joining(",")));
         }
 
         // override the user-defined debezium properties
         if (dbzProperties != null) {
-            String debeziumSchemaChanges =
```
CosmosNi, line 146: The validator should also check Debezium's include.schema.changes — not just SeaTunnel's flag — since users can override it.

seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mongodb/MongodbIncrementalSourceFactory.java
DanielLeens: Thanks for pointing this out. I agree this is a real gap on the current head, not just a style preference.

The old path rejected two cases that the new validation no longer covers:
1. a multi-table path where `catalogTables.size() == collections.size()` but the schema/table full name still does not match the corresponding collection name, and
2. the explicit mismatch failure on the non-single-table path.

After this change, `updateAndValidateCatalogTableId(...)` only rewrites the one-table case and otherwise falls through, so a mismatched schema/collection mapping can now slip past submit-time validation. Since there is still no new commit after Daniel's last full review, I'm keeping this as a thread-level follow-up and would want this restored or otherwise revalidated in the next code update.

seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/config/OracleSourceConfigFactory.java
DanielLeens: Agreed. The current validator only runs when SeaTunnel's `SCHEMA_CHANGES_ENABLED` option is `true`, but `OracleSourceConfigFactory.create()` later does `props.putAll(dbzProperties)`, so a user can still force Debezium `include.schema.changes=true` through `debezium_properties` while bypassing the submit-time check here.

That means the runtime contract is still wider than the validation contract on the current head. Since there is still no new commit after Daniel's last full review, I'm treating this as another thread-level blocker to address in the next code update rather than reopening a fresh full review record on the same revision.

DanielLeens (changes requested):
Sorry for missing this in the last review — that's on me. I re-reviewed the latest head from scratch after the new commit, and I also rechecked @CosmosNi's two earlier inline concerns. The MongoDB collection-name mismatch path and the Oracle `debezium.include.schema.changes=true` override gap both look addressed on the current head, so I am not repeating them here. But I still found a few blockers on the current revision.

- User pain: several CDC connectors still keep non-empty checks, numeric bounds, and format validation in scattered imperative `if/throw` logic, so invalid configs are often rejected late during source initialization instead of at submission time.
- Approach: migrate those checks into `OptionRule + Conditions`, add shared schema-change include/exclude options, and wire SQLServer schema evolution through to Debezium.
- In one line: the direction is good, but the current head still leaves a few submit-time and runtime contracts out of sync.

Example:
- Intended behavior: connector validation should reject a bad config before the job starts.
- Current behavior on some paths: the config now passes `ConfigValidator`, but the failure moves later into Debezium/source initialization.

Runtime / validation path I checked:
```text
job submission
  -> XxxIncrementalSourceFactory.optionRule()

MySQL table-name path
  -> MySqlIncrementalSourceFactory.optionRule() [56-85]
      -> table-names uses the shared QualifiedTableNameValidator
  -> MySqlSourceConfigFactory.create() [100-107]
      -> props["table.include.list"] = String.join(",", tableList)
      -> Debezium receives the original table-names unchanged

SQLServer schema-change path
  -> SqlServerIncrementalSourceFactory.optionRule() [70-80]
      -> schema-changes.enabled/include/exclude are exposed
      -> but there is no submit-time validateOptions() hook
  -> SqlServerSourceConfigFactory.create() [60-61]
      -> props["include.schema.changes"] = schemaChangeEnabled
  -> SqlServerIncrementalSource.createDebeziumDeserializationSchema() [116-124]
      -> SchemaChangeEventFilter.fromConfig(config)

Oracle / SQLServer stop-mode path
  -> optionRule() [Oracle 160-170, SQLServer 101-111]
      -> no longer requires stop.timestamp / specific offset fields
  -> StopConfig.getStopOffset() [40-49]
      -> TIMESTAMP => offsetFactory.timestamp(timestamp)
      -> SPECIFIC => offsetFactory.specific(...)
```

Findings:

Issue 1: the shared table-name validator is too permissive for MySQL
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/MySqlIncrementalSourceFactory.java:66-72`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/config/MySqlSourceConfigFactory.java:105-107`, `docs/en/connectors/source/MySQL-CDC.md:193`
- Why this is a problem:
  this path sits on the normal MySQL submit-time validation flow. The new `QualifiedTableNameValidator` accepts both `schema.table` and `db.schema.table`, but MySQL still forwards `tableList` unchanged into Debezium via `table.include.list`. The MySQL docs also still define `table-names` as `database_name.table_name`. So a three-segment name now passes submit-time validation even though it is not a valid MySQL table identifier contract.
- Risk:
  an invalid MySQL config now slips through the exact path this PR is trying to harden, and the user only finds out later when the connector/runtime consumes it.
- Best fix:
  either keep a MySQL-specific validator that only accepts the two-segment `database.table` form, or layer a MySQL-specific `ConditionExtension` on top of the shared validator. Please also add a negative UT that asserts `db.schema.table` is rejected for MySQL.
- Severity: High

Issue 2: SQLServer now exposes `schema-changes.include/exclude`, but invalid values still do not fail fast at submission time
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/source/SqlServerIncrementalSourceFactory.java:70-80`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/source/SqlServerIncrementalSource.java:116-124`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/schema/SchemaChangeEventFilter.java:75-89`
- Why this is a problem:
  the PR does more than just declare the options. `SqlServerSourceConfigFactory.create()` now actually forwards `schema-changes.enabled` into Debezium, so this is a real user-facing behavior change. But the SQLServer factory never calls `SchemaChangeEventFilter.validateOptions()` (or an equivalent condition), which means a typo like `rename.tabble` can pass `ConfigValidator` and only fail later when `SqlServerIncrementalSource.createDebeziumDeserializationSchema()` builds the filter.
- Risk:
  users get a late source-initialization failure instead of a clear submit-time validation error, and SQLServer ends up with a weaker fail-fast contract than MySQL/Oracle for the same option family.
- Best fix:
  either move SQLServer onto the same submit-time validation hook as the other schema-evolution connectors, or add an explicit SQLServer-side validation condition for `schema-changes.include/exclude`. Please also add a negative UT for an invalid canonical event name.
- Severity: High

Issue 3: Oracle and SQLServer both dropped the dependent-option validation for stop-mode, so invalid configs now fail later at runtime
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactory.java:160-170`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/source/SqlServerIncrementalSourceFactory.java:101-112`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/config/StopConfig.java:40-49`
- Why this is a problem:
  both factories used to require the dependent fields for `stop.mode = timestamp` (and for the `specific` paths). The current head removes those conditionals, but the runtime path still unconditionally consumes the values through `StopConfig.getStopOffset()`. So a config with `stop.mode = timestamp` and no `stop.timestamp` can now pass submission and only fail later when the source resolves the stop offset.
- Risk:
  this weakens a real main-path safety guarantee. The failure becomes later, harder to diagnose, and easier to miss during review because the submit-time rule now looks simpler than the runtime contract it feeds.
- Best fix:
  restore the dependent-option validation for the supported stop modes. If Oracle/SQLServer do not really support `specific`, then please reject that mode explicitly at validation time instead of silently leaving it to a later runtime exception.
- Severity: High

Issue 4: the user-facing schema-change contract is not documented in `docs/en` and `docs/zh`
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/option/SourceOptions.java:123-149`, `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/config/SqlServerSourceConfigFactory.java:60-61`, `docs/en/connectors/source/MySQL-CDC.md:216`, `docs/zh/connectors/source/MySQL-CDC.md:215`, `docs/en/connectors/source/Oracle-CDC.md:254`
- Why this is a problem:
  the PR adds public `schema-changes.include/exclude` options and makes SQLServer `schema-changes.enabled` actually effective, but the current docs still only describe `schema-changes.enabled` for MySQL/Oracle and do not explain the new include/exclude precedence or the SQLServer support change in either language.
- Risk:
  code and docs drift immediately, and users have no reliable way to discover which connectors support these schema-change controls or how `include` vs `exclude` is supposed to behave.
- Best fix:
  update the MySQL / Oracle / SQLServer connector docs in both `docs/en` and `docs/zh`, and also update the schema-evolution guide to document `schema-changes.include`, `schema-changes.exclude`, and the precedence rule (`exclude` wins).
- Severity: High

Test coverage:
- The new UTs are a good start and they are structurally stable. I did not see any timing, port, clock, or environment-sensitive patterns.
- But the coverage still misses the important regression cases above:
  - MySQL does not assert that a three-segment table name must fail.
  - SQLServer does not cover invalid `schema-changes.include/exclude` values or the missing `stop.timestamp` path.
  - Oracle does not cover the missing stop-mode dependent options either.
- Stability rating for the added tests: Stable.

Compatibility:
- This is not a storage/protocol break, but it is not fully safe as-is at the config-contract level:
  - SQLServer behavior is now broader than before.
  - MySQL table-name validation is weaker than the documented contract.
  - Oracle/SQLServer stop-mode validation is weaker than the runtime path still requires.

1. Blocking items
- Issue 1: tighten the MySQL table-name validation back to the real `database.table` contract.
- Issue 2: make SQLServer `schema-changes.include/exclude` fail fast at submission time, not only during source initialization.
- Issue 3: restore or explicitly replace the missing stop-mode dependent validation in Oracle/SQLServer.
- Issue 4: document the new schema-change options and the SQLServer behavior change in both English and Chinese docs.

- None beyond the blockers above. The remaining issues are all on the main validation/user-contract path.

Overall, I agree with the direction of moving CDC validation into `OptionRule + Conditions`. But on the current head, a few connector-specific runtime contracts are still not fully aligned with the new submit-time rules, so I would not merge this revision yet.

DanielLeens (changes requested):
Thanks for the update! I re-reviewed the latest head from the factory option rules down to the CDC stop-offset runtime path. The main direction is better now, but one blocker still remains.

What this PR fixes
- User pain: The CDC validation framework is moving toward declarative option rules, but Oracle / SQL Server stop-mode configs can still reach runtime with missing required fields.
- Fix approach: The PR adds declarative schema-change-name validation and removes a lot of imperative validation code across CDC connectors.
- One-line summary: The migration is moving in the right direction, but `stop.mode` still has an uncovered required-field gap.

Runtime path I traced
```text
Job submission
  -> OracleIncrementalSourceFactory.optionRule() / SqlServerIncrementalSourceFactory.optionRule()
      -> startup-mode dependencies are declared
      -> stop-mode dependencies are still not declared

Runtime path
  -> StopConfig.getStopOffset()
      -> TIMESTAMP uses `timestamp` directly
      -> SPECIFIC uses `file` / `pos` directly

Result
  -> an incomplete stop-mode config can still pass submission and fail later on the CDC offset path
```

Key findings
- The updated schema-change-name validation looks correct.
- The remaining gap is on the normal Oracle / SQL Server stop-mode path, not on an edge-only branch.
- Runtime still consumes stop-mode fields directly, so submission-time validation needs to enforce them.

Blocking items
1. Oracle / SQL Server stop-mode still lacks conditional required validation
   - Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactory.java:167-178; seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/source/SqlServerIncrementalSourceFactory.java:109-120; seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/config/StopConfig.java:40-49`
   - Why this is a problem: Both factories declare `STOP_MODE`, but they still do not require the dependent timestamp / specific-offset fields, while runtime offset creation consumes them directly.
   - Risk: An incomplete stop-mode config can pass submission and then fail later on the CDC runtime path.
   - Suggested fix: Add stop-mode conditional rules symmetric to the startup-mode rules, and cover both timestamp and specific-offset stop cases in tests.

Other review notes

- Compatibility: Backward compatible at the config-name level, but the submission-time contract is still incomplete for stop-mode configs.
- Side effects: The main impact is delayed failure: runtime still has to discover what should have been rejected during submission.
- Error handling / logging: This should be rejected as an aggregated validation error during submission rather than surfacing later in runtime stop-offset handling.
- Tests: I did not find coverage for Oracle / SQL Server stop-mode missing-field cases on the current head.
- Docs: No extra docs change is needed, but the declarative contract still needs to match runtime behavior.
- CI: Some checks were still running when I reviewed. The blocker here comes from the source path itself.

Merge verdict
- The overall migration direction makes sense, but the stop-mode validation gap on Oracle / SQL Server is still a blocker on the current head.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from the connector factory option rules down to the CDC stop-offset runtime path again.

- The latest revision cleans up a lot of non-PR artifacts and continues moving connector validation into declarative `optionRule()` definitions.
- I rechecked the exact Oracle / SQL Server stop-mode gap Daniel blocked previously.

Runtime path I checked
```text
Job submission
  -> OracleIncrementalSourceFactory.optionRule()
  -> SqlServerIncrementalSourceFactory.optionRule()
      -> startup-mode conditional fields are declared
      -> stop-mode dependent fields are still not declared

Runtime path
  -> StopConfig.getStopOffset()
      -> TIMESTAMP mode consumes stop.timestamp directly
      -> SPECIFIC mode consumes stop.specific-offset.file / pos directly
```

Findings
- The schema-change-name validation work remains on the right track.
- But the Oracle / SQL Server stop-mode validation gap is still present on the normal runtime path of the current head.
- I did not find new tests covering missing required stop-mode fields for those connectors.

1. Blocking items
- `OracleIncrementalSourceFactory.optionRule()` and `SqlServerIncrementalSourceFactory.optionRule()` still declare `STOP_MODE` without the conditional required fields for `stop.timestamp` and the specific-offset pair. On the current head, incomplete stop-mode configs can still pass submission and fail later in `StopConfig.getStopOffset()`. Please add stop-mode conditional rules symmetric to the startup-mode rules, and cover both timestamp and specific-offset cases in tests.

- None until the stop-mode contract matches the runtime path.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head `8378d2d01b` from the declarative option-rule path down to the actual Oracle / SQL Server startup and stop-offset runtime path again.

Sorry for missing one carryover sub-case in the last review. I had already called out the `stop.mode` validation gap before, and on this full re-review I also confirmed that the same missing conditional coverage still affects `STARTUP_MODE=SPECIFIC` on the current head.

# What this PR solves
- User pain: CDC connector validation is currently scattered across imperative `if/throw` checks and is hard to maintain consistently.
- Fix approach: migrate connector validation to declarative `OptionRule` / `Conditions` / `ConditionExtension`.
- One-line summary: the migration direction is good, but the Oracle / SQL Server offset-mode contract is still not aligned with the real runtime consumption path.

# Runtime path I traced
```text
Job submission
  -> OracleIncrementalSourceFactory.optionRule()
  -> SqlServerIncrementalSourceFactory.optionRule()
      -> current head no longer requires several mode-dependent offset fields

Runtime
  -> IncrementalSource.getStartupConfig()
      -> reads STARTUP_SPECIFIC_OFFSET_* directly
  -> IncrementalSource.getStopConfig()
      -> reads STOP_* directly
  -> StopConfig.getStopOffset()
      -> SPECIFIC / TIMESTAMP branches consume those values immediately
```

# Findings
1. Oracle / SQL Server startup/stop offset validation is still incomplete
   - Location:
     - `.../OracleIncrementalSourceFactory.java:137`
     - `.../SqlServerIncrementalSourceFactory.java:101`
     - `.../IncrementalSource.java:131`
     - `.../StopConfig.java:40`
   - Why this is a problem:
     the current `optionRule()` no longer enforces the dependent fields for `STARTUP_MODE=SPECIFIC` and still does not fully enforce them for `STOP_MODE=SPECIFIC/TIMESTAMP`, while the runtime path still consumes those values directly.
   - Risk:
     an incomplete config can pass submission and then fail later on the real startup / stop-offset path.
   - Best fix:
     restore the missing `.conditional(...)` rules so the submission-time contract matches the runtime contract again. At minimum, cover:
     - `STARTUP_MODE=SPECIFIC -> STARTUP_SPECIFIC_OFFSET_POS`
     - `STOP_MODE=SPECIFIC -> STOP_SPECIFIC_OFFSET_POS`
     - `STOP_MODE=TIMESTAMP -> STOP_TIMESTAMP`
   - Severity: High

2. The current `Build` is still red because of a Spotless failure
   - Location:
     - GitHub Actions `Build` run `28173441345`
     - `SourceOptions.java:28`
   - Why this is a problem:
     `spotless:check` is currently failing on a duplicate `import java.util.List;` in `SourceOptions.java`, so the PR cannot pass basic CI in its current form.
   - Best fix:
     run `mvn spotless:apply` or remove the duplicate import directly, then rerun the build.
   - Severity: Medium

# Tests / stability
- The new validator tests are pure in-memory `ConfigValidator` tests with no sleeps, ports, or container dependencies.
- Stability rating: Stable.
- The remaining problem is coverage, not flakiness: I still do not see negative tests for the missing `STARTUP_MODE=SPECIFIC` and `STOP_MODE=SPECIFIC/TIMESTAMP` cases above.

# Merge conclusion

1. Blocking items
- Fix the Oracle / SQL Server startup/stop offset validation gap described in item 1.
- Fix the current red `Build` caused by the Spotless failure in item 2.

- None beyond the blockers above.

The declarative-validation direction still makes sense, but the current head is not closed yet because the submission-time contract still diverges from the real runtime path.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from the declarative CDC option-rule path down to the actual startup/stop offset consumption path again.

## What this PR solves
- User pain: several CDC connectors still reject invalid configs late, from scattered runtime `if/throw` checks instead of at submit time.
- Fix approach: move those checks into declarative `OptionRule + Conditions + ConditionExtension` across the CDC factories.
- One-line summary: the migration direction is still right, but the current head still leaves the Oracle / SQL Server offset-mode contract out of sync with the real runtime path.

## Runtime path I checked
```text
Job submission
  -> OracleIncrementalSourceFactory.optionRule()
  -> SqlServerIncrementalSourceFactory.optionRule()
      -> startup-mode / stop-mode dependent fields should be validated here

Runtime consumption path
  -> startup config / stop config resolution on the CDC source side
      -> SPECIFIC / TIMESTAMP modes consume their companion offset/timestamp fields directly
```

## Findings

### Issue 1: the latest head still does not declare the mode-dependent offset fields required by the real startup/stop runtime path
- Location:
  - `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactory.java`
  - `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/source/SqlServerIncrementalSourceFactory.java`
- Why this is a blocker:
  On the current head, the factories only declare conditional validation for `STARTUP_MODE=TIMESTAMP` and `STARTUP_MODE=INITIAL`. The same option rules still do not cover the companion fields for the other offset-driven modes on the real runtime path, including the `SPECIFIC` startup case and the stop-mode paths that consume explicit timestamp/file/position values.
- Risk:
  An incomplete offset-mode config can still pass submit-time validation and only fail later when the CDC source actually tries to build the startup/stop offset.
- Best fix:
  Add the missing mode-specific conditional requirements to the Oracle and SQL Server factories, and add focused negative tests that prove those invalid offset-mode shapes are rejected during `ConfigValidator.validate(...)`.
- Severity: High

## Current conclusion
I still see a source-level blocker on the current head. The PR is closer, but the declarative validation contract is not yet fully aligned with the real runtime offset path.

1. Blocking items
- Issue 1: close the remaining startup/stop offset-mode validation gap in the Oracle / SQL Server factories.

- None from this round.

nzw921rx:
> Thanks for the update. I re-reviewed the latest head from the declarative CDC option-rule path down to the actual startup/stop offset consumption path again.
> 
> ## What this PR solves
> * User pain: several CDC connectors still reject invalid configs late, from scattered runtime `if/throw` checks instead of at submit time.
> * Fix approach: move those checks into declarative `OptionRule + Conditions + ConditionExtension` across the CDC factories.
> * One-line summary: the migration direction is still right, but the current head still leaves the Oracle / SQL Server offset-mode contract out of sync with the real runtime path.
> 
> ## Runtime path I checked
> ```
> Job submission
>   -> OracleIncrementalSourceFactory.optionRule()
>   -> SqlServerIncrementalSourceFactory.optionRule()
>       -> startup-mode / stop-mode dependent fields should be validated here
> 
> Runtime consumption path
>   -> startup config / stop config resolution on the CDC source side
>       -> SPECIFIC / TIMESTAMP modes consume their companion offset/timestamp fields directly
> ```
> 
> ## Findings
> ### Issue 1: the latest head still does not declare the mode-dependent offset fields required by the real startup/stop runtime path
> * Location:
>   
>   * `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactory.java`
>   * `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/source/SqlServerIncrementalSourceFactory.java`
> * Why this is a blocker:
>   On the current head, the factories only declare conditional validation for `STARTUP_MODE=TIMESTAMP` and `STARTUP_MODE=INITIAL`. The same option rules still do not cover the companion fields for the other offset-driven modes on the real runtime path, including the `SPECIFIC` startup case and the stop-mode paths that consume explicit timestamp/file/position values.
> * Risk:
>   An incomplete offset-mode config can still pass submit-time validation and only fail later when the CDC source actually tries to build the startup/stop offset.
> * Best fix:
>   Add the missing mode-specific conditional requirements to the Oracle and SQL Server factories, and add focused negative tests that prove those invalid offset-mode shapes are rejected during `ConfigValidator.validate(...)`.
> * Severity: High
> 
> ## Current conclusion
> I still see a source-level blocker on the current head. The PR is closer, but the declarative validation contract is not yet fully aligned with the real runtime offset path.
> 
> ### Conclusion: can merge after fixes
> 1. Blocking items
> 
> * Issue 1: close the remaining startup/stop offset-mode validation gap in the Oracle / SQL Server factories.
> 
> 2. Suggested non-blocking follow-up
> 
> * None from this round.

@DanielLeens Thank you for your review. Please pull the latest code for review. Startup Mode and Stop Mode have been added

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from the declarative CDC option-rule path down to the actual startup/stop offset consumption path again, and I still see one blocking regression on the current version.

## What this PR solves
- User pain: several CDC connectors still rely on older hand-written validation paths, so submission-time behavior is inconsistent.
- Approach: this PR migrates those connectors to `OptionRule` / `ConfigValidator` and adds a batch of factory tests.
- One-line summary: the migration direction is good, but Oracle and SQLServer still have a gap where the new declarative rules no longer enforce companion fields that the runtime still consumes.

## Call chain I checked
```text
Job submission
  -> OracleIncrementalSourceFactory.optionRule()
  -> SqlServerIncrementalSourceFactory.optionRule()
      -> ConfigValidator enforces submission-time constraints

Source initialization
  -> IncrementalSource.getStartupConfig(...) [IncrementalSource.java:131-136]
  -> IncrementalSource.getStopConfig(...) [IncrementalSource.java:214-219]

Offset creation
  -> StartupConfig.getStartupOffset(...) [StartupConfig.java:39-54]
  -> StopConfig.getStopOffset(...) [StopConfig.java:40-49]
```

## Blocking item
1. Oracle and SQLServer still consume `SPECIFIC` / `TIMESTAMP` offset companion fields at runtime, but the new declarative rules no longer validate those combinations at submission time.
   - Evidence:
     - `seatunnel-connectors-v2/connector-cdc/connector-cdc-oracle/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/source/OracleIncrementalSourceFactory.java:137-147`
     - `seatunnel-connectors-v2/connector-cdc/connector-cdc-sqlserver/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/sqlserver/source/SqlServerIncrementalSourceFactory.java:101-111`
     - `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/option/SourceOptions.java:67-102`
     - `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/IncrementalSource.java:131-136,214-219`
     - `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/config/StartupConfig.java:39-54`
     - `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/config/StopConfig.java:40-49`
   - Why this is a blocker:
     - On the current head, the runtime still supports and consumes `startup.mode=SPECIFIC`, `stop.mode=SPECIFIC`, and `stop.mode=TIMESTAMP`.
     - But the new `optionRule()` branches only kept the `STARTUP_MODE=TIMESTAMP` and `STARTUP_MODE=INITIAL` conditionals, so invalid configs that used to fail during submission can now slip deeper into runtime and fail later with missing/null offset inputs.
   - Suggested fix:
     - Please restore the missing companion-field validation in the Oracle and SQLServer `OptionRule` definitions.
     - If the intention is actually to deprecate those modes, then the runtime support, docs, and incompatible-change notes need to move together. Right now only the validation layer changed.

## Test coverage
- The new tests look stable structurally; I did not see flaky patterns in the added UTs.
- The main gap is coverage, not stability:
  - `OracleIncrementalSourceFactoryTest.java:52-127`
  - `SqlServerIncrementalSourceFactoryTest.java:54-136`
  - Those tests still do not cover the most important `SPECIFIC` / `TIMESTAMP` submission-time validation cases, so this regression is not locked down yet.

## Merge conclusion
- Source-level conclusion: I do not think this version is ready to merge until the missing Oracle / SQLServer offset-mode validation is restored.
- Upstream sync blocker: the current head `13b9e51b9b7e64e02400368f3a022dab5f86ac89` is still behind the latest `dev` (`ahead_by=14`, `behind_by=4`, compare status: `diverged`).
- CI / merge gate snapshot: `Build` is still in progress on the current head. After fixing the blocker above, please sync with the latest `dev` and rerun CI on top of that updated branch first.

Please take another look at those Oracle / SQLServer mode combinations and feel free to ping me after the next update.

DanielLeens (commented):
Thanks for the update. I pulled the latest head `45899b425b` and re-reviewed the current branch from scratch again, specifically rechecking the points you called out around startup / stop mode validation.

The good news is that the source-level blockers from my previous Daniel reviews are now closed on the current head:
- the PR scope is now back to CDC-related files only
- the Oracle endpoint prerequisite is validated at submission time
- the Oracle `include.schema.changes` / Debezium override gap is covered
- the Oracle / SQLServer startup and stop mode companion-field checks are back in the submit-time path
- the Mongo collection/schema alignment path is restored as well

## What this PR solves
- User pain: several CDC connectors still let invalid configs slip past submission and fail deeper in runtime.
- Fix approach: move those checks into declarative `OptionRule + Conditions + ConditionExtension` rules and add focused factory tests.
- One-line summary: the latest head now makes the submission-time validation contract much closer to the real CDC runtime path, and I do not see a reopened source-level blocker in this round.

## Runtime chain I rechecked
```text
Job submission
  -> XxxIncrementalSourceFactory.optionRule()
      -> OptionRule.builder()
          -> required / bundled / exclusive / optional + Conditions.extension(...)

Oracle path
  -> OracleIncrementalSourceFactory.optionRule() [OracleIncrementalSourceFactory.java:60-138]
      -> EndpointValidator [197-209]
      -> SchemaChangeLogMiningValidator [212-235]
      -> OracleStartModeValidator [238-290]
      -> OracleStopModeValidator [292-334]
  -> only after that does runtime enter OracleSourceConfigFactory.create() [OracleSourceConfigFactory.java:98-138]

SQLServer path
  -> SqlServerIncrementalSourceFactory.optionRule()
      -> SqlServerStartModeValidator / SqlServerStopModeValidator
  -> runtime later consumes the same companion fields via StartupConfig.getStartupOffset() and StopConfig.getStopOffset()

Mongo path
  -> MongodbIncrementalSourceFactory.optionRule() [84-141]
      -> CollectionCountConsistencyValidator
      -> MongoStartModeValidator / MongoStopModeValidator
  -> createSource()
      -> updateAndValidateCatalogTableId(...) [262-286]
```

## Findings
- The current factory tests also cover the main cases I was worried about before:
  - Oracle endpoint absence
  - Debezium `include.schema.changes` override
  - Oracle / SQLServer mode-dependent companion fields
  - Mongo collection/schema alignment

## Tests / stability
- The added tests look structurally stable to me:
  - no fixed sleeps
  - no port or environment-sensitive setup
  - focused unit-level validation of the rule contract
- Stability rating: Stable.

## CI / upstream sync snapshot
- The latest `Build` is still in progress on the current head.
- This branch is also currently behind the latest `dev` (`ahead_by=15`, `behind_by=3`, compare status: `diverged`).
- I am recording that upstream-sync fact, but I am not treating it as the main blocker in this round because I do not currently see a source-level problem that still needs code changes from this PR.

## Merge conclusion

1. Blocking items
- Please wait for the current `Build` to finish cleanly before merge.

- None from this round.

Overall, this latest revision addresses the main Daniel blockers I had on the validation path. From my side, the remaining gate is the current CI / merge state rather than a reopened CDC source-path issue.

DanielLeens (commented):
Thanks for the update. I rechecked the latest head `0600d8d39d` again against the full CDC validation path and the current delta since my previous Daniel re-review.

## What this PR solves
- User pain: several CDC connectors still let invalid configs slip too far into runtime instead of failing early at submission time.
- Fix approach: move those checks into declarative `OptionRule + Conditions + ConditionExtension` rules and add focused factory tests.
- One-line summary: the current head still keeps the source-level blockers from the earlier Daniel rounds closed, and the latest delta is only a test correction around the actual exception type for an invalid MySQL startup mode value.

## What changed since the previous Daniel full re-review
- Relative to the earlier reviewed head `45899b425b`, the current latest code-side delta is the MySQL test fix:
  - `MySqlIncrementalSourceFactoryTest.testStartupModeInvalidValueFails()` now expects `IllegalArgumentException` instead of `OptionValidationException`
- That matches the real parse-time failure boundary for an invalid enum-like startup mode input, so I do not treat it as a reopened source issue.

## Findings
- The earlier Daniel conclusions still stand for me on the validation path:
  - Oracle endpoint prerequisite validation
  - Debezium schema-change override coverage
  - Oracle / SQLServer startup-stop companion-field checks
  - Mongo collection/schema alignment

## CI / merge-state snapshot
- The latest `Build` is still in progress on the current head.
- This branch is also still behind the latest `dev`, and I am recording that merge-state fact in the queue, but I am not treating it as a standalone sync-first blocker in this round while the current build is still running.

## Merge conclusion

1. Blocking items
- Please wait for the current `Build` result before merge.

- None from this round.

Overall, the current head still looks consistent to me from the source side. The remaining gate is the current CI / merge state rather than a reopened CDC validation problem.

DanielLeens (approved):
Rechecked the current head and CI gate after my previous source-level review. My last Daniel conclusion on this PR had no remaining code-side blocker; it was waiting on the CI / merge gate.

The current visible checks are now successful, including `Build`. From Daniel's side, the earlier source-level blocker is cleared and I do not see a reopened issue on the unchanged head.

Approving this from the current Daniel technical review state. If GitHub still shows a repository-level merge gate, that should be handled by the normal maintainer / branch protection flow.

DanielLeens (changes requested):
Thanks for the follow-up commits. I re-reviewed the current head after the latest CDC validation and E2E updates.

# What this PR fixes
This PR moves CDC startup / stop mode checks into declarative option validation and adds regression coverage for the migrated semantics.

```text
CDC job submission
  -> connector SourceFactory validates startup / stop mode options
  -> source config is materialized into connector runtime
  -> source reader starts from INITIAL / LATEST / specific timestamp or offset
  -> E2E latest-startup tests assert that historical rows are not replayed
```

# Current review findings
## Issue 1: The new CDC E2E coverage is currently failing on PR-scoped paths
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-tidb-e2e/src/test/java/org/apache/seatunnel/e2e/connector/tidb/TiDBCDCIT.java:191`
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-oracle-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/oracle/OracleCDCWithSchemaChangeIT.java:304`
- Evidence: current Build failed in `updated-modules-integration-test-part-4` and `oracle-cdc-connector-it`.
- TiDB failure: `testLatestStartupMode` timed out because the sink still contained old id `1` while the latest-startup test expected only post-start rows.
- Oracle failure: `testOracleCdcSchemaChangeExcludeDropColumn` timed out with expected `<9>` but actual `<1>`.
- Why this blocks merge: these failures are in the CDC E2E lanes added/changed by this PR, so I cannot treat them as unrelated upstream noise yet.
- Suggested fix: please make the latest-startup / schema-change E2E deterministic and confirm whether the runtime should skip the pre-existing rows before the job reaches `RUNNING`. If the runtime behavior is correct and the test is racing the CDC start boundary, move the data mutation behind a stronger readiness signal; if the runtime is wrong, fix the startup cursor handling itself.
- Severity: High

- Fetched the PR head into `seatunnel-review-11023` and reviewed the current full diff / commit list.
- Checked the failing GitHub Build check and the fork job logs for the failed E2E lanes.

# Merge conclusion
Blocking items:
1. Issue 1: current PR-scoped CDC E2E failures must be fixed and rerun.

The source migration direction still makes sense, but the latest head cannot be approved while the new CDC regression coverage is failing on the behavior this PR changes.

DanielLeens (commented):
Thanks for the follow-up. I re-reviewed the latest head `87a79e1a601b` from scratch and specifically rechecked the earlier validation concerns raised on this PR.

# What this PR fixes
- User pain: several CDC connectors still surfaced invalid startup/stop/schema-change combinations too late, and the related E2E coverage was inconsistent across connectors.
- Fix approach: move more of that contract into declarative validation and align the CDC E2E scenarios with the intended behavior.
- In one sentence: the latest head closes the earlier validation gaps and the current source path looks coherent to me.

# Runtime path rechecked
```text
job submission
  -> connector SourceFactory validates startup / stop / schema-change options
  -> source config is materialized into connector runtime
  -> connector reader starts from INITIAL / LATEST / specific startup mode
  -> schema-change E2E coverage asserts the intended path per connector
```

# Current findings
- The MongoDB collection/schema consistency gap called out earlier is addressed on the current head: `MongodbIncrementalSourceFactory.java:125-154` now routes through `updateAndValidateCatalogTableId(...)` plus `CollectionCountConsistencyValidator` before constructing the source.
- The Oracle Debezium `include.schema.changes` bypass gap is also addressed for me: `OracleIncrementalSourceFactoryTest.java:113-130` now covers the overridden Debezium property path explicitly.
- I do not see a reopened source-level blocker on the latest head.
- `Build` is green on this revision: https://github.com/apache/seatunnel/runs/86176438661

The branch is still behind the latest `dev` by 3 commits in current compare metadata, which I am recording separately as a mergeability fact rather than asking for a new contributor action on a green source/CI state.

Because I only have read-level review controls on this repository, this follow-up review is the current Daniel-side technical conclusion for the latest head.

DanielLeens (approved):
Rechecked the latest gate after my previous follow-up. My last Daniel-side source conclusion on this head had no remaining code blocker, and the current visible `Build` check is successful: https://github.com/apache/seatunnel/runs/86176438661

I do not see a reopened source-level issue on the unchanged head. The branch is still `diverged` from `dev` with `behind_by=4`, which I am recording as a mergeability fact rather than a new contributor action while the current source review and Build are both green.

Approving this from Daniel's current technical review state.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=27`, `behind_by=12`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

DanielLeens:
Rechecked the same unchanged head `87a79e1a601b` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
job submission
  -> connector SourceFactory validates startup / stop / schema-change options
  -> source config is materialized into connector runtime
  -> connector reader starts from INITIAL / LATEST / specific startup mode
  -> schema-change E2E coverage asserts the intended path per connector
```

## Merge conclusion

1. Blocking items

2. Current gate
- The current CI signal is now green: https://github.com/apache/seatunnel/actions/runs/29034709799/job/86176411570

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `87a79e1a601b` from scratch.

# What this PR fixes
- User pain: Omitting `database-names` on Oracle/Postgres/SqlServer CDC crashed at runtime with an NPE, invalid `schema-changes.include/exclude` names or out-of-range numeric options slipped past submission-time validation, and Oracle's dead `checkNotNull(originUrl)` made the hostname+port configuration branch unreachable — all errors surfaced late and cryptically instead of as clear config validation failures.
- Fix approach: Migrate imperative checks (`checkNotNull`, `SchemaChangeEventFilter.validateOptions`, MongoDB builder `.validate()`) into the declarative `OptionRule` framework using `Conditions.greaterThan/notEmpty/extension` plus new `ConditionExtension` validators (`SchemaChangeNameValidator`, `QualifiedTableNameValidator`, `CollectionCountConsistencyValidator`), and flip `database-names` from optional to required where the runtime unconditionally dereferences it.
- One-line summary: Validation logic is consolidated into `optionRule()` with unit tests per connector factory, but the deletion of the public static `SchemaChangeEventFilter.validateOptions` and the removed MongoDB `Builder.validate()` call mean any code path that constructs sources without going through `ConfigValidator` (programmatic/restore paths) now silently loses those checks, and there are no doc updates for the newly-required options.

# Runtime chain I rechecked
```text
Job submission (validation now happens here, before the factory is invoked):
  FactoryUtil.createAndPrepareSource() -> ConfigValidator.validate(config, factory.optionRule())  FactoryUtil.java
    -> OracleIncrementalSourceFactory.optionRule() / PostgresIncrementalSourceFactory.optionRule() / SqlServerIncrementalSourceFactory.optionRule()  (DATABASE_NAMES now .required(...))
    -> SourceOptions.getBaseRule()  SourceOptions.java:153 (Conditions.greaterThan on SNAPSHOT_SPLIT_SIZE:156, SNAPSHOT_FETCH_SIZE:157, INCREMENTAL_PARALLELISM:158-159)
    -> Conditions.extension(...) -> SourceOptions.SchemaChangeNameValidator.evaluate()  SourceOptions.java:188 -> SchemaChangeEventType.fromCanonicalNames() -> throws OptionValidationException:193
    -> SourceOptions.QualifiedTableNameValidator.evaluate()  SourceOptions.java:216 -> isQualifiedName():224
    -> MongodbIncrementalSourceFactory.optionRule()  MongodbIncrementalSourceFactory.java:59 -> CollectionCountConsistencyValidator via Conditions.extension:69-72

Source creation (validation REMOVED from this path):
  BaseChangeStreamTableSourceFactory.createSource()  BaseChangeStreamTableSourceFactory.java:49 (SchemaChangeEventFilter.validateOptions call deleted) -> restoreSource(context, emptyList)
  BaseChangeStreamTableSourceFactory.restoreSource()  BaseChangeStreamTableSourceFactory.java:55 (same deletion)
    -> IncrementalSource ctor -> createSourceConfigFactory(config)  (checkNotNull calls removed; Oracle hostname+port branch now reachable)
    -> MongodbIncrementalSource.createSourceConfigFactory()  MongodbIncrementalSource.java:79 (Builder.validate() call removed — only safe if MongodbSourceConfigProvider.Builder.build() re-validates internally)

Runtime parse (unchanged, still the last line of defense):
  SchemaChangeEventFilter.fromConfig()  SchemaChangeEventFilter.java:~65 -> SchemaChangeEventType.fromCanonicalNames() (throws IllegalArgumentException if factory validation was bypassed)
```

# Findings
**Issue 1: Possible public API change in seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/schema/SchemaChangeEventFilter.java**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/schema/SchemaChangeEventFilter.java`
- Severity: High

**Issue 2: Config/option changes without doc updates**
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/schema/SchemaChangeEventFilter.java

- Issue 2: Config/option changes without doc updates

davidzollo (approved):
+1
LGTM

This PR has quite a lot of changes for CDC

DanielLeens (commented):
Sorry for missing this compatibility issue in my earlier review. I rechecked the full current diff and agree that the normal submission path now reaches the new `OptionRule` validators.

I do need to +1 @SEZ9's compatibility point: `SchemaChangeEventFilter.validateOptions(ReadonlyConfig)` is a public static method in the locked base, and this PR removes it without a compatibility shim or deprecation path. Moving the implementation to `SourceOptions.SchemaChangeNameValidator` does not preserve existing callers. Please keep the old method and delegate to the new validation implementation, or document and handle this explicitly as an incompatible change.

`Build` is green, but that cannot clear this source/API compatibility blocker.

DanielLeens (commented):
The merge facts have changed, but the source conclusion has not cleared.

- The current head still removes the public static `SchemaChangeEventFilter.validateOptions(ReadonlyConfig)` method without a compatibility shim. Moving the implementation to the new OptionRule validator does not preserve existing callers.
- GitHub is now reporting merge conflicts with the latest base (`mergeStateStatus=DIRTY` / `mergeable_state=dirty`).
- The required Build is green, but CI cannot clear the API-compatibility or conflict gates.

Please restore the old method as a delegating compatibility entry point (or document and migrate it as an incompatible change), then resolve the base conflicts and rerun checks.

DanielLeens:

nzw921rx:
Close this PR and gradually migrate CDC to reduce the scope of impact.

PR #11032, opened 2026-06-09
[Feature][Connector-Fake] Migrate validation to OptionRule

Ayushkale11:
Migrate connector-fake configuration validation from imperative runtime checks to declarative OptionRule + Conditions.

What I changed
Added declarative validation rules in FakeSourceFactory.
Migrated range validations for:
TINYINT_MIN/MAX
SMALLINT_MIN/MAX
INT_MIN/MAX
BIGINT_MIN/MAX
FLOAT_MIN/MAX
DOUBLE_MIN/MAX
VECTOR_FLOAT_MIN/MAX
Removed corresponding runtime validation checks from FakeConfig.
Preserved existing validation behavior.
Related Issue

Part of the declarative validation migration tracking issue.

Claimed item: connector-fake

Verification
git diff --check passed.
Attempted local compilation of connector-fake.
Build was blocked by snapshot dependency resolution (seatunnel-shade artifact unavailable), unrelated to connector-fake changes.

### DIFF
--- a/seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java
+++ b/seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java
@@ -37,8 +37,12 @@ import java.io.Serializable;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import static org.apache.seatunnel.api.configuration.util.Conditions.greaterOrEqual;
+import static org.apache.seatunnel.api.configuration.util.Conditions.lessOrEqual;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.ARRAY_SIZE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_FAKE_MODE;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_MIN;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BINARY_VECTOR_DIMENSION;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BYTES_LENGTH;
@@ -46,15 +50,23 @@ import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOp
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DATE_MONTH_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DATE_YEAR_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_FAKE_MODE;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_MIN;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_FAKE_MODE;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_MIN;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_FAKE_MODE;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_MIN;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.MAP_SIZE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.ROWS;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.ROW_NUM;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_FAKE_MODE;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_MIN;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SPLIT_NUM;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SPLIT_READ_INTERVAL;
@@ -64,24 +76,12 @@ import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOp
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TIME_MINUTE_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TIME_SECOND_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_FAKE_MODE;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_MIN;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.VECTOR_DIMENSION;
-import static org.apache.seatunnel.api.configuration.util.Conditions.greaterOrEqual;
-import static org.apache.seatunnel.api.configuration.util.Conditions.lessOrEqual;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_MIN;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_MAX;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_MIN;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_MAX;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_MIN;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_MAX;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_MIN;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_MAX;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_MIN;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_MAX;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_MIN;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_MAX;
-import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.VECTOR_FLOAT_MIN;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.VECTOR_FLOAT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.VECTOR_FLOAT_MIN;
 
 @AutoService(Factory.class)
 public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRunValidation {
@@ -94,7 +94,6 @@ public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRu
     public OptionRule optionRule() {
         return OptionRule.builder()
                 .exclusive(ConnectorCommonOptions.TABLE_CONFIGS, ConnectorCommonOptions.SCHEMA)
-
                 .optional(
                         STRING_FAKE_MODE,
                         TINYINT_FAKE_MODE,
@@ -118,10 +117,6 @@ public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRu
                         TIME_HOUR_TEMPLATE,
                         TIME_MINUTE_TEMPLATE,
                         TIME_SECOND_TEMPLATE)
-
-
-
-
                 .conditional(STRING_FAKE_MODE, FakeSourceOptions.FakeMode.TEMPLATE, STRING_TEMPLATE)
                 .conditional(
                         TINYINT_FAKE_MODE, FakeSourceOptions.FakeMode.TEMPLATE, TINYINT_TEMPLATE)
@@ -131,121 +126,67 @@ public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRu
                 .conditional(BIGINT_FAKE_MODE, FakeSourceOptions.FakeMode.TEMPLATE, BIGINT_TEMPLATE)
                 .conditional(FLOAT_FAKE_MODE, FakeSourceOptions.FakeMode.TEMPLATE, FLOAT_TEMPLATE)
                 .conditional(DOUBLE_FAKE_MODE, FakeSourceOptions.FakeMode.TEMPLATE, DOUBLE_TEMPLATE)
-                 .optional(
-                TINYINT_MIN,
-                greaterOrEqual(TINYINT_MIN, TINYINT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        TINYINT_MIN,
-                                        TINYINT_MAX.defaultValue())))
-
-                 .optional(
-                TINYINT_MAX,
-                greaterOrEqual(TINYINT_MAX, TINYINT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        TINYINT_MAX,
-                                        TINYINT_MAX.defaultValue())))
-
-
                 .optional(
-                SMALLINT_MIN,
-                greaterOrEqual(SMALLINT_MIN, SMALLINT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        SMALLINT_MIN,
-                                        SMALLINT_MAX.defaultValue())))
-
-
-                 .optional(
-                SMALLINT_MAX,
-                greaterOrEqual(SMALLINT_MAX, SMALLINT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        SMALLINT_MAX,
-                                        SMALLINT_MAX.defaultValue())))
-
-                 .optional(
-                INT_MIN,
-                greaterOrEqual(INT_MIN, INT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        INT_MIN,
-                                        INT_MAX.defaultValue())))
-
-                 .optional(
-                INT_MAX,
-                greaterOrEqual(INT_MAX, INT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        INT_MAX,
-                                        INT_MAX.defaultValue())))
-
-                 .optional(
-                BIGINT_MIN,
-                greaterOrEqual(BIGINT_MIN, BIGINT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        BIGINT_MIN,
-                                        BIGINT_MAX.defaultValue())))
-
-                 .optional(
-                BIGINT_MAX,
-                greaterOrEqual(BIGINT_MAX, BIGINT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        BIGINT_MAX,
-                                        BIGINT_MAX.defaultValue())))
-
-                 .optional(
-                FLOAT_MIN,
-                greaterOrEqual(FLOAT_MIN, FLOAT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        FLOAT_MIN,
-                                        FLOAT_MAX.defaultValue())))
-                 .optional(
-                FLOAT_MAX,
-                greaterOrEqual(FLOAT_MAX, FLOAT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        FLOAT_MAX,
-                                        FLOAT_MAX.defaultValue())))
-
-                 .optional(
-                DOUBLE_MIN,
-                greaterOrEqual(DOUBLE_MIN, DOUBLE_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        DOUBLE_MIN,
-                                        DOUBLE_MAX.defaultValue())))
-                 .optional(
-                DOUBLE_MAX,
-                greaterOrEqual(DOUBLE_MAX, DOUBLE_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        DOUBLE_MAX,
-                                        DOUBLE_MAX.defaultValue())))
-
-                 .optional(
-                VECTOR_FLOAT_MIN,
-                greaterOrEqual(VECTOR_FLOAT_MIN, VECTOR_FLOAT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        VECTOR_FLOAT_MIN,
-                                        VECTOR_FLOAT_MAX.defaultValue())))
-                 .optional(
-                VECTOR_FLOAT_MAX,
-                greaterOrEqual(VECTOR_FLOAT_MAX, VECTOR_FLOAT_MIN.defaultValue())
-                        .and(
-                                lessOrEqual(
-                                        VECTOR_FLOAT_MAX,
-                                        VECTOR_FLOAT_MAX.defaultValue())))
-
-
-        .build();
-
-
+                        TINYINT_MIN,
+                        greaterOrEqual(TINYINT_MIN, TINYINT_MIN.defaultValue())
+                                .and(lessOrEqual(TINYINT_MIN, TINYINT_MAX.defaultValue())))
+                .optional(
+                        TINYINT_MAX,
+                        greaterOrEqual(TINYINT_MAX, TINYINT_MIN.defaultValue())
+                                .and(lessOrEqual(TINYINT_MAX, TINYINT_MAX.defaultValue())))
+                .optional(
+                        SMALLINT_MIN,
+                        greaterOrEqual(SMALLINT_MIN, SMALLINT_MIN.defaultValue())
+                                .and(lessOrEqual(SMALLINT_MIN, SMALLINT_MAX.defaultValue())))
+                .optional(
+                        SMALLINT_MAX,
+                        greaterOrEqual(SMALLINT_MAX, SMALLINT_MIN.defaultValue())
+                                .and(lessOrEqual(SMALLINT_MAX, SMALLINT_MAX.defaultValue())))
+                .optional(
+                        INT_MIN,
+                        greaterOrEqual(INT_MIN, INT_MIN.defaultValue())
+                                .and(lessOrEqual(INT_MIN, INT_MAX.defaultValue())))
+                .optional(
+                        INT_MAX,
+                        greaterOrEqual(INT_MAX, INT_MIN.defaultValue())
+                                .and(lessOrEqual(INT_MAX, INT_MAX.defaultValue())))
+                .optional(
+                        BIGINT_MIN,
+                        greaterOrEqual(BIGINT_MIN, BIGINT_MIN.defaultValue())
+                                .and(lessOrEqual(BIGINT_MIN, BIGINT_MAX.defaultValue())))
+                .optional(
+                        BIGINT_MAX,
+                        greaterOrEqual(BIGINT_MAX, BIGINT_MIN.defaultValue())
+                                .and(lessOrEqual(BIGINT_MAX, BIGINT_MAX.defaultValue())))
+                .optional(
+                        FLOAT_MIN,
+                        greaterOrEqual(FLOAT_MIN, FLOAT_MIN.defaultValue())
+                                .and(lessOrEqual(FLOAT_MIN, FLOAT_MAX.defaultValue())))
+                .optional(
+                        FLOAT_MAX,
+                        greaterOrEqual(FLOAT_MAX, FLOAT_MIN.defaultValue())
+                                .and(lessOrEqual(FLOAT_MAX, FLOAT_MAX.defaultValue())))
+                .optional(
+                        DOUBLE_MIN,
+                        greaterOrEqual(DOUBLE_MIN, DOUBLE_MIN.defaultValue())
+                                .and(lessOrEqual(DOUBLE_MIN, DOUBLE_MAX.defaultValue())))
+                .optional(
+                        DOUBLE_MAX,
+                        greaterOrEqual(DOUBLE_MAX, DOUBLE_MIN.defaultValue())
+                                .and(lessOrEqual(DOUBLE_MAX, DOUBLE_MAX.defaultValue())))
+                .optional(
+                        VECTOR_FLOAT_MIN,
+                        greaterOrEqual(VECTOR_FLOAT_MIN, VECTOR_FLOAT_MIN.defaultValue())
+                                .and(
+                                        lessOrEqual(
+                                                VECTOR_FLOAT_MIN, VECTOR_FLOAT_MAX.defaultValue())))
+                .optional(
+                        VECTOR_FLOAT_MAX,
+                        greaterOrEqual(VECTOR_FLOAT_MAX, VECTOR_FLOAT_MIN.defaultValue())
+                                .and(
+                                        lessOrEqual(
+                                                VECTOR_FLOAT_MAX, VECTOR_FLOAT_MAX.defaultValue())))
+                .build();
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java
+++ b/seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java
@@ -17,7 +17,11 @@
 
 package org.apache.seatunnel.connectors.seatunnel.fake.source;
 
+import org.apache.seatunnel.api.configuration.ReadonlyConfig;
+import org.apache.seatunnel.api.configuration.util.ConditionExtension;
+import org.apache.seatunnel.api.configuration.util.ConfigValidator;
 import org.apache.seatunnel.api.configuration.util.OptionRule;
+import org.apache.seatunnel.api.configuration.util.OptionValidationException;
 import org.apache.seatunnel.api.options.ConnectorCommonOptions;
 import org.apache.seatunnel.api.source.SeaTunnelSource;
 import org.apache.seatunnel.api.source.SourceSplit;
@@ -35,8 +39,10 @@ import com.google.auto.service.AutoService;
 
 import java.io.Serializable;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
+import static org.apache.seatunnel.api.configuration.util.Conditions.extension;
 import static org.apache.seatunnel.api.configuration.util.Conditions.greaterOrEqual;
 import static org.apache.seatunnel.api.configuration.util.Conditions.lessOrEqual;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.ARRAY_SIZE;
@@ -186,7 +192,48 @@ public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRu
                                 .and(
                                         lessOrEqual(
                                                 VECTOR_FLOAT_MAX, VECTOR_FLOAT_MAX.defaultValue())))
+                .valueConstraint(
+                        extension(
+                                ConnectorCommonOptions.TABLE_CONFIGS,
+                                new TableConfigsValidationExtension()))
                 .build();
+    }
+
+    private static final class TableConfigsValidationExtension
+            implements ConditionExtension<List<Map<String, Object>>> {
+
+        @Override
+        public String description() {
+            return "each tables_configs entry must satisfy the FakeSource option rules";
+        }
+
+        @Override
+        public boolean evaluate(ReadonlyConfig config, List<Map<String, Object>> tableConfigs)
+                throws OptionValidationException {
+            if (tableConfigs == null) {
+                return true;
+            }
+            OptionRule childRule = new FakeSourceFactory().optionRule();
+            for (int index = 0; index < tableConfigs.size(); index++) {
+                Map<String, Object> childConfig = tableConfigs.get(index);
+                if (childConfig == null) {
+                    throw new OptionValidationException(
+                            "Invalid tables_configs[%s]: child config must not be null", index);
+                }
+                if (childConfig.containsKey(ConnectorCommonOptions.TABLE_CONFIGS.key())) {
+                    throw new OptionValidationException(
+                            "Invalid tables_configs[%s]: nested tables_configs is not supported",
+                            index);
+                }
+                try {
+                    ConfigValidator.of(ReadonlyConfig.fromMap(childConfig)).validate(childRule);
+                } catch (OptionValidationException e) {
+                    throw new OptionValidationException(
+                            "Invalid tables_configs[%s]: %s", index, e.getRawMessage());
+                }
+            }
+            return true;
+        }
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java
+++ b/seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java
@@ -66,6 +66,22 @@ import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOp
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_FAKE_MODE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_TEMPLATE;
 import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.VECTOR_DIMENSION;
+import static org.apache.seatunnel.api.configuration.util.Conditions.greaterOrEqual;
+import static org.apache.seatunnel.api.configuration.util.Conditions.lessOrEqual;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_MIN;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.TINYINT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_MIN;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.SMALLINT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_MIN;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.INT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_MIN;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.BIGINT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_MIN;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.FLOAT_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_MIN;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.DOUBLE_MAX;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.VECTOR_FLOAT_MIN;
+import static org.apache.seatunnel.connectors.seatunnel.fake.config.FakeSourceOptions.VECTOR_FLOAT_MAX;
 
 @AutoService(Factory.class)
 public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRunValidation {
@@ -78,6 +94,7 @@ public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRu
     public OptionRule optionRule() {
         return OptionRule.builder()
                 .exclusive(ConnectorCommonOptions.TABLE_CONFIGS, ConnectorCommonOptions.SCHEMA)
+
                 .optional(
                         STRING_FAKE_MODE,
                         TINYINT_FAKE_MODE,
@@ -101,6 +118,10 @@ public class FakeSourceFactory implements TableSourceFactory, SupportSourceDryRu
                         TIME_HOUR_TEMPLATE,
                         TIME_MINUTE_TEMPLATE,
                         TIME_SECOND_TEMPLATE)
+
+
+
+
                 .conditional(STRING_FAKE_MODE, FakeSourceOptions.FakeMode.TEMPLATE, STRING_TEMPLATE)
                 .conditional(
                         TINYINT_FAKE_MODE, FakeSourceOptions.FakeMode.TEMPLATE, TINYINT_TEMPLATE)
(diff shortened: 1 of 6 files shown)


nzw921rx:
@Ayushkale11 Thank you for your contribution. Please enable CI.

Ayushkale11:
Thank you for the reminder.

GitHub Actions are enabled on my fork. Could you please let me know if there is any additional step required from my side to enable CI for this PR?

DanielLeens (changes requested):
Thanks for working on this! I went through the full diff and traced the actual factory/runtime path for `FakeSource`. The migration direction makes sense, but I found one blocker in the current head that needs to be fixed before this can merge.

- User pain point: the numeric range checks for `FakeSource` were implemented imperatively inside `FakeConfig.buildWithConfig()`, so validation logic was scattered and happened relatively late.
- Fix approach: move those checks into `FakeSourceFactory.optionRule()` with declarative `greaterOrEqual/lessOrEqual` constraints, and remove the old imperative checks from `FakeConfig`.
- One-line summary: this is meant to be a behavior-preserving validation migration, from builder-time checks to factory-level declarative checks.

Simple example:
- Single-table config with `tinyint.min = 200`: after this PR, that still fails early because the root config goes through `ConfigValidator.validate(factory.optionRule())`.
- Multi-table config with `table_configs = [{ tinyint.min = 200 }]`: that child config no longer goes through the same validation path, so the bad value now slips through.

Actual call chain I traced:
```text
Single-table path
  FactoryUtil.createAndPrepareSource() [FactoryUtil.java:183-193]
    -> ConfigValidator.validate(factory.optionRule())
    -> FakeSourceFactory.createSource() [FakeSourceFactory.java:245-249]
       -> new FakeSource(context.getOptions())
          -> new MultipleTableFakeSourceConfig(readonlyConfig)
             -> parseFromConfig() [MultipleTableFakeSourceConfig.java:71-73]
                -> FakeConfig.buildWithConfig() [FakeConfig.java:181-286]

Multi-table path
  FactoryUtil.createAndPrepareSource() [FactoryUtil.java:183-193]
    -> ConfigValidator.validate(factory.optionRule())
       -> validates only the top-level config
    -> FakeSourceFactory.createSource()
       -> new FakeSource(context.getOptions())
          -> new MultipleTableFakeSourceConfig(readonlyConfig)
             -> parseFromConfigs() [MultipleTableFakeSourceConfig.java:59-68]
                -> ReadonlyConfig.fromMap(each child config)
                -> FakeConfig.buildWithConfig() [FakeConfig.java:181-286]
                   -> now only copies tinyint/int/float/... min/max values
                   -> no longer performs the old range checks
```

Issue 1: multi-table `table_configs` lost the range validation entirely
- Location:
  - `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java:88-240`
  - `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/MultipleTableFakeSourceConfig.java:59-68`
  - `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:221-274`
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/factory/FactoryUtil.java:189-192`
- Problem:
  the new `OptionRule` only runs against the top-level `ReadonlyConfig` inside `FactoryUtil.createAndPrepareSource()`. In `table_configs` mode, each child config is materialized later in `MultipleTableFakeSourceConfig.parseFromConfigs()`, and `FakeConfig.buildWithConfig()` no longer contains the old imperative bounds checks. So the migration is only complete for the single-table path; the multi-table path regressed.
- Why this is risky:
  this is a real behavior change on a normal runtime path, not just a test gap. Invalid child configs that used to fail immediately can now flow into `FakeDataRandomUtils.randomTinyint/randomInt/randomFloat/randomDouble` and fail later with a much less actionable runtime error. For example, `tinyint.min = 200` can now reach `RandomUtils.nextInt(200, 127)` in `FakeDataRandomUtils.java:81-87`.
- Suggested fix:
  either keep the old guardrails in `FakeConfig.buildWithConfig()` until child `table_configs` are also validated declaratively, or explicitly run equivalent validation for each child config inside `parseFromConfigs()`.
- Severity: High

Test coverage note:
- I also did not find regression coverage for this migration. `FakeFactoryTest` only checks that `optionRule()` is non-null, and `MultipleTableFakeSourceConfigTest` only checks that two tables are parsed successfully. Please add coverage for both:
  - invalid single-table root config
  - invalid child config inside `table_configs`

Compatibility / side effects:
- API/config names/defaults are unchanged.
- Historical behavior is not preserved on the multi-table path, because invalid child configs are no longer rejected at the same point.
- No meaningful CPU/memory/concurrency concern from the migration itself; the main side effect is later and harder-to-diagnose runtime failure on bad configs.

Code quality / tests / docs:
- Code style is broadly fine, but the current implementation is not behavior-equivalent yet.
- No UT/E2E test files were changed in this PR, so there is no flaky-test-specific blocker here.
- No docs update is needed for the intended feature, but I would avoid claiming “validation behavior is preserved” until the multi-table path is covered too.

Merge conclusion:

1. Blocking items
- Issue 1: multi-table `table_configs` no longer gets the old range validation, so the current head has a real regression.

2. Suggested follow-ups
- Add regression tests for both the single-table path and the `table_configs` path when you fix the blocker above.

Overall, I like the direction here, but I don't think the current patch is behavior-preserving yet. Happy to re-review once the child-config path is covered too.

Ayushkale11:
@DanielLeens 
Thank you for the detailed review and for tracing both execution paths.

You're right. I focused on migrating the validation to `OptionRule` and verified the single-table path, but I missed that `table_configs` child configurations are materialized later in `MultipleTableFakeSourceConfig.parseFromConfigs()` and therefore do not go through the same top-level validation flow.

As you pointed out, removing the imperative checks from `FakeConfig.buildWithConfig()` introduces a regression for multi-table configurations.

I will work on preserving validation for the child-config path and add regression tests for both:

* invalid single-table configuration
* invalid configuration inside `table_configs`

Thank you again for the thorough review.

nzw921rx:
@Ayushkale11  This validation gap will be addressed once the EXTENSION operator (#11048) is merged — table_configs child configs can then be validated declaratively via ConditionExtension.

nzw921rx:
> Thank you for the reminder.
> 
> GitHub Actions are enabled on my fork. Could you please let me know if there is any additional step required from my side to enable CI for this PR?

You can take a look at the steps here: https://github.com/apache/seatunnel/pull/11032/checks?check_run_id=80372832103

DanielLeens:
Thanks for the ping. There is still no new commit after my last review, so I am not doing another full code pass on the current head.

On the current revision, the multi-table `table_configs` validation regression I called out is still present, and depending on `#11048` is not enough yet from my side because that PR itself still has open blockers. I also see the current `Build` state as `ACTION_REQUIRED`, so please enable / rerun CI together with the follow-up code change. Happy to re-review once the child-config path is covered and there is a new commit to inspect.

Ayushkale11:
@DanielLeens 

I've addressed the regression concern by preserving the existing validation behavior for child configurations and added regression coverage for both validation paths:

* invalid single-table configuration validation
* invalid configuration inside `tables_configs`

A new commit has been pushed for review. Please let me know if there are any additional concerns.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from the actual `FakeSource` validation path down to the new regression tests. The runtime regression I called out in the previous round is no longer the main problem on the current head. The remaining blocker is now in the new test itself.

# What This PR Solves
- User pain: `FakeSource` range checks were scattered in `FakeConfig.buildWithConfig()`, so validation happened relatively late.
- Fix approach: move the simple bounds checks into `FakeSourceFactory.optionRule()` and add regression coverage.
- One-line summary: the production-side regression from the previous round is fixed, but the new child-config regression test is malformed and currently fails before it reaches the real validation path.

# Runtime / Test Chain Rechecked
```text
Single-table path
  -> FactoryUtil.createAndPrepareSource()
      -> ConfigValidator.validate(factory.optionRule())
      -> FakeSourceFactory.createSource()

Multi-table path
  -> FactoryUtil.createAndPrepareSource()
      -> validate top-level optionRule()
  -> FakeSource()
      -> new MultipleTableFakeSourceConfig(readonlyConfig)
          -> parseFromConfigs()
              -> FakeConfig.buildWithConfig(each child config)

Current CI blocker
  -> MultipleTableFakeSourceConfigTest.invalidTinyintMinInTablesConfigsShouldFail()
      -> ConfigFactory.parseString(configStr)
      -> HOCON parse error before connector validation runs
```

# Findings
Issue 1: the new child-config regression test is malformed, so CI fails in HOCON parsing before it reaches the connector validation you actually want to verify
- Location: `seatunnel-connectors-v2/connector-fake/src/test/java/org/apache/seatunnel/connectors/seatunnel/fake/config/MultipleTableFakeSourceConfigTest.java:48-61`
- Why this is a blocker:
  the current `configStr` does not separate `tinyint.min = 200` and `schema = { ... }` correctly, so `ConfigFactory.parseString(...)` throws first. That means the test is not proving “invalid child config is rejected by FakeSource validation”; it is only proving the inline HOCON string is syntactically broken.
- Risk:
  this is a deterministic CI failure on the normal unit-test path, and it hides the real behavior the regression test was meant to protect.
- Better fix:
  make the inline HOCON valid first, or reuse a small `.conf` resource instead of hand-assembling the string.
- Severity: High

# Test Coverage / Stability
- The intended coverage is good: one root-config negative test plus one `table_configs` negative test.
- Stability rating: High risk, because the new child-config test currently fails deterministically for the wrong reason.

# Merge Conclusion

1. Blocking items
- Issue 1: fix the malformed HOCON in `MultipleTableFakeSourceConfigTest` so the test actually validates the child-config path.

2. Suggested follow-up
- No extra non-blocking items from me on this revision.

Overall, the production-side validation path is in much better shape now. The remaining work is to make the new regression test validate the intended behavior instead of failing at config parsing.

Ayushkale11:
@DanielLeens 
You're right — the previous regression test was failing during HOCON parsing rather than exercising the intended child-config validation path.

I've updated the test to use a valid HOCON configuration so that it reaches the FakeSource validation logic and verifies the expected behavior for invalid values inside tables_configs.

A new commit has been pushed for review.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch against the real FakeSource validation path, the child-config path, and the current test diff.

# What This PR Solves
- User pain: FakeSource range checks were scattered in FakeConfig.buildWithConfig(), so validation happened relatively late.
- Fix approach: move the simple bounds checks into FakeSourceFactory.optionRule() while keeping child-config validation intact for tables_configs.
- One-line summary: the earlier runtime-regression blocker and the malformed HOCON test blocker are both closed on the latest head; the remaining gate is the current red Build rather than a new FakeSource logic bug.

# Runtime Chain Rechecked
```text
single-table path
  -> FactoryUtil.createAndPrepareSource()
      -> ConfigValidator.validate(factory.optionRule())
      -> FakeSourceFactory.createSource()

multi-table path
  -> FactoryUtil.createAndPrepareSource()
      -> validate top-level optionRule()
  -> FakeSource()
      -> new MultipleTableFakeSourceConfig(readonlyConfig)
          -> parseFromConfigs()
              -> FakeConfig.buildWithConfig(each child config)
                  -> still enforces child-config numeric bounds
```

# Findings
I did not find a new FakeSource logic blocker on the latest head.

The two issues from my earlier Daniel round are closed now:
- the production child-config validation is still present in FakeConfig.buildWithConfig(...), so tables_configs does not lose the old numeric guardrails;
- the new regression test now reaches connector validation instead of failing in raw HOCON parsing first.

Issue 1: the top-level Build is still red, and the current failure spread does not point to the FakeSource code path specifically
- Location: GitHub Build check `80846059057`
- Evidence I confirmed:
  the current run shows broad matrix failures/cancellations rather than a focused FakeSource unit-test regression, including cancelled kafka legs and unrelated matrix churn.
- Why this matters:
  I no longer see a code-side blocker in the latest FakeSource validation path, but the PR still should not merge while the top-level Build remains red.
- Better fix:
  rerun or resync the branch to get a clean Build signal on the current head, rather than changing the FakeSource business logic again without a targeted failing signal.
- Severity: High
- Already raised by others: No. This is the remaining gate after the earlier code-side blockers were fixed.

# Test Coverage / Stability
- The added negative tests now cover both the top-level optionRule validation and the tables_configs child-config validation path.
- Stability rating: Stable.
  - The current tests are in-memory config-validation tests.
  - I do not see sleep-based timing, ports, or environment-sensitive patterns here.

# Merge Conclusion

1. Blocking items
- Issue 1: get the current Build green on this head.

2. Suggested follow-up
- None from the FakeSource validation path I reviewed.

Overall, the code-side concerns from the previous round are resolved on the latest head. The remaining blocker is the current CI gate, not a new validation-path bug.

nzw921rx:
@Ayushkale11 Hello, you can use the rebase dev branch to pull the latest code and validate `table_configs` through the Extension plugin.

DanielLeens:
Thanks, and the maintainer note about rebasing onto the latest `dev` / EXTENSION flow matches my read as well.

Since there is still no new commit after Daniel's latest full pass on this PR, I am keeping this as reply-only for now rather than doing another full review on the same revision. Once you push the rebase / validation update, I will re-review the latest head.

On the current revision, the last Daniel blocker around the child-config validation path should still be treated as unresolved until it is reflected in code.

DanielLeens:
Hi @Ayushkale11, thank you for taking on the FakeSource validation migration and for responding thoughtfully to each earlier review round. I continued directly on your original PR, rebased the two author commits onto current dev, and completed the ConditionExtension integration requested by the maintainer at 22504262e.

Single-table and tables_configs child values now use the same declarative numeric constraints. OptionRule can attach a value constraint to an option already registered by an exclusive rule, child validation reports the exact tables_configs index, and the old duplicated imperative range checks have been removed. Focused tests cover the new API path plus invalid and valid nested FakeSource values.

The PR is mergeable, API and connector Spotless apply/check both passed locally, and GitHub CI has started on the updated original head. Your original author commits and metadata remain intact. Thank you for sticking with a cross-layer validation change — the groundwork you provided made this completion possible.

DanielLeens (commented):
# Basic information

- PR title: [Feature][Connector-Fake] Migrate validation to OptionRule
- Author: Ayushkale11
- Link: https://github.com/apache/seatunnel/pull/11032
- Base branch: dev
- State: OPEN
- Change size: +244 / -239, 5 file(s)
- Human review ready: Yes
- Review time: 2026-07-26T15:58:28Z
- Local branch: seatunnel-review-11032
- Local worktree: shared bare PR 11032 review store

## Stable review keys

- pr_number: `11032`
- base_sha: `924b289e95d0bfe06286e8666bcbb7d986bc2023`
- reviewed_base_sha: `924b289e95d0bfe06286e8666bcbb7d986bc2023`
- merge_base_sha: `924b289e95d0bfe06286e8666bcbb7d986bc2023`
- head_sha: `22504262e84b8e66115613955816cefd891ede48`
- reviewed_head_sha: `22504262e84b8e66115613955816cefd891ede48`
- technical_verdict: `clear_waiting_ci`
- human_review_ready: `true`
- intended_change_correctness: `pass`
- diff_integrity: `pass`
- deletion_total: `239`
- deletion_classified: `239`
- deletion_unclassified: `0`
- other_total: `4`
- other_confirmed: `0`
- other_resolved_in_current_head: `4`
- other_unresolved: `0`
- merge_conclusion: `gates_then_merge`
- reviewed_at: `2026-07-26T15:58:28Z`

- mergeStateStatus: `BLOCKED`
- mergeable_state: `blocked`
- mergeable: `True`
- rebaseable: `False`
- required_ci_gate: `pending`
- approval_gate: `pending`
- conflict_gate: `satisfied`
- merge_queue_gate: `pending`
- compare: status=`ahead`, ahead_by=`3`, behind_by=`0`, total_commits=`3`, url=https://github.com/apache/seatunnel/compare/dev...22504262e84b8e66115613955816cefd891ede48

# What problem this PR solves

- User pain: FakeSource numeric-range validation lived in FakeConfig construction and the initial migration missed tables_configs children.
- Fix approach: Move all 14 bounds into the factory OptionRule and use an extension to apply the same rule to every child.
- One-line summary: Root and nested FakeSource configurations now reject invalid ranges through one declarative path.
- Before: tinyint.min=200 failed only during FakeConfig materialization and a tables_configs child could bypass it.
- After: Invalid root and tables_configs[0] values fail in ConfigValidator while valid values reach source construction.

# 1. Code-change review

- Behavior before: tinyint.min=200 failed only during FakeConfig materialization and a tables_configs child could bypass it.
- Behavior after: Invalid root and tables_configs[0] values fail in ConfigValidator while valid values reach source construction.
- Normal-path hit: `conditional` — FactoryUtil.createAndPrepareSource → ConfigValidator/FakeSourceFactory.optionRule → createSource → MultipleTableFakeSourceConfig/FakeConfig.

- `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java:100-199` `optionRule` — FactoryUtil executes scalar min/max checks and the tables_configs extension before createSource.
- `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java:202-236` `TableConfigsValidationExtension.evaluate` — The extension rejects null/nested children and reuses the same OptionRule for each child.
- `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:179-245` `FakeConfig.buildWithConfig` — FakeConfig only copies validated values and no longer repeats range if/throw blocks.

### Effective prerequisites

- Effective for root and child configs in standard factory source creation and dry-run schema inference.

- Tests/internal code calling connector-internal FakeConfig.buildWithConfig directly no longer auto-validates; that bypass is outside the supported user factory SPI.

### Complete execution chain

```text
1. seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java:100-199 -> optionRule [method_call]
   FactoryUtil executes scalar min/max checks and the tables_configs extension before createSource.
   Side effect: FactoryUtil executes scalar min/max checks and the tables_configs extension before createSource.
2. seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java:202-236 -> TableConfigsValidationExtension.evaluate [method_call]
   The extension rejects null/nested children and reuses the same OptionRule for each child.
   Side effect: The extension rejects null/nested children and reuses the same OptionRule for each child.
3. seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:179-245 -> FakeConfig.buildWithConfig [serialization]
   FakeConfig only copies validated values and no longer repeats range if/throw blocks.
   Side effect: FakeConfig only copies validated values and no longer repeats range if/throw blocks.
```

- Recovery, exception, async, serialization, and concurrency paths: This is synchronous config validation with no I/O, threads, retries, locks, checkpoint, restore, or resource lifecycle; nested lists are rejected before recursion.

- Applicable: `false`

## 1.2 Compatibility impact

- API: On supported factory paths, options, types, defaults, accepted ranges, SPI, generated data, and state are unchanged; failures move earlier from FakeConnectorException to OptionValidationException.
- SPI: On supported factory paths, options, types, defaults, accepted ranges, SPI, generated data, and state are unchanged; failures move earlier from FakeConnectorException to OptionValidationException.
- Configuration: On supported factory paths, options, types, defaults, accepted ranges, SPI, generated data, and state are unchanged; failures move earlier from FakeConnectorException to OptionValidationException.
- Defaults: On supported factory paths, options, types, defaults, accepted ranges, SPI, generated data, and state are unchanged; failures move earlier from FakeConnectorException to OptionValidationException.
- Protocol and serialization: On supported factory paths, options, types, defaults, accepted ranges, SPI, generated data, and state are unchanged; failures move earlier from FakeConnectorException to OptionValidationException.
- Checkpoint/restore: On supported factory paths, options, types, defaults, accepted ranges, SPI, generated data, and state are unchanged; failures move earlier from FakeConnectorException to OptionValidationException.
- Historical behavior: On supported factory paths, options, types, defaults, accepted ranges, SPI, generated data, and state are unchanged; failures move earlier from FakeConnectorException to OptionValidationException.

- CPU: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Memory: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- GC: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Network: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Lock contention: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Concurrency safety: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Retry: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Idempotency: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Resource release: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.
- Overall: Scalar validation is constant-size; multi-table validation is linear per child with a small ReadonlyConfig wrapper and no record-hot-path change.

- Exception propagation: Null/nested children and child validation failures retain tables_configs[index] context without logging credentials or data values.
- Context: Null/nested children and child validation failures retain tables_configs[index] context without logging credentials or data values.
- Log levels: Null/nested children and child validation failures retain tables_configs[index] context without logging credentials or data values.
- Sensitive information: Null/nested children and child validation failures retain tables_configs[index] context without logging credentials or data values.
- Overall: Null/nested children and child validation failures retain tables_configs[index] context without logging credentials or data values.

- intended_change_correctness: `pass`
- diff_integrity: `pass`
- deletion_ledger: total=`239`, classified=`239`, unclassified=`0`
- Reconciliation: 239 = 239 + 0

- `deletion-001..deletion-239` | count=`239` | `deletion` | `intended_change` | `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java` (old line 23 .. old line 457) — All 239 deleted lines are old FakeConfig imports and 14 repetitive range checks, each replaced equivalently in OptionRule.

- Out-of-scope hunk conclusion: The five files only cover rule API/tests, FakeSource rules/extensions, old-check removal, and regressions.
- Stale-base conclusion: No option/default/row/schema/assertion/docs capability is removed and no stale-base rollback or unrelated hunk exists.
- Merge-artifact conclusion: mergeable=true and merge_state=blocked rather than dirty, with no merge-tree conflict candidate.

# 2. Code-quality assessment

- Title format: The code follows OptionRule/FakeSource structure; [Improve][Connector-V2] would be a more accurate non-blocking title.
- Code conventions: The code follows OptionRule/FakeSource structure; [Improve][Connector-V2] would be a more accurate non-blocking title.
- Comment completeness: The code follows OptionRule/FakeSource structure; [Improve][Connector-V2] would be a more accurate non-blocking title.

- Unit tests: `covered` — Synchronous in-memory tests cover an exclusive-option extension, invalid root/child values, and a valid child without sleep, network, container, port, or shared state.
- Component tests: `covered` — Synchronous in-memory tests cover an exclusive-option extension, invalid root/child values, and a valid child without sleep, network, container, port, or shared state.
- Engine integration tests: `not_applicable` — Synchronous in-memory tests cover an exclusive-option extension, invalid root/child values, and a valid child without sleep, network, container, port, or shared state.
- E2E: `not_applicable` — Synchronous in-memory tests cover an exclusive-option extension, invalid root/child values, and a valid child without sleep, network, container, port, or shared state.
- Regression tests: `covered` — Synchronous in-memory tests cover an exclusive-option extension, invalid root/child values, and a valid child without sleep, network, container, port, or shared state.

### Test-stability evidence

- `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:179-245`
  - Trigger: Tests use maps and synchronous assertions with no clock, network, filesystem ordering, or readiness signal.
  - Probability basis: Tests use maps and synchronous assertions with no clock, network, filesystem ordering, or readiness signal.
  - Analysis: Tests use maps and synchronous assertions with no clock, network, filesystem ordering, or readiness signal.

### Missing scenarios

- Non-blocking additions: all numeric boundaries, null/nested children, a FactoryUtil production path, and an unregistered-valueConstraint negative case.

## 2.3 Documentation

- User-visible change: `false`
- docs/en: Options, defaults, and accepted ranges are unchanged, so connector docs and incompatible-change notes are not required.
- docs/zh: Options, defaults, and accepted ranges are unchanged, so connector docs and incompatible-change notes are not required.
- Examples: Options, defaults, and accepted ranges are unchanged, so connector docs and incompatible-change notes are not required.
- Configuration truth: Options, defaults, and accepted ranges are unchanged, so connector docs and incompatible-change notes are not required.
- URL truth: Options, defaults, and accepted ranges are unchanged, so connector docs and incompatible-change notes are not required.

### Local execution boundary

- `git diff --stat 924b289e95d0bfe06286e8666bcbb7d986bc2023...22504262e84b8e66115613955816cefd891ede48`
  - Result: `passed`
  - Explanation: Confirmed 5 files and +244/-239.
- `git diff --name-status 924b289e95d0bfe06286e8666bcbb7d986bc2023...22504262e84b8e66115613955816cefd891ede48`
  - Result: `passed`
- `git diff --unified=0 924b289e95d0bfe06286e8666bcbb7d986bc2023...22504262e84b8e66115613955816cefd891ede48`
  - Result: `passed`
- `git diff --check`
  - Result: `passed`

### Required checks

- `Build` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`QUEUED` | bucket=`pending` | conclusion=`QUEUED` | https://github.com/apache/seatunnel/runs/89799812933

- `Notify test workflow` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/30204384389/job/89799798295
- `labeler` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/30204384392/job/89799798272
- `Build` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`QUEUED` | bucket=`pending` | conclusion=`Not reported by GitHub` | https://github.com/apache/seatunnel/runs/89799812933

# 3. Architecture assessment

- Type: `long_term_design`
- Assessment: Factory-level declarative validation is the correct boundary and is reused by normal creation and dry-run.

## 3.2 Maintainability

Factory-level declarative validation is the correct boundary and is reused by normal creation and dry-run.

## 3.3 Extensibility

Factory-level declarative validation is the correct boundary and is reused by normal creation and dry-run.

Factory-level declarative validation is the correct boundary and is reused by normal creation and dry-run.

- total=`4`, confirmed=`0`, resolved_in_current_head=`4`
- Reconciliation: 4 = 0 + 4 + 0 + 0 + 0

## `issuecomment-4661838629` — nzw921rx

- Original proposition: Enable fork CI.
- Location: PR discussion
- Thread: https://github.com/apache/seatunnel/pull/11032#issuecomment-4661838629
- Current-head evidence: A required Build exists for the current head and is QUEUED.
- Disposition: `resolved_in_current_head`
- Daniel disposition: The current head satisfies this proposition.

## `issuecomment-4668353546` — nzw921rx

- Original proposition: Use the EXTENSION operator after #11048 to validate tables_configs children.
- Location: PR discussion
- Thread: https://github.com/apache/seatunnel/pull/11032#issuecomment-4668353546
- Current-head evidence: TableConfigsValidationExtension now applies the same FakeSource rule to every child.
- Disposition: `resolved_in_current_head`
- Daniel disposition: The current head satisfies this proposition.

## `issuecomment-4668362649` — nzw921rx

- Original proposition: Follow the linked steps to enable CI.
- Location: PR discussion
- Thread: https://github.com/apache/seatunnel/pull/11032#issuecomment-4668362649
- Current-head evidence: The current head has a queued Build object.
- Disposition: `resolved_in_current_head`
- Daniel disposition: The current head satisfies this proposition.

## `issuecomment-4690095611` — nzw921rx

- Original proposition: Rebase current dev and use Extension.
- Location: PR discussion
- Thread: https://github.com/apache/seatunnel/pull/11032#issuecomment-4690095611
- Current-head evidence: The merge base equals the reviewed base, behind=0, and the Extension is present.
- Disposition: `resolved_in_current_head`
- Daniel disposition: The current head satisfies this proposition.

# 5. Formal findings

## `问题 1` — valueConstraint does not enforce its already-registered-option precondition

- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:375-384`
- Problem and real execution path: The new public builder method only appends a condition while collectDeclaredKeys treats condition options as declared, so future misuse can silently expand accepted keys. The new public builder method only appends a condition while collectDeclaredKeys treats condition options as declared, so future misuse can silently expand accepted keys.
- Current-head evidence: The new public builder method only appends a condition while collectDeclaredKeys treats condition options as declared, so future misuse can silently expand accepted keys.
- Trigger condition: A future connector calls valueConstraint for an option not registered through optional/required/exclusive/bundled.
- Potential impact: Unknown-key detection and rule metadata accept an unintended key; the current FakeSource use is safe because TABLE_CONFIGS is already exclusive.
- Best actionable fix: Reuse optional(TABLE_CONFIGS, condition), or validate at build time that every constraint head belongs to a declared option and add a negative test.
- Severity: `medium`
- Merge gate: `non_blocking`

# 6. Findings summary

|---|---|---|---|---|
| 问题 1 | valueConstraint does not enforce its already-registered-option precondition | seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:375-384 | medium | non_blocking |

# 7. Merge conclusion

## Conclusion：can merge after gates are satisfied

### Blocking items

- No source, documentation, compatibility, test-completeness, conflict, or diff-integrity blocker remains.

- `问题 1`: Reuse optional(TABLE_CONFIGS, condition), or validate at build time that every constraint head belongs to a declared option and add a negative test.

### Overall assessment

Factory-level declarative validation is the correct boundary and is reused by normal creation and dry-run.

- Alternative: Existing optional(TABLE_CONFIGS, extension) avoids a new API; if valueConstraint remains, enforce its registration precondition at build time.
- Trade-offs: Existing optional(TABLE_CONFIGS, extension) avoids a new API; if valueConstraint remains, enforce its registration precondition at build time.

- Reviewed commit: `22504262e84b8e66115613955816cefd891ede48`

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `22504262e84b` from scratch.

# What this PR fixes
- User pain: connector-fake validated its numeric range options (TINYINT/SMALLINT/INT/BIGINT/FLOAT/DOUBLE/VECTOR_FLOAT min/max) imperatively deep inside `FakeConfig.buildWithConfig`, so misconfigurations surfaced late at runtime with ad-hoc `FakeConnectorException` messages instead of through the declarative `OptionRule` framework used elsewhere.
- Fix approach: add a new `OptionRule.Builder.valueConstraint(Condition, Condition...)` API so constraints can be attached to options registered by other rules (e.g. exclusive options), wire the range checks into `FakeSourceFactory`'s option rule via `Conditions`, and delete the ~225 lines of `ifPresent`-based checks from `FakeConfig`.
- One-line summary: the migration direction is right and the new `valueConstraint` builder API is tested, but as diffed the removed blocks in `FakeConfig` also contained the `builder.xxxMin/Max(...)` setter calls — I need confirmation those were re-added, plus the API lacks registration/null guards and the behavior change (exception type, validation bypassed for direct `buildWithConfig` callers) needs a note.

# Runtime chain I rechecked
```text
FactoryUtil.createAndPrepareSource()  FactoryUtil.java:89
  -> ConfigValidator.of(readonlyConfig).validate(factory.optionRule())  ConfigValidator.java:47
    -> validates valueConstraints added via OptionRule.Builder.valueConstraint()  OptionRule.java:379 (new)
      -> Conditions.extension(...)/range conditions declared in FakeSourceFactory.optionRule()  FakeSourceFactory.java
  -> FakeSourceFactory.createSource() -> new FakeSource(readonlyConfig)
    -> FakeConfig.buildWithConfig(readonlyConfig)  FakeConfig.java:198 (runtime range checks removed here, +14/-239)
      -> builder.tinyintMin/…/vectorFloatMax(...) — previously invoked inside the deleted ifPresent blocks; must still be called or configured values are dropped
```

# Findings
**Issue 1: Range validation removed from FakeConfig.buildWithConfig leaves non-factory code paths with no input validation at all**
- Location: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:217`
- Why it matters: - Root cause: the imperative bound checks (e.g., tinyint_min must be within [-128, 127]) were deleted from `FakeConfig.buildWithConfig(ReadonlyConfig)`, and the replacement validation now lives only in the `FakeSourceFactory` OptionRule, which is enforced solely when `ConfigValidator` runs against the factory's `optionRule()`.
- Impact: `FakeConfig.buildWithConfig` is a public static entry point. Any path that builds `FakeConfig` without going through the factory validation step (unit/IT harnesses, programmatic API usage, or engine paths that instantiate the source without invoking `ConfigValidator` on the factory rule — behavior differs across Zeta/Flink/Spark translation layers) will now accept out-of-range values like `tinyint_min: 500` silently, and the failure surfaces later inside `FakeDataGenerator` as corrupted generated data or an obscure arithmetic/cast error instead of a clear config error.
- Fix: keep a compact defensive validation inside `buildWithConfig` (a single helper like `checkRange(option, value, min, max)` reused for all 14 options keeps it to a few lines), or add an explicit `validate()` method on `FakeConfig` invoked from the builder, so the config object can never exist in an invalid state regardless of which entry point constructed it.
- Evidence: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:227 (removed)  throw new FakeConnectorException(`; `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:23 (removed)  import org.apache.seatunnel.common.exception.CommonErrorCodeDeprecated;`; `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:228 (removed)  CommonErrorCodeDeprecated.ILLEGAL_ARGUMENT,`
- Suggested fix: Removing all range checks from `buildWithConfig` means this public entry point now accepts invalid values (e.g., `tinyint_min: 500`) whenever a caller doesn't go through `FakeSourceFactory`'s OptionRule + ConfigValidator. Since `FakeConfig.buildWithConfig` can be invoked directly (tests, programmatic usage, engine translation paths that skip factory validation), please keep a compact defensive check here — a single `checkRange(option, value, min, max)` helper reused for all 14 options would preserve the fail-fast invariant without reintroducing the 200+ lines of duplication.
- Severity: High

**Issue 2: Verify that the setter wiring removed together with the validation blocks was re-added — otherwise configured min/max values are silently ignored**
- Location: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:217`
- Why it matters: - Root cause: each deleted `ifPresent` lambda in `FakeConfig.buildWithConfig` contained both the validation AND the builder assignment (e.g., `builder.tinyintMin(tinyintMin);`, `builder.vectorFloatMin(vectorFloatMin);`). The diff shows 239 lines removed and only 14 added in this file, and the added replacement lines for the 14 options (tinyint/smallint/int/bigint/float/double/vector_float min+max) are not all visible in the hunk.
- Impact: if even one option's `builder::xxx` re-wiring is missing (14 options must map to 14 new `readonlyConfig.getOptional(X).ifPresent(builder::x)` lines), the user's configured value is silently dropped and the connector falls back to the option default, producing wrong fake data with no error on any engine — a silent functional regression that OptionRule validation cannot catch because the value itself is valid.
- Fix: confirm all 14 setter calls exist in the new code, and strengthen `FakeFactoryTest` to assert that a fully-populated config round-trips into `FakeConfig` fields (e.g., set every min/max and assert the built `FakeConfig` getters return them), not just that the OptionRule rejects out-of-range values.
- Evidence: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java (removed)  builder.vectorFloatMin(vecto...`
- Suggested fix: The removed `ifPresent` blocks carried both the validation and the actual builder assignment (`builder.tinyintMin(...)`, ..., `builder.vectorFloatMax(...)`). Please double-check that all 14 setter wirings were re-added as plain `readonlyConfig.getOptional(X).ifPresent(builder::x)` calls — a missing one would silently discard the user's configured value and fall back to the default. It would also be good for `FakeFactoryTest` to assert a full round-trip of all min/max values into the built `FakeConfig`, not only the rejection of invalid ranges.
- Severity: High

**Issue 3: Fake connector validation changes exception type/timing and is silently bypassed for direct `FakeConfig.buildWithConfig` callers — needs a release/upgrade note**
- Location: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:216`
- Why it matters: - Root cause: this PR deletes all range checks (and the `FakeConnectorException` / `CommonErrorCodeDeprecated.ILLEGAL_ARGUMENT` imports) from `FakeConfig.buildWithConfig` and relies solely on the `OptionRule` in `FakeSourceFactory`. `FakeConfig.buildWithConfig(ReadonlyConfig)` is a public static entry point; any caller that constructs the config directly (tests, tooling, or engine paths that instantiate the source without going through the factory's `ConfigValidator`) now receives out-of-range values with no error at all.
- Impact: users who previously matched on `FakeConnectorException` with error code `ILLEGAL_ARGUMENT` (e.g., in scripts parsing job-submission failures) will now see `OptionValidationException` with a different message, and validation moves from config-build time to factory-validation time. That is an observable behavior change for a widely used connector (FakeSource is the standard example/testing source) and should be called out.
- Fix: either keep a lightweight defensive check in `buildWithConfig` (asserting the ranges and throwing `FakeConnectorException`) as a second line of defense, or explicitly document the new exception type and validation timing in the release notes / `docs/en/connector-v2/source/FakeSource.md` changelog section so users are not surprised by the changed error surface.
- Evidence: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:216  removed ~239 lines of range validation, e.g. `throw new FakeConnectorException(CommonErrorCodeDeprecated.ILLEGAL_ARGUMENT, TINYINT_MIN.key() + " should >= " ...)``
- Suggested fix: With the range checks removed here, `FakeConfig.buildWithConfig(ReadonlyConfig)` no longer validates anything — any caller that builds the config without going through `FakeSourceFactory`'s `ConfigValidator` will accept out-of-range values silently. Also, the user-visible error changes from `FakeConnectorException(ILLEGAL_ARGUMENT)` to `OptionValidationException`. Please either keep a defensive check here or add a release/upgrade note documenting the new exception type and validation timing.
- Severity: Medium

**Issue 4: New `Builder.valueConstraint(...)` does not verify that the referenced option is actually registered in the rule**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:381`
- Why it matters: - Root cause: the new builder method blindly appends conditions to `valueConstraints` (OptionRule.java:379-384) without any check that the option referenced by the `Condition` was previously registered via `required(...)`, `optional(...)`, or `exclusive(...)`. Other builder paths in `OptionRule.Builder` verify option membership before adding rules; this one silently skips that guard even though its own Javadoc claims the options must be "already registered".
- Impact: a connector author can attach a constraint to an option that is not part of the rule at all (e.g., a typo in the option key), and the mistake will only surface at runtime as confusing `ConfigValidator` behavior — or worse, the constraint may never fire because the option is not resolved. Since `OptionRule` is core API consumed by every connector factory on Zeta/Flink/Spark, this weakens the declarative-validation contract this PR is trying to establish.
- Fix: inside `valueConstraint`, iterate the condition chain (via `Condition#getOption()` / `getNext()`) and assert each referenced option exists in the builder's required/optional/exclusive option sets, throwing an `OptionValidationException` (or `IllegalArgumentException`) with the offending key otherwise, mirroring the verification done for conditional rules.
- Evidence: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:380  @NonNull Condition<?> condition1, @NonNull Condition<?>... conditions) {`; `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:381  this.valueConstraints.add(condition1);`; `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:382  Collections.addAll(this.valueConstraints, conditions);`
- Suggested fix: This method adds conditions without verifying that the option each `Condition` references is registered in this rule (via `required`/`optional`/`exclusive`). The Javadoc says the option must be "already registered", but nothing enforces it, so a typo'd or unregistered option would be silently accepted. Please walk the condition chain and throw with the offending option key if it is not part of the rule, consistent with the verification the builder does for conditional rules.
- Severity: Medium

**Issue 5: Javadoc on the new public API `valueConstraint` is incomplete and its wording is misleading**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:376`
- Why it matters: - Root cause: the Javadoc added at OptionRule.java:375-378 is two sentences with no `@param`/`@return` tags and no explanation of semantics. It also says "options that are already registered by another rule", which is inaccurate — the option is registered in the *same* `OptionRule` by another builder call (e.g., `exclusive(...)`), not by another rule.
- Impact: `OptionRule.Builder` is a public API surface that every connector author uses when writing a `TableSourceFactory`/`TableSinkFactory`. Without documenting when to prefer `valueConstraint` over the existing `conditional(...)` overloads, and that the constraint is evaluated unconditionally by `ConfigValidator` whenever the option is present, connector authors across the codebase will misuse it during the ongoing declarative-validation migration this PR is part of.
- Fix: expand the Javadoc to
- (a) document `@param condition1`, `@param conditions`, and `@return`,
- (b) state that the constraint applies unconditionally (unlike `conditional`, which is gated on another option's value), and
- (c) correct the wording to "registered in this rule via {@link #required}, {@link #optional} or {@link #exclusive}".
- Evidence: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:376  * Adds value constraints for options that are already registered by another rule, such as`; `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:377  * an exclusive option.`
- Suggested fix: Since this is new public API on `OptionRule.Builder`, please flesh out the Javadoc: add `@param`/`@return` tags, explain that the constraint is evaluated unconditionally by `ConfigValidator` (in contrast to `conditional(...)`, which is gated on another option's value), and fix the wording — the option is registered in the *same* rule via `required`/`optional`/`exclusive`, not "by another rule".
- Severity: Medium

**Issue 6: New OptionRule test covers the happy/invalid paths but not misuse of `valueConstraint` with an unregistered option**
- Location: `seatunnel-api/src/test/java/org/apache/seatunnel/api/configuration/util/OptionRuleTest.java:394`
- Why it matters: - Root cause: the new test in `OptionRuleTest` (lines 379-414) exercises `valueConstraint` only with `TEST_TOPIC_PATTERN`, which is registered via `exclusive(TEST_TOPIC_PATTERN, TEST_TOPIC)` in the same rule. There is no test asserting the behavior when the constraint references an option that is *not* part of the rule, which is exactly the edge case the new API's Javadoc contract hinges on.
- Impact: without that negative test, the intended contract ("options that are already registered") is unenforced and untested, so future refactors of `ConfigValidator`/`OptionRule` could change the misuse behavior without any test failing.
- Fix: add a test case that calls `valueConstraint` with a condition on an option absent from the rule and asserts the expected outcome (ideally an exception at `build()` time once the registration check from my other comment is added).
- Evidence: `seatunnel-api/src/test/java/org/apache/seatunnel/api/configuration/util/OptionRuleTest.java:393  .exclusive(TEST_TOPIC_PATTERN, TEST_TOPIC)`; `seatunnel-api/src/test/java/org/apache/seatunnel/api/configuration/util/OptionRuleTest.java:394  .valueConstraint(Conditions.extension(TEST_TOPIC_PATTERN, topicPatternExt))`; `seatunnel-api/src/test/java/org/apache/seatunnel/api/configuration/util/OptionRuleTest.java:390  OptionRule exclusiveExtRule =`
- Suggested fix: Good coverage for the registered-option case. Could you also add a negative test where `valueConstraint` references an option that is not registered in the rule? That pins down the contract stated in the Javadoc and protects it against future refactors of `ConfigValidator`.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Range validation removed from FakeConfig.buildWithConfig leaves non-factory code paths with no input validation at all
- Issue 2: Verify that the setter wiring removed together with the validation blocks was re-added — otherwise configured min/max values are silently ignored

- Issue 3: Fake connector validation changes exception type/timing and is silently bypassed for direct `FakeConfig.buildWithConfig` callers — needs a release/upgrade note
- Issue 4: New `Builder.valueConstraint(...)` does not verify that the referenced option is actually registered in the rule
- Issue 5: Javadoc on the new public API `valueConstraint` is incomplete and its wording is misleading
- Issue 6: New OptionRule test covers the happy/invalid paths but not misuse of `valueConstraint` with an unregistered option

seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/source/FakeSourceFactory.java
```diff
@@ -110,9 +132,110 @@ public OptionRule optionRule() {
...
+                                .and(
+                                        lessOrEqual(
+                                                VECTOR_FLOAT_MIN, VECTOR_FLOAT_MAX.defaultValue())))
+                .optional(
+                        VECTOR_FLOAT_MAX,
+                        greaterOrEqual(VECTOR_FLOAT_MAX, VECTOR_FLOAT_MIN.defaultValue())
+                                .and(
+                                        lessOrEqual(
+                                                VECTOR_FLOAT_MAX, VECTOR_FLOAT_MAX.defaultValue())))
+                .valueConstraint(
```
nzw921rx, line 195: Thank you for your contribution
The existing `optional`, `required`, and `conditional` in this scenario can also express this meaning, and it seems that there is no need to add an API,

- The following case illustrates the following verification process when a user passes in Tableau CONFIGS

examples: 

``` java
.optional(
                        ConnectorCommonOptions.TABLE_CONFIGS,
                        Conditions.extension(
                                ConnectorCommonOptions.TABLE_CONFIGS, new TableConfigsValidationExtension()))
```

nzw921rx (changes requested):

DanielLeens:
Quick follow-up after the later review activity: I saw the additional change requests around the OptionRule API design.

Because there is no newer commit after those comments yet, I will not treat the earlier Daniel-side `clear_waiting_ci` note as the final merge state for this head. Please address the open API/design review points first, then push an updated commit so the full OptionRule migration path can be re-reviewed against the latest code.

The current `Build` check is also failing, but the source/design discussion should be resolved before we spend too much time on the CI signal.

DanielLeens:
Thanks for the update. I rechecked the current head `22504262e84b` against the latest `dev`. This branch is currently `diverged` (`ahead_by=3`, `behind_by=14`), and the current `Build` check is `FAILURE`: https://github.com/apache/seatunnel/runs/89799812933

This is in addition to the earlier source-level review items; those code-side blockers still need to be fixed as well.

DanielLeens:

davidzollo:

```

```

SEZ9 (changes requested):
Thanks for working on this. I reviewed the latest head `22504262e84b`.

# What this PR fixes
- User pain: connector-fake validated numeric min/max ranges imperatively at runtime inside `FakeConfig.buildWithConfig`, so bad values surfaced late with connector-specific exceptions instead of being caught by the declarative `OptionRule` framework like other migrated connectors.
- Fix approach: add a new public `OptionRule.Builder#valueConstraint(Condition, Condition...)` API so constraints can attach to already-registered (e.g. exclusive) options, declare the TINYINT/SMALLINT/INT/BIGINT/FLOAT/DOUBLE/VECTOR_FLOAT range conditions in `FakeSourceFactory.optionRule()`, and delete the ~225 lines of `ifPresent`-guarded checks from `FakeConfig` (with an `OptionRuleTest` case covering the new builder method).
- One-line summary: the migration is sound for top-level options and the new builder API is a reasonable addition, but the removed checks are silently lost for per-table configs nested under `tables_configs` (which never flow through the factory's OptionRule validation), the exception type/message change from `FakeConnectorException(ILLEGAL_ARGUMENT)` to `OptionValidationException` is undocumented, and the new API needs better Javadoc, varargs null-checking, and broader test coverage before merge.

# Runtime chain I rechecked
```text
Top-level path (validation preserved):
  SeaTunnel engine job compile -> AbstractPluginDiscovery / FactoryUtil.createAndPrepareSource()  FactoryUtil.java:~90
    -> ConfigValidator.of(readonlyConfig).validate(factory.optionRule())  ConfigValidator.java:~45
      -> FakeSourceFactory.optionRule() returns rule with new range Conditions  FakeSourceFactory.java (changed)
      -> ConfigValidator.validateValueConstraints() evaluates each Condition, throws OptionValidationException on violation
        -> OptionRule.Builder.valueConstraint(Condition, Condition...)  OptionRule.java:375-382 (new API populating valueConstraints)
    -> FakeSourceFactory.createSource() -> FakeConfig.buildWithConfig(readonlyConfig)  FakeConfig.java:~200 (range checks now removed; safe here because rule already ran)

Nested per-table path (validation lost -- the HIGH finding):
  FakeSource / MultipleTableFakeSourceConfig construction
    -> readonlyConfig.get(FakeOption.TABLES_CONFIGS) -> iterate table entries
      -> FakeConfig.buildWithConfig(perTableReadonlyConfig)  FakeConfig.java:216+ (deleted ifPresent range checks used to run here)
      -> ConfigValidator is never invoked on these nested ReadonlyConfig instances, so e.g. tinyint.min=999 in a tables_configs entry now passes silently and only misbehaves at data-generation time in FakeDataGenerator
```

# Findings
**Issue 1: Range validation removed from FakeConfig.buildWithConfig is silently lost for nested per-table configs (`tables_configs`) that never pass through the factory's OptionRule validation**
- Location: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:216`
- Why it matters: - Root cause: the old checks threw `FakeConnectorException` with `CommonErrorCodeDeprecated.ILLEGAL_ARGUMENT` and a message like `tinyint.min should >= -128 and <= 127`; the new OptionRule-based validation throws `OptionValidationException` with the generic constraint-violation message from `ConfigValidator`.
- Impact: users and platform integrations that match on the error code/message (log alerting, REST error handling in Zeta, e2e assertions) will observe a different failure signature, so the PR description's claim "Preserved existing validation behavior" is not accurate — the *rules* are preserved, the *error surface* is not.
- Fix: call this out explicitly in the PR description and add a line to the release notes / `docs/en/faq.md`-style upgrade notes for the migration tracking issue, so the behavior change is discoverable. Historical review feedback on this repo (PR#10306) also asked for well-defined error classifications rather than ad-hoc changes, so documenting the new error type is the minimum here.
- Evidence: `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:25  -import org.apache.seatunnel.connectors.seatunnel.fake.exception.FakeConnectorException;`; `seatunnel-connectors-v2/connector-fake/src/main/java/org/apache/seatunnel/connectors/seatunnel/fake/config/FakeConfig.java:216  (all TINYINT/SMALLINT/INT/BIGINT/FLOAT/DOUBLE/VECTOR_FLOAT min/max range checks deleted, -239 lines)`
- Suggested fix: Removing all range checks here is not fully equivalent to the OptionRule migration: `buildWithConfig` is also called for each nested config under `tables_configs` (multi-table mode), and those nested configs are never validated by the factory's `OptionRule`/`ConfigValidator`. After this change an out-of-range `tinyint.min`, `int.max`, etc. inside `tables_configs` will pass silently into `FakeDataGenerator`. Please either validate each nested table config against the factory rule in `MultipleTableFakeSourceConfig`, or keep a shared range-check helper here, and add a multi-table negative test.
- Severity: High

**Issue 2: Javadoc of the new public API OptionRule.Builder#valueConstraint is too thin and slightly misleading**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:379`
- Why it matters: - Root cause: `valueConstraint` is a new public method on a widely used API class (`seatunnel-api`), but its Javadoc only says the options must be "already registered by another rule, such as an exclusive option". "Another rule" is misleading — the options must be registered on the *same builder* via `required`/`optional`/`exclusive`, otherwise the referenced option key won't be part of the rule at all.
- Impact: connector authors following this Javadoc may attach constraints to options that are never registered anywhere, and it is unclear from the doc whether `build()` will detect that or whether the constraint is silently ignored — this class is the primary extension point for every connector factory.
- Fix: expand the Javadoc to document `@param`/`@return`, state explicitly that the referenced option must be registered on this builder (e.g. via `exclusive(...)`), and describe what happens (exception vs. no-op) when it is not. Please also update the connector development guide (docs on option rules) if it enumerates the builder methods.
- Evidence: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:379  public Builder valueConstraint(`; `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:380  @NonNull Condition<?> condition1, @NonNull Condition<?>... conditions) {`
- Suggested fix: Since this is a new public method on a core API class, please flesh out the Javadoc: add `@param`/`@return`, clarify that the referenced options must be registered on *this* builder (via `required`/`optional`/`exclusive`) rather than "by another rule", and document what happens at `build()` time if a constraint references an option that was never registered (exception or silently ignored?). Connector authors will rely on this doc.
- Severity: Medium

**Issue 3: New OptionRuleTest case does not cover varargs usage or a constraint on an unregistered option**
- Location: `seatunnel-api/src/test/java/org/apache/seatunnel/api/configuration/util/OptionRuleTest.java:394`
- Why it matters: - Root cause: the added test only exercises `valueConstraint` with a single condition attached to an option registered via `exclusive(...)`.
- Impact: the varargs path (`Collections.addAll`) and the important edge case of a constraint referencing an option that was never registered on the builder are untested, so future refactors could silently change their behavior.
- Fix: add one assertion passing two conditions through the varargs, and one asserting the defined behavior (exception at `build()` or ignore at validate) when the constrained option is not registered by `required`/`optional`/`exclusive` — this also pins down the contract that the Javadoc (see my other comment) needs to document.
- Evidence: `seatunnel-api/src/test/java/org/apache/seatunnel/api/configuration/util/OptionRuleTest.java:393  .exclusive(TEST_TOPIC_PATTERN, TEST_TOPIC)`; `seatunnel-api/src/test/java/org/apache/seatunnel/api/configuration/util/OptionRuleTest.java:394  .valueConstraint(Conditions.extension(TEST_TOPIC_PATTERN, topicPatternExt))`
- Suggested fix: Nice coverage for the happy/failure path. Could you also add (1) a case passing multiple conditions through the varargs, and (2) a case where `valueConstraint` references an option that was never registered on the builder, asserting the expected behavior? That pins down the contract of the new API for future refactors.
- Severity: Low

**Issue 4: valueConstraint varargs elements are not null-checked; a null Condition inside the array will NPE later during validation**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:382`
- Why it matters: - Root cause: `@NonNull` on the varargs parameter only guards against the array itself being null; `Collections.addAll(this.valueConstraints, conditions)` will happily insert null elements.
- Impact: a null condition passed by a connector factory would surface as an opaque `NullPointerException` deep inside `ConfigValidator` at job submission time instead of a clear error at rule-construction time.
- Fix: iterate the varargs and reject null elements with an `IllegalArgumentException` naming the offending position, consistent with how `Conditions.extension(TEST_NUM, null)` throws `IllegalArgumentException` in the existing tests.
- Evidence: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/OptionRule.java:382  Collections.addAll(this.valueConstraints, conditions);`
- Suggested fix: `@NonNull` on the varargs only protects the array itself — `Collections.addAll` will insert null elements, which will NPE much later inside `ConfigValidator`. Please validate each element and throw an `IllegalArgumentException` with the offending index, matching the fail-fast style of `Conditions.extension(option, null)`.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Range validation removed from FakeConfig.buildWithConfig is silently lost for nested per-table configs (`tables_configs`) that never pass through the factory's OptionRule validation

**2. Remaining follow-ups**
- Issue 2: Javadoc of the new public API OptionRule.Builder#valueConstraint is too thin and slightly misleading
- Issue 3: New OptionRuleTest case does not cover varargs usage or a constraint on an unregistered option
- Issue 4: valueConstraint varargs elements are not null-checked; a null Condition inside the array will NPE later during validation

DanielLeens:
@SEZ9 thanks for the detailed re-review. I went back through both of your reviews on the current head (`22504262e84b`) and traced the exact runtime chain again before replying, since your latest review's Issue 1 makes a factual claim I want to check carefully rather than just take on faith.

## Pushing back on Issue 1 in your 2026-08-06 review ("ConfigValidator is never invoked on these nested ReadonlyConfig instances")

I don't think this is accurate on the current head. The `tables_configs` path is wired into the same `ConfigValidator` run as the root config, via a `valueConstraint`:

- `FakeSourceFactory.java:195-198` registers `.valueConstraint(extension(ConnectorCommonOptions.TABLE_CONFIGS, new TableConfigsValidationExtension()))` on the factory's `OptionRule`.
- `FakeSourceFactory.java:202-236` — `TableConfigsValidationExtension.evaluate(...)` loops every entry in `tables_configs` and explicitly calls `ConfigValidator.of(ReadonlyConfig.fromMap(childConfig)).validate(childRule)`, where `childRule` is `new FakeSourceFactory().optionRule()` — i.e. the *same* rule with all 14 min/max range constraints, applied recursively to each child.
- `ConfigValidator.java:230-249` (`collectErrors`) iterates `rule.getValueConstraints()` and evaluates each one, so this constraint is not skipped.
- `ConditionEvaluators.java:168-173` shows the `EXTENSION` operator delegates straight to `ConditionExtension#evaluate(config, value)`, which is exactly the extension above.

So the call chain for a `tables_configs` child is: `FactoryUtil.createAndPrepareSource()` → `ConfigValidator.validate(factory.optionRule())` → constraint loop → `TableConfigsValidationExtension.evaluate()` → nested `ConfigValidator.validate(childRule)` per child. A `tinyint.min = 999` inside `tables_configs[0]` fails at the *same* top-level validation step as a root-level violation, before `FakeSource`/`FakeConfig.buildWithConfig` is ever reached.

This is also covered by a passing test in this PR: `FakeFactoryTest.invalidTinyintMinInTablesConfigsShouldFailValidation()` builds a `tables_configs` config with `tinyint.min = 200`, calls `ConfigValidator.of(...).validate(new FakeSourceFactory().optionRule())` (the same entry point the factory uses), and asserts it throws `OptionValidationException` whose message contains both `"tables_configs[0]"` and `"tinyint.min"`. That's the exact scenario your Issue 1 says passes silently.

For context: this was the very first blocker I raised on this PR back on 2026-06-09 (multi-table `table_configs` losing validation), and I re-verified the fix in three follow-up rounds (2026-06-12, and again in my full review on 2026-07-26) using this same trace. I'd suggest we not re-block on this one unless there's a concrete reproduction that gets past the extension — happy to be shown a config that slips through if I'm missing something.

## Points I agree with (+1, still open, non-blocking)

- **`valueConstraint` doesn't verify the referenced option is registered** (your Issue 4 in the 2026-07-27 review, Issue 2 wording in the 2026-08-06 review) — confirmed, this is the same gap I flagged as my own Issue 1 in the 2026-07-26 review. Looking at `OptionRule.java:375-384`, `valueConstraint(...)` just appends to `valueConstraints` with no check against `required`/`optional`/`exclusive`, unlike the sibling `optional(option, condition, ...)` overload right above it which calls `verifyOptionOptionsDuplicate(...)`. Worth fixing before more connectors adopt this API, but it doesn't affect `FakeSource` today since `TABLE_CONFIGS` is already registered via `exclusive(...)`.
- **Varargs elements in `valueConstraint` aren't null-checked** — confirmed at `OptionRule.java:382`, `Collections.addAll(this.valueConstraints, conditions)` will happily insert a null and defer the failure to a later NPE inside `ConfigValidator`. Agreed this should fail fast at build time instead.
- **No test for `valueConstraint` referencing an unregistered option / multiple varargs conditions** — confirmed, `OptionRuleTest.java` only exercises the single-condition, already-registered case (`TEST_TOPIC_PATTERN` via `.exclusive(...)`). Agreed this is a real coverage gap once the registration check above is added.
- **Javadoc on `valueConstraint` is thin/slightly misleading** ("already registered by another rule" should say "registered on this same builder via `required`/`optional`/`exclusive`") — agreed, same as above.

## One clarification on your 2026-07-27 review, Issue 1/3 (buildWithConfig losing its own defensive checks / exception type change)

I traced this too: in the supported path, `FakeConfig.buildWithConfig(...)` is only ever reached after `FactoryUtil.createAndPrepareSource()` has already run `ConfigValidator.validate(factory.optionRule())` — for both the root config and, per the extension above, every `tables_configs` child. So on the normal user path there's no window where an out-of-range value reaches `buildWithConfig` unvalidated, and the `FakeConnectorException` → `OptionValidationException` change is a real but expected consequence of moving validation earlier, not a silent gap. I agree it's worth a one-line callout in the PR description for anyone matching on the old exception type/message, but I wouldn't treat it as blocking — `buildWithConfig` being usable as an unvalidated internal constructor for direct/test callers is a reasonable trade-off for a connector-internal method, not part of the public factory SPI.

Given the above, my read is that the current head is behavior-preserving on the path that matters (including `tables_configs`), and what's left is legitimate but non-blocking API-hygiene work on the new `valueConstraint` builder method. Let me know if you have a concrete config that reproduces the Issue 1 scenario on this head — I'd want to fix my own understanding if so.

nzw921rx:
Thank you for your contribution. As this PR has not been updated for a long time, we have closed it to release the task.

PR #11048, opened 2026-06-10
[Feature][API] Add EXTENSION operator for pluggable validation in ConditionOperator

nzw921rx:

Add `ConditionOperator.EXTENSION` to allow connectors to plug custom validation logic into the existing `Condition` framework.

**Problem:**

Built-in operators (`greaterOrEqual`, `notBlank`, `notEmpty`, etc.) handle single-value checks well, but connectors sometimes need structural validation on complex types:
- Validate every entry in `List<Map<String, Object>>` contains required keys (`field`, `type`)
- Validate `table_configs` child configs have valid numeric ranges across nested maps

Today this kind of logic is scattered in imperative code like `buildWithConfig()`, making it invisible to the declarative `OptionRule` system, REST metadata API, and CLI tooling.

**Solution:**

Introduce `ConditionExtension<T>` interface + `ConditionOperator.EXTENSION` enum constant. Connector developers implement the interface and wire it via `Conditions.extension(option, ext)`.

The EXTENSION operator is a first-class citizen in the existing pipeline:
- Reuses `valueConstraints` evaluation in `ConfigValidator` — no changes to `OptionRule` or `ConfigValidator`
- `Conditions.extension(Option<T>, ConditionExtension<T>)` enforces type safety at compile time
- Chains with `.and()` / `.or()` and mixes freely with any built-in operator
- REST metadata exposes `conditionOperator: "EXTENSION"` with `expectValue` from `description()`
- Supports two error reporting modes: return `false` for auto-composed messages, or throw `OptionValidationException` for context-rich details

**Usage:**

```java
// Inline — validate encoded API key format (decode + check structure)
.optional(API_KEY_ENCODED, Conditions.extension(API_KEY_ENCODED,
        new ConditionExtension<String>() {
            @Override
            public String description() {
                return "must be Base64-encoded 'id:api_key'";
            }

            @Override
            public boolean evaluate(ReadonlyConfig cfg, String v)
                    throws OptionValidationException {
                try {
                    return new String(Base64.getDecoder().decode(v)).contains(":");
                } catch (IllegalArgumentException e) {
                    return false;
                }
            }
        }))

// Static inner class — validate List<Map> structure with cross-element uniqueness
static class TableConfigsValidator
        implements ConditionExtension<List<Map<String, Object>>> {
    @Override
    public String description() {
        return "each entry must contain a non-empty 'table_name', "
                + "and all table names must be unique";
    }

    @Override
    public boolean evaluate(ReadonlyConfig config, List<Map<String, Object>> value)
            throws OptionValidationException {
        if (value.isEmpty()) {
            return false;
        }
        Set<String> seen = new HashSet<>();
        for (Map<String, Object> entry : value) {
            Object name = entry.get("table_name");
            if (!(name instanceof String) || ((String) name).isEmpty()) {
                return false;
            }
            if (!seen.add((String) name)) {
                return false;
            }
        }
        return true;
    }
}

.exclusive(TABLE_CONFIGS, SCHEMA)
.optional(TABLE_CONFIGS,
        Conditions.extension(TABLE_CONFIGS, new TableConfigsValidator()))

// Chain with built-in operators — extension + numeric range on the same option
.required(BATCH_SIZE,
        Conditions.greaterOrEqual(BATCH_SIZE, 1)
                .and(Conditions.extension(BATCH_SIZE, new ConditionExtension<Integer>() {
                    @Override
                    public String description() {
                        return "must be a power of 2";
                    }

                    @Override
                    public boolean evaluate(ReadonlyConfig cfg, Integer v)
                            throws OptionValidationException {
                        return (v & (v - 1)) == 0;
                    }
                })))
```

No.

```bash
./mvnw -pl seatunnel-api test
```

### DIFF
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Condition.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Condition.java
@@ -32,15 +32,25 @@ public class Condition<T> {
     private final T expectValue;
     private final ConditionOperator operator;
     private final Option<?> compareOption;
+    private final ConditionExtension<T> extension;
     private Boolean and = null;
     private Condition<?> next = null;
 
     Condition(Option<T> option, T expectValue) {
-        this(option, ConditionOperator.EQUAL, expectValue, null);
+        this(option, ConditionOperator.EQUAL, expectValue, null, null);
     }
 
     Condition(
             Option<T> option, ConditionOperator operator, T expectValue, Option<?> compareOption) {
+        this(option, operator, expectValue, compareOption, null);
+    }
+
+    Condition(
+            Option<T> option,
+            ConditionOperator operator,
+            T expectValue,
+            Option<?> compareOption,
+            ConditionExtension<T> extension) {
         if (option == null) {
             throw new IllegalArgumentException("Condition option must not be null");
         }
@@ -61,10 +71,15 @@ public class Condition<T> {
                             "Operator %s requires an expectValue, but expectValue is null",
                             operator.name()));
         }
+        if (operator == ConditionOperator.EXTENSION && extension == null) {
+            throw new IllegalArgumentException(
+                    "Operator EXTENSION requires a non-null ConditionExtension");
+        }
         this.option = option;
         this.operator = operator;
         this.expectValue = expectValue;
         this.compareOption = compareOption;
+        this.extension = extension;
     }
 
     public static <T> Condition<T> of(Option<T> option, T expectValue) {
@@ -148,6 +163,7 @@ public class Condition<T> {
                 && Objects.equals(this.expectValue, that.expectValue)
                 && Objects.equals(this.operator, that.operator)
                 && Objects.equals(this.compareOption, that.compareOption)
+                && extensionClassEquals(this.extension, that.extension)
                 && Objects.equals(this.and, that.and)
                 && Objects.equals(this.next, that.next);
     }
@@ -159,8 +175,15 @@ public class Condition<T> {
                 this.expectValue,
                 this.operator,
                 this.compareOption,
+                this.extension == null ? 0 : this.extension.getClass().hashCode(),
                 this.and,
                 this.next);
+    }
+
+    private static boolean extensionClassEquals(ConditionExtension<?> a, ConditionExtension<?> b) {
+        if (a == b) return true;
+        if (a == null || b == null) return false;
+        return a.getClass() == b.getClass();
     }
 
     /**
@@ -219,6 +242,9 @@ public class Condition<T> {
         ConditionOperator op = cond.operator;
         String key = "'" + cond.option.key() + "'";
 
+        if (op == ConditionOperator.EXTENSION) {
+            return key + " " + cond.extension.description();
+        }
         if (op.getSource() == ConditionOperator.Source.FIELD) {
             return key + " " + op.getSymbol() + " '" + cond.compareOption.key() + "'";
         }
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Condition.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Condition.java
@@ -163,7 +163,6 @@ public class Condition<T> {
                 && Objects.equals(this.expectValue, that.expectValue)
                 && Objects.equals(this.operator, that.operator)
                 && Objects.equals(this.compareOption, that.compareOption)
-                && extensionClassEquals(this.extension, that.extension)
                 && Objects.equals(this.and, that.and)
                 && Objects.equals(this.next, that.next);
     }
@@ -175,15 +174,8 @@ public class Condition<T> {
                 this.expectValue,
                 this.operator,
                 this.compareOption,
-                this.extension == null ? 0 : this.extension.getClass().hashCode(),
                 this.and,
                 this.next);
-    }
-
-    private static boolean extensionClassEquals(ConditionExtension<?> a, ConditionExtension<?> b) {
-        if (a == b) return true;
-        if (a == null || b == null) return false;
-        return a.getClass() == b.getClass();
     }
 
     /**
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
@@ -169,6 +169,14 @@ public final class ConditionEvaluators {
                     return compareNumbers(v, other) >= 0;
                 });
 
+        // Extension (custom logic delegated to ConditionExtension)
+        m.put(
+                ConditionOperator.EXTENSION,
+                (v, c, cfg) -> {
+                    ConditionExtension<Object> ext = (ConditionExtension<Object>) c.getExtension();
+                    return ext.evaluate(cfg, v);
+                });
+
         for (ConditionOperator op : ConditionOperator.values()) {
             if (!m.containsKey(op)) {
                 throw new IllegalStateException(
--- a/seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java
+++ b/seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java
@@ -420,10 +420,14 @@ public class MetadataExportCommand implements Command<MetadataExportCommandArgs>
         }
         ObjectNode node = mapper.createObjectNode();
         node.put("key", condition.getOption().key());
-        if (condition.getExpectValue() != null) {
-            node.put("expectValue", String.valueOf(condition.getExpectValue()));
-        }
         ConditionOperator op = condition.getOperator();
+        Object expectValue = condition.getExpectValue();
+        if (op == ConditionOperator.EXTENSION && condition.getExtension() != null) {
+            expectValue = condition.getExtension().description();
+        }
+        if (expectValue != null) {
+            node.put("expectValue", String.valueOf(expectValue));
+        }
         if (op != null) {
             node.put("conditionOperator", op.name());
             node.put("conditionOperatorCategory", op.getCategory().name());
--- a/seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/OptionRulesService.java
+++ b/seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/OptionRulesService.java
@@ -282,9 +282,13 @@ public class OptionRulesService extends BaseService {
                         : null;
         String conditionOperator = (op != null) ? op.name() : null;
         String conditionOperatorCategory = (op != null) ? op.getCategory().name() : null;
+        Object expectValue = condition.getExpectValue();
+        if (op == ConditionOperator.EXTENSION && condition.getExtension() != null) {
+            expectValue = condition.getExtension().description();
+        }
         return new OptionRuleResponse.ConditionNode(
                 toOptionMetadata(condition.getOption()),
-                condition.getExpectValue(),
+                expectValue,
                 compareOperatorSymbol,
                 compareOptionMeta,
                 conditionOperator,
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -441,7 +441,7 @@ static class ChildConfigValidator
 ```
 
 :::caution
-`ConditionExtension.evaluate()` runs during job submission and REST metadata queries. Implementations **must not** perform I/O (database connections, HTTP calls, file access). Keep the logic pure — validate structure and values only.
+`ConditionExtension.evaluate()` runs during job submission validation only. REST metadata queries only serialize `description()` and do not invoke `evaluate()`. Implementations should avoid I/O (database connections, HTTP calls, file access) and only validate structure and values.
 :::
 
 ## Why It Matters For Operators
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -409,7 +409,7 @@ Inline anonymous class:
 .optional(API_KEY_ENCODED, Conditions.extension(API_KEY_ENCODED,
         new ConditionExtension<String>() {
             @Override public String description() { return "must be Base64-encoded 'id:api_key'"; }
-            @Override public boolean evaluate(ReadonlyConfig cfg, String v) {
+            @Override public boolean evaluate(ReadonlyConfig cfg, String v) throws OptionValidationException {
                 try {
                     return new String(Base64.getDecoder().decode(v)).contains(":");
                 } catch (IllegalArgumentException e) {
@@ -430,7 +430,7 @@ static class TableConfigsValidator
     }
 
     @Override
-    public boolean evaluate(ReadonlyConfig config, List<Map<String, Object>> value) {
+    public boolean evaluate(ReadonlyConfig config, List<Map<String, Object>> value) throws OptionValidationException {
         if (value.isEmpty()) return false;
         Set<String> seen = new HashSet<>();
         for (Map<String, Object> entry : value) {
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -408,8 +408,14 @@ Inline anonymous class:
 ```java
 .optional(API_KEY_ENCODED, Conditions.extension(API_KEY_ENCODED,
         new ConditionExtension<String>() {
-            @Override public String description() { return "must be Base64-encoded 'id:api_key'"; }
-            @Override public boolean evaluate(ReadonlyConfig cfg, String v) throws OptionValidationException {
+            @Override
+            public String description() {
+                return "must be Base64-encoded 'id:api_key'";
+            }
+
+            @Override
+            public boolean evaluate(ReadonlyConfig cfg, String v)
+                    throws OptionValidationException {
                 try {
                     return new String(Base64.getDecoder().decode(v)).contains(":");
                 } catch (IllegalArgumentException e) {
@@ -431,7 +437,9 @@ static class TableConfigsValidator
 
     @Override
     public boolean evaluate(ReadonlyConfig config, List<Map<String, Object>> value) throws OptionValidationException {
-        if (value.isEmpty()) return false;
+        if (value.isEmpty()) {
+            return false;
+        }
         Set<String> seen = new HashSet<>();
         for (Map<String, Object> entry : value) {
             Object name = entry.get("table_name");
--- a/docs/en/architecture/configuration-and-option-system.md
+++ b/docs/en/architecture/configuration-and-option-system.md
@@ -131,6 +131,7 @@ Available operators (all accessed via the `Conditions` factory class):
 | Cross-field | `lessOrEqualField(option, other)` | value <= another option's value |
 | Cross-field | `greaterThanField(option, other)` | value > another option's value |
 | Cross-field | `greaterOrEqualField(option, other)` | value >= another option's value |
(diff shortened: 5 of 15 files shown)


nzw921rx:
@dybyte @davidzollo @yzeng1618 Can you help me review it? thank you very much.

DanielLeens (changes requested):
Thanks for working on this. I went through the full diff locally and traced both the runtime validation path and the existing metadata consumer paths. The core `EXTENSION` hook makes sense, but I still see two blockers before this is ready to merge.

# What This PR Solves
- User pain: the built-in `OptionRule` operators are not expressive enough for structural checks such as `List<Map>` child configs or nested cross-key validation.
- Fix approach: add `ConditionOperator.EXTENSION`, carry a `ConditionExtension<T>` inside `Condition`, evaluate it from `ConditionEvaluators`, and expose the description through the REST `option-rules` path.
- One-line summary: the runtime validation path is wired correctly, but the current head has not finished converging all existing metadata consumers on the new extension semantics.

# Runtime Chain Rechecked
```text
Job submission validation
  -> FactoryUtil.createAndPrepareSource()/createAndPrepareSink()/createAndPrepareTransform()
      -> ConfigValidator.validate(factory.optionRule())
          -> ConfigValidator.validate(Condition) [ConfigValidator.java:458-479]
              -> ConditionEvaluators.evaluate(cur, config)
                  -> EXTENSION delegates to ext.evaluate(cfg, value) [ConditionEvaluators.java:172-177]

REST metadata path
  -> GET /option-rules
      -> OptionRulesService.buildResponse()
      -> toConditionNode()
          -> EXTENSION writes description() into expectValue [OptionRulesService.java:285-297]

CLI metadata export path
  -> SeaTunnelMetadataExporter.main() [SeaTunnelMetadataExporter.java:27-40]
      -> MetadataExportCommand.exportCondition() [MetadataExportCommand.java:417-444]
          -> only writes expectValue when condition.getExpectValue() != null
          -> EXTENSION description is still dropped here
```

# Findings
Issue 1: the existing CLI metadata exporter still drops the extension description from the structured condition tree
- Location: `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java:417-444`
- Why this is a blocker:
  `OptionRulesService` now exposes the extension description on the REST path, but `SeaTunnelMetadataExporter` still serializes the old condition shape. That means one official metadata consumer gets the new semantics while another official metadata consumer still loses them.
- Better fix:
  mirror the same EXTENSION handling in `MetadataExportCommand.exportCondition()` that you already added in `OptionRulesService.toConditionNode()`, and add a focused metadata-export regression test for an extension-based rule.
- Severity: High

Issue 2: the `/option-rules` REST docs still describe the old condition-tree contract
- Location: `docs/en/engines/zeta/rest-api-v2.md:153-155`, `docs/zh/engines/zeta/rest-api-v2.md:151-153`
- Why this matters:
  the current implementation now emits `conditionOperator=EXTENSION`, `conditionOperatorCategory=EXTENSION`, `compareOperator=extension`, and uses `expectValue` for the human-readable extension description. The REST docs still say non-comparison conditions leave these fields `null`.
- Better fix:
  update both the English and Chinese REST API docs to explain the EXTENSION shape and add a minimal JSON example.
- Severity: Medium

# Test Coverage
- The runtime validation tests in `ConfigValidatorTest` / `OptionRuleTest` look good for the evaluation path itself.
- The missing coverage is on the metadata consumer side: there is still no regression test for REST/CLI export of an extension-based rule.
- Stability rating: Stable. The added tests are pure in-memory validation checks with no timing, network, or environment coupling.

# Merge Conclusion

1. Blocking items
- Issue 1: finish the existing CLI metadata-export path so extension-based rules are represented consistently across official metadata consumers.

- Issue 2: sync the REST API docs with the new EXTENSION condition-tree shape.

Overall, I think the runtime validation design is on the right track. The remaining gap is not the `ConditionExtension` evaluation itself, but the fact that the metadata consumer chain is still only partially updated on the current head.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head end to end, from the runtime validation path down to both official metadata consumers. The two blockers from my last round are fixed on the current head, so I don't see anything that should block merge now.

# What This PR Solves
- User pain: the built-in `OptionRule` operators are still not expressive enough for structural validation such as `List<Map>` children or nested cross-key checks.
- Fix approach: add `ConditionOperator.EXTENSION`, carry a `ConditionExtension<T>` inside `Condition`, evaluate it from `ConditionEvaluators`, and expose the rule description through both REST metadata and the CLI metadata exporter.
- One-line summary: the extension hook is now wired through the real validation path and the two official metadata consumer paths consistently.

# Runtime Chain Rechecked
```text
Job submission validation
  -> FactoryUtil.createAndPrepareSource()/Sink()/Transform()
      -> factory.optionRule()
          -> ConfigValidator.validate(Condition) [ConfigValidator.java:458-479]
              -> ConditionEvaluators.evaluate(cur, config)
                  -> EXTENSION delegates to ConditionExtension.evaluate(cfg, value)

REST metadata path
  -> GET /option-rules
      -> OptionRulesService.buildResponse()
          -> toConditionNode()
              -> for EXTENSION, expectValue = condition.getExtension().description()

CLI metadata export path
  -> SeaTunnelMetadataExporter.main()
      -> MetadataExportCommand.exportConnector()
          -> exportCondition()
              -> for EXTENSION, expectValue = condition.getExtension().description()
```

# Findings
I did not find any merge-blocking issue on the current head.

Issue 1: the Extension docs still describe the `evaluate()` trigger point a bit too broadly
- Location: `docs/en/architecture/configuration-and-option-system.md:443-444`, `docs/zh/architecture/configuration-and-option-system.md:443-444`
- Why this matters:
  the docs currently say `ConditionExtension.evaluate()` runs during both job submission and REST metadata queries. Looking at the actual code path, REST metadata only consumes `description()` through `OptionRulesService.toConditionNode()` and does not call `ConditionEvaluators.evaluate(...)`.
- Suggestion:
  tighten the wording to say that `evaluate()` runs during validation/job submission, while REST metadata queries only serialize the human-readable rule description.
- Severity: Medium

Issue 2: the new public extension entry points could use one more layer of API docs
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionExtension.java:22`, `seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Conditions.java:144-145`
- Why this matters:
  this is now a public extension surface for connector authors, so a short type-level Javadoc on `ConditionExtension` plus a method-level Javadoc on `Conditions.extension(...)` would make the contract easier to use correctly.
- Suggestion:
  add a concise API-level description for the interface and the factory method, especially around the pure-validation expectation and error-reporting guidance.
- Severity: Low

# Test Coverage
- The runtime coverage in `ConfigValidatorTest` / `OptionRuleTest` looks good for the validation path itself.
- `MetadataExportCommandTest` and `OptionRulesServiceTest` now cover the two metadata consumer paths that were missing before.
- Stability rating: Stable. The added tests are all in-memory unit tests with no timing, network, or environment coupling.

# Merge Conclusion
### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- Issue 1: tighten the wording in the Extension docs so it matches the actual runtime contract.
- Issue 2: add a bit more API-level Javadoc around the new public extension entry points.

Overall, I think this is in good shape now. The main runtime path works, the REST and CLI metadata views are aligned, and the remaining items are documentation polish rather than correctness blockers.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch against `dev`, including the runtime validation path and both official metadata-consumer paths. The newest head includes a CI-trigger commit, but after rechecking the full current diff I still do not see a code blocker.

# What This PR Solves
- User pain: the built-in `OptionRule` operators are not expressive enough for some real validation cases, such as `List<Map>` structural checks or nested cross-key validation.
- Fix approach: add `ConditionOperator.EXTENSION` plus `ConditionExtension<T>`, route it through the normal validation chain, and expose its human-readable description consistently through REST metadata and CLI metadata export.
- One-line summary: SeaTunnel now has a first-class pluggable validation hook, and the current head keeps the runtime path and both official metadata views aligned.

# Runtime Chain Rechecked
```text
Job submission validation
  -> FactoryUtil.createAndPrepareSource()/Sink()/Transform()
      -> ConfigValidator.validate(factory.optionRule())
          -> ConfigValidator.validate(Condition)
              -> ConditionEvaluators.evaluate(...)
                  -> EXTENSION -> ConditionExtension.evaluate(config, value)

REST metadata path
  -> OptionRulesService.toConditionNode() [OptionRulesService.java:272-297]
      -> for EXTENSION, serializes description() into expectValue

CLI metadata export path
  -> MetadataExportCommand.exportCondition() [MetadataExportCommand.java:417-445]
      -> for EXTENSION, serializes description() into expectValue
```

# Review Result

Key evidence I rechecked locally:
- `ConditionExtension.java:22-62` now documents the contract clearly, including that `evaluate()` runs during validation while metadata paths only serialize `description()`.
- `Conditions.java:142-145` exposes the extension hook cleanly.
- `OptionRulesService.java:272-297` and `MetadataExportCommand.java:417-445` now handle `EXTENSION` consistently.
- The metadata-consumer gap from my earlier round is closed on the latest head.

# Test Coverage
- The runtime validation coverage is good.
- The previously missing metadata-consumer coverage is now present.
- Stability rating: Stable.
  - The new tests are in-memory unit tests.
  - I do not see timing, environment, network, or shared-state flaky-test patterns here.

# Merge Conclusion
### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- None from my side.

Overall, this looks ready from a code-review perspective. The validation hook is wired correctly, the REST and CLI metadata outputs are aligned, and I do not see a protocol or compatibility blocker on the current head. The current `Build` is still queued, so merge should still wait for the check result, but I do not see a code-side reason to hold this revision.

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
```diff
@@ -169,6 +169,14 @@ private static Map<ConditionOperator, Evaluator> createRegistry() {
                     return compareNumbers(v, other) >= 0;
                 });
 
+        // Extension (custom logic delegated to ConditionExtension)
+        m.put(
+                ConditionOperator.EXTENSION,
+                (v, c, cfg) -> {
+                    ConditionExtension<Object> ext = (ConditionExtension<Object>) c.getExtension();
+                    return ext.evaluate(cfg, v);
```
yzeng1618, line 177: Add a try-catch block at the call site to catch all Throwables, wrap them into an OptionValidationException, and ensure the error message contains the context "Custom validation plugin execution failed".

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
```diff
@@ -169,6 +169,14 @@ private static Map<ConditionOperator, Evaluator> createRegistry() {
                     return compareNumbers(v, other) >= 0;
                 });
 
+        // Extension (custom logic delegated to ConditionExtension)
+        m.put(
+                ConditionOperator.EXTENSION,
+                (v, c, cfg) -> {
+                    ConditionExtension<Object> ext = (ConditionExtension<Object>) c.getExtension();
```
yzeng1618, line 176: The return value of getExtension() is forcibly cast to ConditionExtension<Object>. Creating mismatched Condition via unconventional methods like reflection will trigger a runtime ClassCastException and fail job submission. Type validation logic is recommended.

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Condition.java
```diff
@@ -219,6 +234,9 @@ private static String conditionToString(Condition<?> cond) {
         ConditionOperator op = cond.operator;
         String key = "'" + cond.option.key() + "'";
 
+        if (op == ConditionOperator.EXTENSION) {
+            return key + " " + cond.extension.description();
```
yzeng1618, line 238: If description() returns null, the concatenated string will be "option_key null", and NPE may occur in certain serialization scenarios.

seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java
```diff
@@ -420,10 +420,14 @@ private ObjectNode exportCondition(Condition<?> condition) {
...
         ObjectNode node = mapper.createObjectNode();
         node.put("key", condition.getOption().key());
-        if (condition.getExpectValue() != null) {
-            node.put("expectValue", String.valueOf(condition.getExpectValue()));
-        }
         ConditionOperator op = condition.getOperator();
+        Object expectValue = condition.getExpectValue();
+        if (op == ConditionOperator.EXTENSION && condition.getExtension() != null) {
+            expectValue = condition.getExtension().description();
+        }
```
yzeng1618, line 427: Code Duplication (OptionRulesService.java 286-288)

seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/OptionRulesService.java
```diff
@@ -282,9 +282,13 @@ private OptionRuleResponse.ConditionNode toConditionNode(Condition<?> condition)
                         : null;
         String conditionOperator = (op != null) ? op.name() : null;
         String conditionOperatorCategory = (op != null) ? op.getCategory().name() : null;
+        Object expectValue = condition.getExpectValue();
+        if (op == ConditionOperator.EXTENSION && condition.getExtension() != null) {
+            expectValue = condition.getExtension().description();
+        }
```
yzeng1618, line 288: ditto

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
nzw921rx: This is handled by the extension plugin itself, and exceptions are not encapsulated at the framework level

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
nzw921rx: No ClassCastException will be generated, and the Externalization operator will be created uniformly by `Conditions.extension()` without any type conversion issues

seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java
nzw921rx: sorry, I have a question here. In which scenarios does null have serialization issues?

seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/OptionRulesService.java
nzw921rx: same issue.

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Condition.java
nzw921rx: sorry, I still believe that these scenarios should not be covered by interfaces, as it is the responsibility of creators and reviews. We cannot refuse null string scenarios.

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
yzeng1618: These are mostly suggestions to facilitate troubleshooting.

seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java
yzeng1618: Code Duplication

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/Condition.java
DanielLeens: Thanks for calling this out. Rechecking the current contract, ConditionExtension.description() is now documented as non-null and both metadata consumers only serialize that descriptive string, so I would still treat a null return as a plugin-implementation bug rather than a framework-side blocker on this revision. If we later decide to harden against buggy extensions, we can follow up with a defensive null normalization separately.

seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java
DanielLeens: Makes sense as a cleanup suggestion. I do see the same small EXTENSION-to-description mapping pattern here and in OptionRulesService, but on the unchanged head I would treat that as non-blocking duplication rather than a correctness issue. If the author wants to refactor it afterward, a tiny shared helper would be fine.

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
DanielLeens: Thanks for raising this. Rechecking the current head, I still read the contract here the same way the new ConditionExtension Javadoc describes it: evaluate() may throw OptionValidationException for richer validation context, while other unchecked exceptions are allowed to propagate rather than being wrapped generically. So on the unchanged head I would treat the extra catch-and-wrap as an optional troubleshooting refinement, not a new blocker.

seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
DanielLeens: Thanks. On the current runtime path, ConditionOperator.EXTENSION conditions are created through Conditions.extension(...), which pairs the Option<T> and ConditionExtension<T> at construction time. Given that factory contract, I do not see a new blocker on the unchanged head here. If the framework later adds a looser deserialization or reflection-based construction path for Condition, then adding a defensive type guard would be worth revisiting.

seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/OptionRulesService.java
DanielLeens: Same take here: I agree it is the same pattern repeated, but I do not see a behavior bug from it on the current head. Fine as a small follow-up refactor if the author wants to reduce duplication, but I would not reopen the approved code path over this alone.

seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/MetadataExportCommand.java
nzw921rx: > Makes sense as a cleanup suggestion. I do see the same small EXTENSION-to-description mapping pattern here and in OptionRulesService, but on the unchanged head I would treat that as non-blocking duplication rather than a correctness issue. If the author wants to refactor it afterward, a tiny shared helper would be fine.

I agree with your point of view that there is indeed duplicate code between CLI and Rest API at present, but this should be the next step to unify a utility class to close the duplicate code on both sides, which is a good defense line.

yzeng1618 (approved):
There are minor details that do not affect functionality and can be optimized later.

davidzollo (approved):
+1

PR #11057, opened 2026-06-10
[Feature][Connector-V2][MySQL-CDC] Add snapshot-only startup mode

### LINKED ISSUE #11036: [Feature][Connector-V2][MySQL-CDC] Add snapshot-only startup mode
## Background
`MySQL-CDC` currently supports startup paths that either:
- read snapshot and continue into binlog streaming, or
- skip snapshot and start from a binlog-oriented position.

What is still missing is a first-class startup mode for a bounded bootstrap job: read snapshot data only, then stop cleanly.

This matters for one-time backfill, initial warehouse/table bootstrap, and controlled migration stages where users want a deterministic snapshot result without leaving a continuous CDC task running.

## Scope
Add a dedicated snapshot-only startup mode to `MySQL-CDC`.

This issue is only about bounded snapshot execution. It does **not** include GTID startup, skip-events/skip-rows trimming, or dynamic newly-added table capture.

## Suggested implementation approach
1. **Configuration surface**
- Extend `MySqlIncrementalSourceOptions.STARTUP_MODE` with a snapshot-only mode.
- Keep this as an explicit source option instead of forcing users to rely on opaque Debezium passthrough properties.
- Reject incompatible combinations such as snapshot-only together with stop-mode or binlog-specific startup offset options.

2. **Runtime behavior**
- Reuse the existing snapshot split planning and reading path.
- After all snapshot splits are finished, do **not** start the incremental/binlog reader.
- The source should report boundedness correctly so the job can finish naturally instead of hanging in an idle streaming state.

3. **Checkpoint / finish semantics**
- Snapshot-only jobs should still be able to checkpoint during snapshot if the current source path already supports it.
- Completion must be deterministic: once the enumerator confirms all snapshot work is done, the source should emit end-of-input and finish.

4. **Documentation and examples**
- Document exact semantics in both `docs/en` and `docs/zh`.
- Include at least one example showing that the mode is for bootstrap-only jobs and does not continue to consume binlog.

## Suggested config shape
Example target shape:
```hocon
startup.mode = "snapshot"
```

Exact naming can be refined during implementation, but the behavior must be explicit and connector-owned.

## Acceptance criteria
- `MySQL-CDC` accepts a snapshot-only startup mode.
- The mode reads configured snapshot data and exits without entering incremental/binlog streaming.
- The source exposes bounded completion behavior correctly.
- Validation fails fast for incompatible option combinations.
- Add at least one focused integration/E2E validation path.

## Non-goals
- GTID-based startup.
- skip-events / skip-rows trimming.
- Schema evolution policy work.

GabrielBBaldez:

Closes #11036.

Adds a `snapshot` startup mode to `MySQL-CDC`: a bounded bootstrap job that reads the snapshot of the captured tables and then finishes on its own, without entering the incremental/binlog phase. Useful for one-time backfill, initial warehouse/table bootstrap, and controlled migration stages.

### What changed

**Option surface (`MySqlIncrementalSourceOptions`)**
- `startup.mode` now accepts `snapshot` (choices: `initial`, `snapshot`, `earliest`, `latest`, `specific`, `timestamp`), with the description explaining the bounded semantics.

**Runtime (`connector-cdc-base`)**
- `StartupMode` gains a `SNAPSHOT` constant; `StartupConfig.getStartupOffset` treats it like `INITIAL` (the snapshot phase needs no stream offset).
- `HybridSplitAssigner` gains a `snapshotOnly` flag: the existing snapshot split planning/reading path is fully reused, but once the snapshot phase completes no incremental split is handed out and `waitingForCompletedSplits()` no longer considers the incremental assigner — so the enumerator's existing logic signals no-more-splits and the readers finish.
- `IncrementalSource#getBoundedness` reports `BOUNDED` in snapshot mode, letting the job run as `BATCH` and finish naturally instead of idling in a streaming state.
- Fail-fast validation at source creation for incompatible combinations: `startup.mode = snapshot` together with `stop.mode != never`, or with `startup.specific-offset.file` / `startup.specific-offset.pos` / `startup.timestamp`, is rejected with a clear message.

**Checkpoint / finish semantics**
- Snapshot-only jobs checkpoint through the unchanged `HybridPendingSplitsState` path, so checkpointing during the snapshot keeps working. On restore the assigner is rebuilt with the same snapshot-only flag (derived from config), so a restarted job resumes the snapshot phase instead of re-entering it or falling into streaming.

**Tests**
- `HybridSplitAssignerTest#testSnapshotOnlyFinishesAfterSnapshotPhase`: with a completed snapshot phase, the snapshot-only assigner returns no next split and is not waiting (job can finish), while the default hybrid behavior on the same state keeps waiting to hand out the incremental split.
- `MySqlIncrementalSourceFactoryTest#testSupportedStartUpModes`: asserts the supported startup modes, including `snapshot`.
- New e2e case `testMysqlCdcSnapshotOnlyStartupMode` (+ `mysqlcdc_snapshot_only.conf`, `BATCH` job): seeds the source table, runs the job synchronously and asserts it exits 0 on its own, asserts the sink equals the snapshot, then mutates the source after completion and asserts the sink is unchanged (no binlog consumption).

**Docs (EN + ZH)**
- `startup.mode` option row updated and a "Snapshot-only bootstrap" section added with a `BATCH` example and the incompatible-options note.

### Scope notes (matching the issue's non-goals)

- No GTID-based startup, no skip-events/skip-rows trimming, no dynamic newly-added table capture, no schema evolution policy changes.

### Verification

- `mvn install -pl connector-cdc-base,connector-cdc-mysql` — all module tests passing locally (JDK 11), including the new ones.
- `mvn test-compile -pl connector-cdc-mysql-e2e` — e2e module compiles; the new IT runs in CI (needs Docker).
- `mvn spotless:apply` clean.

### Check list

* [x] Code changed are covered with tests, or it does not need tests for reason
* [ ] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)


DanielLeens (changes requested):
Thanks for working on this. I re-reviewed the latest head from scratch, including the full diff, the CDC base call chain, and the new tests. The user need here is real, but I don’t think the current version is ready to merge yet.

- User pain point: some MySQL-CDC users want a one-time bounded snapshot job and do not want the source to stay in the streaming/binlog phase.
- Proposed fix: add `startup.mode = snapshot`, make the source bounded, and stop the hybrid assigner after the snapshot phase.
- In one sentence: the goal makes sense, but the current implementation short-circuits the existing consistency path, and this head also includes an unrelated MongoDB-CDC change set.

Actual runtime path I traced:

```text
MySQL-CDC job startup
  -> IncrementalSource.createEnumerator(...) [IncrementalSource.java:330-369]
      -> startup.mode == SNAPSHOT
      -> new HybridSplitAssigner(..., snapshotOnly = true)

Snapshot phase
  -> HybridSplitAssigner.getNext() hands out snapshot splits
  -> readers finish snapshot splits
  -> CompletedSnapshotSplitsReportEvent delivers snapshot split watermarks
  -> IncrementalSplitAssigner records those watermarks

Normal INITIAL semantics today
  -> IncrementalSplitAssigner.createIncrementalSplit(...) [IncrementalSplitAssigner.java:204-264]
      -> uses snapshot split low/high watermarks
      -> computes the changelog start offset
      -> closes the consistency window between snapshot scanning and later changes

Current SNAPSHOT implementation in this PR
  -> snapshot phase completes
  -> HybridSplitAssigner.getNext() hits the snapshotOnly branch [HybridSplitAssigner.java:117-120]
      -> returns Optional.empty()
      -> enumerator signals no more splits
      -> job finishes immediately
```

That is the key issue for me: in the current CDC architecture, the incremental split is not only “the streaming phase”; it is also part of how the source closes the snapshot consistency window. Skipping it entirely changes the semantics more than the PR description suggests.

Main findings:

1. This PR is not scoped to one problem anymore; it also carries an unrelated MongoDB-CDC feature.
- Evidence:
  `git log upstream/dev..head` on this PR contains both:
  - `01fcb9007b [Feature][Connector-V2][MySQL-CDC] Add snapshot-only startup mode`
  - `9900ec2558 [Feature][Connector-V2][MongoDB-CDC] Support latest-offset startup mode without initial snapshot`
- Why this is a blocker:
  The diff also includes `MongoDB-CDC` docs/options/tests, which makes review, CI attribution, and later rollback much harder than necessary.
- Suggested fix:
  Please split the MongoDB-CDC part back into its own PR and keep this PR focused on MySQL-CDC snapshot mode.

2. `snapshotOnly` currently skips the incremental split entirely, which breaks the existing consistent-snapshot closure path.
- Location:
  `seatunnel-connectors-v2/.../HybridSplitAssigner.java:117-120`
  related runtime path:
  `seatunnel-connectors-v2/.../IncrementalSplitAssigner.java:64-69, 218-255`
- Why this is a blocker:
  `IncrementalSplitAssigner` uses snapshot split low/high watermarks to compute the changelog start offset. That logic is part of how the current CDC runtime covers concurrent changes happening during snapshot scanning. With the new `snapshotOnly` branch, the job ends before that path ever runs.
- Risk:
  Updates/inserts/deletes happening while snapshot splits are being scanned can be lost from the final bounded result.
- Suggested fix:
  The safer design would be a bounded catch-up phase: keep the job finite, but still create a bounded incremental phase that consumes until the snapshot watermark boundary is closed.

3. The new E2E does not cover the critical concurrent-change window during snapshot execution.
- Location:
  `seatunnel-e2e/.../AbstractMysqlCDCITBase.java:186-210`
- Why this matters:
  The new test only proves:
  - data that exists before startup is copied
  - data changed after the job has already finished is not consumed
  It does not test what happens when changes arrive while the snapshot is still in progress, which is the exact gap introduced by issue 2.
- Severity:
  Medium from my side. I would still like this covered before we rely on the new mode.

Compatibility:
- Partially incompatible at the behavior-contract level for the new mode.
- Existing defaults are unchanged, but once users opt into `startup.mode = snapshot`, the current behavior is not equivalent to the existing consistent snapshot semantics.

Performance / side effects:
- Lower runtime cost is not the real question here.
- The real side effect is correctness: the job may finish earlier by dropping the catch-up path that currently protects snapshot consistency.

Tests:
- The new tests are structurally stable. I did not see obvious flaky-test patterns in the added UT/E2E code.
- But coverage is still incomplete for the most important path, so I don’t think the current tests are sufficient proof of correctness.

CI:
- The PR does not currently have a clean Build signal.
- The latest visible failing check is `Build`, with at least `unit-test (8, ubuntu-latest)` failing at `run all modules unit test`.
- I am not using that as my main blocker because the code-level blockers above are already sufficient, but the head is not green yet either.

Merge conclusion:
### Conclusion: do not merge in the current form

1. Blocking items
- Issue 1: split the unrelated MongoDB-CDC change set out of this PR.
- Issue 2: redesign the snapshot-only main path so it still closes the snapshot/change-log consistency window before finishing.

2. Non-blocking suggestions
- Issue 3: add coverage for concurrent updates during the snapshot phase, so the new mode has a real regression guard.

Overall, I agree the feature is worth having. The current version just needs a tighter PR scope and a stronger semantic guarantee on the bounded snapshot path before it is mergeable. Happy to re-review once you push an updated head.

GabrielBBaldez:
Thanks for the thorough review, Daniel — all three points are fair.

**1. Unrelated MongoDB-CDC commit:** done. I'd accidentally branched this off my MongoDB-CDC branch, so its commit rode along. Rebased onto `dev` — this PR now contains only the MySQL-CDC snapshot-mode commit.

**2. Consistency closure:** you're right, and this is the important one. Skipping the incremental phase entirely drops the catch-up that reconciles changes occurring *during* the snapshot scan, so concurrent writes on a busy table can be lost. I'll rework it so snapshot-only still runs a **bounded** incremental phase that closes the snapshot/changelog window before finishing, instead of short-circuiting it — most likely by routing through the existing bounded-stop machinery so the guarantee matches `initial` semantics rather than introducing a new code path.

**3. Concurrent-change coverage:** agreed — I'll add an e2e that injects changes while the snapshot is still in progress, so the mode has a real regression guard for exactly that window.

I've converted this to a draft while I rework it, and I'll ping you for re-review once the bounded catch-up and the concurrent-change test are in. Thanks again for catching the consistency gap before it shipped.

GabrielBBaldez:
Pushed the rework — thanks again, this is much better for it.

Instead of short-circuiting the incremental phase, `snapshot` mode is now implemented as **`initial` snapshot + a forced bounded `LATEST` stop**:

- `StartupMode.SNAPSHOT` resolves to a null startup offset (snapshot runs) and routes through `HybridSplitAssigner`, exactly like `initial`.
- `IncrementalSource` forces `StopConfig` to `LATEST` for snapshot mode. Since the incremental split is only created *after* the snapshot phase completes, `offsetFactory.latest()` resolves to the change-log position at snapshot completion. The existing bounded-stop machinery (the binlog reader already terminates at a non-`NEVER` stop offset) then runs the catch-up from the snapshot low watermark up to that boundary and finishes.

So **Issue 2** is resolved by reusing the exact catch-up path `initial` uses — the snapshot/change-log consistency window is closed by the same `IncrementalSplitAssigner.createIncrementalSplit` watermark logic, not a bespoke path. The only difference from `initial` is the bounded stop. (`getBoundedness` reports `BOUNDED` via that stop; `HybridSplitAssigner` is back to its original form.)

On **Issue 3**: because the during-snapshot reconciliation is now the same code as `initial` (already covered by `testMysqlCdcCheckDataE2e`), I kept the snapshot e2e deterministic — it asserts bounded completion, the snapshot is captured, and changes made after the boundary are not consumed — rather than adding a timing-dependent "mutate during the snapshot scan" test that tends to be flaky on the shared runners. Happy to add one if you'd prefer it; just wanted to avoid introducing flakiness.

**Issue 1** is also done — the MongoDB-CDC commit is split out, so this PR is MySQL-only now. Marked ready for re-review. 🙏

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch again, including the full diff, the MySQL CDC startup path, and the new snapshot-mode tests.

# What This PR Solves
- User pain: some MySQL CDC users want a one-time bounded snapshot job and do not want the source to stay in the streaming/binlog phase.
- Fix approach: add startup.mode = snapshot, make the source bounded, and stop after the snapshot phase.
- One-line summary: the earlier consistency-window blocker is largely closed on the latest head, but the new revision still has one correctness gap: snapshot mode silently accepts stop-position options that it will never honor.

# Runtime Chain Rechecked
```text
job submission validation
  -> IncrementalSource.validateAndGetBoundedness() [IncrementalSource.java:132-159]
      -> validates startup.mode-specific options

snapshot-only stop config
  -> IncrementalSource.getStopConfig() [IncrementalSource.java:245-255]
      -> for SNAPSHOT mode, hard-forces StopMode.LATEST

enumerator/runtime path
  -> HybridSplitAssigner / IncrementalSplitAssigner
      -> snapshot mode no longer uses user-provided stop offset/timestamp settings
      -> runtime always behaves as bounded snapshot-only, regardless of supplied stop.* options
```

# Findings
The earlier concern about the snapshot/binlog consistency boundary is much better on the latest head than it was before. But there is still one concrete contract bug left.

Issue 1: snapshot mode still accepts stop-position options that it silently ignores at runtime
- Location:
  - seatunnel-connectors-v2/.../IncrementalSource.java:132-159
  - seatunnel-connectors-v2/.../IncrementalSource.java:245-255
  - related stop-consumption logic in IncrementalSplitAssigner and MySqlBinlogFetchTask
- Why this is a blocker:
  on the current head, users can still provide stop.timestamp or stop.specific-offset.* together with startup.mode = snapshot. Validation does not reject that combination, but getStopConfig() hard-forces StopMode.LATEST for snapshot mode, so the supplied stop options are ignored silently.
- Risk:
  this is a bad user-facing contract because the config appears accepted but the runtime never honors it.
- Better fix:
  explicitly reject unsupported stop.* options in snapshot mode during validation, and document the restriction so the bounded-mode contract is unambiguous.
- Severity: High
- Already raised by others: No. This is the current blocker I still see after re-reviewing the latest head.

# Test Coverage / Stability
- The new snapshot-mode tests cover the intended bounded path better than before.
- Stability rating: Stable.
  - I did not find a new flaky-test pattern in the added coverage.
- Missing coverage:
  there still should be a negative validation test proving snapshot mode rejects unsupported stop.* options once Issue 1 is fixed.

# Merge Conclusion

1. Blocking items
- Issue 1: reject unsupported stop.* options in snapshot mode instead of silently ignoring them.

2. Suggested follow-up
- Add a focused validation regression test for that rejected configuration combination.

Overall, the feature direction makes sense and the latest head is much closer. The remaining blocker is now a concrete config-contract problem rather than the earlier broad consistency concern.

GabrielBBaldez:
Thanks @DanielLeens, good catch. Addressed in 681ed9e:

`validateSnapshotStartupConfig` now rejects `stop.timestamp` and `stop.specific-offset.*` when `startup.mode = snapshot`, instead of silently overriding them with `StopMode.LATEST` at runtime. The existing `stop.mode != NEVER` check already covered the mode flag — this closes the offset-options gap you pointed out.

Added `testSnapshotModeRejectsStopOptions` (asserts both the rejection and the error message). cdc-base and cdc-mysql tests pass locally.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the MySQL CDC startup validation path through the bounded snapshot runtime path again.

# What this PR solves
- User pain: some users want a one-time MySQL CDC snapshot job that finishes after the bounded snapshot phase.
- Fix approach: add `startup.mode = snapshot`, keep the source bounded, and validate away unsupported stop options.
- One-line summary: the earlier config-contract blocker is fixed on the current head, and I did not find a new source-level blocker.

# Runtime path rechecked
job config
  -> `IncrementalSource.validateAndGetBoundedness()`
  -> snapshot-mode validation rejects unsupported `stop.*` combinations
runtime
  -> bounded snapshot startup path
  -> no silent acceptance of ignored stop-position settings

The new follow-up commit closes the issue I raised last round: snapshot mode no longer silently accepts stop-position options that it will never honor.

# CI
The PR-level `Build` check is currently red, so I would still wait for the latest workflow result before merge.

1. Blocking items
- No new source-level blocker from me on the current head.
- Please let the latest `Build` finish cleanly before merge.

2. Suggested follow-up
- A focused negative validation test for unsupported snapshot+stop combinations is still nice to keep long-term, but it is not a new blocker from me.

Thanks for tightening the config contract here.

DanielLeens (approved):
Thanks for the update and for syncing the branch with the latest `dev`. I re-reviewed the latest head from the MySQL CDC startup validation path through the bounded snapshot runtime path again.

Runtime path I checked
```text
job config
  -> IncrementalSource.validateSnapshotStartupConfig(...)
      -> rejects unsupported startup/stop combinations
runtime
  -> getStopConfig()
      -> SNAPSHOT forces bounded catch-up to LATEST boundary
  -> getBoundedness()
      -> bounded source lifecycle
```

- The earlier config-contract blocker is still closed on the current head: snapshot mode no longer silently accepts stop-position settings that it will never honor.
- The docs also now describe the mode as a bounded snapshot plus catch-up, which matches the actual runtime semantics much better than the earlier wording.
- I did not find a new source-level blocker on this revision. The new tests look structurally stable from my side, although a future regression test around the concurrent-change window would still be nice to keep long term.

Merge conclusion
- No new blocker from my side on the current head.
- This can merge once the latest Build / CI finishes cleanly.

DanielLeens:

- base branch: `dev`
- head SHA: `aa9c78066c340034adba3ffcd66df72605a4fdaa`
- compare status: `diverged`
- ahead_by: `3`
- behind_by: `8`
- CI / merge gate snapshot: `Build CANCELLED`

DanielLeens:
Hi @GabrielBBaldez, thank you for working on snapshot-only startup mode for MySQL CDC. This is a useful operational mode, and your PR contributed to the discussion around how snapshot completion and connector shutdown should behave.

The common snapshot-only foundation has now been merged through #11380, and the actively maintained MySQL implementation is continuing in #11487 with restore, lifecycle, and E2E coverage against the latest code. Keeping this older branch open would split review effort across two implementations of the same feature.

We are closing this PR as superseded so the community can focus on the current implementation. Thank you again for investing in the feature; the work and feedback here remain part of the path that moved it forward.

GabrielBBaldez:
Okay!😁

PR #11063, opened 2026-06-11
[Fix][Docs] Fix zh user page link

zhangshenghang:
## Purpose

Fix the Chinese docs About page user link so it navigates to the Chinese user page instead of the English user page.

## Changes

- Update the zh About page user link from `https://seatunnel.apache.org/user` to `https://seatunnel.apache.org/zh-CN/user`.

## Verification

- `curl -I -L https://seatunnel.apache.org/zh-CN/user` returns 200.
- `./mvnw spotless:apply` passed.
- `git diff --check` passed.
- `./mvnw -q -DskipTests verify` was not run to completion per requester instruction.


davidzollo (approved):
+1
Good job

PR #11068, opened 2026-06-11
[Fix][Connector-V2] Fix Cassandra null timestamp conversion

### LINKED ISSUE #9849: [Bug] [connector-cassandra] when field in source(cassancra) is timestamp type and data is null, will cause error.
### Search before asking

### What happened

source table in cassandra, when the field is timestamp type, and the data is null ,will cause nullpointer exception.

### SeaTunnel Version

2.3.3 and dev

### SeaTunnel Config

```conf
test1.config's content as below:


env {
    execution.parallelism = 1
    job.mode = "BATCH"
    checkpoint.interval = 10000
}
source {
    Cassandra {
        host = "192.168.0.100:9042"
        keyspace = "store"
        username = "admin"
        password = "123456"
        datacenter = "datacenter1"
        cql = "select userid,item_count,last_update_timestamp from shopping_cart"
    }
}

sink {
    doris {
        fenodes = "192.168.0.101:8030"
        username = "root"
        password = "123456"
        database = "cassandradb"
        table = "shopping_cart"
        table.identifier = "cassandradb.shopping_cart"
        sink.label-prefix = "seatunnel_offline"
        sink.enable-2pc = "false"
        doris.config = {
            format = "json",
            read_json_by_line = "true"
        }
        sink.properties.columns = "userid,item_count,last_update_timestamp"
    }
}
```

### Running Command

```shell
bin/seatunnel.sh --config test1.confg -e local
```

### Error Exception

```log
Exception in thread "main" org.apache.seatunnel.core.starter.exception.CommandExecuteException: SeaTunnel job executed failed
        at org.apache.seatunnel.core.starter.seatunnel.command.ClientExecuteCommand.execute(ClientExecuteCommand.java:191)
        at org.apache.seatunnel.core.starter.SeaTunnel.run(SeaTunnel.java:40)
        at org.apache.seatunnel.core.starter.seatunnel.SeaTunnelClient.main(SeaTunnelClient.java:34)
Caused by: org.apache.seatunnel.engine.common.exception.SeaTunnelEngineException: java.lang.NullPointerException
        at java.util.Objects.requireNonNull(Objects.java:203)
        at org.apache.seatunnel.connectors.seatunnel.cassandra.util.TypeConvertUtil.buildSeaTunnelRow(TypeConvertUtil.java:185)
        at org.apache.seatunnel.connectors.seatunnel.cassandra.source.CassandraSourceReader.lambda$pollNext$0(CassandraSourceReader.java:73)
        at java.lang.Iterable.forEach(Iterable.java:75)
        at org.apache.seatunnel.connectors.seatunnel.cassandra.source.CassandraSourceReader.pollNext(CassandraSourceReader.java:73)
        at org.apache.seatunnel.engine.server.task.flow.SourceFlowLifeCycle.collect(SourceFlowLifeCycle.java:150)
        at org.apache.seatunnel.engine.server.task.SourceSeaTunnelTask.collect(SourceSeaTunnelTask.java:95)
        at org.apache.seatunnel.engine.server.task.SeaTunnelTask.stateProcess(SeaTunnelTask.java:168)
        at org.apache.seatunnel.engine.server.task.SourceSeaTunnelTask.call(SourceSeaTunnelTask.java:100)
        at org.apache.seatunnel.engine.server.TaskExecutionService$BlockingWorker.run(TaskExecutionService.java:613)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:750)

        at org.apache.seatunnel.engine.client.job.ClientJobProxy.waitForJobComplete(ClientJobProxy.java:122)
        at org.apache.seatunnel.core.starter.seatunnel.command.ClientExecuteCommand.execute(ClientExecuteCommand.java:184)
        ... 2 more
```

_No response_

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:

Fixes #9849.

Cassandra `timestamp` columns can be null. The current conversion path calls `Objects.requireNonNull(row.getInstant(i))`, so a null timestamp from Cassandra fails with `NullPointerException` before the row reaches downstream connectors.

This patch keeps null timestamp values as null, and only converts non-null `Instant` values to `LocalDateTime`.

No. It only fixes null handling for Cassandra timestamp columns.

- `./mvnw -pl seatunnel-connectors-v2/connector-cassandra -Dtest=TypeConvertUtilTest#testBuildSeaTunnelRowKeepsNullTimestamp test`
- `./mvnw -pl seatunnel-connectors-v2/connector-cassandra test`
- `./mvnw spotless:apply`

### DIFF
--- a/seatunnel-connectors-v2/connector-cassandra/src/main/java/org/apache/seatunnel/connectors/seatunnel/cassandra/util/TypeConvertUtil.java
+++ b/seatunnel-connectors-v2/connector-cassandra/src/main/java/org/apache/seatunnel/connectors/seatunnel/cassandra/util/TypeConvertUtil.java
@@ -44,6 +44,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.sql.Timestamp;
+import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
@@ -181,9 +182,8 @@ public class TypeConvertUtil {
                     fields[i] = row.getLocalDate(i);
                     break;
                 case ProtocolConstants.DataType.TIMESTAMP:
-                    fields[i] =
-                            Timestamp.from(Objects.requireNonNull(row.getInstant(i)))
-                                    .toLocalDateTime();
+                    Instant instant = row.getInstant(i);
+                    fields[i] = instant == null ? null : Timestamp.from(instant).toLocalDateTime();
                     break;
                 case ProtocolConstants.DataType.BLOB:
                     fields[i] =
--- a/seatunnel-connectors-v2/connector-cassandra/src/test/java/org/apache/seatunnel/connectors/seatunnel/cassandra/util/TypeConvertUtilTest.java
+++ b/seatunnel-connectors-v2/connector-cassandra/src/test/java/org/apache/seatunnel/connectors/seatunnel/cassandra/util/TypeConvertUtilTest.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.connectors.seatunnel.cassandra.util;
+
+import org.apache.seatunnel.api.table.type.SeaTunnelRow;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import com.datastax.oss.driver.api.core.cql.ColumnDefinition;
+import com.datastax.oss.driver.api.core.cql.ColumnDefinitions;
+import com.datastax.oss.driver.api.core.cql.Row;
+import com.datastax.oss.driver.api.core.type.DataTypes;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class TypeConvertUtilTest {
+
+    @Test
+    void testBuildSeaTunnelRowKeepsNullTimestamp() {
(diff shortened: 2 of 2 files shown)


DanielLeens (approved):
Thanks for the fix here. I reviewed the full current diff against `dev`, traced the real Cassandra source conversion path, and checked the new regression test plus the current CI signal.

# What This PR Solves
- User pain: Cassandra `timestamp` columns can legitimately be `null`, but the old conversion path treated them as always present and could throw `NullPointerException` before the row even reached downstream processing.
- Fix approach: keep `null` timestamp values as `null`, and only convert non-null `Instant` values into SeaTunnel `LocalDateTime` values.
- One-line summary: this is a focused null-handling fix on the real Cassandra read path, so rows with nullable timestamps no longer fail during conversion.

Simple example: if a Cassandra row has `created_at = null`, the old code would fail inside `buildSeaTunnelRow(...)` because of `Objects.requireNonNull(row.getInstant(i))`. After this patch, that field stays `null` and the row can continue downstream.

# Runtime chain checked
```text
Cassandra source read
  -> CassandraSourceReader.pollNext(...)
      -> fetch Cassandra Row
      -> TypeConvertUtil.buildSeaTunnelRow(row)
          -> per-column conversion
          -> TIMESTAMP column uses row.getInstant(i)
              -> old behavior: require non-null, then convert
              -> new behavior: keep null, otherwise convert to LocalDateTime
      -> emit SeaTunnelRow to downstream
```

This is the normal main path rather than an edge-only branch, so the fix will be hit whenever a real Cassandra row contains a nullable timestamp.

# Review result

Key evidence I checked:
- `TypeConvertUtil.java:181-186` removes the incorrect non-null assumption only for the `TIMESTAMP` branch.
- The non-null conversion behavior is otherwise unchanged.
- `TypeConvertUtilTest.java:32-49` adds a focused regression test for the exact failure mode this PR is fixing.
- Current PR checks are green, including `Build`.

- Fully compatible.
- No API, config, default-value, protocol, or serialization contract changes.
- CPU / memory impact is negligible; this is just one null check on a field conversion path.
- No new concurrency, retry, or resource-lifecycle concerns introduced here.

# Test coverage and stability
- The new unit test covers the key regression target: `null` Cassandra timestamps.
- Stability rating: Stable.
  - The test is fully in-memory with mocks.
  - I do not see sleep-based timing, port usage, shared global state, or other flaky-test patterns.

# Merge Conclusion
### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- None from my side.

Overall, this is a precise and useful bug fix. The change hits the real failure point, keeps the behavior minimal, and the regression test is targeted in the right place.

davidzollo (approved):
+1

PR #10724, opened 2026-04-08
[SEATUNNEL-10685] prevent timestamp_ntz from being saved as timestamp_ltz

doyong365:
Purpose of this pull request
Fixes #10685
JDBC connector could not distinguish between `TIMESTAMP` (NTZ, No Time Zone) and `TIMESTAMP_TZ` (LTZ, Local Time Zone) types when reading from databases. Previously, both timezone-naive and timezone-aware timestamp columns were mapped to SeaTunnel's internal `TIMESTAMP` type, causing timezone information to be lost when writing to timezone-aware sinks such as Iceberg. This PR explicitly separates them by fixing the `TypeConverter` and `JdbcRowConverter` for MySQL, PostgreSQL, Oracle, and SQL Server:

| DB Column Type | Before (wrong) | After (correct) |
|---|---|---|
| MySQL `DATETIME` | `TIMESTAMP` | `TIMESTAMP` (NTZ) ✅ |
| MySQL `TIMESTAMP` | `TIMESTAMP` | `TIMESTAMP_TZ` (LTZ) ✅ |
| PostgreSQL `timestamp` | `TIMESTAMP` | `TIMESTAMP` (NTZ) ✅ |
| PostgreSQL `timestamptz` | `TIMESTAMP` | `TIMESTAMP_TZ` (LTZ) ✅ |
| Oracle `TIMESTAMP` | `TIMESTAMP` | `TIMESTAMP` (NTZ) ✅ |
| Oracle `TIMESTAMP WITH LOCAL TIME ZONE` | `TIMESTAMP` | `TIMESTAMP_TZ` (LTZ) ✅ |
| SQL Server `datetime` | `TIMESTAMP` | `TIMESTAMP` (NTZ) ✅ |
| SQL Server `datetimeoffset` | `TIMESTAMP` | `TIMESTAMP_TZ` (LTZ) ✅ |

Also fixes `IcebergTypeMapper` and `DefaultDeserializer` in `connector-iceberg` to correctly map `TIMESTAMP_TZ` → Iceberg `timestamptz` (withZone).

Does this PR introduce _any_ user-facing change?
Yes.

**Previous behavior:**
Both `DATETIME` (NTZ) and `TIMESTAMP` (LTZ) columns from MySQL were read as SeaTunnel's `TIMESTAMP` type. When written to Iceberg, both were stored as `timestamp` (without timezone), losing timezone semantics.

**New behavior:**
[...]

PR #11069, opened 2026-06-11
[Bug][CDC][Postgres] Support TIMESTAMP_TZ type in Debezium row converter

### LINKED ISSUE #11005: [Bug][CDC][Postgres] Postgres-CDC fails with UnsupportedOperationException: Unsupported type TIMESTAMP_TZ for TIMESTAMP WITH TIME ZONE columns
### Search before asking

### What happened

When using Postgres-CDC on a table containing a PostgreSQL TIMESTAMP WITH TIME ZONE (TIMESTAMPTZ) column, SeaTunnel fails during source initialization.

The job never starts processing records and fails while building the CDC deserialization schema.

Actual behavior:

Source initialization fails with:

java.lang.UnsupportedOperationException:
Unsupported type: TIMESTAMP_TZ
Stack Trace
Caused by: java.lang.UnsupportedOperationException: Unsupported type: TIMESTAMP_TZ
    at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializationConverters.createNotNullConverter(SeaTunnelRowDebeziumDeserializationConverters.java:185)
    at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializationConverters.createConverter(SeaTunnelRowDebeziumDeserializationConverters.java:119)
    at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializeSchema.createTableRowConverters(SeaTunnelRowDebeziumDeserializeSchema.java:349)
    at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializeSchema.<init>(SeaTunnelRowDebeziumDeserializeSchema.java:94)
    at org.apache.seatunnel.connectors.seatunnel.cdc.postgres.source.PostgresIncrementalSource.createDebeziumDeserializationSchema(PostgresIncrementalSource.java:113)

### SeaTunnel Version

2.3.13

### SeaTunnel Config

```conf
env {
  parallelism = 1
  job.mode = "STREAMING"
}

source {
  Postgres-CDC {
    plugin_output = "source_data"
    base-url = "jdbc:postgresql://<host>:5432/<database>"
    username = "<username>"
    password = "<password>"

    database-names = ["postgres"]
    schema-names = ["public"]
    table-names = ["public.cdc_timestamp_test"]

    slot.name = "seatunnel_timestamp_test"
  }
}

sink {
  Console {}
}
```

### Running Command

```shell
cd {seatunnel_path}/ 
./bin/seatunnel.sh --config ./config/new_config/postgres_cdc_test -n test1
```

### Error Exception

```log
Caused by: java.lang.UnsupportedOperationException: Unsupported type: TIMESTAMP_TZ
        at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializationConverters.createNotNullConverter(SeaTunnelRowDebeziumDeserializationConverters.java:185)
        at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializationConverters.createConverter(SeaTunnelRowDebeziumDeserializationConverters.java:119)
        at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializationConverters.lambda$new$0(SeaTunnelRowDebeziumDeserializationConverters.java:77)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
        at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
        at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)
        at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializationConverters.<init>(SeaTunnelRowDebeziumDeserializationConverters.java:79)
        at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializeSchema.createTableRowConverters(SeaTunnelRowDebeziumDeserializeSchema.java:349)
        at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializeSchema.<init>(SeaTunnelRowDebeziumDeserializeSchema.java:94)
        at org.apache.seatunnel.connectors.cdc.debezium.row.SeaTunnelRowDebeziumDeserializeSchema$Builder.build(SeaTunnelRowDebeziumDeserializeSchema.java:374)
        at org.apache.seatunnel.connectors.seatunnel.cdc.postgres.source.PostgresIncrementalSource.createDebeziumDeserializationSchema(PostgresIncrementalSource.java:113)
        at org.apache.seatunnel.connectors.cdc.base.source.IncrementalSource.<init>(IncrementalSource.java:127)
        at org.apache.seatunnel.connectors.seatunnel.cdc.postgres.source.PostgresIncrementalSource.<init>(PostgresIncrementalSource.java:63)
        at org.apache.seatunnel.connectors.seatunnel.cdc.postgres.source.PostgresIncrementalSourceFactory.lambda$createSource$1(PostgresIncrementalSourceFactory.java:110)
        at org.apache.seatunnel.api.table.factory.FactoryUtil.createAndPrepareSource(FactoryUtil.java:180)
        at org.apache.seatunnel.api.table.factory.FactoryUtil.restoreAndPrepareSource(FactoryUtil.java:136)
        ... 8 more
```

_No response_

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

lazyfetch:

Resolves #11005 

**Bug:**
When using the Postgres-CDC connector to sync a table containing a `TIMESTAMP WITH TIME ZONE` (`TIMESTAMPTZ`) column, SeaTunnel throws an `UnsupportedOperationException: Unsupported type: TIMESTAMP_TZ` during source initialization. This occurs because `SeaTunnelRowDebeziumDeserializationConverters#createNotNullConverter` lacks a branch to handle `TIMESTAMP_TZ`.

**Fix:**
* Added the `SqlType.TIMESTAMP_TZ` mapping case to `createNotNullConverter`.
* Implemented `convertToTimestampTz()` to correctly parse the ISO-8601 string natively emitted by Debezium into a Java `OffsetDateTime`.

Yes. 

**Previous Behavior:** Syncing a PostgreSQL table with a `TIMESTAMPTZ` column would result in an immediate job crash (`UnsupportedOperationException`).
**New Behavior:** `TIMESTAMPTZ` columns are successfully parsed and synced as `OffsetDateTime` types without throwing errors. No new `.conf` configuration changes are required by the user.

* **Unit Tests Added:** Added `testTimestampTzConversion` to `SeaTunnelRowDebeziumDeserializationConvertersTest.java` to explicitly mock the Debezium string emission for `TIMESTAMPTZ` and verify it correctly parses into a Java `OffsetDateTime`.
* **Local Verification:** Compiled and tested locally using `mvn clean install` to ensure no compilation issues or regressions in the data type formatting suite.

### Check list


DanielLeens (commented):
Thanks for the fix. I reviewed the latest head from the PostgreSQL CDC Debezium conversion path down to the added TIMESTAMP_TZ support.

# What This PR Solves
- User pain: PostgreSQL CDC rows carrying TIMESTAMP_TZ values are not converted correctly in the Debezium row-converter path.
- Fix approach: add explicit TIMESTAMP_TZ handling in the Debezium row converter so the value lands in the SeaTunnel row with the right logical type.
- One-line summary: I did not find a new code-side blocker in the TIMESTAMP_TZ conversion path on the latest head; the remaining blocker is the current red Build signal.

# Runtime Chain Rechecked
```text
PostgreSQL CDC path
  -> Debezium change event
      -> Postgres row converter
          -> detect TIMESTAMP_TZ column
          -> convert to the matching SeaTunnel logical value type
      -> downstream SeaTunnel row consumer sees the converted timestamp-with-time-zone value
```

# Findings
I did not find a new runtime-path blocker in the current TIMESTAMP_TZ conversion logic.

Issue 1: the top-level Build is still red, and the visible failures look broad rather than Postgres TIMESTAMP_TZ-specific
- Location: GitHub Build check `80863855909`
- Evidence I confirmed:
  the current run shows failures in updated-modules-integration-test-part-* and paimon-connector-it jobs rather than a targeted PostgreSQL CDC converter regression.
- Why this matters:
  the code-side fix itself looks coherent, but the PR still does not have a clean merge gate while the top-level Build is red.
- Better fix:
  get the current Build green on the latest head. If a targeted Postgres CDC failure remains afterward, that would be the point to revisit the converter logic again.
- Severity: High

# Test Coverage / Stability
- The added converter coverage is directionally correct for the changed path.
- Stability rating: Stable.
  - I did not see timing, network, or environment-sensitive test patterns in the change.

# Merge Conclusion

1. Blocking items
- Issue 1: get the current Build green on this head.

2. Suggested follow-up
- None from the Postgres CDC converter path I reviewed.

Overall, the TIMESTAMP_TZ fix itself looks reasonable. The remaining blocker is the CI gate, not a new converter-side correctness issue.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the PostgreSQL CDC Debezium conversion path down to the new `TIMESTAMP_TZ` branch.

Runtime path rechecked:
```text
Debezium source value
  -> SeaTunnelRowDebeziumDeserializationConverters.createConverter(...)
      -> case TIMESTAMP_TZ
      -> convertToTimestampTz()
  -> SeaTunnelRow field becomes OffsetDateTime
```

The current conversion logic is aligned with the SeaTunnel type system here: `TIMESTAMP_TZ` now produces `OffsetDateTime` instead of falling through to an unsupported path, and the added regression test locks the expected row value down with a real Debezium-style string payload. I did not find a new source-level blocker on the current head.

The remaining merge gate is the PR-level `Build` check, which is still red right now.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the Debezium row-conversion path instead of only checking the new unit test.

What this PR fixes:
- User pain: a PostgreSQL CDC table with `TIMESTAMPTZ` currently fails during source-side conversion with `Unsupported type: TIMESTAMP_TZ`, so the job cannot even start reading that column type.
- Fix approach: add the missing `TIMESTAMP_TZ` branch in `SeaTunnelRowDebeziumDeserializationConverters#createNotNullConverter()` and parse the Debezium ISO-8601 payload into `OffsetDateTime`.
- In plain words: SeaTunnel now understands the timezone-aware timestamp that Debezium is already sending.

Runtime path I checked:
```text
Postgres CDC event
  -> Debezium schema exposes TIMESTAMP_TZ
  -> SeaTunnelRowDebeziumDeserializationConverters.createNotNullConverter()
      -> convertToTimestampTz()
  -> SeaTunnel row carries OffsetDateTime for downstream stages
```

I did not find a blocking issue in the current revision. The change matches SeaTunnel's existing `TIMESTAMP_TZ` / `OffsetDateTime` type contract, and the added test covers the normal Debezium string payload that was previously crashing. From Daniel's side this can move forward.

lazyfetch:
Hi @DanielLeens,

Thanks so much for the detailed review and the approval!

I just synced with the latest dev branch to get a fresh build
Everything is looking good, but one of the broad integration tests (SeaTunnel : E2E : Connector V2 : JDBC : Schema Evolution) failed on the latest run. 

Given the scope of the Debezium TIMESTAMP_TZ changes, this looks like an unrelated flaky CI test:
https://github.com/lazyfetch/seatunnel/actions/runs/27789387660/job/82235692598

Since there are no source-level blockers, let me know if i can do anything for this

DanielLeens:
Hi @lazyfetch,

I rechecked the current CI state on the same head. The only red gate is still `Build`, and the concrete red lane on the linked fork run is `updated-modules-integration-test-part-1`, where the repo-owned failure is in the JDBC schema-evolution suite: `SqlServerSchemaChangeIT.testMysqlCdcWithSchemaEvolutionCaseExactlyOnce` timing out inside `AbstractSchemaChangeBaseIT.assertSchemaEvolution(...)`.

That lane is outside the PostgreSQL CDC `TIMESTAMP_TZ` conversion path touched by this PR. Since the head is unchanged from the version I already reviewed, I still do not see a new source-side blocker in the Debezium row-converter chain here. From Daniel's side this stays in the "code looks okay, waiting for a clean broad CI signal / maintainer pickup" bucket.

If someone wants the fastest next step, I would rerun or isolate that broad JDBC schema-evolution lane rather than changing the PostgreSQL CDC conversion code again.

DanielLeens:
Hi @lazyfetch, thanks again for contributing this fix and for patiently following up on the CI state. I rechecked this against the current  branch today before trying to resolve the conflict.

The PostgreSQL CDC  capability is now already covered on  by the later merged #10724. The current implementation has the  branch wired to  and includes broader parsing coverage for ISO offsets, IANA zone ids, space-separated datetime strings, short hour-only offsets, UTC literals, plus the invalid-format path. That is a superset of the older implementation in this PR.

Because this PR is now conflicting with that newer implementation, merging it would no longer be the right rescue path and could accidentally regress the current converter behavior. I am closing this one as already covered by the newer merged implementation, not because the original direction was wrong. The contribution was useful and helped push this area forward. Please keep contributing when you spot CDC type gaps like this.

DanielLeens:
Correction to my previous comment: the shell stripped the inline code spans while I was posting it, so the technical references were rendered incomplete.

The intended evidence is:
- Current `dev` already covers PostgreSQL CDC `TIMESTAMP_TZ` through the later merged #10724.
- The current `dev` implementation wires `TIMESTAMP_TZ` to `convertToTimestampTz(serverTimeZone)`.
- The current tests cover ISO offsets, IANA zone ids, space-separated datetime strings, short hour-only offsets, UTC literals, and invalid-format handling.

So the conclusion is unchanged: this PR was a useful contribution, but the current branch has a newer and broader implementation now. Because this PR conflicts with that newer implementation, merging it would not be the right rescue path anymore.

DanielLeens:

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `29c3f664f70b` from scratch.

# What this PR fixes
- User pain: Syncing a Postgres table with a `TIMESTAMPTZ` column immediately crashes the CDC job with `UnsupportedOperationException: Unsupported type: TIMESTAMP_TZ` because the Debezium row converter has no branch for that SqlType.
- Fix approach: Adds a `TIMESTAMP_TZ` case in `createNotNullConverter` and a new `convertToTimestampTz()` converter that parses the ISO-8601 string Debezium emits into a Java `OffsetDateTime`, plus a unit test covering the happy path.
- One-line summary: The fix is minimal and functionally correct for the standard Debezium string emission, but the converter is brittle (unhandled `DateTimeParseException`, String-only input, no schema-aware error context), the test mocks the type with an ad-hoc anonymous `SeaTunnelDataType` instead of the real TIMESTAMP_TZ type, and connector docs plus small style issues (fully-qualified `java.time.OffsetDateTime`, gratuitous license-header edit) still need attention before merge.

# Runtime chain I rechecked
```text
SeaTunnelRowDebeziumDeserializeSchema.deserialize(SourceRecord, Collector)
  -> SeaTunnelRowDebeziumDeserializationConverters.convert(SourceRecord, Struct, Schema)  SeaTunnelRowDebeziumDeserializationConverters.java (row loop over field converters)
    -> createConverter(SeaTunnelDataType) -> wrapIntoNullableConverter(createNotNullConverter(...))
      -> createNotNullConverter() switch on SqlType: new case TIMESTAMP_TZ  SeaTunnelRowDebeziumDeserializationConverters.java:168-169
        -> convertToTimestampTz()  SeaTunnelRowDebeziumDeserializationConverters.java:459-475
          -> java.time.OffsetDateTime.parse((String) dbzObj)  SeaTunnelRowDebeziumDeserializationConverters.java:466 (strict ISO_OFFSET_DATE_TIME; DateTimeParseException propagates uncaught)
          -> else throw IllegalArgumentException for non-String dbzObj  SeaTunnelRowDebeziumDeserializationConverters.java:468-472
    -> SeaTunnelRow.setField(i, OffsetDateTime) consumed downstream by sinks expecting SqlType.TIMESTAMP_TZ
```

# Findings
**Issue 1: Unhandled DateTimeParseException in convertToTimestampTz() — any string that is not strict ISO_OFFSET_DATE_TIME crashes the job with no context about the column/value**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConverters.java:466`
- Why it matters: - Root cause: `java.time.OffsetDateTime.parse((String) dbzObj)` uses the strict `ISO_OFFSET_DATE_TIME` formatter and throws an unchecked `DateTimeParseException` for anything it cannot parse. Debezium's `io.debezium.time.ZonedTimestamp` values are usually compatible, but edge cases exist: Postgres `infinity`/`-infinity` timestamptz values, BC dates (Debezium can emit era-suffixed or out-of-range strings), and other dialects that emit `+00` style offsets.
- Impact: instead of the informative `IllegalArgumentException` this converter builds for non-string inputs, a parse failure escapes as a bare `DateTimeParseException` that kills the reader thread on Zeta/Flink/Spark with no field name or value in the message, making it very hard for users to diagnose which row/column broke the sync — this is exactly the happy-path-only pattern we try to avoid.
- Fix: wrap the parse in try/catch (`DateTimeParseException`) and rethrow the same `IllegalArgumentException` with the offending value included; prefer parsing with `io.debezium.time.ZonedTimestamp.FORMATTER` (or falling back to `ISO_ZONED_DATE_TIME`) so the converter accepts everything Debezium can actually emit.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConverters.java:465  if (dbzObj instanceof String) {`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConverters.java:466  return java.time.OffsetDateTime.parse((String) dbzObj);`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConverters.java:468-472  throw new IllegalArgumentException("Unable to convert to OffsetDateTime from unexpected value '" + dbzObj + "' of type " + (dbzObj != null ? dbzObj.getClass().getName() : "null"));`
- Suggested fix: `OffsetDateTime.parse` throws an unchecked `DateTimeParseException` for anything that isn't strict `ISO_OFFSET_DATE_TIME` (e.g. Postgres `infinity` timestamptz, BC/era-suffixed values, `+00`-style offsets from other dialects). That exception escapes this converter with no field/value context and crashes the job. Please wrap the parse in try/catch and rethrow the same descriptive `IllegalArgumentException` including the raw value, and consider parsing via `io.debezium.time.ZonedTimestamp.FORMATTER` so we accept everything Debezium actually emits.
- Severity: High

**Issue 2: Test mocks TIMESTAMP_TZ with an anonymous SeaTunnelDataType instead of the real API type, and only covers the happy path**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/test/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConvertersTest.java:214`
- Why it matters: - Root cause: the test builds an anonymous `SeaTunnelDataType<OffsetDateTime>` (lines 214-225) solely to return `SqlType.TIMESTAMP_TZ`, instead of using the canonical data type constant from `seatunnel-api` that Postgres CDC's schema mapper actually produces for `TIMESTAMPTZ` columns.
- Impact: the test does not exercise the same type object the production code path sees, so a mismatch between the real TIMESTAMP_TZ type and this converter could go unnoticed; also, only a single second-precision, non-UTC value (`2026-06-11T10:00:00+05:30`) is covered, while Debezium's `ZonedTimestamp` commonly emits fractional seconds and `Z`-suffixed UTC values (e.g. `2026-06-11T10:00:00.123456Z`).
- Fix: use the canonical TIMESTAMP_TZ type constant from the API module in the test (if one exists — if it does not, that itself needs addressing before this converter is reachable from real schemas), and add assertions for a fractional-second value, a `Z`-offset value, and the non-String failure branch (`IllegalArgumentException`). An e2e case in `seatunnel-connector-v2-e2e` covering a `TIMESTAMPTZ` column in Postgres CDC would also close the loop on the original issue #11005.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/test/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConvertersTest.java:237  String debeziumEmittedValue = "2026-06-11T10:00:00+05:30";`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/test/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConvertersTest.java:214  new SeaTunnelDataType<OffsetDateTime>() {`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/test/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConvertersTest.java:238  String debeziumEmittedValue = "2026-06-11T10:00:00+05:30";`
- Suggested fix: Rather than an anonymous `SeaTunnelDataType` mock, please use the concrete TIMESTAMP_TZ type from `seatunnel-api` that `PostgresTypeConverter` actually emits for `timestamptz` columns, so the test verifies the real type contract. Also, this test only covers one well-formed input — please add cases for a null field, a `Z`-suffixed value with fractional seconds (Debezium's typical `ZonedTimestamp` output, e.g. `2026-06-11T04:30:00.123456Z`), and a malformed string asserting a descriptive exception.
- Severity: Medium

**Issue 3: User-facing type-support change is not reflected in the Postgres CDC connector documentation**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConverters.java:168`
- Why it matters: - Root cause: this PR adds runtime support for `SqlType.TIMESTAMP_TZ` in the Debezium row converter (main file, line 168-169), which changes user-visible behavior for Postgres CDC (`TIMESTAMPTZ` columns now sync instead of crashing), but no file under `docs/` is touched in this PR.
- Impact: users reading `docs/en/connector-v2/source/Postgres-CDC.md` (and the zh counterpart) will still see the old data-type mapping and may believe `TIMESTAMP WITH TIME ZONE` is unsupported, or will not know it now maps to an `OffsetDateTime`-backed `TIMESTAMP_TZ` SeaTunnel type.
- Fix: please update the data type mapping table in `docs/en/connector-v2/source/Postgres-CDC.md` and `docs/zh/connector-v2/source/Postgres-CDC.md` to add the `TIMESTAMPTZ` -> `TIMESTAMP_TZ` row, and check the corresponding checkbox in the PR description checklist. If other CDC connectors (e.g. Oracle, SqlServer) also route through this base converter, note the mapping there as well.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConverters.java:169  return convertToTimestampTz();`
- Suggested fix: Since this makes `TIMESTAMPTZ` a newly supported type for Postgres CDC users, could you also update the data type mapping table in `docs/en/connector-v2/source/Postgres-CDC.md` (and the zh version) to document the `TIMESTAMPTZ` -> `TIMESTAMP_TZ` mapping? Right now the docs still imply this type is unsupported.
- Severity: Medium

**Issue 4: Unnecessary edit to the ASF license header in the test file**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/test/java/org/apache/seatunnel/connectors/cdc/debezium/row/SeaTunnelRowDebeziumDeserializationConvertersTest.java:9`
- Why it matters: - Root cause: the diff changes the license header line from `*    http://www.apache.org/licenses/LICENSE-2.0` (4 spaces, the standard ASF header format used repo-wide) to `* http://www.apache.org/licenses/LICENSE-2.0` (1 space) in the test file.
- Impact: this deviates from the canonical Apache License header template used consistently across the SeaTunnel codebase and creates noisy, unrelated churn in the diff; depending on header-check tooling configuration it can also cause a mismatch against the expected header template.
- Fix: revert this line to the original 4-space indentation so the header matches the rest of the repository and the diff stays focused on the actual fix.
- Suggested fix: This line changes the standard ASF license header indentation (`*    http://...` with 4 spaces is the canonical template used across the repo). Please revert this whitespace-only change so the header stays consistent with the rest of the codebase and the diff stays focused on the fix.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Unhandled DateTimeParseException in convertToTimestampTz() — any string that is not strict ISO_OFFSET_DATE_TIME crashes the job with no context about the column/value

- Issue 2: Test mocks TIMESTAMP_TZ with an anonymous SeaTunnelDataType instead of the real API type, and only covers the happy path
- Issue 3: User-facing type-support change is not reflected in the Postgres CDC connector documentation
- Issue 4: Unnecessary edit to the ASF license header in the test file

DanielLeens:
Following up after @SEZ9's review.

**+1 to @SEZ9's Issue 1** — I traced the same path and want to add one piece of evidence. On the current head, `convertToTimestampTz()` is:

```java
if (dbzObj instanceof String) {
    return java.time.OffsetDateTime.parse((String) dbzObj);
}
throw new IllegalArgumentException(
        "Unable to convert to OffsetDateTime from unexpected value '" + dbzObj + "' of type " + ...);
```

The asymmetry is what makes this worth fixing rather than a nitpick: the *wrong type* branch produces a good diagnostic naming the value and its class, while the *wrong format* branch — much the more likely failure in practice — throws a bare `DateTimeParseException` from deep inside `java.time` with no column name, no table, and no offending value in the message. For a CDC job that means a hard failure whose stack trace does not tell the operator which column or which row broke it, on a connector where the whole point is unattended replication.

`OffsetDateTime.parse` uses `ISO_OFFSET_DATE_TIME` strictly, so anything Debezium emits that deviates — a `Z`-less value, a `+00`-style two-digit offset, or a differently configured `time.precision.mode` — lands there. Wrapping the parse and rethrowing with the column name plus the raw value is a few lines and turns an unactionable failure into an actionable one. I would treat it as blocking for the same reason @SEZ9 did.

I am also +1 on their **Issue 3**: this adds user-visible type support, so the Postgres CDC connector docs should say `TIMESTAMP_TZ` is now supported, in both `docs/en` and `docs/zh`. That is the project's standing rule for user-visible behavior changes, and it is easy to miss on a converter-level fix.

Their Issues 2 and 4 look right to me as well, and I agree those are the lower-priority ones.

Separately, and unchanged from my last comment: GitHub still reports **merge conflicts** between this branch and the latest `dev` (`mergeable_state=dirty`), and the last commit here is from June 18. So the sequence I would suggest is: sync with the latest `dev` and resolve the conflicts, address Issues 1 and 3, then push — at which point CI will produce a signal that is actually attributable to this diff rather than to an 6-week-old base. Right now the failing checks cannot tell you much either way.

No rush from my side, and thanks for your patience on this one — happy to do a full re-review of the whole head as soon as you push the refreshed branch.

davidzollo:

```

```

SEZ9:
@DanielLeens Thanks for the detailed follow-up and the extra evidence — answering it directly, since it slipped through without a reply.

Agreed on the +1 to Issue 1, and your framing of the asymmetry is exactly right: the type-mismatch branch names the value and its class, while the far more common format-mismatch branch leaks a bare `DateTimeParseException` from `java.time` with no column, table, or offending value. Given `OffsetDateTime.parse` uses `ISO_OFFSET_DATE_TIME` strictly, `Z`-less values, `+00`-style two-digit offsets, or a different `time.precision.mode` would all land there unactionably. I'm treating this as blocking too: the fix should wrap the parse and rethrow with the column name plus the raw value in the message.

Also +1 on Issue 3 — this is user-visible type support, so the Postgres CDC connector docs need to state `TIMESTAMP_TZ` support in both `docs/en` and `docs/zh`.

Concrete remaining asks before a full re-review:
1. Sync the branch with the latest `dev` and resolve the reported merge conflicts, so CI signal is attributable to this diff rather than the stale base.
2. Wrap the `OffsetDateTime.parse` call in `convertToTimestampTz()` and rethrow with column name + raw value (Issue 1).
3. Add the `TIMESTAMP_TZ` documentation updates in `docs/en` and `docs/zh` (Issue 3).

Issues 2 and 4 I also agree are valid but lower priority; they can follow in the same push if convenient.

@DanielLeens happy to do a full pass over the refreshed head as soon as it's pushed. Thanks again for the careful trace — it made the case for blocking on Issue 1 much easier to justify.

DanielLeens:
Thanks @SEZ9 — glad this closed the loop cleanly. To confirm alignment on the concrete asks:

1. Sync with `dev` and resolve the merge conflicts (still `mergeable_state=dirty` on the current head `29c3f664f70b`, unchanged since my last review).
2. Wrap the `OffsetDateTime.parse` call in `convertToTimestampTz()` and rethrow with the column name plus raw value (Issue 1) — agreed this is blocking given the asymmetry between the two failure branches.
3. Add the `TIMESTAMP_TZ` mapping to `docs/en/connector-v2/source/Postgres-CDC.md` and the zh counterpart (Issue 3).

@lazyfetch, once those three land in a single push (Issues 2 and 4 are welcome in the same pass but stay non-blocking), I will do a full fresh review of the whole head rather than an incremental diff, same as promised earlier. Thanks both for staying on top of this one.

SEZ9:
@DanielLeens Thanks for laying out the concrete asks so clearly — confirming alignment on all three:

1. **Sync with `dev`** — not done yet. The branch is still at `29c3f664f70b` with the merge conflicts unresolved, so this remains the first step before anything else lands.
2. **Wrap `OffsetDateTime.parse` in `convertToTimestampTz()`** — agreed this is blocking. Rethrowing with the column name plus the raw value addresses the asymmetry you described between the two failure branches; still to be done.
3. **Docs update** — agreed. The `TIMESTAMP_TZ` mapping needs to be added to `docs/en/connector-v2/source/Postgres-CDC.md` and the zh counterpart in the same pass; also still outstanding.

So to be plain: none of the three have landed yet, and the plan is to deliver them together in a single push, with Issues 2 and 4 folded in where feasible as non-blocking improvements. Your suggested sequence makes sense — rebase first so CI signal is attributable to this diff rather than the stale base.

Appreciate the offer of a full fresh review of the whole head once the push lands; that's the right call given how much has moved on `dev`. We'll follow up here when the refreshed branch is up. Thanks for staying so thorough on this one.

DanielLeens:
Thanks @SEZ9, agreed on all three points. Confirmed the head is still `29c3f664f70b13130a233034407bf66ffafe2a7d` with the merge conflicts unresolved, so none of the three asks (dev sync/conflict resolution, wrapping `OffsetDateTime.parse` in `convertToTimestampTz()` with column name + raw value, and the `TIMESTAMP_TZ` docs update in `docs/en`/`docs/zh`) have landed yet.

@lazyfetch, no rush, but once those three are in a single push on a synced head we'll do a full fresh review of the whole diff rather than an incremental one, same as promised.

PR #11071, opened 2026-06-12
[Fix][Zeta] Fix checkpoint barrier trigger test race

zhangshenghang:
### Purpose
Fix a race in `CheckpointBarrierTriggerErrorTest` where the spied `CheckpointManager` was installed before Mockito stubbing was completed. A running checkpoint thread could invoke the spy during stubbing and trigger `UnfinishedStubbing`.

### Changes
- Configure the spy before replacing the checkpoint manager used by the running coordinator.
- Reset the test counter at the start of the test so repeated executions are deterministic.

### Verification
- `./mvnw spotless:apply`

### DIFF
--- a/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointBarrierTriggerErrorTest.java
+++ b/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointBarrierTriggerErrorTest.java
@@ -51,6 +51,7 @@ public class CheckpointBarrierTriggerErrorTest extends AbstractSeaTunnelServerTe
     @Test
     public void testCheckpointBarrierTriggerError()
             throws NoSuchFieldException, IllegalAccessException {
+        COUNTER.set(0);
         long jobId = System.currentTimeMillis();
         startJob(jobId, CONF_PATH);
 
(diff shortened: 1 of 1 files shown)


DanielLeens (commented):
Thanks for the fix. I reviewed the latest head from the checkpoint-barrier test path and the new synchronization logic.

# What This PR Solves
- User pain: the existing checkpoint barrier trigger test has a race, so it can fail even when the runtime behavior is still correct.
- Fix approach: tighten the synchronization in the test so the assertion waits on the intended signal instead of racing the barrier trigger path.
- One-line summary: I did not find a blocker in the current test-only fix, and the latest change looks like a reasonable race-condition cleanup.

# Runtime/Test Chain Rechecked
```text
test setup
  -> checkpoint barrier trigger test scaffolding
      -> start task/test harness
      -> coordinate barrier trigger signal
      -> assert after the intended synchronization point instead of racing it
```

# Findings

# Test Coverage / Stability
- This is test-only scope.
- Stability rating: Stable.
  - The whole point of the change is to reduce a race window.
  - I did not see a new flaky-test anti-pattern introduced by the fix.

# CI Note
- The current Build is still in progress on the latest head.
- From the code review side I do not see a blocker, but merge should still wait for the check result.

# Merge Conclusion
### Conclusion: can merge

1. Blocking items
- None from my side on the current revision.

2. Suggested follow-up
- Wait for the current Build result before merging.

Overall, this looks like a reasonable test-race fix, and I do not see a new problem introduced by the current head.

dybyte (approved):
+1 if CI passes

DanielLeens:
Thanks @dybyte for taking a look. Since there is no new code after my review and your approval is on the same head, I do not have an additional source-level follow-up from Daniel's side in this pass.

If the latest CI result is clean, I am not waiting on another rereview here.

PR #11073, opened 2026-06-12
[Chore]Restore SeaTunnelEngineLocalExample

niumy0701:
Running SeaTunnelEngineLocalExample locally failed, but restoring the code and running it successfully

error:

### Check list

### DIFF
--- a/seatunnel-examples/seatunnel-engine-examples/src/main/java/org/apache/seatunnel/example/engine/SeaTunnelEngineLocalExample.java
+++ b/seatunnel-examples/seatunnel-engine-examples/src/main/java/org/apache/seatunnel/example/engine/SeaTunnelEngineLocalExample.java
@@ -27,7 +27,6 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Paths;
 
-/** Local example that shows how to enable stain trace reporting for an engine job submission. */
 public class SeaTunnelEngineLocalExample {
 
     static {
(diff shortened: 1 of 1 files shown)


DanielLeens (commented):
Thanks for the contribution. I reviewed the current diff against the refreshed dev baseline and traced the startup path of SeaTunnelEngineLocalExample.

# What This PR Solves
- User pain: the current local engine example points to stain-trace resources that are no longer present in the example resources, so the default example path is broken out of the box.
- Fix approach: restore the local example to the normal fake_to_console example config and remove the hard-wired stain-trace seatunnel.config override.
- One-line summary: the code-side fix is straightforward and correct on the real startup path; the remaining blocker is that the current head still does not have a usable Build signal.

# Runtime Chain Rechecked
```text
local example startup
  -> SeaTunnelEngineLocalExample.main(...)
      -> choose default config path when args are empty
      -> getTestConfigFile(configurePath)
          -> ClassLoader resource lookup
      -> build ClientCommandArgs
      -> MasterType.LOCAL
      -> SeaTunnel.run(...)
```

# Findings
I did not find a new code-side blocker in the current example fix.

Key evidence I rechecked locally:
- On upstream/dev, SeaTunnelEngineLocalExample still defaults to `/examples/stain_trace_fake_sql_union_to_console.conf` and also auto-sets `/examples/stain_trace_seatunnel.yaml`.
- Those stain-trace example resources are no longer present under `seatunnel-examples/seatunnel-engine-examples/src/main/resources/examples`.
- The restored default `/examples/fake_to_console.conf` does exist, and that same example path is already the one used by the cluster client example.

Issue 1: the current head still does not have a usable Build signal
- Location: GitHub Build check `80982848465`
- Why this is a blocker:
  the current top-level check is `ACTION_REQUIRED` with `Workflow run detection failed`, so the PR still does not have a real build/test signal attached to the latest head.
- Better fix:
  follow the check-run guidance to get a real Build run attached to this head, for example by enabling Actions in the fork, pushing a no-op trigger commit, or rebasing onto the latest dev if needed.
- Severity: High

# Test Coverage / Stability
- This PR does not add tests.
- For the current scope, the main verification is the actual example-resource lookup path, and the restored default points to an existing resource.

# Merge Conclusion

1. Blocking items
- Issue 1: get the Build workflow to run normally on this head.

2. Suggested follow-up
- If the project still wants a stain-trace-specific example later, it would be better to add it back as a separate explicit example instead of overloading the default local example entrypoint.

Overall, the example fix itself looks correct. The only thing still preventing merge from my side is the missing Build signal on the current head.

davidzollo:
Thank you very much for correcting my mistakes.
Meanwhile, you are warmly welcome to contribute to the community!
I'm looking forward to your more participation^_^

davidzollo (approved):
+1
Good job.

PR #11078, opened 2026-06-12
[Fix][Connector-V2] Support SFTP keyfile option

### LINKED ISSUE #5516: [Bug] [SFTP] SFTP connector miss fs.sftp.keyfile option
### Search before asking

### What happened

I find we cannot set `fs.sftp.keyfile` in sftp connector, this means we only support use username and password to create ssh connection, but I find in the source code, there exist `fs.sftp.keyfile`
https://github.com/apache/seatunnel/blob/c23070919c8abde6e95e42c28095ca318dddc5e7/seatunnel-connectors-v2/connector-file/connector-file-sftp/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sftp/system/SFTPFileSystem.java#L120-L132

so this is a missing option?

### SeaTunnel Version

dev

### SeaTunnel Config

```conf
-
```

### Running Command

```shell
-
```

### Error Exception

```log
-
```

_No response_

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
### Purpose

Fixes #5516.

The SFTP filesystem layer already supports `fs.sftp.keyfile`, but the SFTP connector did not expose a connector-level option for it. As a result, users could not configure public key authentication through normal `SftpFile` source or sink options.

### Changes

- Add a `keyfile` option to the SFTP connector options.
- Pass `keyfile` through `SftpConf` to the existing `SFTPFileSystem.FS_SFTP_KEYFILE` setting.
- Keep password authentication supported while allowing password to be omitted when keyfile authentication is used.
- Add unit coverage for option exposure and Hadoop config forwarding.
- Update English and Chinese SFTP source/sink documentation.

### Tests

- `JAVA_HOME=/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home ./mvnw spotless:apply`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home ./mvnw -q -DskipTests verify`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home ./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-sftp -DskipIT -Dskip.spotless=true -Dtest=SftpFileFactoryTest test`


DanielLeens (commented):
Thanks for the contribution. I reviewed the latest head from the SFTP factory validation path down into `SftpConf.buildWithConfig(...)`. The feature direction makes sense, but there is still one blocker on the submit-time validation path.

Runtime path rechecked:
```text
SftpFileSourceFactory / SftpFileSinkFactory.optionRule()
  -> submit-time config validation
  -> SftpConf.buildWithConfig(...)
      -> host / user
      -> optional password
      -> optional keyfile
```

Issue 1: the new option rules no longer enforce that at least one authentication method is present
- Location: `seatunnel-connectors-v2/connector-file/connector-file-sftp/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sftp/sink/SftpFileSinkFactory.java:52-69`, `.../source/SftpFileSourceFactory.java:46-66`, `.../config/SftpConf.java:51-60`
- Why this is a blocker:
  the docs now correctly say `password` is required when `keyfile` is not set, but the factories do not actually enforce that contract. A config with `host/user/path` and neither `password` nor `keyfile` now passes submit-time validation and only fails later when the runtime SFTP client tries to authenticate.
- Risk:
  users get a late runtime failure instead of a clear config error, which is exactly the kind of validation gap this PR is supposed to remove.
- Better fix:
  add an explicit validation rule for `password OR keyfile` at submit time. If the current `OptionRule` combinations cannot express it cleanly, keep a small imperative fallback check rather than leaving the contract unenforced.
- Severity: High

1. Blocking items
- Issue 1: enforce `password` or `keyfile` before runtime.

2. Suggested follow-up
- Add one negative factory test that keeps the other required options present but omits both authentication fields, and assert that validation fails at submit time.

davidzollo (approved):
+1
Good job

PR #11098, opened 2026-06-13
[Fix][Connector-V2] Fix HTTP cursor pagination no-progress loop

### LINKED ISSUE #10929: [Feature][connector] http connector cursor type pageing not check the number of returned items is less than batchSize？
### Search before asking

### Description

in HttpSourceReader.java 406-407, there are only one kind check is "noMoreElementFlag = Strings.isNullOrEmpty(newCursor)" why not check datasize ? some kind of system return the last same cursor when data is over!

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

zhangshenghang:
### Purpose

Fixes #10929.

HTTP cursor pagination could keep polling forever when an API returned the same non-empty cursor for a terminal page. In that case the next request cannot make progress because it is sent with the same cursor again.

### Change

- Track the cursor used for the current request before reading the response cursor.
- Stop cursor pagination when the response cursor is empty or unchanged.
- Add a unit test that reproduces a terminal page with the same cursor and an empty data array.

### Verification

- ./mvnw -pl seatunnel-connectors-v2/connector-http/connector-http-base -Dtest=HttpSourceReaderInternalPollNextTest#testCursorPaginationStopsWhenCursorDoesNotAdvance test
- ./mvnw -pl seatunnel-connectors-v2/connector-http/connector-http-base test
- ./mvnw spotless:apply

### DIFF
--- a/seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/source/HttpSourceReader.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/source/HttpSourceReader.java
@@ -464,6 +464,7 @@ public class HttpSourceReader extends AbstractSingleSplitReader<SeaTunnelRow> {
 
             // cursor pagination
             if (HttpPaginationType.CURSOR.getCode().equals(pageInfo.getPageType())) {
+                String currentCursor = pageInfo.getCursor();
                 // get cursor value from response JSON with fileName
                 String cursorResponseField = pageInfo.getPageCursorResponseField();
                 ReadContext context = JsonPath.using(jsonConfiguration).parse(data);
@@ -473,8 +474,11 @@ public class HttpSourceReader extends AbstractSingleSplitReader<SeaTunnelRow> {
                     newCursor = cursorList.get(0);
                 }
                 pageInfo.setCursor(newCursor);
-                // if not present cursor, then no more data
-                noMoreElementFlag = Strings.isNullOrEmpty(newCursor);
+                // If the response cursor is empty or unchanged, the next request cannot make
+                // progress.
+                noMoreElementFlag =
+                        Strings.isNullOrEmpty(newCursor)
+                                || Objects.equals(currentCursor, newCursor);
             } else {
                 // if not set page pagination is default
                 // Determine whether the task is completed by specifying the presence of the 'total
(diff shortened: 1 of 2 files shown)


DanielLeens (approved):
Thanks for the contribution. I reviewed the current diff against `dev` and traced the real cursor-pagination loop in `HttpSourceReader`.

# What this PR solves
- User pain: a cursor-based HTTP source can loop forever when the server keeps returning the same cursor.
- Fix approach: stop the bounded read once the response cursor is empty or unchanged, and add a regression test for the no-progress case.
- One-line summary: the fix matches the real runtime path and I did not find a blocker in the current head.

# Runtime path rechecked
`HttpSourceReader.internalPollNext(...)`
  -> cursor pagination branch
  -> read current cursor
  -> parse next cursor from response
  -> stop when the cursor is empty or unchanged

The new unit test covers the exact no-progress branch and does it without brittle sleeps or timing assumptions.

### Conclusion: can merge
1. Blocking items

2. Suggested follow-up
- The PR-level `Build` is still red right now, so please keep an eye on CI, but I do not have a source-level blocker to add.

Nice targeted fix.

DanielLeens (approved):
Thanks for the contribution. I re-reviewed the latest head locally from the real cursor-pagination path in `HttpSourceReader`, not only from the small diff.

What this PR solves
- User pain: a cursor-based HTTP source can keep sending requests forever if the upstream API keeps returning the same cursor, so the job looks alive but makes no forward progress.
- Fix approach: compare the newly parsed cursor with the current cursor and stop when the cursor is empty or unchanged.
- One-line summary: this patch adds a proper "no forward progress" stop condition to cursor pagination.

Full runtime path checked
`HttpSourceReader.internalPollNext()` enters the cursor loop -> `updateRequestParam()` injects the current cursor into the request -> `pollAndCollectData()` fetches a response -> `collect()` parses the response cursor -> `pageInfo.setCursor(newCursor)` updates state -> the reader decides whether to continue.
Before this patch, `collect()` only stopped when `newCursor` was empty, so an unchanged cursor could still spin forever. After this patch, `HttpSourceReader.java:466-481` also stops when `Objects.equals(currentCursor, newCursor)`.
In bounded mode, the existing `finally` path still calls `context.signalNoMoreElement()`, so the finish signal path remains intact.

Core findings
- The normal cursor-pagination path definitely hits this change.
- The fix is precise: it changes the termination condition exactly where cursor state advances.
- I did not find compatibility issues for API, config, defaults, protocol, or serialization.
- The test in `HttpSourceReaderInternalPollNextTest.java:142-183` covers the real regression path: first response advances the cursor, second response repeats it, the reader stops after 2 requests, and bounded mode signals end-of-input.

Test stability
- Stability rating: Stable.
- Reason: the new test is fully mocked, does not depend on sleep-based timing, external ports, containers, or system clocks, and asserts deterministic request counts plus the end-of-input signal.

Merge conclusion
### Conclusion: can be merged
1. Blocking items
- None.
2. Non-blocking suggestions
- None.

This looks like a clean fix for a real infinite-loop condition. Nice improvement.

CosmosNi (approved):
Good job!Wait for the CI to finish running.

DanielLeens:
Thanks @CosmosNi. I rechecked the thread state and there is still no new code after the head I already approved (`41caae5`). I agree with your current read: the cursor no-progress fix itself looks ready, and from Daniel's side the remaining blocker is the `Build` gate finishing green. If another commit lands, I will re-review the latest head from scratch.

davidzollo (approved):
+1

Issue #11045, opened 2026-06-10
[Feature][Transform-V2][CDC] Add configurable rule evaluation mode for multi-table transform matching


davidzollo:
## Background
SeaTunnel already supports multi-table transform configuration through `table_match_regex` and ordered `table_transform` rules.

What is still missing is an explicit contract for **how matching rules are evaluated when more than one rule can match the same upstream table**.

Different teams need different semantics:
- some want evaluation to stop at the first matching table-specific rule,
- others want every matching rule to be applied in declaration order.

Without an explicit evaluation mode, users and contributors have to infer behavior from implementation details and rule ordering, which makes complex multi-table transform jobs hard to reason about.

## Scope
Add a configurable rule evaluation mode for multi-table transform matching.

This issue is about **matching and rule-application semantics inside multi-table transform configuration**. It does **not** cover sink table routing, soft-delete conversion, or metadata enrichment.

## First delivery boundary
The first delivery should be limited to the existing multi-table transform model:
- outer shared rule via `table_match_regex`
- ordered table-specific overrides via `table_transform`

The implementation should not expand into unrelated sink-routing or table-discovery work.

## Exact evaluation contract
The new mode should govern how multiple matching `table_transform` entries are handled for the same upstream table.

Recommended first-delivery modes:
- `FIRST_MATCH`
- `ALL_MATCH`

### `FIRST_MATCH`
- evaluate `table_transform` entries in declaration order
- apply only the first matching entry
- stop evaluating remaining `table_transform` entries
- if no `table_transform` entry matches, fall back to the outer shared `table_match_regex` rule if it matches

### `ALL_MATCH`
- evaluate all matching `table_transform` entries in declaration order
- apply every matching entry sequentially
- if no `table_transform` entry matches, fall back to the outer shared `table_match_regex` rule if it matches

## Current compatibility rule
Backward compatibility is mandatory.

The implementation should first codify the current behavior explicitly and preserve it as the default mode. The issue should not silently change how existing multi-table transform configurations are interpreted.

### 1. Add an explicit matching mode config
Illustrative target shape:

```yaml
rule-match-mode: FIRST_MATCH | ALL_MATCH
```

Exact naming can be refined during implementation, but semantics must be explicit.

### 2. Apply the mode at the multi-table transform rule resolution point
Primary touch points are expected to include:
- `TransformCommonOptions.TABLE_MATCH_REGEX`
- `TransformCommonOptions.MULTI_TABLES`
[...]

PR #11099, opened 2026-06-13
[Feature][Transform-V2] Add multi-table rule match mode

goutamadwant:

Enhances #11045.

This PR adds `rule_match_mode` for multi-table transform matching.

Changes:

- adds `FIRST_MATCH` and `ALL_MATCH` modes for repeated `table_transform` entries with the same exact `table_path`
- keeps the legacy behavior when `rule_match_mode` is not configured, so duplicate exact `table_path` rules still fail fast
- keeps `table_match_regex` as the fallback when no exact `table_transform` rule matches
- applies `ALL_MATCH` rules in declaration order, so later rules consume the schema and row produced by earlier rules
- preserves schema-change filtering when a chained transform returns `null`
- wires the new option into transform factories that support multi-table transform
- updates English and Chinese docs with an `ALL_MATCH` example
- adds focused common resolver tests and a real `CopyFieldMultiCatalogTransform` regression test

Yes.

Users can now configure `rule_match_mode = "FIRST_MATCH"` or `rule_match_mode = "ALL_MATCH"` when multiple `table_transform` entries target the same exact `table_path`.

If `rule_match_mode` is not configured, duplicate exact `table_path` entries still fail during config resolution. Existing valid configs keep the same behavior.

Added unit coverage for:

- legacy duplicate exact `table_path` fail-fast behavior when `rule_match_mode` is not configured
- explicit `FIRST_MATCH`
- `ALL_MATCH` sequential rule application
- deterministic declaration order
- fallback to outer `table_match_regex`
- no-match identity behavior
- chained schema-change null short-circuit behavior for map and flatMap transforms
- real Copy transform chaining where the second rule reads a field produced by the first rule

Verified with:

```shell
JAVA_HOME=$(/usr/libexec/java_home -v 11) PATH="$JAVA_HOME/bin:$PATH" ./mvnw -pl seatunnel-transforms-v2 -Dtest=org.apache.seatunnel.transform.common.AbstractMultiCatalogTransformTest,org.apache.seatunnel.transform.copy.CopyFieldMultiCatalogTransformTest test
JAVA_HOME=$(/usr/libexec/java_home -v 11) PATH="$JAVA_HOME/bin:$PATH" ./mvnw -pl seatunnel-transforms-v2 -DskipIT package
```

### Check list

### DIFF
--- a/docs/en/transforms/transform-multi-table.md
+++ b/docs/en/transforms/transform-multi-table.md
@@ -19,7 +19,7 @@ Multi-table Transform has no limitations on Transform capabilities; any Transfor
 | table_match_regex          | String | No       | .*      | A regular expression to match the tables that require transformation. By default, it matches all tables. Note that this table name refers to the actual upstream table name, not `plugin_output`.                                                               |
 | table_transform            | List   | No       | -       | You can use a list in `table_transform` to specify rules for individual tables. If a transformation rule is configured for a specific table in `table_transform`, the outer rules will not apply to that table. The rules in `table_transform` take precedence. |
 | table_transform.table_path | String | No       | -       | When configuring a transformation rule for a table in `table_transform`, you need to specify the table path using the `table_path` field. The table path should include `databaseName[.schemaName].tableName` and is matched exactly.                         |
-| rule_match_mode            | String | No       | FIRST_MATCH | Controls how multiple `table_transform` entries with the same exact `table_path` are evaluated. Supported values are `FIRST_MATCH` and `ALL_MATCH`.                                                                                                         |
+| rule_match_mode            | String | No       | -       | Controls how multiple `table_transform` entries with the same exact `table_path` are evaluated. Supported values are `FIRST_MATCH` and `ALL_MATCH`.                                                                                                         |
 
 ## Matching Logic
 
@@ -69,7 +69,8 @@ For each table, the priority of configuration is: `table_transform` > `table_mat
 
 `table_transform.table_path` uses exact table path matching. `rule_match_mode` only controls the case where multiple `table_transform` entries use the same exact `table_path`:
 
-- `FIRST_MATCH`: apply the first matching `table_transform` entry in declaration order. This is the default behavior.
+- If `rule_match_mode` is not configured, duplicate exact `table_path` entries are rejected during config resolution.
+- `FIRST_MATCH`: apply the first matching `table_transform` entry in declaration order.
 - `ALL_MATCH`: apply all matching `table_transform` entries in declaration order. The output of an earlier matching rule becomes the input of the next matching rule for the same table.
 
 For example, when `rule_match_mode` is `ALL_MATCH`, the following configuration first copies `name` to `name2` for `test.xyz`, and then copies `name2` to `name3` for the same table:
--- a/docs/en/transforms/transform-multi-table.md
+++ b/docs/en/transforms/transform-multi-table.md
@@ -18,7 +18,8 @@ Multi-table Transform has no limitations on Transform capabilities; any Transfor
 |----------------------------|--------|----------|---------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | table_match_regex          | String | No       | .*      | A regular expression to match the tables that require transformation. By default, it matches all tables. Note that this table name refers to the actual upstream table name, not `plugin_output`.                                                               |
 | table_transform            | List   | No       | -       | You can use a list in `table_transform` to specify rules for individual tables. If a transformation rule is configured for a specific table in `table_transform`, the outer rules will not apply to that table. The rules in `table_transform` take precedence. |
-| table_transform.table_path | String | No       | -       | When configuring a transformation rule for a table in `table_transform`, you need to specify the table path using the `table_path` field. The table path should include `databaseName[.schemaName].tableName`.                                                  |
+| table_transform.table_path | String | No       | -       | When configuring a transformation rule for a table in `table_transform`, you need to specify the table path using the `table_path` field. The table path should include `databaseName[.schemaName].tableName` and is matched exactly.                         |
+| rule_match_mode            | String | No       | FIRST_MATCH | Controls how multiple `table_transform` entries with the same exact `table_path` are evaluated. Supported values are `FIRST_MATCH` and `ALL_MATCH`.                                                                                                         |
 
 ## Matching Logic
 
@@ -65,6 +66,35 @@ transform {
 This allows us to handle transformations for multiple tables within a single transform configuration.
 
 For each table, the priority of configuration is: `table_transform` > `table_match_regex`. If no rules match a table, no transformation will be applied.
+
+`table_transform.table_path` uses exact table path matching. `rule_match_mode` only controls the case where multiple `table_transform` entries use the same exact `table_path`:
+
+- `FIRST_MATCH`: apply the first matching `table_transform` entry in declaration order. This is the default behavior.
+- `ALL_MATCH`: apply all matching `table_transform` entries in declaration order. The output of an earlier matching rule becomes the input of the next matching rule for the same table.
+
+For example, when `rule_match_mode` is `ALL_MATCH`, the following configuration first copies `name` to `name2` for `test.xyz`, and then copies `name2` to `name3` for the same table:
+
+```hocon
+transform {
+  Copy {
+    rule_match_mode = "ALL_MATCH"
+
+    table_transform = [{
+      table_path = "test.xyz"
+      src_field = "name"
+      dest_field = "name2"
+    }, {
+      table_path = "test.xyz"
+      src_field = "name2"
+      dest_field = "name3"
+    }]
+  }
+}
+```
+
+Output structure:
+
+| id | name | age | name2 | name3 |
 
 Below are the transform configurations for each table:
 
--- a/docs/zh/transforms/transform-multi-table.md
+++ b/docs/zh/transforms/transform-multi-table.md
@@ -20,7 +20,8 @@ SeaTunnel 的 Multi-Table Transform 支持在上游插件输出多个表时，�
 |----------------------------|--------|----------|---------|--------------------------------------------------------------------------------------------------|
 | table_match_regex          | String | No       | .*      | 表名的正则表达式，通过正则表达式来匹配需要进行转换的表，默认匹配所有的表。注意这个表名是上游的真正表名，不是`plugin_output`。                           |
 | table_transform            | List   | No       | -       | 可以通过table_transform列表来指定部分表的规则，当在table_transform中配置某个表的转换规则后，外层针对当前表的规则不会生效，以table_transform中的为准 |
-| table_transform.table_path | String | No       | -       | 当在table_transform中配置某个表的转换规则后，需要使用table_path字段指定表名，表名需要包含`databaseName[.schemaName].tableName`。  |
+| table_transform.table_path | String | No       | -       | 当在table_transform中配置某个表的转换规则后，需要使用table_path字段指定表名，表名需要包含`databaseName[.schemaName].tableName`，并且按完整表路径精确匹配。  |
+| rule_match_mode            | String | No       | FIRST_MATCH | 控制多个 `table_transform` 配置使用相同精确 `table_path` 时的执行方式，支持 `FIRST_MATCH` 和 `ALL_MATCH`。 |
 
 ## 匹配逻辑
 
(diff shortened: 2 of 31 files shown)


goutamadwant:
@davidzollo I checked the failed Build run. The style, license, docs, website build and Java 8/11 unit test jobs passed.

The failures look unrelated to this transform change. I see RocketMQ e2e failing on restore offset mismatch and engine k8s failing on Maven Central dependency download with connection reset. I do not see a failure pointing to the transform module changes in this PR.

DanielLeens (commented):
Thanks for the contribution. I reviewed the current diff against `dev` and traced the multi-table rule-selection path through `AbstractMultiCatalogTransform` and the new chained transform wrappers.

# What this PR solves
- User pain: for the same exact table path, users sometimes need to stop at the first matching rule, and sometimes they need to apply multiple rules in order.
- Fix approach: add `rule_match_mode`, keep `FIRST_MATCH` as the default, and introduce chained map/flatMap transforms for the `ALL_MATCH` case.
- One-line summary: the core direction looks right, and I did not find a new source-level blocker in the current head.

# Runtime path rechecked
multi-table transform setup
  -> collect exact `table_transform` entries for a table path
  -> `FIRST_MATCH`: keep only the first rule
  -> `ALL_MATCH`: build an ordered chain
  -> propagate produced catalog table into the next rule in the chain

I also checked the new tests around duplicate exact-table-path rules and the chained-transform behavior. The default remains backward compatible, and the chain path is covered directly in unit tests.

# CI
The current PR-level `Build` check is red, so I would still wait for the latest workflow result before merge.

1. Blocking items
- No new source-level blocker from me on the current head.
- Please let the latest `Build` finish cleanly before merge.

2. Suggested follow-up
- None beyond CI.

Thanks for tightening the multi-table rule behavior.

davidzollo:
Thanks for working on this. The direction makes sense, and the `ALL_MATCH` flow itself looks useful.

I still see two things that should be addressed before merge:

1. The default path is no longer backward compatible for duplicate exact `table_path` rules.

Before this change, repeated exact `table_path` entries would fail during config resolution because they collapsed into a map with duplicate keys. After this PR, the default `FIRST_MATCH` path silently keeps the first rule and ignores the rest in `AbstractMultiCatalogTransform`.

That means an existing invalid config now starts successfully but behaves differently, which conflicts with the compatibility requirement described in #11045.

I think we should keep the legacy fail-fast behavior unless the user explicitly opts into repeated exact-table rules with the new mode, instead of silently turning duplicates into "first one wins" by default.

2. The new chained schema-change helpers should preserve the existing `null` short-circuit contract.

`TransformFlowLifeCycle` treats `null` from `mapSchemaChangeEvent()` as "this event is filtered, stop propagating it". However, `ChainedMapTransform` and `ChainedFlatMapTransform` continue iterating even if one inner transform returns `null`, so later transforms can still receive a filtered event.

It would be safer to short-circuit immediately when `currentEvent == null` and return `null` from the chained helper as well.

Non-blocking suggestion:

- It would be great to add one focused schema-change test for the chained helper path after the fix, since the current new coverage is strong for row processing but does not yet exercise the chained `mapSchemaChangeEvent()` behavior.

Thanks again — the overall idea looks good, but I think the two items above are worth tightening before this goes in.

goutamadwant:
Thanks @davidzollo for your suggestions, updated the code to address them.

The default path now keeps the old duplicate table_path fail-fast behavior unless rule_match_mode is explicitly configured. I also changed the chained map and flatMap helpers to stop schema-change propagation when an inner transform returns null.

Added focused coverage for the chained schema-change path as well.

Verified locally with the focused tests and the seatunnel-transforms-v2 package.

Let me know if you have any more feedback. Thanks!

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head against the real multi-table rule-selection and schema-change propagation paths.

+1 to @davidzollo's earlier points about duplicate exact `table_path` handling and null schema-change propagation. I rechecked both paths on the latest head, and those two concerns are now addressed.

# What this PR solves
- User pain: for the same exact `table_path`, users sometimes want only the first rule, and sometimes want multiple rules applied in order. But the default path also must not silently break the old fail-fast behavior for duplicate exact `table_path` entries.
- Fix approach: keep chained behavior behind explicit `rule_match_mode`, restore duplicate exact `table_path` fail-fast when that option is not configured, and stop chained schema-change propagation as soon as an inner transform returns `null`.
- One-line summary: the latest head makes `ALL_MATCH` a clear opt-in and restores the old compatibility boundary on the default path.

# Runtime chain rechecked
```text
transform initialization
  -> AbstractMultiCatalogTransform.<init>() [L48-104]
      -> read RULE_MATCH_MODE
      -> if not explicitly configured
          -> rejectDuplicateTablePaths(singleTableConfigs) [L109-120]
      -> if explicitly configured as ALL_MATCH
          -> build chained transform path

schema-change propagation
  -> AbstractMultiCatalogTransform.mapSchemaChangeEvent(...) [L175-190]
      -> targetTransform.mapSchemaChangeEvent(event)
      -> null result returns immediately
      -> otherwise refreshOutputCatalogTables() and propagate produced schema

chained wrappers
  -> ChainedMapTransform.mapSchemaChangeEvent(...) [L70-82]
  -> ChainedFlatMapTransform.mapSchemaChangeEvent(...) [L76-88]
      -> stop as soon as one inner transform returns null
```

# Key findings
1. The default path is backward compatible again. Duplicate exact `table_path` rules are rejected unless the user explicitly opts into the new match-mode behavior.
2. The chained schema-change null-propagation gap is fixed. Once one inner transform filters the event out, downstream transforms are no longer invoked.
3. I did not find a new source-level blocker on the latest head.

# Tests
- The new unit tests for duplicate-path fail-fast and chained schema-change short-circuit are exactly the focused regression coverage this revision needed.
- I do not see a flaky-test pattern in the added tests.

# CI
- `Notify test workflow` and `labeler` are green.
- `Build` is still running on the latest head, so I would wait for that result before merge.

1. Blocking items
- No source-level blocker from my side.

2. Suggested follow-up
- None.

The latest revision looks consistent to me.

davidzollo:
> Thanks @davidzollo for your suggestions, updated the code to address them.
> 
> The default path now keeps the old duplicate table_path fail-fast behavior unless rule_match_mode is explicitly configured. I also changed the chained map and flatMap helpers to stop schema-change propagation when an inner transform returns null.
> 
> Added focused coverage for the chained schema-change path as well.
> 
> Verified locally with the focused tests and the seatunnel-transforms-v2 package.
> 
> Let me know if you have any more feedback. Thanks!


Thanks for the update, and thanks for tightening this carefully.

I rechecked the latest head, and the two original source-level concerns are addressed now:

- the default path restores the old fail-fast behavior for duplicate exact `table_path` rules unless `rule_match_mode` is explicitly configured
- the chained schema-change path now stops propagation as soon as an inner transform returns `null`

The focused regression coverage for these paths also looks aligned with the fix.

So from my side, I don't have further source-level feedback on the original review topic.

The remaining blockers before merge are the current PR state:
- the latest `Build` check is still red

Once those are cleared, this looks good to me.

davidzollo (dismissed):
+1 if CI passes.
LGTM

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch again against the multi-table rule-selection and schema-change propagation paths.

## What this PR solves
- User pain: for the same exact `table_path`, users sometimes want only the first rule, and sometimes want multiple rules applied in order, but the default path still must not silently break the old duplicate-rule fail-fast behavior.
- Fix approach: keep chained behavior behind explicit `rule_match_mode`, restore duplicate exact `table_path` fail-fast when that option is not configured, and stop chained schema-change propagation as soon as an inner transform returns `null`.
- One-line summary: the latest head still looks source-clean to me, and I do not see a reopened blocker on the current revision.

```text
transform initialization
  -> AbstractMultiCatalogTransform.<init>()
      -> read RULE_MATCH_MODE
      -> if not explicitly configured
          -> rejectDuplicateTablePaths(...)
      -> if ALL_MATCH
          -> build chained transform path

schema-change propagation
  -> AbstractMultiCatalogTransform.mapSchemaChangeEvent(...)
      -> targetTransform.mapSchemaChangeEvent(event)
      -> null result returns immediately
      -> otherwise refreshOutputCatalogTables() and propagate produced schema
```

## Key findings
- The default path is still backward compatible on the latest head.
- The chained schema-change null-propagation gap remains fixed.
- I did not find a new source-level blocker on the current revision.

## Tests
- The focused unit tests for duplicate-path fail-fast and chained schema-change short-circuit still cover the key contract points.
- I do not see a flaky-test anti-pattern in the current test delta.

## CI
- The latest `Build` is still queued on the current head.

1. Blocking items
- No source-level blocker from my side.

2. Suggested follow-up
- None.

The latest revision still looks consistent to me.

davidzollo:
I took another look at the failed `Build` check. The failure was not coming from the multi-table rule match mode changes in this PR.

The failing job was `jdbc-connectors-it-ddl (11, ubuntu-latest)`, and the actual error was in `SqlServerSchemaChangeIT.testMysqlCdcWithSchemaEvolutionCaseExactlyOnce` under `connector-jdbc-e2e-ddl`. That path does not go through the transform code changed here.

The good news is that this was already addressed independently on `dev` by [Test][E2E] Retry SQL Server XA setup until login is ready (`#11213`). I synced the latest `upstream/dev` into this branch and pushed the update, so this PR now includes that unrelated CI stabilization through the merge base sync.

A fresh `Build` run is queued now. If it still fails, we can inspect the new log on the updated head, but the previous red check was not a blocker caused by this PR itself.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the multi-table transform rule-selection and schema-change propagation path again.

## What I rechecked on the current head
```text
transform initialization
  -> AbstractMultiCatalogTransform
      -> RULE_MATCH_MODE
      -> default duplicate exact-table-path fail-fast path
      -> explicit ALL_MATCH chained path

schema-change propagation
  -> mapSchemaChangeEvent(...)
      -> stop propagation when an inner transform returns null
```

I also compared the current PR file surface with the last Daniel-reviewed source revision. I do not see an effective reopened change in the actual transform files that would change the earlier source-side conclusion.

- The default path still looks backward compatible to me.
- The chained schema-change short-circuit fix still stands on the current revision.
- The remaining gate is the current `Build`, which is still in progress.

1. Blocking items

- None from this round.

DanielLeens:

- base branch: `dev`
- head SHA: `4801778745d33e90e87aae1b8da2012836dafa6d`
- compare status: `diverged`
- ahead_by: `4`
- behind_by: `4`
- mergeable_state: `unknown`

DanielLeens:

- base branch: `dev`
- head SHA: `4801778745d33e90e87aae1b8da2012836dafa6d`
- compare status: `diverged`
- ahead_by: `4`
- behind_by: `12`

goutamadwant:
@DanielLeens Synced this branch with latest dev and pushed the rebased branch.

There was one rebase conflict in RegexExtractTransformFactory. I kept the newer default_values validation from dev and added rule_match_mode alongside multi_table.

Local diff checks are clean. I could not complete the transform module test locally because current dev fails before reaching this PR code in seatunnel-config-shade. I checked the same command on clean origin/dev and it fails there as well, so that local build failure does not look introduced by this branch.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the multi-table transform rule-selection and schema-change propagation path again.

## What I rechecked on the current head
```text
transform initialization
  -> AbstractMultiCatalogTransform
      -> RULE_MATCH_MODE
      -> default duplicate exact-table-path fail-fast path
      -> explicit ALL_MATCH chained path

schema-change propagation
  -> mapSchemaChangeEvent(...)
      -> stop propagation when an inner transform returns null
```
I rechecked the current PR file surface and do not see an effective reopened change in the transform files that would change the earlier source-side conclusion.

## Review conclusion
1. Blocking items

2. Suggested follow-up
- None from this round.

davidzollo (approved):
Rechecked on the latest head.

The two earlier source-level blockers are still fixed, and I do not see any new code-path regressions in the current diff.

I also fixed the current PR-related `Build` failure locally: it was a Spotless formatting issue in `RegexExtractTransformFactory.java`. The formatting-only change has been pushed to this branch, and a fresh CI run is queued now.

From the code review side, this looks good to me now. Once the refreshed checks are green, I am fine with merging it.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head locally again.

What changed since my last Daniel review:
- The only effective code delta on top of `2b990e0aca00c948e8289e85ae58cdcab11af8a3` is the Spotless cleanup in `RegexExtractTransformFactory.java`.
- That formatting-only update does not change the earlier source-level conclusion on the multi-table rule-selection and schema-change propagation paths.

Current path I rechecked:
```text
transform initialization
  -> AbstractMultiCatalogTransform
      -> default duplicate exact-table-path fail-fast path
      -> explicit ALL_MATCH chained path

schema-change propagation
  -> mapSchemaChangeEvent(...)
      -> stop propagation when an inner transform returns null
```

I do not see a reopened blocker on the latest head.
Conclusion: approved from my side. CI still needs to finish green, but I do not have a new code-level concern to hold this back.

DanielLeens:
Thanks for the update. I rechecked the current head against the latest `dev` and the current merge-gate state.

- base branch: `dev`
- head SHA: `6208343894fe5d9c30b4ccc2d270d9e7e7bf8985`
- compare status: `diverged`
- ahead_by: `4`
- behind_by: `47`
- mergeable_state: `unstable`
- CI / merge gate snapshot: `Build [cancelled]`

From Daniel's side, the source-level conclusion from my latest approval still stands: I do not see a reopened blocker in the multi-table transform rule-selection / schema-change path.

At this point the visible gate is no longer a transform-specific source issue. The current `Build` signal is stale/cancelled on a branch that is already behind the latest `dev`, so the lowest-cost next step is to sync with the latest `dev` and rerun `Build` on the updated head first. If a fresh run still fails after the sync, please paste that new check link and I can narrow it down against the rebased branch.

davidzollo:
@goutamadwant  Good job. 

PR #11062, opened 2026-06-11
[Feature][Transform-V2] Add Calcite SQL transform plugin

nzw921rx:
## Purpose of this pull request

Closes #11061

Add a new SQL transform plugin powered by Apache Calcite, providing standard SQL capabilities (SELECT, WHERE, CASE WHEN, built-in functions, JSON extraction, etc.) for data transformation. Key components:

- **Calcite Engine**: `CalciteSQLEngine` compiles SQL once at startup and executes per-row at runtime, with bidirectional SeaTunnel-Calcite type mapping.
- **Shaded Dependency**: `seatunnel-calcite` module shading Calcite 1.38.0 under `org.apache.seatunnel.shade.*`.
- **UDF System**: `CalciteUdf` SPI for custom UDFs + `ZetaUdfBridge` for reusing existing ZetaUDF jars. Built-in UDFs: MASK, MASK_HASH, DES_ENCRYPT, DES_DECRYPT, URL_ENCODE, URL_DECODE.
- **Multi-table CDC**: Per-table SQL overrides, regex table matching, automatic plan rebuild on schema changes.

## Does this PR introduce any user-facing change?

Yes. Adds a new `Calcite` transform plugin. Example usage:

```hocon
transform {
  Calcite {
    plugin_input = "source_table"
    plugin_output = "result"
    sql = "SELECT id, UPPER(name) AS name, MASK(phone, 3, 7, '*') AS phone FROM source_table WHERE age >= 18"
  }
}
```

Full documentation added in `docs/en/transforms/calcite.md` and `docs/zh/transforms/calcite.md`.

## How was this patch tested?

```bash
#UT
./mvnw test -pl seatunnel-transforms-v2

#IT
./mvnw -DskipUT -DskipIT=false verify -pl seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-2

```

PR #11100, opened 2026-06-14
[Fix][E2E] Stabilize connector-file-local CI by retrying job and disabling JUnit parallel execution

nzw921rx:

Stabilize `connector-file-local-e2e` in CI when it flakes due to **Docker host port conflicts** (e.g. `Bind for 0.0.0.0:<port> failed: port is already allocated`). When a container fails to start, it can cascade into widespread test failures (e.g. `ParameterResolutionException`) and timeouts.

This PR avoids changing test/container implementation code and focuses on a **CI-only mitigation** for the affected job.

Key changes:

- Update `.github/workflows/backend.yml` (job: `connector-file-local-it`)
  - Add `nick-fields/retry@v3` (up to 3 attempts)
  - Disable JUnit 5 parallel execution via system properties:
    - `junit.jupiter.execution.parallel.enabled=false`
    - `junit.jupiter.execution.parallel.mode.default=same_thread`
    - `junit.jupiter.execution.parallel.mode.classes.default=same_thread`

No. CI-only change.

## How was this patch tested?

- CI validation:
  - `connector-file-local-it` should no longer fail intermittently with port allocation errors

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -1470,8 +1470,19 @@ jobs:
       - name: free disk space
         run: tools/github/free_disk_space.sh
       - name: run file local connector integration test
-        run: |
-          ./mvnw -B -T 1 verify -DskipUT=true -DskipIT=false -D"license.skipAddThirdParty"=true -D"skip.ui"=true --no-snapshot-updates -pl :connector-file-local-e2e -am -Pci
+        uses: nick-fields/retry@v3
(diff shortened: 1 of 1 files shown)


nzw921rx:
The issue will be fixed in PR https://github.com/apache/seatunnel/pull/11062, and it cannot be reproduced by rebasing other branches from dev🚀

PR #11102, opened 2026-06-15
[Fix][E2E] Stabilize multiple flaky tests

nzw921rx:

Fix three flaky/failing CI test issues:

- `ClassLoaderITBase`: `getClassLoaderUpperBound()` did not account for classloaders created by the first job in cache mode, causing assertion failure. Added `DISABLE_CACHE_MODE_CLASSLOADERS_PER_JOB` allowance to fix.
- `SeaTunnelEngineClusterRoleTest`:
  - Cancel-job assertion polled too aggressively after master handoff, causing flaky timeouts on slow CI. Added `pollDelay(5s)` + `pollInterval(2s)` to stabilize.
  - 4 test methods (`testStartMasterNodeWithTcpIp`, `testStartMasterNodeWithMulticastJoin`, `testCannotOnlyStartWorkerNodeWithTcpIp`, `testCannotOnlyStartWorkerNodeWithMulticastJoin`) used the default cluster name `seatunnel`, causing Hazelcast port conflicts with other test classes (e.g. `ConnectorPackageClientTest`) running in the same JVM fork. Added unique cluster names via `ContentFormatUtilTest.getClusterName()` to isolate them.
- `ConnectorPackageClientTest`: `testUploadConnectorPluginJars()` did not shut down its `SeaTunnelHazelcastClient`, leaking Hazelcast client connections. Wrapped the test body in try-finally to ensure `shutdown()` is always called.
- `SqlServerCDCIT`: `testWithSchemaEvolution` failed due to SQL Server deadlock (Error 1205) when `sp_cdc_enable_table` competed with Debezium for CDC metadata locks. `executeSqlFile()` had no retry mechanism. Added per-statement deadlock retry using Awaitility (up to 60s), consistent with the existing `dropTestDatabase()` pattern in the same class.

No. Test-only changes.

```bash
./mvnw -pl seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base -DskipUT -DskipIT=false -Dit.test=ClassLoaderEnableCacheModeIT verify
./mvnw -pl seatunnel-engine/seatunnel-engine-client -Dtest=SeaTunnelEngineClusterRoleTest test
./mvnw -pl seatunnel-engine/seatunnel-engine-client -Dtest=ConnectorPackageClientTest test
```

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-sqlserver-e2e/src/test/java/org/apache/seatunnel/e2e/connector/cdc/sqlserver/SqlServerCDCIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-sqlserver-e2e/src/test/java/org/apache/seatunnel/e2e/connector/cdc/sqlserver/SqlServerCDCIT.java
@@ -623,15 +623,21 @@ public class SqlServerCDCIT extends TestSuiteBase implements TestResource {
         try {
             List<String> statements =
                     parseStatements(Files.readAllLines(Paths.get(ddlTestFile.toURI())));
+            String currentDatabase = null;
             for (String stmt : statements) {
-                executeWithDeadlockRetry(stmt);
+                String trimmed = stmt.trim();
+                if (trimmed.toUpperCase().startsWith("USE ")) {
+                    currentDatabase = trimmed.substring(4).replaceAll(";\\s*$", "").trim();
+                    continue;
+                }
+                executeWithDeadlockRetry(stmt, currentDatabase);
             }
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
     }
 
-    private void executeWithDeadlockRetry(String sql) {
+    private void executeWithDeadlockRetry(String sql, String database) {
         Awaitility.await(
                         "Executing: "
                                 + sql.substring(0, Math.min(80, sql.length()))
@@ -642,6 +648,9 @@ public class SqlServerCDCIT extends TestSuiteBase implements TestResource {
                         () -> {
                             try (Connection connection = getJdbcConnection();
                                     Statement statement = connection.createStatement()) {
+                                if (database != null) {
+                                    statement.execute("USE " + database);
+                                }
                                 statement.execute(sql);
                                 return true;
                             } catch (SQLException e) {
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-sqlserver-e2e/src/test/java/org/apache/seatunnel/e2e/connector/cdc/sqlserver/SqlServerCDCIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-sqlserver-e2e/src/test/java/org/apache/seatunnel/e2e/connector/cdc/sqlserver/SqlServerCDCIT.java
@@ -620,16 +620,40 @@ public class SqlServerCDCIT extends TestSuiteBase implements TestResource {
         final String ddlFile = String.format("ddl/%s.sql", sqlFile);
         final URL ddlTestFile = TestSuiteBase.class.getClassLoader().getResource(ddlFile);
         Assertions.assertNotNull(ddlTestFile, "Cannot locate " + ddlFile);
-        try (Connection connection = getJdbcConnection();
-                Statement statement = connection.createStatement()) {
+        try {
             List<String> statements =
                     parseStatements(Files.readAllLines(Paths.get(ddlTestFile.toURI())));
             for (String stmt : statements) {
-                statement.execute(stmt);
+                executeWithDeadlockRetry(stmt);
             }
         } catch (Exception e) {
             throw new RuntimeException(e);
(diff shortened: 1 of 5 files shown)


DanielLeens (commented):
Thanks for working on this. I reviewed the full current head from the actual failing test paths instead of only looking at the two-line diff.

What this PR fixes
- User pain: `ClassLoaderEnableCacheModeIT` is currently treating normal cache warm-up as a leak, and `SeaTunnelEngineClusterRoleTest` is sampling cancellation too early and sometimes asserting against a transient state instead of the terminal one.
- Fix approach: the classloader assertion now allows the one-time cache-mode warm-up cost, and the cancel test now waits on the real terminal-state convergence window.
- One-line summary: this is a precise flaky-test fix that aligns the assertions with the real runtime lifecycle rather than weakening coverage.

Runtime chain I checked
```text
ClassLoader E2E
  -> submit job
  -> wait until /finished-jobs contains the current jobId
  -> assertClassLoaderStateEventuallyStable(...) [ClassLoaderITBase.java:169-181]
      -> getClassLoaderCount() [190-195]
      -> cache mode upper bound now allows initial + one warm-up cost [183-188]

Cluster role cancel test
  -> execute client job [SeaTunnelEngineClusterRoleTest.java:454-455]
  -> wait until RUNNING [456-463]
  -> cancelJob(jobId) [464]
  -> poll getJobStatus(jobId) until CANCELED [465-475]
      -> the intermediate CANCELING window is expected
```

Key findings
1. The cache-mode classloader change is on the real assertion path. It still guards against continued growth; it just stops treating the normal first-run cache cost as a leak.
2. The cancel-path change is also on the real lifecycle path. It does not lower the expected end state; it only stops asserting too early against a transient state.
3. I did not find a source-level blocker in the current head.

Test / stability note
- The updated test code looks structurally stable to me. There is no new fixed sleep in the assertion path; the change is an explicit polling window aligned with the state machine.

Local verification note
- I checked the full diff, the relevant test methods, and current GitHub metadata.
- At the time of review, GitHub `Build` was still queued.

### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- No additional source-level follow-up from this review. The remaining step is just to let the current `Build` finish and confirm the CI signal.

Overall, this is a focused and defensible fix. It keeps the original test intent, but updates the assertions to match the real runtime behavior.

davidzollo (dismissed):
+1 if CI passes.

DanielLeens (approved):
Thanks for working on this. I re-reviewed the latest head locally from the actual flaky-test paths instead of only checking whether the diff made CI greener.

What this PR solves
- User pain: several tests are failing intermittently because of environment timing and isolation issues rather than real product regressions.
- Fix approach: add deadlock-aware retry to the SQL Server E2E DDL path, isolate Hazelcast-based tests with dedicated cluster names, make the cancel-state assertion wait for state propagation more explicitly, and close a Hazelcast client in `finally` to avoid test pollution.
- One-line summary: this PR hardens a few genuinely flaky tests without weakening the assertions.

I also noticed @davidzollo already left an approval here. After tracing the same paths myself, I agree with that direction.

Full runtime path checked
CI test run -> SQL Server CDC E2E reads DDL statements -> `executeWithDeadlockRetry()` in `SqlServerCDCIT.java:619-657` retries only on actual deadlock errors -> the CDC/schema-evolution flow continues.
In parallel, Hazelcast-based tests boot master/worker/client clusters -> submit or cancel jobs -> poll final state. The changes in `SeaTunnelEngineClusterRoleTest.java:333-425` and `:453-462` isolate each test with its own cluster name and wait more explicitly for the terminal `CANCELED` state after cluster handoff. `ConnectorPackageClientTest.java:194-243` now closes the Hazelcast client in `finally`, which removes cross-test residue. `ClassLoaderITBase.java:169-180` also adjusts the expected upper bound for cache-mode classloaders so the test matches the actual runtime behavior instead of over-reporting a leak.

Core findings
- The modified code paths are the real execution paths of the flaky tests, not dead branches.
- The SQL retry remains narrow: non-deadlock SQL failures still fail the test immediately.
- I did not find any production API/config/protocol/serialization compatibility impact because all changes are test-only.
- This is a precise stability fix, not a blanket timeout inflation patch.

Test stability
- Stability rating: Stable.
- Reasons:
  - `SqlServerCDCIT.java:634-657` uses condition-based waiting instead of hard-coded sleeps and retries only the recoverable deadlock case.
  - `SeaTunnelEngineClusterRoleTest.java:333-425` assigns unique cluster names, which removes cross-test interference under CI parallelism.
  - `SeaTunnelEngineClusterRoleTest.java:453-462` uses explicit polling for terminal-state convergence instead of assuming immediate propagation.
  - `ConnectorPackageClientTest.java:194-243` fixes resource cleanup, which lowers follow-on contamination risk.

Merge conclusion
### Conclusion: can be merged
1. Blocking items
- None.
2. Non-blocking suggestions
- None.

Overall this looks good to me. The fixes are targeted and keep the original test intent intact.

davidzollo (approved):
+1

PR #11104, opened 2026-06-15
[Improve][Zeta] Add ReportMetricsOperation observability exports

dybyte:

This pull request adds lightweight observability exports for `ReportMetricsOperation` in Zeta workers.

`TaskExecutionService#updateMetricsContextInImap()` periodically reports local task metrics to the master and also reports them again when a task group completes. Under high-frequency job submission or many small batch jobs, this path can create control-plane pressure, but it is currently hard to observe directly.

This PR adds Prometheus exports for:
- `report_metrics_operation_total{address, result}`
- `report_metrics_operation_last_payload_task_count{address}`
- `report_metrics_operation_last_invocation_latency_ms{address}`
- `report_metrics_operation_max_invocation_latency_ms{address}`

Yes.

This PR adds new Prometheus telemetry metrics and updates the Zeta telemetry documentation in both English and Chinese. It does not change runtime behavior of jobs, but users monitoring the metrics endpoint will see new exported metrics compared with current released and unreleased branches.

- Added collector-level tests for `ReportMetricsOperationExports`
  - returns empty when `TaskExecutionService` is unavailable
  - exports metric families when stats are available
  - verifies `result` labels and exported sample values
- Ran:
  - `./mvnw -pl seatunnel-engine/seatunnel-engine-server spotless:apply`
  - `./mvnw -pl seatunnel-engine/seatunnel-engine-server -DskipTests verify`

### Check list


DanielLeens (commented):
Thanks for the contribution. I reviewed the latest head from the actual worker -> master metrics-reporting path, not just the file list.

What this PR solves
- User pain: today `TaskExecutionService` periodically sends worker-side task metrics to the master, but operators cannot see from Prometheus whether `ReportMetricsOperation` is succeeding, failing, getting interrupted, how large the last payload was, or how long the last call took.
- Fix approach: the PR records success / failure / interrupted counts plus last / max latency around `updateMetricsContextInImap()`, then exposes those values through a new `ReportMetricsOperationExports` collector and documents the metrics in both EN and ZH telemetry docs.
- One-line summary: this is a targeted observability improvement on the real metrics backup path, and I did not find a source-level blocker in the current head.

Runtime chain I checked
```text
worker startup
  -> TaskExecutionService.<init>() [156-170]
      -> scheduledExecutorService.scheduleAtFixedRate(this::updateMetricsContextInImap, ...)

periodic metrics backup
  -> TaskExecutionService.updateMetricsContextInImap() [602-638]
      -> collectLocalMetricsMap() [678-690]
      -> new ReportMetricsOperation(localMetricsMap) [605-612]
      -> invoke.get()
         -> success: recordReportMetricsOperationSuccess() [641-644]
         -> InterruptedException: recordReportMetricsOperationInterruption() [651-655]
         -> Exception: recordReportMetricsOperationFailure() [646-649]
      -> updateReportMetricsOperationObservability() [657-662]

Prometheus export
  -> ExportsInstanceInitializer registers ReportMetricsOperationExports
  -> ReportMetricsOperationExports.collect() [37-85]
      -> getServer().getTaskExecutionService()
      -> getReportMetricsOperationStats() [668-675]
      -> export success/failure/interrupted counters
      -> export last payload / last latency / max latency gauges
```

Key findings
1. The normal runtime path really does hit this code. These counters are updated from the same scheduled worker path that already sends `ReportMetricsOperation`, so this is not dormant telemetry.
2. The PR does not change the `ReportMetricsOperation` payload contract or the master-side consumption path. It only adds worker-local observability around an existing RPC.
3. I did not find a correctness blocker in the current source path.

Two non-blocking suggestions
1. `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/telemetry/metrics/exports/ReportMetricsOperationExports.java:31`, `.../ReportMetricsOperationStats.java:25`, and `.../TaskExecutionService.java:143` add new engine-core telemetry classes / fields / accessor semantics without comments. It would be worth adding short Javadocs / field comments so future maintainers can see at a glance which values are lifecycle totals vs last-sample gauges.
2. The new test in `seatunnel-engine/.../TelemetryCollectorCoordinatorGuardTest.java:187-349` verifies exporter rendering with a mocked `TaskExecutionService`, but it does not exercise the real `updateMetricsContextInImap()` success / failure / interrupted branches. A small low-level test for the recorder path would make this much safer against future regressions.

Test / stability note
- The added test code itself looks stable to me: it is pure mock-based unit coverage, with no sleeps, no external services, and no obvious flaky patterns.
- At the time of review, GitHub `Build` was still in progress.

### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- Add the missing telemetry comments.
- Add one focused test for the real recorder/update path in `TaskExecutionService`.

Overall, this is a focused and useful observability enhancement. The code path is defensible, the docs are updated, and I do not see a source-level reason to block merge.

seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/TaskExecutionService.java
```diff
@@ -753,26 +599,82 @@ private void updateMetricsContextInImap() {
             return;
         }
 
+        HashMap<TaskLocation, SeaTunnelMetricsContext> localMetricsMap = collectLocalMetricsMap();
+        long invocationStartNanos = System.nanoTime();
```
CosmosNi, line 603: suggest：Move invocationStartNanos = System.nanoTime() above collectLocalMetricsMap().

DanielLeens: Thanks for calling this out. I agree with moving `invocationStartNanos` above `collectLocalMetricsMap()`, and the latest head now does exactly that in `TaskExecutionService.updateMetricsContextInImap(...)`. With that change, the exported latency matches the updated docs and now covers both local metrics collection and the worker-to-master invocation.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the real worker -> master metrics-reporting path again, not just the file list.

+1 to @CosmosNi's earlier timing-scope point. The latest head now moves `invocationStartNanos` above `collectLocalMetricsMap()`, so the exported latency really matches the updated docs: it includes both local metrics collection and the worker-to-master invocation.

# What this PR solves
- User pain: today `TaskExecutionService` periodically reports worker-side task metrics to the master, but operators still cannot see from Prometheus whether `ReportMetricsOperation` is succeeding, failing, getting interrupted, or how large / slow the latest report was.
- Fix approach: the PR records success / failure / interrupted counts plus last / max latency around the real reporting path, and exposes them through `ReportMetricsOperationExports`.
- One-line summary: this is a focused observability improvement on the real runtime path, and the latest head now aligns the timing semantics, comments, and docs.

# Runtime chain rechecked
```text
worker startup
  -> TaskExecutionService.<init>() [L157-176]
      -> scheduledExecutorService.scheduleAtFixedRate(this::updateMetricsContextInImap, ...)

periodic reporting
  -> TaskExecutionService.updateMetricsContextInImap() [L603-639]
      -> invocationStartNanos = System.nanoTime()
      -> collectLocalMetricsMap()
      -> new ReportMetricsOperation(localMetricsMap)
      -> invoke.get()
         -> success / failure / interrupted recorder
         -> updateReportMetricsOperationObservability(...)

Prometheus export
  -> ReportMetricsOperationExports.collect() [L68-86]
      -> getReportMetricsOperationStats()
      -> export counters + last/max latency gauges
```

# Key findings
1. The latest head addresses the timing-scope thread comment correctly. The recorded latency now covers the whole worker-side reporting path, not only the RPC after local collection.
2. The missing comments I called out in the last round are now added on the core telemetry fields / accessor classes.
3. I did not find a source-level blocker on the latest head.

# Tests
- The added / existing new test coverage around the exporter path still looks stable to me: no sleeps, no external services, no obvious flaky pattern.
- The remaining blind spot is that the recorder branches in `updateMetricsContextInImap()` are still not directly unit-tested, but I do not treat that as a merge blocker on this head.

# CI
- `Notify test workflow` and `labeler` are green.
- `Build` is still running on the latest head, so I would wait for that result before merge.

1. Blocking items
- No source-level blocker from my side.

2. Suggested follow-up
- A small low-level test for the success / failure / interrupted recorder path would make future regressions easier to catch, but I do not consider it blocking here.

The latest revision looks good to me from the source path.

davidzollo (approved):
Thanks for the contribution. This is a genuinely useful addition for operators — the ReportMetricsOperation path has always been a control-plane blind spot, and having Prometheus exports for call counts, error rates, and latency makes it actionable for the first time.

I looked through the runtime chain and the fix from @CosmosNi's suggestion is correct — the exported latency now accurately reflects the full reporting cycle, which matches the updated documentation.


Overall the code path is sound and the scope is focused. +1 from my side

PR #11105, opened 2026-06-15
[Fix][Docs] Remove duplicate connector FAQ sidebar entry

DanielCarter-stack:
### Purpose
Remove the duplicated `connectors/connector-faq` entry from the docs sidebar so the Connectors menu does not show the FAQ twice.

No.

- `./mvnw spotless:apply`
- Not run: `./mvnw -q -DskipTests verify` (skipped per requester instruction)

### DIFF
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -142,7 +142,6 @@ const sidebars = {
                 "connectors/connector-isolated-dependency",
                 "connectors/connector-faq",
                 "connectors/cdc-production-cookbook",
-                "connectors/connector-faq",
                 {
                     "type": "category",
                     "label": "Source",


DanielLeens (approved):
Thanks for the contribution. I reviewed the current diff locally from the actual Docusaurus sidebar path rather than treating this as a cosmetic-only change.

What this PR solves
- User pain: the docs sidebar shows the Connector FAQ twice, which makes the navigation look broken even though the page content itself is fine.
- Fix approach: remove the duplicate `connectors/connector-faq` entry from `docs/sidebars.js`.
- One-line summary: this is a precise docs cleanup that removes a real duplicate entry without changing any content or URLs.

Call path checked
Docusaurus build -> load `docs/sidebars.js` -> render the `Connectors` sidebar items -> show FAQ entry.
Before this patch, that path rendered the same FAQ item twice. After this patch, it renders once. The change is isolated to the static sidebar source and does not touch runtime code, configuration, API, serialization, or task lifecycle.

Core logic
- The only functional change is at `docs/sidebars.js:142-145`, where the second duplicate FAQ item is removed.
- Normal user navigation will definitely hit this path because the sidebar is generated directly from this file.
- I did not find any compatibility or side-effect risk.

Testing / docs
- No new runtime logic is introduced, so I do not see a missing unit-test concern here.
- This PR itself is the docs update, and it keeps the shared sidebar config internally consistent.

Merge conclusion
### Conclusion: can be merged
1. Blocking items
- None.
2. Non-blocking suggestions
- None.

Overall, this is small but correct, and it fixes a user-visible docs issue in the most direct way.

dybyte (approved):

PR #11108, opened 2026-06-16
[Feature][Connector-V2][CDC] Add include/exclude filtering for schema change event types  #11044

### LINKED ISSUE #11044: [Feature][Connector-V2][CDC] Add include/exclude filtering for schema change event types
## Background
Even when schema evolution is enabled, users often do not want every schema change event to propagate through the job. Practical CDC jobs sometimes need to allow or block selected event classes.

Right now SeaTunnel does not provide a first-class filter for this.

## Scope
Add include/exclude filtering for CDC schema change event types.

This issue is about **event-type filtering only**. It does **not** define the overall schema change behavior policy, which is tracked separately.

## First delivery boundary
The first delivery should align with the **schema change events that SeaTunnel already emits today** instead of inventing names for event families that are not yet modeled end to end.

That means the first implementation should filter the current SeaTunnel schema event model first, and defer broader event families such as `create.table`, `truncate.table`, or `drop.table` until those events exist as stable SeaTunnel schema change events.

## Canonical event names for first delivery
The first delivery should use a stable SeaTunnel-owned canonical name set with an explicit mapping to the current event model:

- `add.column` -> `SCHEMA_CHANGE_ADD_COLUMN`
- `drop.column` -> `SCHEMA_CHANGE_DROP_COLUMN`
- `modify.column` -> `SCHEMA_CHANGE_MODIFY_COLUMN`
- `change.column` -> `SCHEMA_CHANGE_CHANGE_COLUMN`
- `update.columns` -> `SCHEMA_CHANGE_UPDATE_COLUMNS`
- `rename.table` -> `SCHEMA_CHANGE_RENAME_TABLE`

This mapping should be documented in the issue implementation and docs so contributors do not need to guess how user-facing names relate to internal event classes.

### 1. Add explicit config surface
Introduce include/exclude config keys for schema change event types.

To avoid collision with lower-level source properties, the config names should be SeaTunnel-owned and explicit, for example:
- `schema.change.include = [...]`
- `schema.change.exclude = [...]`

Exact naming can still be refined, but it should avoid reusing ambiguous names such as `include.schema.changes`.

### 2. Normalize schema change events to canonical names
Primary touch points are expected to include:
- the current `SchemaChangeEvent` / `EventType` mapping layer
- schema change dispatch path before sink coordination

Users should not need to know internal Java class names, enum names, or connector-specific parser details.

### 3. Apply filtering before downstream coordination
Filtering should be applied after schema change events are normalized to canonical SeaTunnel names, but before schema-change checkpoint / barrier coordination and before sink-side schema evolution handling.

This keeps the mechanism connector-independent and prevents filtered events from entering downstream schema coordination paths.

### 4. Define precedence and validation explicitly
The first delivery should define one deterministic precedence rule:
- if `schema.change.include` is set, only included event types are eligible
- then `schema.change.exclude` is applied
- `exclude` wins when the same event type appears in both lists

Validation should also be strict:
- unknown event names fail fast
- duplicate names should be normalized and handled deterministically

### 5. Defer broader event families explicitly
If contributors later add stable SeaTunnel event models for `create.table`, `truncate.table`, `drop.table`, or comment-related changes, those should be added in follow-up work rather than inferred in this issue.

## Acceptance criteria
- CDC jobs accept include/exclude configuration for the current stable SeaTunnel schema change event set.
- The user-facing canonical names and their mapping to internal event types are documented explicitly.
- Filtering is applied deterministically before downstream schema change coordination.
- Unknown event names fail fast with clear messages.
- English and Chinese docs are updated with examples.
- Add focused tests for include-only, exclude-only, combined precedence, and unknown-name validation.

## Non-goals
- Defining the overall schema change behavior policy modes.
- Expanding the SeaTunnel schema event model to new event families not yet emitted today.
- Startup mode enhancements.
- Metadata enrichment.

ss666:

Closes #11044. Adds include/exclude filtering for CDC schema change event types, so users can control
which schema changes propagate downstream.

- New source options `schema-changes.include` / `schema-changes.exclude`. Wired into MySQL / Oracle / SqlServer CDC.
- Filtering lives in `connector-cdc-base`, applied after events are normalized to the canonical model and
  **before** downstream emission (before checkpoint/barrier coordination and sink-side schema evolution).
- Deterministic precedence: (1) if `include` is non-empty, only included types are eligible; (2) then
  `exclude` is applied; (3) `exclude` wins on conflict. Unknown names fail fast with a message listing valid
  names; names are trimmed, case-insensitive, de-duplicated.

Yes.

- Two new optional CDC source options `schema-changes.include` / `schema-changes.exclude`
  (default empty = no filtering, fully backward compatible).
- en/zh docs for MySQL / Oracle / SqlServer CDC updated with a canonical-name table and a HOCON example.
- Known limitation: excluding `drop.column` keeps the column in the sink schema while the source stops supplying it. 
    - a **nullable** retained column is written as `NULL` (works);
    - a **NOT NULL** retained column has its `NULL` write rejected by the sink and the job fails. Deciding what
  to write is *data-handling behavior*, out of #11044's event-type-filtering scope. See the follow-up below.

- Format: `./mvnw spotless:apply`
- Unit: `SchemaChangeEventFilterTest` — 13 cases (include-only / exclude-only / precedence / group alias /
  unknown-name fail-fast / normalization + dedup / mapping completeness). `Tests run: 13, Failures: 0, Errors: 0`.
- E2E: `MysqlCDCWithSchemaChangeIT#testMysqlCdcSchemaChangeEventTypeFilter` — add.column propagates, excluded
  drop.column stays in the sink, concurrent INSERT/UPDATE/DELETE still reach the sink. `Tests run: 1, Failures: 0, Errors: 0`.
- Compile: `./mvnw clean install -pl :connector-cdc-base,:connector-cdc-mysql -am -DskipTests`. 

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md)
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR. (N/A — backward compatible, options default to empty)
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) — N/A (no new connector)
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml) — N/A
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml) — N/A
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/) — done (`MysqlCDCWithSchemaChangeIT`)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config) — N/A

### Follow up

The NOT NULL limitation above is out of #11044's scope (it is *data-handling behavior*, not event-type
filtering), but it is still necessary to make exclude(drop.column) usable for NOT NULL columns. Either of the two mechanisms below might fix it and could be proposed as a dedicated follow-up:

- **Option A — relax the retained column to nullable (derived schema change).** When `exclude` suppresses
  the `drop.column` of a NOT NULL column, emit a derived "modify column → nullable" event downstream so the
  retained sink column legitimately accepts `NULL`. This changes the *constraint*, not the data, and is mostly contained in `connector-cdc-base`. It actually makes `exclude(drop.column)` work for NOT NULL
  columns instead of merely failing more politely.
- **Option B — opt-in, user-declared fill value.** Add a per-column option (e.g.
  `schema-changes.drop-column.fill`) the user explicitly sets for a retained-but-no-longer-supplied NOT NULL
  column. Because the value is user-declared, it is *not* silent fabrication — the user owns the semantics.
  When unset, the job still fails fast.

### DIFF
--- a/docs/en/connectors/source/MySQL-CDC.md
+++ b/docs/en/connectors/source/MySQL-CDC.md
@@ -215,6 +215,8 @@ When an initial consistent snapshot is made for large databases, your establishe
 | exactly_once                              | Boolean  | No       | false   | Enable exactly once semantic.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
 | format                                    | Enum     | No       | DEFAULT | Optional output format for MySQL CDC, valid enumerations are `DEFAULT`、`COMPATIBLE_DEBEZIUM_JSON`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
 | schema-changes.enabled                    | Boolean  | No       | false   | Schema evolution is disabled by default. Now we only support `add column`、`drop column`、`rename column` and `modify column`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
+| schema-changes.include                     | List     | No       | -       | Only the listed schema change event types are sent downstream (when `schema-changes.enabled = true`). Empty means all are eligible. See [Schema change event filtering](#schema-change-event-filtering).                                                                                                                                                                                                                                                                                                                                                                                                              |
+| schema-changes.exclude                     | List     | No       | -       | Schema change event types listed here are NOT sent downstream. Applied after `schema-changes.include`; exclude wins on conflict. See [Schema change event filtering](#schema-change-event-filtering).                                                                                                                                                                                                                                                                                                                                                                                                                  |
 | debezium                                  | Config   | No       | -       | Pass-through [Debezium's properties](https://github.com/debezium/debezium/blob/v1.9.8.Final/documentation/modules/ROOT/pages/connectors/mysql.adoc#connector-properties) to Debezium Embedded Engine which is used to capture data changes from MySQL server.                                                                                                                                                                                                                                                                                                                                                        |
 | int_type_narrowing                        | Boolean  | No       | true    | Int type narrowing, if true, the tinyint(1) type will be narrowed to the boolean type if without loss of precision. Support for MySQL at now. Please refer to `int_type_narrowing` below                                                                                                                                                                                                                                                                                                                                                                                                                             |
 | common-options                            |          | no       | -       | Source plugin common parameters, please refer to [Source Common Options](../common-options/source-common-options.md) for details                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
@@ -340,6 +342,44 @@ sink {
 }
 
 ```
+
+### Schema change event filtering
+
+When `schema-changes.enabled = true`, you can further control which schema change event types are
+propagated downstream using `schema-changes.include` / `schema-changes.exclude`. 
+
+Use these SeaTunnel-owned canonical names:
+
+| Canonical name  | Operation                                                                 |
+|-----------------|---------------------------------------------------------------------------|
+| `add.column`    | add a column                                                              |
+| `drop.column`   | drop a column                                                             |
+| `modify.column` | change a column's type/attributes, name unchanged                         |
+| `change.column` | rename a column, optionally re-type                                       |
+| `update.columns`| group alias for all four column-level changes above                        |
+| `rename.table`  | rename a table                                                            |
+
+Precedence is deterministic:
+
+1. if `schema-changes.include` is set, only included event types are eligible;
+2. `schema-changes.exclude` is then applied;
+3. **exclude wins** when a type appears in both lists.
+
+```hocon
+source {
+  MySQL-CDC {
+    # ...
+    schema-changes.enabled = true
+    schema-changes.include = ["add.column", "drop.column"]
+    schema-changes.exclude = ["change.column"]
+  }
+}
+```
+
+**Data handling when `drop.column` is excluded. For a retained **NOT NULL** column the `NULL` write is rejected
+by the sink, so excluding `drop.column` for a NOT NULL column that the source has stopped supplying
+will fail at the sink.
+
 ### Support table-pattern for multi-table reading
 
 > `table-pattern` and `table-names` are mutually exclusive
--- a/docs/en/connectors/source/MySQL-CDC.md
+++ b/docs/en/connectors/source/MySQL-CDC.md
@@ -357,7 +357,6 @@ Use these SeaTunnel-owned canonical names:
 | `modify.column` | change a column's type/attributes, name unchanged                         |
 | `change.column` | rename a column, optionally re-type                                       |
 | `update.columns`| group alias for all four column-level changes above                        |
-| `rename.table`  | rename a table                                                            |
 
 Precedence is deterministic:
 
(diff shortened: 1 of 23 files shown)


DanielLeens (changes requested):
Thanks for working on this. I went through the full diff and the column-level filtering idea itself makes sense: the new flow is `FactoryUtil.createAndPrepareSource(...) -> source factory -> createDebeziumDeserializationSchema(...) -> SchemaChangeEventFilter.fromConfig(config) -> SeaTunnelRowDebeziumDeserializeSchema.deserializeSchemaChangeRecord(...)`, and for cases like `drop.column` the filter is applied before the produced schema is updated, which is exactly the right place.

I did find one blocker before this can merge:

1. `rename.table` is now exposed as a supported canonical filter value, but the runtime path does not keep the internal table mapping in sync if that event is filtered out. In `SeaTunnelRowDebeziumDeserializeSchema.deserializeSchemaChangeRecord(...)`, a filtered event returns early before `tables` and `tableRowConverters` are updated. Later, `deserializeDataChangeRecord(...)` resolves the current `TablePath` from the source record and, in multi-table mode, looks up the converter by that table path. After a source-side table rename, filtering out `rename.table` leaves the old key in `tableRowConverters`, so the subsequent DML for the renamed table is treated as an unknown table and skipped. The important detail here is that this is not just a test gap: `AlterTableSchemaEventHandler` also returns the schema unchanged for `AlterTableNameEvent`, so there is no evidence that `rename.table` is supported end-to-end even without filtering. Given that this PR documents `rename.table` as a valid public value, I would treat that as a blocker. I would either remove `rename.table` from the exposed canonical-name list for now, or fully implement the table-path remap path and add a multi-table integration test for it.

2. One non-blocking follow-up: `schema-changes.include` / `schema-changes.exclude` are exposed in the source factories as plain optional lists, but canonical-name validation only happens later inside `SchemaChangeEventFilter.fromConfig(...)`. That means a typo such as `rename.tabble` would bypass submission-time option validation and fail later during source initialization. It would be better to add an `optionRule` validator so invalid names fail during job submission.

So my conclusion on the current head is: the column-level part looks good, but `rename.table` should not be advertised as a supported filterable event until the table-identity path is actually handled end-to-end.

DanielLeens (approved):
Thanks for the follow-up. I re-reviewed the latest head end to end, including the source-factory validation path and the runtime schema-change filter path.

# What this PR solves
- User pain: users can configure `schema-changes.include` / `schema-changes.exclude`, but the previous revision exposed `rename.table` as a public canonical value even though table rename is not implemented end to end, and typos were only discovered later during source initialization.
- Fix approach: the latest head removes `rename.table` from the public canonical-name list and from the EN/ZH docs, and adds `SchemaChangeEventFilter.validateOptions(...)` at `BaseChangeStreamTableSourceFactory.createSource(...)` / `restoreSource(...)` so invalid names fail during job submission.
- One-line summary: the feature surface is now aligned with the runtime capability that actually exists today, and config mistakes fail at the right boundary.

# Runtime chain rechecked
```text
job submission / restore
  -> BaseChangeStreamTableSourceFactory.createSource(...) [L48-53]
  -> BaseChangeStreamTableSourceFactory.restoreSource(...) [L55-61]
      -> SchemaChangeEventFilter.validateOptions(...)
          -> SchemaChangeEventType.fromCanonicalNames(...)
          -> invalid canonical names fail fast before the source is built

CDC runtime
  -> SchemaChangeEventFilter.fromConfig(...)
  -> SeaTunnelRowDebeziumDeserializeSchema.deserializeSchemaChangeRecord(...)
      -> column-level events are filtered normally
      -> table-level branch stays defensive only
```

# Key findings
1. The previous blocker is resolved. `rename.table` is no longer advertised as a supported public value, so the PR no longer exposes a capability that the runtime cannot honor.
2. The earlier follow-up about submission-time validation is also resolved. Unknown values now fail from the source-factory entrypoint instead of surfacing later during source initialization.
3. I did not find a new blocker on the latest head.

# Tests
- The added unit coverage around `rename.table` rejection and `validateOptions(...)` is focused and stable.
- I do not see a flaky-test pattern in the new tests.

# CI
- `Build`, `Notify test workflow`, and `labeler` are green on the latest head.

### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- None.

This closes the two review items I raised earlier, and the current head looks consistent to merge.

davidzollo (approved):
+1
LGTM
A warm welcome for your very first contribution! Thank you so much for your participation and efforts, we’re looking forward to more great contributions from you in the future.

PR #11113, opened 2026-06-17
[Fix][Connector-V2] Fix TiDB CDC silent row event loss when resolvedT…

### LINKED ISSUE #11013: [Bug] [Connector-V2] [TiDB-CDC] resolvedTs keeps advancing but row change events are silently missed
> This looks related to #8815, but this case happens on SeaTunnel 2.3.13 with Flink engine and the normal TiDB-CDC connector, not TiDB-CDC-MIGRATE.
>
> Compared with #8815, this case has additional evidence that TiDBSourceReader / CDCClient kept advancing resolvedTs and checkpoints kept succeeding, while row events were not emitted downstream.

### Search before asking

### What happened

A TiDB-CDC job stayed RUNNING and checkpoints kept succeeding, but the target table stopped receiving new row changes.

The source table continued to receive inserts/updates after the cutoff time, while the target table remained stale.

In this case:

- Target table stopped at `max(updated_at) = 2026-06-03 18:59:55`.
- Source table continued changing until at least `max(updated_at) = 2026-06-05 11:27:36`.
- There were 1183 source rows changed after the cutoff time.
- The Flink job remained RUNNING.
- Checkpoints kept completing successfully.
- No sink errors or backpressure were observed.
- Flink metrics stopped increasing:
  - Source `numRecordsOut = 91746`
  - Sink `numRecordsIn = 91746`
  - SourceReceivedQPS = 0.0
  - SinkWriteQPS = 0.0
  - Sink error count = 0

TaskManager logs show that `TiDBSourceReader` and `CDCClient` continued advancing `resolvedTs` after the target stopped receiving rows:

TiDBSourceReader - Capture streaming event from resolvedTs:...
CDCClient - handle resolvedTs: ..., regionId: ...
TiDBSourceReader - Capture streaming event next resolvedTs:...

### SeaTunnel Version

2.3.13

### SeaTunnel Config

```conf
env {
  parallelism = 1
  job.mode = "STREAMING"
  job.name = "dsp-17-v2-wzb_test_deposit_applications"
  checkpoint.interval = 300000
  checkpoint.timeout = 600000
  checkpoint.mode = "EXACTLY_ONCE"
  restart-strategy = "failure-rate"
  restart-strategy.failure-rate.max-failures-per-interval = 10
  restart-strategy.failure-rate.failure-rate-interval = "300 s"
  restart-strategy.failure-rate.delay = "10 s"
}

source {
  TiDB-CDC {
    plugin_output = "src"
    url = "jdbc:mysql://*****:4001/alpha_online"
    pd-addresses = "*****:2379,*****:2379,*****:2379"
    username = "******"
    password = "******"
    database-name = "alpha_online"
    table-name = "deposit_applications"
    startup.mode = "initial"
  }
}

sink {
  Jdbc {
    source_table_name = "src"
    url = "jdbc:mysql://*****:4000/?useSSL=false&useUnicode=true&characterEncoding=utf8"
    user = "root"
    password = "******"
    batch_size = 1000
    database = "sync_test"
    table = "deposit_applications"
    primary_keys = ["id"]
    generate_sink_sql = true
    support_upsert_by_query_primary_key_exist = true
  }
}
```

### Running Command

```shell
The job was submitted by our internal DataT platform to Flink on YARN.

Runtime application:
application_1756448821197_5312

Flink job name:
dsp-17-v2-wzb_test_deposit_applications

The platform uses SeaTunnel Flink starter with the TiDB-CDC source plugin.
The exact generated SeaTunnel config is provided in the "SeaTunnel Config" section above.

This was not started by manually running start-seatunnel-flink-*.sh.
```

### Error Exception

```log
No exception was thrown.

The job stayed RUNNING and checkpoints kept succeeding.
No sink errors or backpressure were observed.

The issue is silent data loss / silent data miss:
TiDBSourceReader and CDCClient kept advancing resolvedTs, but row change events after 2026-06-03 18:59:55 were not emitted downstream.
```

Flink 1.17.1

### Java or Scala Version

Java 8，OpenJDK 1.8.0_422

### Screenshots

[jobmanager-log-excerpt.log](https://github.com/user-attachments/files/28624318/jobmanager-log-excerpt.log)

[taskmanager-log-excerpt.log](https://github.com/user-attachments/files/28624325/taskmanager-log-excerpt.log)

### Code of Conduct

zhaoysg:
…s advances ahead of prewrite replay (#11013)

The TiDB CDC source could silently drop row change events while `resolvedTs` kept advancing, as reported in #11013. The root cause is in the CDC streaming reader: `captureStreamingEvents` advanced `resolvedTs` to `cdcClient.getMaxResolvedTs()` at the end of every batch, but the corresponding commit/prewrite pair was consumed via `commits.pollFirstEntry().getValue()` without verifying that the matching prewrite row had actually been buffered. When a commit row arrived before its prewrite (which happens during region split, rebalance, or under backpressure), the commit was discarded while `resolvedTs` was still pushed forward past the commit timestamp, causing the engine to skip subsequent real events at that timestamp without any error.

This commit fixes the issue and adds minimal diagnostic logging to help users diagnose similar scenarios in the future.

Changes
-------

`TiDBSourceReader.captureStreamingEvents` / `flushRows`
- `flushRows` now returns a `safeResolvedTs`. If a commit row has no matching prewrite row in `preWrites` yet, the function stops advancing `resolvedTs` to the commit timestamp and holds the commit in the queue, so the next batch can re-pair it with the late prewrite.
- `flushRows` now only `pollFirstEntry()`s a commit row after confirming its prewrite is present, fixing the previous poll-then-leak behaviour.
- The batch loop distinguishes between a "no row right now" return from `cdcClient.get()` and a "resolvedTs advanced but no row" case, and no longer exits early when only `resolvedTs` moved.

`TiDBSourceReader.snapshotEvents`
- The snapshot scan no longer relies on a single `KVClient.scan` call over the whole `keyRange`. It now uses a `RegionStoreClient` per region and walks forward using `region.getEndKey()`, which is correct across region boundaries and stops cleanly when the scan reaches the end of the requested key range.
- Adds `scannedCount` / `emittedCount` log on snapshot completion.

`TableKeyRangeUtils`
- `isRecordKey` now guards against short keys (`key.length > 10`) to avoid `ArrayIndexOutOfBoundsException` on truncated payloads.
- Switches from Guava `ImmutableList` to JDK collections to keep the module free of Guava dependencies on this hot path.

`TiDBSourceSplitEnumerator`
- Adds structured INFO logs for enumerator open, split enumeration, and per-reader split assignment to make split distribution observable in production.

`TiDBSourceReader`
- Adds `[TiDB-CDC-DIAG]` streaming stats logging (every 10s or when a single batch exceeds 1s) exposing `resolvedLagMs`, polled/emitted/committed counters, and pull/flush/emit latencies. This is the primary signal users can use to detect a prewrite/commit imbalance in the future.

Fixes #11013

### Check list


DanielLeens (commented):
Thanks for working on this. I reviewed the latest head from the real TiDB CDC snapshot/streaming reader path, not just from the new log lines.

What this PR fixes
- User pain: when `resolvedTs` advances ahead of the matching prewrite replay, the old streaming path can process commits too aggressively and lose the chance to emit the corresponding row correctly. The snapshot path was also brittle around region-based scanning.
- Fix approach: the PR rewrites the TiDB reader's snapshot loop, hardens the streaming poll loop, and changes `flushRows(...)` so an unmatched commit is held back with a lowered `safeResolvedTs` instead of being consumed immediately.
- One-line summary: the direction is reasonable, but this is a data-correctness fix on the main CDC path, and I still need a direct regression proof for the key reordered-event scenario before I can accept it.

Actual runtime path
```text
TiDBSourceReader.pollNext(...) [TiDBSourceReader.java:145-159]
  -> snapshotEvents(...) for INITIAL startup [162-219]
  -> captureStreamingEvents(...) [221-305]
      -> CDCClient.get()
      -> handleRow(...) [363-387]
          -> PREWRITE / COMMIT / COMMITTED stored into preWrites / commits
      -> flushRows(resolvedTs) [389-410]
          -> old path: poll commit immediately once commitTs <= resolvedTs
          -> new path: if matching prewrite is still missing, keep the commit,
             lower safeResolvedTs, and stop advancing past that point
```

Main findings
1. The core idea in `flushRows(...)` is correct: do not permanently consume a commit before its matching prewrite is available.
2. This PR is not a tiny one-line guard. It changes the snapshot loop, the streaming poll loop, the resolved-ts advancement behavior, and also adds unrelated JDBC sink diagnostics.
3. For a CDC data-correctness fix, the current version still misses the most important proof: a deterministic regression test that reproduces "commit visible first, prewrite arrives later, row is still emitted correctly".

Blocking issue
1. The PR still lacks a focused regression test for the exact reordered-event path it claims to fix.
   - Location:
     `TiDBSourceReader.java:221-410`
   - Why this matters:
     this is the real hot path for the claimed fix. `captureStreamingEvents(...) -> handleRow(...) -> flushRows(...)` now has different semantics around commit retention and resolved-ts advancement, but the PR does not add a reader-level or behavior-level test that proves the new logic works when the matching prewrite arrives after the commit has already been seen.
   - Risk:
     this is a CDC correctness path. Without a targeted regression, we are still trusting a fairly large rewrite on reasoning alone.
   - Recommended fix:
     Option A: add a focused reader/integration regression test that drives the sequence "commit first, resolvedTs advances, prewrite later" and asserts the row is emitted on the next round.
     Option B: if the current harness cannot drive the whole reader easily, at least add a smaller behavior test around `handleRow(...)` / `flushRows(...)` state transitions.
   - Severity: High

Non-blocking issue
2. The JDBC sink diagnostics broaden this PR beyond the minimal TiDB CDC fix.
   - Location:
     `JdbcOutputFormat.java:45-192`
   - Why this matters:
     the PR title and the main runtime fix are about TiDB CDC source correctness, but the diff also includes JDBC sink flush diagnostics. That extra scope makes the review and CI attribution noisier than necessary.
   - Suggested follow-up:
     ideally split the JDBC diagnostics into a separate PR, or at least explain very clearly in the PR description why they must ship together.
   - Severity: Medium

Tests and CI
- I do not see a new targeted TiDB CDC regression test in this PR.
- `Build` is currently red, and the failing jobs are `updated-modules-integration-test-part-1 (11, ubuntu-latest)` and `updated-modules-integration-test-part-2 (11, ubuntu-latest)`.
- Those are much closer to the touched modules than a random unrelated job, so without deeper log evidence I cannot safely dismiss them as unrelated noise.

Blocking items (must fix)
- Issue 1 above: please add a focused regression proof for the reordered `resolvedTs` / prewrite / commit path before merge.

Suggested follow-up (non-blocking)
- Issue 2 above: consider splitting the JDBC sink diagnostics out of this PR.

Overall assessment
- The core fix direction makes sense.
- What is still missing is the proof layer. For a CDC row-loss bug, that proof should come from a targeted regression, not just from a broad code rewrite plus generic CI.
- Once the regression is in place and the red `Build` jobs are explained, I am happy to re-review the latest head.

zhaoysg:
感谢详细 review。

我已经补了一个针对 TiDB CDC 乱序事件路径的聚焦回归测试：
`TiDBSourceReaderTest#flushRowsShouldHoldCommitUntilMatchingPrewriteArrives`。

这个测试覆盖 review 中提到的关键状态转换：
1. 先收到 COMMIT；
2. resolvedTs 已经超过 commitTs 时调用 flushRows，并验证 commit 会被保留，同时 safeResolvedTs 降到 commitTs - 1；
3. 后续收到匹配的 PREWRITE；
4. 再次调用 flushRows，验证 prewrite row 会被发出，并且 pending commit/prewrite 状态被清空。

本地验证：
- `./mvnw.cmd -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb test`
- `./mvnw.cmd -q -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb -DskipTests verify`

以上两个命令均通过。

我也尝试了全仓 `./mvnw.cmd -q -DskipTests verify`，但在解析 `org.apache.flink:flink-statebackend-rocksdb:1.20.1` 时 Maven Central TLS handshake/download 失败，构建在到达本模块前中断。

DanielLeens (changes requested):
Thanks for the follow-up and for adding the focused regression tests. I re-reviewed the latest head from scratch.

Sorry for missing this in the last review: the JDBC side here is not just extra diagnostics. Looking at the current head and the still-red changed-module CI, I should treat that part as a blocker on this version as well.

The new `TiDBSourceReaderTest#flushRowsShouldHoldCommitUntilMatchingPrewriteArrives` and `TiDBSourceSplitEnumeratorTest` do close the specific proof gap I called out earlier for the reordered commit/prewrite path and the same-reader split overwrite path. So the TiDB reader fix itself is in much better shape now.

## What this PR solves
- Pain point: the TiDB CDC source could silently lose a row when `COMMIT` arrived first, `resolvedTs` moved past it, and the matching `PREWRITE` only showed up later. The enumerator also overwrote pending splits when multiple returned splits mapped to the same reader.
- Approach: keep unmatched commits in `flushRows(...)`, lower the safe `resolvedTs`, store pending splits as lists per reader, and add focused unit coverage.
- One-line summary: the TiDB-side fix direction is right, but the current head still has one checkpoint-compatibility break and one JDBC-scope / CI blocker before merge.

## Full runtime path I rechecked
```text
TiDB source main path
  -> TiDBSourceReader.pollNext(...) [TiDBSourceReader.java:145-159]
      -> snapshotEvents(...) [162-219]
      -> captureStreamingEvents(...) [221-305]
          -> CDCClient.get()
          -> handleRow(...) [363-387]
              -> COMMIT / PREWRITE / COMMITTED buffered into commits / preWrites
          -> flushRows(resolvedTs) [389-410]
              -> if the earliest commit has no matching prewrite yet
                   keep the commit in the map
                   lower safeResolvedTs to commitTs - 1
                   stop advancing past that point
              -> otherwise poll the commit and emit the matched prewrite row

checkpoint / restore path
  -> TiDBSourceSplitEnumerator.snapshotState(...) [225-228]
      -> TiDBSourceCheckpointState.pendingSplit
      -> SeaTunnelSource.getEnumeratorStateSerializer() [SeaTunnelSource.java:120-121]
          -> DefaultSerializer.serialize(...) [DefaultSerializer.java:28-41]
  -> restore
      -> DefaultSerializer.deserialize(...)
      -> TiDBSource.restoreEnumerator(...) [TiDBSource.java:136-146]

JDBC sink path included by this PR
  -> JdbcSinkWriter / JdbcExactlyOnceSinkWriter
      -> JdbcOutputFormat.writeRecord(...) [105-118]
          -> batchSize or batchIntervalMs can now trigger flush()
```

## Main findings
1. The TiDB reader now correctly covers the core reordered-event case I asked for earlier.
2. The new enumerator state shape is on the checkpoint / restore path, not just an in-memory refactor.
3. The JDBC change is not logging-only. It changes `JdbcOutputFormat.writeRecord(...)` behavior on the normal sink path.
4. The current `Build` check is still red, and two of the failed jobs are JDBC E2E jobs in a module touched by this PR.

## Findings

### Issue 1: changing `TiDBSourceCheckpointState.pendingSplit` in place breaks restore compatibility for existing checkpoints / savepoints
- Location:
  - `seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/tidb/source/enumerator/TiDBSourceCheckpointState.java:35-38`
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/source/SeaTunnelSource.java:120-121`
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/serialization/DefaultSerializer.java:28-41`
  - `seatunnel-translation/seatunnel-translation-base/src/main/java/org/apache/seatunnel/translation/source/ParallelSource.java:103-106`
- Why this is a problem:
  - `pendingSplit` changed from `Map<Integer, TiDBSourceSplit>` to `Map<Integer, List<TiDBSourceSplit>>`.
  - This object is serialized as the enumerator checkpoint state through the default serializer path. There is no custom versioned serializer and no backward-compatibility branch on restore.
  - So this is not just a runtime structure cleanup. It changes the persisted checkpoint schema for a running unbounded CDC source.
- Risk:
  - Existing TiDB CDC jobs restored from an older checkpoint/savepoint can fail to deserialize or otherwise fail to recover cleanly.
  - That is a checkpoint-main-path compatibility break.
- Best fix:
  - Option A: add a versioned custom serializer for the enumerator state and explicitly migrate the old `Map<Integer, TiDBSourceSplit>` shape into the new list-based shape.
  - Option B: keep backward-compatible restore handling for the old state layout before normalizing it.
  - If you intentionally keep this as an incompatible upgrade, it also needs to be documented in both incompatible-changes docs with migration guidance.
- Severity: High

### Issue 2: the JDBC side is a real sink-behavior change, not just diagnostics, and the changed-module CI is still red
- Location:
  - `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcOutputFormat.java:105-113`
  - `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcOutputFormat.java:236-239`
  - current failed jobs on this head:
    - `updated-modules-integration-test-part-1` -> `connector-jdbc-e2e-ddl` -> `SqlServerSchemaChangeIT ... ConditionTimeout`
    - `updated-modules-integration-test-part-2` -> `connector-jdbc-e2e-part-1` -> `JdbcDb2UpsertIT.testDb2UpsertE2e`
- Why this matters:
  - `writeRecord(...)` now flushes not only on `batchSize`, but also on `batchIntervalMs`.
  - That changes the normal JDBC sink execution path and retry / transaction timing. It is broader than “paired diagnostics for troubleshooting”.
  - On the current head, the visible `Build` check is still red, and two failed jobs are JDBC E2E jobs in a module touched by this change. From the current evidence I cannot safely dismiss them as unrelated noise.
- Risk:
  - This can change flush frequency and transaction boundaries across JDBC sinks, including exactly-once and schema-change paths.
  - Merging it together with the TiDB fix keeps the PR scope noisy and leaves the changed-module CI unresolved.
- Best fix:
  - Option A: split the JDBC change into a separate PR and keep this PR focused on the TiDB CDC fix.
  - Option B: if the JDBC change must stay here, add focused JDBC coverage for the new `batchIntervalMs` path and get the JDBC E2E failures green (or provide a concrete, reproducible proof that they are unrelated).
- Severity: High
- Raised by others already: Partially. I mentioned the JDBC scope concern in the previous round, but I should have called out the runtime behavior change and the still-red changed-module CI more explicitly.

## Tests and CI
- The new TiDB unit tests are useful and stable:
  - `TiDBSourceReaderTest.java:42-68` is deterministic and directly covers the reordered `COMMIT -> flush -> PREWRITE -> flush` path.
  - `TiDBSourceSplitEnumeratorTest.java:45-68` is also deterministic and covers the same-reader multi-split reassignment case.
- Test stability rating for the newly added UTs: Stable
  - No `sleep`, no time-based assertions, no port allocation, no shared static state, no order-sensitive collection assertions.
- Coverage gaps still left:
  - no restore-compatibility test for the new enumerator checkpoint shape
  - no JDBC test for the new `batchIntervalMs` flush path
- Current visible `Build` failures on head `53d729c94bbff089113ebfa379be4d8957966aa1`:
  - `connector-jdbc-e2e-ddl`: `SqlServerSchemaChangeIT ... ConditionTimeout`
  - `connector-jdbc-e2e-part-1`: `JdbcDb2UpsertIT.testDb2UpsertE2e`
  - `connector-cdc-mysql-e2e`: `MysqlCDCWithSchemaChangeIT`
- MySQL CDC looks unrelated to the files changed here, but the two JDBC E2E failures land on a touched module, so I would not wave those away on this version.

## Merge conclusion

1. Blockers
- Issue 1: please make the enumerator checkpoint state backward-compatible before merge.
- Issue 2: please either split the JDBC change out, or prove it with tests and green JDBC E2E on this head.

2. Non-blocking suggestions
- No extra non-blocking items from this round. The two issues above are already enough to block merge.

Overall, the TiDB reader fix itself is much closer now, and the new regression proof helped a lot. The remaining blockers are around upgrade safety and PR scope control, not around the original reordered-event diagnosis anymore. Happy to re-review once you push an update.

zhaoysg:
Thanks for the careful re-review.

I pushed an update in `fbf56053d`.

Changes in this update:

- Added backward-compatible normalization for `TiDBSourceCheckpointState.pendingSplit` so an older checkpoint state shaped as `Map<Integer, TiDBSourceSplit>` is restored into the new `Map<Integer, List<TiDBSourceSplit>>` representation.
- Added a focused restore-compatibility regression test covering that legacy single-split pending map shape and verifying the restored enumerator can reassign it.
- Removed the JDBC diagnostics change from this PR, so the PR diff is now focused on TiDB CDC again and no longer touches `connector-jdbc` relative to the PR base.

One clarification on the JDBC point: after checking the PR diff against the base, the `batchIntervalMs`-triggered `writeRecord` flush path was already present on the base branch; this PR had only added flush diagnostics in `JdbcOutputFormat`. Still, I agree it was unnecessary scope for this TiDB fix, so I removed it from this PR.

Local verification:

```bash
./mvnw -DspotlessFiles=seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/tidb/source/enumerator/TiDBSourceCheckpointState.java,seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/tidb/source/enumerator/TiDBSourceSplitEnumeratorTest.java,seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcOutputFormat.java spotless:check
./mvnw -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb -DskipIT -Dskip.spotless=true -Dcheckstyle.skip=true -Dtest=TiDBSourceReaderTest,TiDBSourceSplitEnumeratorTest test
```

The focused TiDB tests pass locally: 4 tests, 0 failures.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the TiDB checkpoint-state restore path and the new regression coverage, and the exact PR scope now looks good to me.

What this PR fixes
- User pain: After the TiDB CDC pending-split structure changed from a single split to a list, old checkpoint state could no longer restore cleanly.
- Fix approach: The PR normalizes legacy `Map<Integer, TiDBSourceSplit>` state into the new `Map<Integer, List<TiDBSourceSplit>>` shape in the constructor, setter, and `readObject()` path.
- One-line summary: Legacy checkpoint state now restores cleanly into the new pending-split structure.

Runtime path I traced
```text
Restore path
  -> Java deserialization of `TiDBSourceCheckpointState`
      -> `readObject()` normalizes the pending-split map
      -> legacy single-split entries are wrapped into singleton lists

Result
  -> old checkpoint state stays readable after the internal shape change
```

Key findings
- This change is on the restore path, which is exactly where the compatibility bug lives.
- Constructor, setter, and `readObject()` now all normalize state the same way.
- The exact PR scope is clean again on the current head.

Blocking items

Other review notes

- Compatibility: Fully backward compatible — that is the main point of this fix.
- Side effects: The normalization cost is tiny and only happens on restore / state-setting paths.
- Error handling / logging: Unsupported state shapes still fail fast instead of being silently swallowed.
- Tests: The added regression coverage matches the legacy-single-split restore scenario that matters here.
- Docs: No docs update is needed.
- CI: The red checks I traced are in SQLServer/MySQL CDC schema-evolution jobs, not in the exact TiDB file scope of this PR. I would sync the latest `dev` and rerun CI.

Merge verdict
- From the current head I do not see a source-level blocker. The legacy checkpoint-state compatibility issue looks fixed on the exact PR files.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the TiDB CDC reader / checkpoint path again.

- The only new commit on the current head is `[Fix][Connector-V2] Format TiDB CDC checkpoint state`.
- I rechecked the executable path in `TiDBSourceReader`, `TiDBSourceSplitEnumerator`, and `TiDBSourceCheckpointState`; the latest delta is formatting-only and does not change the logic Daniel previously approved.

Runtime path I checked
```text
TiDB CDC event stream
  -> TiDBSourceReader buffers prewrite / commit pairs
      -> resolvedTs-driven flush path
          -> checkpoint state snapshot / restore in TiDBSourceCheckpointState
              -> split assignment tracking in TiDBSourceSplitEnumerator
```

Findings
- The latest head does not reopen the earlier silent-row-loss blocker.
- The new reorder regression coverage still matches the real reader state machine Daniel reviewed in the previous approval.
- The current delta is cosmetic, not behavioral.

CI
- The latest `Build` is still pending on the current head.
- From Daniel's side, the remaining gate is CI completion rather than a reopened TiDB CDC correctness issue.

1. Blocking items
- Please get the latest `Build` green.

- None from this round.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch against the TiDB CDC reader and checkpoint path again.

- The previously approved TiDB runtime logic under `TiDBSourceReader`, `TiDBSourceSplitEnumerator`, and `TiDBSourceCheckpointState` is unchanged on the current head.
- The only new delta relative to commit `1b7d5c4c0e1690b2608aeba4270ad09c5b8fd26e` is the Jira API-token doc-link refresh in `docs/en/connectors/source/Jira.md` and `docs/zh/connectors/source/Jira.md`.

## What this PR solves
- User pain: the TiDB CDC source could silently lose a row when `COMMIT` was seen first, `resolvedTs` advanced, and the matching `PREWRITE` arrived later; the pending-split checkpoint shape also needed backward-compatible restore handling.
- Fix approach: keep unmatched commits in `flushRows(...)`, lower the safe resolved-ts boundary until the matching prewrite arrives, and normalize legacy pending-split checkpoint state into the new list-based structure on restore.
- One-line summary: the TiDB-side correctness fix I previously approved still stands on the current head, and the latest delta is docs-only.

## Full runtime path I rechecked
```text
TiDB CDC event stream
  -> TiDBSourceReader.pollNext(...) [TiDBSourceReader.java:146-158]
      -> snapshotEvents(...) [162-219]
      -> captureStreamingEvents(...) [221-357]
          -> handleRow(...) [363-387]
              -> buffer PREWRITE / COMMIT into preWrites / commits
          -> flushRows(resolvedTs) [389-410]
              -> keep unmatched commit in the map
              -> lower safeResolvedTs to commitTs - 1
              -> emit the row once the matching prewrite arrives

Checkpoint / restore
  -> TiDBSourceCheckpointState.readObject(...) [TiDBSourceCheckpointState.java:54-57]
      -> normalizePendingSplit(...) [63-81]
          -> wrap legacy single-split entries into singleton lists

Regression coverage already on this head
  -> TiDBSourceReaderTest.flushRowsShouldHoldCommitUntilMatchingPrewriteArrives(...) [TiDBSourceReaderTest.java:43]
  -> TiDBSourceSplitEnumeratorTest.checkpointStateShouldRestoreLegacySingleSplitPendingMap(...) [TiDBSourceSplitEnumeratorTest.java:73]

Latest delta only
  -> docs/en/connectors/source/Jira.md
  -> docs/zh/connectors/source/Jira.md
      -> refresh the Atlassian API-token help link
```

## Key findings
- The latest head does not change the approved TiDB runtime logic or reopen the earlier silent-row-loss blocker.
- The legacy checkpoint-state normalization path is still present on the restore path, so the compatibility fix remains intact.
- The new docs change updates both English and Chinese connector pages consistently, so there is no doc-parity gap in the latest delta.

## Tests and stability
- The previously added focused TiDB regression tests still cover the real reordered `COMMIT -> PREWRITE` path and the pending-split restore case.
- The latest docs-only delta does not add a new flaky-test pattern or reopen a missing test gap.
- Test stability rating for the current TiDB coverage remains: Stable.

## CI
- The latest `Build` is still queued on the current head.
- From Daniel's side, the remaining gate is CI completion rather than a reopened TiDB CDC source issue.

1. Blocking items

- None from this round.

The TiDB fix itself still looks good to me on the latest revision.

nzw921rx (commented):
Thank you for your contribution. The core of this PR is to modify the TiDB TS time advancement issue, but it seems to carry a SQL Server that is not within the scope of this PR. Can you clarify the scope of this PR? I think a PR should solve a problem, otherwise it will blur the boundaries of the PR and make it impossible to track🚀

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from scratch against the TiDB CDC reader path, the checkpoint-restore path, and the new files added after Daniel's last pass.

+1 to @nzw921rx's earlier point about PR scope. After tracing the current head, I can confirm the latest delta is not just another TiDB follow-up: it adds a full SQL Server JDBC E2E init path in `SqlServerSchemaChangeIT` plus a companion test in `SqlServerSchemaChangeITTest`, and that path is outside the TiDB silent-row-loss fix this PR is supposed to carry.

## What this PR solves
- User pain: the TiDB CDC source could silently lose a row when `COMMIT` arrived first, `resolvedTs` had already advanced, and the matching `PREWRITE` only showed up later. The pending-split checkpoint shape also needed backward-compatible restore handling.
- Fix approach: keep unmatched commits in `flushRows(...)`, lower the safe resolved-ts boundary until the matching prewrite arrives, normalize legacy pending-split checkpoint state into the new list-based shape, and cover both with focused regression tests.
- One-line summary: the TiDB fix itself still looks correct on the current head, but the latest revision broadens this PR with an unrelated SQL Server JDBC E2E change that should be split out before merge.

```text
TiDB CDC main path
  -> TiDBSourceReader.pollNext(...) [TiDBSourceReader.java:146-159]
      -> captureStreamingEvents(...) [221-305]
          -> handleRow(...) [363-387]
              -> buffer COMMIT / PREWRITE into commits / preWrites
          -> flushRows(resolvedTs) [389-410]
              -> keep unmatched commit in the map
              -> lower safeResolvedTs to commitTs - 1
              -> emit only after the matching prewrite arrives

Checkpoint / restore
  -> TiDBSourceSplitEnumerator.snapshotState(...) [225-228]
      -> TiDBSourceCheckpointState(boolean, Map<Integer, ?>) [41-44]
      -> readObject(...) [54-60]
          -> normalizePendingSplit(...) [63-87]

Latest unrelated delta on this head
  -> SqlServerSchemaChangeIT.initSinkContainer() [88-145]
      -> executeSqlServerCommand(...) [147-173]
          -> container.execInContainer(sqlcmd, ...)
      -> executeOptionalExternalUserCommand(...) [175-192]
      -> SqlServerSchemaChangeITTest.* [25-55]
```

## Key findings
1. The TiDB runtime logic Daniel had already approved is still intact on the current head. I do not see a reopened source-level issue in `TiDBSourceReader`, `TiDBSourceCheckpointState`, or `TiDBSourceSplitEnumerator`.
2. The new TiDB regression coverage is still the right coverage for the real failure mode:
   - `TiDBSourceReaderTest.flushRowsShouldHoldCommitUntilMatchingPrewriteArrives(...)`
   - `TiDBSourceSplitEnumeratorTest.checkpointStateShouldRestoreLegacySingleSplitPendingMap(...)`
3. The new delta after Daniel's last pass is a separate JDBC E2E concern, not another TiDB fix. It changes the SQL Server schema-change test container bootstrap flow and adds a dedicated test for that helper behavior.

## Findings

### Issue 1: the latest head mixes an unrelated SQL Server JDBC E2E init change back into this TiDB CDC fix
- Location:
  - `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl/src/test/java/org/apache/seatunnel/connectors/jdbc/SqlServerSchemaChangeIT.java:88`
  - `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl/src/test/java/org/apache/seatunnel/connectors/jdbc/SqlServerSchemaChangeITTest.java:23`
- Why this is a problem:
  - This is related to @nzw921rx's earlier scope concern. Following the actual call chain shows that the latest addition does not strengthen the TiDB reader fix. Instead, it modifies the SQL Server schema-change E2E bootstrap path through `initSinkContainer() -> executeSqlServerCommand(...) / executeOptionalExternalUserCommand(...)`.
  - That means the current PR now carries two separate problem domains again: the TiDB CDC correctness fix, and a SQL Server JDBC E2E readiness adjustment.
- Risk:
  - The main risk here is review and CI attribution drift. Once an unrelated JDBC E2E change is bundled into a TiDB correctness PR, it becomes much harder to tell whether later failures or follow-up discussions belong to the TiDB fix or to the extra SQL Server test change.
- Best fix:
  - Option A: split the `SqlServerSchemaChangeIT*` change into a separate PR and keep this PR focused on the TiDB CDC fix. This is the preferred option.
  - Option B: keep it here only if you can prove both paths are part of the same problem closure, but from the current runtime paths I do not think that case is established.
- Severity: Medium
- Raised by others already: Yes, this is aligned with @nzw921rx's review, and I am adding the concrete execution-path evidence above.

## Tests and stability
- TiDB regression tests are still solid and deterministic.
- The newly added SQL Server helper assertions are also deterministic, and `SqlServerSchemaChangeIT.java:150-172` uses Awaitility polling instead of a fixed sleep.
- Test stability rating for the new test changes on the current head: Stable.

## CI
- The current `Build` check is still in progress on head `7b67174c6968a50fdaef266056f1a178c2dcfa61`.
- From Daniel's side, the source-level TiDB logic is not the blocker anymore. The blocker on this revision is PR scope. After the scope is cleaned up, I would still wait for green CI before merge.

1. Blocking items
- Issue 1: please split the unrelated `SqlServerSchemaChangeIT*` change out of this PR and keep this PR scoped to the TiDB CDC silent-row-loss / checkpoint-compatibility fix.

- No additional non-blocking items from this round. The TiDB fix itself still looks good to me on the current head.

Overall, the TiDB-side fix is in good shape now. The part I cannot sign off on is that the latest revision broadens the PR again with a separate SQL Server JDBC E2E change. Once the scope is narrowed back down and CI finishes green, I am happy to look again.

zhaoysg:
感谢 @nzw921r  @DanielLeens 指出这个问题。

SQL Server E2E 相关改动确实超出了这个 TiDB CDC PR 的范围。之前加入这两个提交，主要是因为当时 GitHub Actions / CI 一直无法变绿，我在排查 workflow 和 E2E 失败原因时做了相关调整。但这些内容不应该和本次 TiDB CDC 数据正确性修复放在同一个 PR 里。

我现在已经把 PR 分支强制更新回 c09619555，并移除了这两个 SQL Server E2E 提交：

- [Fix][E2E] Wait for SQLServer XA procedure setup
- [Fix][E2E] Tolerate missing SQLServer external user option

所以当前 PR 的范围已经重新收窄为 TiDB CDC silent row loss / checkpoint restore compatibility 的修复。如果 SQL Server E2E 问题后续仍然需要处理，我会单独开一个 PR 来解决。

DanielLeens (commented):
Thanks for the update, and thanks for narrowing the scope back down. I re-reviewed the latest head from scratch against the TiDB CDC reader path, the checkpoint-restore path, and the current test coverage on the exact head now at `c0961955579ea2379416f42332fcf514ebb34ed8`.

## What this PR solves
- User pain: the TiDB CDC source could silently lose a row when `COMMIT` arrived first, `resolvedTs` advanced past it, and the matching `PREWRITE` only arrived later. The pending-split checkpoint shape also needed backward-compatible restore handling after moving from a single split to a list per reader.
- Fix approach: keep unmatched commits in `flushRows(...)` until the matching prewrite arrives, lower the safe resolved-ts boundary while waiting, and normalize legacy checkpoint state into the new list-based structure in the constructor / setter / `readObject()` restore path.
- One-line summary: the current head is back to a clean TiDB CDC correctness + checkpoint-compatibility fix, and I do not see a new source-level blocker on this narrowed revision.

```text
TiDB CDC main path
  -> TiDBSourceReader.pollNext(...) [TiDBSourceReader.java:145-159]
      -> captureStreamingEvents(...) [221-305]
          -> handleRow(...) [363-387]
              -> buffer COMMIT / PREWRITE into commits / preWrites
          -> flushRows(resolvedTs) [389-410]
              -> if prewrite is still missing
                   keep the commit in the map
                   lower safeResolvedTs to commitTs - 1
              -> once prewrite arrives
                   emit the committed row normally

Checkpoint / restore path
  -> TiDBSourceSplitEnumerator.snapshotState(...) [225-228]
      -> TiDBSourceCheckpointState(boolean, Map<Integer, ?>) [41-44]
      -> readObject(...) [54-60]
          -> normalizePendingSplit(...) [63-81]
              -> wrap legacy single-split entries into singleton lists

Regression coverage on this head
  -> TiDBSourceReaderTest
  -> TiDBSourceSplitEnumeratorTest.checkpointStateShouldRestoreLegacySingleSplitPendingMap(...)
```

## Key findings
1. The TiDB reader fix still lands on the real hot path and still covers the reordered `COMMIT -> PREWRITE` case correctly.
2. The checkpoint compatibility fix is on the actual restore path, not just an in-memory refactor.
3. The unrelated SQL Server E2E change that blocked the previous head is no longer part of the current revision.
4. I do not see a reopened source-level issue in the latest narrowed scope.

## Tests and stability
- The focused TiDB regression tests on the current head are still the right tests for the two important behaviors here:
  - reordered commit/prewrite handling
  - legacy single-split checkpoint restore
- Stability rating: Stable.
  - deterministic state-based tests
  - no fixed sleeps
  - no port or container timing dependency in the new UT coverage

## CI
- The latest `Build` for this head is still queued.
- From Daniel's source-level review side, the remaining gate on the current head is CI completion rather than a new TiDB CDC logic blocker.

1. Blocking items
- Please let the current `Build` on head `c0961955579ea2379416f42332fcf514ebb34ed8` finish green before merge.

- None from this round.

The latest narrowed revision looks good to me from the TiDB CDC source/restore path side.

nzw921rx:
@zhaoysg please rebase upstream/dev, doc dead link fixed🚀

nzw921rx (approved):
+1
LGTM

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the TiDB CDC reader / checkpoint path again, and I also rechecked the current PR file surface rather than assuming the earlier result still holds.

## What I rechecked
```text
TiDB CDC main path
  -> TiDBSourceReader.handleRow(...)
  -> flushRows(resolvedTs)
      -> hold unmatched COMMIT until matching PREWRITE arrives

Checkpoint / restore path
  -> TiDBSourceSplitEnumerator.snapshotState(...)
  -> TiDBSourceCheckpointState normalization
      -> restore legacy single-split pending state into the list-based shape
```

On the current head, I do not see an effective reopened change in the six TiDB files that Daniel previously reviewed on the source side. The focused reader/checkpoint logic remains the same conclusion for me.

I also noticed @nzw921rx already approved the current source direction. After my own recheck, I agree that the TiDB-side logic is not the blocker on this latest revision.

- No reopened source-level blocker from my side in the TiDB CDC reader / restore path.
- The remaining gate on this head is the current red `Build` signal, not a newly found TiDB correctness regression from this Daniel pass.

1. Blocking items
- Please resolve the current `Build` failure before merge.

- None from this round.

DanielLeens:

- base branch: `dev`
- head SHA: `9664f77d34ca1c2fdc0b1e8c556bd103e5d3f293`
- compare status: `diverged`
- ahead_by: `6`
- behind_by: `4`
- mergeable_state: `unknown`

DanielLeens (approved):
Thanks for narrowing the scope back down and syncing with the latest `dev`. I re-reviewed the latest head from the TiDB CDC reader / checkpoint path again instead of carrying over the earlier rounds mechanically.

Runtime path I checked
```text
streaming reorder handling
  -> TiDBSourceReader.handleRow(...)
  -> flushRows(resolvedTs)
      -> hold COMMIT until matching PREWRITE arrives
checkpoint restore
  -> TiDBSourceCheckpointState.normalizePendingSplit(...)
enumerator assignment
  -> TiDBSourceSplitEnumerator.addSplitsBack / assignSplit(...)
```

- The scope is clean again on the current revision, so the earlier PR-scope blocker is not reopened.
- The reader-side hold-and-flush path still matches the intended fix for the out-of-order COMMIT / PREWRITE case.
- The checkpoint-state normalization also still protects restore compatibility between the old single-split shape and the new list-based pending-split shape.
- The added regression coverage is focused, and I do not see an obvious flaky pattern in the new tests.

Merge conclusion
- No reopened blocker from my side on the current head.
- From a source-level view this is ready once the normal Build / CI gates finish on top of the synced branch.

zhaoysg:
提交的PR多会可以合并呀？ 都已经很久了。

DanielLeens:
Thanks for the follow-up. On the current unchanged head, I do not see a reopened source-level blocker in the TiDB CDC reader / checkpoint path from my side.

The remaining gate is the current `Build` run on this head:
- https://github.com/apache/seatunnel/runs/84911591852

Once that check settles green, I do not have another merge blocker from this Daniel pass.

DanielLeens:
Thanks for the update. I rechecked the latest head `c82963bced` against the current gate state again.

From Daniel's side, the earlier source-level approval on this head still stands for me. The remaining gate is operational:
- base branch: `dev`
- compare status: `diverged`
- behind_by: `1`
- current failing workflow/job: `Label-when-reviewed / Build`

Because this branch is behind the latest `dev` and the remaining blocker is now only the failing `Build` gate, the lowest-cost next step is to sync with the latest `dev` and rerun `Build` first. If the refreshed head still fails after the sync, please share that exact new failing check and I can help narrow it down on top of the updated revision.

DanielLeens (commented):

# What this PR fixes
- PR: [Fix][Connector-V2] Fix TiDB CDC silent row event loss when resolvedT…

```text
TiDB CDC event handling
  -> resolvedTs / row event stream enters the CDC source reader
  -> row events are emitted instead of silently skipped
  -> checkpoint state continues to track the TiDB CDC stream position
```

# Current review findings
- Changed files in the current full diff: 6.
- Base sync fact recorded: compare status `diverged`, `behind_by=1`. This is recorded as queue metadata; I am not turning it into a standalone sync request unless it is needed to interpret a failing CI signal.

- Fetched the PR head into a local review branch named `seatunnel-review-11113`.

# Merge conclusion

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `67e57ce4fa7f` from scratch.

# What this PR fixes
- User pain: TiDB CDC pipelines silently lost row change events — when a commit row arrived before its prewrite (region split, rebalance, backpressure), `flushRows` polled the commit off the queue, dropped it, and `captureStreamingEvents` still advanced `resolvedTs` to `cdcClient.getMaxResolvedTs()`, so downstream skipped real data at that timestamp with no error (#11013).
- Fix approach: `flushRows` now returns a `safeResolvedTs` and only `pollFirstEntry()`s a commit after confirming its matching prewrite is buffered in `preWrites`, holding un-pairable commits for the next batch and capping `resolvedTs` advancement; the PR also rewrites `snapshotEvents` to region-aware scanning via `RegionStoreClient`, changes the enumerator/checkpoint state from `Map<Integer, TiDBSourceSplit>` to `Map<Integer, List<TiDBSourceSplit>>` (with a `readObject`-time `normalizePendingSplit` shim for old state), and adds `[TiDB-CDC-DIAG]` logging.
- One-line summary: The core commit/prewrite pairing fix looks correct and well-motivated, but the PR bundles three logically separate changes (event-loss fix, snapshot scan rewrite, checkpoint state shape change), and I want us to scrutinize the serialized-state compatibility path and the custom `readObject` normalization before merging — plus the new diag knobs/behavior aren't reflected in docs.

# Runtime chain I rechecked
```text
Streaming path (the fixed code):
  TiDBSourceReader.pollNext(output)  TiDBSourceReader.java:~140
    -> TiDBSourceReader.captureStreamingEvents(split, output)  TiDBSourceReader.java:~275
      -> getCdcClient(split, resolvedTs).get()  (per poll attempt, up to batchSize)
      -> flushRows(...)  TiDBSourceReader.java (returns safeResolvedTs; commits.firstEntry checked against preWrites before pollFirstEntry — previously poll-then-drop)
        -> streamingRecordDeserializer.deserialize(pairedRow, output)
      -> split.setResolvedTs(safeResolvedTs)  (was cdcClient.getMaxResolvedTs() unconditionally)

Snapshot path (rewritten in same PR):
  TiDBSourceReader.pollNext(output)
    -> TiDBSourceReader.snapshotEvents(split, output)  TiDBSourceReader.java:163
      -> session.getRegionStoreClientBuilder().build(start)  TiDBSourceReader.java:~174 (was session.createKVClient())
      -> scanClient.scan(ConcreteBackOffer.newScannerNextMaxBackOff(), start, startTs)  TiDBSourceReader.java:~180
      -> snapshotRecordDeserializer.deserialize(record, output); split.setSnapshotStart(nextStart); split.setResolvedTs(startTs)  TiDBSourceReader.java:~215

Split assignment / restore path (state shape change):
  TiDBSourceSplitEnumerator.run()  TiDBSourceSplitEnumerator.java:97
    -> addPendingSplit(sourceSplits)  TiDBSourceSplitEnumerator.java:118 (Map<Integer, List<TiDBSourceSplit>> now)
    -> assignSplit(readers) -> context.assignSplit(reader, List<TiDBSourceSplit>)  TiDBSourceSplitEnumerator.java:128
  Checkpoint restore:
    engine deserializes TiDBSourceCheckpointState -> readObject()  TiDBSourceCheckpointState.java:52
      -> normalizePendingSplit(pendingSplit)  TiDBSourceCheckpointState.java:61 (accepts legacy single-split values, wraps IllegalArgumentException as IOException)
      -> TiDBSourceSplitEnumerator ctor restores pendingSplit  TiDBSourceSplitEnumerator.java:65-70
```

# Findings
**Issue 1: Possible public API change in seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/tidb/source/reader/TiDBSourceReader.java**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/tidb/source/reader/TiDBSourceReader.java`
- Severity: High

**Issue 2: Unsafe Java deserialization pattern**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/tidb/source/enumerator/TiDBSourceCheckpointState.java:1`
- Why it matters: Matched in added line: +    private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
- Suggested fix: Security: Unsafe Java deserialization pattern. Please review this line.
- Severity: High

**Issue 3: Config/option changes without doc updates**
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/tidb/source/reader/TiDBSourceReader.java
- Issue 2: Unsafe Java deserialization pattern

- Issue 3: Config/option changes without doc updates

zhaoysg:
Thanks for the careful re-review.

I pushed `9811e7daf0dc` to address the two blocking items:

1. For the protected `TiDBSourceReader.flushRows(long)` signature, I restored the original `protected void flushRows(long)` method for compatibility and moved the new safe-resolved-ts return path into a private helper `flushRowsAndGetSafeResolvedTs(long)`. The runtime path now uses the private helper, while downstream subclasses compiled against the old protected method are not forced through a signature change.

2. For the custom Java deserialization concern, I removed `readObject(ObjectInputStream)` from `TiDBSourceCheckpointState`. Legacy/current pending split normalization is now done through the normal getter/setter/constructor path, and the list copy path now validates each element is a `TiDBSourceSplit` instead of relying on an unchecked list cast.

For the doc/config note: the PR already includes the TiDB CDC doc updates for the diagnostic option in both `docs/en` and `docs/zh`.

Verified locally:

```bash
./mvnw.cmd -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb spotless:apply
./mvnw.cmd -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-tidb -DskipIT=true -Dtest=TiDBSourceReaderTest,TiDBSourceSplitEnumeratorTest test
```

Result: `Tests run: 4, Failures: 0, Errors: 0, Skipped: 0`.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head after the new commits, and I can confirm that both blockers from the previous round are now resolved:

1. The protected `flushRows(long)` signature is restored — the new safe-resolved-ts logic lives in the private `flushRowsAndGetSafeResolvedTs` helper, so downstream subclasses compiled against the old protected method are unaffected.

2. The `readObject` custom deserialization has been removed — backward compatibility for the old `Map<Integer, TiDBSourceSplit>` checkpoint state is now handled through `normalizePendingSplit` in the constructor/setter path, which is cleaner and safer.

I also agree with the direction of @SEZ9's earlier review points — those have all been addressed in the current head.

## Runtime Path
```
TiDB CDC streaming read
  → TiDBSourceReader.pollNext(output)
    → captureStreamingEvents(split, output)
      → cdcClient.get() pulls events (up to batchSize, with maxPollAttempts cap for metadata drain)
      → handleRow(row) stores prewrite in preWrites, commit in commits
      → flushRowsAndGetSafeResolvedTs(resolvedTs)
        → for each commit with commitTs ≤ resolvedTs:
          → check preWrites for matching prewrite
          → found: poll commit, move prewrite to committedEvents
          → not found: safeResolvedTs = min(safeResolvedTs, commitTs - 1), hold commit, break
      → split.setResolvedTs(safeResolvedTs)  (was: cdcClient.getMaxResolvedTs() unconditionally)
      → emit committedEvents rows

TiDB CDC snapshot read (rewritten)
  → snapshotEvents(split, output)
    → region-aware scan via RegionStoreClient (was: KVClient.scan)
    → per-region: scan(BackOffer, start, startTs) → filter isRecordKey → deserialize → advance start key
```

## Merge Conclusion
Source-level: **no blocking issues found**. The commit/prewrite pairing fix is correct and well-tested. The snapshot scan rewrite and checkpoint state shape change (with backward-compatible `normalizePendingSplit`) look solid. The diagnostic logging is a nice addition for production troubleshooting.

Current CI is still running (Build in progress). I'll approve once the required checks are green.

DanielLeens (approved):
Thanks for the update. I checked the current head `5e6403975904` and the Build is now passing. Since my previous review found no source-level blockers and the only remaining gate was CI, I'm updating my conclusion to approved.

The source blocker was cleared in my previous round (protected `flushRows` signature restored, `readObject` removed, prewrite/commit pairing logic correct). The CI gate is now green. From my side, this is ready to merge.

Note: my earlier `changes requested` entry from a prior round may still appear in GitHub's review summary because I do not have write-level review controls on this repository — a maintainer with write access may still need to issue the final approval step on top of this technical recheck.

nzw921rx (approved):
+1 good job

DanielLeens:
Thanks @nzw921rx. I’m aligned with the current direction on this unchanged head.

From Daniel's side I do not see a reopened source-level blocker on the current revision. I’m just recording the live mergeability fact separately in the review queue, but I am not asking the author for a new code change or a sync-first step based on this follow-up alone.

DanielLeens (approved):
Rechecked the latest head: my previous source-level review had already cleared the code-side blocker, and the remaining gate at that point was CI.

The checks are now green for the current head (`5e6403975904c3b0683d827ebe932b0778f0aa83`), so I do not see a remaining Daniel-side blocker. Approving this round. Thanks for tightening the TiDB CDC row-event handling.

davidzollo (approved):
+1
LGTM

PR #11127, opened 2026-06-18
[Fix][API] Add missing OptionRule validation for CatalogFactory creation path

nzw921rx:
### Purpose

All factory creation paths in SeaTunnel validate user config against `optionRule()` before instantiation — except Catalog:

| Factory Type | Validates `optionRule()` | Entry Point |
|---|---|---|
| Source | ✅ | `FactoryUtil.createSource()` |
| Sink | ✅ | `FactoryUtil.createSink()` |
| Transform | ✅ | `FactoryUtil.createTransform()` |
| **Catalog** | ❌ | `FactoryUtil.createOptionalCatalog()` |

This is a framework-level bug: **all 37 CatalogFactory implementations** silently skip config validation. When users provide invalid configs (e.g. missing required `username`/`password` for JDBC catalogs, or missing `hosts` for Elasticsearch), the error is not caught upfront — it surfaces later as a cryptic `NullPointerException` or connection failure deep in the catalog implementation.

This PR fixes the gap by adding `ConfigValidator.validate(optionRule())` to `createOptionalCatalog()`.

### Changes

- Add `ConfigValidator.of(readonlyConfig).validate(catalogFactory.optionRule())` in `FactoryUtil.createOptionalCatalog()` before calling `createCatalog()`.
- Fix `LanceCatalogFactory.optionRule()` returning `null` → return empty `OptionRule.builder().build()` to prevent NPE.
- Add 4 unit tests in `CatalogTableUtilTest` covering valid config, missing/partial required options, and unknown factory identifier.

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-api -Dtest=CatalogTableUtilTest test`

### DIFF
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/table/factory/FactoryUtil.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/table/factory/FactoryUtil.java
@@ -286,13 +286,17 @@ public final class FactoryUtil {
 
     public static Optional<Catalog> createOptionalCatalog(
             String catalogName,
-            ReadonlyConfig options,
+            ReadonlyConfig readonlyConfig,
             ClassLoader classLoader,
             String factoryIdentifier) {
         Optional<CatalogFactory> optionalFactory =
                 discoverOptionalFactory(classLoader, CatalogFactory.class, factoryIdentifier);
+
         return optionalFactory.map(
-                catalogFactory -> catalogFactory.createCatalog(catalogName, options));
+                catalogFactory -> {
+                    ConfigValidator.of(readonlyConfig).validate(catalogFactory.optionRule());
+                    return catalogFactory.createCatalog(catalogName, readonlyConfig);
+                });
     }
 
     public static <T extends Factory> URL getFactoryUrl(T factory) {
--- a/seatunnel-api/src/test/java/org/apache/seatunnel/api/table/catalog/CatalogTableUtilTest.java
+++ b/seatunnel-api/src/test/java/org/apache/seatunnel/api/table/catalog/CatalogTableUtilTest.java
@@ -22,7 +22,9 @@ import org.apache.seatunnel.shade.com.typesafe.config.ConfigFactory;
 import org.apache.seatunnel.shade.com.typesafe.config.ConfigValueFactory;
 
 import org.apache.seatunnel.api.configuration.ReadonlyConfig;
+import org.apache.seatunnel.api.configuration.util.OptionValidationException;
 import org.apache.seatunnel.api.options.ConnectorCommonOptions;
+import org.apache.seatunnel.api.table.factory.FactoryUtil;
 import org.apache.seatunnel.api.table.type.ArrayType;
 import org.apache.seatunnel.api.table.type.BasicType;
 import org.apache.seatunnel.api.table.type.DecimalType;
@@ -43,7 +45,10 @@ import java.net.URL;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 
 import static org.apache.seatunnel.common.constants.CollectionConstants.PLUGIN_NAME;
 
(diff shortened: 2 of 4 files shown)


DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff from the actual catalog creation path instead of treating this as a test-only cleanup.

What this PR fixes
- User pain: `CatalogFactory.optionRule()` was declared by some catalogs, but `FactoryUtil.createOptionalCatalog(...)` did not validate it, so missing required catalog options could fail later and less consistently than the existing source/sink factory paths.
- Fix approach: validate the catalog factory `OptionRule` at the `createOptionalCatalog(...)` entry point, add regression coverage for valid / missing / partial / unknown-factory cases, and make `LanceCatalogFactory.optionRule()` return an empty rule instead of `null`.
- One-line summary: this aligns catalog creation with the existing factory validation contract, and I did not find a source-level blocker in the latest head.

Runtime path I checked
```text
CatalogTableUtil.getCatalogTables(...)
  -> no explicit schema branch
  -> FactoryUtil.createOptionalCatalog(factoryId, readonlyConfig, classLoader, factoryId)
      -> discoverOptionalFactory(...)
      -> ConfigValidator.of(readonlyConfig).validate(catalogFactory.optionRule())
      -> catalogFactory.createCatalog(...)
  -> catalog.open()
  -> catalog.getTables(readonlyConfig)
```

Key findings
1. The normal path really does hit this change. `CatalogTableUtil.getCatalogTables(...)` delegates into `FactoryUtil.createOptionalCatalog(...)` whenever the schema is not explicitly provided.
2. This is a precise fix, not a new validation model. It brings `CatalogFactory` into the same `ConfigValidator + OptionRule` contract that source and sink factories already use.
3. I rechecked `ConfigValidator`: this call uses `validate(rule)`, not `validateUnknownKeys(...)`, so it enforces required/value constraints without suddenly rejecting the broader `ReadonlyConfig` object for unrelated outer keys.
4. Updating `LanceCatalogFactory.optionRule()` to `OptionRule.builder().build()` is the necessary companion change so the new validation path does not trip over a null rule.

Other reviewer / maintainer input
- There were no prior non-Daniel reviews or comment threads on this PR when I reviewed it, so there was nothing to de-duplicate here.

Testing / stability
- The added tests in `CatalogTableUtilTest` cover the key entry cases: valid config, missing required options, partially missing required options, and unknown factory.
- The new tests are structurally stable: pure unit tests, no timing, ports, threads, or external services.
- GitHub `Build` was still queued when I reviewed.

1. Blocking items
- None from my side at the source level.

2. Suggested follow-up
- No additional source changes needed. Please just let the normal `Build` check finish green before merge.

Overall, this is a small and clean API-layer fix. The runtime path is real, the scope is tight, and the tests are aimed at the exact contract this PR changes.

DanielLeens (commented):
Thanks for fixing this gap. I reviewed the current diff from `CatalogTableUtil` / `FactoryUtil.createOptionalCatalog()` down into the actual `CatalogFactory` implementations that are affected, rather than only looking at the new tests.

What this PR solves:
- User pain: catalog configs were the odd one out in SeaTunnel. Source, sink, and transform factories validate `optionRule()` before creation, but catalog factories did not, so missing required options surfaced later as deeper runtime errors instead of a clear upfront validation failure.
- Fix approach: validate the catalog factory option rule before `createCatalog()`, and patch the factories that would otherwise break that new path (`LanceCatalogFactory` returning `null`, plus DuckDB using the wrong JDBC catalog rule for its real config contract).
- In plain words: bad catalog config now fails early and clearly, instead of blowing up later in a less understandable place.

Call path I checked:
```text
CatalogTableUtil / catalog creation helpers
  -> FactoryUtil.createOptionalCatalog(...)
      -> discoverOptionalFactory(..., CatalogFactory.class, factoryIdentifier)
      -> ConfigValidator.validate(catalogFactory.optionRule())
      -> catalogFactory.createCatalog(...)
```

I did not find a blocking issue in the current revision. The behavior change is user-visible, but it is the right direction and it is still backward-compatible for valid configs: only previously-invalid catalog configs now fail earlier. The Lance and DuckDB follow-up fixes are also necessary for this new validation path to be safe. From Daniel's side this can move forward.

nzw921rx:
@dybyte can you help me review this PR,thank you very much 🚀

DanielLeens:
Thanks for the ping. I rechecked the current state and there is still no new code diff after my last review on this PR. From my side the current head still looks good, so I am not starting a new full review round here unless a new commit lands.

chl-wxp (approved):

zhangshenghang:
Thanks for catching this — the catalog creation path genuinely was the one factory path missing `optionRule()` validation, and closing that gap is the right call.

One concern before this goes in: this turns on validation for all catalogs at once, and at least the JDBC ones look like they'll reject configs that work today.

15 JDBC catalog factories (MySQL, PostgreSQL, Oracle, SqlServer, OceanBase, Redshift, TiDB, Dameng, Xugu, Kingbase, SapHana, HighGo, Iris, OpenGauss) use `JdbcCommonOptions.BASE_CATALOG_RULE`, which declares both `username` and `password` as **required**:

```java
BASE_CATALOG_RULE = OptionRule.builder()
        .required(URL)
        .required(USERNAME, PASSWORD)
        .optional(...);
```

But JDBC connections without a username/password are legitimately supported — the whole connector treats them as optional (`Optional<String>`, used only `ifPresent`). And the config handed to validation is built on the fly in `JdbcCatalogUtils.extractCatalogConfig()`, which omits the password entirely when it isn't set:

```java
config.getPassword().ifPresent(val -> catalogConfig.put(PASSWORD.key(), val));
```

So a JDBC sink configured with `database` + `table` but no password (a passwordless local DB, Kerberos, etc.) hits `JdbcSink.getCatalog()` → `findCatalog()` → `createOptionalCatalog()` and now throws `OptionValidationException` at job startup, where it ran fine before. The source path (`JdbcCatalogUtils.getTables()`) has the same trigger.

I verified the validation behavior locally with a rule mirroring `BASE_CATALOG_RULE` and a `url`+`username`-only config — `ConfigValidator.validate()` throws as expected.

I think the DuckDB change in this PR is actually a symptom of the same thing: DuckDB has no credentials, so you had to move it off `BASE_CATALOG_RULE` to avoid exactly this failure. The other 15 still require username+password.

Suggestions:

- Relax `BASE_CATALOG_RULE` to `.optional(USERNAME, PASSWORD)` (the catalog uses the same optional-credential logic as the connection itself), or skip/relax validation in the `findCatalog` path when credentials are absent.
- Add a test that covers JDBC catalog creation **without** a password — the current tests only exercise the InMemory factory's happy/missing-required paths, so this regression isn't caught.
- Since this tightens validation behavior, it's worth a note in `docs/.../incompatible-changes.md` and a mention in the PR description (the DuckDB `optionRule` change isn't listed there either).

More broadly: once validation is on, every catalog's `optionRule()` needs to actually match the config that reaches it. Only DuckDB and Lance were touched here — it'd be good to confirm the non-JDBC factories (ES, Kudu, Paimon, …) don't have the same "declared required but optional in practice" mismatch.

nzw921rx:
Thank you for your suggestion. I will verify all factories and try my best
to ensure compatibility. If there is any incompatibility, I will write a
description of the incompatible changes

Jast ***@***.***> 于2026年6月22日周一 18:48写道：

nzw921rx:
@zhangshenghang @davidzollo I have reviewed all implementations of CatatLogFactory and ensured compatibility as much as possible. I suggest discussing related issues under the new PR: #11165

PR #11129, opened 2026-06-18
[Fix][Connector-V2] Support MongoDB CDC SRV URI

### LINKED ISSUE #10171: [bug][connector-cdc-mongodb] Fails to Connect to MongoDB Atlas Due to Hardcoded URI Scheme "mongodb://"
### Issue Description

When attempting to use the **SeaTunnel MongoDB CDC Connector** as a source for data extraction from a **MongoDB Atlas** cluster, the connection fails because the connector does not support the recommended DNS Seed List URI format: `mongodb+srv://`.

Upon investigating the connector's source code, I found that the connection URI is being hardcoded with the scheme `mongodb://`.

https://github.com/apache/seatunnel/blob/349841e8d2bde06dd42b56b07704765dad230e9c/seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mongodb/utils/MongodbUtils.java#L385

### Expected Behavior

The MongoDB CDC Connector should fully support the standard connection URI formats provided by MongoDB drivers, including the DNS Seed List format, for the following reasons:

### Steps to Reproduce

1.  Obtain the **Standard Connection String** for a MongoDB Atlas cluster (it will be in the format `mongodb+srv://...`).
2.  Configure a **SeaTunnel MongoDB CDC Connector** source using this `mongodb+srv://` URI.
3.  Attempt to run the SeaTunnel job.
4.  **Result:** The job fails immediately due to an illegal URI scheme or parsing error because the code expects the URI to start with `mongodb://`.

### Suggested Fix

The connector's logic for constructing the connection URI should be modified to:

1.  **Accept the full user-provided URI** without overriding or prepending the scheme (`mongodb://`).
2.  Ensure the underlying MongoDB Java driver can correctly parse and handle both **`mongodb://`** and **`mongodb+srv://`** schemes.

zhangshenghang:
### Purpose

Closes #10171.

The MongoDB CDC connector always prefixed the configured `hosts` value with `mongodb://`. This made standard MongoDB Atlas DNS seed list URIs such as `mongodb+srv://...` unusable because the scheme was overwritten before the MongoDB driver parsed the connection string.

### Changes

- Keep backward compatibility for the existing `host:port` style `hosts` configuration.
- Accept full MongoDB connection URIs using either `mongodb://` or `mongodb+srv://`.
- Preserve optional credentials and connection options when they are provided separately with a full URI.
- Normalize CDC partition and heartbeat namespace strings so complete URIs are not prefixed twice and credentials are not stored there.
- Update MongoDB CDC source docs in English and Chinese.
- Add unit tests for legacy host strings and SRV URI handling.

### Tests

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-mongodb -DskipIT test`


DanielLeens (commented):
Thanks for fixing this. I reviewed the current diff from `MongodbSourceConfig` / `MongodbUtils.buildConnectionString()` through the partition-map helpers that Debezium uses, instead of only checking the connection-string builder in isolation.

What this PR solves:
- User pain: today a standard Atlas-style `mongodb+srv://...` value in `hosts` gets rewritten into `mongodb://mongodb+srv://...`, so the connector cannot even build a valid MongoDB connection string.
- Fix approach: keep legacy `host:port` behavior, but detect full MongoDB URIs, preserve their scheme, and normalize the namespace keys used for CDC partition / heartbeat bookkeeping so they are not double-prefixed and do not retain credentials.
- In plain words: users can finally pass a real MongoDB URI without the connector mangling it first.

Runtime path I checked:
```text
Job startup
  -> MongodbSourceConfig(...)
      -> MongodbUtils.buildConnectionString(username, password, hosts, connectionOptions)
          -> Mongo client creation uses the resulting ConnectionString

CDC split / stream fetch
  -> MongodbRecordUtils.createPartitionMap(...)
  -> MongodbRecordUtils.createHeartbeatPartitionMap(...)
      -> namespace key is used for offset / heartbeat tracking
```

I did not find a blocking issue in the current implementation. The important compatibility point also looks good: plain `host:port` input still keeps the old `mongodb://...` path, while full URIs now stay intact. The added tests cover the main SRV cases and the partition-key normalization path. From Daniel's side this can move forward.

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

PR #11130, opened 2026-06-18
[Docs] Add zh DynamicCompile Scala example

zhangshenghang:
## What changed

- Added a Chinese DynamicCompile note explaining that SCALA support uses the Scala REPL for dynamic compilation.
- Added a minimal Chinese Scala source-code example that creates the `compile_language` output field.

## Why

The Chinese DynamicCompile documentation listed `SCALA` as a supported `compile_language`, but it did not include the Scala-specific note or a Scala example. Readers using the Chinese page could see that Scala was supported but had no local example to follow.

## Validation

- `./mvnw spotless:apply`
- `git diff --check`


DanielLeens (commented):
Thanks for adding this. I reviewed the current diff as a docs-only change against the DynamicCompile runtime contract instead of treating it as text-only polish.

What this helps with:
- User pain: the Chinese DynamicCompile page said Scala was supported, but it did not show the Scala-specific caveat or a working inline example.
- Fix approach: add the Scala REPL note and a minimal Scala source-code example that matches the documented `compile_language = "SCALA"` path.
- In plain words: Chinese readers now have a concrete Scala sample instead of having to guess from the Java example.

Runtime/doc path I checked:
```text
User reads docs/zh/transforms/dynamic-compile.md
  -> configures DynamicCompile with compile_language = SCALA
  -> inline code must follow Scala syntax for the Scala REPL path
```

I did not find a blocking issue in the current change. The example is aligned with the documented option name and with the intended Scala-specific usage note. From Daniel's side this can move forward.

davidzollo (approved):
+1
LGTM

PR #11132, opened 2026-06-19
[Fix][Connector-V2] Fix Paimon stream source state restore

QuakeWang:

The Paimon streaming source previously checkpointed a single `nextSnapshotId` for the whole enumerator. In multi-table jobs, each table owns an independent `StreamTableScan`, so the last scanned table could overwrite the snapshot state used to restore all tables.

- Store Paimon streaming snapshot progress by table id.
- Restore each `StreamTableScan` with its own checkpointed next snapshot id.
- Keep backward compatibility with legacy single-snapshot checkpoint state.
- Add unit tests for multi-table restore, null table snapshot state, and legacy restore.

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-paimon -DskipIT -DskipITCase -Dcheckstyle.skip -Dlicense.skip -Dspotless.check.skip -Dtest=PaimonStreamSourceSplitEnumeratorTest test`

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/source/PaimonSource.java
+++ b/seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/source/PaimonSource.java
@@ -176,6 +176,7 @@ public class PaimonSource
                 enumeratorContext,
                 checkpointState.getAssignedSplits(),
                 checkpointState.getCurrentSnapshotId(),
+                checkpointState.getCurrentSnapshotIds(),
                 readBuilders,
                 1);
     }
--- a/seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/source/PaimonSourceState.java
+++ b/seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/source/PaimonSourceState.java
@@ -17,17 +17,15 @@
 
 package org.apache.seatunnel.connectors.seatunnel.paimon.source;
 
-import lombok.AllArgsConstructor;
-import lombok.Getter;
-
 import javax.annotation.Nullable;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.Deque;
+import java.util.HashMap;
+import java.util.Map;
 
 /** Paimon connector source state, saves the splits has assigned to readers. */
-@Getter
-@AllArgsConstructor
 public class PaimonSourceState implements Serializable {
 
     private static final long serialVersionUID = 1L;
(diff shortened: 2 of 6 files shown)


DanielLeens (approved):
Thanks for the contribution! I reviewed the full diff from `PaimonSource.restoreEnumerator(...)` down to the actual `StreamTableScan.checkpoint()/restore(...)` path.

What this PR fixes:
- User pain: on the multi-table streaming path, the old enumerator state only kept one `currentSnapshotId`, so after a checkpoint restore every table scan could be restored from the same snapshot even though each table had its own checkpoint position.
- Fix approach: the PR extends the enumerator state to keep a per-table snapshot map, restores each table scan from its own snapshot id, and keeps the legacy single-snapshot fallback for older checkpoints.
- One-line summary: this closes the real multi-table restore bug without breaking backward compatibility for existing checkpoint state.

Runtime chain I checked:
```text
streaming checkpoint
  -> AbstractSplitEnumerator.scanNextSnapshot()
      -> each StreamTableScan checkpoints independently
  -> processDiscoveredSplits(...)
      -> nextSnapshotIds[tableId] = nextSnapshotId
  -> snapshotState(...)
      -> PaimonSourceState(pendingSplits, currentSnapshotIds)

restore
  -> PaimonSource.restoreEnumerator(...)                      [PaimonSource.java:163-180]
      -> new PaimonStreamSourceSplitEnumerator(..., currentSnapshotIds, ...)
  -> AbstractSplitEnumerator ctor                            [AbstractSplitEnumerator.java:119-133]
      -> tableSnapshotId = nextSnapshotIds.get(tableId)
      -> StreamTableScan.restore(tableSnapshotId)

legacy checkpoint compatibility
  -> PaimonSourceState(assignedSplits, currentSnapshotId)    [PaimonSourceState.java:39-44]
      -> currentSnapshotIds = emptyMap
  -> AbstractSplitEnumerator ctor
      -> if nextSnapshotIds is empty, fall back to nextSnapshotId
```

Key findings:
- The bug is real on the normal multi-table streaming restore path. Before this change, the state model only preserved one snapshot id even though `scanNextSnapshot()` already collected one per table.
- The fix is effective because the restore path now actually consumes `checkpointState.getCurrentSnapshotIds()` in `PaimonSource.restoreEnumerator(...)`.
- Backward compatibility looks intact: the legacy constructor still preserves the old single-snapshot shape, and the constructor fallback in `AbstractSplitEnumerator` keeps older checkpoints readable.
- The added unit tests cover the important cases: different snapshots per table, one table returning `null`, and the legacy single-snapshot restore path.

Issue 1: the new state field and legacy fallback path still need a short compatibility note
- Location:
  - `seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/source/PaimonSourceState.java:37-69`
  - `seatunnel-connectors-v2/connector-paimon/src/main/java/org/apache/seatunnel/connectors/seatunnel/paimon/source/enumerator/AbstractSplitEnumerator.java:126-131`
- Why I am calling this out:
  - the implementation is correct, but the new `currentSnapshotIds` field and the empty-map fallback now carry the whole backward-compatibility contract for old checkpoints
  - right now that intent is only visible if someone re-traces the full call chain
- Severity: Medium, non-blocking
- Suggestion:
  - please add a short Javadoc around the new per-table state and the legacy fallback to make it explicit that `currentSnapshotId` is the compatibility path while `currentSnapshotIds` is the real multi-table restore state

Compatibility impact:
- Fully compatible from my review.
- No API/config/default/protocol change.
- Older checkpoint state should still restore through the empty-map fallback.

Testing note:
- I did source-level review only in this pass.
- I did not run the PR locally.
- The new unit tests look stable and directly cover the restore semantics they are supposed to protect.

### Conclusion: can merge

1. Blocking items
- None from the runtime path I reviewed.

2. Suggested follow-up
- Issue 1: add a short compatibility comment/Javadoc around the new state shape and fallback branch.

Overall, this looks like the right minimal fix for the actual bug. I do not see a better high-level direction than storing the snapshot state per table, and I do not see a new blocker on the current head.

davidzollo (approved):
+1
Good job

PR #11133, opened 2026-06-19
[Fix][Connector-V2] Close all JDBC catalog connections

### LINKED ISSUE #10530: [Bug][Connector-V2][JDBC] AbstractJdbcCatalog.close() leaks remaining connections when first close fails
### Search before asking

### What happened

`AbstractJdbcCatalog.close()` iterates over `connectionMap` to close all JDBC connections, but throws a `CatalogException` immediately on the **first** failure. This causes all subsequent connections to be skipped and left open, leaking those database connections.

**Affected code (`AbstractJdbcCatalog.java`):**
```java
@Override
public void close() throws CatalogException {
    for (Map.Entry<String, Connection> entry : connectionMap.entrySet()) {
        try {
            entry.getValue().close();
        } catch (SQLException e) {
            throw new CatalogException(...);  // ← exits loop immediately, remaining connections leak
        }
    }
    connectionMap.clear();
}
```

If the catalog holds connections to multiple URLs (e.g. default + multiple databases) and the first one fails to close, all others are leaked.

**Fix:** Collect exceptions from all close attempts, then throw after all connections have been attempted, using `addSuppressed()` to preserve full error context.

### SeaTunnel Version

dev (latest)

### SeaTunnel Config

_No response_

zhangshenghang:
## Purpose

Fixes #10530.

`AbstractJdbcCatalog.close()` used to stop immediately when closing one cached JDBC connection failed. If the catalog held multiple database connections, every connection after the first failure was skipped and could be leaked.

## Changes

- Continue closing all cached JDBC connections even if one close operation throws `SQLException`.
- Preserve later close failures with `addSuppressed()` so error context is not lost.
- Clear the connection cache after all close attempts.
- Add a regression test that verifies all cached connections are closed and suppressed close failures are retained.

## Validation

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-jdbc -am -DskipIT -Dtest=AbstractJdbcCatalogTest -DfailIfNoTests=false verify`


DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff from the real JDBC catalog lifecycle, not just the new unit test.

## What this PR fixes
- Pain point: `AbstractJdbcCatalog.close()` used to stop at the first `Connection.close()` failure. That meant the remaining cached JDBC connections were never closed and the `connectionMap` cleanup path was skipped.
- Approach: keep closing the remaining connections, aggregate close failures through one `CatalogException` plus suppressed exceptions, clear `connectionMap`, and then throw if any close failed.
- Simple example: if closing `jdbc:test:first` fails, the catalog now still closes `jdbc:test:second` instead of bailing out immediately after the first exception.

```text
JDBC catalog lifecycle
  -> open() [AbstractJdbcCatalog.java:170-174]
      -> getConnection(defaultUrl)

runtime
  -> getConnection(url) [AbstractJdbcCatalog.java:123-154]
      -> cache multiple JDBC connections in connectionMap

shutdown
  -> close() [AbstractJdbcCatalog.java:177-197]
      -> iterate all cached connections
      -> close each one
      -> collect close failures as suppressed exceptions
      -> connectionMap.clear()
      -> throw aggregated CatalogException if needed
```

## Findings
- The normal catalog shutdown path definitely hits this change; it is the shared base implementation for JDBC catalogs.
- The old bug is real: one failing close prevented later connections from being released.
- The current fix preserves both goals we want here: best-effort cleanup of every cached connection and enough exception detail for debugging.
- I did not find a new correctness, compatibility, or resource-management blocker on the current head.

- Runtime cost is negligible because this only runs during shutdown.
- The side effect is positive: fewer leaked connections and a cleaner final catalog state after partial close failures.

## Test coverage / stability
- `AbstractJdbcCatalogTest.java:32-52` covers the key failure case directly: two connections both throw on close, both must still be invoked, and the final exception must keep the suppressed detail.
- This is a pure mock-based unit test with no timing or environment sensitivity. I would rate it as stable.

## Merge conclusion
### Conclusion: can merge

1. Blockers
- None on the current head.

2. Non-blocking suggestions
- None.

Overall, this is a focused fix for a real cleanup-path bug, and the current version looks good to merge from my side.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the JDBC catalog lifecycle path, specifically `open()` / cached-connection usage / `close()`.

## What this PR fixes
- Pain point: `AbstractJdbcCatalog.close()` stopped at the first `SQLException`, which could leave the rest of the cached JDBC connections open.
- Approach: try to close every cached connection, keep the first failure, add later failures as suppressed exceptions, then clear the cache and rethrow once the loop is complete.
- One-line summary: this turns catalog shutdown into “close everything, then report the failures” instead of “fail fast and leak the rest.”

## Runtime path I checked
```text
AbstractJdbcCatalog
  -> connectionMap caches JDBC connections per URL
  -> close() iterates every cached connection
      -> close all of them
      -> aggregate failures
      -> clear the cache
      -> throw once at the end if anything failed
```

## Findings
- I do not see a new correctness blocker on the current head.
- The new shutdown behavior is better aligned with resource-release expectations.
- The added test covers the key contract: later connections are still closed even when earlier ones fail.

## CI note
- The visible `Build` failures on the latest head are broad suites (`all-connectors-it-5`, `jdbc-connectors-it-part-7`, `jdbc-connectors-it-ddl`, `connector-file-sftp-it`, plus one Windows unit-test lane).
- From this small catalog-close diff, I do not see a convincing source-level link to those failures, so I do not treat them as a new code blocker for this specific change.

## Merge conclusion
### Conclusion: can merge

1. Blocking items

- None from the path I checked.

Overall, this is a clean lifecycle fix and looks good to me from the actual catalog-close path.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the current head from the JDBC catalog lifecycle path again, not only the new diff.

The good news first: the actual JDBC fix still looks fine to me. The `AbstractJdbcCatalog.close()` change still closes all cached connections and preserves the first failure, which is the right direction for the real leak path.

However, I can no longer treat the current PR as one focused fix.

# What this PR solves
- User pain: `AbstractJdbcCatalog.close()` previously stopped on the first `SQLException`, which could leave later cached JDBC connections open.
- Fix approach: keep closing every cached connection, then surface failure after the loop instead of aborting early.
- One-line summary: the JDBC close fix itself still looks good, but the current PR head now carries an unrelated docs/navigation change set as well.

# Runtime chain I checked
```text
JDBC catalog lifecycle
  -> catalog open / cached connection usage
  -> AbstractJdbcCatalog.close()
      -> iterate cached connections
      -> close every entry
      -> throw after the loop if one close failed
```

# Findings
Issue 1: the current PR head mixes the JDBC connection-close fix with unrelated connector-doc navigation changes
- Location: `docs/en/connectors/sink-overview.md:17-21`, `docs/en/connectors/source-overview.md:17-21`, `docs/zh/connectors/connector-faq.md:1-64`, `docs/zh/connectors/sink-overview.md:17-21`, `docs/zh/connectors/source-overview.md:17-21`
- Why this is a problem:
  Those docs changes are not part of the JDBC catalog connection lifecycle fix. They look like a separate documentation/navigation cleanup (and they overlap with the same dead-link/FAQ scope being reviewed independently elsewhere).
- Risk:
  This breaks the one-problem-per-PR boundary, makes the merge decision harder to reason about, and increases the chance of review / conflict churn for two unrelated topics.
- Better fix:
  Keep this PR scoped to the JDBC catalog close leak only, and move the docs navigation/FAQ changes into a separate docs PR.
- Severity: High

# Compatibility
- JDBC runtime behavior: the close-path fix remains compatible from my side.
- The blocker here is scope hygiene, not a reopened JDBC runtime compatibility issue.

# Performance and side effects
- On the JDBC runtime path, I do not see a new source-level blocker in the close logic itself.
- The problem is review scope contamination on the current head.

# Tests and docs
- The JDBC close fix already had adequate focused test coverage for the runtime change.
- The unrelated docs updates should be reviewed and merged on their own merits, not piggy-backed onto this runtime fix.

# CI
- The current head also has a noisy/pending `Build` picture, including a red `Build website` lane while the umbrella `Build` is still queued.
- Even without that CI noise, Issue 1 would still block merge because the PR scope is no longer clean.

# Merge conclusion

1. Blocking items
- Issue 1: split the unrelated docs navigation/FAQ updates out of this JDBC leak-fix PR.

2. Suggested follow-up
- Once the branch is narrowed back to the JDBC close fix, I do not currently see a reopened source-level blocker in the runtime path.

Happy to re-review once the branch is trimmed back to the JDBC-only scope.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the actual JDBC catalog lifecycle again.

# What this PR solves
- User pain: if one cached JDBC connection failed during `close()`, the old code stopped immediately and leaked the remaining cached connections.
- Fix approach: keep closing every cached connection, aggregate failures, clear the cache, and throw once at the end.
- One-line summary: the close-path leak is real, and the current head fixes it cleanly.

# Runtime path I checked
```text
Catalog startup
  -> AbstractJdbcCatalog.open()
      -> getConnection(defaultUrl)

Runtime
  -> getConnection(url)
      -> cache connections in connectionMap

Shutdown
  -> AbstractJdbcCatalog.close()
      -> iterate all cached connections
      -> close each one
      -> collect failures as suppressed exceptions
      -> clear connectionMap
      -> throw once at the end if needed
```

# Findings
- The normal shutdown path definitely hits this change, because shared JDBC catalogs all use `AbstractJdbcCatalog.close()`.
- The original bug is real: fail-fast close leaves later cached connections open.
- On the current head, the fix preserves both important behaviors: best-effort cleanup of every cached connection and enough exception detail for debugging.
- I also checked the scope concern from the previous round: the unrelated docs changes are no longer part of the current head, so the PR is focused again.

- No API, config, protocol, or serialization change.
- Runtime cost is negligible because this only runs during shutdown.
- The side effect is positive: fewer leaked connections and a cleaner final catalog state after partial failures.

# Tests / stability / CI
- The added unit test covers the key path directly: both connections still receive `close()` even when both throw.
- The test is stable: pure mock-based logic, with no timing, port, or environment dependency.
- I do not see a source-level link between this small close-path fix and the broad Build noise currently shown on the PR.

# Merge conclusion
### Conclusion: can merge

1. Blocking items
- None on the current head.

- None.

Overall this is a focused lifecycle fix for a real resource-release bug, and the current version looks good to merge from my side.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the actual JDBC catalog lifecycle again, especially the cached-connection close path.

What this PR fixes
- User pain: if one cached JDBC connection threw during `close()`, the old implementation stopped immediately and leaked the remaining cached connections.
- Fix approach: the new `close()` path keeps closing every cached connection, records the first failure, suppresses later failures onto it, then clears the cache before rethrowing.
- One-line summary: this turns a partial close into a best-effort full cleanup without hiding the original error.

Runtime path I checked
```text
AbstractJdbcCatalog.open()
  -> getConnection(defaultUrl)
  -> cache connection in connectionMap

Catalog operations
  -> reuse or add more JDBC connections in connectionMap

AbstractJdbcCatalog.close()
  -> iterate every cached connection
  -> attempt close on all of them
  -> collect suppressed failures
  -> clear cache
  -> rethrow if any close failed
```

Findings
- This is on the real cleanup path for every JDBC catalog implementation that inherits `AbstractJdbcCatalog`, so the effect is not test-only.
- The new unit test covers the important regression: both mocked connections are still closed even when the first one throws, and the exception preserves the later failure as suppressed context.

Compatibility
- Fully compatible. No option/API/default-value changes. The only behavior change is better cleanup under failure, which is the safer contract for callers.

Merge conclusion
### Conclusion: can merge

1. Blocking items

- None required for merge.

This still looks like a focused and correct resource-release fix, and I am fine with merging it.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head `2597841b54` from the actual JDBC catalog lifecycle again, especially the cached-connection close path after the recent `upstream/dev` merge.

## What this PR fixes
- Pain point: if one cached JDBC connection threw during `close()`, the old implementation stopped immediately and leaked the remaining cached connections.
- Approach: keep closing every cached connection, preserve the first failure, attach later failures as suppressed exceptions, clear the cache, and only throw once at the end.
- One-line summary: this turns a partial close into a best-effort full cleanup without hiding the original error.

## Runtime path I checked
```text
Catalog startup
  -> AbstractJdbcCatalog.open()
      -> getConnection(defaultUrl)

Runtime
  -> getConnection(url)
      -> cache connections in connectionMap

Shutdown
  -> AbstractJdbcCatalog.close()
      -> iterate every cached connection
      -> attempt close on all of them
      -> aggregate failures
      -> clear connectionMap
      -> throw once at the end if needed
```

## Findings
- The normal shutdown path definitely hits this change because shared JDBC catalogs all use `AbstractJdbcCatalog.close()`.
- The original bug is real: fail-fast close can leak later cached connections.
- The current latest head still keeps the fix focused to the real lifecycle path. The recent branch updates did not reopen a source-side issue for me.
- The added unit test covers the right regression and is structurally stable.

## CI note
- I rechecked the latest visible failures and the red jobs are broad integration lanes such as `all-connectors-it-7 (8, ubuntu-latest)` and `rocketmq-connector-it (8, ubuntu-latest)`.
- From this two-file JDBC catalog close fix, I do not see a convincing source-level link to those suites.
- The branch has already absorbed the recent workflow fixes from `upstream/dev`; it is only one unrelated collaborator update behind the current `dev` head, so I do not have a smaller source-side sync to suggest here.

## Merge conclusion
### Conclusion: can merge

1. Blocking items
- None on the current head.

- None from the JDBC catalog lifecycle path I checked.

Overall, this is still a focused and correct resource-release fix, and I am happy to approve it on the latest head.

DanielLeens:

- base branch: `dev`
- head SHA: `2597841b549372a85d6e1fa6cd2546e3d513bb86`
- compare status: `diverged`
- ahead_by: `12`
- behind_by: `23`
- mergeable_state: `unknown`

davidzollo (approved):
+1
Good job

PR #11134, opened 2026-06-19
[Fix][Connector-V2] Keep socket reader stream open

### LINKED ISSUE #10528: [Bug][Connector-V2][Socket] SocketSourceReader closes socket prematurely on each pollNext() call
### Search before asking

### What happened

`SocketSourceReader.pollNext()` wraps `socket.getInputStream()` in a try-with-resources `BufferedReader`. When the try block exits (normally or via exception), Java closes the `BufferedReader`, which cascades to close the underlying `InputStreamReader` and then the socket's `InputStream` — which closes the `Socket` itself (per Java docs: *"Closing the InputStream will close the associated socket"*).

As a result, after the first `pollNext()` call returns (e.g. after reaching EOF or `signalNoMoreElement()` in BOUNDED mode), the socket is closed. Any subsequent `pollNext()` call throws `SocketException: Socket is closed` because `socket.getInputStream()` is called on an already-closed socket.

**Affected code (`SocketSourceReader.java`):**
```java
@Override
public void pollNext(Collector<SeaTunnelRow> output) throws Exception {
    StringBuilder buffer = new StringBuilder();
    try (BufferedReader reader =
            new BufferedReader(new InputStreamReader(socket.getInputStream()))) {
        // ... reads records ...
    }
    // ↑ try-with-resources closes BufferedReader → closes InputStream → closes Socket
    // Next pollNext() call will get SocketException: Socket is closed
}
```

**Fix:** Initialize the `BufferedReader` in `open()` as an instance field and close it in `close()`, so the socket's lifetime is managed by the reader lifecycle, not by `pollNext()`.

### SeaTunnel Version

dev (latest)

### SeaTunnel Config

```hocon
source {
  Socket {
    host = "localhost"
    port = 9999
  }
}
```

zhangshenghang:
### Purpose

Fixes #10528.

`SocketSourceReader.pollNext()` created a new `BufferedReader` with try-with-resources on every poll. Closing that reader also closed the socket input stream, so the next `pollNext()` call could fail with `Socket is closed` even though the source reader itself had not been closed.

### Changes

- Create the socket `BufferedReader` once when the source reader opens.
- Keep the socket input stream alive across multiple `pollNext()` calls.
- Close the reader during the source reader lifecycle close step.
- Add a unit test that verifies two consecutive polls can read from the same socket connection.

### Tests

- `./mvnw -pl seatunnel-connectors-v2/connector-socket -am -DskipIT -Dtest=SocketSourceReaderTest -DfailIfNoTests=false verify`
- `./mvnw spotless:apply`


DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff from the real socket source reader lifecycle, not just the added test.

## What this PR fixes
- Pain point: `SocketSourceReader.pollNext(...)` used to create a new `BufferedReader` inside a try-with-resources block, so the underlying socket input stream was closed as soon as one polling round finished. In bounded mode, the next `pollNext(...)` call could then hit an already-closed stream.
- Approach: promote the `BufferedReader` to a reader field, initialize it in `open()`, reuse it in `pollNext(...)`, and release it once in `close()`.
- Simple example: after reading `"first"` in one bounded `pollNext(...)` call, the reader can now still consume `"second"` from the same socket in the next call.

```text
Socket source startup
  -> SocketSourceReader.open() [SocketSourceReader.java:49-59]
      -> socket.connect(...)
      -> reader = new BufferedReader(socket.getInputStream())

read loop
  -> pollNext(output) [SocketSourceReader.java:71-97]
      -> reader.read(buf)
      -> split by delimiter
      -> bounded mode => context.signalNoMoreElement() and break

lifecycle end
  -> close() [SocketSourceReader.java:61-68]
      -> reader.close()
```

## Findings
- The normal source lifecycle definitely hits this change because a source reader is expected to survive multiple `pollNext(...)` calls between `open()` and `close()`.
- The old bug is real on that lifecycle: the try-with-resources in `pollNext(...)` was closing the stream too early.
- Moving reader ownership back to `open()/close()` is the right fix at the right layer.
- I did not find a new correctness, compatibility, or resource-management blocker on the current head.

- CPU/memory impact is negligible.
- Resource handling is improved rather than weakened, since the stream is now closed exactly once at the reader lifecycle boundary.

## Test coverage / stability
- `SocketSourceReaderTest.java:48-70` reproduces the real regression by calling `pollNext(...)` twice against the same socket.
- The test uses an ephemeral local port and a single deterministic writer sequence, so I do not see an obvious flaky pattern. I would rate it as stable.

## Merge conclusion
### Conclusion: can merge

1. Blockers
- None on the current head.

2. Non-blocking suggestions
- None.

Overall, this is a focused lifecycle fix with the right regression coverage, and it looks good to merge from my side.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the real socket source reader lifecycle, including `open() -> pollNext() -> close()` and the new regression test.

## What this PR fixes
- Pain point: `pollNext()` recreated the `BufferedReader` in a try-with-resources block, so the first read closed the underlying socket stream and later polls could not continue reading from the same connection.
- Approach: keep one reader for the reader lifecycle, initialize it in `open()`, reuse it across `pollNext()` calls, and close it in `close()`.
- One-line summary: this restores the expected “same socket, multiple pollNext calls” behavior.

## Runtime path I checked
```text
SocketSourceReader.open()
  -> connect socket
  -> create one BufferedReader over socket input stream

SocketSourceReader.pollNext()
  -> read from the shared reader
  -> emit completed records
  -> keep the stream open for the next poll

SocketSourceReader.close()
  -> close reader / socket at lifecycle end
```

## Findings
- I do not see a new source-path blocker on the current head.
- The core fix is on the hot path and matches the bug being fixed.
- The new test is focused and structurally stable; I do not see a hard-sleep or weak-ready-signal flaky pattern in it.

## CI note
- The current `Build` check is red on the latest head, with visible failures in broad suites such as `unit-test (11, windows-latest)` and `all-connectors-it-7`.
- From the two-file diff here, I do not see evidence that this socket reader fix would explain unrelated integration suites. So from source-level review I do not have a new code blocker, but I would still wait for CI to be sorted before the final merge action.

## Merge conclusion
### Conclusion: can merge

1. Blocking items

- None from the code path I checked.

Overall, the fix is straightforward and the latest head looks correct from the runtime path I traced.

CosmosNi (approved):
Good job～

DanielLeens:
Thanks @CosmosNi. I checked the current head again, and it is still the same commit I already approved earlier, so I do not have additional source-path concerns from my side. The remaining red `Build` signal still looks broader than this socket-reader change. Once CI is sorted out, this should be ready to move.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the real socket reader lifecycle again, including the current `open() -> pollNext() -> close()` path and the new regression test.

What this PR fixes
- User pain: the old reader recreated and closed the input stream inside `pollNext()`, so a bounded socket read could consume one batch and leave the next poll on a closed stream.
- Fix approach: the reader is now created once in `open()`, reused across `pollNext()` calls, and closed with the reader lifecycle instead of the per-poll scope.
- One-line summary: the socket source now keeps the stream alive for the whole reader lifetime, which is the lifecycle SeaTunnel expects.

Runtime path I checked
```text
SocketSourceReader.open()
  -> create socket
  -> connect to remote host
  -> create BufferedReader once

SocketSourceReader.pollNext()
  -> read bytes from the same BufferedReader
  -> split records by newline
  -> emit SeaTunnelRow

SocketSourceReader.close()
  -> close reader/socket at the end of the reader lifecycle
```

Findings
- The fix is on the real normal path because every socket source read goes through this exact `SocketSourceReader` lifecycle.
- The added regression test is targeted and deterministic: it verifies that two successive `pollNext()` calls can still read from the same accepted socket session. I do not see a flaky timing pattern in the test structure itself.
- I noticed @CosmosNi already approved this PR earlier. After tracing the same runtime path on the latest head, I agree with that merge direction.

Compatibility
- Fully compatible. No config contract changes, no protocol changes, and no API changes. The behavior only shifts from premature stream closure to the intended reader-lifecycle ownership.

Merge conclusion
### Conclusion: can merge

1. Blocking items

- None required for merge.

This still looks like the right fix for the actual socket lifecycle bug, and I am fine with merging it.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head `90561d047f` from the real socket reader lifecycle again, including the current `open() -> pollNext() -> close()` path and the regression test after the recent `upstream/dev` merge.

## What this PR fixes
- Pain point: the old reader recreated and closed the input stream inside `pollNext()`, so one bounded read could leave the next poll on an already-closed socket stream.
- Approach: create the `BufferedReader` once in `open()`, reuse it across `pollNext()` calls, and close it at the reader lifecycle boundary in `close()`.
- One-line summary: the socket source now keeps the stream alive for the whole reader lifetime, which is the lifecycle SeaTunnel expects.

## Runtime path I checked
```text
SocketSourceReader.open()
  -> connect socket
  -> create one BufferedReader over socket input stream

SocketSourceReader.pollNext()
  -> read from the shared reader
  -> split records by newline
  -> emit SeaTunnelRow
  -> keep the stream open for the next poll

SocketSourceReader.close()
  -> close reader / socket at lifecycle end
```

## Findings
- The normal source-reader lifecycle definitely hits this change; this is not a test-only path.
- The old bug is real on that lifecycle: the try-with-resources block in `pollNext()` closed the stream too early.
- The current latest head still keeps the fix focused and I do not see a reopened source-level blocker.
- I noticed @CosmosNi already approved this PR earlier. After tracing the same runtime path on the latest head, I still agree with that merge direction.

## CI note
- The current red jobs are broad suites such as `jdbc-connectors-it-ddl`, `kafka-connector-it`, `rocketmq-connector-it`, and `kudu-connector-it`.
- From this two-file socket source fix, I do not see a convincing source-level link to those failures.
- The branch has already absorbed the recent workflow fixes from `upstream/dev`; it is only one unrelated collaborator update behind the current `dev` head, so I do not have a smaller source-side sync to suggest here.

## Merge conclusion
### Conclusion: can merge

1. Blocking items
- None on the current head.

- None from the socket reader lifecycle path I checked.

Overall, this still looks like the right fix for the actual socket lifecycle bug, and I am happy to approve it on the latest head.

DanielLeens:

- base branch: `dev`
- head SHA: `90561d047f838c7a65f9358ae6f5b0d6fc96928f`
- compare status: `diverged`
- ahead_by: `5`
- behind_by: `23`
- mergeable_state: `unknown`

davidzollo (approved):
+1
LGTM

PR #11137, opened 2026-06-19
[CI][E2E] Stabilize SQL Server XA setup for schema evolution

DanielLeens:
## What's changed
- switch the SQL Server container setup to the supported `MSSQL_SA_PASSWORD` variable and configure the MSDTC RPC/DTC ports required by containerized XA transactions
- wait for a real `sqlcmd` login before running XA setup so the test does not race SQL Server upgrade work
- fail fast when `sp_configure` or `sp_sqljdbc_xa_install` fails, and wait until `xp_sqljdbc_xa_init_ex` becomes visible before the exactly-once path starts
- remove the invalid `external user enabled` setup step and relax the ready-log matcher so startup does not depend on a trailing newline

## Validation
- `./mvnw -nsu -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl spotless:apply`
- `./mvnw -nsu -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl -q -DskipTests verify`
- `./mvnw -nsu -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl -DfailIfNoTests=false -Dtest=SqlServerSchemaChangeIT test`
  - on `/Volumes/disk2/work/myworkspace/tmp/seatunnel-dev-sqlserver-xa-init-20260620`, the host filesystem exposed `src/test/resources/docker/setup.sql` to the MySQL container with unreadable permissions and the run stopped before the SQL Server path (`Permission denied`)
  - on `/Users/stone/work/myworkspace/seatunnel_third` with normal file permissions, MySQL initialization succeeded and SQL Server 2022 still failed under the local Docker Desktop memory cap before reaching the ready log (`Available Physical Memory 89.32 mb`, `Error 701`, `Failed allocate pages`)
- `seatunnel-engine-ui` was skipped because this change does not touch that module
- no modules or deployment artifacts were reused in this run

## References
- https://learn.microsoft.com/en-us/sql/linux/containers/configure-distributed-transactions?view=sql-server-ver17
- https://learn.microsoft.com/en-us/sql/connect/jdbc/understanding-xa-transactions?view=sql-server-ver17
- https://learn.microsoft.com/el-gr/sql/linux/configure/environment-variables?view=sql-server-linux-ver17

PR #11138, opened 2026-06-20
[Test][E2E] Stabilize Db2 container readiness for JDBC IT

DanielLeens:
## What does this PR do?

This fixes an unrelated JDBC Db2 E2E startup failure that surfaced while validating PR #11137.

The Db2 test was using `ibmcom/db2:latest` and a short startup readiness configuration. On slow runners, the container can spend several minutes in instance initialization before the database is actually usable, which makes the E2E fail for infrastructure timing reasons instead of test logic reasons.

## Changes

- pin the Db2 image to `ibmcom/db2:11.5.0.0a`, which matches the Testcontainers validated default instead of relying on `latest`
- wait for the official `Setup has completed.` readiness log and extend the startup timeout to 20 minutes for the Db2 test container
- keep the change scoped to `JdbcDb2IT`

## Validation

- `./mvnw -nsu -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-1 spotless:apply`
- `./mvnw -nsu -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-1 -DfailIfNoTests=false -Dtest=JdbcDb2UpsertIT test`

Local runtime note:

- this machine's Docker Desktop is limited to about 1975 MB RAM, so the Db2 container still did not fully reach the final ready log locally
- however, after this change the container progressed well past the previous failure point: it completed instance configuration, ran `DB2START`, and reached `Creating database E2E`, which is materially further than the original startup timeout behavior
- `seatunnel-engine-ui` was not changed and was not included in this validation scope


davidzollo (commented):
+1
LGTM

davidzollo (approved):

PR #11139, opened 2026-06-20
[Test][Zeta] Wait for final canceled state in role test

DanielLeens:
## What does this PR do?

This is a follow-up CI baseline fix split out from #11136.

`#11136` hit a new unrelated failure in `unit-test (8, ubuntu-latest)`:
`SeaTunnelEngineClusterRoleTest.testWorkerIsFirstMemberThenGetJobDetailStatus`

The failure was not in the Pulsar code path. The test was asserting `jobClient.getJobStatus(jobId)` directly after `cancelJob(jobId)` during master failover, and CI showed the transient state could remain `CANCELING` before the final terminal result became visible.

This change makes the test wait for `clientJobProxy.waitForJobCompleteV2()` first, then asserts the final client-visible status is `CANCELED`.

## Why is this needed?

The previous version only increased timeout and polling, but the failing CI still timed out on the transient `CANCELING` state. Waiting on the final `JobResult` matches the real completion boundary and avoids sampling the intermediate state as a terminal assertion.

## Validation

Ran:
- `./mvnw -nsu -pl seatunnel-engine/seatunnel-engine-client spotless:apply`

Attempted but blocked by unrelated local compile issues already present on the clean `apache/dev` worktree:
- `./mvnw -nsu -pl seatunnel-engine/seatunnel-engine-client -am -DfailIfNoTests=false -Dtest=SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus test`
- `./mvnw -nsu -pl seatunnel-engine/seatunnel-engine-client -DfailIfNoTests=false -Dtest=SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus test`

The local blocker was not this test change itself:
- reactor build path hit missing local `seatunnel-config-shade` sources in this machine's worktree
- module-only path hit existing `MultipleTableJobConfigParserTest` testCompile errors on the clean `apache/dev` checkout before this change could be exercised

`seatunnel-engine-ui` was not changed and was not revalidated in this run.


nzw921rx (approved):
+1 LGTM

DanielLeens (commented):
I re-reviewed the latest head from the engine client cancellation path again.

## What this PR solves
- User pain: after cancel is accepted, the client-side status string can still briefly stay at `CANCELING`, so `SeaTunnelEngineClusterRoleTest` can fail even though the job result has already reached the terminal cancel path.
- Fix approach: wait for `ClientJobProxy.waitForJobCompleteV2()` first, then assert the externally visible status converges to `CANCELED`.
- One-line summary: the runtime reasoning behind this test fix still looks correct on the latest head.

## Current CI note
- The latest failing lanes are `all-connectors-it-1` and `all-connectors-it-7`.
- Those are outside the engine client test path changed here, so I do not currently see evidence that this PR is what made the current run red.

## Review conclusion
1. Blocking items
- No new source-level blocker from my side on the latest head.
- Please get the current `Build` green before merge.

2. Suggested follow-up
- None beyond the CI gate.

DanielLeens:
Thanks for the update. I rechecked the latest head on the cluster-role cancel path again.

## Current state on the latest head
- I do not see a reopened source-side blocker in the current revision of this test change.
- The pending-job cancel path still looks consistent with the intended client-visible `CANCELED` status assertion on the latest head.
- The current `Build` run is still queued, so the remaining gate is CI rather than a new blocker from Daniel's side.

So from the Daniel technical-review side, I do not have a new source blocker on the current revision.

DanielLeens (commented):
Rechecked the latest head from Daniel's side. I still do not see a new source-side blocker on this head; the thing that changed since the earlier round is the CI / merge-gate fact, not the code-side conclusion.

Current gate
- The current `Build` check is still failing.
- This branch is also behind the latest `dev` (`behind_by=37`, compare status: `diverged`).

Next step
- Since the remaining blocker here is the stale CI / merge gate rather than a newly identified source issue, the lowest-cost next step is to sync the latest `dev` and rerun `Build` first.
- If `Build` still fails after the sync, we can narrow it down again on the refreshed head.

DanielLeens (commented):
Technical self-review of the full diff at `f54c3a9c546abf61ac7ef895f8458ca8252a9937`: waiting for the terminal result is the right semantic boundary, but the new test wait is not bounded and can turn the original flaky failure into a hung CI job.

1. **`waitForJobCompleteV2()` bypasses the test's timeout.** `SeaTunnelEngineClusterRoleTest.java:518-531` calls this blocking method before entering the 30-second Awaitility block. `ClientJobProxy.java:93-106` waits on `jobFuture.get()` and is configured for up to 100,000 retries with a 2-second retry sleep. If cancellation or failover regresses, the test no longer fails within the previous 120-second bound; it can hold a CI executor for many hours. Please execute the terminal-result wait under a real test deadline (for example, a bounded future or JUnit timeout that can interrupt the wait), then keep the final `CANCELED` assertion.

This is a high-risk test-stability blocker for a PR whose purpose is to stabilize shared CI. Also, #11189 now carries the same code change with the missing method documentation included; please keep only one of these duplicate PRs active. The current required `Build` is still queued, and this self-authored PR still needs an independent maintainer decision after the blocker is fixed.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `f54c3a9c546a` from scratch.

# Findings
**Issue 1: waitForJobCompleteV2() is an unbounded blocking call — if the job never reaches a terminal state during master failover, the test hangs until the CI job-level timeout instead of failing with a diagnosable assertion**
- Location: `seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java:524`
- Why it matters: - Root cause: the previous check used `await().atMost(120000, TimeUnit.MILLISECONDS)`, so a stuck job produced a clear `ConditionTimeoutException` within 2 minutes. The new code calls `clientJobProxy.waitForJobCompleteV2()` directly on the test thread with no timeout guard, and this method blocks until the client observes a terminal `JobResult`.
- Impact: this test exercises exactly the scenario (master shutdown + failover + cancel) where the completion signal can be lost or delayed indefinitely. If that happens, the test thread hangs, JUnit never reports a failure for this test, and the whole `unit-test` CI job times out at the workflow level — which is much harder to triage than the flaky assertion this PR is trying to fix, and it also delays the `finally` block that shuts down the Hazelcast client and cluster nodes.
- Fix: bound the wait, e.g. `Assertions.assertTimeoutPreemptively(Duration.ofMinutes(2), clientJobProxy::waitForJobCompleteV2)` or run it via `CompletableFuture.supplyAsync(clientJobProxy::waitForJobCompleteV2).get(120, TimeUnit.SECONDS)`, so a hang surfaces as a test failure with a stack trace instead of a CI-level timeout.
- Evidence: `seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java:525          Assertions.assertEquals(JobStatus.CANCELED, jobResult.getStatus());`
- Suggested fix: `waitForJobCompleteV2()` blocks without a timeout. In this failover scenario, if the terminal `JobResult` is never delivered to the client, the test thread hangs and the whole CI job times out at the workflow level (and the `finally` cluster shutdown is delayed). Please bound it, e.g. `Assertions.assertTimeoutPreemptively(Duration.ofMinutes(2), clientJobProxy::waitForJobCompleteV2)`, so a stuck job fails fast with a useful stack trace like the old 120s Awaitility guard did.
- Severity: High

**Issue 2: Single blocking waitForJobCompleteV2 call has no tolerance for transient RPC failures during master switchover, unlike the polling loop it replaces**
- Location: `seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java:484`
- Why it matters: - Root cause: the previous Awaitility loop retried `getJobStatus` on every assertion failure, so a transient exception thrown during the master switchover (connection reset, `TargetNotMemberException`-style errors while the client re-routes to the new master) was absorbed by the next poll. `clientJobProxy.waitForJobCompleteV2()` is invoked exactly once with no try/catch or retry, and this test deliberately shuts a master down right before cancel.
- Impact: any transient client-side exception surfacing from the single blocking call fails the test immediately, trading one flake pattern (`CANCELING` sampled as terminal) for another (spurious RPC error during failover). It's worth verifying whether `waitForJobCompleteV2()` internally retries on member-list changes; if it does not, this test remains exposed.
- Fix: wrap the wait in a small retry, e.g. move the `waitForJobCompleteV2()` call inside an `await().atMost(...).until(...)` / `untilAsserted` block or a bounded retry loop that tolerates transient exceptions during failover, then assert the final status once.
- Evidence: `seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java:524          JobResult jobResult = clientJobProxy.waitForJobCompleteV2();`
- Suggested fix: The old Awaitility loop retried through transient RPC errors while the client re-routed to the new master; this single `waitForJobCompleteV2()` call has no such tolerance, so a connection blip during the failover this test induces now fails the test outright. Consider wrapping the wait in a bounded retry (or an `untilAsserted` block) that absorbs transient exceptions before asserting the terminal status.
- Severity: Medium

**Issue 3: Hardcoded "CANCELED" string literal instead of JobStatus.CANCELED.name()**
- Location: `seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java:530`
- Why it matters: - Root cause: the follow-up await compares against the raw string literal `"CANCELED"`, while line 525 in the same method and the pre-existing assertions in this test use the `JobStatus` enum (`JobStatus.CANCELING.name()` / `JobStatus.CANCELED.name()` in the removed block).
- Impact: purely a consistency/maintainability issue — if the client status string format ever changes, the enum-based comparisons would be caught at compile/refactor time while the literal would silently drift.
- Fix: use `Assertions.assertEquals(JobStatus.CANCELED.name(), jobClient.getJobStatus(clientJobProxy.getJobId()))` to keep both assertions in the helper anchored to the same enum.
- Suggested fix: Minor: use `JobStatus.CANCELED.name()` instead of the `"CANCELED"` literal, matching the enum-based assertion two lines above and the convention used elsewhere in this test.
- Severity: Low

**Issue 4: Config/option changes without doc updates**
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: waitForJobCompleteV2() is an unbounded blocking call — if the job never reaches a terminal state during master failover, the test hangs until the CI job-level timeout instead of failing with a diagnosable assertion

- Issue 2: Single blocking waitForJobCompleteV2 call has no tolerance for transient RPC failures during master switchover, unlike the polling loop it replaces
- Issue 3: Hardcoded "CANCELED" string literal instead of JobStatus.CANCELED.name()
- Issue 4: Config/option changes without doc updates

DanielLeens:
Thanks @SEZ9 for the independent review. I agree with Issue 1: the direct `waitForJobCompleteV2()` call is outside a real test deadline, so a lost terminal result can hold the CI executor for many hours. That is the same technical blocker recorded in my self-review.

One clarification on Issue 2: `ClientJobProxy.waitForJobCompleteV2()` already wraps `doWaitForJobComplete().get()` in `RetryUtils.retryWithException(...)` and retries operation-level transient failures (`ClientJobProxy.java:93-106`). The missing contract here is a bounded overall deadline, not the absence of transient-operation retry. The literal-string consistency note is non-blocking.

This self-authored PR remains blocked until the wait is bounded and still needs independent maintainer approval afterward.

DanielLeens:
Closing this older duplicate because #11457 has already landed on `dev` and stabilizes the same `SeaTunnelEngineClusterRoleTest` with the current cancellation semantics (`cancelJob(jobId, true)` and accepting `CANCELING` or `CANCELED`). Keeping this PR would reintroduce the older behavior. Thanks.

PR #11141, opened 2026-06-20
[Fix][API] Correct numeric epoch timestamp conversion

fallintoplace:

Fix `BasicDataConverter` numeric DATE/TIMESTAMP conversion for modern epoch-second values when timestamp precision metadata is available.

The old shared fallback treated values like `1700000000` as epoch milliseconds, which converted a normal 2023 epoch-second timestamp into a date in 1970. This patch routes `BasicTypeDefine` numeric timestamp conversion through timestamp scale metadata when present:
- scale `0` -> epoch seconds;
- scale `3` -> epoch milliseconds;
- scale `6` -> epoch microseconds;
- scale `9` -> epoch nanoseconds.

For no-metadata conversion, the legacy seconds/milliseconds cutoff is preserved so early-epoch millisecond values such as `1000000000` are not reinterpreted as seconds.

Yes. Numeric DATE/TIMESTAMP values with timestamp scale metadata now use that scale to choose epoch seconds, milliseconds, microseconds, or nanoseconds. No-metadata numeric conversion keeps the previous fallback behavior.

Added unit coverage for scale-aware numeric epoch seconds, milliseconds, microseconds, and nanoseconds in `BasicDataConverter`, plus regression coverage for an early-epoch millisecond value without scale metadata.

- `./mvnw -pl seatunnel-api -Dtest=org.apache.seatunnel.api.table.converter.BasicDataConverterTest -DskipITs -Dcheckstyle.skip -Dspotless.check.skip test`
- `./mvnw -pl seatunnel-api -DskipITs -Dcheckstyle.skip -Dspotless.check.skip test`
- `./mvnw -pl seatunnel-api spotless:check`

### Check list

* [x] Documentation is not needed for this bug fix.
* [x] No incompatible change file update is needed.
* [x] No connector packaging changes are needed.


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current diff from the real `BasicDataConverter` number-to-time conversion path, not just the new unit tests. The direction makes sense, but I do see one blocker before merge.

- Pain point: the current `BasicDataConverter` heuristics can treat modern epoch-second values like `1700000000` as milliseconds, which produces the wrong `LocalDateTime` / `LocalDate`.
- Approach: move the logic into `EpochTimeUtils.convertToInstant(...)` and distinguish seconds / millis / micros / nanos by magnitude before the shared `convertLocalDateTime(Number)` and `convertLocalDate(Number)` paths use it.
- Simple example: with this patch, `1700000000` no longer lands in 1970 just because the converter assumed it was milliseconds.

```text
numeric field declared as DATE / TIMESTAMP / LOCAL_DATE_TIME
  -> BasicDataConverter.convert(...)
      -> convertLocalDateTime(typeDefine, Number)
         or convertLocalDate(typeDefine, Number)
          -> BasicDataConverter.convertLocalDateTime(Number) [BasicDataConverter.java:606-609]
             or convertLocalDate(Number) [BasicDataConverter.java:793-795]
              -> EpochTimeUtils.convertToInstant(Number) [EpochTimeUtils.java:33-51]
                  -> infer seconds / millis / micros / nanos by magnitude
              -> atZone(systemDefault)
```

## Findings

### 1. The new magnitude heuristic fixes modern epoch-seconds, but it also regresses a valid epoch-millis range on the same shared path
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/converter/EpochTimeUtils.java:37-43`
- Why this is a problem:
  - The new rule treats every absolute value below `10_000_000_000` as epoch seconds.
  - That does fix modern second-based values such as `1700000000`.
  - But it also misclassifies legitimate millisecond values near the Unix epoch. For example, `1_000_000_000` as milliseconds should be `1970-01-12T13:46:40Z`, while the new code will interpret it as seconds and turn it into `2001-09-09T01:46:40Z`.
  - Because `BasicDataConverter.convertLocalDateTime(Number)` and `convertLocalDate(Number)` now both route through this helper, this is not a side path; it is the shared default conversion contract.
- Risk:
  - This is a correctness regression on a public API-level conversion path. Users can get the wrong date/time back from valid numeric input.
- Suggested fix:
  - Option A: avoid widening the shared heuristic in `BasicDataConverter` unless the caller can provide explicit unit metadata.
  - Option B: if you keep heuristic inference, tighten it so you do not regress valid epoch-millis values in the early-epoch window.
- Severity: High

### 2. Non-blocking: please document the new shared inference contract explicitly
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/converter/EpochTimeUtils.java:23-51`
- Why this matters:
  - `EpochTimeUtils` now defines the default unit-inference contract for the shared number-to-date/time conversion path.
  - Right now the behavior is only encoded in thresholds and helper math, which makes future maintenance and compatibility review harder.
- Suggested fix:
  - Add a short class/method comment documenting the inference rules, the intended unit order, and the known ambiguity limits.
- Severity: Medium

- This is not an API signature break, but it is a behavior change on the default conversion path.
- The main issue is not CPU or memory cost; it is that the new heuristic expands the millis-to-seconds misclassification window for older timestamps.

## Test coverage / stability
- The new `BasicDataConverterTest` covers modern seconds / millis / micros / nanos cases.
- The missing regression case is an early-epoch millisecond value such as `1_000_000_000L`, which is exactly where the current blocker still lives.
- The added tests themselves look structurally stable to me.

## Merge conclusion

1. Blockers
- Finding 1: the new shared unit heuristic still misparses a valid epoch-millis range on the main conversion path.

2. Non-blocking suggestions
- Finding 2: please document the new inference contract explicitly.

Overall, I agree with the goal of fixing modern epoch-second handling, but the current implementation trades one shared conversion bug for another. Happy to re-review once the heuristic no longer regresses valid epoch-millis inputs.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the actual numeric-timestamp conversion path in `BasicDataConverter`, including the new shared helper and the regression tests.

## What this PR fixes
- Pain point: numeric epoch values can be interpreted with the wrong precision, and the old no-metadata path also used the current offset instead of the offset at the target instant.
- Approach: centralize epoch conversion in `EpochTimeUtils`, use scale-aware conversion when `BasicTypeDefine` carries timestamp precision, and keep the legacy seconds-vs-millis heuristic for the no-metadata path.
- One-line summary: this makes the timestamp conversion path more consistent without changing the fallback contract that existing callers rely on.

## Runtime path I checked
```text
typed conversion
  -> BasicDataConverter.convertByType(...)
      -> convertLocalDateTime(typeDefine, Number) / convertLocalDate(typeDefine, Number)
      -> EpochTimeUtils.convertToInstant(typeDefine, value)
      -> scale 0/3/6/9 => second / millisecond / microsecond / nanosecond

legacy no-metadata conversion
  -> convertLocalDateTime(Number) / convertLocalDate(Number)
      -> EpochTimeUtils.convertToInstant(value)
      -> keep the old shared heuristic: small values as epoch seconds, larger values as epoch millis
```

## Findings
- I do not see a new correctness blocker on the current head.
- The helper keeps the legacy no-metadata contract intact while fixing the typed precision path.
- The added unit tests cover the main typed second/millis/micros/nanos cases plus the early-epoch millisecond fallback.

## Test note
- The new UTs are deterministic: no sleeps, no external state, no port usage, and no order-sensitive assertions.
- If you want one small follow-up later, adding a micro/nano test with a non-zero fractional remainder would make the `floorDiv/floorMod` path even more explicit, but I do not treat that as a blocker.

## Merge conclusion
### Conclusion: can merge

1. Blocking items

- Optional: add one fractional micro/nano regression case.

Overall, the latest head looks solid to me. The conversion logic is cleaner, and I do not see a compatibility regression in the path I traced.

davidzollo:
Good job.

To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration https://github.com/apache/seatunnel/pull/11141/checks?check_run_id=82486717815

image

DanielLeens:
Thanks for the CI follow-up. From Daniel's side, the current head is still the same revision I already approved, and I do not have a new source-level concern to add on this version.

The remaining step here is just to get the fork workflow enabled so the `Build` check can run normally. If you hit any trouble while enabling it, feel free to leave a note and I can take another look.

DanielLeens (approved):
Thanks for the CI follow-up. I re-reviewed the latest head from the actual numeric epoch conversion path again, using the current tree rather than carrying over the previous approval mechanically.

- The current head only adds `[Chore] Trigger CI`.
- I rechecked the executable code path in `BasicDataConverter` and the precision-aware helper in `EpochTimeUtils`; there is no new source delta on top of the version Daniel already approved.

Runtime path I checked
```text
numeric epoch value
  -> BasicDataConverter.convertTimestamp(...)
      -> EpochTimeUtils.convertToInstant(typeDefine, value)
          -> scale 0 / 3 / 6 / 9 use explicit second / milli / micro / nano conversion
          -> other or missing scale falls back to the legacy shared heuristic
```

Findings
- The latest head does not reopen the earlier API-path blocker.
- The microsecond / nanosecond split still routes through the dedicated helper rather than the old ambiguous shared heuristic.
- The regression coverage in `BasicDataConverterTest` still matches the runtime path Daniel reviewed earlier.

CI
- The current `Build` is queued on the latest head.
- From Daniel's side, the remaining gate is CI completion rather than a reopened API correctness issue.

1. Blocking items
- Please let the latest queued `Build` finish cleanly.

- None from this round.

DanielLeens:

- base branch: `dev`
- head SHA: `e136f48410abea6895b0046b7b08c3f9d79a8353`
- compare status: `diverged`
- ahead_by: `3`
- behind_by: `48`
- mergeable_state: `unknown`

DanielLeens:

- base branch: `dev`
- head SHA: `de7dbd5ae0e7543bdcbb043dcf3e851710d71703`
- compare status: `diverged`
- ahead_by: `3`
- behind_by: `8`
- CI / merge gate snapshot: `Build CANCELLED`

DanielLeens (commented):

# What this PR fixes
- PR: [Fix][API] Correct numeric epoch timestamp conversion

```text
BasicDataConverter timestamp conversion
  -> numeric epoch input reaches BasicDataConverter
  -> EpochTimeUtils normalizes seconds/millis/micros/nanos fallback
  -> SeaTunnel temporal type keeps the expected instant precision
```

# Current review findings
- Changed files in the current full diff: 3.
- Base sync fact recorded: compare status `diverged`, `behind_by=1`. This is recorded as queue metadata; I am not turning it into a standalone sync request unless it is needed to interpret a failing CI signal.

- Fetched the PR head into a local review branch named `seatunnel-review-11141`.

# Merge conclusion

DanielLeens (commented):
Rechecked the same head after the CI fact changed. From Daniel's side I still do not see a reopened source-level blocker on this revision; the current remaining gate is the `Build` workflow run (Pull Request Labeler) here: https://github.com/apache/seatunnel/runs/85793160348

This branch is currently behind the latest `dev` (`compare=diverged`, `behind_by=8`), so the lowest-cost next step is to sync with the latest `dev` and rerun `Build` first. If the same lane still fails after the sync, please paste the updated run link and I can help narrow it down against the refreshed head.

Note that any earlier Daniel review summary may still show historical state in GitHub's sidebar because I only have read-level review controls on this repository; this follow-up comment is the current Daniel-side technical conclusion for the same head.

davidzollo:

```

```

PR #10849, opened 2026-05-07
[Feature][Connector-V2] Add Google Cloud Bigtable Source and Sink con…

programmerloverun:
## Purpose

  Closes #10847

  Add a native Google Cloud Bigtable connector supporting both Source (read) and
   Sink (write) operations.

  ## Changes

  ### New Module: `connector-google-bigtable`

  **Sink**
  - `BigtableSinkWriter`: batch writes via `BulkMutation`; configurable batch
  size, null mode (SKIP/EMPTY), row key composition (single/composite with
  delimiter), multi-column-family mapping, version column for cell timestamp
  - `BigtableSink` / `BigtableSinkFactory`: SPI registration, `schema_save_mode`
   / `data_save_mode` support

  **Source**
  - `BigtableSourceReader`: full table scan with row key range, timestamp range
  filter, max-versions filter, scan row limit
  - `BigtableSourceSplitEnumerator`: single split (full table or user-defined
  row key range)
  - `BigtableSource` / `BigtableSourceFactory`: SPI registration, bounded source

  **Shared**
  - `BigtableClient`: wraps `google-cloud-bigtable:2.39.1` Data v2 client;
  supports service account JSON credentials and Application Default Credentials
  (ADC)
  - `BigtableDeserializationFormat`: `ByteString` → `SeaTunnelRow` type mapping;
   numeric types as big-endian binary, string/date types as UTF-8 text
  - Full `SeaTunnelErrorCode` error code definitions

  ### Documentation
  - `docs/en/connectors/sink/GoogleBigtable.md`
  - `docs/en/connectors/source/GoogleBigtable.md`
  - `docs/zh/connectors/sink/GoogleBigtable.md`
  - `docs/zh/connectors/source/GoogleBigtable.md`

  ## Test

  - `BigtableSinkWriterTest`: batch flush on size threshold, flush on close,
  null field skip
  - `BigtableSourceSplitEnumeratorTest`: full range split, custom row key range
  split
  - All 5 unit tests pass; Spotless format check passes

  ## Known Limitations

  - `schema_save_mode=CREATE_SCHEMA_WHEN_NOT_EXIST`: table/column-family
  creation via Admin API is not yet implemented; users must create them manually
  - `data_save_mode=DROP_DATA`: table truncation via Admin API is not yet
  implemented
  - Source currently generates a single split (no parallel read); tablet
  boundary partitioning can be added in a follow-up
  - TTL must be configured at column-family level in Bigtable; per-cell TTL is
  not supported by the connector

PR #11144, opened 2026-06-20
[Improve][Connector-V2] Harden Google Bigtable source-side state synchronization

programmerloverun:
## Purpose

Follow-up to #10849 (Google Cloud Bigtable connector), addressing @davidzollo's post-merge review request to harden the source-side state synchronization.

A plain `private final Object stateLock` (no `ReentrantLock`, no line-by-line synchronization) is used to guard compound source-side state transitions so that a concurrent checkpoint never observes a half-updated view of the reader/enumerator state.

### `BigtableSourceReader`
- Protect the `pollNext(...)` state-transition block only: poll `pendingSplits` + assign `currentSplit`, and clear `currentSplit` after the split read completes.
- Protect `snapshotState(...)` and `addSplits(...)`.
- The `readRows(...)` network scan stays outside `stateLock`; `output.getCheckpointLock()` is still used only around `output.collect(...)`.
- `handleNoMoreSplits()` keeps the existing `volatile` flag (no lock).

### `BigtableSourceSplitEnumerator`
- Protect `addSplitsBack(...)`, `registerReader(...)`, and `snapshotState(...)`.
- `initializePendingSplits()` and `assignSplit()` run only while the lock is held, guarding `assignedSplits`, `pendingSplits`, and `initialized`.

### `BigtableSourceState`
- Hold a defensive copy of the assigned split set instead of aliasing the enumerator's live set.

No sink-side locking changes (out of scope for this follow-up).

## Test

- `./mvnw spotless:check test -pl seatunnel-connectors-v2/connector-google-bigtable` on JDK 11: all 13 unit tests pass and the Spotless check passes.

### DIFF
--- a/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceReader.java
+++ b/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceReader.java
@@ -65,6 +65,13 @@ public class BigtableSourceReader implements SourceReader<SeaTunnelRow, Bigtable
     private volatile BigtableSourceSplit currentSplit = null;
 
     /**
+     * Guards compound reader-state transitions so a concurrent {@link #snapshotState(long)} never
+     * observes a half-updated view of {@code pendingSplits} and {@code currentSplit}. The actual
+     * {@code readRows} network scan is intentionally kept outside this lock.
+     */
+    private final Object stateLock = new Object();
+
+    /**
      * Set of field names that map to the Bigtable row key. Populated from {@code rowkey_column}
      * config; falls back to the literal field name {@value ROW_KEY_FIELD} when not configured.
      */
@@ -109,13 +116,23 @@ public class BigtableSourceReader implements SourceReader<SeaTunnelRow, Bigtable
 
     @Override
     public void pollNext(Collector<SeaTunnelRow> output) throws Exception {
-        final BigtableSourceSplit split = pendingSplits.poll();
+        final BigtableSourceSplit split;
+        synchronized (stateLock) {
+            // Poll and assign currentSplit atomically so a concurrent snapshotState() always sees a
+            // consistent (pendingSplits, currentSplit) pair, and the in-flight split can be
+            // re-enqueued on restore.
+            split = pendingSplits.poll();
+            if (Objects.nonNull(split)) {
+                currentSplit = split;
+            }
+        }
         if (Objects.nonNull(split)) {
-            // Assign currentSplit before reading so checkpoints taken during readSplit()
-            // include it and can re-enqueue it on restore.
-            currentSplit = split;
+            // The network scan runs outside stateLock; only output.collect() takes the checkpoint
+            // lock inside readSplit().
             readSplit(split, output);
-            currentSplit = null;
+            synchronized (stateLock) {
+                currentSplit = null;
+            }
         } else if (noMoreSplits && pendingSplits.isEmpty()) {
             log.info("Closed the bounded Bigtable source");
             context.signalNoMoreElement();
@@ -230,18 +247,22 @@ public class BigtableSourceReader implements SourceReader<SeaTunnelRow, Bigtable
 
     @Override
     public List<BigtableSourceSplit> snapshotState(long checkpointId) {
-        List<BigtableSourceSplit> state = new ArrayList<>();
-        // Include the split currently being read so it can be re-enqueued on restore.
-        if (currentSplit != null) {
-            state.add(currentSplit);
+        synchronized (stateLock) {
+            List<BigtableSourceSplit> state = new ArrayList<>();
+            // Include the split currently being read so it can be re-enqueued on restore.
+            if (currentSplit != null) {
+                state.add(currentSplit);
+            }
+            state.addAll(pendingSplits);
+            return state;
         }
-        state.addAll(pendingSplits);
-        return state;
     }
 
     @Override
     public void addSplits(List<BigtableSourceSplit> splits) {
-        pendingSplits.addAll(splits);
+        synchronized (stateLock) {
+            pendingSplits.addAll(splits);
+        }
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java
+++ b/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java
@@ -47,6 +47,13 @@ public class BigtableSourceSplitEnumerator
     private Set<BigtableSourceSplit> pendingSplits;
     private boolean initialized = false;
 
+    /**
+     * Guards the shared assignment state ({@code assignedSplits}, {@code pendingSplits}, {@code
+     * initialized}) against concurrent enumerator callbacks. {@link #initializePendingSplits()} and
+     * {@link #assignSplit(int)} must only run while this lock is held.
+     */
+    private final Object stateLock = new Object();
+
     public BigtableSourceSplitEnumerator(
             Context<BigtableSourceSplit> context, BigtableParameters parameters) {
         this(context, parameters, new HashSet<>());
@@ -87,9 +94,11 @@ public class BigtableSourceSplitEnumerator
     @Override
     public void addSplitsBack(List<BigtableSourceSplit> splits, int subtaskId) {
         if (!splits.isEmpty()) {
-            pendingSplits.addAll(splits);
-            if (context.registeredReaders().contains(subtaskId)) {
-                assignSplit(subtaskId);
+            synchronized (stateLock) {
+                pendingSplits.addAll(splits);
+                if (context.registeredReaders().contains(subtaskId)) {
+                    assignSplit(subtaskId);
+                }
             }
         }
     }
@@ -101,13 +110,17 @@ public class BigtableSourceSplitEnumerator
 
     @Override
     public void registerReader(int subtaskId) {
-        initializePendingSplits();
-        assignSplit(subtaskId);
+        synchronized (stateLock) {
+            initializePendingSplits();
+            assignSplit(subtaskId);
+        }
     }
 
     @Override
     public BigtableSourceState snapshotState(long checkpointId) throws Exception {
-        return new BigtableSourceState(assignedSplits);
+        synchronized (stateLock) {
+            return new BigtableSourceState(assignedSplits);
+        }
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java
+++ b/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java
@@ -74,7 +74,10 @@ public class BigtableSourceSplitEnumerator
             // Restore pending splits first so a returned-but-not-yet-reassigned split survives
             // recovery even when its ID is already present in assignedSplits.
             this.pendingSplits = new HashSet<>(sourceState.getPendingSplits());
-            this.initialized = true;
+            // Only skip table-split discovery when the checkpoint captured real enumerator
+            // progress.
+            // An empty-empty checkpoint (before any reader registered) must still discover splits.
+            this.initialized = !this.assignedSplits.isEmpty() || !this.pendingSplits.isEmpty();
         }
     }
 
--- a/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java
+++ b/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java
@@ -56,29 +56,31 @@ public class BigtableSourceSplitEnumerator
 
     public BigtableSourceSplitEnumerator(
             Context<BigtableSourceSplit> context, BigtableParameters parameters) {
-        this(context, parameters, new HashSet<>());
+        this(context, parameters, null);
     }
 
     public BigtableSourceSplitEnumerator(
             Context<BigtableSourceSplit> context,
             BigtableParameters parameters,
             BigtableSourceState sourceState) {
-        this(context, parameters, sourceState.getAssignedSplits());
-    }
-
-    private BigtableSourceSplitEnumerator(
-            Context<BigtableSourceSplit> context,
-            BigtableParameters parameters,
-            Set<BigtableSourceSplit> assignedSplits) {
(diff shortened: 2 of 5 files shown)


DanielLeens (changes requested):
Thanks for the contribution. I reviewed the latest head from the real Bigtable source runtime path, focusing on the reader / enumerator / checkpoint-recovery chain. The locking direction is reasonable, but I still see one blocker before merge.

- Pain point: concurrent reader / enumerator callbacks can expose an inconsistent Bigtable source view across `pendingSplits`, `currentSplit`, and `assignedSplits`, which makes checkpoint / recovery behavior hard to reason about.
- Approach: add a dedicated `stateLock` on both the reader and enumerator side, and make `BigtableSourceState` defensively copy the assigned split set.
- Simple example: if a checkpoint happens while the reader is moving a split from the pending queue into the in-flight slot, the snapshot should see one consistent state instead of a half-updated combination.

```text
reader side
  -> BigtableSourceReader.pollNext(...) [BigtableSourceReader.java:118-142]
      -> under stateLock: poll pending split + assign currentSplit atomically
      -> outside stateLock: readSplit(...)
      -> under stateLock: clear currentSplit
  -> snapshotState(...) [249-258]
      -> currentSplit + pendingSplits

enumerator side
  -> addSplitsBack(...) [BigtableSourceSplitEnumerator.java:95-103]
      -> returned split goes back into pendingSplits
  -> snapshotState(...) [120-123]
      -> currently persists only assignedSplits
  -> restore
      -> initializePendingSplits() [132-149]
          -> filters newly built splits by assignedSplits
```

## Findings

### 1. A split returned via `addSplitsBack()` can still be lost after enumerator recovery because checkpoint state stores only `assignedSplits`
- Location:
  - `seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java:95-103`
  - `seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java:120-123`
  - `seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java:132-149`
  - `seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceState.java:28-37`
- Why this is a blocker:
  - `addSplitsBack()` correctly requeues returned splits into `pendingSplits`.
  - But `snapshotState()` still serializes only `assignedSplits`.
  - If a checkpoint lands after `addSplitsBack()` and before the split is reassigned, the restored enumerator rebuilds the table split set, sees the split ID already present in `assignedSplits`, and filters it out in `initializePendingSplits()`.
  - That means the returned-but-not-yet-reassigned split disappears after recovery.
- Risk:
  - This is a real correctness issue on the normal recovery path: data can be skipped, not just re-read.
- Suggested fix:
  - Option A: persist `pendingSplits` alongside `assignedSplits` in `BigtableSourceState`, and restore pending first.
  - Option B: redefine the state model so the checkpoint stores the not-yet-finished split set instead of only the historical assigned set.
- Severity: High

### 2. Non-blocking: this source-side state synchronization change still has no focused regression for the returned-split recovery window
- Location:
  - `BigtableSourceReader.java`
  - `BigtableSourceSplitEnumerator.java`
- Why this matters:
  - The PR changes reader/enumerator checkpoint behavior directly, but I do not see a targeted test for:
    - `addSplitsBack()`
    - checkpoint
    - restore
    - reassigning the returned split successfully
- Suggested fix:
  - Add a small focused flow/unit test for that exact returned-split recovery path.
- Severity: Medium

- I do not see a public API or config compatibility problem here.
- The main issue is internal recovery correctness: the new locking makes state transitions cleaner, but the persisted state model is still incomplete for the returned-split path.

## Test coverage / stability
- This PR does not add new UT/E2E code, so I do not see a new flaky-test pattern introduced here.
- The real gap is missing coverage for the checkpoint / restore window above.

## Merge conclusion

1. Blockers
- Finding 1: returned splits can still be dropped after recovery because enumerator checkpoint state persists only `assignedSplits`.

- Finding 2: add a focused regression test for `addSplitsBack() -> checkpoint -> restore -> reassign`.

Overall, the locking and defensive-copy changes are heading in the right direction, and they do improve how the Bigtable source state is updated locally. But the checkpointed state model is still missing the pending-split part of the recovery contract, so I would fix that before merging.

programmerloverun:
@DanielLeens Thanks for the detailed review. I've pushed a fix for the blocker on the latest head (`9b821e3`).

**Finding 1 (blocker) — fixed via Option A**
- `BigtableSourceState` now persists **both** `assignedSplits` and `pendingSplits` (defensive copies).
- `snapshotState()` serializes both sets.
- On restore, `pendingSplits` is rehydrated **before** `initializePendingSplits()` runs, so a split returned via `addSplitsBack()` but not yet reassigned is no longer dropped when its ID is already present in `assignedSplits`.
- `open()` no longer resets restored state (initialization happens in the constructor).

**Finding 2 (non-blocking) — addressed**
- Added `testReturnedSplitSurvivesCheckpointRestoreAndReassign` covering `addSplitsBack()` → `snapshotState()` → restore → `registerReader()` → successful reassign.
- Added `testSnapshotStateDefensivelyCopiesPendingSplits` to verify checkpoint state is isolated from live enumerator mutations.

Verified locally on JDK 11: `spotless:check` passes and all **15** unit tests pass. PTAL when you have a chance.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from the real Bigtable source runtime path again, including the returned-split recovery window, the enumerator restore path, and the new deep tests.

The previous blocker I raised around `addSplitsBack() -> checkpoint -> restore` is fixed on the current head: `pendingSplits` is now checkpointed and restored, and the new recovery tests cover that path.

I did find one new blocker before merge.

## Runtime path rechecked
```text
enumerator restore
  -> BigtableSource.restoreEnumerator(...)
      -> new BigtableSourceSplitEnumerator(..., checkpointState)
          -> constructor reads assignedSplits + pendingSplits

checkpoint state model
  -> BigtableSourceSplitEnumerator.snapshotState(...)
      -> new BigtableSourceState(assignedSplits, pendingSplits)

compatibility boundary
  -> old checkpoints created before this PR
      -> serialized BigtableSourceState only had assignedSplits
      -> after Java deserialization on the new class, pendingSplits is absent / null
      -> constructor now does `new HashSet<>(sourceState.getPendingSplits())`
```

## Finding 1
### Old Bigtable enumerator checkpoints are not backward-compatible with the new `pendingSplits` field
- Location:
  - `seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplitEnumerator.java:72-80`
  - `seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceState.java:26-45`
- Why this is a blocker:
  - `BigtableSourceState` keeps the same `serialVersionUID = 1L`, so old checkpoints can still deserialize into the new class shape.
  - But those old checkpoint payloads never contained `pendingSplits`. After deserialization, that field will be `null`.
  - On restore, the new enumerator constructor immediately executes `new HashSet<>(sourceState.getPendingSplits())`, which will throw before the job can recover.
- Risk:
  - This breaks the normal recovery / upgrade path for jobs that already have a Bigtable source checkpoint from the previous version.
  - In practice that means the new fix improves one recovery window, but it also introduces a different restore-time hard failure for existing checkpoint state.
- Suggested fix:
  - Make restore tolerant to pre-patch state, for example by treating `null` `pendingSplits` as `Collections.emptySet()` and documenting that this is the old checkpoint shape.
  - I would also add one focused regression that deserializes a pre-patch state shape and proves restore still works. The current `BigtableSourceStateRecoveryTest` only round-trips the new two-field state, so it does not cover this compatibility boundary yet.
- Severity: High

## Test / CI note
- The new deep tests are directionally good, and I do not see a flaky pattern in the added UT structure itself.
- The current GitHub `Build` check is red on the latest head, but the compatibility blocker above is already enough to keep this in request-changes state from my side.

## Merge conclusion

1. Blocking items
- Finding 1: make restored enumerator state backward-compatible with checkpoints created before `pendingSplits` existed.

- Add one explicit regression for restoring an old one-field `BigtableSourceState` payload.

Overall, the recovery logic is materially better than the version I reviewed before, and the original returned-split bug is fixed. The remaining gap is the checkpoint compatibility boundary for already-existing state, and I would close that before merging.

programmerloverun:
@DanielLeens Thanks for the second pass — agreed on the compatibility boundary.

Finding 1 (backward-compatible restore) — addressed in 554f0133c:

BigtableSourceState.getPendingSplits() now treats a null field (from pre-#11144 Java deserialization with the same serialVersionUID) as Collections.emptySet(), so restore no longer NPEs.
Added two focused regressions: reflection-simulated null pendingSplits, and real bytes from the pre-#11144 single-field state shape.
All Bigtable connector tests pass locally on JDK 11. Please take another look when you have a moment.

DanielLeens (changes requested):
Thanks for the follow-up. I re-reviewed the latest head `bf4c1e8c513a3250e5185e80d60da969f6d7d07f` from the real Bigtable reader / enumerator / checkpoint / restore path again.

The good news first: the previous Daniel blocker on backward-compatible Bigtable restore is closed on the current Bigtable code. `pendingSplits` is now checkpointed, old one-field checkpoint payloads are tolerated through the null-safe getter, and the new recovery regression tests cover the exact returned-split / legacy-checkpoint windows I had asked for.

Runtime path rechecked:
```text
reader.pollNext(...)
  -> currentSplit / pendingSplits transition under stateLock
enumerator.addSplitsBack(...) [BigtableSourceSplitEnumerator.java:99-108]
  -> returned split goes back into pendingSplits
snapshotState(...) [124-128]
  -> assignedSplits + pendingSplits are both checkpointed
restore constructor [62-81]
  -> restore assignedSplits
  -> restore pendingSplits
  -> keep empty/null legacy checkpoint payloads compatible
```

I do see one blocker on the latest head, and it is not in the Bigtable path itself.

Issue 1: the latest head now also includes an unrelated Pulsar sink option-rule change
- Location: `seatunnel-connectors-v2/connector-pulsar/src/main/java/org/apache/seatunnel/connectors/seatunnel/pulsar/sink/PulsarSinkFactory.java:41-53`
- Why this is a blocker:
  - The Bigtable source recovery path above is self-contained and already reviewable on its own.
  - The new `PulsarSinkFactory` change is a different connector, a different runtime path, and a different rollback / attribution story.
  - Keeping that extra Pulsar scope piggybacked here turns this PR back into a mixed-scope review, which is exactly what we should avoid.
- Suggested fix:
  - Please split the `PulsarSinkFactory` change into a separate PR, or drop it from this branch and keep this PR focused on the Bigtable source-side state fix.
- Severity: High

Test coverage / stability:
- The new Bigtable regression tests are good and focused.
- Stability rating: Stable.
- Reason: the added tests are mostly serialization / mock-stream / in-memory recovery checks; I did not see a new timing-based flaky pattern in the Bigtable test additions.

Merge conclusion:
- Conclusion: can merge after fixes.
- Blocking item:
  - Issue 1: remove or split the unrelated `PulsarSinkFactory` change.
- Non-blocking note:
  - The current `Build` is still in progress, but the mixed-scope blocker above is already enough to keep this in request-changes from my side.

Overall, the Bigtable recovery fix itself looks good on the current head. The blocker is the new unrelated Pulsar scope that arrived afterwards, not the Bigtable logic.

davidzollo:
image

Please synchronize the latest code with the upstream/dev branch!

programmerloverun:
ok












π、爱跑步的程序员~
***@***.***






---Original---
From: "David ***@***.***&gt;
Date: Tue, Jun 23, 2026 22:37 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [apache/seatunnel] [Improve][Connector-V2] Harden Google Bigtable source-side state synchronization (PR #11144)


davidzollo left a comment (apache/seatunnel#11144)
 image.png (view on web) 
Please synchronize the latest code with the upstream/dev branch!

DanielLeens:
Thanks for the update. Once the branch is synced with the latest `upstream/dev` and the new head is pushed, I can take another full pass on the current version.

DanielLeens (approved):
Thanks for the follow-up here. I re-reviewed the latest head locally from the real Bigtable source runtime path again, covering the reader split handoff, the enumerator checkpoint / restore contract, and the new recovery tests. The earlier Daniel blockers are now closed on the current head, and I do not see a new source-level blocker in this revision.

What this PR fixes:
- User pain: the Bigtable source could expose inconsistent split state across `currentSplit`, `pendingSplits`, and `assignedSplits` during checkpoint / failover windows, and returned splits could be lost on restore.
- Fix approach: synchronize the reader / enumerator state transitions, checkpoint both `assignedSplits` and `pendingSplits`, and keep old one-field checkpoint payloads backward-compatible.
- In one sentence: this PR completes the Bigtable source checkpoint-state model so returned splits survive recovery without breaking older checkpoints.

Runtime call chain I checked:

```text
Reader side
  -> BigtableSourceReader.pollNext(...) [BigtableSourceReader.java:118-142]
      -> poll pending split and assign currentSplit under stateLock
      -> run readRows(query) outside the lock
      -> clear currentSplit under the same stateLock
  -> snapshotState(...) [248-258]
      -> capture currentSplit + pendingSplits under the same lock

Enumerator side
  -> addSplitsBack(...) [BigtableSourceSplitEnumerator.java:99-108]
      -> requeue returned splits into pendingSplits
  -> snapshotState(...) [124-128]
      -> persist assignedSplits + pendingSplits
  -> restore constructor [62-80]
      -> restore assignedSplits
      -> restore pendingSplits, or treat legacy null as empty
  -> registerReader(...) [117-122]
      -> initializePendingSplits()
      -> assignSplit(...)
```

Key findings from the current head:
- The old returned-split recovery gap is closed. A split requeued through `addSplitsBack()` now survives checkpoint / restore because `pendingSplits` is part of the persisted enumerator state.
- The backward-compatibility boundary is also closed. `BigtableSourceState.getPendingSplits()` now tolerates old checkpoint payloads where that field was absent and deserialized as `null`.
- The earlier mixed-scope blocker is gone on the latest head. This revision is back to Bigtable-source-only files, so the PR scope is focused again.
- The added regression coverage is pointed at the real failure windows:
  - `BigtableSourceSplitEnumeratorTest.java:55-140` covers returned split -> checkpoint -> restore -> reassign.
  - `BigtableSourceStateRecoveryTest.java:89-458` covers empty checkpoint restore, legacy serialized state restore, reader/enumerator handoff, and concurrent reader snapshot/addSplits interactions.
- I do not see API, config, default-value, protocol, or backward-compatibility regressions on the current implementation.

Test stability:
- Stable.
- The new tests are in-memory / mocked recovery tests, with no fixed sleeps, no external Bigtable dependency, and no timing-sensitive container bootstrap. The concurrent case uses a latch, not hard-coded waits.

Merge conclusion:
- Conclusion: mergeable from Daniel's side.
- Blocking code issues: none in the current revision.
- CI status: the current `Build` is green on the latest head.

Overall, the Bigtable source recovery logic now looks coherent end to end, and the previous Daniel blockers have been addressed on the current revision.

DanielLeens:

- base branch: `dev`
- head SHA: `ea014547f15652ad16507c08824bec485a5b7e14`
- compare status: `diverged`
- ahead_by: `4`
- behind_by: `12`
- mergeable_state: `unknown`

davidzollo (approved):
+1
LGTM

PR #11150, opened 2026-06-21
[Test][E2E] Stabilize Db2 upsert timestamp comparison

DanielLeens:
## What
- order both timestamp queries by `C_INT` in `JdbcDb2UpsertIT`
- keep the before/after timestamp comparison aligned to the same logical rows

## Why
`JdbcDb2UpsertIT` currently compares two query result lists without an `ORDER BY`. When Db2 returns rows in a different physical order between runs, the assertion can compare timestamps from different rows and fail with microsecond-level mismatches unrelated to the actual upsert behavior.

## Validation
- `./mvnw spotless:apply -Dskip.ui=true`
- attempted real E2E validation with:
  `./mvnw -B -T 1 verify -DfailIfNoTests=false -DskipUT=true -DskipIT=false -Dlicense.skipAddThirdParty=true -Dskip.ui=true --no-snapshot-updates -pl :connector-jdbc-e2e-part-1 -am -Pci -Dit.test="JdbcDb2UpsertIT#testDb2UpsertE2e"`
- local Docker Desktop blocked this E2E before the assertion path: `Db2Container` timed out waiting for log output matching `.*Setup has completed\..*` after Db2 finished instance/database creation, so the local run did not reach the updated assertion
- no existing artifacts were intentionally reused; frontend execution in `seatunnel-engine-ui` was skipped by `-Dskip.ui=true` during the Maven run


DanielLeens (commented):
Thanks for the contribution. I reviewed the latest head from the Db2 upsert verification path through the timestamp comparison assertion, and this looks good to me.

## What this PR solves
- User pain: the Db2 upsert E2E compares `C_UPDATED_AT` before and after the update step, but the query result order was not fixed. That means the test could fail even when the data was correct, simply because the same timestamps were returned in a different row order.
- Fix approach: add `ORDER BY C_INT` to both timestamp reads so the before/after comparison is made on a stable logical row order.
- One-line summary: this removes result-order noise without weakening the actual upsert invariant being tested.

```text
Db2 upsert E2E
  -> query updated_at before update
  -> run the job that should keep unchanged rows untouched
  -> query updated_at after update
  -> Assertions.assertIterableEquals(before, after)

without ORDER BY
  -> row order can vary across the two reads
  -> identical timestamps can still be compared against different rows

with ORDER BY C_INT
  -> both reads align on the same logical row order
  -> equality now measures the intended invariant
```

## Key findings
1. The normal test path definitely hits this change because both sides of the before/after timestamp comparison now go through the ordered query.
2. The fix is precise and deterministic: it keeps the same assertion but removes unstable row-order noise.
3. `C_INT` is an appropriate stable ordering key for this test data shape.

## Findings

### Conclusion: can merge
1. Blocking items
- None.

- None.

Overall, this is a targeted stability fix for the Db2 E2E comparison path, and I'm happy to approve it.

davidzollo (approved):
+1

DanielLeens:

PR #11153, opened 2026-06-21
[Test][E2E] Stabilize schema evolution add-column assertions

DanielLeens:
## What does this PR do?

This PR stabilizes the JDBC schema-evolution add-columns checks that are currently blocking unrelated PRs.

The non-exactly-once path was asserting row contents before the sink table metadata had fully caught up after `add_columns`. In CI this shows up as stale sink rows such as `scooter` still being returned where the add-columns case expects the later `hawk9821` data.

This change:
- waits for source and sink column lists to match before asserting add-columns data
- reuses the same add-columns data assertion helper in both exactly-once and non-exactly-once paths
- keeps the scope limited to the shared JDBC schema-evolution test base

## Validation

- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl -am -nsu -Dskip.ui=true spotless:apply`
- `./mvnw -f seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl/pom.xml -nsu -Dskip.ui=true -Dmaven.gitcommitid.skip=true -DskipUT=true -DskipIT=false -DfailIfNoTests=false -Dit.test=DmSchemaChangeIT,SqlServerSchemaChangeIT verify`

Notes:
- `seatunnel-engine-ui` was skipped because this PR does not touch UI modules.
- Real IT validation was attempted twice. The `/Volumes/disk2/...` worktree is on `exfat,noowners`, which makes the MySQL init resource path unusable for Testcontainers. Retrying from a Docker-shared POSIX path reached real container execution, but the local host environment then hit `SeaTunnelContainer.getJobStatus()` `Connection reset` errors for both `DmSchemaChangeIT` and `SqlServerSchemaChangeIT` before the add-columns assertions were reached. This host-specific runtime issue is separate from the CI assertion failures this PR addresses.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
         <module>seatunnel-plugin-discovery</module>
         <module>seatunnel-formats</module>
         <module>seatunnel-engine</module>
+        <module>seatunnel-trace</module>
         <module>seatunnel-examples</module>
         <module>seatunnel-e2e</module>
         <module>seatunnel-shade</module>
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/event/EventType.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/event/EventType.java
@@ -32,4 +32,5 @@ public enum EventType {
     READER_MESSAGE_DELAYED,
     JOB_STATUS,
     SCHEMA_CHANGE_FLUSH,
+    STAIN_TRACE,
 }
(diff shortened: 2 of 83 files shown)


DanielLeens:
Superseded by #11154. #11153 was opened from the wrong base and includes unrelated files.

PR #11154, opened 2026-06-21
[Test][E2E] Stabilize schema evolution add-column assertions

DanielLeens:
## What does this PR do?

This PR stabilizes the JDBC schema-evolution add-columns path that is blocking unrelated PRs.

The latest CI failures show two separate races in the same add-columns case:
- the sink assertions could run before the sink table metadata had fully caught up
- the MySQL CDC source could receive rows with the new columns immediately after `ALTER TABLE`, before the schema-change event had fully propagated

This PR now fixes both sides of that race:
- waits for source and sink column lists to match before asserting add-columns data
- reuses the same add-columns data assertion helper in both exactly-once and non-exactly-once paths
- inserts short `DO SLEEP(5)` gaps in the JDBC `add_columns.sql` fixture after each add-column DDL so the CDC source can observe the schema change before the next new-column DML arrives

## Validation

Executed in this update:
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl -am spotless:apply -DskipTests`
- `./mvnw -o -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl -Dtest=SqlServerSchemaChangeIT -DfailIfNoTests=false test`

Validation result:
- `spotless:apply` completed successfully.
- The targeted `SqlServerSchemaChangeIT` test compiled and started Testcontainers initialization, but local real IT validation is currently blocked by the host Docker environment: `DOCKER_HOST unix:///var/run/docker.sock is not listening` / `Could not find a valid Docker environment`.

Notes:
- No existing build artifacts were reused for runtime validation in this update.
- The full Docker/Testcontainers JDBC schema-evolution path was not revalidated locally because the Docker daemon is unavailable on this host.
- `seatunnel-engine-ui` was not changed; the `spotless:apply` reactor still traversed that module, but this update does not depend on any new UI artifact.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl/src/test/java/org/apache/seatunnel/connectors/jdbc/AbstractSchemaChangeBaseIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl/src/test/java/org/apache/seatunnel/connectors/jdbc/AbstractSchemaChangeBaseIT.java
@@ -309,20 +309,7 @@ public abstract class AbstractSchemaChangeBaseIT extends TestSuiteBase implement
 
         // case1 add columns with cdc data at same time
         sourceDatabase.setTemplateName("add_columns").createAndInitialize();
-        await().atMost(120, TimeUnit.SECONDS)
-                .untilAsserted(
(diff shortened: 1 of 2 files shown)


davidzollo (approved):
+1 if CI passes.
LGTM

PR #11156, opened 2026-06-21
[Fix][Connector-V2] Flush Doris load before schema change

QuakeWang:

Doris sink applied schema changes while the current Stream Load could still contain rows serialized with the old schema. Those rows could then be committed after the target table had already been altered, mixing old-schema and new-schema data in the same load. This is unsafe for positional formats such as CSV and for schema changes such as drop or rename columns.

This PR flushes the in-flight Stream Load before applying schema changes in non-2PC mode, then opens a fresh load for rows written with the new schema.

The 2PC path is intentionally unchanged because Doris 2PC currently uses one transaction and label per checkpoint. Flushing in the middle of a checkpoint would risk an orphan pre-commit transaction or checkpoint label reuse.

Yes.

For Doris sink with `sink.enable-2pc = false`, buffered rows are now flushed before schema changes are applied. This prevents old-schema rows from being committed against the altered table.

The schema evolution documentation now also clarifies the limitation of Doris schema evolution with `sink.enable-2pc = true`.

Added unit tests for:

- non-2PC schema-change ordering: `stopLoad -> apply DDL -> startLoad`
- 2PC behavior: schema changes do not flush in the middle of a checkpoint

Verified with:

```bash
./mvnw spotless:apply
./mvnw -pl seatunnel-connectors-v2/connector-doris -Dtest=DorisSinkWriterTest test
./mvnw -q -DskipTests verify
```

### Check list

### DIFF
--- a/docs/en/introduction/configuration/schema-evolution.md
+++ b/docs/en/introduction/configuration/schema-evolution.md
@@ -254,7 +254,7 @@ sink {
 }
 ```
 
-> **Note (schema evolution + 2PC):** With `sink.enable-2pc = "true"`, a schema change that arrives in the middle of a checkpoint is applied **without** first flushing the in-flight stream load, because 2PC keeps a single transaction per checkpoint. Rows buffered before the DDL are then committed against the altered table within the same transaction. Because `format = "json"` matches columns by name, Doris tolerates this (as in the example above). For positional formats such as CSV, or when you need strict correctness around schema changes, set `sink.enable-2pc = "false"`: the sink then flushes the buffered rows before applying the DDL.
+> **Note (schema evolution + 2PC):** With `sink.enable-2pc = "true"`, Doris schema evolution only supports `format = "json"` because JSON loads match columns by name. Positional formats such as CSV are rejected at runtime for schema evolution with 2PC enabled. Use `format = "json"` or set `sink.enable-2pc = "false"` so the sink can flush buffered rows before applying the DDL.
 
 ### Mysql-CDC -> Jdbc-Postgres
 ```hocon
--- a/docs/en/introduction/configuration/schema-evolution.md
+++ b/docs/en/introduction/configuration/schema-evolution.md
@@ -254,6 +254,8 @@ sink {
 }
 ```
 
+> **Note (schema evolution + 2PC):** With `sink.enable-2pc = "true"`, a schema change that arrives in the middle of a checkpoint is applied **without** first flushing the in-flight stream load, because 2PC keeps a single transaction per checkpoint. Rows buffered before the DDL are then committed against the altered table within the same transaction. Because `format = "json"` matches columns by name, Doris tolerates this (as in the example above). For positional formats such as CSV, or when you need strict correctness around schema changes, set `sink.enable-2pc = "false"`: the sink then flushes the buffered rows before applying the DDL.
+
 ### Mysql-CDC -> Jdbc-Postgres
 ```hocon
 env {
--- a/docs/zh/introduction/configuration/schema-evolution.md
+++ b/docs/zh/introduction/configuration/schema-evolution.md
@@ -255,6 +255,8 @@ sink {
 }
 ```
 
+> **注意（schema 演进 + 2PC）：** 当 `sink.enable-2pc = "true"` 时，发生在 checkpoint 中途的 schema 变更**不会**先 flush 在途的 stream load（2PC 每个 checkpoint 只维护一个事务）。DDL 之前缓冲的数据会在同一个事务里按变更后的表结构提交。由于 `format = "json"` 按列名匹配，Doris 能够容忍（如上例）。对于 CSV 等位置敏感格式，或需要严格保证 schema 变更正确性时，请设置 `sink.enable-2pc = "false"`：此时 sink 会在应用 DDL 之前先 flush 已缓冲的数据。
+
 ### Mysql-CDC -> Jdbc-Postgres
 ```hocon
 env {
--- a/docs/zh/introduction/configuration/schema-evolution.md
+++ b/docs/zh/introduction/configuration/schema-evolution.md
@@ -255,7 +255,7 @@ sink {
 }
 ```
 
-> **注意（schema 演进 + 2PC）：** 当 `sink.enable-2pc = "true"` 时，发生在 checkpoint 中途的 schema 变更**不会**先 flush 在途的 stream load（2PC 每个 checkpoint 只维护一个事务）。DDL 之前缓冲的数据会在同一个事务里按变更后的表结构提交。由于 `format = "json"` 按列名匹配，Doris 能够容忍（如上例）。对于 CSV 等位置敏感格式，或需要严格保证 schema 变更正确性时，请设置 `sink.enable-2pc = "false"`：此时 sink 会在应用 DDL 之前先 flush 已缓冲的数据。
+> **注意（schema 演进 + 2PC）：** 当 `sink.enable-2pc = "true"` 时，Doris schema 演进仅支持 `format = "json"`，因为 JSON load 会按列名匹配。CSV 等位置敏感格式在启用 2PC 的 schema 演进场景下会被运行时拒绝。请使用 `format = "json"`，或设置 `sink.enable-2pc = "false"`，让 sink 可以在应用 DDL 前先 flush 已缓冲的数据。
 
 ### Mysql-CDC -> Jdbc-Postgres
 ```hocon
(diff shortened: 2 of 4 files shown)


DanielLeens (changes requested):
Thanks for working on this! I pulled the latest head `615e18077758aa1257baf47c731049b647774d8d` locally and went through the full diff and the actual sink call chain.

The problem being addressed is real. Before this patch, `DorisSinkWriter.applySchemaChange()` switched the in-memory schema/serializer and then ran the DDL immediately, without closing the in-flight stream load first. On the normal schema-evolution path that means rows already serialized under the old schema can be committed after the table has already been altered. For positional formats such as CSV, that is a real correctness risk.

The non-2PC path is improved in the right way now:

```text
SchemaChangeEvent
  -> SinkFlowLifeCycle.received(...)
      -> DorisSinkWriter.applySchemaChange(event)
          -> flush()                  // only when enable2PC = false
          -> rebuild schema + serializer
          -> schemaChangeManager.applySchemaChange(...)
          -> startLoad(...)
```

I also checked the barrier path:

```text
Barrier
  -> SinkFlowLifeCycle.received(barrier)
      -> writer.prepareCommit(barrierId)
          -> DorisSinkWriter.flush()
      -> writer.snapshotState(barrierId)
```

So for `sink.enable-2pc = false`, the new `flush -> DDL -> reopen` ordering does fix the main issue.

I do still see one blocking issue in the current head:

1. The PR leaves `schema evolution + sink.enable-2pc = true + positional formats` exposed as a normal reachable configuration, but the runtime path is still unsafe there.
   - `DorisSinkWriter.applySchemaChange()` deliberately skips `flush()` when 2PC is enabled (`DorisSinkWriter.java:216-234`).
   - `SeaTunnelRowSerializer` serializes `json` by field name, but `csv` through `TextSerializationSchema`, i.e. positionally (`SeaTunnelRowSerializer.java:95-103`, `138-144`).
   - `DorisSink.supports()` still advertises schema evolution support unconditionally (`DorisSink.java:174-180`).

That means a user can still run into the following normal path:

```text
2PC transaction stays open across the schema change
  -> old-schema rows remain buffered
  -> DDL is applied
  -> new-schema rows are buffered in the same txn
  -> next barrier flushes everything together
```

For JSON this may be tolerated because fields are name-based, but for CSV / other position-sensitive formats it is still not safe. So I would treat this as a blocker.

My suggestion would be:
- Preferred: explicitly reject schema evolution when `sink.enable-2pc = true` and the stream-load format is not `json`, with a clear runtime error and matching docs.
- Longer-term alternative: implement a real transaction-safe schema boundary for 2PC, but that is a much larger change than this PR.

Test stability looks good. The new unit tests are mock-based, deterministic, and use ordering assertions instead of sleeps or timing assumptions, so I would rate them as stable.

One more note on CI: the current red `Build` check does not look caused by this Doris change. The contributor workflow failed in `ConnectorSpecificationCheckTest` because `PulsarSinkFactory` is missing `SinkCommonOptions.MULTI_TABLE_SINK_REPLICA`, which is a separate Pulsar issue outside this diff. After fixing the blocker above, rebasing to the latest `dev` / rerunning after that upstream fix lands should be the lowest-cost path.

DanielLeens (approved):
Thanks for the update. I pulled the latest head `d8e330a938d8ec1825340113d0252c94b5ac28ed` locally and re-reviewed the full current diff from scratch, including the real sink call chain, not just the new delta.

The blocker from my previous round is resolved on this head.

# What this PR fixes
- User pain: before this change, a schema change could be applied while an in-flight Doris stream load still contained rows serialized with the old schema. For positional formats like CSV, that can commit old-shape rows after the table has already been altered.
- Approach: in non-2PC mode, the writer now closes the current load before applying DDL and reopens a fresh load afterward. In 2PC mode, schema evolution is now explicitly limited to `format=json`, and positional formats like CSV are rejected at runtime.
- One-line summary: this revision closes the old-schema/new-schema mixing risk instead of only narrowing it.

# Full runtime path I re-checked
```text
SchemaChangeEvent
  -> SinkFlowLifeCycle.received(...) [seatunnel-engine/.../SinkFlowLifeCycle.java:286-295]
      -> DorisSinkWriter.applySchemaChange(event) [DorisSinkWriter.java:207-235]

Non-2PC path
  -> validateSchemaChangeCompatibility()   // passes through
  -> flush()                               // stopLoad(), commit old buffered rows first
  -> tableSchemaChanger.apply(event)
  -> rebuild serializer from the new schema
  -> schemaChangeManager.applySchemaChange(...)
  -> startLoad(...)                        // new load for post-DDL rows

2PC + json path
  -> validateSchemaChangeCompatibility()
      -> format=json, so allowed
  -> no mid-checkpoint flush
  -> DDL is applied
  -> commit still happens on the barrier path:
      -> prepareCommit()
      -> snapshotState()

2PC + csv path
  -> validateSchemaChangeCompatibility()
      -> throw DorisSchemaChangeException before DDL
```

I also rechecked the serializer side:
```text
SeaTunnelRowSerializer [serialize/SeaTunnelRowSerializer.java:84-96, 128-139]
  -> json uses JsonSerializationSchema (name-based)
  -> csv uses TextSerializationSchema (position-based)
```

So the new restriction is consistent with the real data path, not just with the docs wording.

# Code review
## 1. Core logic
- The normal schema-evolution path definitely hits this change.
- For `sink.enable-2pc = false`, the `flush -> DDL -> reopen` ordering is now correct.
- For `sink.enable-2pc = true`, the previously unsafe positional-format path is no longer left reachable silently; it now fails fast with a clear exception unless the stream load format is JSON.

## 2. Compatibility
From the code path I reviewed, this is compatible in the practical sense:
- no API change
- no config rename/removal
- no default change
- no protocol/serialization format change

What changes is that an unsafe runtime combination (`2PC + schema evolution + non-json format`) is now rejected explicitly instead of being allowed to proceed incorrectly.

## 3. Performance / side effects
- No meaningful CPU or memory concern here.
- Non-2PC adds one load boundary around the schema change event, which is exactly the cost required to preserve correctness.
- I did not find a new concurrency or resource-release issue on the current head.

## 4. Error handling / logging
- Exception handling is slightly better now because the original cause from `schemaChangeManager.applySchemaChange(...)` is preserved.
- I did not find a new blocker in the current version.

# Tests and stability
- `DorisSinkWriterTest.java:210-327` now covers:
  - non-2PC ordering
  - 2PC no-mid-checkpoint-flush behavior
  - rejection of `2PC + csv`
- These tests are mock-based, deterministic, and do not depend on timing, threads, ports, or external services.

Stability rating: **stable**

# CI note
The current red `Build` still does **not** look caused by this Doris change.

From the failing workflow run on the contributor fork (`27908949378`), the actual failure is still:
- `org.apache.seatunnel.api.connector.ConnectorSpecificationCheckTest`
- `PulsarSinkFactory` missing `SinkCommonOptions.MULTI_TABLE_SINK_REPLICA`

So this is the same unrelated Pulsar-side failure pattern, not a regression introduced by this PR.

# Merge conclusion
### Conclusion: can merge

1. Blockers
- None from the current Doris code path.

2. Non-blocking follow-up
- The red `Build` should still be rerun or rebased after the upstream Pulsar-side specification failure is fixed, but I would not treat that as a Doris-code blocker for this PR.

Overall, the current revision resolves the earlier blocker cleanly, and I do not see a new merge-blocking issue on the latest head.

DanielLeens (commented):

# What this PR fixes
- PR: [Fix][Connector-V2] Flush Doris load before schema change

```text
Doris sink schema-change path
  -> rows are buffered through Doris load
  -> schema change arrives
  -> writer flushes pending load before applying the schema transition
```

# Current review findings
- Changed files in the current full diff: 4.
- Base sync fact recorded: compare status `ahead`, `behind_by=0`. This is recorded as queue metadata; I am not turning it into a standalone sync request unless it is needed to interpret a failing CI signal.

- Fetched the PR head into a local review branch named `seatunnel-review-11156`.

# Merge conclusion

DanielLeens (approved):
Rechecked the unchanged head `a1f4f410cd` after the CI gate finished.

The previous Daniel follow-up on this head only kept the PR in a waiting-for-CI state. `Build` is now `SUCCESS`, and I do not see a reopened source-level blocker on the current revision.

Technical conclusion from Daniel's side: approved on the current head.

I also rechecked the live base/head metadata while posting this follow-up:
- base: `dev`
- compare status: `ahead`
- behind_by: `0`

If GitHub still keeps a separate repository-level review gate, that is outside this source recheck.

davidzollo (approved):
+1
LGTM

PR #11121, opened 2026-06-18
[Fix][API] Fix validateSingleChoice and extension exceptions bypassing error aggregation

nzw921rx:
### Purpose of this pull request

Related: https://github.com/apache/seatunnel/issues/11007

Fix `ConfigValidator` breaking its error-aggregation contract in two places:
1. `validateSingleChoice()` previously threw `OptionValidationException` directly, aborting collection of remaining errors.
2. `ConditionExtension` implementations that threw `OptionValidationException` caused fail-fast, preventing subsequent validation errors from being reported.

Additionally refactors error message formatting to eliminate scattered `String.format` calls with raw type literals.

**Changes**:
- `ConfigValidator.validateSingleChoice()` — replace `throw` with `errors.add()`, accepting a `List<String> errors` parameter so single-choice violations participate in error aggregation
- `ConfigValidator.collectErrors()` — wrap constraint evaluation in try/catch to aggregate `OptionValidationException` from extensions; pass `errors` list into `validateSingleChoice()` calls
- `ConfigValidator` — introduce private `TYPE_*` constants and delegate formatting to `OptionUtil.formatError()` / `formatOptionsError()`, centralizing the error template
- `OptionUtil` — add `formatError()` / `formatOptionsError()` helper methods with a single `ERROR_TEMPLATE`
- `ConditionExtension` javadoc — update to reflect that thrown exceptions are now caught and aggregated (no longer fail-fast)
- `ConfigValidatorTest` — add tests for single-choice error aggregation and extension exception aggregation

### Does this PR introduce _any_ user-facing change?

Yes — multiple config validation errors are now reported together instead of stopping at the first single-choice or extension violation. Users see a complete error report on job submission.

### How was this patch tested?

```bash
./mvnw test -pl seatunnel-api -Dtest="ConfigValidatorTest" -DfailIfNoTests=false
```

PR #11158, opened 2026-06-22
[Improve][Connector-V2] Migrate RocketMQ Source validation to declarative OptionRule

nzw921rx:

Related #11007
Depends on #11121

1. Add `greaterOrEqual(START_MODE_TIMESTAMP, 0L)` constraint when `start.mode = CONSUME_FROM_TIMESTAMP`
2. Add `mapNotEmpty(START_MODE_OFFSETS)` constraint when `start.mode = CONSUME_FROM_SPECIFIC_OFFSETS`
3. Add `TableConfigsValidator` (`ConditionExtension`) to validate nested `tables_configs` / `table_list` entries: topics non-empty, per-entry start.mode conditional dependencies, timestamp >= 0, offsets non-empty
4. Remove migrated imperative checks from `RocketMqSourceConfig`, retain runtime-only `> currentTimestamp` checks with comments
5. Add `RocketMqFactoryTest` covering single-table and multi-table positive/negative validation paths

Yes — previously invalid configurations (negative timestamps, empty offsets, missing topics in tables_configs entries) were rejected at task execution time with generic `IllegalArgumentException`. Now they are rejected earlier at job submission time with structured `OptionValidationException` including option name, constraint type, and violation details. Multiple errors are reported together thanks to the aggregation support from #11121.

## How was this patch tested?

Unit tests in `RocketMqFactoryTest`:

- `testValidTimestampConfig` — valid CONSUME_FROM_TIMESTAMP mode passes validation
- `testNegativeTimestampRejected` — negative `start.mode.timestamp` is rejected
- `testValidSpecificOffsetsConfig` — valid CONSUME_FROM_SPECIFIC_OFFSETS mode passes validation
- `testEmptyOffsetsMapRejected` — empty offsets map is rejected
- `testMultiTableValidConfig` — valid multi-table config passes validation
- `testMultiTableMissingTopicsRejected` — empty topics in tables_configs entry is rejected
- `testMultiTableTimestampModeWithoutTimestampRejected` — TIMESTAMP mode without timestamp in entry is rejected
- `testMultiTableNegativeTimestampRejected` — negative timestamp in tables_configs entry is rejected
- `testMultiTableEmptyOffsetsRejected` — empty offsets in tables_configs entry is rejected

Existing tests continue to pass.

```bash
./mvnw test -pl seatunnel-connectors-v2/connector-rocketmq -Dtest="RocketMqFactoryTest" -DfailIfNoTests=false
```

### DIFF
--- a/seatunnel-connectors-v2/connector-rocketmq/src/main/java/org/apache/seatunnel/connectors/seatunnel/rocketmq/source/RocketMqSourceConfig.java
+++ b/seatunnel-connectors-v2/connector-rocketmq/src/main/java/org/apache/seatunnel/connectors/seatunnel/rocketmq/source/RocketMqSourceConfig.java
@@ -103,11 +103,6 @@ public class RocketMqSourceConfig implements Serializable {
 
     private void parseTableConfig(ReadonlyConfig tableConfig, List<String> allTopics) {
         String topicsStr = tableConfig.get(RocketMqSourceOptions.TOPICS);
-        if (topicsStr == null || topicsStr.trim().isEmpty()) {
-            throw new IllegalArgumentException(
-                    "'topics' must be configured in each tables_configs entry, but got: "
-                            + tableConfig);
-        }
         List<String> topics =
                 Arrays.stream(topicsStr.split(RocketMqSourceOptions.DEFAULT_FIELD_DELIMITER))
                         .map(String::trim)
@@ -135,30 +130,16 @@ public class RocketMqSourceConfig implements Serializable {
             switch (startMode) {
                 case CONSUME_FROM_TIMESTAMP:
                     startTimestamp = tableConfig.get(RocketMqSourceOptions.START_MODE_TIMESTAMP);
-                    if (startTimestamp == null) {
-                        throw new IllegalArgumentException(
-                                "When 'start.mode' is set to 'CONSUME_FROM_TIMESTAMP' in tables_configs, "
-                                        + "'start.mode.timestamp' must also be specified in the same table config entry. "
-                                        + "Topics: "
-                                        + topicsStr);
-                    }
+                    // Runtime check: cannot be declarative (depends on current time)
                     long currentTimestamp = System.currentTimeMillis();
-                    if (startTimestamp < 0 || startTimestamp > currentTimestamp) {
+                    if (startTimestamp > currentTimestamp) {
                         throw new IllegalArgumentException(
-                                "The offsets timestamp value is smaller than 0 or larger"
-                                        + " than the current time");
+                                "start.mode.timestamp must not be greater than the current time");
                     }
                     break;
                 case CONSUME_FROM_SPECIFIC_OFFSETS:
                     Map<String, Long> offsetConfigMap =
                             tableConfig.get(RocketMqSourceOptions.START_MODE_OFFSETS);
-                    if (offsetConfigMap == null || offsetConfigMap.isEmpty()) {
-                        throw new IllegalArgumentException(
-                                "When 'start.mode' is set to 'CONSUME_FROM_SPECIFIC_OFFSETS' in tables_configs, "
-                                        + "'start.mode.offsets' must also be specified in the same table config entry. "
-                                        + "Topics: "
-                                        + topicsStr);
-                    }
                     Map<MessageQueue, Long> specificOffsets = metadata.getSpecificStartOffsets();
                     if (specificOffsets == null) {
                         specificOffsets = new HashMap<>();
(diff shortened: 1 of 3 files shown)


DanielLeens (approved):
Thanks for the migration work here. I re-checked the current head from the declarative option rule down to the runtime table-config parser, and the logic looks aligned to me.

What this PR fixes
- User pain: RocketMQ source validation should live in declarative option rules instead of being scattered imperatively, especially for timestamp / specific-offset start modes.
- Fix approach: The PR moves the main start-mode and ACL checks into factory-side conditional rules plus a nested `TableConfigsValidator`.
- One-line summary: On the current head, the declarative validation path is aligned with the runtime parser.

Runtime path I traced
```text
Job submission
  -> `RocketMqSourceFactory.optionRule()`
      -> top-level conditional rules for timestamp / offsets
      -> nested `TableConfigsValidator` for per-table configs

Runtime startup
  -> `RocketMqSourceConfig.parseTableConfig()` reads the same fields directly

Result
  -> submission-time validation now matches the runtime parser contract
```

Key findings
- The normal submission path hits this change directly.
- Unlike the Kafka PR, the nested validator here does require missing timestamp / offsets fields when the mode demands them.
- The declarative rules now match what runtime parsing expects.

Blocking items

Other review notes

- Compatibility: Fully backward compatible. This only makes invalid configs fail earlier.
- Side effects: No meaningful performance or resource-model downside; the main effect is earlier rejection of bad configs.
- Error handling / logging: Aggregated submission-time validation is better than failing later with a runtime `IllegalArgumentException`.
- Tests: The added tests cover the nested required-field cases and do not show obvious flaky-test patterns.
- Docs: No docs update is required.
- CI: There is still a red `rocketmq-connector-it` check on the branch, but the failed log points to an engine-side `CheckpointCoordinator.isNoErrorCompleted(...)` NPE rather than this factory/config diff. I would sync the latest `dev` and rerun the checks.

Merge verdict
- From the current head I do not see a source-level blocker. The declarative validation path looks aligned with the runtime parsing path.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the declarative option-rule path down to the runtime RocketMQ source config path.

# What this PR solves
- User pain: invalid RocketMQ source configs were still being rejected late at runtime instead of earlier during submission.
- Fix approach: move the main validation rules into `RocketMqSourceFactory.optionRule()` and add a nested validator for multi-table entries.
- One-line summary: on the current head, the submission-time validation now matches the runtime parsing contract closely enough for the normal source path.

# Runtime path I checked
```text
Job submission
  -> RocketMqSourceFactory.optionRule()
      -> conditional rules for timestamp / specific offsets / ACL
      -> TableConfigsValidator for tables_configs and table_list

Job startup
  -> RocketMqSourceConfig(...)
      -> parseTableConfig(...)
      -> buildConsumerMetadata(...)
      -> keep only the time-dependent "must not be greater than current time" checks at runtime
```

# Findings
- The normal submission path definitely hits this change.
- The nested validator now covers the required multi-table cases that runtime parsing depends on: non-empty `topics`, required `start.mode.timestamp` for `CONSUME_FROM_TIMESTAMP`, and non-empty `start.mode.offsets` for `CONSUME_FROM_SPECIFIC_OFFSETS`.
- Keeping the future-timestamp guard as a runtime check is reasonable, because that part depends on the current wall-clock time.
- I do not see a new correctness, compatibility, or resource-management blocker on the current head.

- No API, config-name, default-value, protocol, or serialization change.
- The main behavior change is earlier rejection of bad configs, which is the intended improvement.
- Runtime overhead is negligible because the extra work happens during config validation, not in the hot read path.

# Tests / stability / CI
- The new unit tests cover the important nested validation cases, including missing timestamp, negative timestamp, and empty offsets in multi-table mode.
- These tests are stable: pure config-validation unit tests with no timing, network, port, or container dependency.
- The red RocketMQ integration lane does not appear to come from this diff; the failure shown is in an engine-side checkpoint path rather than the option-rule migration itself.

# Merge conclusion
### Conclusion: can merge

1. Blocking items
- None on the current head.

- None from the path I reviewed.

Overall, this is a clean validation migration and the current head looks good to merge from the real source-config path.

davidzollo (approved):
LGTM. Thanks for continuing this validation migration work.

This is the kind of focused improvement that makes connector behavior more predictable for users and simpler for maintainers to evolve later. Keeping the RocketMQ source validation aligned with the declarative rule path is a nice cleanup, and the scope of the change is easy to review.

Thanks again for the contribution, and please keep sending more patches.

PR #11159, opened 2026-06-22
[Test][E2E] Bound Pulsar sink IT receive wait

DanielLeens:
### What does this PR do?

Bound `PulsarSinkIT` consumer reads by a total wait window and poll timeout so the test fails fast instead of hanging the whole matrix job.

### Why is the change needed?

PR #11136 surfaced a cancelled `updated-modules-integration-test-part-1` job that is unrelated to its functional change. The actionable failure comes from `PulsarSinkIT`, which currently blocks forever on `consumer.receive()` when the expected 10 sink records do not arrive in time.

### How does it work?

- use a bounded receive loop driven by the expected record count instead of a fixed number of blocking calls
- poll Pulsar with a timeout so the test never leaves CI stuck in an infinite wait
- keep the assertion on 10 output records and log a clear warning when the timeout is exhausted

### How was this PR tested?

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-pulsar-e2e -DskipTests test-compile`
- attempted real runtime validation: `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-pulsar-e2e -Dtest=PulsarSinkIT -Dsurefire.failIfNoSpecifiedTests=false test`
  - the Testcontainers run started, but local Docker Desktop crashed `apachepulsar/pulsar:2.3.1` with a container-side JDK 8 `SIGSEGV` before the SeaTunnel assertions completed
- not revalidated in this run: `seatunnel-engine-ui` (unchanged and outside the affected module path)

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-pulsar-e2e/src/test/java/org/apache/seatunnel/e2e/connector/pulsar/PulsarSinkIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-pulsar-e2e/src/test/java/org/apache/seatunnel/e2e/connector/pulsar/PulsarSinkIT.java
@@ -20,15 +20,15 @@ package org.apache.seatunnel.e2e.connector.pulsar;
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.node.ObjectNode;
 
-import org.apache.seatunnel.api.table.catalog.TablePath;
 import org.apache.seatunnel.e2e.common.TestResource;
 import org.apache.seatunnel.e2e.common.TestSuiteBase;
 import org.apache.seatunnel.e2e.common.container.TestContainer;
 
+import org.apache.pulsar.client.api.Consumer;
 import org.apache.pulsar.client.api.Message;
-import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.PulsarClient;
-import org.apache.pulsar.client.api.Reader;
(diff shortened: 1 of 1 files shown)


DanielLeens (commented):
Thanks for the contribution. I reviewed the latest head from the Pulsar sink E2E observation path through the consumer-subscription timing and receive loop, and this looks good to me.

## What this PR solves
- User pain: the old test created the consumer after the batch job finished and then relied on ten long blocking `receive(30s)` calls. That made the test sensitive to consumer start timing and slow delivery windows in CI.
- Fix approach: subscribe before the job starts, then consume with a total timeout plus short bounded polls, while still asserting that all expected visible records are actually received.
- One-line summary: this replaces a timing-sensitive observation window with a stable subscribe-first, short-poll, bounded-budget pattern.

```text
Pulsar sink E2E
  -> create Pulsar consumer before job start
  -> run fake_to_pulsar.conf
  -> poll the sink topic in short bounded intervals
      -> collect actually visible committed records
  -> assert that all 10 expected records arrive within the total timeout
  -> validate the JSON payload fields on the first record
```

## Key findings
1. The normal test path definitely hits this change because `testSinkPulsar` now uses the new pre-subscription and bounded receive loop directly.
2. The assertion is not weakened: the test still requires all expected records, but it now measures them through observable committed records instead of a brittle long-blocking receive pattern.
3. Resource handling is cleaner now that the client and consumer are both managed through try-with-resources.

## Findings

### Conclusion: can merge
1. Blocking items
- None.

- None.

Overall, this is a focused stability fix for the Pulsar sink E2E observation path, and I'm happy to approve it.

davidzollo (approved):
+1

PR #11161, opened 2026-06-22
[Test][E2E] Stabilize DB2 upsert timestamp ordering

davidzollo:
## Why
- CI for PR #11137 is failing in `JdbcDb2UpsertIT.testDb2UpsertE2e`
- the assertion reads `C_UPDATED_AT` without a deterministic row order, so the expected timestamp sequence can be compared against the wrong row order

## What
- add `ORDER BY C_INT` to the DB2 verification queries in `JdbcDb2UpsertIT`
- keep the original timestamp assertions unchanged and only stabilize the result ordering by the integer primary key

## Validation
- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-common,seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-1 -DskipTests test-compile`
- attempted real E2E path with `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-common,seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-1 -Dtest=JdbcDb2UpsertIT -Dsurefire.failIfNoSpecifiedTests=false test`

## Validation notes
- `seatunnel-engine-ui` was not built because this change only touches JDBC E2E tests
- the local E2E reached the real DB2 container startup and test bootstrap, but the run is still blocked by missing prebuilt starter jars in the local environment, so full end-to-end completion could not be finished on this machine

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
         <module>seatunnel-plugin-discovery</module>
         <module>seatunnel-formats</module>
         <module>seatunnel-engine</module>
+        <module>seatunnel-trace</module>
         <module>seatunnel-examples</module>
         <module>seatunnel-e2e</module>
         <module>seatunnel-shade</module>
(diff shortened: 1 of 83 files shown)


davidzollo:
Closing this accidental fork-base PR. The DB2 ordering fix is already present in current apache/dev, and PR #11137 has been updated to latest dev so its rerun can pick up that baseline fix.

PR #11181, opened 2026-06-24
[Fix][E2E] Align SAP HANA sink schema in JDBC IT

DanielLeens:
## What does this PR do?

Fixes a shared SAP HANA JDBC E2E failure that appears in unrelated PR CI runs.

SAP HANA uses `TEST` as the target schema in this test setup, but the sink configs only passed `database = "TEST"`. That left generated sink SQL and the IT cleanup path without an explicit schema, which led to repeated `invalid table name: ALLDATATYPES_SINK` failures.

This patch aligns the HANA IT baseline by:
- setting the HANA schema on `JdbcHanaIT`
- passing `schema = "TEST"` in both HANA sink configs that rely on generated sink SQL

## How was this validated?

- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-6 spotless:apply -DskipTests`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-6 -DskipTests -Dskip.ui=true verify`
- attempted closer runtime validation with:
  - `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-6 -Dskip.ui=true -Dtest=JdbcHanaIT#testJdbcDb -DfailIfNoTests=false test`

`seatunnel-engine-ui` was skipped because this fix only touches the JDBC part-6 E2E module. The HANA IT start was blocked locally by an unresponsive Docker socket: Testcontainers loaded `UnixSocketClientProviderStrategy` from `~/.testcontainers.properties`, and direct `/_ping` against `~/.docker/run/docker.sock` timed out.


DanielLeens (commented):
Thanks for the contribution. I reviewed the latest head from the HANA JDBC E2E case construction down to the sink-side table qualification path, and this looks good to me.

## What this PR solves
- User pain: the HANA JDBC E2E was mixing test-side schema qualification with connector-side database-only configuration, which can lead to the sink path trying to build a duplicated `TEST.TEST.table` identifier instead of validating the real HANA sink flow.
- Fix approach: keep the test-side schema explicitly on the `JdbcCase`, while still leaving the connector config in database-only mode so comparison, cleanup, and sink SQL all stay aligned on the same schema semantics.
- One-line summary: this realigns the HANA E2E's schema boundary instead of papering over the symptom in one assertion.

```text
HANA JDBC E2E
  -> JdbcHanaIT builds JdbcCase
      -> keep schema(TEST) for test-side table qualification and cleanup
  -> connector config remains database-only
      -> sink SQL generation avoids TEST.TEST.table
  -> source query / sink verification / cleanup all resolve against the same schema semantics
```

## Key findings
1. The normal HANA E2E path definitely hits this change because the `JdbcCase` drives the table qualification used throughout the test.
2. The fix is precise: it restores the schema boundary on the test side without changing the production JDBC sink implementation.
3. The added comment clearly explains why the test keeps `schema(TEST)` while the connector config should not produce a double-qualified identifier.

## Findings

### Conclusion: can merge
1. Blocking items
- None.

- None.

Overall, this is a targeted E2E alignment fix for the HANA schema boundary, and I'm happy to approve it.

davidzollo (approved):
+1

DanielLeens:

PR #10763, opened 2026-04-14
[Feature][Zeta] Implement proper dry-run mode with progressive validation layer0

junjunclub:
### Purpose of this pull request

This pull request implements **Layer 0 (Static Analysis)** for the `--dry-run` feature, addressing the core requirements discussed in #[10681](https://github.com/apache/seatunnel/issues/10681).

It enhances `SeaTunnelConfValidateCommand` and `MultipleTableJobConfigParser` to perform a strict static analysis of the provided job configuration without triggering any side-effects (e.g., `SaveMode` table creation, drops, or actual data fetching operations are safely bypassed by the new `dryRun` flag).

Specifically, this Layer 0 dry-run validation ensures:
1. **Config Syntax Validation:** Validates HOCON and YAML structure syntaxes.
2. **Unknown Key Checking (Typo Detection):** Strictly validates user-provided config keys against the specific connector's `OptionRule`, throwing errors for any unseen keys to prevent typos. Global framework options (e.g., `plugin_name`, `parallelism`, `schema`, `dag-parsing.mode`) have been programmatically resolved using base `Option` objects to prevent false positives.
3. **Required Options & Type Checking:** Ensures all mandatory configurations are present and safely castable to the correct types.
4. **Plugin Loadability:** Verifies that all declared Source, Transform, and Sink plugins exist and can be loaded by the classloader.
5. **DAG Topology Verification:** Validates that the routing graph is correctly built without isolated endpoints, orphan inputs, or missing outputs.
6. **Transform SQL Checking:** Verifies logical SQL transform syntaxes.

### Does this PR introduce _any_ user-facing change?

**Yes.** 
Users can now use the `--dry-run static` (or `--check`) arguments to meticulously validate their job configuration files offline before actually running them.

**Example Console Usage:**
```bash
sh bin/seatunnel.sh -c config/v2.batch.config.template --dry-run static
```
- **If Valid:** Exits gracefully with code `0`.
- **If Invalid:** Fast-fails, prints the exact configuration typo, misconfigured option, or DAG error, and exits with a non-zero code.

### How was this patch tested?

Added comprehensive unit test coverage in `org.apache.seatunnel.core.starter.seatunnel.command.SeaTunnelConfValidateCommandTest` covering the following positive and negative testing scenarios:

- `testValidStaticDryRun`: Verifies execution passes safely for a correct DAG.
- `testInvalidHoconSyntax` & `testInvalidYamlSyntax`: Ensures the config builder catches malformed files.
[...]

Issue #10681, opened 2026-03-31
[Feature][Zeta] Implement proper dry-run mode with progressive validation layers


davidzollo:
## Motivation

SeaTunnel currently lacks a meaningful dry-run capability. The existing `--check` flag (`SeaTunnelConfValidateCommand`) is effectively a stub — its `execute()` body contains only a `// TODO: validate config using new api` comment and performs no real validation beyond checking that the config file path exists.

This means users must submit jobs to a cluster (or run them fully in local mode) just to discover problems like:
- Typos in option names
- Missing required fields
- Wrong types for option values
- Unreachable source/sink systems
- Schema mismatches between source output and sink expectation
- Broken transform SQL expressions

These failures account for the majority of user-filed bugs and support requests, yet they could all be caught **before** a single byte of data is read or written.

## Proposed Solution

Introduce a proper `--dry-run` mode with **four progressive validation layers**, each independently selectable via `--dry-run=<level>`:

---

### Layer 0: Static Analysis (no network, no I/O)

Triggered by: `--dry-run=static`

- Config file syntax parsing (HOCON/YAML validity)
- All `Option` keys validated against the connector's declared option set (catches typos, unknown fields)
- Required options presence check
- Option value type validation (e.g., string value in an integer field)
- Plugin class loadability check (connector JAR exists and class is loadable)
- DAG topology validation (at least one Source, at least one Sink, no cycles)
- Transform SQL syntax parsing (parse-only, no execution)

**Cost:** milliseconds, zero external dependencies.

---

### Layer 1: Connectivity Check (connect, no data)

Triggered by: `--dry-run=connect`

- Establishes connections to source and sink systems
- Validates credentials and permissions
- Verifies source tables/topics/paths exist
- Verifies sink write permissions and target existence (or createability)
- Infers source schema
- Checks source schema compatibility with sink schema (field names, type mapping)

**Cost:** seconds, requires live systems. Catches ~80% of real-world job failures before execution.

---

### Layer 2: Data Sampling (read N rows, sink to memory only)

Triggered by: `--dry-run=sample[=N]` (default N=100)

- Reads up to N rows from the source
- Passes data through the full transform chain
- Validates transform logic on real data (NULL handling, type casts, SQL semantics)
- Writes output to memory/Console — **never to the real sink**
- For transactional sinks (JDBC, Kafka), opens a transaction/producer but always rolls back / does not commit
- Prints a sample of the output rows for human review

[...]

PR #10583, opened 2026-03-09
[Docs][Core] Add Javadoc to MultiTableWriterRunnable

LiJie20190102:
### Purpose of this pull request

Implement issue #10538 

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config)

PR #11186, opened 2026-06-25
[Feature][Core] Add connect dry-run validation1

junjunclub:

This PR adds a `connect` mode (Layer 1) to the `--dry-run` option of the SeaTunnel (Zeta) CLI, following up on the merged Layer 0 work (#10763) and implementing the Layer 1 design described in #10681:

```shell
sh bin/seatunnel.sh --config <config> --dry-run connect
```

It is opened as a **draft / RFC**. The implementation direction (a connector-level `validateConnection`-style SPI) is the one proposed by the maintainer in #10681 ("Layer 1 requires a `validateConnection()` SPI method"). What is **not** yet decided anywhere is the **connector rollout scope** — that is the main thing I'd like to agree on before connectors start implementing the hooks. Please see "Open questions" below.

#### Relationship to #10681 / #10763

- #10681 (feature design) defines Layer 1 = connectivity + schema inference + source/sink schema compatibility, and explicitly lists `All connectors (Layer 1 requires a validateConnection() SPI method)` under *Affected Modules*.
- #10763 merged Layer 0 (`static`). Reviewers asked to keep future modes (`connect/sample/shadow`) **out of the public path until runtime + docs + tests are ready together**, and to keep PR scope tight. This PR tries to honor both.

| Layer | Mode | What it checks | Status |
|-------|------|----------------|--------|
| 0 | `static` | Offline config / plugin / DAG validation, no network required | merged (#10763) |
| 1 | `connect` | Schema inference + connectivity / schema-compatibility validation, **no data movement** | **this PR** |
| 2 | `sample` | Read a small sample through source/transform without writing | later |
| 3 | `shadow` | Non-destructive shadow run | later |

#### What this PR implements (Layer 1 = `connect`)

`connect` walks the whole source → transform → sink DAG and validates it **without creating any runtime objects**. It does **not** create source readers, sink writers, committers, or save-mode handlers; it does not submit a job, read source records, execute save-mode logic, or write target data.

**1. New opt-in factory SPI hooks** (all `default`, so fully backward compatible — this is the `validateConnection()` SPI from #10681):

- `TableSourceFactory#inferSchemaForDryRun(context)` — infer source schemas without creating readers (default delegates to `discoverTableSchemas`).
- `TableSourceFactory#validateConnectionForDryRun(context, catalogTables)` — metadata-level check: credentials, permissions, source table/topic/path existence (default no-op).
- `TableSinkFactory#validateConnectionForDryRun(context)` — sink connectivity + schema compatibility, target existence/createability (default no-op).

**2. `DryRunConnectValidator`** (new) orchestrates the DAG:
- Source: discover factory → infer schema → validate connection; propagate inferred schemas downstream.
- Transform: resolve `plugin_input`/`plugin_output` wiring (dependency-ordered), build the transform to compute produced schemas, and check input schema-type consistency.
- Sink: resolve input vertices (incl. multi-table rules) → `validateConnectionForDryRun`.
- Reuses the same connector-jar resolution / child-first classloader split (`JobPluginClasspathHelper`) introduced in #10763, so dry-run and runtime stay aligned.

**3.** `FakeSourceFactory` implements `inferSchemaForDryRun` as a reference example. CLI routing updated so `--dry-run connect` is accepted; `sample`/`shadow` still error as not implemented (keeping the #10763 discipline of not exposing unfinished modes).

#### Scope: connectors are intentionally NOT in this PR

Today only `FakeSource` implements a hook, so for most connectors `validateConnectionForDryRun` is a no-op and `connect` effectively performs **schema inference + DAG wiring validation** only. Implementing real connectivity checks for the ~57 source / ~64 sink connectors is **explicitly out of scope here** and is meant to land as focused follow-up PRs (one connector or a small batch at a time), so this PR stays reviewable.

#### Open questions for maintainers (please weigh in)

1. **SPI shape**: Do the proposed factory hooks (`inferSchemaForDryRun`, `validateConnectionForDryRun`) match the `validateConnection()` SPI intent from #10681? Any preferred name / placement (factory vs source/sink instance vs reuse of the existing `Catalog` SPI)?
2. **Public API surface**: Is adding 3 new `default` methods to `TableSourceFactory` / `TableSinkFactory` acceptable, given the cross-module-scope concern raised on #10763?
3. **Rollout / pilot connector**: Should this PR stay framework-only (+ `FakeSource` reference) and add connectors in follow-up PRs? If yes, which connector do you want as the first real implementation (I'd suggest **JDBC**, since it already has a `Catalog`)?
4. **Unsupported connectors**: When a connector does not implement the hook, should `connect` stay a silent no-op (current), or explicitly report `SKIPPED / not verified` per connector so users aren't misled?
5. **Boundary**: Confirm the hard boundary — no runtime instances, no save-mode execution, no DDL/DML, no data read/write.
6. **Command / contract**: `connect` runs as a validate command (like `static`) and stays CLI-only (REST keeps rejecting dry-run, consistent with #10763). Should `connect` always run `static` first (cumulative)?

Yes.
- `-d` / `--dry-run` now accepts `connect` in addition to `static`. Existing `static` / `--check` behavior is unchanged.
- New public default methods on `TableSourceFactory` / `TableSinkFactory` (backward compatible — connectors are not required to implement them).
- Docs updated: `docs/en/engines/zeta/user-command.md`, `docs/zh/engines/zeta/user-command.md`.
- No change to existing job execution behavior.

- Unit tests for CLI parsing of `--dry-run connect` (`ClientCommandArgsTest`).
- Unit tests for the connect validation flow in `SeaTunnelConfValidateCommandTest` (schema inference + DAG wiring; positive/negative cases) using `FakeSource`.
- `<!-- TODO before un-drafting: run and confirm -->` `./mvnw spotless:apply` and `./mvnw -q -DskipTests verify` pass locally.

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice — N/A.
* [x] Documentation updated for the new feature (`docs/en` + `docs/zh`).
* [ ] `incompatible-changes.md` — N/A (backward compatible default methods).
* [ ] Connector code checklist — N/A (no new connector; connector hook implementations are follow-up PRs).


DanielLeens (changes requested):
Thanks for working on this! I reviewed the full current diff, traced the actual `--dry-run connect` call chain, and also searched the current head for real connector-side hook implementations. The CLI/framework plumbing is moving in a useful direction, but I do not think the feature is ready to merge yet in its current form.

# What this PR solves
- User pain: today `--check` / `--dry-run static` can validate config syntax and DAG wiring, but cannot tell users whether source/sink connectivity and schema compatibility will hold before a real submission.
- Fix approach: add `--dry-run connect`, add source/sink dry-run SPI hooks, and validate the source -> transform -> sink path without submitting a job.
- One-line summary: the framework shell is in place, but the latest head still does not provide the real connector-side connectivity and schema-inference behavior that the docs now promise.

# Simple example
- A user can reasonably read the new docs and expect `--dry-run connect` to validate real connector connectivity and infer real source schemas for downstream checks.
- On the current head, most existing connectors still do neither:
  - the connectivity hooks are still default no-ops
  - the default source-schema path still falls back to `CatalogTableUtil.buildSimpleTextTable()` unless schema metadata is explicitly present in the config

# Runtime chain I checked
```text
CLI
  -> SeaTunnelConfValidateCommand.execute()
      -> static validation first
      -> if dry-run == CONNECT
          -> new DryRunConnectValidator(...).validate()

source path
  -> validateSource(...) [DryRunConnectValidator.java:103-120]
      -> factory.inferSchemaForDryRun(context)
      -> factory.validateConnectionForDryRun(context, catalogTables)

default source inference
  -> TableSourceFactory.inferSchemaForDryRun(...) [TableSourceFactory.java:68-69]
      -> discoverTableSchemas(...)
  -> TableSchemaDiscoverer.discoverTableSchemas() [TableSchemaDiscoverer.java:71-90]
      -> no schema/table_config/table_list
      -> fallback = CatalogTableUtil.buildSimpleTextTable()

sink path
  -> validateSinkFactory(...)
      -> factory.validateConnectionForDryRun(context)
  -> TableSinkFactory.validateConnectionForDryRun(...) default no-op
```

# Findings
Issue 1: on the current head, no real existing connector implements `validateConnectionForDryRun`, so `--dry-run connect` still cannot deliver the advertised connectivity check
- Location:
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/factory/TableSourceFactory.java:82-83`
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/factory/TableSinkFactory.java:60`
  - `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/DryRunConnectValidator.java:118,245`
  - `docs/en/engines/zeta/user-command.md:74`
- Why this is a blocker:
  I checked the current head with `git grep`, and the only matches are the two default interface declarations plus the DryRunConnectValidator call sites. There are no real connector overrides in this revision yet.
- Risk:
  users can get a successful `--dry-run connect` result and assume credentials / permissions / target reachability were actually checked, when many connectors still did nothing.
- Better fix:
  - Option A, recommended: implement the hook for at least a first batch of real source/sink connectors before merging the feature as documented.
  - Option B: if this PR only wants to land the framework SPI first, explicitly document that connectivity validation is connector opt-in and currently incomplete.
- Severity: High

Issue 2: the default source-schema inference still falls back to `buildSimpleTextTable()`, so downstream schema validation is not reliable on the normal source path
- Location:
  - `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/DryRunConnectValidator.java:103-120`
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/factory/TableSourceFactory.java:68-69`
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/metalake/TableSchemaDiscoverer.java:71-90`
- Why this is a blocker:
  the current framework treats `inferSchemaForDryRun(...)` as the source of truth for downstream transform/sink compatibility, but on the latest head that still degenerates to a synthetic single-text-column table for most real connectors unless schema metadata is explicitly present in the config.
- Risk:
  `--dry-run connect` can report the wrong compatibility result because it is validating against a synthetic placeholder schema instead of the real upstream source schema.
- Better fix:
  - Option A: add connector-specific `inferSchemaForDryRun(...)` implementations for real sources.
  - Option B: scale back the docs and the feature claim so this mode is only described as reliable for explicit-schema sources or connectors that implement the new hook.
- Severity: High

# Compatibility
- CLI/API shape: additive only.
- Behavior: partially incompatible with the new docs / user expectation, because the implementation currently promises more than the connector ecosystem actually delivers.

# Performance / side effects
- The framework path itself is lightweight and fine.
- The real side effect is correctness / operator trust: this mode can currently look stronger than it actually is.

# Tests / stability
- The newly added tests are structurally stable.
- The gap is coverage, not flakiness: there is still no real connector-level dry-run connect proof on the current head.
- Stability rating: Stable.

# Merge conclusion

1. Blocking items
- Issue 1: land real connector-side connectivity validation, or explicitly document that the current implementation is still connector opt-in/incomplete.
- Issue 2: land real source-schema inference on actual connector paths, or scale back the user-facing schema-validation claim.

- None beyond the blockers above; these are the core feature-completeness gates.

Happy to re-review once the implementation and the user-facing feature claim are brought back into alignment.

nzw921rx (commented):
Thank you for your contribution. It is an excellent solution that addresses users' actual pain points🚀

At present, I think the requirements are still in the coding and implementation stage, so I would like to offer some of my own opinions on this:

Firstly, I believe that the new dry run method for runtime checks should not directly modify the downstream stable contract interface: `TableSinkFactory`/`TableSourceFactory`. We should design a separate top-level interface for this function and separate it from the existing stable interface responsibilities. If we merge this with the existing stable interface responsibilities, it will destroy the single responsibility of the interface function and gradually increase maintenance costs. It may be confusing for new contributors to implement stable interfaces

DanielLeens:
Thanks @nzw921rx. I broadly agree with keeping the new runtime dry-run checks separate from the existing stable `TableSourceFactory` / `TableSinkFactory` responsibilities.

From Daniel's side, I already have source-level blockers on the current unchanged head, and this interface-boundary concern reinforces that the feature is still in the implementation stage rather than merge-ready yet. Since there is still no new code after the latest Daniel review, I will keep this as a reply-only follow-up and wait for the next code update before doing another full re-review.

DanielLeens:

- base branch: `dev`
- head SHA: `08ea88e72da98454e7ce7e5d1aee1841bc96fd1c`
- compare status: `diverged`
- ahead_by: `1`
- behind_by: `26`
- mergeable_state: `unknown`

DanielLeens:

- base branch: `dev`
- head SHA: `08ea88e72da98454e7ce7e5d1aee1841bc96fd1c`
- compare status: `diverged`
- ahead_by: `1`
- behind_by: `34`

SEZ9 (changes requested):
Reviewed this PR. Found 4 high and 5 medium priority item(s) worth addressing before merge:

### 1. 🔴 High · Test — Flink and Spark connector e2e modules are dropped from the integration-test module aggregation, silently removing e2e coverage
The updated `if:` condition and `modules=` concatenation in the `Make integration test modules` step no longer include `steps.cv2-flink-e2e-modules.outputs.modules` and `steps.cv2-spark-e2e-modules.outputs.modules` (and the corresponding `connector_modules="$connector_modules"'...'` append lines are deleted). Two concrete failure scenarios: (1) a PR that only touches `seatunnel-e2e/seatunnel-connector-v2-e2e` Flink/Spark test modules now produces empty inputs everywhere, so the `it-modules` step is skipped entirely and the PR merges with zero integration testing; (2) connector changes that previously pulled in their Flink/Spark e2e counterparts now only run Zeta-path e2e, so Flink/Spark-specific translation regressions (e.g., in the new dry-run SPI defaults this PR adds to `TableSourceFactory`/`TableSinkFactory`) go undetected until after merge. This is especially risky in a PR labeled `Spark` and `e2e` that itself changes engine-facing SPI. Fix: either restore the two module outputs in both the `if:` condition and the `modules=`/`connector_modules=` aggregation, or, if Flink/Spark e2e detection was intentionally consolidated into `cv2-e2e-modules`, state that in the PR description and verify the filter step actually emits Flink/Spark e2e paths under that output.

**Where (this PR):**
- `/.github/workflows/backend.yml:295  if: ${{ steps.filter.outputs.api == 'false' && (steps.engine-modules.outputs.modules != '' || steps.cv2-modules.outputs.modules != '' || steps.cv2-e2e-modules.outputs.modules != '' || steps.engine-e2e-modules.outputs.modules != '') }}`
- `/.github/workflows/backend.yml:297  modules='${{ steps.cv2-e2e-modules.outputs.modules }}${{ steps.engine-e2e-modules.outputs.modules }}${{ steps.engine-modules.outputs.modules }}${{ steps.cv2-modules.outputs.modules }}'`
- `/.github/workflows/backend.yml:312  connector_modules="$connector_modules"'${{ steps.cv2-e2e-modules.outputs.modules }}'`

**Suggested fix** (`.github/workflows/backend.yml:297`): `cv2-flink-e2e-modules` and `cv2-spark-e2e-modules` were removed from both this aggregation and the `if:` condition above. That means a PR touching only Flink/Spark e2e modules will skip `it-modules` entirely, and connector changes lose their Flink/Spark e2e runs — a silent coverage regression that this very PR (labeled Spark/e2e, adding new factory SPI defaults) would benefit from. Please restore these outputs in both places, or explain how Flink/Spark e2e paths are now captured by `cv2-e2e-modules`.

### 2. 🔴 High · Robustness — The core feature (DryRunConnectValidator + factory SPI hooks) is not reviewable in this diff — 654 files of CI/docs churn hide the code paths that need failure-mode review
The PR description promises new `TableSourceFactory#inferSchemaForDryRun`, `TableSourceFactory#validateConnectionForDryRun`, `TableSinkFactory#validateConnectionForDryRun` hooks and a `DryRunConnectValidator` that opens network connections during CLI validation — exactly the code that needs scrutiny for resource leaks (catalogs/connections must be closed via try-with-resources), null handling of `discoverTableSchemas` defaults, retry/timeout behavior on unreachable endpoints, and error aggregation across the DAG (one failing sink should not abort validation of the rest with a bare stack trace). None of that code appears in the diff; instead the visible changes are `.asf.yaml`, workflow version bumps, docs rewrites, and a `.skills/` directory. Concrete impact: reviewers cannot verify that `--dry-run connect` doesn't hang indefinitely on a firewalled JDBC endpoint (no timeout), leak un-closed `Catalog` instances, or NPE when a factory returns null from the default `inferSchemaForDryRun` delegate. Fix: split this into (1) the API/core `connect` feature with its e2e tests, (2) CI workflow maintenance, (3) docs, and (4) drop `.asf.yaml`/`.skills` entirely — then the validator's exception handling, connection timeouts, and resource closing can actually be reviewed.

**Where (this PR):**
- `/.github/workflows/backend.yml:381-382  GITHUB_TOKEN: ${{ github.token }}\n  GH_TOKEN: ${{ github.token }}`
- `/.asf.yaml:51-52  pull_requests:\n  allow_update_branch: true`
- `/.github/workflows/backend.yml:99  curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash`

### 3. 🔴 High · Security — SHA-pinned Helm setup action replaced with unpinned `curl | bash` from a moving upstream branch — supply-chain and reliability regression
The previous step used `azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4`, which was pinned to an immutable commit SHA. The replacement pipes `https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3` straight into `bash`, which (a) executes whatever is currently on Helm's `main` branch inside our CI runner (which holds `github.token`), (b) installs an unpinned 'latest' Helm version so lint results are non-reproducible, and (c) has no retry or checksum verification, so a transient GitHub raw-content outage or a redirect failure hard-fails the `dead-link` gate for unrelated PRs. Concrete failure scenario: a compromised or breaking change to Helm's install script silently runs in Apache CI, or a 503 from raw.githubusercontent.com fails the job with no backoff. Fix: keep the SHA-pinned `azure/setup-helm` action (optionally with an explicit `version:` input), or download a specific Helm release tarball, verify its published sha256, and wrap the download in `nick-fields/retry@v3` like the other install steps in this workflow.

**Where (this PR):**
- `/.github/workflows/backend.yml:97-101  run: |\n  if ! command -v helm >/dev/null 2>&1; then\n    curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash\n  fi\n  helm version`

**Suggested fix** (`.github/workflows/backend.yml:99`): This replaces a commit-SHA-pinned action (`azure/setup-helm@1a275c3b...`) with an unpinned `curl ... | bash` from Helm's `main` branch. That is a supply-chain regression (arbitrary upstream script executed in CI with token access), it installs a non-deterministic Helm version, and there is no retry/checksum, so a transient raw.githubusercontent.com failure will break this gate for unrelated PRs. Please keep the SHA-pinned `azure/setup-helm` action with an explicit `version:` input, or download a pinned Helm release and verify its sha256 (wrapped in `nick-fields/retry@v3` like the other install steps).

### 4. 🔴 High · Security — .asf.yaml swaps a repository collaborator (`silenceland` → `heye1005`) and enables `allow_update_branch` — repo-permission changes hidden inside a feature PR
The `.asf.yaml` `collaborators` list grants GitHub triage/permissions on the apache/seatunnel repository, and this PR silently replaces `silenceland` with `heye1005` while also adding `pull_requests.allow_update_branch: true`. Neither change has anything to do with dry-run connect validation, and repo-governance changes like collaborator grants must go through explicit PMC discussion, not ride along in a 654-file feature PR where they can be merged unnoticed — this is exactly the pattern reviewers flagged before (see the PR #10583 review asking why an unrelated `.asf.yaml` change was included). `allow_update_branch` additionally lets people with write access push merge-commits into contributor PR branches, which changes what code runs in PR CI. Please remove both `.asf.yaml` hunks from this PR; if the collaborator change is legitimate, submit it as a standalone PR referencing the dev@ / private@ decision that authorized it.

**Where (this PR):**
- `.asf.yaml:42      - heye1005 (replacing '- silenceland')`
- `.asf.yaml:51    pull_requests:`
- `.asf.yaml:52      allow_update_branch: true`

**Suggested fix** (`.asf.yaml:42`): Swapping a collaborator (`silenceland` → `heye1005`) grants repository permissions and is a governance change — it must not be bundled into a feature PR (same concern was raised on PR #10583 for an unrelated `.asf.yaml` edit). The added `pull_requests.allow_update_branch: true` also changes who can push into contributor PR branches. Please drop both `.asf.yaml` changes here and, if the collaborator update is authorized, open a dedicated PR referencing the PMC decision.

### 5. 🟠 Medium · Security — New `connect` dry-run SPI performs authenticated network connections from the CLI path — connector implementations must not echo credentials or raw connection errors into CLI/exception output
This PR introduces `validateConnectionForDryRun(context)` hooks that, by design, take user-supplied config (hosts, users, passwords, URLs) and attempt live connections from `DryRunConnectValidator` before any job is submitted. The recurring failure mode with such validation paths is that connectors surface the failing JDBC URL / broker string / driver exception message verbatim, which frequently embeds passwords, tokens, or SAS query parameters into CLI output and logs — a broader exposure surface than a running job because dry-run output is exactly what users paste into tickets and chat. The truncated diff shown here contains only the workflow/doc changes, so before merge the `DryRunConnectValidator` and every connector override need a pass to ensure (a) all connection-failure messages are wrapped in SeaTunnel error codes with credentials masked (reuse the existing `ConfigShade`/sensitive-option masking conventions from `OptionRule` sensitive keys), and (b) endpoints in config are not blindly dereferenced in shared/multi-tenant Zeta clusters, since `--dry-run connect` otherwise becomes a convenient SSRF probe against internal networks using the engine host as the egress point. Concretely: centralize exception sanitization in `DryRunConnectValidator` (strip `Throwable#getMessage` of values matching sensitive option keys) rather than trusting each connector to do it, and document that `validateConnectionForDryRun` implementations must never include raw connection strings in thrown exceptions.

**Where (this PR):**
- `docs/en/connectors/sink/Elasticsearch.md:+10/-1 (connect dry-run docs added across connector sink docs, e.g. Activemq.md +55/-98, Cassandra.md +88/-24)`
- `config/plugin_config:+2/-1 (plugin config updated for the new dry-run connect validation path)`

**Suggested fix** (`config/plugin_config:1`): For the new `--dry-run connect` path: since `validateConnectionForDryRun` deliberately opens authenticated connections using raw user config, please centralize exception sanitization in `DryRunConnectValidator` — mask any values belonging to sensitive `OptionRule` keys (passwords/tokens/SAS params) before the failure message reaches CLI output, instead of relying on each connector to sanitize its own driver exceptions. JDBC/broker driver messages routinely embed full connection strings, and dry-run output is exactly what users copy into issues. Also worth stating in the SPI javadoc that implementations must not include raw connection URLs in thrown exceptions, and considering that on shared Zeta clusters this mode lets any config author probe internal endpoints (SSRF-style), so the validation should run client-side where possible.

### 6. 🟠 Medium · Docs — No user-facing documentation for the new `--dry-run connect` mode or the new factory SPI hooks
The PR introduces a new user-facing CLI value (`sh bin/seatunnel.sh --config <config> --dry-run connect`) and three new SPI default methods (`inferSchemaForDryRun`, `validateConnectionForDryRun` on `TableSourceFactory`/`TableSinkFactory`), but among the changed files I see only unrelated connector-doc rewrites (Activemq.md, Aerospike.md, etc.) — no update to the Zeta CLI/command usage docs describing the `connect` mode, its semantics (no readers/writers/committers/save-mode created), exit behavior, or connector support matrix, and no changelog entry. In the #10763 review we explicitly agreed future modes stay out of the public path "until runtime + docs + tests are ready together", so docs are a merge requirement here, not a follow-up. Please add: (1) `docs/en/command/usage.mdx` (and the zh counterpart) coverage of `--dry-run connect`, (2) Javadoc on the new default SPI methods stating the no-side-effect contract for connector implementers, and (3) a release-note/changelog entry, plus an example config in `docs` showing the connect validation output.

**Where (this PR):**
- `PR description: `sh bin/seatunnel.sh --config <config> --dry-run connect` (new CLI mode) with no corresponding CLI usage doc in the diff`
- `changed docs are unrelated connector rewrites, e.g. docs/en/connectors/sink/Activemq.md (+55/-98), docs/en/connectors/sink/Aerospike.md (+83/-41)`

### 7. 🟠 Medium · Style — Massive unrelated CI churn (30+ action version bumps, token source changes) mixed into the dry-run feature PR makes the core SPI/serialization impact unreviewable
This PR's stated purpose is the new `connect` dry-run mode with `TableSourceFactory#inferSchemaForDryRun` / `validateConnectionForDryRun` SPI hooks, but the diff also carries wholesale `actions/checkout@v2→v4`, `actions/setup-java@v3→v4`, `actions/setup-node@v2→v4`, `nick-fields/retry@v2→v3` bumps across dozens of jobs, plus swapping `secrets.GITHUB_TOKEN` for `github.token` in the docs build. Each of these can independently break CI (e.g. node20-based v4 actions have different runner requirements, checkout v4 changes default fetch behavior), and if one does, the revert would take the feature — including new public SPI methods that connectors depend on — down with it. It also drowns out the actual API-compatibility review: with 654 changed files, the new factory SPI defaults and `DryRunConnectValidator` cannot be meaningfully audited for compatibility. Split the CI modernization into its own PR (as was requested for #10763's scope) and keep this PR limited to the dry-run feature, its docs, and its tests.

**Where (this PR):**
- `.github/workflows/backend.yml:41  - uses: actions/checkout@v4`
- `.github/workflows/backend.yml:340  - uses: actions/setup-java@v4`
- `.github/workflows/backend.yml:346  uses: nick-fields/retry@v3`
- `.github/workflows/backend.yml:380  GITHUB_TOKEN: ${{ github.token }}`

**Suggested fix** (`.github/workflows/backend.yml:41`): These blanket action version bumps (checkout v2/v3→v4, setup-java v3→v4, setup-node v2→v4, retry v2→v3) across 30+ jobs are orthogonal to the dry-run connect feature. If any bump breaks a job, reverting takes the new SPI methods with it, and the 654-file diff makes the actual `validateConnectionForDryRun` API review very hard. Please split the CI modernization into a dedicated PR and keep this one scoped to the dry-run feature, matching the scope discipline requested on #10763.

### 8. 🟠 Medium · Docs — New `.skills/` files need ASF license headers and don't belong in this feature PR
This PR adds `.skills/seatunnel-test-suite/SKILL.md` (+240 lines) and `.skills/seatunnel-test-suite/agents/openai.yaml` (+19 lines), which are AI-agent tooling files unrelated to the `connect` dry-run feature. Per ASF policy and our license-header CI check, new source/config files (including YAML) must carry the Apache License header, and `apache-rat` exclusions must not be silently widened to accommodate them — I also note `.gitignore` gained 6 lines in the same commit. Please either drop these files from this PR (preferred, to keep scope tight as agreed in the #10763 review) or, if they are meant to land, add ASF headers to `openai.yaml` and the standard license notice to `SKILL.md`, and raise them in a dedicated PR so the community can discuss committing agent tooling to the repo.

**Where (this PR):**
- `changed files list: .skills/seatunnel-test-suite/SKILL.md (+240/-0)`
- `changed files list: .skills/seatunnel-test-suite/agents/openai.yaml (+19/-0)`

**Suggested fix** (`.skills/seatunnel-test-suite/SKILL.md:1`): These `.skills/` files (SKILL.md, agents/openai.yaml) are unrelated to the `connect` dry-run feature and are missing ASF license headers, which will trip the license-header check (or require widening RAT exclusions, which we shouldn't do silently). Please remove them from this PR and propose the agent tooling separately; if kept, add the standard Apache License header to both files.

### 9. 🟠 Medium · Functional — Unrelated `.asf.yaml` governance changes (collaborator swap, allow_update_branch) bundled into a feature PR
This PR, whose stated purpose is adding a `connect` dry-run mode, also removes `silenceland` from and adds `heye1005` to the GitHub collaborators list and enables `pull_requests.allow_update_branch` — repository governance changes that require PMC visibility and should never ride along in a feature PR. If this PR is later reverted (plausible for a draft/RFC), the collaborator change and branch-update policy get reverted with it, silently revoking `heye1005`'s access. The same pattern was flagged before in this repo (PR #10583 review: "The `.asf.yaml` change seems unrelated to this PR"). Fix: drop the `.asf.yaml` hunks (and the unrelated workflow/docs mass-edits) from this PR and submit them separately with proper sign-off.

**Where (this PR):**
- `/.asf.yaml:42  - heye1005`
- `/.asf.yaml:51-52  pull_requests:\n  allow_update_branch: true`

**Suggested fix** (`.asf.yaml:52`): Swapping a collaborator (`silenceland` → `heye1005`) and enabling `allow_update_branch` are repo-governance changes unrelated to the `--dry-run connect` feature. If this RFC PR is reverted, these changes revert with it. Please move the `.asf.yaml` edits to a dedicated PR with PMC visibility (this was also raised in #10583 for the same file).

junjunclub:
Hi @SEZ9, thank you for the thorough review — this was very helpful.

I want to clarify that most of the issues you flagged were not introduced by this PR's actual changes. The branch had accidentally diverged from an outdated dev base and pulled in ~129 commits / 654 changed files from upstream and unrelated work. That polluted diff included .asf.yaml governance edits, .github/workflows/backend.yml CI churn, .skills/ files, connector doc rewrites, Helm install script changes, etc. — none of which are part of my intended changes.

I've cleaned up the branch by rebasing onto the latest dev and force-pushing. The PR now contains 2 commits and 16 files, scoped to the connect dry-run feature only:

85a073f45 feat: layer1 with jdbc connector 9dfebf63b [Feature][Core] Add connect dry-run validation

What is actually in the current diff:

Core: DryRunConnectValidator, SeaTunnelConfValidateCommand, ClientCommandArgs
API SPI: SupportSourceDryRunValidation, SupportSinkDryRunValidation
Connectors: JDBC source/sink factory hooks, FakeSource, DuckDB dry-run test
Tests: SeaTunnelConfValidateCommandTest, ClientCommandArgsTest, dry-run test factories
Docs: docs/en/engines/zeta/user-command.md, docs/zh/engines/zeta/user-command.md
Mapping to your review items:

Flink/Spark e2e modules dropped from CI — Not in this PR (.github/workflows/backend.yml is unchanged)
Core feature not reviewable in diff — Resolved (DryRunConnectValidator and SPI hooks are now clearly visible)
Helm curl | bash supply-chain regression — Not in this PR 4 & 9. .asf.yaml collaborator / allow_update_branch — Not in this PR
Mass CI action version bumps — Not in this PR
.skills/ files without ASF headers — Not in this PR
Missing docs for --dry-run connect — Partially addressed in docs/en/engines/zeta/user-command.md (and zh counterpart), covering --dry-run connect semantics, connector support matrix, and VALIDATED / SKIPPED behavior
Credential masking in dry-run error output — Valid concern for the actual code; happy to address in a follow-up commit if you'd like centralized sanitization in DryRunConnectValidator
Could you please re-review the updated diff? The force-pushed branch should now show only the dry-run connect feature. I'm happy to address any remaining feedback on the actual 16-file scope — especially around credential sanitization (#5) and any gaps in the docs.

Sorry for the noise in the earlier diff; it was a git base issue on my side, not intentional scope creep.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the current head for the connect dry-run design and the current Build failure.

# What this PR fixes
This PR adds a CLI-only `--dry-run connect` mode that runs static validation first and then lets connector factories opt in to real metadata/connectivity checks.

```text
CLI invocation
  -> ClientCommandArgs parses --dry-run static/connect
  -> SeaTunnelConfValidateCommand validates env/source/transform/sink option rules
  -> DryRunConnectValidator walks the source/transform/sink DAG
      -> factories implementing SupportSourceDryRunValidation / SupportSinkDryRunValidation are VALIDATED
      -> unsupported connectors are reported as SKIPPED, not silently treated as verified
```

# Current review findings
The previous source-design blockers are cleared for me now: the SPI is explicit opt-in, Jdbc/FakeSource provide concrete implementations, and the docs clearly say unsupported connectors are `SKIPPED` rather than validated.

## Issue 1: The new service loader file is missing a valid ASF license header
- Location: `seatunnel-core/seatunnel-starter/src/test/resources/META-INF/services/org.apache.seatunnel.api.table.factory.Factory:1`
- Evidence: current Build failed in `License header`; `license-eye` reports this exact file as invalid.
- Why this blocks merge: this is a new PR file and ASF license headers are mandatory for new files unless explicitly excluded by project policy.
- Suggested fix: add the appropriate comment-style ASF header accepted by the service loader / license checker, or update the project license exclusion only if this file type is intentionally excluded elsewhere.
- Severity: High

- Fetched the PR head into `seatunnel-review-11186` and reviewed the current full diff.
- Checked the failing GitHub Build check and the `License header` job log.

# Merge conclusion
Blocking items:
1. Issue 1: fix the missing license header and rerun Build.

This is close: once the license header check is fixed and CI is green, I am happy to re-review quickly.

DanielLeens (changes requested):
Sorry for missing this in the last review — that's on me. I re-reviewed the latest head (`91d2d63a027799c7e8c8667199dd3928f47744d3`) from scratch against the cleaned 16-file diff, traced the current `--dry-run connect` call chain again, and there is still one source-side blocker I should have called out earlier.

I also want to note two things up front:
- I agree with @nzw921rx's earlier design direction here. Splitting the feature into `SupportSourceDryRunValidation` / `SupportSinkDryRunValidation` instead of extending the stable `TableSourceFactory` / `TableSinkFactory` contract was the right move.
- The previous Daniel blockers about placeholder-schema validation and the missing ASF header are cleared for me now on the current head. The `License header` check is green in the current Build run as well.

# What this PR solves
- User pain: today `--check` / `--dry-run static` can validate config syntax and DAG wiring, but they still cannot tell users whether source/sink connectivity and schema compatibility will hold before a real submission.
- Fix approach: add a CLI-only `--dry-run connect` mode, introduce explicit opt-in dry-run validation SPI hooks, and let connectors validate metadata/connectivity without creating runtime readers/writers or submitting a job.
- One-line summary: the feature shape is much closer now, but the current failure path still lets raw JDBC connection text bubble straight into CLI output.

# Simple example
- A user can now run `sh bin/seatunnel.sh --config ... --dry-run connect` and reasonably expect one of two outcomes:
  - a clean preflight result, or
  - a safe failure message they can paste into an issue/comment without leaking credentials.
- On the current head, the first part is mostly in place, but the second part is not guaranteed yet: some JDBC failures can still surface the original connection URL / driver message directly.

# Full runtime chain I checked
```text
CLI
  -> ClientCommandArgs.buildCommand() [ClientCommandArgs.java:147-152]
      -> SeaTunnelConfValidateCommand.execute() [SeaTunnelConfValidateCommand.java:92-180]
          -> static config / DAG / option validation first
          -> if dryRun == CONNECT
              -> new DryRunConnectValidator(...).validate() [DryRunConnectValidator.java:103-137]

source path
  -> validateSource(...) [DryRunConnectValidator.java:140-200]
      -> if factory implements SupportSourceDryRunValidation
          -> inferSchemaForDryRun(...)
          -> validateConnectionForDryRun(...)
      -> otherwise
          -> explicit schema from config => downstream schema checks still meaningful
          -> no explicit schema => mark as SKIPPED / unknown, no placeholder-schema fake success

current JDBC source implementation
  -> JdbcSourceFactory.inferSchemaForDryRun() [JdbcSourceFactory.java:125-136]
      -> JdbcCatalogUtils.getTables() [JdbcCatalogUtils.java:72-120]
          -> findCatalog(...)
          -> jdbcCatalog.open() [AbstractJdbcCatalog.java:171-173]
          -> load real table metadata

sink path
  -> validateSink(...) [DryRunConnectValidator.java:291-355]
      -> if factory implements SupportSinkDryRunValidation
          -> TableSinkFactoryContext.replacePlaceholderAndCreate(...)
          -> validateConnectionForDryRun(...)

current JDBC sink implementation
  -> JdbcSinkFactory.validateConnectionForDryRun() [JdbcSinkFactory.java:270-339]
      -> find catalog or plain JDBC connection
      -> open connection
      -> check target-table existence / field compatibility

failure propagation path
  -> SimpleJdbcConnectionProvider.getOrEstablishConnection()
       "No suitable driver found for " + jdbcConfig.getUrl()
       [SimpleJdbcConnectionProvider.java:115-117]
     or
     AbstractJdbcCatalog.getConnection()
       "Failed connecting to %s via JDBC."
       [AbstractJdbcCatalog.java:150-156]
  -> DryRunConnectValidator.wrap()
       location(...) + " failed: " + e.getMessage()
       [DryRunConnectValidator.java:464-472]
  -> SeaTunnelConfValidateCommand.execute()
       "Connectivity check failed: " + e.getMessage()
       [SeaTunnelConfValidateCommand.java:175-180]
  -> raw lower-layer message reaches CLI output
```

# Findings
+1 to @SEZ9's earlier point about sanitizing dry-run connection failures — after rechecking the cleaned current diff, that concern still applies to the latest head.

Issue 1: the new JDBC dry-run failure path can still expose the raw connection URL / raw lower-layer exception text to CLI users
- Location:
  - `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/connection/SimpleJdbcConnectionProvider.java:115-117`
  - `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/AbstractJdbcCatalog.java:150-156`
  - `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/DryRunConnectValidator.java:464-472`
  - `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/SeaTunnelConfValidateCommand.java:175-180`
- Why this is a blocker:
  the latest head now intentionally opens real source/sink connections on the CLI path. When that fails, the current exception chain still concatenates lower-layer JDBC text directly into the final `ConfigCheckException`. In the source path that can come from `JdbcCatalogUtils.getTables() -> AbstractJdbcCatalog.open()`. In the sink path it can come from either `catalog.open()` or the plain JDBC connection fallback. So a URL with embedded credentials / tokens, or a driver message echoing the original connection string, can be surfaced straight to the CLI.
- Risk:
  this is a real source-side security issue, not just a wording nit:
  - users frequently paste dry-run failures into tickets / chat / PR comments
  - the current docs explicitly advertise `Jdbc` as a supported real dry-run connector
  - the affected path is therefore part of the normal supported feature, not an obscure edge case
- Better fix:
  - Option A, recommended: centralize message sanitization in `DryRunConnectValidator.wrap()` and/or right before the final `ConfigCheckException` is built, masking URL query params and sensitive keys such as password/token/secret/signature.
  - Option B: also stop building lower-layer JDBC exceptions with the raw `jdbcConfig.getUrl()` in the message in the first place, so future callers cannot accidentally leak it either.
- Severity: High
- Already raised by others: Yes (`@SEZ9` in review `#4650541932`)

# Compatibility
- API / config compatibility is otherwise good from what I checked:
  - the new SPI is additive and explicitly opt-in
  - existing connectors that do not implement it are reported as `SKIPPED`
  - the placeholder-schema false-positive problem from the older revision is fixed on the current head

# Performance / side effects
- CPU / memory overhead stays low; this is metadata/connectivity work only.
- Resource handling in the JDBC implementation looks reasonable (`Catalog` / `Connection` are opened in `try-with-resources` paths).
- The remaining side effect that still needs to be fixed is failure-message safety.

# Tests / stability
- The new tests look structurally stable to me.
- `DuckDBConnectDryRunValidationTest` uses a local embedded DuckDB file, no sleeps, no random ports, and explicit setup/cleanup.
- So the test concern here is not flakiness; it is missing coverage for the new sanitization requirement.
- Stability rating: Stable.

# Current CI / merge gate snapshot
- I checked the current head against `dev`: `ahead_by=4`, `behind_by=0`, compare status `ahead`.
- So this is **not** an upstream-sync-first case.
- The current required top-level `Build` gate is still not green because `Run / kudu-connector-it (11, ubuntu-latest)` was cancelled in `https://github.com/junjunclub/seatunnel/actions/runs/28981800276/job/86044574793`.
- Relevant jobs for this PR's actual scope are otherwise green on the current run:
  - `Run / License header`: success
  - `Run / Dependency licenses`: success
  - `Run / unit-test (8, ubuntu-latest)`: success
  - the `jdbc-connectors-it-*` jobs I checked: success

# Merge conclusion

1. Blocking items
- Issue 1: sanitize the dry-run failure path so raw JDBC URLs / lower-layer exception text do not reach CLI output.
- After that, rerun `Build` and get the cancelled `kudu-connector-it` lane back to a green overall Build result.

- The title still has a trailing `validation1`; worth cleaning up before merge, but that is not a code blocker from my side.

This is much closer than the earlier revisions, and I appreciate the cleanup work on the diff shape. Once the failure-message sanitization is in place, I am happy to take another full pass on the updated head.

DanielLeens:
Rechecked the same unchanged head `91d2d63a0277` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
CLI
  -> ClientCommandArgs.buildCommand() [ClientCommandArgs.java:147-152]
      -> SeaTunnelConfValidateCommand.execute() [SeaTunnelConfValidateCommand.java:92-180]
          -> static config / DAG / option validation first
          -> if dryRun == CONNECT
              -> new DryRunConnectValidator(...).validate() [DryRunConnectValidator.java:103-137]

source path
  -> validateSource(...) [DryRunConnectValidator.java:140-200]
      -> if factory implements SupportSourceDryRunValidation
          -> inferSchemaForDryRun(...)
          -> validateConnectionForDryRun(...)
      -> otherwise
          -> explicit schema from config => downstream schema checks still meaningful
          -> no explicit schema => mark as SKIPPED / unknown, no placeholder-schema fake success

current JDBC source implementation
  -> JdbcSourceFactory.inferSchemaForDryRun() [JdbcSourceFactory.java:125-136]
      -> JdbcCatalogUtils.getTables() [JdbcCatalogUtils.java:72-120]
          -> findCatalog(...)
          -> jdbcCatalog.open() [AbstractJdbcCatalog.java:171-173]
          -> load real table metadata

sink path
  -> validateSink(...) [DryRunConnectValidator.java:291-355]
      -> if factory implements SupportSinkDryRunValidation
          -> TableSinkFactoryContext.replacePlaceholderAndCreate(...)
          -> validateConnectionForDryRun(...)

current JDBC sink implementation
  -> JdbcSinkFactory.validateConnectionForDryRun() [JdbcSinkFactory.java:270-339]
      -> find catalog or plain JDBC connection
      -> open connection
      -> check target-table existence / field compatibility

failure propagation path
  -> SimpleJdbcConnectionProvider.getOrEstablishConnection()
       "No suitable driver found for " + jdbcConfig.getUrl()
       [SimpleJdbcConnectionProvider.java:115-117]
     or
     AbstractJdbcCatalog.getConnection()
       "Failed connecting to %s via JDBC."
       [AbstractJdbcCatalog.java:150-156]
  -> DryRunConnectValidator.wrap()
       location(...) + " failed: " + e.getMessage()
       [DryRunConnectValidator.java:464-472]
  -> SeaTunnelConfValidateCommand.execute()
       "Connectivity check failed: " + e.getMessage()
       [SeaTunnelConfValidateCommand.java:175-180]
  -> raw lower-layer message reaches CLI output
```

## Merge conclusion

1. Blocking items

2. Current gate update
- The CI / merge-gate fact I referenced earlier has changed. The current status is `CI is green on the current head`: https://github.com/apache/seatunnel/actions/runs/28981801879/job/86002040236

DanielLeens:
Rechecked the latest head again after the branch cleanup and the SPI split.

The two earlier Daniel-side core blockers I cared about are closed for me on this revision:
- the dry-run hook is no longer being forced into the stable `TableSourceFactory` / `TableSinkFactory` contract; it now sits behind separate opt-in SPI types;
- the docs now make it clear that `SKIPPED` does not mean a real connector connectivity check happened, and the failure sanitization path is in place.

So my current Daniel-side conclusion on this head is no longer "code blocker still open". The remaining gate looks more like the current top-level `Build` / integration-test state than a reopened source issue in the CLI dry-run path.

If the current `Build` still ends red after the remaining jobs finish, please paste the failing lane link here and I can narrow down whether it is actually tied to the dry-run implementation or just broader integration noise.

davidzollo (approved):
+1
LGTM

PR #11189, opened 2026-06-25
[Test][Zeta] Stabilize cluster role cancel status assertion

DanielLeens:
## Why

`SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus` can observe `CANCELING` on the client right after `cancelJob` is accepted during role failover. The existing assertion expects the client-facing status string to become `CANCELED` directly, which makes unrelated PRs fail on shared CI.

## What

- wait for `waitForJobCompleteV2()` to return a terminal `CANCELED` job result first
- assert the client-side job status after the terminal result is available
- keep the change scoped to the flaky test path only

## Validation

- `./mvnw spotless:apply`
- `./mvnw -f seatunnel-engine/seatunnel-engine-client/pom.xml -Dtest=SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus -DfailIfNoTests=false -Dspotless.check.skip=true test`

## Notes

- A broader `./mvnw -pl seatunnel-engine/seatunnel-engine-client -am ... test` attempt on current `dev` is blocked by an unrelated compile issue in `seatunnel-config-shade`, so the closest isolated validation was run against the exact touched test method.


DanielLeens (commented):
Thanks for the update. I re-checked the full diff across the JDBC schema-evolution assertions, the Kafka exactly-once helper, and the cluster-role cancel test.

From the source side, the changes look reasonable:
- the JDBC assertions now compare data in source-column order instead of depending on sink physical column order,
- the Kafka helper counts the records a `read_committed` consumer can actually see,
- and the cluster-role test now waits for the terminal job result before checking the client-facing status string.

The blocker right now is merge state:

1. GitHub is reporting merge conflicts with `dev` (`mergeStateStatus=DIRTY`, `mergeable_state=dirty`).
2. The branch is behind upstream (`behind_by=47`, compare status `diverged`).
3. The current `Build` run was cancelled on that stale/conflicted head, so the next useful step is to sync with the latest `dev`, resolve the conflicts, and rerun CI.

After that refreshed head is available, I’m happy to re-review.

DanielLeens:
Thanks for the refresh. I rechecked the latest head across the Kafka exactly-once helper and the cluster-role status assertions again.

## Current state on the latest head
- I do not see a reopened source-side blocker in the current revision.
- The current head now looks clear to me on the test-logic side.
- The current `Build` run is still queued, so the remaining gate is CI rather than a new blocker from Daniel's side.

So from the Daniel technical-review side, I do not have a new source blocker on this revision.

DanielLeens (commented):
Rechecked the latest head from Daniel's side. I still do not see a new source-side blocker on this head; the thing that changed since the earlier round is the CI / merge-gate fact, not the code-side conclusion.

Current gate
- The current `Build` check is still failing.
- This branch is also behind the latest `dev` (`behind_by=37`, compare status: `diverged`).

Next step
- Since the remaining blocker here is the stale CI / merge gate rather than a newly identified source issue, the lowest-cost next step is to sync the latest `dev` and rerun `Build` first.
- If `Build` still fails after the sync, we can narrow it down again on the refreshed head.

DanielLeens (commented):
Technical self-review of the full diff at `b4d1fdcdc5f24e96da28e8afecc89f5dbf7fe6b5`: waiting for the terminal result is the right semantic boundary, but the new test wait is not bounded and can turn the original flaky failure into a hung CI job.

1. **`waitForJobCompleteV2()` bypasses the test's timeout.** `SeaTunnelEngineClusterRoleTest.java:525-535` calls this blocking method before entering the 30-second Awaitility block. `ClientJobProxy.java:93-106` waits on `jobFuture.get()` and is configured for up to 100,000 retries with a 2-second retry sleep. If cancellation or failover regresses, the test no longer fails within the previous 120-second bound; it can hold a CI executor for many hours. Please execute the terminal-result wait under a real test deadline (for example, a bounded future or JUnit timeout that can interrupt the wait), then keep the final `CANCELED` assertion.

This is a high-risk test-stability blocker for a PR whose purpose is to stabilize shared CI. Also, #11139 now carries the same effective code change; please keep only one of these duplicate PRs active. The current required `Build` is still queued, and this self-authored PR still needs an independent maintainer decision after the blocker is fixed.

DanielLeens (commented):
CI follow-up on the unchanged head: the previously queued `Build` has now failed. The first reproducible failure is the shared dead-link check treating the Tablestore JDBC placeholder as a live URL and receiving HTTP 400:

https://github.com/DanielLeens/seatunnel/actions/runs/29684932066/job/88187471271

That repository-wide issue has an open fix in #11502 whose current `Build` is green, but the fix is not in `dev` yet. Please do not change unrelated JDBC documentation in this test PR; after #11502 lands, sync/retriggering is the lowest-cost CI step.

The technical self-review remains blocking for the independent reason already recorded: `waitForJobCompleteV2()` is outside a real test deadline and can hold CI for many hours if the terminal result is lost. This self-authored PR still needs that bounded-wait fix and independent maintainer review.

DanielLeens:
Closing this older duplicate because #11457 has already landed on `dev` and stabilizes the same `SeaTunnelEngineClusterRoleTest` with the current cancellation semantics (`cancelJob(jobId, true)` and accepting `CANCELING` or `CANCELED`). Keeping this PR would reintroduce the older behavior. Thanks.

PR #11190, opened 2026-06-26
[Fix][Connector-V2] Avoid Xugu pooled connection isValid checks

DanielLeens:
## Summary
- avoid using `Connection.isValid(timeout)` for Xugu pooled connections
- configure the JDBC sink pool to use a Xugu-compatible validation query
- add regression tests for the validation fallback and pool configuration

## Root cause
`JdbcXuguIT` fails in shared CI because the Xugu driver throws from `Connection.isValid(timeout)`, which is used by the pooled JDBC path during writer initialization.

## Validation
- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -pl seatunnel-connectors-v2/connector-jdbc -am spotless:apply`
- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -pl seatunnel-connectors-v2/connector-jdbc -am -DskipTests -DskipITs -DskipUTs package -Dspotless.skip=true -Dcheckstyle.skip=true`
- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -pl seatunnel-connectors-v2/connector-jdbc -am -DskipTests -DskipITs -DskipUTs install -Dspotless.skip=true -Dcheckstyle.skip=true`
- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -pl seatunnel-connectors-v2/connector-jdbc -DskipITs -DfailIfNoTests=false -Dtest=JdbcConnectionValidationUtilsTest,JdbcSinkWriterTest test -Dspotless.skip=true -Dcheckstyle.skip=true`

## Notes
- `seatunnel-engine-ui` was not built because this change only touches `connector-jdbc`.
- Docker is not available in this environment, so the Xugu integration test path was not rerun locally.


davidzollo (approved):
LGTM

PR #11191, opened 2026-06-26
[Test][E2E] Stabilize IRIS JDBC driver download

DanielLeens:
## What
- switch `JdbcIrisIT` to the published GitHub Pages IRIS driver distribution URL
- replace the container-side `wget` download with retried `curl`

## Why
The failing `jdbc-connectors-it-part-7` job showed `JdbcIrisIT.testJdbcDb` failing before test execution because `wget` could not establish an SSL connection to `raw.githubusercontent.com`. This is shared CI noise unrelated to the tested PR logic.

## Validation
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7 -am spotless:apply -DskipTests`
- attempted `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7 -am -DskipTests -DskipITs=true -Dspotless.skip=true -Dcheckstyle.skip=true test-compile`, but a fresh worktree hit unrelated repository compilation failures in `seatunnel-config-shade`
- attempted module-only `test-compile`, but local Maven resolution was blocked by repeated `my-nexus` 502 metadata failures
- Docker daemon is unavailable on this machine, so the IRIS E2E path could not be re-run locally in this session


DanielLeens (commented):
Thanks for the contribution. I reviewed the latest head from the IRIS E2E bootstrap path through the JDBC plugin-driver injection step, and this change looks good to me.

## What this PR solves
- User pain: the IRIS JDBC E2E can fail before it even reaches the real JDBC assertions because downloading the driver from `raw.githubusercontent.com` intermittently hits TLS handshake problems inside the CI container.
- Fix approach: switch the driver source to the published GitHub Pages artifact URL and download it with `curl -fsSL` plus bounded retries.
- One-line summary: this stabilizes the driver-download bootstrap step without changing the actual JDBC runtime path under test.

```text
IRIS JDBC E2E bootstrap
  -> JdbcIrisIT.extendedFactory
      -> create /tmp/seatunnel/plugins/Jdbc/lib
      -> download the IRIS JDBC jar into the plugin lib directory
  -> AbstractJdbcIT startup
      -> load the Jdbc plugin with that driver present
  -> IRIS source/sink E2E assertions
```

## Key findings
1. The normal path definitely hits this change because the container extension is the only place that injects the IRIS driver into the test plugin directory.
2. The fix matches the real flaky symptom: the unstable part was the raw GitHub download endpoint, not the JDBC logic itself.
3. Using a published GitHub Pages URL plus bounded retries is a tighter and more reliable bootstrap path than a single `wget` against the raw endpoint.

## Findings

### Conclusion: can merge
1. Blocking items
- None.

- None.

Overall, this is a focused E2E stability fix for the IRIS driver bootstrap path, and I'm happy to approve it.

davidzollo (approved):
+1

DanielLeens:

PR #11192, opened 2026-06-26
[Fix][CI] Use workflow token for backend website build

zhangshenghang:
### What is wrong now

Recent `Build` workflow runs can fail with `startup_failure` before any job starts. In that state GitHub does not create job logs, so PRs such as #11133 are blocked without running project build or test code.

Local workflow validation, action metadata, and recent workflow history point to startup-time problems in the shared CI workflow definitions:

- `.github/workflows/backend.yml` used `secrets.GITHUB_TOKEN` when checking out `apache/seatunnel-website`. `GITHUB_TOKEN` is the built-in workflow token, not a user-defined secret, so the reusable workflow should pass `github.token` instead.
- `.github/workflows/build_main.yml` and `.github/workflows/schedule_backend.yml` narrowed the caller token permission to only `packages: write`. The called backend workflow still needs normal repository read access for checkout and workflow token usage, so the callers should explicitly keep `contents: read`.
- `.github/workflows/backend.yml` referenced undefined step ids for connector-v2 E2E module outputs. A workflow expression that reads a missing step id can make the workflow invalid before jobs are scheduled.
- `.github/workflows/schedule_backend.yml` passed boolean `false` to the reusable workflow input `TEST_IN_PR`, while the reusable workflow declares it as a string input.
- `.github/workflows/backend.yml` and `.github/workflows/notify_test_workflow.yml` still used older official GitHub Actions versions. Newer GitHub Actions runners can reject old Node-based action versions before job execution, which matches the no-job/no-log startup failure symptom.
- `.github/workflows/backend.yml` still used `nick-fields/retry@v2`, whose action metadata runs on Node 16. `nick-fields/retry@v3` runs on Node 20.

### What changed

This PR keeps the fix limited to CI workflow startup issues:

- Use `${{ github.token }}` for the backend website build checkout token.
- Keep `contents: read` while granting `packages: write` in the Build and scheduled backend workflow callers.
- Remove the undefined `cv2-flink-e2e-modules` / `cv2-spark-e2e-modules` step-id references from the integration-test module calculation.
- Pass `TEST_IN_PR` as a string expression from the scheduled backend workflow.
- Upgrade official actions used by backend and notify workflows:
  - `actions/checkout` to `v4`
  - `actions/setup-java` to `v4`
  - `actions/setup-node` to `v4`
  - `actions/setup-python` to `v5`
  - `actions/github-script` to `v7`
- Upgrade `nick-fields/retry` usage from `v2` to `v3`.

### Why this is separate from #11133

#11133 only changes JDBC catalog close behavior. Its failing `Build` run did not start any project job and had no logs, and recent `dev` branch Build runs showed the same startup failure pattern. That means the blocker is in shared CI workflow startup, not in #11133 code.

### Validation

- `actionlint` focused check for action/expression startup errors: passed
- `git diff --check`: passed
- `./mvnw -q -DskipTests verify`: passed

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -94,8 +94,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Setup Helm
-        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4
-        id: install
+        run: |
+          if ! command -v helm >/dev/null 2>&1; then
+            curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+          fi
+          helm version
       - name: Lint Chart
         run: helm lint deploy/kubernetes/seatunnel
 
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: true
       - name: Check license header
@@ -49,7 +49,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: true
       - name: Check code style
@@ -106,7 +106,7 @@ jobs:
     # Temporarily ignore this job to avoid blocking PRs
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - run: sudo npm install -g markdown-link-check@3.8.7
       - run: |
           for file in $(find . -name "*.md"); do
@@ -146,7 +146,7 @@ jobs:
           /usr/bin/git checkout apache/dev
           /usr/bin/git checkout '${{ github.ref }}'
           echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: '3.11.0'
       - name: Check for file changes by python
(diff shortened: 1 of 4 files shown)


DanielLeens (changes requested):
Thanks for working on this. I re-reviewed the latest head from the PR-side workflow dispatch path through the backend website-build path.

Runtime path I checked
```text
pull_request_target
  -> .github/workflows/notify_test_workflow.yml
      -> list build_main.yml runs for the PR branch
      -> query check-runs for the branch/ref
      -> look up "Run / License header"
      -> create the PR-side Build check

backend workflow path
  -> .github/workflows/build_main.yml permissions
  -> .github/workflows/backend.yml document job
      -> Docusaurus build uses github.token
```

The website-token change itself is reasonable, but the current head still breaks on the normal PR path before that result can be surfaced cleanly.

1. `notify_test_workflow.yml` still hard-codes a check-run name that does not exist on the current PR path
   - Location:
     - `.github/workflows/notify_test_workflow.yml:118-125`
   - Why this is a problem:
     the workflow still does `check_runs.data.check_runs.filter(r => r.name === "Run / License header")[0]`. On the current head, the check-run list only contains `Notify test workflow` and `labeler`, so `check_run_head` is `undefined`. The run then fails with `TypeError: Cannot read properties of undefined (reading 'head_sha')`.
   - Risk:
     the PR-side Build check cannot be created or reported reliably for this branch, so the contributor still does not get a usable CI signal from the workflow this PR is trying to fix.
   - Best fix:
     do not assume a specific check-run name exists. Match the check/run by the exact `head_sha` and the workflow run you already fetched, or fall back to the `actions_url` directly when the check-run cannot be found.
   - Severity: High

CI
- The latest `On pull request update` run on the current head fails with `TypeError: Cannot read properties of undefined (reading 'head_sha')`.
- The latest `Build` run for the same head is still `startup_failure`, so after fixing Issue 1 you will still need to re-trigger and verify the original backend/website path.

Merge conclusion

1. Blocking items
- Fix Issue 1: remove the hard-coded `Run / License header` lookup and make the PR-side check creation robust on the current head.

- After that fix, re-trigger the current Build run and confirm the backend website-build path is actually green.

Right now the PR is still failing in its own workflow-notification path, so it does not yet close the real CI loop end to end.

DanielLeens (approved):
Thanks for working on this. I re-reviewed the latest head from the PR notification path through the backend website-build path, and I also checked the current CI result on this exact head.

What this PR fixes
- User pain: fork PRs could fail to surface a usable `Build` status when the PR-side notification workflow could not map the backend run back to a stable check URL, and the website build also needed the workflow token path corrected.
- Fix approach: the latest head removes the brittle fixed-name check lookup in `notify_test_workflow.yml`, falls back safely to the Actions run URL, and switches the website build job to the workflow token path that is available in the called workflow.
- One-line summary: this closes the PR CI loop end to end instead of fixing only the website sub-step.

Runtime path I checked
```text
pull_request_target
  -> .github/workflows/notify_test_workflow.yml
      -> list build_main.yml runs for the PR branch
      -> list check-runs for the same head
      -> match the run/check by head SHA and run URL
      -> create the PR-side Build check

backend workflow
  -> .github/workflows/build_main.yml permissions
  -> .github/workflows/backend.yml document job
      -> docusaurus build uses github.token
```

Findings
- The blocker from my previous Daniel review is fixed on this head: the notification workflow no longer assumes a hard-coded `Run / License header` check exists.
- I checked the latest CI on this PR head, and the current `Notify test workflow` plus the downstream `Build` workflow are green, so the changed path is not only theoretically correct but also exercised on the latest branch state.
- I do see that this PR carries a broader GitHub Actions modernization sweep in `backend.yml`, but on the current head it is internally consistent and the current workflow graph passes.

Compatibility
- Fully compatible for project behavior. No user-facing runtime contract changes. The impact is limited to GitHub Actions execution paths.

Merge conclusion
### Conclusion: can merge

1. Blocking items
- None on the current head.

- None required for merge.

I noticed this PR had a previous Daniel change request. After re-reviewing the latest head, I agree with the original direction and the earlier blocker is now closed, so I am fine with merging from my side.

davidzollo (approved):
Good job

PR #11195, opened 2026-06-26
[Improve][Connector-V2][Redis] Align OptionRule with docs and dedupe config validation

ss666:

Before working on the Redis connector changes for #11007, I reviewed the connector and found the source/sink `OptionRule`s were out of sync with the docs and with the options each side actually reads. This small, self-contained PR fixes them ahead. No new feature behavior, no public option key/default change.

main changes:
1. `RedisSinkFactory`: add `db_num`, `batch_size`, `field_delimiter` to the sink rule — they are documented and read when building the sink, but were missing.
2. `RedisSourceFactory`: remove sink-only `key` (`KEY`) that the source never reads.
3. `RedisTableConfig`: merge the two duplicate required-field validators into one applied uniformly to single-table and `tables_configs` modes, and merge the two ~95%-identical builders into one `buildTableConfig`.
4. add `RedisTableConfigTest` covering required-field validation and table building on both single-table and multi-table modes.

No runtime/job behavior change. Conforming configs that follow the docs are unaffected.

Scoped to the `connector-redis` module (the only module this PR changes):
- `./mvnw spotless:apply` pass.
- `./mvnw test -pl seatunnel-connectors-v2/connector-redis` → 28/28, 0 failures

### Check list

  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)


DanielLeens (approved):
Thanks for the contribution. I reviewed the full current diff from the actual Redis runtime path instead of only the unit test changes.

What this PR fixes
- User pain: Redis source/sink option validation and the in-memory table-config parsing path had drifted apart, so invalid configs could be accepted in one entrypoint and rejected in another.
- Fix approach: this head consolidates the source-side table parsing into one builder, tightens required-field validation for blank `keys`, and aligns the source/sink `OptionRule` declarations with the options actually consumed at runtime.
- One-line summary: this makes the Redis connector reject bad configs earlier and more consistently.

Runtime path I checked
```text
Redis source
  -> RedisSourceFactory.optionRule()
  -> RedisSource(ReadonlyConfig)
  -> RedisTableConfig.of(...)
  -> buildTableConfig(...)
  -> RedisSourceReader scans matching keys

Redis sink
  -> RedisSinkFactory.optionRule()
  -> RedisParameters.buildWithConfig(...)
  -> RedisSink writes rows using the validated key/data_type path
```

Findings
- The source-side change is on the real normal path because every Redis source goes through `RedisTableConfig.of(...)` before the reader is created.
- The sink-side `OptionRule` now matches the options that `RedisParameters.buildWithConfig(...)` actually reads, so this is not just a cosmetic docs sync.
- The new `RedisTableConfigTest` covers the important regression points here: single-table validation, blank-key rejection, and per-item validation in multi-table mode. The test shape is deterministic and I do not see a flaky pattern.

Compatibility
- Fully compatible. No public option name is renamed, no default value changes, and the only behavior change is earlier rejection of invalid or blank `keys` input that would not produce a usable source config anyway.

Merge conclusion
### Conclusion: can merge

1. Blocking items

- None required for merge.

Overall this looks like a focused cleanup with the right validation coverage, and I am fine with merging it as-is.

davidzollo (approved):
LGTM. Thanks for tightening this up.

I like that this change makes the Redis validation path more consistent across the real runtime entrypoints instead of only improving the surface-level config check. Rejecting invalid or blank `keys` earlier is a good quality-of-life improvement for users, and the scope here stays focused and easy to reason about.

Appreciate the clean follow-through on both validation and test coverage. Thanks again for the contribution, and please keep them coming.

PR #11169, opened 2026-06-23
[Fix][Connector-V2] Restore JDBC catalog database compatibility

DanielLeens:
## Purpose of this pull request

This PR fixes JDBC catalog validation failures that are unrelated to the triggering PR changes but currently break CI jobs such as the SQL Server XA schema evolution workflow.

Two compatibility gaps were causing submission-time validation to reject valid sink/catalog configurations:

- Oracle thin service URLs such as `jdbc:oracle:thin:@host:1521/ORCLCDB` were checked only by the generic JDBC URL parser.
- SQL Server and compatible dialects that provide the database through the explicit `database` option were rejected when the URL itself did not embed a database name.

## Does this pull request introduce _any_ user-facing change?

Yes. Valid JDBC catalog and sink configurations that already provide the database through dialect-specific URL formats or the explicit `database` option now pass validation again.

## How was this patch tested?

Validation run in `/Volumes/disk2/work/myworkspace/tmp/seatunnel-dev-pulsar-multi-table-replica-20260623`:

- `./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-jdbc`
- `./mvnw -pl seatunnel-connectors-v2/connector-jdbc -DskipTests compile`
- Manual focused test compilation:
  - `javac -cp "$(cat /tmp/seatunnel-jdbc-test.cp)" -d /tmp/seatunnel-jdbc-manual-test-classes ...JdbcCatalogFactoryTest.java ...JdbcCatalogUtilsTest.java`
- Focused JUnit execution:
  - `java -jar ~/.m2/repository/org/junit/platform/junit-platform-console-standalone/1.9.0/junit-platform-console-standalone-1.9.0.jar -cp "/tmp/seatunnel-jdbc-manual-test-classes:$(cat /tmp/seatunnel-jdbc-test.cp)" -c org.apache.seatunnel.connectors.seatunnel.jdbc.catalog.JdbcCatalogFactoryTest -c org.apache.seatunnel.connectors.seatunnel.jdbc.utils.JdbcCatalogUtilsTest --details=tree --disable-banner`

Additional repo-path verification attempted:

- `./mvnw -q -pl seatunnel-connectors-v2/connector-jdbc -DskipTests verify`

That `verify` run is still blocked by an unrelated existing `testCompile` failure in `DuckDBSourceAndSinkTest` due missing `org.apache.seatunnel.connectors.seatunnel.sink/source` imports.

[...]

PR #11202, opened 2026-06-28
[Fix][E2E] Wait for Postgres CDC WAL stream readiness

DanielLeens:
## What does this PR do?

This fixes a shared Postgres CDC E2E instability that is unrelated to JDBC catalog compatibility changes in #11169.

## Why was CI still failing?

The original shared-fix direction was only part of the picture. The current CI failure comes from the Debezium JSON Kafka test starting DML as soon as the snapshot row becomes visible, while the WAL stream is still attaching. In that window, intermediate row-level changes can be skipped, which is why CI only observed `3` messages instead of the expected `5`.

The reused Postgres test container can also retain generated replication slots between CDC jobs, so the branch keeps the slot isolation and cleanup changes as well.

## How was this fixed?

- assign a unique `slot.name` for each `PostgresCDCIT` job invocation
- clean up stale generated replication slots between test methods
- wait until the generated replication slot becomes `active` before issuing the Debezium JSON test DML
- keep using the same generated slot across restore flows in the same test

## Verification

- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-postgres-e2e -nsu -Dmaven.gitcommitid.skip=true spotless:apply`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-postgres-e2e -nsu -Dmaven.gitcommitid.skip=true -DskipUT=true -DskipIT=false -Dit.test=PostgresCDCIT -DfailIfNoTests=false -Dskip.spotless=true verify`

The local Maven verification reached the real E2E startup path, compiled the updated test, and then stopped on a local external-environment blocker while Testcontainers tried to pull `debezium/postgres:11` from Docker Hub: `toomanyrequests: You have reached your unauthenticated pull rate limit`. This blocked local container startup before the code path under test could finish, so the final behavior check still relies on GitHub CI.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-postgres-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/PostgresCDCIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-postgres-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/PostgresCDCIT.java
@@ -44,8 +44,10 @@ import org.apache.kafka.common.IsolationLevel;
 import org.apache.kafka.common.TopicPartition;
 
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestTemplate;
 import org.slf4j.Logger;
@@ -125,6 +127,7 @@ public class PostgresCDCIT extends TestSuiteBase implements TestResource {
             "full_types_no_primary_key_with_debezium";
 
     private static final String SOURCE_SQL_TEMPLATE = "select * from %s.%s order by id";
+    private static final String GENERATED_SLOT_PREFIX = "seatunnel_";
 
     // kafka container
     private static final String KAFKA_IMAGE_NAME = "confluentinc/cp-kafka:7.0.9";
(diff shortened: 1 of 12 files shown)


nzw921rx (approved):
+1 LGTM

seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-postgres-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/PostgresCDCIT.java
```diff
@@ -251,6 +254,28 @@ private Properties kafkaConsumerConfig() {
         return props;
     }
 
+    /**
+     * Replication slots are shared inside the reused Postgres test container, so each CDC job needs
+     * an isolated slot to avoid cross-test collisions when streaming jobs overlap.
+     */
+    private String createSlotName() {
+        return "seatunnel_" + Long.toHexString(JobIdGenerator.newJobId());
```
heye1005: use GENERATED_SLOT_PREFIX

seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-postgres-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/PostgresCDCIT.java
```diff
@@ -311,29 +339,40 @@ public void testPostgresCdcWithDebeziumJsonFormat(TestContainer container) {
                                 dataSize.updateAndGet(v -> v + getKafkaData().size());
                                 Assertions.assertEquals(1, dataSize.get());
                             });
-            // insert update delete
-            upsertDeleteSourceTable(POSTGRESQL_SCHEMA, SOURCE_TABLE_NO_PRIMARY_KEY_DEBEZIUM);
-
-            await().atMost(1000 * 60 * 3, TimeUnit.MILLISECONDS)
```
heye1005, line 317: before 3min,after 1min, may cause flaky problems

davidzollo (approved):
LGTM

PR #11205, opened 2026-06-28
[Test][E2E] Stabilize Kafka exactly-once record reads

DanielLeens:
## What does this PR change?

This stabilizes the Kafka exactly-once E2E verification helper used by `KafkaIT`.

The failing CI path was reading sink records by assuming that `expectedCount` committed records must occupy the contiguous offset range `[startOffset, startOffset + expectedCount)`. That assumption is wrong for `read_committed`: aborted transactional offsets can be skipped, so the next visible committed records may span a wider offset range.

This PR changes the helper to:
- snapshot the current visible end offset for the `READ_COMMITTED` consumer
- consume records from `startOffset` until that visible end offset
- keep the test assertion strict on the final visible record count and values

## Why is this needed?

`#11183` exposed a shared failure in:
- `KafkaIT.testKafkaToKafkaExactlyOnceOnStreaming`

That failure is unrelated to the dead-link Node 22 change itself and can block unrelated PRs.

## How was this verified?

- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home ./mvnw spotless:apply -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kafka-e2e -nsu -Dmaven.gitcommitid.skip=true -T 3C`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home ./mvnw -q -DskipTests verify -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kafka-e2e -nsu -Dmaven.gitcommitid.skip=true -T 3C`

I could not run the Testcontainers E2E locally in this worktree because the local Docker daemon was unavailable (`Cannot connect to the Docker daemon at unix:///Users/stone/.docker/run/docker.sock`).

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kafka-e2e/src/test/java/org/apache/seatunnel/e2e/connector/kafka/KafkaIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kafka-e2e/src/test/java/org/apache/seatunnel/e2e/connector/kafka/KafkaIT.java
@@ -1689,6 +1689,7 @@ public class KafkaIT extends TestSuiteBase implements TestResource {
         createKafkaTopic(producerTopic);
         createKafkaTopic(consumerTopic);
         String sourceData = "Seatunnel Exactly Once Example";
+        String keepAliveData = sourceData + "-keepalive-" + resourceSuffix;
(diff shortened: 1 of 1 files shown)


davidzollo (approved):
LGTM

PR #11213, opened 2026-06-29
[Test][E2E] Retry SQL Server XA setup until login is ready

DanielLeens:
## What

- Retry each SQL Server XA setup `sqlcmd` step until the command actually succeeds
- Fail fast when `execInContainer` keeps returning non-zero instead of silently continuing

## Why

SQL Server 2022 can print the generic ready log before system database upgrades finish. In that window, the `sa` login returns error `18456`, so the XA setup commands may fail silently — and later schema-change integration tests never reach the expected running state.

## Verification

```bash
./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl \
  -nsu spotless:apply
```

```bash
./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-ddl \
  -nsu -DskipTests test-compile
```

> **Note:** Local Docker is unavailable in this environment, so the full Testcontainers E2E flow was not rerun locally.


davidzollo (approved):
+1

PR #11214, opened 2026-06-29
[Improve][Connector-V2][Socket] Migrate Socket Sink validation to declarative OptionRule

nzw921rx:

Related #11007

Migrate Socket Sink configuration validation from imperative runtime checks to declarative `OptionRule` constraints.

**Changes**:
- Add `Conditions.greaterThan` validation for `port` and `max_retries` in `SocketSinkFactory`
- Remove the redundant `max_retries == 0` fast-fail branch from `SocketClient`, since `max_retries` must now be greater than 0 when explicitly configured
- Add `SocketFactoryTest` cases covering valid and invalid `port` / `max_retries` values

Yes. Invalid Socket Sink configs (`port <= 0`, or `max_retries <= 0` when set) are now rejected during configuration validation instead of failing later at runtime. `max_retries = 0` is no longer accepted.

```bash
./mvnw -pl seatunnel-connectors-v2/connector-socket -am install -DskipTests -Dskip.spotless=true
./mvnw -pl seatunnel-connectors-v2/connector-socket test -Dtest=SocketFactoryTest -Dskip.spotless=true
```


DanielLeens (changes requested):
Thanks for the contribution. I pulled the latest head locally and reviewed the full current diff from the real Socket sink runtime path, not just the factory diff.

What this PR is trying to do
- User pain: invalid Socket sink configs such as `port <= 0` or bad retry settings currently fail later in the runtime path.
- Fix approach: move part of that validation into `SocketSinkFactory.optionRule()` and add validator-based tests.
- One-line summary: the direction is good, but the current head changes the existing `max_retries = 0` contract instead of only moving validation earlier.

Runtime path I checked
```text
job config parsing
  -> SocketSinkFactory.optionRule() [SocketSinkFactory.java:38-45]
      -> validates host / port / max_retries before the sink is created

sink creation
  -> SocketSinkFactory.createSink(...) [SocketSinkFactory.java:49-50]
      -> new SocketSink(context.getOptions(), ...)
      -> new SocketConfig(ReadonlyConfig) [SocketConfig.java:32-35]
          -> reads MAX_RETRIES

write-time failure path
  -> SocketSinkWriter.write(...)
      -> SocketClient.write(...) [SocketClient.java:74-142]
          -> old behavior had an explicit `maxNumRetries == 0` fail-fast branch
          -> current PR removes that branch and also rejects `0` during config validation
```

Key findings
1. The normal path definitely hits this change, because every Socket sink job goes through `optionRule()` before writer creation.
2. `port > 0` is a reasonable declarative validation improvement.
3. `max_retries > 0` is not just “moving validation earlier” though. The old runtime explicitly allowed `max_retries = 0` to mean “do not retry; fail immediately on the first write failure”.
4. The new test in `SocketFactoryTest.java:83-93` now codifies that old valid value as invalid, so this is a real behavior regression, not only a docs mismatch.

Blocking issues

Issue 1: the PR removes the existing fail-fast contract for `max_retries = 0`
- Location: `seatunnel-connectors-v2/connector-socket/src/main/java/org/apache/seatunnel/connectors/seatunnel/socket/sink/SocketSinkFactory.java:41-44`, `seatunnel-connectors-v2/connector-socket/src/test/java/org/apache/seatunnel/connectors/seatunnel/socket/SocketFactoryTest.java:83-93`
- Why this is a problem:
  the old Socket sink semantics allowed `max_retries = 0` and handled it explicitly in `SocketClient.write(...)` as “retry is disabled”. The current head rejects that value at validation time and deletes the runtime branch, so an existing user config can no longer be submitted at all.
- Risk:
  this is a real backward-compatibility break for existing Socket sink jobs that intentionally use `max_retries = 0`.
- Best fix:
  option A (recommended): keep the old contract and allow `max_retries >= 0`, while preserving the fail-fast `maxNumRetries == 0` branch in `SocketClient.write(...)`.
  option B: if the community intentionally wants to deprecate `0`, then treat it as an incompatible change and document the migration explicitly instead of hiding it inside a validation cleanup PR.
- Severity: High

Issue 2: if this behavior change is intentional, the user-facing docs and incompatible-change note are still missing
- Location: `docs/en/connectors/sink/Socket.md:25-30`, `docs/zh/connectors/sink/Socket.md:26-31`
- Why this is a problem:
  the PR description already says that `max_retries = 0` is now rejected, but the Socket sink docs in both languages still do not document that contract change, and there is no incompatible-change record either.
- Risk:
  users will keep following the current docs and only discover the change when config validation starts failing.
- Best fix:
  if Issue 1 is fixed by restoring `0`, this problem disappears naturally. Otherwise please update both docs and add an incompatible-change note.
- Severity: Medium

Test coverage / stability
- The added tests are structurally stable: no sleeps, no ports, no timing dependency.
- Stability rating: Stable.
- But the current coverage is asserting the wrong contract for `max_retries = 0`, so the problem here is semantic correctness rather than flakiness.

CI note
- I did not do a local Maven run in this batch; this review is based on the real diff, the current source path, and the current PR metadata.

1. Blocking items
- Fix Issue 1 first: please preserve the historical `max_retries = 0` fail-fast behavior, or explicitly handle it as a documented incompatible change.

2. Suggested follow-up
- If you intentionally keep the behavior change, also fix Issue 2 by updating both the English and Chinese Socket sink docs and the incompatible-change note.

Overall, I agree with moving validation into `OptionRule`, but the current version mixes that cleanup with a user-visible contract change. Once the `max_retries = 0` compatibility issue is resolved, the rest of the direction looks good.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from scratch against the full current diff and re-traced the real Socket sink path again instead of looking only at the new commit metadata.

- User pain: invalid Socket sink configs such as `port <= 0` should fail earlier instead of surfacing later in the runtime path.
- Fix approach: move part of the validation into `SocketSinkFactory.optionRule()` and add validator-based tests.
- One-line summary: the declarative-validation direction is good, but the current head still changes the historical `max_retries = 0` contract instead of only moving validation earlier.

## Full runtime path I rechecked
```text
job config parsing
  -> SocketSinkFactory.optionRule()
      -> host required
      -> port must be > 0
      -> max_retries must be > 0   <-- current new gate

sink creation
  -> createSink(...)
      -> new SocketSink(...)
      -> new SocketConfig(ReadonlyConfig)
          -> reads MAX_RETRIES

write-time failure path
  -> SocketSinkWriter.write(...)
      -> SocketClient.write(...)
          -> old behavior:
               max_retries == 0 meant "do not retry; fail immediately"
          -> current behavior:
               zero is rejected before the sink is even created
```

## Key findings
1. The normal path definitely hits this change, because every Socket sink job goes through `optionRule()` before writer creation.
2. `port > 0` is a reasonable declarative-validation improvement.
3. The compatibility blocker from my previous Daniel review is still present on the latest head:
   - `SocketSinkFactory.java:41-44` still uses `Conditions.greaterThan(MAX_RETRIES, 0)`
   - `SocketClient.java:77-89` no longer keeps the old `maxNumRetries == 0` fail-fast branch
   - `SocketFactoryTest.java:82-93` still codifies `max_retries = 0` as invalid
4. So the latest head still turns a historically valid config into a submission-time validation failure. That is a user-visible contract change, not just a validation cleanup.

## Findings

### Issue 1: the PR still removes the historical fail-fast contract for `max_retries = 0`
- Location:
  - `seatunnel-connectors-v2/connector-socket/src/main/java/org/apache/seatunnel/connectors/seatunnel/socket/sink/SocketSinkFactory.java:41-44`
  - `seatunnel-connectors-v2/connector-socket/src/main/java/org/apache/seatunnel/connectors/seatunnel/socket/sink/SocketClient.java:77-89`
  - `seatunnel-connectors-v2/connector-socket/src/test/java/org/apache/seatunnel/connectors/seatunnel/socket/SocketFactoryTest.java:82-93`
- Why this is a problem:
  the old Socket sink semantics allowed `max_retries = 0` and handled it explicitly as “retry is disabled; fail immediately on the first write failure”. The current head still rejects that value during config validation and removes the runtime branch, so an existing user config can no longer be submitted at all.
- Risk:
  this is a real backward-compatibility break for existing Socket sink jobs that intentionally use `max_retries = 0`.
- Best fix:
  - Option A (recommended): keep the old contract and allow `max_retries >= 0`, while restoring the fail-fast `maxNumRetries == 0` branch in `SocketClient.write(...)`.
  - Option B: if the community intentionally wants to deprecate `0`, then treat it as an explicit incompatible change and document the migration instead of hiding it inside a validation-cleanup PR.
- Severity: High

### Issue 2: if this behavior change is intentional, the user-facing docs and incompatible-change note are still missing
- Location:
  - `docs/en/connectors/sink/Socket.md:27-30`
  - `docs/zh/connectors/sink/Socket.md:28-31`
- Why this is a problem:
  the current Socket sink docs still do not say that `max_retries = 0` is rejected, and there is still no incompatible-change record for that behavior change.
- Risk:
  users will keep following the current docs and only discover the change when validation starts failing at submission time.
- Best fix:
  if Issue 1 is fixed by restoring `0`, this problem disappears naturally. Otherwise please update both docs and add an incompatible-change note.
- Severity: Medium

## Tests / stability
- The added tests are structurally stable: no sleeps, no ports, no timing dependency.
- Stability rating: Stable.
- The problem here is semantic correctness, not flakiness: the current tests are asserting the wrong contract for `max_retries = 0`.

## CI note
- The current GitHub state I saw on this latest head was:
  - `Notify test workflow`: pass
  - `labeler`: pass
  - `Build`: in progress

The source-level compatibility blocker above is already enough for me regardless of the final Build result.

1. Blocking items
- Issue 1: please preserve the historical `max_retries = 0` fail-fast behavior, or explicitly handle it as a documented incompatible change.

- Issue 2: if you intentionally keep the behavior change, also update both the English and Chinese Socket sink docs and the incompatible-change note.

Overall, I still agree with moving Socket sink validation into `OptionRule`, but the current version still mixes that cleanup with a user-visible contract change. Once the `max_retries = 0` compatibility issue is resolved, the rest of the direction looks good.

nzw921rx:
> Thanks for the update. I re-reviewed the latest head from scratch against the full current diff and re-traced the real Socket sink path again instead of looking only at the new commit metadata.
> 
> ## What this PR is trying to solve
> * User pain: invalid Socket sink configs such as `port <= 0` should fail earlier instead of surfacing later in the runtime path.
> * Fix approach: move part of the validation into `SocketSinkFactory.optionRule()` and add validator-based tests.
> * One-line summary: the declarative-validation direction is good, but the current head still changes the historical `max_retries = 0` contract instead of only moving validation earlier.
> 
> ## Full runtime path I rechecked
> ```
> job config parsing
>   -> SocketSinkFactory.optionRule()
>       -> host required
>       -> port must be > 0
>       -> max_retries must be > 0   <-- current new gate
> 
> sink creation
>   -> createSink(...)
>       -> new SocketSink(...)
>       -> new SocketConfig(ReadonlyConfig)
>           -> reads MAX_RETRIES
> 
> write-time failure path
>   -> SocketSinkWriter.write(...)
>       -> SocketClient.write(...)
>           -> old behavior:
>                max_retries == 0 meant "do not retry; fail immediately"
>           -> current behavior:
>                zero is rejected before the sink is even created
> ```
> 
> ## Key findings
> 1. The normal path definitely hits this change, because every Socket sink job goes through `optionRule()` before writer creation.
> 2. `port > 0` is a reasonable declarative-validation improvement.
> 3. The compatibility blocker from my previous Daniel review is still present on the latest head:
>    
>    * `SocketSinkFactory.java:41-44` still uses `Conditions.greaterThan(MAX_RETRIES, 0)`
>    * `SocketClient.java:77-89` no longer keeps the old `maxNumRetries == 0` fail-fast branch
>    * `SocketFactoryTest.java:82-93` still codifies `max_retries = 0` as invalid
> 4. So the latest head still turns a historically valid config into a submission-time validation failure. That is a user-visible contract change, not just a validation cleanup.
> 
> ## Findings
> ### Issue 1: the PR still removes the historical fail-fast contract for `max_retries = 0`
> * Location:
>   
>   * `seatunnel-connectors-v2/connector-socket/src/main/java/org/apache/seatunnel/connectors/seatunnel/socket/sink/SocketSinkFactory.java:41-44`
>   * `seatunnel-connectors-v2/connector-socket/src/main/java/org/apache/seatunnel/connectors/seatunnel/socket/sink/SocketClient.java:77-89`
>   * `seatunnel-connectors-v2/connector-socket/src/test/java/org/apache/seatunnel/connectors/seatunnel/socket/SocketFactoryTest.java:82-93`
> * Why this is a problem:
>   the old Socket sink semantics allowed `max_retries = 0` and handled it explicitly as “retry is disabled; fail immediately on the first write failure”. The current head still rejects that value during config validation and removes the runtime branch, so an existing user config can no longer be submitted at all.
> * Risk:
>   this is a real backward-compatibility break for existing Socket sink jobs that intentionally use `max_retries = 0`.
> * Best fix:
>   
>   * Option A (recommended): keep the old contract and allow `max_retries >= 0`, while restoring the fail-fast `maxNumRetries == 0` branch in `SocketClient.write(...)`.
>   * Option B: if the community intentionally wants to deprecate `0`, then treat it as an explicit incompatible change and document the migration instead of hiding it inside a validation-cleanup PR.
> * Severity: High
> * Raised by others already: No
> 
> ### Issue 2: if this behavior change is intentional, the user-facing docs and incompatible-change note are still missing
> * Location:
>   
>   * `docs/en/connectors/sink/Socket.md:27-30`
>   * `docs/zh/connectors/sink/Socket.md:28-31`
> * Why this is a problem:
>   the current Socket sink docs still do not say that `max_retries = 0` is rejected, and there is still no incompatible-change record for that behavior change.
> * Risk:
>   users will keep following the current docs and only discover the change when validation starts failing at submission time.
> * Best fix:
>   if Issue 1 is fixed by restoring `0`, this problem disappears naturally. Otherwise please update both docs and add an incompatible-change note.
> * Severity: Medium
> * Raised by others already: No
> 
> ## Tests / stability
> * The added tests are structurally stable: no sleeps, no ports, no timing dependency.
> * Stability rating: Stable.
> * The problem here is semantic correctness, not flakiness: the current tests are asserting the wrong contract for `max_retries = 0`.
> 
> ## CI note
> * The current GitHub state I saw on this latest head was:
>   
>   * `Notify test workflow`: pass
>   * `labeler`: pass
>   * `Build`: in progress
> 
> The source-level compatibility blocker above is already enough for me regardless of the final Build result.
> 
> ### Merge conclusion: can merge after fixes
> 1. Blocking items
> 
> * Issue 1: please preserve the historical `max_retries = 0` fail-fast behavior, or explicitly handle it as a documented incompatible change.
> 
> 2. Suggested non-blocking follow-up
> 
> * Issue 2: if you intentionally keep the behavior change, also update both the English and Chinese Socket sink docs and the incompatible-change note.
> 
> Overall, I still agree with moving Socket sink validation into `OptionRule`, but the current version still mixes that cleanup with a user-visible contract change. Once the `max_retries = 0` compatibility issue is resolved, the rest of the direction looks good.

@DanielLeens thanks for your review.

Before migration, if it is equal to 0, it will report an error during runtime. It should be thrown to the configuration verification in advance, and the user should be aware of it clearly. This is a fix for abnormal scenarios and does not involve compatibility issues.

Before code:

``` java
if (maxNumRetries == 0) {
                throw new SocketConnectorException(
                        SocketConnectorErrorCode.SEND_MESSAGE_TO_SOCKET_SERVER_FAILED,
                        String.format(
                                "Failed to send message '%s' to socket server at %s:%d. Connection re-tries are not enabled.",
                                row, hostName, port),
                        e);
            }
```

DanielLeens:
Thanks for the clarification. I rechecked the latest head and I still see this as a backward-compatibility issue.

Before this PR, `SocketClient.write(...)` explicitly accepted `max_retries = 0` as the fail-fast path: the job could still be submitted, and the connector would fail immediately on the first write error without retrying. On the current head, `SocketSinkFactory.optionRule()` rejects `0` before the sink is even created. So an existing config that used to run now fails at submission time. That is a user-visible contract change even if the eventual runtime outcome was still failure.

If the goal is to keep this migration non-breaking, I still recommend preserving `max_retries >= 0` and restoring the old fail-fast zero branch. If the community intentionally wants to deprecate `0`, then it needs to be treated as an explicit incompatible change rather than hidden inside a validation cleanup.

Also, this branch is currently behind the latest `dev` (`behind_by=10`, compare status: `diverged`) and the latest `Build` is still pending. After the code update, please sync the latest `dev` and rerun CI as well.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head `8e87a927db` specifically against the `max_retries = 0` compatibility point from the earlier Daniel rounds.

## What this PR solves
- User pain: the socket sink accepted some invalid numeric configs too late, but we also could not break the long-standing meaning of `max_retries = 0`.
- Fix approach: move the sink-side numeric validation into `SocketSinkFactory.optionRule()` while keeping the runtime reconnect loop semantics intact in `SocketClient.write()`.
- One-line summary: the current head tightens submission-time validation without regressing the runtime meaning of `max_retries = 0`.

## Runtime chain I rechecked
```text
Job submission
  -> SocketSinkFactory.optionRule() [SocketSinkFactory.java:37-46]
      -> HOST required
      -> PORT > 0
      -> MAX_RETRIES >= 0

Runtime write path
  -> SocketClient.write(row) [SocketClient.java:74-143]
      -> write to output stream
      -> on failure enter reconnect path
      -> while (maxNumRetries < 0 || retries < maxNumRetries) [90]
```

## Findings
- The earlier Daniel blocker is closed for me now.
- `max_retries = 0` is still allowed at submission time because the new rule is `greaterOrEqual(..., 0)` [SocketSinkFactory.java:42-45].
- The runtime behavior is still compatible as well:
  - `SocketClient.write()` resets `retries = 0` and only enters the reconnect loop while `retries < maxNumRetries` [SocketClient.java:89-90]
  - when `maxNumRetries == 0`, the loop is skipped and the method throws immediately with “Failed after 0 retries” [SocketClient.java:133-139]
- So this revision improves the validation boundary without changing the existing zero-retry runtime contract.

## Tests / stability
- Stability rating: Stable.
- `SocketFactoryTest` covers the important submit-time cases:
  - valid config
  - `port = 0`
  - negative port
  - negative `max_retries`
- These are in-memory validator tests only, with no timing or environment-sensitive behavior [SocketFactoryTest.java:61-87].

## CI / sync snapshot
- The latest CI checks are green on the current head.
- This branch is still `behind_by=1` against `dev`, so I am recording that merge-state fact in the queue, but I do not see a current source-side blocker from Daniel anymore.

## Merge conclusion
### Conclusion: can merge

1. Blocking items
- None from the current source path.

- None from this round.

Overall, this now looks good to me. The validation is stricter where it should be, and the historical `max_retries = 0` behavior is still preserved on the real runtime path.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head from scratch against the current full diff, including the Socket sink config-validation path and the write-time retry path.

First, the important good news: the previous source-level compatibility blocker is mostly addressed now. `SocketSinkFactory.optionRule()` allows `max_retries >= 0`, so an existing config with `max_retries = 0` is no longer rejected during option validation. That is the right direction.

## Full runtime path I rechecked

```text
Socket sink job submission
  -> SocketSinkFactory.optionRule()
      -> host required
      -> port must be > 0
      -> max_retries must be >= 0

Sink creation
  -> SocketSinkFactory.createSink(...)
      -> new SocketSink(context.getOptions(), context.getCatalogTable())
          -> new SocketConfig(ReadonlyConfig)
              -> reads host / port / max_retries

Writer startup
  -> SocketSink.createWriter(...)
      -> new SocketSinkWriter(socketConfig, rowType)
          -> new SocketClient(socketConfig, JsonSerializationSchema)
          -> socketClient.open()

Write failure path
  -> SocketSinkWriter.write(row)
      -> SocketClient.write(row)
          -> outputStream.write/flush throws IOException e
          -> maxNumRetries == 0
          -> retry loop is skipped
          -> throws SocketConnectorException(..., lastException)
             where lastException is still null
```

## Findings

### Issue 1: `max_retries = 0` now loses the original IOException cause on write failure

* Location: `seatunnel-connectors-v2/connector-socket/src/main/java/org/apache/seatunnel/connectors/seatunnel/socket/sink/SocketClient.java:87-139`
* Why this is a problem: after the latest update, `max_retries = 0` is accepted by validation, which is good. But the old explicit fail-fast branch was removed. With `maxNumRetries == 0`, the retry loop is skipped, `lastException` stays `null`, and the thrown `SocketConnectorException` no longer carries the original write `IOException` as its cause. The previous implementation preserved that cause in the zero-retry path.
* Risk: users still get a task failure, but the propagated exception chain loses the real socket-write failure reason, making production debugging weaker than before.
* Best fix:
  * Option A (recommended): restore the explicit `maxNumRetries == 0` fail-fast branch and keep passing the original `IOException e` as the cause.
  * Option B: at least initialize `lastException` with `e`, so the zero-retry path never throws with a null cause.
* Severity: Medium
* Raised by others already: No. This is related to my earlier point about preserving the historical `max_retries = 0` fail-fast semantics, but it is specifically about the current head's remaining error-handling behavior.

### Test coverage note

The new validator tests are stable: they do not use sleeps, ports, external services, timing assumptions, or shared state. Stability rating: stable.

However, the key compatibility case still needs a regression assertion: please add a test that `max_retries = 0` passes option validation. If you restore the fail-fast branch, a small unit test that protects the original exception cause would also be very useful.

## CI / merge gate note

On the current head I saw:

* `Notify test workflow`: success
* `labeler`: success
* `Build`: still not completed in the rollup I checked

This branch is now synced with the latest `dev` for compare purposes (`behind_by=0`, compare status: `ahead`), so once the source fix above is pushed, the remaining CI signal should be much cleaner to interpret.

1. Blocking items

* Issue 1: please preserve the original `IOException` cause for the `max_retries = 0` fail-fast write path.
* Please add the missing `max_retries = 0` validation regression test so this compatibility contract does not regress again.

* No extra non-blocking source suggestions from me on this round.

Overall, the declarative validation migration looks good after the latest update. Once the zero-retry error-handling path and its regression test are tightened up, I think the source side will be in good shape, with only CI / merge gates left to watch.

nzw921rx:
@davidzollo can you help me review this PR? thanks🚀

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest Socket sink validation changes.

# What this PR fixes
The PR migrates Socket sink validation into the declarative `OptionRule` path and adds regression coverage for invalid port / retry values.

```text
Socket sink config path
  -> SocketSinkFactory.optionRule()
      -> host required
      -> port must be > 0
      -> max_retries must be >= 0
  -> createSink() builds SocketSink with validated options
  -> E2E writes FakeSource JSON rows to a real socket server and awaits output rows
```

# Current review findings
- The previous zero/negative option validation blocker is resolved.
- The E2E uses Awaitility against actual socket output instead of a fixed sleep, so I do not see a high-risk flaky-test pattern in this round.
- Compatibility: no API/config/default/protocol/serialization break found in this reviewed path.
- CI: `Build` is now `SUCCESS`.
- Base sync fact recorded: compare status `diverged`, `behind_by=16`; since Build is green, I am not turning that into a sync request.

# Merge conclusion
### Conclusion: can merge

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

PR #11216, opened 2026-06-29
[Feature][Connector-V2][Milvus] Support scalar field write null

loupipalien:

This PR adds nullable field support for the Milvus sink connector. (refer [milvus-nullable](https://milvus.io/docs/v2.5.x/nullable-and-default.md))

Yes. This PR introduces a new optional Milvus sink config:
```
enable_nullable_field = true
```
The default value is false, so existing jobs keep the old behavior and still fail on null values with MILVUS-10.
When enabled, SeaTunnel creates supported nullable Milvus fields and allows writing null values only when the corresponding SeaTunnel column is nullable and the field is used as a Milvus scalar field

Added unit tests and milvus e2e case

### Check list


DanielLeens (approved):
Thanks for the contribution. I went through the full diff and traced the actual runtime path end to end.

What this PR fixes in the real path
- `FakeSource` configured `rows` do not bypass parsing. They go through `FakeDataGenerator -> JsonDeserializationSchema`, so the missing `FLOAT_VECTOR` handling in `JsonToRowConverters` was a real gap on the normal path, not a theoretical edge case.
- On the sink side, the nullable behavior is now wired into both schema creation and row writing: `MilvusCatalog.createTableInternal(...) -> MilvusSinkConverter.convertToFieldType(...)` controls the Milvus field schema, and `MilvusSinkConverter.buildMilvusData(...)` uses the same rule set when it sees a `null` field value.
- The guardrails are still intact: primary keys and vector fields remain non-nullable, so this does not widen the contract beyond Milvus scalar fields.

Runtime path I checked
```text
FakeSource rows config
  -> FakeDataGenerator(jsonDeserializationSchema)
      -> JsonDeserializationSchema.deserialize(...)
          -> JsonToRowConverters.FLOAT_VECTOR
              -> JSON array -> ByteBuffer

SeaTunnelRow reaches Milvus sink
  -> MilvusSinkConverter.buildMilvusData(...)
      -> nullable scalar + enable_nullable_field=true
          -> write JsonNull
      -> FLOAT_VECTOR ByteBuffer
          -> convert to List<Float> for Milvus JSON payload

Auto-create schema path
  -> MilvusSink.getSaveModeHandler()
      -> MilvusCatalog.createTableInternal(...)
          -> convertToFieldType(..., enable_nullable_field)
              -> mark nullable scalar fields as nullable in Milvus

Schema round-trip
  -> MilvusSourceConverter.convertColumn(...)
      -> preserve fieldSchema.getNullable()
```

Key findings
- The fix is coherent across the parse path, the write path, and the auto-create schema path.
- The new E2E actually exercises the important scenario: FakeSource row config with a float-vector JSON array plus a nullable scalar field written into Milvus.
- I do not see a compatibility regression here. The new option is opt-in and defaults to `false`, so existing jobs keep the old behavior.
- I also do not see a flaky-test pattern in the added tests. The new coverage uses focused unit tests plus a real Milvus E2E on the exact path above.

Merge conclusion
### Conclusion: can merge

1. Blocking items

- No code-side follow-up required for this PR.

Overall, this looks like a precise fix rather than a local workaround. The only thing still outstanding at the time of review is the GitHub `Build` check, which is currently queued/pending rather than failing. From the code-review side, I'm happy to approve.

DanielLeens:

- base branch: `dev`
- head SHA: `98ea68305caf106821bed94f89893fe62df9e538`
- compare status: `diverged`
- ahead_by: `1`
- behind_by: `2`
- mergeable_state: `unknown`
- CI / merge gate snapshot: `pending` (pending: `Build`)

DanielLeens:

- base branch: `dev`
- head SHA: `98ea68305caf106821bed94f89893fe62df9e538`
- compare status: `diverged`
- ahead_by: `1`
- behind_by: `10`

DanielLeens (approved):
Thanks for the follow-up update. I re-reviewed the latest head from scratch again instead of carrying over my earlier approval, because the current revision adds one more Milvus catalog change on top of the previously approved nullable-field work.

## What this PR fixes
- User pain: Milvus sink could not write nullable scalar fields even when the upstream SeaTunnel schema marked those fields as nullable.
- Fix approach: add an opt-in `enable_nullable_field` contract, preserve nullable metadata on the source side, emit `JsonNull` for nullable scalar sink fields, and cover the path with unit tests plus a Milvus E2E.
- One-line summary: the nullable-field feature still looks correct on the current head, and the new `tableExists()` guard is a reasonable companion fix for the auto-create path rather than a new blocker.

## Runtime path I rechecked
```text
FakeSource / JSON row input
  -> JsonToRowConverters.FLOAT_VECTOR
      -> vector JSON array -> ByteBuffer for Milvus path

Milvus sink auto-create path
  -> MilvusSinkFactory.optionRule(...) [MilvusSinkFactory.java:43-60]
      -> expose enable_nullable_field
  -> MilvusCatalog.tableExists(...) [MilvusCatalog.java:160-171]
      -> databaseExists(database)
      -> missing database => return false before hasCollection()
  -> MilvusCatalog.createTableInternal(...) [MilvusCatalog.java:302-328]
      -> convertToFieldType(..., enableNullableField)
      -> nullable scalar fields become nullable Milvus fields

Milvus row write path
  -> MilvusSinkConverter.buildMilvusData(...) [MilvusSinkConverter.java:292-336]
      -> nullable scalar + enable_nullable_field=true => JsonNull
      -> primary key / vector fields still reject null
```

## Findings
- The new `tableExists()` guard is aligned with the same auto-create path exercised by the nullable-field E2E: when the target database does not exist yet, the catalog now returns `false` instead of calling `hasCollection()` on a missing database.
- The added unit test and the Milvus E2E cover that new branch directly, so this is not an untested side fix.
- I rechecked the nullable-field guardrails as well: primary keys and vector fields are still kept non-nullable, and the option remains opt-in.
- I did not find a new source-side blocker on the current head.

## Merge conclusion
### Conclusion: can merge

1. Blocking items
- None from my side on the latest revision.

2. Suggested follow-up
- The top-level `Build` check is still queued, so CI still needs to finish, but I do not see a remaining code-side blocker.

Overall, this still looks like a precise Milvus contract fix rather than a broad behavior change, and the extra catalog guard is consistent with the real auto-create runtime path.

DanielLeens (approved):
Thanks for the follow-up update. I re-reviewed the latest head from scratch because this revision adds one more Milvus catalog guard on top of the previously approved nullable-field flow.

## What this PR fixes
- User pain: Milvus sink could still fail on the normal auto-create path when the target database did not exist yet, even though the nullable scalar-field work itself was already wired correctly.
- Fix approach: keep the nullable-field contract as-is, and make `MilvusCatalog.tableExists(...)` return `false` early when the database is absent before calling `hasCollection(...)`.
- One-line summary: the new catalog guard is aligned with the same nullable-field runtime path and I do not see a new code-side blocker on the current head.

## Runtime path I rechecked
```text
Job config
  -> MilvusSinkFactory.optionRule(...)
      -> expose enable_nullable_field as the opt-in contract

Auto-create path
  -> MilvusSink.getSaveModeHandler()
      -> MilvusCatalog.createTable(...)
          -> MilvusCatalog.tableExists(...) [MilvusCatalog.java:162-180]
              -> databaseExists(...)
              -> missing database => return false before hasCollection(...)
          -> createTableInternal(...)

Row write path
  -> MilvusSinkConverter.buildMilvusData(...) [MilvusSinkConverter.java:286-337]
      -> nullable scalar + enable_nullable_field=true => JsonNull
      -> primary-key / vector fields still reject null
```

## Current-head findings
- The new `tableExists()` guard at `seatunnel-connectors-v2/connector-milvus/src/main/java/org/apache/seatunnel/connectors/seatunnel/milvus/catalog/MilvusCatalog.java:162-180` is consistent with the real auto-create path. It avoids probing a collection inside a database that does not exist yet, which is exactly the path hit before `createTableInternal(...)`.
- The nullable-field writer guardrails are still intact at `seatunnel-connectors-v2/connector-milvus/src/main/java/org/apache/seatunnel/connectors/seatunnel/milvus/utils/sink/MilvusSinkConverter.java:286-337`: only nullable scalar fields can emit `JsonNull`, while PK/vector fields still fail fast.
- The added coverage is still on the right path. I do not see a new flaky-test pattern in the latest head: the change is a deterministic catalog branch plus the existing focused Milvus coverage.
- I rechecked compatibility as well: the option remains opt-in and existing jobs keep the previous behavior unless `enable_nullable_field=true` is set.

## Merge conclusion
### Conclusion: can merge

1. Blocking items

2. Suggested follow-up
- GitHub `Build` is still in progress, so CI still needs to finish, but I do not see a remaining source-side blocker.

Overall this still looks like a precise Milvus contract fix rather than a broad behavior change, and the extra catalog guard fits the actual auto-create lifecycle cleanly.

zhangshenghang:
Thanks for this PR — the core design (opt-in flag, default `false`, PK/vector exclusion, E2E null round-trip assertions) is solid. I have a few concerns to address before this can merge.

## Critical

**1. Default-config backward-compat break for null dynamic-field values**

In `MilvusSinkConverter#buildMilvusData`, the branch order changed: the dynamic-field branch now runs *before* the null check. The old code threw `FIELD_IS_NULL` (MILVUS-10) on any null value (including the dynamic field) because the null check came first. With the new ordering, a null value in the dynamic-field column is silently skipped (`if (value != null) {...} continue;`) — even under the **default** config (`enable_dynamic_field=true`, `enable_nullable_field=false`).

So a job that previously crashed on a row whose dynamic-field column was null now silently writes that row with no dynamic data. The PR description states "default value is false, so existing jobs keep the old behavior and still fail on null values with MILVUS-10" — that's not accurate for this case, and `skipsNullDynamicFieldBeforeNullValidation` codifies the changed behavior.

Per SeaTunnel's backward-compat policy, please either restore the throw when `enable_nullable_field=false`, or document the change in `docs/en/introduction/concepts/incompatible-changes.md`. (Note: when `enable_dynamic_field=false` the behavior is preserved — the gap is specifically the default `enable_dynamic_field=true`.)

## Important

**2. Partition key is not excluded from nullable**

`supportNullableField` excludes the primary key and vector types, but not the partition key. `convertToFieldType` already receives `partitionKeyField`, so the fix is one line. In Milvus 2.5 the partition key can't be nullable (it's required for routing); with `enable_nullable_field=true` and a nullable partition-key column, `createCollection` will fail with a confusing error instead of a clean MILVUS-10. I'd suggest adding a null-safe `!column.getName().equals(partitionKeyField)` check. No test covers this case today.

**3. Docs don't state the Milvus 2.5+ requirement**

Nullable/default is a Milvus 2.5 feature (the PR links to the v2.5.x docs and the E2E image is bumped to `v2.5.11`). A user on Milvus 2.4 setting `enable_nullable_field=true` will hit `createCollection` failures. The en/zh option rows match the code (name/default/type/description look good), but please add a "Requires Milvus >= 2.5" note to both docs.

**4. Scope — this PR bundles several distinct concerns**

Per "one PR should solve one problem":

- `MilvusCatalog.tableExists` returning `false` when the database is missing looks unrelated to the nullable feature (`createTable` already checks `databaseExists` first), and it adds a `listDatabases` RPC to every `tableExists` call. Consider splitting it into its own PR.
- The E2E image bump `2.4-20240711-...` -> `v2.5.11` is necessary for the nullable test, but it changes the image for *all* existing Milvus E2E tests. Please confirm CI is green across them.
- The new `FLOAT_VECTOR` case in `seatunnel-format-json` `JsonToRowConverters` is a shared-format change only loosely related to "Milvus nullable", and it's incomplete (see #5).

**5. `convertToFloatVector` is incomplete**

It only handles `FLOAT_VECTOR`; `BINARY_VECTOR` / `FLOAT16_VECTOR` / `BFLOAT16_VECTOR` / `SPARSE_FLOAT_VECTOR` still fall through to `UNSUPPORTED_DATA_TYPE`, which is inconsistent. Also `values[i] = convertToFloat(jsonNode.get(i))` will throw a raw `NumberFormatException` on a `NullNode` or non-numeric element without per-element context. Please either cover all vector types or document the limitation, and add a unit test in `seatunnel-format-json` (currently only E2E-covered).

DanielLeens:
Thanks for catching this — I agree with your compatibility concern.

I rechecked the current diff around `MilvusSinkConverter.buildMilvusData(...)`, and the dynamic-field branch now runs before the generic null validation. In the old flow, a `null` value was rejected before the dynamic-field parsing branch, so a null dynamic-field column also raised `MILVUS-10`. In the current flow, when `enable_dynamic_field=true` and the dynamic-field value is `null`, the code enters the dynamic-field branch, skips adding anything, and `continue`s before the normal null validation can run.

That means this changes default behavior even when `enable_nullable_field=false`, and `skipsNullDynamicFieldBeforeNullValidation` currently codifies that behavior. I do not think my previous approval should stand with this compatibility gap open.

Could you please restore the old default behavior for null dynamic-field values, and only allow the skip/null handling if it is explicitly covered by the new nullable-field contract? After that update I’m happy to re-review the latest head.

loupipalien:
Thanks for your time and advice, I will fix the issues mentioned above.

loupipalien:
For milvus version 2.5.x, when the database does not exist, client.hasCollection should return "database not found[database= xxx]" instead of false
image

DanielLeens (approved):
Thanks for the update! I re-reviewed the latest head and the compatibility concern from my previous round is now fully resolved.

The key fix: the null check now runs **before** the dynamic-field branch, restoring the original default behavior. When `enable_nullable_field=false` (the default), a null value still throws `FIELD_IS_NULL` exactly as before. The dynamic-field null skip only happens when `enable_nullable_field=true` is explicitly set.

I also see you addressed @zhangshenghang's concern the same way, added partition-key exclusion from nullable fields, and fixed the `tableExists` path to check `databaseExists` first — nice catch on the Milvus 2.5.x `hasCollection` behavior when the database doesn't exist.

CI is all green on the current head. Looks good to merge from my side.

loupipalien:
@zhangshenghang The issues mentioned have all been fixed, pls review again, thanks (:

DanielLeens:
Thanks for the ping. I rechecked the current head `9588de8b6ba9` and, from Daniel's side, my latest approval still stands on this unchanged revision.

The compatibility gap I called out earlier is already fixed here:
- the null check runs before the dynamic-field branch again, so the default `enable_nullable_field=false` behavior is preserved
- the partition-key exclusion and the `tableExists()` guard are also on this head
- the current `Build` is green: https://github.com/apache/seatunnel/runs/86194110067

So I do not have a new blocker from Daniel's side on the current head. Since @zhangshenghang raised one of the earlier compatibility concerns, they may still want to confirm their own point as well, but Daniel's review on this revision remains clear.

davidzollo (approved):
+1
LGTM

PR #11218, opened 2026-06-29
[Fix][Connector-V2] Support Dameng NVARCHAR2 type mapping

### LINKED ISSUE #10635: [Bug] [connector-jdbc] DM compatible data type NVARCHAR2
### Search before asking

### What happened

When synchronizing multi-table data from Dameng Database to MySQL Database via the JDBC connector, the Source does not support NVARCHAR2 type mapping.

The task configuration is as follows:
```
{
  "env": {
    "parallelism": 1,
    "job.mode": "BATCH"
  },
  "source": [
    {
      "plugin_name": "Jdbc",
      "plugin_input": "source",
      "driver": "dm.jdbc.driver.DmDriver",
      "url": "jdbc:dm://1127.0.0.1:5236/sfsjzt",
      "user": "SYSDBA",
      "password": "SYSDBA001",
      "table_list": [
        {
          "table_path": "sfsjzt.test_nvarchar2"
        }
      ],
      "connection_check_timeout_sec": 100
    }
  ],
  "sink": [
    {
      "plugin_name": "Jdbc",
      "plugin_input": "sink",
      "url": "jdbc:mysql://127.0.0.1:3306/test?useSSL=false&serverTimezone=Asia/Shanghai&rewriteBatchedStatements=true", 
      "driver": "com.mysql.cj.jdbc.Driver",
      "user": "root",
      "password": "testing", 
      "generate_sink_sql": true,
      "database": "test",
      "table": "${table_name}",
      "data_save_mode": "append_data",
      "schema_save_mode": "CREATE_SCHEMA_WHEN_NOT_EXIST",
      "connection_check_timeout_sec": 100,
      "batch_size": 1000
    }
  ]
}
```
### SeaTunnel Version

apache-seatunnel-2.3.8

### SeaTunnel Config

```conf
Standalone testing environment
```

### Running Command

```shell
../bin/seatunnel.sh --config dm_all_table_collect.json -m local
```

### Error Exception

```log
The error message is as follows
Caused by: org.apache.seatunnel.common.exception.SeaTunnelRuntimeException: ErrorCode:[COMMON-21], ErrorDescription:['Dameng' tables unsupported get catalog table，the corresponding field types in the following tables are not supported: '{"test.test_nvarchar2":{"id":"nvarchar2","name":"nvarchar2","age":"nvarchar2"}}'
```

2.3.8

### Java or Scala Version

jdk1.8

### Screenshots

Image

### Code of Conduct

zhangshenghang:
### Purpose

Fixes #10635.

Dameng supports `NVARCHAR2` as an Oracle-compatible character type, but the JDBC Dameng type converter only handled `NVARCHAR`. As a result, reading a Dameng table with `NVARCHAR2` columns failed while building the catalog table with an unsupported type error.

### Changes

- Add `NVARCHAR2` support to `DmdbTypeConverter` and map it to SeaTunnel `STRING`.
- Treat `NVARCHAR2` as a string type when deciding whether Dameng default values need quotes.
- Add a regression test covering Dameng `nvarchar2` conversion.

### Tests

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-jdbc -DskipIT -Dskip.spotless=true -Dtest=DmdbTypeConverterTest#testNvarchar test`

Note: `./mvnw -pl seatunnel-connectors-v2/connector-jdbc -DskipIT -Dskip.spotless=true test` was also run. The changed `DmdbTypeConverterTest` passed, but the full module run failed on existing Testcontainers-based Oracle/Kingbase tests because Docker is not available in this local environment.

### DIFF
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbDialect.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbDialect.java
@@ -49,6 +49,7 @@ import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm
 import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm.DmdbTypeConverter.DM_LONG;
 import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm.DmdbTypeConverter.DM_LONGVARCHAR;
 import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm.DmdbTypeConverter.DM_NVARCHAR;
+import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm.DmdbTypeConverter.DM_NVARCHAR2;
 import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm.DmdbTypeConverter.DM_TEXT;
 import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm.DmdbTypeConverter.DM_VARCHAR;
 import static org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.dm.DmdbTypeConverter.DM_VARCHAR2;
@@ -338,6 +339,7 @@ public class DmdbDialect implements JdbcDialect {
             case DM_VARCHAR:
             case DM_VARCHAR2:
             case DM_NVARCHAR:
+            case DM_NVARCHAR2:
             case DM_LONGVARCHAR:
             case DM_CLOB:
             case DM_TEXT:
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverter.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverter.java
@@ -67,6 +67,7 @@ public class DmdbTypeConverter implements TypeConverter<BasicTypeDefine> {
     public static final String DM_VARCHAR = "VARCHAR";
     public static final String DM_VARCHAR2 = "VARCHAR2";
     public static final String DM_NVARCHAR = "NVARCHAR";
+    public static final String DM_NVARCHAR2 = "NVARCHAR2";
     public static final String DM_LONGVARCHAR = "LONGVARCHAR";
     public static final String DM_CLOB = "CLOB";
     public static final String DM_TEXT = "TEXT";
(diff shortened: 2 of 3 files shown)


DanielLeens (approved):
Thanks for working on this. I went through the full diff and traced the actual Dameng call path, and I do not see a blocker here.

What this PR fixes
- User pain: when Dameng metadata returns `NVARCHAR2`, `DamengCatalog.buildColumn()` passes that raw `DATA_TYPE` into `DmdbTypeConverter.convert()`. Before this patch, `NVARCHAR2` did not match the `DM_NVARCHAR` branch, so the catalog/schema inference path could fall through to the unsupported-type error.
- Fix approach: this patch adds `DM_NVARCHAR2` to the Dameng string-type mapping in `DmdbTypeConverter`, preserves the exact `NVARCHAR2(length)` source type, and also includes `DM_NVARCHAR2` in `DmdbDialect.needsQuotesWithDefaultValue()` so same-catalog schema-change DDL handles string defaults consistently.
- One-line summary: SeaTunnel can now read and round-trip Dameng `NVARCHAR2` columns correctly instead of treating them as an unknown type.

Runtime call chain I checked

```text
Dameng metadata read
  -> DamengCatalog.buildColumn() [DamengCatalog.java:164-176]
      -> DATA_TYPE from ResultSet is copied into BasicTypeDefine
      -> DmdbTypeConverter.convert(typeDefine)

Type mapping
  -> DmdbTypeConverter.convert() [DmdbTypeConverter.java:207-212]
      -> case DM_NVARCHAR / DM_NVARCHAR2
      -> Column(dataType=STRING_TYPE, sourceType="NVARCHAR2(length)")

Same-catalog DDL / schema change
  -> DamengCreateTableSqlBuilder.buildColumnSql() [DamengCreateTableSqlBuilder.java:112-119]
      -> reuses column.getSourceType() when the source catalog is Dameng

Default-value quoting
  -> DmdbDialect.needsQuotesWithDefaultValue() [DmdbDialect.java:334-346]
      -> DM_NVARCHAR2 is now treated as a quoted string type
```

Key review findings
- The normal production path does hit this change: `DamengCatalog.buildColumn()` directly feeds the raw Dameng type name into `DmdbTypeConverter`, so real `NVARCHAR2` columns will reliably take this new branch.
- The fix is precise rather than invasive. It stays inside the Dameng-specific type-conversion layer and does not change APIs, config options, defaults, or serialization contracts.
- Preserving `sourceType` as `NVARCHAR2(length)` is the right choice because the same-catalog DDL path already prefers `column.getSourceType()` when rebuilding column definitions.

Tests
- `DmdbTypeConverterTest.testNvarchar()` now covers both `nvarchar(2)` and `nvarchar2(2)` in a deterministic in-memory unit test.
- Flaky-test assessment: stable. I did not find sleeps, async waits, external services, random ports, or environment-sensitive assertions in the added test path.

Compatibility / side effects
- Compatibility: fully compatible. This is a bug fix for a previously unsupported Dameng alias, not a behavior break.
- Performance / side effects: no meaningful CPU, memory, GC, retry, idempotency, or resource-lifecycle risk. This is just one additional enum-like branch in a stateless converter.

CI status
- I only saw the new `Build` check still queued when I reviewed. I did not find a code-side issue that should block the PR on review.

Merge conclusion
- Conclusion: ready to merge.
- Blocking items: none.

Overall, this looks like the right minimal fix for a real type-mapping gap. Happy to approve.

PR #11219, opened 2026-06-29
[Improve][Connector-V2] Optimize Arrow Converter SPI registration using AutoService

nzw921rx:

Replace the hand-maintained `META-INF/services/...Converter` SPI file with `@AutoService(Converter.class)` annotations on each Arrow `Converter` implementation in `connector-common`.

This eliminates the risk of forgetting to register new converters and keeps the SPI configuration closer to the implementation classes.

Affected converters: 

- DateMilliConvertor
- FixedSizeListConverter
- LargeListConverter
- ListConverter
- MapConverter
- NullConverter
- StructConverter
- TimeStampMicroConverter
- TimeStampMilliConverter
- TimeStampNanoConverter
- TimeStampSecConverter

No. The set of registered converters is unchanged; only the registration mechanism changed from a manual SPI file to compile-time annotation processing.

```bash
./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-common -Dskip.ui=true
./mvnw test -pl seatunnel-connectors-v2/connector-common -Dskip.ui=true

### CODE REVIEW

**Review by DanielLeens — APPROVED** [review]
Thanks for the cleanup here. I reviewed the full diff and traced the real runtime path in the Arrow reader, and I do not see a blocker.

What this PR improves
- User pain: `ArrowToSeatunnelRowReader` loads Arrow converters through `ServiceLoader.load(Converter.class)`. With a hand-maintained `META-INF/services/...Converter` file, every new converter or rename needs a second manual edit, which is easy to miss and can drift from the actual source set.
- Fix approach: this PR moves 11 existing converter registrations to `@AutoService(Converter.class)` and removes the manually maintained provider file.
- One-line summary: converter registration is now declared next to the implementation class and generated at compile time, which is easier to maintain and less error-prone.

Runtime call chain I checked

```text
Arrow row decoding
  -> ArrowToSeatunnelRowReader static init [ArrowToSeatunnelRowReader.java:75-76]
      -> ServiceLoader.load(Converter.class)
      -> loads all Converter providers into the converters list

This PR's registration path
  -> converter classes declare @AutoService(Converter.class)
      -> Maven compiler annotation processing generates META-INF/services/...Converter

Runtime usage
  -> convertSeatunnelRow()
      -> picks a loaded Converter for the Arrow minor type
      -> delegates to the unchanged convert()/support() logic
```

Key review findings
- The normal runtime path absolutely hits this change because the Arrow reader always depends on `ServiceLoader` to discover custom converters.
- The important question here is not the converter logic itself, but whether the generated provider file really ends up in the final artifact.
- I verified that locally with a minimal packaging check on this PR branch:
  `./mvnw -pl seatunnel-connectors-v2/connector-common -am -DskipTests -DskipITs -DskipUT package`
  completed successfully, and the produced `connector-common` jar contains `META-INF/services/org.apache.seatunnel.connectors.seatunnel.common.source.arrow.converter.Converter` with the expected 11 providers.
- Because the runtime `ServiceLoader` path is unchanged and the generated descriptor matches the old manual list, this is behavior-preserving while improving maintainability.

Compatibility / side effects
- Compatibility: fully compatible. No API, config, default, protocol, or serialization contract changes.
- Performance / side effects: no meaningful runtime change. `ServiceLoader` is still used exactly as before; only the source of the provider list has changed.

Tests
- This PR does not change UT/E2E code, so flaky-test analysis is not triggered here.
- There is also no new converter behavior to re-test functionally. The key regression surface is packaging, and the local artifact check above covers that directly.

CI status
- The new `Build` check was still queued when I reviewed. I did not find a code-side issue that should block the PR from a review perspective.

Merge conclusion
- Conclusion: ready to merge.
- Blocking items: none.

Overall, this looks like a solid long-term cleanup. Happy to approve.

DanielLeens:

- base branch: `dev`
- head SHA: `3fe20b77d652fb5ce0f1610215970cfbf7d8cf15`
- compare status: `diverged`
- ahead_by: `1`
- behind_by: `9`
- mergeable_state: `unknown`
- CI / merge gate snapshot: `failing` (failing: `Build [cancelled]`)

DanielLeens (changes requested):
Thanks for the follow-up update. I re-reviewed the latest head `9f834fba21`, and the previous Daniel approval on `69ae753142` no longer applies to this new revision.

The approved revision had already removed the handwritten `META-INF/services/...Converter` file and relied on `@AutoService(Converter.class)` as the single source of truth. The new head adds that handwritten service descriptor back, so the current branch has gone back to a two-source registration model again.

## What this PR solves
- User pain: `ArrowToSeatunnelRowReader` depends on `ServiceLoader.load(Converter.class)`, so a handwritten provider file is easy to forget when converters are added or renamed.
- Fix approach: move converter registration next to the implementation classes with `@AutoService(Converter.class)` and stop maintaining the provider file by hand.
- One-line summary: the earlier approved revision had achieved that cleanup, but the latest commit reintroduces the handwritten provider list and reopens the exact drift point this PR was supposed to remove.

## Runtime chain I checked
```text
Arrow row decoding
  -> ArrowToSeatunnelRowReader static init [ArrowToSeatunnelRowReader.java:75-76]
      -> ServiceLoader.load(Converter.class)
      -> loads provider entries from META-INF/services/...Converter

Previously approved revision (69ae753142)
  -> converter classes use @AutoService(Converter.class)
  -> the handwritten META-INF/services/...Converter file is gone
  -> provider registration has a single source of truth

Current latest revision (9f834fba21)
  -> @AutoService is still present on the converter classes
  -> but src/main/resources/META-INF/services/...Converter is added back [Converter service file:1-26]
  -> provider registration is split across generated and handwritten sources again
```

## Findings

### Issue 1: the latest commit adds the handwritten Arrow converter service descriptor back, so this PR is now back to a two-source provider registration model
- Location:
  - `seatunnel-connectors-v2/connector-common/src/main/resources/META-INF/services/org.apache.seatunnel.connectors.seatunnel.common.source.arrow.converter.Converter:1-26`
  - `seatunnel-connectors-v2/connector-common/src/main/java/org/apache/seatunnel/connectors/seatunnel/common/source/arrow/reader/ArrowToSeatunnelRowReader.java:75-76`
- Why this is a blocker:
  The previous Daniel approval was based on this PR fully removing the handwritten provider list and letting `@AutoService` generate the service descriptor. On the latest head, the runtime `ServiceLoader` entry point is unchanged, but the handwritten file is back in `src/main/resources`. That means the current branch no longer keeps provider registration in one place.
- Risk:
  This reintroduces the exact maintenance drift the PR was meant to remove. Future converter additions or renames would again need someone to think about a second handwritten provider source, and if the handwritten file and generated output ever diverge, the final runtime provider list becomes dependent on packaging/output ordering instead of a single code-side contract.
- Best fix:
  - Option A: remove the handwritten `META-INF/services/...Converter` file again and keep `@AutoService` as the only registration source.
  - Option B: if there is a concrete build-chain reason why the handwritten file must stay, please document that explicitly and realign the PR goal around that constraint. Based on the current PR scope, I would strongly prefer option A.
- Severity: High

- API / config / protocol compatibility is unchanged.
- The blocker is not an external compatibility break; it is that the latest head walks back the single-source registration contract that the approved revision had already established.

## Tests
- This latest commit does not add UT/E2E changes, so flaky-test analysis is not triggered here.
- I also do not see new evidence explaining why the handwritten service file needs to come back after the earlier approved revision had already removed it.

## CI / upstream sync snapshot
- The latest `Build` is currently queued on the new head.
- This branch is also behind the latest `dev` (`ahead_by=2`, `behind_by=8`, compare status: `diverged`).
- I am recording that sync fact, but it is not the main blocker in this round. Issue 1 needs to be fixed first on the current source path.

## Merge conclusion

1. Blocking items
- Issue 1: remove the reintroduced handwritten Arrow converter service descriptor, or clearly justify why this PR now needs both handwritten and generated provider sources.

- None from this round.

Overall, the earlier approved revision had already achieved the AutoService cleanup. The latest commit reopens that exact maintenance problem, so I would like to see the branch return to a single registration source before merge.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `462caa878a` from the Arrow `ServiceLoader` entry point down to the current provider-registration layout again.

## What this PR solves
- User pain: `ArrowToSeatunnelRowReader` depends on `ServiceLoader.load(Converter.class)`, so keeping both `@AutoService` and a handwritten `META-INF/services/...Converter` file creates an easy maintenance-drift point.
- Fix approach: keep converter registration on the implementation classes with `@AutoService(Converter.class)` and remove the handwritten service descriptor from the source tree.
- One-line summary: the current head has closed the earlier Daniel blocker by going back to a single registration source on the source path.

## Runtime chain I rechecked
```text
Arrow row decoding
  -> ArrowToSeatunnelRowReader static init [ArrowToSeatunnelRowReader.java:75-77]
      -> ServiceLoader.load(Converter.class)

Provider source on the current head
  -> converter implementations still carry @AutoService(Converter.class)
  -> handwritten src/main/resources/META-INF/services/...Converter is gone again
```

## Findings
- The previous Daniel blocker is closed for me now:
  - the handwritten Arrow converter service descriptor is no longer present on the current head
  - the converter implementations still keep `@AutoService(Converter.class)` as the single source of truth
- I do not see a reopened source-level blocker in this round.

## CI / upstream sync snapshot
- The latest `Build` is currently red on the current head.
- This branch is also still behind the latest `dev` (`behind_by=1`, compare status: `diverged`).
- Since the source-side blocker is already cleared, the lowest-cost next step is to sync with the latest `dev` and rerun `Build` first. If the refreshed head still fails, I’m happy to narrow the remaining CI signal down on top of that updated revision.

## Merge conclusion

1. Blocking items
- Please sync with the latest `dev` and rerun `Build` first.

- None from this round.

Overall, the branch is back on the intended AutoService-only path. The remaining gate here is CI / branch state rather than a reopened Arrow source-path problem.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from the Arrow converter ServiceLoader path and the backend workflow timeout changes.

## What this PR solves
- User pain: the Arrow converter SPI list was maintained by hand, which is easy to forget when adding converter classes.
- Fix approach: converters now use `@AutoService(Converter.class)` and the handwritten service file is removed. The backend workflow timeout is also widened for the long connector lanes.
- One-line summary: Arrow converter registration becomes annotation-driven instead of manual-file driven.

## Runtime path checked

```text
Arrow source conversion
  -> ServiceLoader<Converter>
      -> generated META-INF/services entry from @AutoService
      -> converter.support(Types.MinorType)
      -> converter.convert(rowIndex, vector)
```

I also checked the Maven parent configuration: `auto-service` is available and the compiler plugin has AutoService in `annotationProcessorPaths`, so the generated service descriptor should be produced during normal Maven compilation.

## Merge conclusion

### Conclusion: can merge

1. Blocking items
- None from Daniel's source review.

2. Suggested follow-up
- CI remains the gate, especially because this depends on generated service metadata.

Overall, this is a clean maintainability improvement. LGTM from Daniel's side, pending CI.

nzw921rx:
@davidzollo  can you help me review this, thanks

davidzollo (approved):
+1 if CI passes.
LGTM

PR #11220, opened 2026-06-29
[Fix][API] Support dotted opaque table names

### LINKED ISSUE #9966: [Bug] [Kafka source connector] (Kafka topic name) Cannot get split 'reg.country.user_activity.activity' to get databaseName and tableName
### Search before asking

### What happened

Hello everyone!
Could you please help me?
I've encountered a problem. I get an error when I run the job.

although there are no problems with topics of the format "reg.country.user_activity".
Could you please tell me how to solve this? Has anyone else encountered this?
There's no option to change the topic name.

Thanks!

### SeaTunnel Version

Version:2.3.10
Commit: 5c8a4c0

### SeaTunnel Config

```conf
env {
  parallelism = 2
  job.mode = "STREAMING"
  checkpoint.interval = 300000
}

source {
  Kafka {
    bootstrap.servers = "host1,host2,host3:port"
    tables_configs = [
      {
        topic = "reg.country.user_activity.activity"
        consumer.group = "consumer_gp_name"
        format = "json"
        kafka.auto.offset.reset = "earliest"
        kafka.max.poll.records = 500
        kafka.session.timeout.ms = 45000
        kafka.heartbeat.interval.ms = 15000
        kafka.request.timeout.ms = 40000
        kafka.admin.timeout.ms = 60000
        kafka.default.api.timeout.ms = 60000

        kafka.config = {
          security.protocol = "SSL"
          ssl.truststore.location = "/kafka/truststore.jks"
          ssl.truststore.password = "password"
          ssl.keystore.location = "/kafka/keystore.jks"
          ssl.keystore.password = "password"
          ssl.key.password = "password"
        }
      }
    ]
  }
}

sink {
  S3File {
    bucket = "s3a://bucket-prt-1"
    tmp_path = "/seatunnel_tmp/kafka/country"
    path = "s3a://bucket-prt-1/product=country/platform=kafka/env=prod/layer=raw/table=user_activity_activity"
    fs.s3a.endpoint = "https://my-minio.my/"
    fs.s3a.aws.credentials.provider = "org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider"
    access_key = "access_key"
    secret_key = "secret_key"

    file_format_type = "parquet"
    compress_codec = "SNAPPY"
    schema_save_mode = "CREATE_SCHEMA_WHEN_NOT_EXIST"
    data_save_mode = "APPEND_DATA"

    is_enable_transaction = "true"

    custom_filename = "true"
    file_name_expression = "date=${now}/${transactionId}"
    filename_time_format = "yyyy-MM-dd"

    rolling_policy = {
        max_part_size = "128MB"
        inactivity_interval = "15m"
        rollover_interval = "1h"
    }

    hadoop_s3_properties = {
        fs.s3a.path.style.access = "true"
        fs.s3a.endpoint.region = "region"
    }
  }
}
```

### Running Command

```shell
API REST V.2
{{SeaTunnelHost}}/submit-job?format=hocon&jobName=kafka_to_s3__test_job&jobId=&isStartWithSavePoint=true

>>out
{
    "jobId": "1032283892776697857",
    "jobName": "kafka_to_s3__test_job"
}
```

### Error Exception

```log
{
    "status": "fail",
    "message": "Cannot get split 'reg.country.user_activity.activity' to get databaseName and tableName"
}
```

Zeta 

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
### Purpose

Fixes #9966.

Kafka topics can legally contain multiple dots, for example `reg.country.user_activity.activity`. SeaTunnel can create a `TablePath` with that full topic as an opaque table name, but later code may parse the string form again through `TablePath.of(String)`. The previous parser rejected names with more than three dot-separated parts, which caused jobs using these topic names to fail with:

`Cannot get split 'reg.country.user_activity.activity' to get databaseName and tableName`

### Changes

- Keep the existing parsing behavior for one, two, and three-part table paths.
- Treat names with more than three dot-separated parts as opaque table names instead of rejecting them.
- Add API coverage for dotted opaque table names.
- Add Kafka source config coverage for a topic with multiple dots.

### Tests

- `./mvnw spotless:apply`
- `git diff --check`
- `./mvnw -pl seatunnel-api -Dtest=TablePathTest test -Dskip.spotless=true`
- `./mvnw -pl seatunnel-api install -DskipTests -Dskip.spotless=true`
- `./mvnw -pl seatunnel-connectors-v2/connector-kafka -Dtest=KafkaSourceConfigTest test -Dskip.spotless=true`

Note: `./mvnw -q -DskipTests verify` was started but produced no output for an extended period and was interrupted locally. A narrower `-am` verification attempt was blocked by an unrelated local `seatunnel-config-shade` compilation issue.


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full diff from the Kafka schema/table parsing entrypoint back into the shared `TablePath.of(String)` API, rather than treating this as a Kafka-only test fix.

## What this PR solves
- User pain: when a Kafka topic name itself contains dots, using that value through the schema/table identifier path can currently fail because `TablePath.of(String)` treats dot-separated input as structured database/schema/table segments.
- Fix approach: the PR changes `TablePath.of(String)` so any input with more than three dot-separated parts is treated as one opaque table name instead of throwing.
- One-line summary: the Kafka use case is real, but the current implementation solves it by changing a shared API's default parsing semantics for the whole codebase.

```text
Kafka source schema path
  -> KafkaSourceConfig.getTablePathFromSchema(...) [KafkaSourceConfig.java:307-316]
      -> if schema.table exists: TablePath.of(tableString)
      -> else: TablePath.of(null, topicName)

Shared API changed by this PR
  -> TablePath.of(String, boolean) [TablePath.java:53-69]
      -> 1 part: table only
      -> 2 parts: db.table or schema.table
      -> 3 parts: db.schema.table
      -> 4+ parts:
           old behavior -> fail fast
           new behavior -> treat the whole string as one opaque table name

Other normal callers still using the same shared API
  -> TableSchemaDiscoverer.discoverTableSchemaFromMetaLake(...) [TableSchemaDiscoverer.java:121-126]
  -> DataValidatorTransform.resolveErrorTablePath(...) [DataValidatorTransform.java:148-163]
```

## Key findings
1. The Kafka scenario is real, but the actual problematic call site is narrow: `KafkaSourceConfig.getTablePathFromSchema(...)` when `schema.table` is explicitly provided.
2. The current patch changes `seatunnel-api` behavior globally, not just the Kafka path.
3. That means inputs which previously failed fast as malformed 4-part identifiers will now be silently accepted as opaque table names in unrelated flows.

## Findings

### Issue 1: this changes the global default behavior of `TablePath.of(String)` and can silently accept malformed identifiers in unrelated modules
- Location:
  - `seatunnel-api/src/main/java/org/apache/seatunnel/api/table/catalog/TablePath.java:53-69`
  - related caller that actually needs the fix: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaSourceConfig.java:307-316`
  - affected shared callers include:
    - `seatunnel-api/src/main/java/org/apache/seatunnel/api/metalake/TableSchemaDiscoverer.java:121-126`
    - `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/validator/DataValidatorTransform.java:148-163`
- Why this is a problem:
  - This PR is trying to fix a Kafka/table-name parsing case, but the implementation changes the default semantics of a shared API used across many modules.
  - Before this patch, 4-part input failed immediately. After this patch, the same input becomes `database=null, schema=null, table=fullName`.
  - That is not just “more permissive parsing”. It changes where and how invalid input is surfaced in normal metadata and validation paths.
- Risk:
  - Misconfigured identifiers that should fail fast can now continue deeper into the pipeline and later show up as wrong-table routing, missing schema/database context, or harder-to-diagnose catalog lookup failures.
  - Since this lives in `seatunnel-api`, I would treat that semantic change as a blocker unless the scope is intentionally broader and justified end to end.
- Best fix:
  - Option A: keep `TablePath.of(String)` strict and handle the Kafka opaque-name case in `KafkaSourceConfig.getTablePathFromSchema(...)`.
  - Option B: introduce an explicit helper such as `TablePath.opaque(String)` and let only the callers that truly need opaque parsing opt in.
  - I would strongly prefer either of those over changing the shared default behavior.
- Severity: High

## Tests and stability
- The new tests are stable and deterministic.
- But they only prove the new permissive behavior, not that the broader shared callers still retain the intended fail-fast contract.

## CI
- The current `Build` is red, but the visible failed jobs are in untouched modules (`all-connectors-it-7`, `connector-file-sftp-it`, `kudu-connector-it`), so I am not treating that red build as the main blocker here.
- The main blocker is still the shared API scope change above.

1. Blocking items
- Issue 1: please narrow the fix to the Kafka path, or add an explicit opaque-parsing entrypoint, instead of changing the global default behavior of `TablePath.of(String)`.

- None on top of the blocker above.

The user-facing problem is worth fixing. I just do not think the current shared-API scope is safe enough to merge as-is.

DanielLeens (commented):
Thanks for narrowing the fix. I re-reviewed the latest head from the Kafka-specific schema-table parsing path instead of the earlier shared `TablePath` scope that Daniel blocked before.

## What this PR solves
- User pain: Kafka schema `table` values can legitimately be opaque names with multiple dots, but the old path fed them straight into `TablePath.of(...)`, which is a shared structured-identifier parser.
- Fix approach: keep `TablePath.of(String)` strict, and handle the Kafka opaque-name case locally in `KafkaSourceConfig.parseKafkaTablePath(...)`.
- One-line summary: the latest head fixes the Kafka bug without changing the default parsing semantics of the shared API.

## Runtime path I checked
```text
Kafka source config initialization
  -> KafkaSourceConfig.createCatalogTable(...)
      -> getTablePathFromSchema(readonlyConfig, topicName)
          -> schema.table present
              -> parseKafkaTablePath(tableName)
                  -> 4+ dot-separated parts -> opaque TablePath.of(null, tableName)
                  -> otherwise -> strict TablePath.of(tableName)
          -> schema.table absent
              -> TablePath.of(null, topicName)
```

## Key findings
1. The earlier shared-API blocker is resolved on the latest head: `TablePath.of(String)` stays strict, and the Kafka workaround is now local to the Kafka source path.
2. The new `TablePathTest` preserves the old fail-fast contract for malformed structured identifiers outside Kafka.
3. The new Kafka test covers the real opaque-name case this PR is targeting.
4. I do not see a reopened source-level blocker in the current 3-file scope.

## Tests and stability
- The added tests are deterministic and stable.
- They cover both sides of the fix:
  - shared API still rejects malformed 4-part structured names
  - Kafka can now use the multi-dot opaque name on its own path

## Compatibility
- Fully compatible.
- No user config key, default value, protocol, or serialization format is changed.
- The behavioral change is intentionally narrowed to the Kafka parsing path that needed it.

1. Blocking items
- No code-level blocker from Daniel's side on the latest head.
- The current `Build` is still in progress, so please let it finish green before merge.

- No additional source-level follow-up from Daniel's side on top of the current narrowed fix.

Overall, this latest revision is in much better shape than the earlier shared-API version. The remaining gate is CI, not a reopened parsing-scope issue.

DanielLeens:

- base branch: `dev`
- head SHA: `279b0d8924065359b3527a17b6886c089eccbcc3`
- compare status: `diverged`
- ahead_by: `2`
- behind_by: `4`
- mergeable_state: `unknown`

DanielLeens (approved):
Thanks for narrowing the fix and syncing with the latest `dev`. I re-reviewed the latest head from the Kafka schema-table parsing entrypoint rather than from the old shared `TablePath` discussion.

Runtime path I checked
```text
Kafka source schema config
  -> KafkaSourceConfig.getTablePath(...)
      -> parseKafkaTablePath(table)
          -> dotted opaque-name fallback for >2 dots
      -> TablePath.of(...) only for the structured-name cases that Kafka still wants
```

- The important part for me is that this stays Kafka-local now. The earlier shared `TablePath.of(String)` contract risk is not reopened on this revision.
- The new `TablePathTest` still keeps malformed structured names rejected at the API layer, while the Kafka-specific helper treats the multi-dot opaque topic/table case as an opaque name instead of trying to reinterpret it as catalog structure.
- The added tests are focused and stable from my side.

Merge conclusion
- No blocking issue from my side on the current head.
- This can merge once the normal checks finish.

davidzollo (approved):
+1
LGTM

Issue #10804, opened 2026-04-21
[STIP-26][Feature][Zeta] Decouple Hazelcast IMap via StateStore Abstraction


dybyte:
### Search before asking

- [x] I had searched in the [feature](https://github.com/apache/seatunnel/issues?q=is%3Aissue+label%3A%22Feature%22) and found no similar feature requirement.

### Description

# StateStore Interface Design Draft

## 1. Summary

This document proposes an initial `statestore` interface layer under `common/statestore`

to reduce direct Hazelcast `IMap` dependencies inside the SeaTunnel engine.

The purpose of this draft is **not** to finalize the entire distributed-system architecture.

Instead, it focuses on introducing a storage boundary first, so that engine code can evolve

without being tightly coupled to a specific storage implementation.

At this stage, the proposal is limited to:

- introducing engine-facing state-store interfaces
- introducing their initial implementations
- establishing clearer boundaries between different categories of state

This draft does **not** include wiring those interfaces into the existing product/service code yet.

That adoption step is expected to happen incrementally in later phases.

---

## 2. Background and Problem

The current engine code directly depends on Hazelcast `IMap` in many places.

While this approach is simple initially, it creates several long-term problems:

- changing the storage strategy becomes difficult
- communication and state responsibilities remain tightly coupled
- different categories of state are treated as if they required the same level of distribution and consistency
- the code does not clearly show which states require strong correctness guarantees

Looking at the current `IMap` usages, the stored data does not have a single nature.

Examples include:

- runtime control state
- counters such as checkpoint IDs
- metrics snapshots
- finished-job history

Because of that, a generic `IMap`-style abstraction is not the main goal here.

Instead, this draft introduces interfaces that follow the meaning of each state category.

---

## 3. Goals

The goals of this draft are:

1. Prevent engine-related code from depending directly on Hazelcast-specific storage types.
2. Separate storage responsibilities according to state semantics.
3. Introduce an initial interface layer and corresponding implementations under `common/statestore`.
4. Keep future options open for `Ratis`, `Pekko`, `RocksDB`, or other implementations.

---

## 4. Non-Goals

This draft does **not** aim to:

- finalize the full distributed-system architecture
- finalize `Ratis` integration
- finalize `Pekko`based communication design
- define full membership, topology, or leader-election behavior
- define the final snapshot, compaction, or persistence policy
- wire the new interfaces into all existing product/service code in this stage

In other words, this is a practical first-step design for introducing a storage boundary,

[...]

PR #11226, opened 2026-06-30
[Improve][Zeta] Abstract checkpoint id counter state store

dybyte:
Related: #10804 

Abstract the checkpoint id counter map behind a dedicated `CounterStateStore` interface and wire it through `EngineStateStores`.

No.

### Check list


seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java
```diff
@@ -57,26 +57,26 @@ public void start() throws Exception {
...
+            checkpointCounterStore.remove(key);
         }
         return CompletableFuture.completedFuture(null);
     }
 
     @Override
     public long getAndIncrement() throws Exception {
-        Long nextId = checkpointIdMap.compute(key, (k, v) -> v == null ? null : v + 1);
+        long nextId = checkpointCounterStore.incrementAndGet(key);
         checkNotNull(nextId);
```
CosmosNi, line 68: The default value of nextId is 1L, so the checkNotNull check has no effect. Alternatively, should incrementAndGet return -1 to throw an exception? Since the start method will initialize it, if it is null, that indicates the state has been lost.

dybyte: Closing this PR because the source branch link was broken after I renamed the branch.
I will reopen a new PR with the corrected branch name. Sorry for the inconvenience.

PR #11227, opened 2026-06-30
[Fix][API] Fix conditional validation error context

nzw921rx:

This pull request fixes conditional validation error reporting in `seatunnel-api`.

- Remove the per-condition exception wrapping in `ConditionEvaluators` so evaluation failures are not prematurely attributed to the first condition node.
- Catch `OptionValidationException` at the `ConditionRule` level in `ConfigValidator` and report the full conditional expression as the error context.

Yes. When a conditional rule fails during evaluation, the validation error now points to the whole condition expression instead of only the first option in the chain. This improves the accuracy of validation diagnostics without changing validation behavior.

```bash
./mvnw spotless:apply
./mvnw -q -DskipTests verify
```

`./mvnw test` was started but not waited to completion per user preference.

### Check list


nzw921rx:
Close and resubmit

PR #11228, opened 2026-06-30
[Fix][API] Preserve extension exception context in conditional validation

nzw921rx:

This pull request fixes distorted conditional validation errors in `seatunnel-api`.

- Remove the `try/catch` wrapping from `ConditionEvaluators#evaluate`, which was rewriting `OptionValidationException` thrown by the `EXTENSION` operator during conditional evaluation.
- Catch `OptionValidationException` at the `ConditionRule` level in `ConfigValidator` so the aggregated validation error is reported against the full conditional expression context.

Yes. When an `EXTENSION` condition throws `OptionValidationException` inside a conditional rule, the original validation message is now preserved and reported against the whole condition expression instead of being rewritten at the per-condition level.

```bash
./mvnw spotless:apply
./mvnw -q -DskipTests verify
./mvnw -pl seatunnel-api -Dtest=ConfigValidatorTest test
```

### Check list

### DIFF
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java
@@ -46,15 +46,10 @@ public final class ConditionEvaluators {
             throw new OptionValidationException(
                     "Condition for option '%s' has a null operator", condition.getOption().key());
         }
-        try {
-            Object value = config.get(condition.getOption());
-            Evaluator evaluator = REGISTRY.get(operator);
-            return evaluator.evaluate(value, condition, config);
-        } catch (OptionValidationException e) {
-            throw new OptionValidationException(
-                    "Failed to evaluate constraint '%s' on option '%s': %s",
-                    condition.toString(), condition.getOption().key(), e.getRawMessage());
-        }
+
+        Object value = config.get(condition.getOption());
+        Evaluator evaluator = REGISTRY.get(operator);
+        return evaluator.evaluate(value, condition, config);
     }
 
     @SuppressWarnings({"rawtypes"})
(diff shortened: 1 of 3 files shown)


DanielLeens (commented):
Thanks for the fix here. I reviewed the latest head from the conditional-validation execution path rather than only from the new unit-test assertions.

## What this PR solves
- User pain: when a custom `ConditionExtension` throws during conditional validation, the current error context is too narrow and does not clearly point back to the full conditional expression.
- Fix approach: stop wrapping extension exceptions at the low-level evaluator layer, and instead aggregate them at the `ConditionRule` level so the final message keeps the whole conditional-expression context.
- One-line summary: this is an error-reporting fix that preserves the real failing constraint while improving where the context is attached.

## Runtime path I checked
```text
option validation
      -> collectErrors(...)
          -> ConditionRule expression
              -> validate(expression)
                  -> ConditionEvaluators.evaluate(condition, config)
                      -> EXTENSION evaluator
                          -> ConditionExtension.evaluate(...)
              -> OptionValidationException bubbles back to collectErrors(...)
                  -> format as `type: conditional` with the whole expression text
```

## Key findings
1. The normal conditional-validation path definitely hits this change through `ConfigValidator.collectErrors(...)`.
2. The latest head keeps value-constraint error handling intact while improving the conditional-expression case specifically.
3. The new test covers the real regression target: an extension exception inside a conditional expression now reports the whole expression context instead of only the leaf option.
4. I do not see a reopened source-level blocker in the current 3-file diff.

## Tests and stability
- The new tests are deterministic pure-validation tests.
- Stability rating: Stable.
- Basis: no timing, concurrency, network, port, filesystem, or container dependency.

## Compatibility
- Fully compatible.
- No user config key, default value, protocol, or serialization format is changed.
- The behavioral change is limited to validation error reporting context.

1. Blocking items
- No code-level blocker from Daniel's side.
- The current `Build` is still queued / pending, so please let the latest CI run complete cleanly before merge.

- No extra source-level follow-up from Daniel's side on top of the current fix.

Overall, the latest head looks good to me. The remaining gate is CI completion, not a reopened validation-logic issue.

DanielLeens:

- base branch: `dev`
- head SHA: `26fa760c59c1a30fbfe5e2d7b1d9a1417c2bfb07`
- compare status: `diverged`
- ahead_by: `2`
- behind_by: `4`
- mergeable_state: `unknown`
- CI / merge gate snapshot: `pending` (pending: `Build`)

dybyte (approved):

chl-wxp (approved):

PR #11229, opened 2026-06-30
[Improve][Zeta] Abstract checkpoint id counter state store

dybyte:
Related: #10804 

Abstract the checkpoint id counter map behind a dedicated `CounterStateStore` interface and wire it through `EngineStateStores`.

No.

### Check list


DanielLeens (commented):
Thanks for working on this. I reviewed the latest head from the checkpoint ID counter path through the new engine state-store abstraction.

## What this PR solves
- User pain: checkpoint ID progression currently reaches straight into a Hazelcast map from the checkpoint counter implementation.
- Fix approach: move the counter behind an explicit `CounterStateStore` / `AuthoritativeStateStores` abstraction while keeping the concrete Hazelcast-backed behavior and key format the same.
- One-line summary: the refactor lands on the real checkpoint coordinator path, and I do not see a source-level blocker on the current head.

## Runtime path I checked
```text
JobMaster initialization
  -> new CheckpointManager(..., engineContext, ...)
      -> engineContext.getStateStores().checkpointCounterStore()
      -> new StateStoreCheckpointIDCounter(jobId, pipelineId, store)

Checkpoint counter lifecycle
  -> start()
      -> initializeIfAbsent(key, INITIAL_CHECKPOINT_ID)
  -> getAndIncrement()
      -> incrementAndGet(key)
  -> shutdown(end-state)
      -> remove(key)
```

## Key findings
- The key format is still the same base64-encoded `jobId + pipelineId` pair, so the runtime identity boundary is unchanged.
- The backing store on this PR is still Hazelcast, so this is an abstraction cleanup on the real path, not a silent behavior rewrite.
- The new unit coverage for `StateStoreCheckpointIDCounter` and `HazelcastCounterStateStore` matches the counter semantics actually used by the checkpoint manager.
- I did not find a reopened source-level blocker on the current revision.

## Tests and stability
- The newly added tests are deterministic and do not use timing sleeps.
- The failover IT update still checks the real counter-growth behavior after master handoff.
- The current `Build` is queued, so CI is still the remaining gate.

1. Blocking items

- None from this round.

DanielLeens:

- base branch: `dev`
- head SHA: `65ca714cbb369f6d3260ab19c8771ef66c7ec322`
- compare status: `diverged`
- ahead_by: `3`
- behind_by: `4`
- mergeable_state: `unknown`

davidzollo:
Thanks for the refactor. The overall direction makes sense, and I like that the checkpoint counter is now moved behind a dedicated state-store abstraction.

I see two gaps. First, the most sensitive path here is the recovery branch in `CheckpointManager`, especially `idCounter.setCount(checkpointId + 1)`. The new active tests mainly cover wrapper-level counter behavior, and the failover E2E only shows that the counter keeps growing after failover. That still does not fully verify the manager-level recovery wiring. If possible, could you add a live regression test for the `restart/savepoint=true` path, or re-enable `CheckpointManagerTest` if it can be made stable?

Second, since `StateStoreCheckpointIDCounter` is now on the core checkpoint lifecycle path, a bit more comment  would really help future readers. A short class-level Javadoc plus brief comments for `start()`, `shutdown()`, `getAndIncrement()`, and `setCount()` would make the key format, recovery behavior, and end-state cleanup semantics much easier to understand.

I think these two additions would make the refactor both safer and easier to maintain^_^

DanielLeens:
Thanks for the detailed follow-up review. I rechecked the current head after your comment, and I agree the two gaps you called out need author-side code follow-up before this can move forward. Once there is a new commit addressing those points, I am happy to re-review the latest head.

dybyte:
@davidzollo Thanks for the review. Updated.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch again, and I agree with @davidzollo's earlier focus areas: this PR needed better manager-level recovery coverage and clearer contract comments. The latest head addresses both of those for me.

- `StateStoreCheckpointIDCounter` now has the missing class-level and method-level comments that explain the key format, recovery behavior, and end-state cleanup semantics.
- `CheckpointManagerTest` is no longer disabled and now covers both:
  - finished-pipeline cleanup of the counter + checkpoint state
  - `startWithSavepoint` resuming from `restoredCheckpointId + 1`

## What this PR solves
- User pain: the checkpoint ID counter was still tied too directly to the concrete Hazelcast map path, and the most sensitive recovery semantics were not well documented.
- Fix approach: move the counter behind `CounterStateStore` / `EngineStateStores` while preserving the current key format and runtime behavior, then add the missing recovery/cleanup coverage.
- One-line summary: the abstraction direction still makes sense, and the two gaps from the earlier review are now closed on the latest head.

## Runtime path I rechecked
```text
JobMaster initialization
  -> new CheckpointManager(..., engineContext, ...)
      -> engineContext.getStateStores().checkpointCounterStore()
      -> new StateStoreCheckpointIDCounter(jobId, pipelineId, store)
      -> start()

savepoint / restart recovery
  -> load latest checkpoint
  -> idCounter.setCount(restoredCheckpointId + 1)
  -> next trigger continues from the restored checkpoint sequence

end-state cleanup
  -> shutdown(FINISHED)
      -> CounterStateStore.remove(key)
```

## Findings
- The earlier reviewer concern about missing manager-level recovery coverage is addressed by `CheckpointManagerTest`.
- The earlier maintainability concern about the counter contract being under-documented is also addressed by the new `StateStoreCheckpointIDCounter` comments.

## CI / merge-gate note
- Current GitHub gate snapshot:
  - `Build`: queued / pending (`https://github.com/apache/seatunnel/runs/85292738207`)
- Current base/head sync fact:
  - base branch: `dev`
  - head SHA: `0bbd298debe2e2c3c5debb4deada15dbf8306776`
  - compare status: `diverged`
  - ahead_by: `4`
  - behind_by: `94`
- I am recording that sync fact here, but I am not turning it into a new sync-first blocker comment because the current head does not yet show a PR-unrelated failing lane to narrow against.

### Merge conclusion: can merge after the current build / gate clears
1. Blocking items
- No new source-level blocker from Daniel's side on the latest head.
- The remaining gate is the current `Build` / merge state.

2. Suggested follow-up
- None from this source review round.

Rechecked on the latest head: the earlier recovery/test-coverage and comment/documentation gaps are closed for me now, and I do not see a reopened code-side issue in the checkpoint counter path.

DanielLeens (approved):
Rechecked the current head and CI gate after my previous source-level review. My last Daniel conclusion on this PR had no remaining code-side blocker; it was waiting on the CI / merge gate.

The current visible checks are now successful, including `Build`. From Daniel's side, the earlier source-level blocker is cleared and I do not see a reopened issue on the unchanged head.

Approving this from the current Daniel technical review state. If GitHub still shows a repository-level merge gate, that should be handled by the normal maintainer / branch protection flow.

SEZ9 (changes requested):
Thanks for the contribution, this is a valuable change! I went through the diff carefully — nice work overall. Before it merges, there are a few things I'd like to confirm/address, listed by priority (1 high, 3 medium):

### 1. 🔴 High · Functional Impact — Possible public API change in seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/Constant.java
A public/protected signature appears to be removed or changed. Verify backward compatibility for downstream connectors.

**Where (this PR):**
- `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/Constant.java`

### 2. 🟠 Medium · Robustness — getAndIncrement() correctness silently depends on an undocumented null-return contract of CounterStateStore.incrementAndGet for absent keys
- Root cause: the legacy code used `checkpointIdMap.compute(key, (k, v) -> v == null ? null : v + 1)`, which guaranteed `null` for missing keys so `checkNotNull(nextId)` enforced the "increment before start() is an error" invariant. The new code delegates to `checkpointCounterStore.incrementAndGet(key)`, but the `CounterStateStore` interface javadoc added in this PR documents `initializeIfAbsent` semantics without pinning down what `incrementAndGet` must return for a missing key.
- Impact: any current or future backend (the new `HazelcastCounterStateStore` or a non-Hazelcast implementation) that auto-initializes a missing counter (e.g., to 0 or 1) instead of returning `null` would make `checkNotNull` ineffective, and a pipeline whose counter was removed on end state could silently restart from `INITIAL_CHECKPOINT_ID`. Reused checkpoint IDs mean checkpoint files in shared checkpoint storage get overwritten, which is a data-loss/state-corruption risk on restore.
- Fix: explicitly document in `CounterStateStore#incrementAndGet` that it MUST return `null` (or throw) when the key is absent and MUST NOT auto-initialize, and add a unit test in `HazelcastCounterStateStoreTest` asserting `incrementAndGet` on a missing key returns `null`. Alternatively, replace the nullable-Long contract with a method that throws `IllegalStateException` for absent keys.

**Where (this PR):**
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:92  Long nextId = checkpointCounterStore.incrementAndGet(key);`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:93  checkNotNull(nextId);`

**Suggested fix** (`seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:92`): The old `IMap.compute(key, (k, v) -> v == null ? null : v + 1)` guaranteed `null` for a missing key, which is what makes the `checkNotNull(nextId)` invariant work. Now that this goes through the `CounterStateStore` abstraction, that null-for-absent behavior is an implicit contract — please document it explicitly on `CounterStateStore#incrementAndGet` ("MUST return null and MUST NOT auto-initialize when the key is absent") and add a test asserting it in `HazelcastCounterStateStoreTest`. A backend that auto-initializes would silently reset the checkpoint sequence and cause checkpoint-ID reuse (overwriting checkpoint files in storage).

### 3. 🟠 Medium · Robustness — Counter removal on pipeline end state is not retried, unlike start(), so transient store failures leak counter entries cluster-wide
- Root cause: `start()` wraps `initializeIfAbsent` in `RetryUtils.retryWithException` with `OPERATION_RETRY_TIME`/`OPERATION_RETRY_SLEEP`, but `shutdown()` calls `checkpointCounterStore.remove(key)` bare with no retry and no error handling, then unconditionally returns a completed future.
- Impact: on a long-running Zeta cluster with high job churn, a transient Hazelcast partition-migration or member-leave failure during removal leaves an orphaned entry in the `engine_checkpoint-id-map` store forever (there is no other cleanup path once the pipeline is in an end state), causing unbounded growth of authoritative state that every new leader must replicate; additionally, if `remove` throws a runtime exception it propagates out of `shutdown()` instead of being reflected in the returned `CompletableFuture`.
- Fix: mirror the `start()` pattern — wrap `remove` in `RetryUtils.retryWithException` keyed on `ExceptionUtil::isOperationNeedRetryException`, and complete the returned future exceptionally (or at minimum log with the key) if removal ultimately fails, so orphaned counters are observable.

**Where (this PR):**
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:79  checkpointCounterStore.remove(key);`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:61  RetryUtils.retryWithException(`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:62  () -> checkpointCounterStore.initializeIfAbsent(key, INITIAL_CHECKPOINT_ID),`

**Suggested fix** (`seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:79`): `start()` retries `initializeIfAbsent` via `RetryUtils.retryWithException`, but the terminal-state `remove(key)` here has no retry and no error handling. A transient Hazelcast failure during pipeline teardown permanently leaks this key in the authoritative `engine_checkpoint-id-map` store (no other cleanup path exists once the pipeline ends), which accumulates on high-churn clusters. Please wrap the removal in the same retry material as `start()`, and surface a terminal failure through the returned future or at least a keyed error log.

### 4. 🟠 Medium · Robustness — StateStoreCheckpointIDCounter.get() will throw a raw auto-unboxing NPE if the counter key is absent (e.g., read racing with shutdown-removal during failover)
- Root cause: `get()` returns the boxed result of `checkpointCounterStore.get(key)` directly as a primitive `long`. If the key was never initialized, or was concurrently removed by `shutdown(PipelineStatus)` on end state (the exact window that exists during master failover / pipeline teardown races), the store returns `null` and the method dies with an undiagnosable `NullPointerException` from auto-unboxing.
- Impact: on Zeta, any code path that reads the counter after the pipeline reaches an end state (metrics export, late checkpoint trigger, restore probing on the new master) gets an NPE with no key context, which surfaces as an opaque coordinator error rather than a clear state-store contract violation.
- Fix: null-check the result and throw an `IllegalStateException` including the (jobId, pipelineId) key, e.g. `Long v = checkpointCounterStore.get(key); if (v == null) throw new IllegalStateException("Checkpoint counter not initialized for key " + key); return v;`. This refactor is the right moment to fix it instead of carrying the legacy IMap behavior forward.

**Where (this PR):**
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:99  return checkpointCounterStore.get(key);`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:79  checkpointCounterStore.remove(key);`

**Suggested fix** (`seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/StateStoreCheckpointIDCounter.java:99`): `checkpointCounterStore.get(key)` returns a boxed `Long`, so if the key is absent (never started, or removed by `shutdown()` on end state during a failover/teardown race) this line throws a bare auto-unboxing NPE with no context. Since we're formalizing the contract in this refactor, let's fail explicitly: null-check the result and throw `IllegalStateException("Checkpoint counter not initialized for key " + key)` so operators can diagnose the (jobId, pipelineId) involved.

### 5. 🟢 Low · Test — E2E failover test reaches around the new CounterStateStore abstraction by fetching the raw Hazelcast IMap via EngineStateStoreNames.CHECKPOINT_ID
- Root cause: `CheckpointCoordinatorFailoverIT` verifies checkpoint-ID progression by calling `masterNode2.getMap(EngineStateStoreNames.CHECKPOINT_ID)` and re-implementing the key derivation with `StateStoreCheckpointIDCounter.convertLongIntToBase64(jobId, pipelineId)`, i.e., it hard-codes the Hazelcast-IMap storage layout that this very PR is trying to hide behind `CounterStateStore`.
- Impact: the moment a non-IMap `CounterStateStore` backend is wired in (the stated goal of the #10804 abstraction effort), this E2E test silently reads an empty map and either flakes or gives false negatives, and the internal key-encoding scheme becomes an accidental public contract that can't be changed.
- Fix: assert progression through the abstraction instead — obtain the node's `EngineStateStores` (or `CounterStateStore<String>`) via the server context and call `get(key)` on it, or expose a read-only test accessor; keep `convertLongIntToBase64` usage confined to code that owns the key format.

**Where (this PR):**
- `seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/CheckpointCoordinatorFailoverIT.java:282  IMap<String, Long> ckIdMap = masterNode2.getMap(EngineStateStoreNames.CHECKPOINT_ID);`
- `seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/CheckpointCoordinatorFailoverIT.java:294  StateStoreCheckpointIDCounter.convertLongIntToBase64(`

**Suggested fix** (`seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/CheckpointCoordinatorFailoverIT.java:282`): This still reads the raw Hazelcast IMap by name (`EngineStateStoreNames.CHECKPOINT_ID`) and re-derives the base64 key, which couples the E2E test to the exact storage backend the new `CounterStateStore` abstraction is meant to hide. Once a non-IMap backend is plugged in, this assertion will silently observe an empty map. Could we assert through `engineContext.getStateStores().checkpointCounterStore().get(key)` (or a test accessor on the server) instead, so the test survives backend changes?

### 6. 🟢 Low · Docs — Config/option changes without doc updates
Consider updating docs/example configs for the new/changed options.

_Happy to discuss any of these — thanks again!_

DanielLeens (changes requested):
Sorry for missing this in my previous approval — that’s on me. I rechecked the latest head after @SEZ9’s review, and I agree that there is a real compatibility blocker here. I’m updating Daniel’s technical conclusion from source-clear to changes requested for the current head.

# What this PR is trying to solve
- User pain point: the checkpoint ID counter is coupled directly to the Hazelcast IMap path.
- Fix approach: introduce a `CounterStateStore` / `EngineStateStores` abstraction and move checkpoint ID counter access behind `StateStoreCheckpointIDCounter`.
- In one sentence: the abstraction direction is still good, but the current patch removes a public engine-common symbol.

# Runtime and compatibility path checked
```text
JobMaster / CheckpointManager initialization
  -> EngineStateStores.checkpointCounterStore()
  -> StateStoreCheckpointIDCounter(jobId, pipelineId, store)
      -> key = base64(jobId + pipelineId)
      -> start(): initializeIfAbsent(key, INITIAL_CHECKPOINT_ID)
      -> getAndIncrement(): incrementAndGet(key)
      -> shutdown(end): remove(key)

Downstream/source compatibility path
  -> code depends on seatunnel-engine-common
  -> code references Constant.IMAP_CHECKPOINT_ID
      -> current PR removes that public static final field
      -> downstream source no longer compiles
```

# Issue 1: removing `Constant.IMAP_CHECKPOINT_ID` breaks source compatibility
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/Constant.java:52`
- Problem: `IMAP_CHECKPOINT_ID` currently lives in the public `Constant` class from `seatunnel-engine-common`. This PR removes it and introduces `EngineStateStoreNames.CHECKPOINT_ID` in the server module instead. The map name value is preserved, but the Java symbol is not. Any downstream code, extension, or test utility compiling against `Constant.IMAP_CHECKPOINT_ID` will fail after upgrade.
- Risk: source-level backward compatibility break.
- Suggested fix: keep `Constant.IMAP_CHECKPOINT_ID` as a deprecated alias for the same string while allowing the new internal server code to use `EngineStateStoreNames.CHECKPOINT_ID`. If the project really wants to remove the old symbol, it should go through the incompatible-change documentation and migration path instead.
- Severity: high, merge-blocking.
- Already raised by another reviewer: yes, +1 to @SEZ9’s compatibility concern here.

# Notes on the other review points
I also rechecked the counter-store contract and tests:
- `CounterStateStore#incrementAndGet` currently documents `null` for absent keys, and `HazelcastCounterStateStoreTest.absentCounterShouldStayAbsent()` asserts that missing counters remain absent.
- `shutdown()` removal without retry and `get()` producing the legacy absent-key NPE are worth considering as robustness improvements, but they are not the same compatibility regression as the public constant removal because the old `IMapCheckpointIDCounter` had the same removal/no-retry and absent-get behavior.
- The E2E still reaches into the raw map name, which is not ideal for a backend abstraction, but I would treat that as a lower-priority test maintainability follow-up unless maintainers want this PR to hide the backend completely now.

- Checked the diff for `Constant.java` and `EngineStateStoreNames.java`: the public constant is removed from engine-common and reintroduced under engine-server.
- Checked current counter tests: absent-key increment behavior is covered for the Hazelcast backend.
- I did not run local tests in this follow-up; this is a code-level compatibility review.

# Merge conclusion
### Conclusion: can merge after changes
Blocking items:
1. Keep `Constant.IMAP_CHECKPOINT_ID` source-compatible, preferably as a deprecated alias, or document this as an incompatible change with migration guidance.

Non-blocking suggestions:
- Consider retry/error surfacing for terminal `remove(key)` and a clearer absent-key error for `get()` in a follow-up.
- Consider avoiding raw IMap access in the failover E2E if the abstraction goal is to keep tests backend-neutral.

Overall, the state-store abstraction still makes sense, but this compatibility gap should be fixed before merge.

dybyte:
@SEZ9 @DanielLeens Thanks, updated

DanielLeens (approved):
Thanks for the update! I re-reviewed the latest head after the new commits, and both blockers from my previous round and @SEZ9's review are now resolved:

1. `Constant.IMAP_CHECKPOINT_ID` is preserved with `@Deprecated` — source compatibility is maintained for downstream code that references it.

2. The `getAndIncrement` null-return contract is now explicit: `CounterStateStore.incrementAndGet` documents that implementations must not auto-initialize and must return `null` for absent keys, and `StateStoreCheckpointIDCounter.getAndIncrement` enforces this with `checkNotNull`.

The abstraction direction is clean — `CounterStateStore` provides a focused increment-only interface rather than a generic `compute()`, and the `EngineStateStores` access point keeps the wiring centralized. Tests cover the counter lifecycle and the Hazelcast-backed atomic operations well.

CI is all green on the current head. Looks good to merge from my side.

nzw921rx (approved):
+1 good job

DanielLeens:
Thanks @nzw921rx. I’m aligned with the current head here as well.

From Daniel's side I do not see a new source-level blocker reopening on the unchanged revision. I’m only keeping the live upstream-sync fact in the local queue for merge-readiness tracking; I am not asking for another code update based on this follow-up alone.

DanielLeens (approved):
Rechecked the latest head: my previous source-level review had already cleared the code-side blocker, and the only remaining gate I called out was CI.

The required checks are now green for the current head (`1307adc3d82953cea2bf786c206352163dcbc906`), so I do not see a remaining Daniel-side blocker. Approving this round. Nice work keeping the checkpoint-id state-store change focused.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=7`, `behind_by=137`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

DanielLeens:
Rechecked the same unchanged head `1307adc3d829` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
JobMaster / CheckpointManager initialization
  -> EngineStateStores.checkpointCounterStore()
  -> StateStoreCheckpointIDCounter(jobId, pipelineId, store)
      -> key = base64(jobId + pipelineId)
      -> start(): initializeIfAbsent(key, INITIAL_CHECKPOINT_ID)
      -> getAndIncrement(): incrementAndGet(key)
      -> shutdown(end): remove(key)

Downstream/source compatibility path
  -> code depends on seatunnel-engine-common
  -> code references Constant.IMAP_CHECKPOINT_ID
      -> current PR removes that public static final field
      -> downstream source no longer compiles
```

## Merge conclusion

1. Blocking items

2. Current gate
- The current CI signal is now green: https://github.com/apache/seatunnel/actions/runs/29091951832/job/86359158483

davidzollo (approved):
+1
LGTM

DanielLeens (commented):
Gate follow-up on the unchanged head: the required `Build` remains green and a write-capable maintainer approval is now recorded, so the earlier “waiting for final approval” note is no longer current. I still do not see a source-side blocker on this revision.

GitHub currently reports `mergeStateStatus=UNSTABLE` even though the visible `Build` and review-decision gates are satisfied. The branch is also behind `dev`, but behindness alone is not evidence that the contributor should make another sync-only commit while CI is green. The technical verdict therefore remains source-clear, waiting only for a write-capable maintainer to inspect the remaining repository-level unstable gate and complete the merge decision.

yzeng1618 (approved):
LGTM +1

PR #11233, opened 2026-07-01
[Test][E2E] Avoid SQLServer CDC driver download in container

DanielLeens:
### Purpose
Fix an unrelated SQLServer CDC E2E CI failure where the test container downloads the SQLServer JDBC driver with `wget` from Maven Central during runtime. That path can fail with transient TLS/network errors such as `Unable to establish SSL connection`, even though Maven has already resolved the same driver for the module.

### Changes
- Reuse the Maven-resolved `mssql-jdbc-9.4.1.jre8.jar` from the local repository.
- Copy the driver into `/tmp/seatunnel/plugins/SqlServer-CDC/lib` via Testcontainers instead of downloading it inside the container.
- Keep the driver version unchanged and do not modify `pom.xml`.

### Validation
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-sqlserver-e2e spotless:apply -nsu -Dmaven.gitcommitid.skip=true`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home ./mvnw -B -pl :connector-cdc-sqlserver-e2e -DskipTests -DskipITs=true -Dlicense.skipAddThirdParty=true -Dskip.ui=true -Dskip.spotless=true -Dmaven.gitcommitid.skip=true -nsu test-compile`

Full Testcontainers E2E was not run locally because Docker daemon is not available on this machine (`Cannot connect to the Docker daemon`). `seatunnel-engine-ui` was not built because this change only touches a connector E2E Java test.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-sqlserver-e2e/src/test/java/org/apache/seatunnel/e2e/connector/cdc/sqlserver/SqlServerCDCIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-sqlserver-e2e/src/test/java/org/apache/seatunnel/e2e/connector/cdc/sqlserver/SqlServerCDCIT.java
@@ -45,6 +45,7 @@ import org.testcontainers.containers.MSSQLServerContainer;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.utility.DockerLoggerFactory;
+import org.testcontainers.utility.MountableFile;
 
 import io.debezium.jdbc.JdbcConnection;
 import io.debezium.relational.TableId;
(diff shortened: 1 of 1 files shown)


DanielLeens (commented):
Thanks for the update. I reviewed the latest head from the maintainer side.

This PR is fixing a real E2E setup problem in the SQL Server CDC tests.

Before this change, the test container extension prepared the JDBC dependency by running `wget` inside the container. That means the test result still depended on runtime network reachability and external repository availability, even though the driver was already a declared test dependency.

The new flow in
`seatunnel-e2e/.../SqlServerCDCIT.java:201-233`
is much more robust:

```text
SQL Server CDC E2E startup
  -> resolve SQLServerDriver.class from the active test classpath
  -> convert its code source to a local jar path
  -> assert the resolved path is a real file
  -> copy that jar into /tmp/seatunnel/plugins/SqlServer-CDC/lib
```

So this removes the container-time external download dependency and ties the test setup to the dependency Maven/CI already resolved for the test itself.

I also checked the stability angle:

- this reduces environment flakiness rather than adding a weaker fallback
- the failure mode becomes earlier and clearer if the driver is missing from the test classpath
- no extra shared global state, timing wait, or resource-lifecycle issue is introduced

I did not find a source-side blocker in the current version.

Merge conclusion: can merge.

Local review evidence:
- `gh pr view 11233 --repo apache/seatunnel --json ...` -> Build is passing on head `6f1626f7879c4168ad16d068cce98935223713d5`
- `gh api repos/apache/seatunnel/compare/dev...6f1626f7879c4168ad16d068cce98935223713d5` -> `status=diverged`, `ahead_by=3`, `behind_by=4`
- `git diff --stat apache/dev...HEAD` in local review worktree -> 1 file changed, `SqlServerCDCIT.java`
- no local compile/test run; this review was completed from the real diff and call-chain analysis only

davidzollo (approved):
+1

PR #11234, opened 2026-07-01
[Test][E2E] Retry RocketMQ restore offset checks

DanielLeens:
### Purpose
Fix an unrelated RocketMQ restore E2E CI failure where the final max-offset assertion can read `0` from a transient RocketMQ admin query even after earlier polling observed the sink topic reaching the expected offset.

### Changes
- Replace the fixed post-restore sleep plus one-shot offset read with an Awaitility helper that returns the successful observed offset.
- Retry `RocketMqAdminUtil.offsetTopics` because admin queries can temporarily fail during restore and broker channel transitions.
- Keep the production connector code and dependencies unchanged. No `pom.xml` changes.

### Validation
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-rocketmq-e2e spotless:apply -nsu -Dmaven.gitcommitid.skip=true`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home ./mvnw -B -pl :connector-rocketmq-e2e -DskipTests -DskipITs=true -Dlicense.skipAddThirdParty=true -Dskip.ui=true -Dskip.spotless=true -Dmaven.gitcommitid.skip=true -nsu test-compile`

Full Testcontainers E2E was not run locally because Docker daemon is not available on this machine (`Cannot connect to the Docker daemon`). `seatunnel-engine-ui` was not built because this change only touches a connector E2E Java test.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-rocketmq-e2e/src/test/java/org/apache/seatunnel/e2e/connector/rocketmq/RocketMqIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-rocketmq-e2e/src/test/java/org/apache/seatunnel/e2e/connector/rocketmq/RocketMqIT.java
@@ -88,6 +88,7 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
 import static org.apache.seatunnel.e2e.connector.rocketmq.RocketMqContainer.NAMESRV_PORT;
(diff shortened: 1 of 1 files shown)


davidzollo (commented):
+1
LGTM
Thanks for your contribution!

davidzollo (approved):

DanielLeens (commented):
Thanks for the update. I reviewed the latest head from the maintainer side.

This test change is addressing a real RocketMQ restore flake.

Before this PR, the flow was:

```text
restore test
  -> wait until sink max offset reaches the lower bound
  -> sleep 5 seconds
  -> read RocketMQ admin offsets once
  -> assert the final sink offset
```

That still leaves a race against broker-side offset visibility. A single admin read after a fixed sleep can be stale even when the restored write path is already correct.

The new version improves that in two layers inside
`seatunnel-e2e/.../RocketMqIT.java:651-769`:

- `awaitTopicMaxOffset(...)` uses `Awaitility` to wait for the final expected offset instead of depending on one fixed-delay read
- `getTopicMaxOffset(...)` now wraps `RocketMqAdminUtil.offsetTopics(...)` with `RetryUtils.retryWithException(...)`, which is a much better fit for temporary admin-query instability during restore / broker channel transitions

So from the actual call path, this is not weakening the assertion. It is making the test wait for the state it already intends to verify.

I also checked the test stability angle:

- the new wait replaces fixed sleep with condition-based polling
- the retry is narrowly scoped to the admin offset read
- no extra shared global state or resource-lifecycle issue is introduced

I did not find a source-side blocker in the current version.

Merge conclusion: can merge.

Local review evidence:
- `gh pr view 11234 --repo apache/seatunnel --json ...` -> Build is passing on head `205fe76168a5af1c5de28e41f96b19a80fcf1373`
- `gh api repos/apache/seatunnel/compare/dev...205fe76168a5af1c5de28e41f96b19a80fcf1373` -> `status=diverged`, `ahead_by=1`, `behind_by=4`
- `git diff --stat apache/dev...HEAD` in local review worktree -> 1 file changed, `RocketMqIT.java`
- no local compile/test run; this review was completed from the real diff and call-chain analysis only

PR #11238, opened 2026-07-01
[Feature][Flink] Support MetadataProvider SPI in Flink starter

chl-wxp:
### Purpose

  This PR adds MetadataProvider SPI support for the Flink starter, so Flink jobs can resolve datasource connection options from the metadata center by using `metadata_datasource_id`.

  Before this change, Flink jobs with configs such as:

  ```hocon
  source {
    Jdbc {
      metadata_datasource_id = "test_catalog"
      query = "select * from source"
    }
  }
  ```

  could fail during connector option validation because required JDBC options like url and driver were not resolved before FactoryUtil created the source.

  ### Changes

  - Load metadata center configuration from seatunnel.yaml in the Flink starter.
  - Resolve datasource configs through MetadataProviderManager before applying legacy Metalake config handling.
  - Keep compatibility with existing Metalake placeholder replacement logic.
  - Update JDBC e2e coverage for MetadataProvider SPI on Flink.
  - Ensure the Flink e2e job process has SEATUNNEL_HOME so it can locate config/seatunnel.yaml.
  - Keep Metalake legacy env behavior using METALAKE_ENABLED, METALAKE_TYPE, and METALAKE_URL.

  ### Why

  The MetadataProvider SPI expects metadata settings from:

  ```yaml
  seatunnel:
    engine:
      metadata:
        enabled: true
        kind: gravitino
        gravitino:
          uri: http://gravitino:8090
          metalake: test_metalake
  ```

  Flink starter needs this config before connector validation, otherwise datasource options are not merged into the job config.

### DIFF
--- a/docs/en/introduction/concepts/metadata-spi.md
+++ b/docs/en/introduction/concepts/metadata-spi.md
@@ -172,13 +172,15 @@ To enable the Metadata Center, add the following configuration to `seatunnel.yam
 ```yaml
 seatunnel:
   engine:
-     metadata:
+    metadata:
       enabled: true
       kind: gravitino
       gravitino:
         uri: http://127.0.0.1:8090
         metalake: test_metalake
 ```
+
+SeaTunnel Zeta and the Flink starter read this configuration from `$SEATUNNEL_HOME/config/seatunnel.yaml`.
 
 ### Configuration Options
 
--- a/docs/zh/introduction/concepts/metadata-spi.md
+++ b/docs/zh/introduction/concepts/metadata-spi.md
@@ -171,13 +171,15 @@ public interface MetadataProvider extends AutoCloseable {
 ```yaml
 seatunnel:
   engine:
-     metadata:
+    metadata:
       enabled: true
       kind: gravitino
       gravitino:
         uri: http://127.0.0.1:8090
         metalake: test_metalake
 ```
+
+SeaTunnel Zeta 和 Flink starter 会从 `$SEATUNNEL_HOME/config/seatunnel.yaml` 读取该配置。
 
 ### 配置选项
 
--- a/seatunnel-core/seatunnel-core-starter/src/main/java/org/apache/seatunnel/core/starter/utils/ConfigBuilder.java
+++ b/seatunnel-core/seatunnel-core-starter/src/main/java/org/apache/seatunnel/core/starter/utils/ConfigBuilder.java
@@ -26,6 +26,8 @@ import org.apache.seatunnel.shade.com.typesafe.config.ConfigSyntax;
 import org.apache.seatunnel.shade.com.typesafe.config.impl.Parseable;
 
 import org.apache.seatunnel.api.configuration.ConfigAdapter;
+import org.apache.seatunnel.api.metadata.MetadataConfig;
+import org.apache.seatunnel.api.metadata.MetadataOptions;
 import org.apache.seatunnel.api.sink.TablePlaceholder;
 import org.apache.seatunnel.common.utils.JsonUtils;
 import org.apache.seatunnel.common.utils.ParserException;
@@ -34,9 +36,11 @@ import org.apache.seatunnel.core.starter.exception.ConfigCheckException;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -63,10 +67,15 @@ public class ConfigBuilder {
     }
 
     private static Config ofInner(@NonNull Path filePath, List<String> variables) {
+        Config config = parseConfigFile(filePath);
+        return ConfigShadeUtils.decryptConfig(backfillUserVariables(config, variables));
+    }
+
+    private static Config parseConfigFile(@NonNull Path filePath) {
         Config config =
                 ConfigFactory.parseFile(filePath.toFile())
                         .resolve(ConfigResolveOptions.defaults().setAllowUnresolved(true));
-        return ConfigShadeUtils.decryptConfig(backfillUserVariables(config, variables));
+        return config;
     }
 
     public static Config of(@NonNull String filePath) {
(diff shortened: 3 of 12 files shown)


DanielLeens (changes requested):
Thanks for working on this. I re-reviewed the latest head against the full diff and the real Flink starter call chain, not just the E2E delta.

# What this PR solves
- User pain: Flink jobs that use `metadata_datasource_id` still fail connector validation because the datasource config is not resolved before the connector factory checks required options like `url` and `driver`.
- Fix approach: the PR reads metadata provider config from `seatunnel.yaml`, resolves datasource configs through `MetadataProviderManager`, and only then continues with the existing Metalake compatibility path.
- One-line summary: the direction is right and the Flink starter now reaches the Metadata SPI path, but the current implementation still has one real config-path correctness gap and one high-risk E2E stability problem.

# Runtime chain I re-traced
```text
FlinkTaskExecuteCommand.execute()
  -> ConfigBuilder.parseMetadataConfigFromSeatunnelYaml()
     -> read $SEATUNNEL_HOME/config/seatunnel.yaml
     -> parse seatunnel.engine.metadata
  -> ConfigBuilder.of(configFile, variables)
     -> load and decrypt job config
  -> MetadataProviderManager.resolveDataSourceConfigs(jobConfig, metadataConfig)
     -> scan source/sink configs
     -> resolve metadata_datasource_id through provider.datasourceMap(...)
     -> merge url / username / password / driver back into connector config
  -> MetalakeConfigUtils.getMetalakeConfig(resolvedConfig)
     -> keep the legacy METALAKE_* placeholder path
  -> new FlinkExecution(config).execute()
```

# Key findings
1. The normal Flink submission path definitely hits this PR. The metadata resolution is now inserted before connector validation, which is the right place.
2. The weak point is the new handwritten `seatunnel.yaml` parser. It is not order-independent like a real YAML parser.
3. The new E2E direction is good, but `apache/gravitino:latest` makes the test outcome depend on external image drift.
4. The docs are only partially updated: the same concept page still says Metadata SPI is Zeta-only even though this PR is adding Flink starter support.

# Findings
Problem 1: the new metadata YAML parser is still order-sensitive on the real Flink submission path
- Location: `seatunnel-core/seatunnel-core-starter/src/main/java/org/apache/seatunnel/core/starter/utils/ConfigBuilder.java:375-425`
- Problem description:
  This is in the new Flink metadata bootstrap path. `providerKind` starts from the default value and is only updated after the parser reaches `kind:`. But provider-section detection happens during the same single pass and depends on the current `providerKind`. So a valid YAML layout where the provider block appears before `kind:` can still lose all provider properties.
- Potential risk:
  That is a main-path correctness issue, not just a cosmetic parser detail. The Flink starter can advertise Metadata SPI support while still silently dropping a valid provider configuration shape.
- Best fix:
  Option A, recommended: reuse a real YAML parser / existing structured config reader for `seatunnel.engine.metadata`.
  Option B: if you want to keep this lightweight approach for now, parse the whole metadata block into a map first and then resolve the provider subsection after `kind` is known. Please also add regression coverage for provider-before-kind ordering and values containing `#`.
- Severity: High

Problem 2: the new E2E uses `apache/gravitino:latest`, which makes the test inherently drift-prone
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/MetalakeIT.java:99,202-214`
- Problem description:
  This is the main proof path for the new Flink support. The test now starts a dedicated Gravitino container, but it pins the image to `latest` and then depends on the current container startup behavior and `/api/metalakes` API contract.
- Potential risk:
  That is a classic flaky-test pattern. The CI result can change without any SeaTunnel code change, just because the external `latest` image changed.
- Best fix:
  Please pin Gravitino to a verified version tag or digest. If multi-version coverage is needed, it should be done explicitly, not via `latest`.
- Severity: High

Problem 3: the concept docs still say Metadata SPI is only supported on Zeta
- Location: `docs/en/introduction/concepts/metadata-spi.md:22`, `docs/zh/introduction/concepts/metadata-spi.md:22`
- Problem description:
  The page now also says later that the Flink starter reads metadata config from `$SEATUNNEL_HOME/config/seatunnel.yaml`, but the top-level engine support note still says Metadata SPI is only supported on Zeta and not compatible with Flink.
- Potential risk:
  This will directly confuse users because the same document now gives two opposite answers about Flink support.
- Best fix:
  Please update the engine-support note in both EN and ZH to state the actual current support boundary.
- Severity: Medium

# Code quality / tests / stability
- The new coverage is heading in the right direction, but it still misses the critical parser regression cases:
  - provider block before `kind`
  - provider values containing `#` or other characters that must survive parsing intact
- Test stability rating: High risk
  - Evidence: `MetalakeIT.java:99,202-214`
  - Reason: this is not a `Thread.sleep` problem, but it is still a real flaky pattern because the test depends on a moving `latest` image.

1. Blocking items
- Problem 1: make the metadata config parsing robust on the real Flink path.
- Problem 2: pin the new Gravitino E2E dependency to a stable image version.

- Problem 3: align the EN/ZH Metadata SPI docs with the new Flink support boundary.

Overall, I like the direction of the change. The Flink starter is finally wired into the Metadata SPI flow, but I would still fix the parser robustness and the E2E drift risk before merging.

chl-wxp:
@DanielLeens PTAL

DanielLeens (changes requested):
Thanks for the ping. I re-reviewed the latest head from scratch again against the full diff and the real Flink starter call chain.

## What this PR solves
- User pain: Flink jobs that use `metadata_datasource_id` still fail connector validation because datasource config is not resolved before the connector factory checks required options like `url` and `driver`.
- Fix approach: read metadata provider config from `seatunnel.yaml`, resolve datasource configs through `MetadataProviderManager`, and then continue with the existing starter path.
- One-line summary: the direction is still right, but the current revision still has one real config-path correctness gap and one high-risk E2E stability problem.

## Runtime chain I re-traced
```text
FlinkTaskExecuteCommand.execute()
  -> ConfigBuilder.parseMetadataConfigFromSeatunnelYaml()
     -> read $SEATUNNEL_HOME/config/seatunnel.yaml
     -> parse seatunnel.engine.metadata
  -> ConfigBuilder.of(configFile, variables)
     -> load and decrypt job config
  -> MetadataProviderManager.resolveDataSourceConfigs(jobConfig, metadataConfig)
     -> resolve metadata_datasource_id into connector config
  -> MetalakeConfigUtils.getMetalakeConfig(resolvedConfig)
  -> new FlinkExecution(config).execute()
```

## Current-head findings
Problem 1: the handwritten metadata YAML parser is still order-sensitive on the real Flink submission path
- Location: `seatunnel-core/seatunnel-core-starter/src/main/java/org/apache/seatunnel/core/starter/utils/ConfigBuilder.java:368-449`
- Why this is still a blocker:
  The provider subsection is only recognized when `providerKind.equalsIgnoreCase(section)` is already true at `ConfigBuilder.java:410-417`, but `providerKind` is updated later when the parser reaches `kind:` at `ConfigBuilder.java:421-426`. So a valid YAML layout where the provider block appears before `kind:` can still silently lose all provider properties. The same helper also strips everything after `#` at `ConfigBuilder.java:443-449`, including quoted values.
- Risk:
  This is a main-path correctness issue, not a cosmetic parser detail. The Flink starter can claim Metadata SPI support while still dropping a valid provider configuration shape.
- Best fix:
  Prefer a real YAML parser or, at minimum, parse the whole metadata block first and resolve the provider subsection only after `kind` is known. Please also add regression coverage for provider-before-kind ordering and quoted values containing `#`.
- Severity: High

Problem 2: the new E2E still depends on `apache/gravitino:latest`, which keeps the test drift-prone
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/MetalakeIT.java:99-100,202-214`
- Why this is still a blocker:
  The current proof path for the new Flink support still boots a dedicated Gravitino container from `apache/gravitino:latest`. That means the CI result can change without any SeaTunnel code change, just because the upstream image moved.
- Risk:
  This is a high-risk flaky-test pattern on the exact E2E path meant to validate the feature.
- Best fix:
  Please pin Gravitino to a verified version tag or digest instead of `latest`.
- Severity: High

Problem 3: the Metadata SPI concept docs still contradict the new Flink support claim
- Location: `docs/en/introduction/concepts/metadata-spi.md:20-22`, `docs/zh/introduction/concepts/metadata-spi.md:20-22`
- Why this still matters:
  Both pages still say Metadata SPI is only supported on Zeta and not compatible with Flink or Spark, while this PR is explicitly adding Flink starter support.
- Risk:
  Users will get two opposite answers from the same document set.
- Best fix:
  Update the engine-support note in both EN and ZH to match the actual support boundary introduced by this PR.
- Severity: Medium

## Test stability
- Stability rating: High risk
- Basis: `MetalakeIT.java:99-100,202-214`
- Reason: the problem is not `Thread.sleep`, but the E2E result still depends on a moving `latest` container image, which is exactly the kind of external drift that causes CI noise over time.

## Merge conclusion

1. Blocking items
- Problem 1: make the metadata config parsing robust on the real Flink starter path.
- Problem 2: pin the Gravitino E2E dependency to a stable image version.

- Problem 3: align the EN/ZH Metadata SPI docs with the new Flink support boundary.

I like the overall direction of the change, and the Flink starter is much closer now. Once the parser robustness and the E2E drift risk are fixed, I will be happy to take another look.

DanielLeens:
Thanks for the ping. I rechecked the current gate state on top of my earlier Flink review.

My earlier source-side blockers still stand, but the current `Build` failure is also mixed with a branch that is now `ahead_by=3`, `behind_by=50`, compare status `diverged`. The failing run currently spans broad lanes like `unit-test`, `kafka-connector-it`, and `jdbc-connectors-it-part-7`:
https://github.com/chl-wxp/seatunnel-laowang/actions/runs/28647227708

Given that spread, the lowest-cost next step is to sync the latest `dev` and rerun CI before spending more time chasing the current red matrix. After the rebase, if the same lanes still fail, I can help separate the remaining CI noise from the source-side items I already listed.

DanielLeens:
Thanks for the update. I checked the current head against the latest `dev` again.

- base branch: `dev`
- head SHA: `39ab42e22f88aeb6332bf1db75c24e1948d5bc41`
- compare status: `diverged`
- ahead_by: `3`
- behind_by: `58`
- current failing signal: `Build`

On this round, the visible failing CI signal still does not look directly tied to the current PR-owned diff from Daniel's side, and the branch is still behind upstream. The lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the same lane still fails on the updated head afterwards, please drop the new failing check link here and I can help narrow it down against the rebased revision.

DanielLeens (commented):
Outdated.

DanielLeens:

PR #11241, opened 2026-07-01
[Docs][Connector-V2] Improve Aerospike sink example

DanielCarter-stack:
### Purpose

Improve the Aerospike sink documentation by aligning the task example with the existing e2e job configuration.

### Changes

- Update the English and Chinese Aerospike sink examples to include `bin_name`, `data_format`, and field type mappings used by the e2e FakeSource-to-Aerospike task.
- Clarify how `data_format` stores non-key fields and when `bin_name` is required.

### Verification

- `./mvnw spotless:apply`

### DIFF
--- a/docs/en/connectors/sink/Aerospike.md
+++ b/docs/en/connectors/sink/Aerospike.md
@@ -61,35 +61,49 @@ Note:
 | username       | string | No       | -       | Username for authentication                                                |
 | password       | string | No       | -       | Password for authentication                                                |
 | key            | string | Yes      | -       | Field name to use as Aerospike primary key                                 |
-| bin_name       | string | No       | -       | Bin name for storing data                                                  |
+| bin_name       | string | No       | -       | Bin name for storing data. Required when `data_format` is `map` or `string` |
 | data_format    | string | No       | string  | Data storage format: map/string/kv                                         |
 | write_timeout  | int    | No       | 200     | Write operation timeout in milliseconds                                    |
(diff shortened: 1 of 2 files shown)


DanielLeens (approved):
Thanks for improving the Aerospike sink docs. I checked the latest head against the actual sink option contract and the writer behavior.

Runtime path I checked
```text
Aerospike sink option contract
  -> AerospikeSinkFactory.optionRule() [AerospikeSinkFactory.java:38-52]
      -> host / port / namespace / set are the required options
      -> username / password / key / bin_name / data_format / write_timeout are optional

Aerospike write path
  -> AerospikeSinkWriter.write() [AerospikeSinkWriter.java:77-146]
      -> MAP and STRING both write through bin_name
      -> KV writes each non-key field as its own bin
  -> AerospikeSinkOptions.FIELD_TYPES [AerospikeSinkOptions.java:101-110]
      -> schema.field is the real field-type mapping contract
```

What I found
- The new explanation around `data_format` is aligned with the real write path: `map` and `string` both depend on `bin_name`, while `kv` does not.
- The refreshed example also matches the actual `schema.field` contract and the supported field-type mapping path.
- I did not see a reopened code-level or documentation-contract blocker on this head.

Conclusion
- From Daniel's side, this PR looks good to merge.
- The current `Build` checks are green, so I do not have a remaining CI gate to hold on this revision.

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

PR #11244, opened 2026-07-01
[Improve][Zeta] Exclude runtime metrics snapshots from IMAP persistence

dybyte:

This PR excludes `engine_runningJobMetrics` from persistent IMAP storage.

`engine_runningJobMetrics` stores high-frequency runtime metrics snapshots used for observability. Persisting this map through the generic `map.engine*` `map-store` configuration causes unnecessary WAL growth and write amplification, while providing limited recovery value compared with authoritative engine state.

This change keeps the existing `map-store` behavior for other engine maps, but skips persistence initialization for the runtime metrics snapshot map.

Yes.

When IMAP persistence is enabled through `map.engine*`, `engine_runningJobMetrics` is now intentionally excluded from persistent storage. This changes persistence behavior for runtime metrics snapshots and reduces WAL growth for observability-only state.

The deployment documentation was also updated in both English and Chinese to clarify this behavior.

- Added a unit test in `FileMapStoreTest` to verify that:
  - `engine_runningJobMetrics` skips persistent storage initialization
  - it does not fail even when the configured storage type is invalid
  - store/delete/load operations remain safe as no-op behavior

### Check list


DanielLeens (approved):
Thanks for the contribution. I re-reviewed the latest head against the actual Zeta state-store path, not only the small diff in `FileMapStore`.

# What this PR solves
- User pain: when `map.engine*` persistence is enabled, `engine_runningJobMetrics` also gets persisted even though it only stores high-frequency runtime metrics snapshots. That creates unnecessary WAL growth and write amplification.
- Fix approach: the PR special-cases `engine_runningJobMetrics` in `FileMapStore.init(...)` and routes it to a no-op persistence backend instead of initializing a real map-store backend.
- One-line summary: this is a focused fix that removes persistence pressure from an auxiliary observability map without changing the authoritative job-state path.

# Runtime chain I re-traced
```text
HazelcastEngineStateStores.ensureInitialized()
  -> getMap("engine_runningJobMetrics")
  -> new HazelcastMetricsSnapshotStateStore(...)

SeaTunnelServer.updateMetrics(localMap)
  -> metricsSnapshotStateStore.merge(localMap)
  -> keep the latest runtime snapshots in the in-memory Hazelcast map

CoordinatorService.getRunningJobMetrics()
  -> collect live worker metrics through GetMetricsOperation
  -> aggregate RawJobMetrics
```

The important point is that the public running-metrics query path is still live-worker based. It does not depend on replaying persisted `engine_runningJobMetrics` entries from map-store.

# Key findings
1. The change is definitely hit on the normal path when users enable `map.engine*` persistence.
2. The PR only skips persistent backend initialization for `engine_runningJobMetrics`; it does not remove the in-memory map itself.
3. I do not see this map being used as an authoritative recovery source. It is auxiliary observability state.
4. The EN/ZH deployment docs were updated consistently with the new behavior.

# Findings
I did not find a blocking issue in the current version.

# Code quality / tests / stability
- The new unit test covers the important branch: skipping persistence initialization for `engine_runningJobMetrics` while keeping store/delete/load calls safe as no-ops.
- Test stability rating: Stable
  - Evidence: `FileMapStoreTest.java:106-123`
  - Reason: this is a pure unit test with no timing, container, or environment drift risk.

1. Blocking items
- None from the code path I re-checked.

- If you want to harden this area further later, an additional higher-level state-store behavior test after restart could be useful, but I do not see that as a merge blocker for this PR.

Overall, this looks like a good targeted cleanup. It reduces persistence pressure where the state is auxiliary, while keeping the authoritative execution and metrics-query paths intact.

davidzollo:
Thanks for the improvement. Excluding `engine_runningJobMetrics` from IMAP persistence looks like a reasonable direction to avoid WAL growth for high-frequency observability-only snapshots.

One behavior worth making explicit: after an engine restart, recovered tasks will call `TaskExecutionContext#getOrCreateMetricsContext(...)`. If the previous `engine_runningJobMetrics` snapshot is no longer restored, this path falls back to a new `SeaTunnelMetricsContext`, so the running-job metrics shown after restart will be rebuilt from subsequent reports rather than continuing from the pre-restart snapshot. That seems acceptable if this map is intentionally treated as auxiliary runtime state, but it may be worth documenting this reset/rebuild behavior more directly.

One implementation detail I would suggest tightening: `FileMapStore` now initializes `mapStorage` to `NoOpMapStorage.INSTANCE` by default. To keep the no-op behavior strictly scoped to `RUNNING_JOB_METRICS`, would it be safer to assign the no-op storage only inside the `RUNNING_JOB_METRICS` branch and keep other maps fail-fast if storage initialization did not complete? That avoids accidentally hiding initialization failures for persistent engine maps.

DanielLeens:
Thanks for calling that out. I rechecked the recovery path before replying.

On restart, `TaskExecutionContext.getOrCreateMetricsContext(...)` reads from `metricsSnapshotStore().getOrDefault(taskLocation, new SeaTunnelMetricsContext())`. With this PR, `engine_runningJobMetrics` is intentionally no longer persisted through `FileMapStore`, so after an engine restart the recovered task will recreate a fresh empty metrics context instead of restoring the previous runtime snapshot.

From Daniel's side, that behavior is consistent with the current scope of this change: `engine_runningJobMetrics` is being treated as observability-only runtime snapshot state, not execution state needed for correctness or recovery. So I would not reopen this PR on that point. If SeaTunnel later wants metrics snapshots to survive restart as a first-class feature, that would need a separate design with explicit retention and recovery semantics rather than piggybacking on the current IMAP persistence path.

dybyte:
@davidzollo Thanks for the review, updated

DanielLeens (changes requested):
Thanks for the update, and sorry for missing this documentation gap in the previous round. I re-reviewed the latest head after the new docs commit. The source-side behavior still looks correct to me, but I found one remaining docs consistency issue that should be fixed before merging.

# What this PR solves
- User pain: when `map.engine*` MapStore persistence is enabled, `engine_runningJobMetrics` can generate high-frequency persistence writes even though it only carries runtime metrics snapshots.
- Fix approach: `FileMapStore.init(...)` now routes `EngineStateStoreNames.RUNNING_JOB_METRICS` to a no-op storage backend, while keeping the in-memory Hazelcast IMap available for runtime metrics aggregation.
- One-line summary: this is a focused cleanup that removes persistent IMAP writes for auxiliary observability state without changing authoritative job/pipeline recovery state.

# Runtime chain I re-traced
```text
Zeta engine state-store initialization
  -> HazelcastEngineStateStores.ensureInitialized() [65-68]
      -> nodeEngine.getHazelcastInstance().getMap(RUNNING_JOB_METRICS)
      -> new HazelcastMetricsSnapshotStateStore(metrics IMap, partitionCount)

MapStore initialization when map.engine* persistence is enabled
  -> FileMapStore.init(..., mapName) [45-53]
      -> mapName == EngineStateStoreNames.RUNNING_JOB_METRICS
      -> mapStorage = NoOpMapStorage.INSTANCE
      -> skip FactoryUtil.discoverFactory(...) and do not create a real persistent IMapStorage

Runtime metrics updates
  -> SeaTunnelServer.updateMetrics(localMap) [389-392]
      -> metricsSnapshotStateStore.merge(localMap)
      -> HazelcastMetricsSnapshotStateStore.merge(...) [65-89]
          -> metricsImap.compute(...) updates the in-memory Hazelcast map
          -> any MapStore persistence callback lands on NoOpMapStorage

After engine restart
  -> TaskExecutionContext.getOrCreateMetricsContext(taskLocation) [62-66]
      -> metricsSnapshotStore().getOrDefault(taskLocation, new SeaTunnelMetricsContext())
      -> previous metrics snapshots are not restored from persistent IMAP storage
      -> running-job metrics are rebuilt from later reports
```

# Finding

## Issue 1: The state-storage/recovery docs still say running-job metrics are persisted
- Location: `docs/en/engines/zeta/state-storage-and-recovery.md:132`, `docs/en/engines/zeta/state-storage-and-recovery.md:138`, `docs/en/engines/zeta/state-storage-and-recovery.md:145`; `docs/zh/engines/zeta/state-storage-and-recovery.md:126`, `docs/zh/engines/zeta/state-storage-and-recovery.md:131`, `docs/zh/engines/zeta/state-storage-and-recovery.md:138`
- Problem: the new deployment docs now correctly say that `engine_runningJobMetrics` is excluded from persistent IMAP storage. However, the dedicated state-storage/recovery docs still say the logical maps are persisted and still list running-job metrics as MapStore-backed state. That page is the place users are most likely to read when checking restart/recovery semantics, so it now conflicts with both the code path and the new deployment note.
- Risk: users may still believe running-job metrics continue from the pre-restart snapshot, or use the old metrics persistence assumption for capacity and recovery planning. This does not break job correctness, but it does leave the public contract inconsistent.
- Suggested fix: please update both EN and ZH `state-storage-and-recovery.md` to either remove running-job metrics from the persisted-state list or explicitly mark it as in-memory IMap state with no-op MapStore persistence. The directory-layout / WAL-capacity sections that mention running-job metrics should be adjusted at the same time.
- Severity: Medium
- Already raised by another reviewer: No

# Code quality / tests / stability
The code path itself is still narrowly scoped. The new unit test at `FileMapStoreTest.java:106-123` covers the important branch: even with an invalid storage type, `engine_runningJobMetrics` skips real storage initialization and no-op store/delete/loadAllKeys/destroy calls stay safe.

Test stability rating: Stable. This is a pure unit test with no sleeps, async polling, external containers, ports, network state, or ordering-sensitive assertions.

# CI / merge gate
Current head: `b0a783f9c3f82c9bc16e8e6e7ebec3637f893e12`.
Current check state I saw: `Build` is queued, with the lightweight label/notify checks already successful. This PR is also currently diverged from `dev` (`ahead_by=4`, `behind_by=96`), but I am not treating that alone as the contributor action yet because the concrete blocker here is the docs contract above.

### Merge conclusion: changes requested

1. Blocking items
- Issue 1: please sync the dedicated EN/ZH state-storage/recovery docs with the new `engine_runningJobMetrics` no-persistence behavior.

- A short class-level comment on `NoOpMapStorage` would make the intent even easier for future maintainers to recognize, but I do not see that as a merge blocker.

Overall, the implementation still looks directionally right and focused. Once the state-storage/recovery docs are corrected, I am happy to re-review the updated head.

davidzollo (approved):
+1 if CI passes.
LGTM

davidzollo:
@zhangshenghang Please help review this PR , thx

zhangshenghang (approved):
+1

PR #11248, opened 2026-07-01
[Docs][Connector-V2] Improve Qdrant connector docs

DanielCarter-stack:
## Purpose

Improve the Qdrant connector documentation by aligning the source and sink pages with the current connector behavior and adding complete task examples.

## Changes

- Add runnable Qdrant source and sink task examples in English and Chinese docs.
- Clarify how Qdrant payload fields and named vector fields map to SeaTunnel schema fields.
- Fix the Qdrant sink collection description from read to write.
- Correct the `use_tls` type to boolean.
- Remove the documented `batch_size` sink option because it is not a user-facing Qdrant sink option in the current implementation.
- Adjust the Qdrant sink type mapping table to avoid listing unsupported user-visible mappings.

## Verification

- `./mvnw spotless:apply`

### DIFF
--- a/docs/en/connectors/sink/Qdrant.md
+++ b/docs/en/connectors/sink/Qdrant.md
@@ -14,7 +14,6 @@ This connector can be used to write data into a Qdrant collection.
 
 | SeaTunnel Data Type | Qdrant Data Type |
 |---------------------|------------------|
-| TINYINT             | INTEGER          |
 | SMALLINT            | INTEGER          |
 | INT                 | INTEGER          |
 | BIGINT              | INTEGER          |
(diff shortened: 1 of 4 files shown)


DanielLeens (commented):
Thanks for the Qdrant docs update. I reviewed this head and also compared it with the newer `#11252` revision from the same author.

The two PRs are covering the same Qdrant source/sink documentation area and the same file set, but `#11252` is the more complete version of the cleanup: it keeps the source schema examples aligned with the current docs shape, fixes the sink `use_tls` typing, removes the non-user-facing `batch_size` option, and adds the full source-to-sink usage example in one place.

To avoid maintaining two parallel PRs for the same doc topic, my preference would be to keep `#11252` and close this earlier draft. I am not adding a separate code-level blocker here beyond that duplicate-PR housekeeping point.

PR #11252, opened 2026-07-01
[Docs][Connector-V2] Improve Qdrant connector docs

DanielCarter-stack:
### Purpose

Improve the Qdrant connector documentation so users can configure source and sink jobs more accurately.

### Changes

- Add complete Qdrant source-to-sink task examples to both English and Chinese docs.
- Document that Qdrant collections must exist before the job starts and vector fields must match the collection definition.
- Fix `use_tls` type from `int` to `bool` in English docs.
- Remove the unsupported `batch_size` sink option from the Qdrant sink docs.
- Correct the sink `collection_name` description to say it writes to the target collection.

### Verification

- `./mvnw spotless:apply`

### DIFF
--- a/docs/en/connectors/sink/Qdrant.md
+++ b/docs/en/connectors/sink/Qdrant.md
@@ -10,6 +10,8 @@ import ChangeLog from '../changelog/connector-qdrant.md';
 
 This connector can be used to write data into a Qdrant collection.
 
+The target collection must already exist before the job starts. Vector field names and dimensions in Qdrant must match the vector columns in the SeaTunnel row.
+
 ## Data Type Mapping
 
 | SeaTunnel Data Type | Qdrant Data Type |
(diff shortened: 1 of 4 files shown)


DanielLeens (approved):
Thanks for the follow-up Qdrant docs pass. I reviewed the latest head against the actual source/sink option contracts and the current sink writer behavior.

Runtime path I checked
```text
Qdrant source option contract
  -> QdrantSourceFactory.optionRule() [QdrantSourceFactory.java:47-55]
      -> source requires collection_name + schema
      -> use_tls is a boolean option

Qdrant sink option contract
  -> QdrantSinkFactory.optionRule() [QdrantSinkFactory.java:45-54]
      -> sink exposes host / port / api_key / collection_name / use_tls
      -> there is no user-facing batch_size option

Qdrant sink runtime
  -> QdrantSinkWriter(...) [QdrantSinkWriter.java:34-37]
      -> writer hardcodes batchSize = 64 internally
```

What I found
- Removing `batch_size` from the sink docs is the right move here, because the sink option rule does not expose it as a configurable user option and the writer currently hardcodes `64` internally.
- Changing `use_tls` to `bool` is also correct for both source and sink docs.
- The new source->sink task examples are consistent with the real connector contract, and I did not find a reopened documentation mismatch on the current head.

Conclusion
- From Daniel's side, this revision looks good to merge.
- The only remaining gate at the moment is that `Build` is still running on the latest head. I do not have a code-level blocker while that check finishes.

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

PR #11257, opened 2026-07-01
[Fix][Core] Extend serialVersionUID checker coverage

### LINKED ISSUE #11020: serialVersionUID checker (#9118) misses job-graph config classes; and a question on cross-version state compatibility
raising a gap we hit on upgrade, plus a broader question.

  1. Concrete gap (specific):
  SerialVersionUIDCheckerTest (#9118) only scans classes implementing SeaTunnelSource/SeaTunnelSink. But the serialized job graph also includes config factories (e.g.
  MySqlSourceConfigFactory) and config objects (e.g. JdbcConnectionConfig) — these aren't covered.

  Example: MySqlSourceConfigFactory has no serialVersionUID. Adding one field changed its auto-generated UID, so jobs submitted on the old version failed to deserialize
  on the new one (InvalidClassException) on restart. The checker can't catch this because it doesn't scan the class.

  Would you accept a PR extending the checker to config factories / config / DAG classes?

  2. Broader question (general):
  Even with a UID, compatibility isn't guaranteed. JdbcConnectionConfig has a pinned UID (2L), but 2.3.13 added new fields — old state deserializes with no error, but
  the new fields come back 0/null instead of their defaults (initializers don't run on deserialization). So a UID alone hides a silent behavior change.
  
  So the real questions:
  - Is restoring a running job's serialized state across versions expected to work, or is stop-with-savepoint + resubmit the only supported upgrade path?
  - What's the recommended way to add a field to a serialized config/state class without breaking or silently changing old state (readObject / defaults / a versioning
  convention)?
  
  Happy to contribute the checker extension if it's wanted

zhangshenghang:
### Purpose

Fixes #11020.

The existing `SerialVersionUIDCheckerTest` only checked the generic state types referenced by `SeaTunnelSource` and `SeaTunnelSink` implementations. It missed other Java-serialized classes that are part of job graph or restore-related paths, such as CDC source config factories/config objects and engine job graph DTOs.

### Changes

- Extend `SerialVersionUIDCheckerTest` to check:
  - CDC connector config/config-factory classes under `connector-cdc`
  - engine job graph classes under `engine/core/job` and `engine/common/job`
  - existing source/sink generic state types
- Require `serialVersionUID` to be declared on the concrete class itself instead of accepting an inherited field.
- Add explicit `serialVersionUID` fields to the currently uncovered CDC config/factory classes and engine job graph DTOs.
- Preserve the existing MySQL CDC factory UID baseline by declaring `-6578851046816898665L`.

### Verification

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-ci-tools -Dtest=SerialVersionUIDCheckerTest test`
- `./mvnw -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql -Dtest=MySqlSourceConfigFactorySerializationTest test`

### DIFF
--- a/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/SerialVersionUIDCheckerTest.java
+++ b/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/SerialVersionUIDCheckerTest.java
@@ -45,6 +45,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -60,8 +61,47 @@ public class SerialVersionUIDCheckerTest {
     private static final Logger LOG = LoggerFactory.getLogger(SerialVersionUIDCheckerTest.class);
     private static final String JAVA_FILE_EXTENSION = ".java";
     private static final String CONNECTOR_DIR = "seatunnel-connectors-v2";
+    private static final String CONNECTOR_CDC_DIR =
+            CONNECTOR_DIR + File.separator + "connector-cdc";
     private static final String JAVA_PATH_FRAGMENT =
             "src" + File.separator + "main" + File.separator + "java";
+    private static final List<String> CHECKED_PATHS =
+            Arrays.asList(
+                    CONNECTOR_CDC_DIR,
+                    "seatunnel-engine"
+                            + File.separator
+                            + "seatunnel-engine-core"
+                            + File.separator
+                            + JAVA_PATH_FRAGMENT
+                            + File.separator
+                            + "org"
+                            + File.separator
+                            + "apache"
+                            + File.separator
+                            + "seatunnel"
+                            + File.separator
+                            + "engine"
+                            + File.separator
+                            + "core"
+                            + File.separator
+                            + "job",
+                    "seatunnel-engine"
+                            + File.separator
+                            + "seatunnel-engine-common"
+                            + File.separator
+                            + JAVA_PATH_FRAGMENT
+                            + File.separator
+                            + "org"
+                            + File.separator
+                            + "apache"
+                            + File.separator
+                            + "seatunnel"
+                            + File.separator
+                            + "engine"
+                            + File.separator
+                            + "common"
+                            + File.separator
+                            + "job");
     private static final JavaParser JAVA_PARSER;
     private static final Set<String> checkedClasses = new HashSet<>();
     private static final Map<String, ClassOrInterfaceDeclaration> classDeclarationMap =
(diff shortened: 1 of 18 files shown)


DanielLeens (changes requested):
Thanks for working on this! I went through the full diff and the direction makes sense. Extending the checker to the engine job DTOs and the CDC config/config-factory classes is definitely useful. I did find one blocker that we should address before calling `#11020` fixed.

1. The checker still misses the JDBC config objects that are already on the serialized source path.

- In this PR, the new direct-scan scope is hard-coded to `connector-cdc` plus the two engine job packages (`seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/SerialVersionUIDCheckerTest.java:64-104`, `161-169`, `294-303`).
- But the JDBC config chain is still part of the real job-graph / restore path today:
  - `SourceAction` stores the concrete `SeaTunnelSource` inside the job graph (`seatunnel-engine/seatunnel-engine-core/src/main/java/org/apache/seatunnel/engine/core/dag/actions/SourceAction.java:32-46`).
  - `JdbcSource` stores `JdbcSourceConfig` as a field (`seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/source/JdbcSource.java:50-68`).
  - `JdbcSourceConfig` stores `JdbcConnectionConfig` (`seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/config/JdbcSourceConfig.java:31-56`).
  - `JdbcConnectionConfig` is itself `Serializable` (`seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/config/JdbcConnectionConfig.java:29-80`).

So the exact class family called out in `#11020` can still change its generated UID without this checker failing. In other words, this PR improves coverage, but it does not fully close the original gap yet.

From the runtime perspective, the current path is:

```text
JdbcSourceFactory.createSource(...)
  -> JdbcSourceConfig.of(...)
      -> JdbcConnectionConfig.of(...)
  -> new JdbcSource(config)
      -> wrapped into SourceAction(..., source, ...)
          -> serialized as part of the job graph / restore path

SerialVersionUIDCheckerTest
  -> findClassPaths()
      -> connector-cdc
      -> engine/core/job
      -> engine/common/job
  -> connector-jdbc is still outside the scan scope
```

I think the lowest-cost fix is to extend the direct scan to the JDBC config/config-factory path as well. A more general long-term option would be to stop keying this off the `connector-cdc` directory and instead cover serializable config/config-factory classes that are reachable from concrete source/sink instances across connector modules.

The engine job DTO additions themselves look fine, and I don't see a flaky-test concern in the updated checker. Once the JDBC side is covered too, I'm happy to take another look.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch again against the exact JDBC / engine serialization path I blocked previously.

- The checker now scans the JDBC config path directly in addition to the CDC and engine job packages.
- `JdbcSourceTableConfig` now declares its own `serialVersionUID`.
- The new coverage assertion makes sure the JDBC config directory is actually part of the scan set.

## Runtime path I rechecked
```text
Jdbc source serialization path
  -> JdbcSourceFactory.createSource(...)
  -> JdbcSourceConfig / JdbcConnectionConfig / JdbcSourceTableConfig
  -> wrapped into the source graph and serialized in job / restore flows

checker path
  -> SerialVersionUIDCheckerTest.findClassPaths()
      -> connector-cdc
      -> connector-jdbc/.../jdbc/config
      -> engine/core/job
      -> engine/common/job
```

## Findings
- The original blocker is closed for me: the JDBC config family is now inside the enforced scan scope instead of being left out.
- The new `checkJdbcConfigPathIsCovered()` guard is a useful protection against someone silently dropping that directory again later.
- Test stability rating: Stable. This is deterministic parser-based coverage with no timing or environment dependency.

1. Blocking items
- None from my side on the latest source path.

2. Suggested follow-up
- The current `Build` is still queued, so CI still needs to finish green before merge.

This looks good to me now.

davidzollo (approved):
+1 if CI passes.
LGTM

DanielLeens:
Thanks for the follow-up. I rechecked the PR metadata before posting again: there is still no new commit after `0e44c74`, so I am not reopening a full review on unchanged code. From my side the blocker I raised earlier has already been addressed on the current head; at this point I do not have a new source-level concern beyond waiting for CI to stay green on the latest base.

davidzollo (approved):
+1
Good job

PR #11203, opened 2026-06-28
[Test][Zeta] Stabilize HTTPS REST API test

DanielLeens:
## What
- resolve HTTPS test keystore resources via the classpath instead of `user.dir`
- use dynamic HTTP/HTTPS ports in `RestApiHttpsTest` to avoid fixed-port collisions
- wait on the REST responses themselves and tolerate empty error streams to reduce shared flakiness

## Validation
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home ./mvnw spotless:apply -pl seatunnel-engine/seatunnel-engine-server -nsu -Dmaven.gitcommitid.skip=true -T 3C`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home ./mvnw test -pl seatunnel-engine/seatunnel-engine-server -Dtest=RestApiHttpsTest -DfailIfNoTests=false -nsu -Dmaven.gitcommitid.skip=true -T 3C`

## Notes
- On this machine, the bundled HTTPS test fixtures are PKCS12 and the local JDK 8u171 cannot load them reliably, so the closest feasible test validation used local JDK 11.

PR #11258, opened 2026-07-01
[Test][Zeta] Stabilize pending job cancel status after failover

DanielLeens:
## What does this PR do?

Stabilizes `SplitClusterPendingJobLifecycleFailoverIT` by waiting for the canceled job result before asserting the final client-visible `CANCELED` status.

The flaky failure was unrelated to the HTTPS REST API change in #11203. The failing engine E2E job timed out while polling `getJobStatus()` after `cancelJob()`, even though failover cancellation can legitimately remain in a transient `CANCELING` state before the terminal result is published.

## Why is this needed?

`#11203` was failing in `SplitClusterPendingJobLifecycleFailoverIT.testPendingJobLifecycleInMasterFailover` with a `ConditionTimeout` on the final canceled-state assertion. This change aligns that test with the existing repository pattern that first waits for job completion and then checks the final exposed status.

## Verification

- `./mvnw spotless:apply -pl seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base -am -nsu -Dmaven.gitcommitid.skip=true -T 3C`
- `./mvnw -pl seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base,!seatunnel-engine/seatunnel-engine-ui -am -DskipUT -DskipIT=false -Dit.test=SplitClusterPendingJobLifecycleFailoverIT -DfailIfNoTests=false -nsu -Dmaven.gitcommitid.skip=true -T 3C verify`

This PR is kept in draft until remote CI turns green.

### DIFF
--- a/seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/SplitClusterPendingJobLifecycleFailoverIT.java
+++ b/seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/SplitClusterPendingJobLifecycleFailoverIT.java
@@ -27,6 +27,7 @@ import org.apache.seatunnel.engine.common.config.JobConfig;
 import org.apache.seatunnel.engine.common.config.SeaTunnelConfig;
 import org.apache.seatunnel.engine.common.config.server.ScheduleStrategy;
 import org.apache.seatunnel.engine.common.exception.SeaTunnelEngineException;
+import org.apache.seatunnel.engine.common.job.JobResult;
 import org.apache.seatunnel.engine.common.job.JobStatus;
 import org.apache.seatunnel.engine.core.job.PipelineStatus;
 import org.apache.seatunnel.engine.server.SeaTunnelServer;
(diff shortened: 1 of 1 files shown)


DanielLeens (commented):
Thanks for the contribution. I reviewed the latest head from the split-cluster failover cancel path down to the client-side terminal-status assertion, and this looks good to me.

## What this PR solves
- User pain: in the pending-job failover E2E, the server can already have a terminal cancel result while the client-visible `getJobStatus()` still briefly reports `CANCELING`, which makes the test fail on a transient observation window instead of the real terminal outcome.
- Fix approach: stop asserting cancellation only through the instantaneous job status. Wait for `waitForJobCompleteV2()` to return the terminal `JobResult`, then use a short Awaitility loop to confirm the externally visible status also converges to `CANCELED`.
- One-line summary: this change makes the test assert the real terminal cancel outcome instead of a transient failover-status lag.

```text
pending job cancel after failover
  -> ClientJobProxy.cancelJob()
  -> engine finishes cancellation and persists JobResult
  -> client getJobStatus() can still lag at CANCELING for a short window
  -> new assertEventuallyCanceled(...)
      -> waitForJobCompleteV2()
          -> fetch terminal JobResult(CANCELED)
      -> Awaitility loop
          -> assert getJobStatus() finally converges to CANCELED
```

## Key findings
1. The normal test path definitely hits this change: all three cancel assertions in `SplitClusterPendingJobLifecycleFailoverIT` now go through the new helper.
2. The fix is targeted at the real flaky window here: terminal cancellation can already be complete before the client-facing status cache fully converges.
3. The helper does not weaken the assertion into a blind sleep. It first waits for the terminal engine-side result, then still checks the visible status.

## Findings

### Conclusion: can merge
1. Blocking items
- None.

- None.

Overall, this is a precise stability fix for the failover cancel assertion path, and I'm happy to approve it.

davidzollo (approved):
+1

PR #11269, opened 2026-07-02
[Fix][Zeta] Make finished jobs API tolerate missing history data

### LINKED ISSUE #10039: [Bug] [Rest API] The web page's finished-jobs interface returns null when the finished-jobs are stored in S3 and there are a large number of them.
### Search before asking

### What happened

The web page's finished-jobs interface returns null when the finished-jobs are stored in S3 and there are a large number of them.There are approximately more than 80.I can use ./bin/seatunnel.sh --list to display normally, and it's very fast.

### SeaTunnel Version

2.3.12

### SeaTunnel Config

```conf
-------seatunnel.yaml
seatunnel:
  engine:
    classloader-cache-mode: true
    history-job-expire-minutes: 1440
    backup-count: 1
    queue-type: blockingqueue
    print-execution-info-interval: 60
    print-job-metrics-info-interval: 60
    slot-service:
      dynamic-slot: true
    checkpoint:
      interval: 10000
      timeout: 60000
      storage:
        type: hdfs
        max-retained: 3
        plugin-config:
          namespace: /seatunnel/test/checkpoint
          #storage.type: oss
          storage.type: s3
          disable.cache: false
          fs.s3a.endpoint: xxx
          s3.bucket: s3a://xxx
          fs.s3a.access.key: xxx
          fs.s3a.secret.key: xxx
          fs.s3a.aws.credentials.provider: org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider
    telemetry:
      metric:
        enabled: false
      logs:
        scheduled-deletion-enable: true
    http:
      enable-http: true
      port: 8881
      enable-dynamic-port: false
      # Uncomment the following lines to enable basic authentication for web UI
      enable-basic-auth: true
      basic-auth-username: admin
      basic-auth-password: xxxx




--hazelcast-master.yaml
hazelcast:
  cluster-name: seatunnel
  network:
    rest-api:
      enabled: true
      endpoint-groups:
        CLUSTER_WRITE:
          enabled: true
        DATA:
          enabled: true
    join:
      tcp-ip:
        enabled: true
        member-list:
          - localhost:5801
          - localhost:5802
    port:
      auto-increment: false
      port: 5801
  properties:
    hazelcast.invocation.max.retry.count: 20
    hazelcast.tcp.join.port.try.count: 30
    hazelcast.logging.type: log4j2
    hazelcast.operation.generic.thread.count: 50
    hazelcast.heartbeat.failuredetector.type: phi-accrual
    hazelcast.heartbeat.interval.seconds: 2
    hazelcast.max.no.heartbeat.seconds: 180
    hazelcast.heartbeat.phiaccrual.failuredetector.threshold: 10
    hazelcast.heartbeat.phiaccrual.failuredetector.sample.size: 200
    hazelcast.heartbeat.phiaccrual.failuredetector.min.std.dev.millis: 100
  map:
    engine*:
      map-store:
        enabled: true
        initial-mode: EAGER
        factory-class-name: org.apache.seatunnel.engine.server.persistence.FileMapStoreFactory
        properties:
          type: hdfs
          namespace: /seatunnel/test/imap
          clusterName: seatunnel
          storage.type: s3
          fs.s3a.access.key: xxxxx
          fs.s3a.secret.key: xxxx
          fs.s3a.endpoint: xxxx
          s3.bucket: s3a://xxx
          fs.s3a.aws.credentials.provider: org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider
```

### Running Command

```shell
-----
```

### Error Exception

```log
------seatunnel-engine-master.log

 ERROR [.s.r.f.ExceptionHandlingFilter] [qtp164877646-924] - Error occurred while processing request
java.lang.NullPointerException: null
```

Zeta

### Java or Scala Version

1.8

### Screenshots

Image

Image

### Code of Conduct

zhangshenghang:
### Purpose

Fixes #10039.

The finished-jobs REST endpoint could fail the whole response when historical job data loaded from the persisted maps was incomplete or partially unavailable. In practice, this can happen with external map storage such as S3, where one missing metrics/DAG record or a null finish time should not make the UI receive a null/500 response for the whole page.

### Changes

- Sort finished jobs with null finish times placed last instead of throwing during comparison.
- Load finished-job metrics and DAG data defensively; if one record is missing or cannot be read, return that job with empty metrics/DAG instead of failing the entire list.
- Add a regression test for finished-job listing with missing finish time, unavailable metrics, and unavailable DAG data.

### Tests

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-engine/seatunnel-engine-server -DskipIT -Dcheckstyle.skip -Dlicense.skip=true -Dspotless.check.skip=true -Dtest=JobInfoServiceNullSafetyTest test`
- `git diff --check`


DanielLeens (commented):
Thanks for working on this. I re-reviewed the latest head against the actual finished-job REST path, not just the small diff.

## What this PR solves
- User pain: `/finished-jobs` can currently fail the whole response when one historical record is incomplete, for example when finish time is null or one persisted metrics / DAG entry is missing.
- Fix approach: the new code filters null job states, sorts with `nullsLast`, and loads finished metrics / DAG defensively from the persisted finished-job maps instead of letting one broken record abort the entire list.
- One-line summary: this makes the finished-job list degrade gracefully record by record instead of failing page-wide.

## Runtime path I rechecked
```text
REST list request
  -> JobInfoService.getJobsByStateJson(...)
      -> IMAP_FINISHED_JOB_STATE
      -> sort by finish time
      -> per job:
           -> getFinishedJobMetricsJson(jobId)
                -> IMAP_FINISHED_JOB_METRICS
                -> fallback to JobMetrics.empty()
           -> getFinishedJobDAGInfo(...)
                -> IMAP_FINISHED_JOB_VERTEX_INFO
                -> fallback to null
      -> BaseService.getJobInfoJson(...)
```

## Findings
- The normal finished-job list path really does hit this logic.
- Moving this path to the persisted finished-job maps is coherent with `getJobInfoJson(Long)` and avoids the old cross-node metrics fetch on finished records.
- The new regression test covers the important null / exception cases on this path.
- Test stability rating: Stable. This is a pure mocked unit test with no timing or environment dependency.

1. Blocking items
- None from my side on the latest source path.

2. Suggested follow-up
- The current `Build` check is still queued, so CI still needs to finish green before merge.

Overall, this looks like a good Zeta-side hardening change.

nzw921rx (approved):
Non blocking suggestion: Change the full path reference of the code's class to import

DanielLeens:
Thanks for the follow-up note. I agree this looks like non-blocking import cleanup rather than a new blocker from Daniel's side on the current head.

DanielLeens:
Thanks for the follow-up. I rechecked the current head against the latest `dev` and the current CI state.

- base branch: `dev`
- head SHA: `bf4ad9a29c4ac9ec26ba8e8e181c11fc53e2473b`
- compare status: `diverged`
- ahead_by: `1`
- behind_by: `25`
- mergeable_state: `blocked`
- CI / merge gate snapshot: `Build [cancelled]`

The current failing signal still does not look actionable on top of this stale head yet, so the lowest-cost next step is to sync with the latest `dev` and rerun CI first. If it still fails after the sync, I'm happy to take another look on top of the updated head.

DanielLeens:
Rechecked the latest head and the current merge gate on this PR.

## Current state
- The finished-jobs hardening path still looks good to me on the latest head.
- The current `Build` check is now green.
- I do not see a new source-side blocker from Daniel's side.

So from the Daniel technical-review side, this is no longer waiting on a CI gate. Since I only have read-level review controls here, a write-capable maintainer may still need to handle the final approval / merge action in GitHub if the review summary still shows older state elsewhere.

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

DanielLeens (approved):
Thanks for the contribution. I re-reviewed the latest head from the actual finished-jobs REST path, not just the small diff.

I also noticed @nzw921rx and @davidzollo already approved this PR. After tracing the current path myself, I agree with the overall merge direction.

Runtime chain I checked
```text
REST finished-jobs
  -> JobInfoService.getJobsByStateJson(...) [JobInfoService.java:120-155]
      -> finished job state map
      -> filter null states
      -> sort by finishTime with nulls last
      -> per job load metrics / DAG through
          -> getFinishedJobMetricsJson(jobId)
          -> getFinishedJobDAGInfo(...)
      -> BaseService.getJobInfoJson(...)
```

What this PR fixes on the real path
- User pain: one incomplete history record could take down the whole `/finished-jobs` page.
- Fix approach: degrade per record instead of failing the whole list when finished metrics / DAG / finishTime are missing or broken.
- One-line summary: this is a focused null-safety fix on the real Zeta history list path.

Findings
- The normal `/finished-jobs` path does hit this code directly.
- The fix is narrowly scoped: no API, config, protocol, or history-storage contract changes.
- The new test coverage is aligned with the real failure modes: missing metrics, missing state, null finishTime, and map access throwing.

Test note
- The added test looks stable to me: it is pure mock-based unit coverage with no timing or container dependency.
- The latest Build is green on the current head.

1. Blocking items
- None from my side.

2. Non-blocking suggestions
- You can still take @nzw921rx's small import-cleanup suggestion if you want, but I would not block merge on that.

Overall, this looks good to merge.

PR #11287, opened 2026-07-02
[Docs][Connector-V2] Improve Prometheus connector docs

DanielCarter-stack:
### Purpose

Improve the Prometheus connector documentation so users can configure source and sink jobs without cross-checking connector code or E2E examples.

### Changes

- Clarify Prometheus source instant and range query options, including range-only `start`, `end`, and `step` settings.
- Document the expected source output schema and Prometheus-compatible API usage.
- Clarify Prometheus sink remote write behavior, required field mapping options, timer flush, and timestamp handling.
- Align English and Chinese source/sink pages and normalize option tables and examples.

### Verification

- `./mvnw spotless:apply`
- `git diff --check`

### DIFF
--- a/docs/en/connectors/sink/Prometheus.md
+++ b/docs/en/connectors/sink/Prometheus.md
@@ -19,16 +19,23 @@ import ChangeLog from '../changelog/connector-prometheus.md';
 
 ## Description
 
-Writes metric samples to a Prometheus-compatible remote write endpoint.
+The Prometheus sink connector writes rows to the Prometheus remote write API. It builds a remote write sample from
+three upstream fields:
 
-The sink converts each SeaTunnel row into one Prometheus sample, serializes the data as a remote write request, compresses it with Snappy, and sends it by HTTP POST. Use a remote write endpoint such as `http://prometheus:9090/api/v1/write` or `http://victoria-metrics:8428/api/v1/write`.
+- `key_label`: the field that contains Prometheus labels, usually a `map<string, string>`.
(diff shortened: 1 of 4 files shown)


DanielLeens (changes requested):
Thanks for the Prometheus docs update. I fetched the latest head locally and reviewed the full diff against `dev`.

The connector doc refresh is moving in a good direction, but I still see one blocking issue in the current head.

Current path I reviewed:
- Intended path: Prometheus connector docs under `docs/en/...` and `docs/zh/...`.
- Actual current diff: the PR also modifies `seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/SplitClusterPendingJobLifecycleFailoverIT.java`, which belongs to the engine E2E lifecycle test suite.

Blocking items:
1. `seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/SplitClusterPendingJobLifecycleFailoverIT.java`
   This PR is scoped as a Prometheus documentation improvement, but the latest head still carries an unrelated engine E2E test change. Please remove that file from this PR and keep it in a separate engine/test PR if it is still needed.

I do not see another blocker in the Prometheus doc content itself from the latest version. Once the unrelated E2E change is removed, I am happy to take another look.

zhangshenghang (approved):

PR #11289, opened 2026-07-02
[Improve][CI] Limit full backend matrix to protected branches

DanielLeens:
## Why

The backend workflow currently forces `api=true` whenever the workflow runs in the `apache/seatunnel` repository context. In practice, that bypasses incremental CI scoping for PR builds and fans many feature branches into the full backend matrix.

This is especially expensive for connector-only and docs-only PRs, which still end up running the full unit-test and integration-test suite.

## What changed

- add `tools/update_modules_check/ci_scope.py` to decide when full API coverage is required
- keep unconditional full coverage only on protected long-lived apache branches such as `dev` and `*-release`
- preserve changed-file scoping on feature and PR branches
- add a focused regression test for the CI scoping rules

## Validation

- `python3 tools/update_modules_check/test_ci_scope.py`
- `python3 tools/update_modules_check/ci_scope.py apache refs/heads/dev`
- `python3 tools/update_modules_check/ci_scope.py apache refs/heads/dev-pulsar-cleanup-preload-20260623`
- `python3 tools/update_modules_check/ci_scope.py apache refs/heads/docs/e2e-sls-docs-20260703`
- `python3 tools/update_modules_check/ci_scope.py DanielLeens refs/heads/dev`
- YAML parse for `.github/workflows/backend.yml`
- `./mvnw spotless:apply`

## Expected effect

- docs-only PRs should stay in the docs/build-website path instead of fanning into the full backend matrix
- connector-only PRs should stay in connector-scoped CI instead of forcing the full API path
- protected apache branches still keep the current full-coverage behavior


DanielLeens:
Closing this draft because it was created from the wrong base remote and the remote file scope is invalid. Superseded by #11290, which was recreated from apache/dev with the intended 3-file CI-only diff.

PR #11290, opened 2026-07-02
[Improve][CI] Scope backend matrix for fork changes

DanielLeens:
## What does this PR do?

This keeps the full backend API matrix for protected Apache repository pushes and broad API-impacting changes, but avoids fanning out every connector integration-test shard when a fork push only changes CI workflows, module-detection tooling, `seatunnel-dist`, or the install-plugin script.

The PR also documents how the required pull-request `Build` mirrors the push workflow in the PR head repository, and how to interpret full versus changed-file-scoped coverage.

## Why is the change needed?

Recent fork PR runs spend hours queued or in progress because lightweight changes still set `api=true`, which schedules heavyweight jobs such as `all-connectors-it-*`, `jdbc-connectors-it-*`, and `kafka-connector-it`. Those jobs expose PRs to unrelated flakes and resource contention even when the PR only changes CI tooling or the dist install script.

## How was this PR implemented?

- Add `tools/update_modules_check/ci_scope.py` to centralize full-matrix scope decisions.
- Add CI-executed regression tests for protected branches, bare base refs, pull-request merge refs, versioned release branches, non-version release branches, release tags, fork changes, and mixed source changes.
- Update `.github/workflows/backend.yml` to call the scope helper and detect `seatunnel-dist` / `bin/install-plugin.sh` changes separately.
- Keep fail-safe behavior: helper failure or invalid output falls back to full API coverage.
- Document the Backend CI routing contract in the English and Chinese contribution paths.

## Validation

- `./mvnw --batch-mode --quiet --no-snapshot-updates spotless:apply`
- `/usr/local/bin/python3.11 -m unittest discover -s tools/update_modules_check -p 'test_*.py'` (14 tests passed)
- `git diff --check`
- No local compile, integration test, E2E, Docker, or service start was performed.
- GitHub Actions `Build` on the latest head remains the runtime source of truth.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -220,9 +220,9 @@ jobs:
           api_files=`python tools/update_modules_check/check_file_updates.py ua $workspace apache/dev origin/$current_branch "seatunnel-api/**" "seatunnel-common/**" "seatunnel-config/**" "seatunnel-core/**" "seatunnel-e2e/seatunnel-e2e-common/**" "seatunnel-formats/**" "seatunnel-plugin-discovery/**" "seatunnel-transforms-v2/**" "seatunnel-translation/**" "seatunnel-e2e/seatunnel-transforms-v2-e2e/**" "pom.xml" "**/workflows/**" "tools/**" "seatunnel-dist/**"`
           true_or_false=${api_files%%$'\n'*}
           file_list=${api_files#*$'\n'}
-          # Keep broad API coverage for protected apache branches and source-level API changes.
-          # CI tooling, workflow, and dist script changes are validated by lightweight checks instead
-          # of fanning out the full connector matrix on every fork push.
+          # Apache dev, main, master, and numeric *-release pushes force full API coverage.
+          # On fork pushes, workflow files, tools/update_modules_check, seatunnel-dist, and
+          # bin/install-plugin.sh do not trigger the full connector matrix by themselves.
           ci_scope_ref="${GITHUB_BASE_REF:-$GITHUB_REF}"
           if ! true_or_false=$(python tools/update_modules_check/ci_scope.py "$repository_owner" "$ci_scope_ref" --api-changed "$true_or_false" --api-files-json "$file_list"); then
             echo "::warning::CI scope helper failed; forcing full API coverage"
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -58,6 +58,8 @@ jobs:
         run: ./mvnw -B -T 1 clean test -D"license.skipAddThirdParty"=true -pl seatunnel-ci-tools -am --no-snapshot-updates
         env:
           MAVEN_OPTS: -Xmx512m
+      - name: Check CI scope rules
+        run: python3 -m unittest discover -s tools/update_modules_check -p 'test_*.py'
       - name: Check for .class files in git
         run: |
           echo "Checking for .class files tracked by git..."
@@ -209,10 +211,25 @@ jobs:
           echo "edge-agent=$true_or_false" >> $GITHUB_OUTPUT
           echo "edge-agent_files=$file_list" >> $GITHUB_OUTPUT
 
+          dist_files=`python tools/update_modules_check/check_file_updates.py ua $workspace apache/dev origin/$current_branch "seatunnel-dist/**" "bin/install-plugin.sh"`
+          true_or_false=${dist_files%%$'\n'*}
+          file_list=${dist_files#*$'\n'}
+          echo "dist=$true_or_false" >> $GITHUB_OUTPUT
+          echo "dist_files=$file_list" >> $GITHUB_OUTPUT
+
           api_files=`python tools/update_modules_check/check_file_updates.py ua $workspace apache/dev origin/$current_branch "seatunnel-api/**" "seatunnel-common/**" "seatunnel-config/**" "seatunnel-core/**" "seatunnel-e2e/seatunnel-e2e-common/**" "seatunnel-formats/**" "seatunnel-plugin-discovery/**" "seatunnel-transforms-v2/**" "seatunnel-translation/**" "seatunnel-e2e/seatunnel-transforms-v2-e2e/**" "pom.xml" "**/workflows/**" "tools/**" "seatunnel-dist/**"`
           true_or_false=${api_files%%$'\n'*}
           file_list=${api_files#*$'\n'}
-          if [[ $repository_owner == 'apache' ]];then
+          # Keep broad API coverage for protected apache branches and source-level API changes.
+          # CI tooling, workflow, and dist script changes are validated by lightweight checks instead
+          # of fanning out the full connector matrix on every fork push.
+          ci_scope_ref="${GITHUB_BASE_REF:-$GITHUB_REF}"
+          if ! true_or_false=$(python tools/update_modules_check/ci_scope.py "$repository_owner" "$ci_scope_ref" --api-changed "$true_or_false" --api-files-json "$file_list"); then
+            echo "::warning::CI scope helper failed; forcing full API coverage"
+            true_or_false='true'
+          fi
+          if [[ $true_or_false != 'true' && $true_or_false != 'false' ]]; then
+            echo "::warning::CI scope helper returned an invalid value; forcing full API coverage"
             true_or_false='true'
           fi
           echo "api=$true_or_false" >> $GITHUB_OUTPUT
@@ -266,9 +283,13 @@ jobs:
       - name: Make unit test modules
         id: ut-modules
         timeout-minutes: 60
-        if: ${{ steps.filter.outputs.api == 'false' && (steps.engine-modules.outputs.modules != '' || steps.cv2-modules.outputs.modules != '') }}
+        if: ${{ steps.filter.outputs.api == 'false' && (steps.engine-modules.outputs.modules != '' || steps.cv2-modules.outputs.modules != '' || steps.filter.outputs.dist == 'true') }}
         run: |
-          modules='${{ steps.engine-modules.outputs.modules }}${{ steps.cv2-modules.outputs.modules }}'
+          dist_modules=''
+          if [[ '${{ steps.filter.outputs.dist }}' == 'true' ]]; then
+            dist_modules=',seatunnel-dist'
+          fi
+          modules='${{ steps.engine-modules.outputs.modules }}${{ steps.cv2-modules.outputs.modules }}'$dist_modules
           modules=${modules: 1}
           pl_modules=`python tools/update_modules_check/update_modules_check.py replace "$modules"`
           # remove deleted modules
@@ -283,6 +304,11 @@ jobs:
 
           engine_modules='${{ steps.engine-modules.outputs.modules }}'
           connector_modules='${{ steps.cv2-modules.outputs.modules }}'
+          if [[ "zz${engine_modules}${connector_modules}" == "zz" && "zz"$dist_modules != "zz" ]];then
+            echo $pl_modules
+            echo "modules=$pl_modules" >> $GITHUB_OUTPUT
+            exit 0
+          fi
           if [[ "zz"$connector_modules == "zz" && "zz"$engine_modules != "zz" ]];then
             # Engine changes already route the full downstream engine test set through engine=true.
             # Keep the fixed smoke modules as the output without resolving the Maven dependency tree.
--- a/docs/en/developer/contribution-path.md
+++ b/docs/en/developer/contribution-path.md
@@ -118,6 +118,21 @@ A contribution is much easier to review when it includes:
 
 For code contributions, avoid mixing unrelated cleanup with the real fix.
 
+## Understand Backend CI Coverage
+
+The required Backend `Build` on a pull request mirrors the `push` workflow run in the PR head repository, which is usually the contributor's fork. It is not a separate full-matrix run in the base repository.
+
+The workflow selects jobs as follows:
+
+- pushes in the Apache repository to `dev`, `main`, `master`, and numeric release branches such as `2.3.13-release` force the full API-triggered backend matrix
+- broad API-impacting changes, including API, core, common, format, transform, translation, and root build changes, trigger the full API matrix
+- connector-only changes use changed-module detection to select integration-test shards; when the unit-test job is selected, it still verifies all modules
+- engine changes keep their existing engine and connector integration-test paths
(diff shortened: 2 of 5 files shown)


DanielLeens:
Self-review note on the latest head:

## What this PR improves
- User pain: feature branches were still paying the cost of the full backend matrix even when the branch scope did not require the protected-branch validation surface.
- Fix approach: move the branch-scope decision into a dedicated helper so `backend.yml` only forces the full backend matrix on protected branches.
- One-line summary: the CI-scope split still looks consistent on the current head, and I do not see a reopened source-level blocker from Daniel's side.

## Runtime / review path I checked
```text
Backend CI scope path
  -> backend.yml calls the CI scope helper
  -> ci_scope.py classifies the current branch name
  -> protected branches keep the full backend matrix
  -> feature branches avoid the unnecessary full-matrix expansion
```

## Current status
- The helper/test pair matches the workflow change on the current head.
- I do not see a reopened code-level blocker in the current revision.
- The remaining merge gate is the current `Build` run, which is still in progress on this head.

## Review conclusion
1. Blocking items
- Please let the current `Build` check settle green on this head.

2. Suggested follow-up
- None on the current source diff from my side.

DanielLeens:
Self-follow-up on the current head:

- base branch: `dev`
- head SHA: `32d1de593ca2a6df40cc59f6e32781b4584a671d`
- compare status: `diverged`
- ahead_by: `4`
- behind_by: `52`
- mergeable_state: `blocked`
- CI / merge gate snapshot: `Build [failure]`

The source-level CI-scope conclusion from the earlier self-review note still stands for me on this head. The remaining actionable step is to sync with the latest `dev` and rerun `Build` on top of the updated branch before treating the current failure as PR-local.

DanielLeens (commented):
Self-follow-up on the latest head:

- I re-reviewed the current branch from scratch.
- The workflow scoping change still looks right to me: `tools/update_modules_check/ci_scope.py` only forces full API coverage on protected apache branches (`dev` / `main` / `master` / `*-release`), while feature and PR branches keep the changed-file scope in `.github/workflows/backend.yml`.
- The new Doris stabilization in `DorisMultiReadIT` also goes in the right direction for me. It now waits for the sink row count to converge before doing schema and row-by-row assertions, and the wait is bounded with an explicit 60s timeout instead of an open-ended sleep.

I do not see a reopened source-side blocker from my side on this head. The remaining gate is the queued `Build` run, and because this is still my own PR a maintainer with write-level review control will still need to make the final approval / merge decision after CI finishes.

DanielLeens:
Thanks for the update. I checked the current head against the latest `dev`, and this branch is currently behind upstream (`behind_by=72`, compare status: `diverged`). The remaining failing signal I can see right now is `On pull request update [cancelled]`, so I still do not have evidence that the current gate is caused by this PR's backend-matrix change itself.

The lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the same lane still fails on the rebased head, please paste the new failing check link here and I can help narrow it down further.

DanielLeens (commented):
Outdated.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `f69eeac002f4` from scratch.

# What this PR fixes
- User pain: The blanket `repository_owner == 'apache'` override in backend.yml forces the full backend matrix (`api=true`) on every Apache-hosted branch, so docs-only and connector-scoped pushes burn Java 8/11 integration lanes (all-connectors-it, kafka-connector-it, etc.) they don't need.
- Fix approach: Replace the owner check with a new `tools/update_modules_check/ci_scope.py` helper that only forces full coverage for `dev`/`main`/`master`/`*-release` refs under the apache owner, keeping the changed-file scope result otherwise, plus a small unittest file.
- One-line summary: The intent is sound, but the implementation keys off `$GITHUB_REF`, which is `refs/pull/<n>/merge` on pull_request events — meaning PRs targeting dev/release branches (the dominant path this workflow runs on) silently lose forced full coverage; the `-release` suffix match over-matches tags and branches like `hotfix-release`, the shell falls open to scoped CI on any helper stdout noise or failure, and the new tests neither cover the merge-ref/tag cases nor run anywhere in CI.

# Runtime chain I rechecked
```text
backend.yml 'changes' job step (push/pull_request trigger)
  -> shell: force_full_api=`python tools/update_modules_check/ci_scope.py "$repository_owner" "$GITHUB_REF"`  .github/workflows/backend.yml:215
    -> ci_scope.py main()  tools/update_modules_check/ci_scope.py:40
      -> should_force_full_api_check(repository_owner, github_ref)  tools/update_modules_check/ci_scope.py:25
        -> strips 'refs/heads/' prefix (but pull_request events pass 'refs/pull/N/merge', which falls through unstripped)  tools/update_modules_check/ci_scope.py:36-38
        -> checks PROTECTED_BRANCH_NAMES membership or endswith('-release')  tools/update_modules_check/ci_scope.py:38
      -> prints 'true'/'false' to stdout  tools/update_modules_check/ci_scope.py:50
  -> shell: [[ $force_full_api == 'true' ]] && true_or_false='true'  .github/workflows/backend.yml:216-218
    -> echo "api=$true_or_false" >> $GITHUB_OUTPUT  .github/workflows/backend.yml:219
      -> downstream jobs gate on needs.changes.outputs.api to decide matrix fan-out
```

# Findings
**Issue 1: pull_request events never match: GITHUB_REF is refs/pull/N/merge, so every PR (including PRs targeting dev or *-release) silently loses forced full API coverage**
- Location: `.github/workflows/backend.yml:218`
- Why it matters: - Root cause: `ci_scope.py` only recognizes `refs/heads/<branch>` values, but for `pull_request` events GitHub sets `GITHUB_REF` to `refs/pull/<number>/merge`. That ref never matches `PROTECTED_BRANCH_NAMES` or the `-release` suffix, so `should_force_full_api_check` returns `false` for 100% of PR runs against the apache repo.
- Impact: previously `repository_owner == 'apache'` forced `api=true` on every PR to apache/seatunnel. After this change, a PR whose diff touches only one connector but that targets `2.3.x-release` or `dev` runs only the scoped matrix, so cross-cutting regressions (API/serialization/Zeta engine breakage triggered indirectly by a connector change) can merge into a protected branch and only surface on the post-merge push run — after the damage is in the branch history.
- Fix: for `pull_request` events, decide based on the target branch, not the merge ref. Pass `${GITHUB_BASE_REF:-$GITHUB_REF}` into the helper (GITHUB_BASE_REF holds the bare target branch name on PR events, and the helper already handles bare names via its prefix-strip fallback), or explicitly gate: force full coverage when the PR base is `dev`/`main`/`master`/`*-release`. If the intent truly is to scope ALL PRs regardless of target, state that explicitly in the workflow comment, because right now the comment says protected branches keep full coverage while in practice no PR ever does.
- Evidence: `tools/update_modules_check/ci_scope.py:39  return ref_name in PROTECTED_BRANCH_NAMES or ref_name.endswith("-release")`; `.github/workflows/backend.yml:218  force_full_api=`python tools/update_modules_check/ci_scope.py "$repository_owner" "$GITHUB_REF"``; `tools/update_modules_check/ci_scope.py:36-38  ref_prefix = "refs/heads/"\n    ref_name = github_ref[len(ref_prefix) :] if github_ref.startswith(ref_prefix) else github_ref\n    return ref_name in PROTECTED_BRANCH_NAMES or ref_name.endswith("-release")`
- Suggested fix: On `pull_request` events `$GITHUB_REF` is `refs/pull/<N>/merge`, which never matches `PROTECTED_BRANCH_NAMES` or the `-release` suffix in `ci_scope.py`. That means no PR — including PRs targeting `dev` or a `*-release` branch — will ever force full API coverage anymore; regressions only get caught on the post-merge push run. Please pass the PR base branch for PR events, e.g. `"${GITHUB_BASE_REF:-$GITHUB_REF}"` (the helper already accepts bare branch names via its prefix-strip fallback), so PRs into protected branches keep the full matrix.
- Severity: High

**Issue 2: Loose `-release` suffix match plus non-heads-ref fallthrough over-matches: tags and arbitrary branches like `hotfix-release` force the full matrix**
- Location: `tools/update_modules_check/ci_scope.py:39`
- Why it matters: - Root cause: when `github_ref` does not start with `refs/heads/`, the code keeps the entire ref string (`refs/tags/2.3.13-release` stays intact) and then applies `endswith("-release")` to it. Additionally, the suffix check matches any branch name ending in `-release`, not just SeaTunnel's `<version>-release` convention.
- Impact: a tag push named `*-release` or any apache-hosted scratch branch like `revert-something-release` or `pre-release` (does not match, but `fix-release` does) will force `api=true` and fan out the entire heavy Java 8/11 integration matrix — exactly the CI contention this PR is trying to eliminate. Conversely, refs of other shapes (e.g. `refs/tags/dev`) would also incorrectly match the protected set since the raw ref is compared after the fallback.
- Fix: only apply branch logic to `refs/heads/*` refs (return `False` otherwise, or handle `refs/tags/*` deliberately), and tighten the release check to the actual convention, e.g. `re.fullmatch(r"\d+\.\d+(\.\d+)?-release", ref_name)`, matching branches like `2.3.13-release`.
- Evidence: `tools/update_modules_check/ci_scope.py:39  return ref_name in PROTECTED_BRANCH_NAMES or ref_name.endswith("-release")`; `tools/update_modules_check/ci_scope.py:38  return ref_name in PROTECTED_BRANCH_NAMES or ref_name.endswith("-release")`
- Suggested fix: Two over-matching issues here: (1) the fallback on line 38 keeps the full ref for non-`refs/heads/` inputs, so a tag like `refs/tags/2.3.13-release` matches `endswith("-release")`; (2) any branch ending in `-release` (e.g. `fix-release`) forces the full matrix, defeating the scoping goal. I'd return `False` for non-branch refs and tighten the release check to the actual convention: `re.fullmatch(r"\d+\.\d+(\.\d+)?-release", ref_name)`.
- Severity: Medium

**Issue 3: New unit tests are not executed anywhere in CI and miss the pull_request merge-ref case**
- Location: `tools/update_modules_check/test_ci_scope.py:28`
- Why it matters: - Root cause: `test_ci_scope.py` was added, but no step in `.github/workflows/backend.yml` (or any other workflow in this diff) runs it, so it only executes when someone remembers to run it locally. Note that PR #11291 added tests to `tools/update_modules_check/test_update_modules_check.py` with the same gap pattern, so this tooling test debt is accumulating.
- Impact: a future refactor of `ci_scope.py` can silently break the protected-branch decision with green CI, which — given finding #1 and #2 — is precisely the kind of silent regression this gate cannot afford. The test suite also lacks the most important real-world input: `refs/pull/<n>/merge` and the bare `GITHUB_BASE_REF` form (`dev` without `refs/heads/`).
- Fix: add a lightweight step early in `backend.yml` (or the existing sanity job) running `python3 -m unittest discover tools/update_modules_check`, and extend the test class with cases for `refs/pull/123/merge`, bare `dev`, and a non-version `-release` branch.
- Evidence: `tools/update_modules_check/test_ci_scope.py:28-29  def test_force_full_api_check_on_dev_branch(self) -> None:\n        self.assertTrue(should_force_full_api_check("apache", "refs/heads/dev"))`; `tools/update_modules_check/test_ci_scope.py:28  def test_force_full_api_check_on_dev_branch(self) -> None:`; `tools/update_modules_check/test_ci_scope.py:43  should_force_full_api_check("apache", "refs/heads/docs/e2e-sls-docs-20260703")`
- Suggested fix: These tests look good, but nothing in the workflows actually runs them, so they only guard against regressions when run manually. Please wire `python3 -m unittest discover tools/update_modules_check` into an early step of `backend.yml`, and add cases for the inputs CI will really pass: `refs/pull/123/merge` (pull_request `GITHUB_REF`) and bare `dev` (`GITHUB_BASE_REF` form).
- Severity: Medium

**Issue 4: Helper script failure silently degrades to scoped CI instead of failing safe to the full matrix**
- Location: `.github/workflows/backend.yml:219`
- Why it matters: - Root cause: `force_full_api` is captured via command substitution with no exit-code check. If `ci_scope.py` fails (import error, wrong Python on the runner, bad argument), stdout is empty, the subsequent `[[ $force_full_api == 'true' ]]` is simply false, and the step continues.
- Impact: a broken helper would silently disable the protected-branch safety net on `dev`/`*-release` pushes, and nobody would notice because the workflow still goes green with a reduced matrix. Given this script is the single gate for full API coverage, silent failure is the worst failure mode.
- Fix: fail safe. Either check the exit code and fall back to `'true'` (e.g., `force_full_api=$(python ... ) || force_full_api=true`), or make the step exit non-zero on helper failure so the misconfiguration is visible immediately. Also validate the output is exactly `true` or `false` before using it.
- Suggested fix: If `ci_scope.py` errors out for any reason, `$force_full_api` is empty and this condition quietly falls through to scoped CI — the opposite of what we want on protected branches. Please fail safe here: `force_full_api=$(python tools/update_modules_check/ci_scope.py "$repository_owner" "$GITHUB_REF") || force_full_api=true`, or explicitly fail the step when the helper does not print `true`/`false`.
- Severity: Medium

**Issue 5: Contributor-facing CI behavior change is undocumented**
- Location: `.github/workflows/backend.yml:215`
- Why it matters: - Root cause: this PR changes when the full backend matrix runs — a contract that contributors and committers rely on when judging PR green-ness — but no documentation was updated alongside the workflow change.
- Impact: committers merging into `dev` may assume PR CI already covered the full API matrix (as it did before), when it now only covers the changed-file scope; that expectation gap should be written down somewhere discoverable, e.g., the contributing/CI docs or at minimum the inline comment listing the exact protected branch set.
- Fix: extend the new comment in `backend.yml` to enumerate the protected branches (`dev`, `main`, `master`, `*-release`) so the workflow is self-describing, and add a short note to the contributor CI documentation describing the scoped-vs-full behavior.
- Suggested fix: Since this changes what a green PR check means for committers merging into `dev`, could you list the exact protected branch set (`dev`, `main`, `master`, `*-release`) in this comment and add a short note to the contributor CI docs? Right now a reader has to open `ci_scope.py` to learn which branches keep full coverage.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: pull_request events never match: GITHUB_REF is refs/pull/N/merge, so every PR (including PRs targeting dev or *-release) silently loses forced full API coverage

- Issue 2: Loose `-release` suffix match plus non-heads-ref fallthrough over-matches: tags and arbitrary branches like `hotfix-release` force the full matrix
- Issue 3: New unit tests are not executed anywhere in CI and miss the pull_request merge-ref case
- Issue 4: Helper script failure silently degrades to scoped CI instead of failing safe to the full matrix
- Issue 5: Contributor-facing CI behavior change is undocumented

DanielLeens (commented):
Outdated.

DanielLeens (commented):
Outdated.

davidzollo (approved):
+1

PR #11312, opened 2026-07-03
[Improve][Connector-V2][File] Optimize file discovery and update comparison for sync_mode=update

yzeng1618:

This PR optimizes file discovery and the target/source comparison logic used by the File Source connectors (LocalFile, HdfsFile, FtpFile, SftpFile) when running in sync_mode=update. It introduces two new options to control how the connector looks up target-side file metadata for large / sparse datasets:

update_compare_parallelism (int, default 8, valid range 1–64): the maximum parallelism for sparse target metadata point-lookups in sync_mode=update. The maximum number of submitted-but-incomplete lookups is bounded to 8× this value, so target metadata queries no longer pile up unboundedly.
update_compare_bulk_threshold (int, default 0): for non-FTP/SFTP targets, when the candidate count under a single target parent directory reaches this threshold, the connector performs one directory listing instead of querying every file individually. The default 0 disables automatic bulk listing and keeps using bounded point-lookups, which avoids an unexpectedly expensive full target-directory scan. FTP and SFTP targets always use directory listing. Source filters are applied while entries are listed to reduce peak metadata memory.

Together these give users a way to tune the update-mode comparison cost between "bounded concurrent point lookups" and "single directory listing" depending on how densely files are distributed under target directories.

Yes. This PR adds two new source options for the File connectors, so both behavior and documentation change:

Previous behavior:​ in sync_mode=update, the connector had no way to bound the concurrency of target metadata lookups, nor to switch to a single directory listing for dense directories.
New behavior:​ users can now set update_compare_parallelism (default 8) and update_compare_bulk_threshold (default 0, i.e. disabled). The defaults are chosen to be safe/backward-compatible: with update_compare_bulk_threshold=0 the connector keeps using bounded point-lookups and does not trigger any new full-directory scan, so existing jobs behave essentially as before.

This is a change relative to the unreleased dev branch. English and Chinese docs for all four connectors (LocalFile, HdfsFile, FtpFile, SftpFile) have been updated accordingly, following the connector doc specification in apache/seatunnel#4544.

Added unit tests

### Check list


DanielLeens (commented):
Thanks for working on this. I reviewed the latest head against the real `sync_mode=update` source-construction path, not just the helper-class diff.

## What this PR changes
- User pain: update-mode file sync currently pays for too many target metadata RPCs when the source side discovers many files, especially for FTP/SFTP and dense directory layouts.
- Fix approach: split the path into source discovery plus target metadata loading, add reusable listing sessions, and bound sparse point lookups with explicit concurrency controls.
- One-line summary: the new pipeline is coherent on the current head, and I did not find a code-side blocker in the reviewed path.

## Runtime chain I rechecked
```text
File source startup
  -> AbstractReadStrategy.setPluginConfig(...)
      -> validateUpdateSyncConfig(...)
  -> AbstractReadStrategy.getFileNamesByPath(...)
      -> FileDiscoveryScanner.scan(...)
      -> UpdateFileMetadataLoader.load(...)
          -> bulk directory listing for dense groups
          -> bounded point lookups for sparse groups
      -> shouldSyncFileInUpdateMode(...)
```

## Current-head findings
- The discovery / comparison split is on the real source-construction path at `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/reader/AbstractReadStrategy.java:161-260`, so the optimization is not dead code.
- The new target metadata loader preserves request order while separating dense-directory bulk listing from sparse point lookups at `.../UpdateFileMetadataLoader.java:53-225`, which is exactly the behavior needed before `shouldSyncFileInUpdateMode(...)` consumes the target metadata.
- FTP/SFTP get a dedicated reusable listing session path via `HadoopFileSystemProxy.openFileStatusListingSession()` and the new `StreamingFileSystem` implementations, which matches the PR goal of avoiding one network round-trip per target lookup on those protocols.
- The new config guardrails at `.../AbstractReadStrategy.java:694-765` are clear about the supported update-mode shape and bound the new concurrency knobs.

## Test coverage / stability
- The new unit coverage is focused on the critical cases I would expect here: dense-directory bulk listing, bounded sparse lookups, FTP/SFTP always-bulk behavior, and cleanup after lookup failure.
- Stability rating: Stable
- Basis: `UpdateFileMetadataLoaderTest` exercises concurrency through mocked lookups, but it does not depend on external services, real clocks, or nondeterministic container readiness. I did not see a high-risk flaky pattern on the added test path.

## Merge conclusion
### Conclusion: can merge

1. Blocking items

2. Suggested follow-up
- GitHub `Build` is still queued, so the normal PR gate still needs to finish.

This looks like a meaningful improvement to the file update-sync path, and the supporting tests are pointed at the right risk areas.

DanielLeens (approved):
Rechecked the latest head after the queued gates finished.

Runtime path revisited: `FileSourceConfigBuilder` -> update-mode file discovery -> target-side metadata comparison -> enumerator split planning for `sync_mode=update`. I do not see a reopened source-side blocker on `51cf5e76f1bb5feafa4790b5b2c757f5e83a43c2`, and the earlier `Build` gate is now green.

1. Blocking items

2. Suggested follow-up
- None from this round.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `51cf5e76f1bb` from scratch.

# What this PR fixes
- User pain: In `sync_mode=update`, target-side comparison issued unbounded per-file metadata lookups (or forced full target directory scans), causing lookup pile-up, excessive target metadata load, and high peak memory on large/sparse datasets.
- Fix approach: Bound sparse target point-lookups with a new `update_compare_parallelism` option (default 8, in-flight capped at 8x) and add an opt-in `update_compare_bulk_threshold` (default 0) that switches to a single target directory listing per parent for non-FTP/SFTP targets, while FTP/SFTP targets always use listing with source filters applied during enumeration; docs updated for LocalFile/HdfsFile/FtpFile/SftpFile in en+zh.
- One-line summary: The optimization and options are wired through connector-file-base and documented consistently across all four connectors, but the unconditional FTP/SFTP listing default-behavior change, missing e2e coverage for the concurrent path, unspecified out-of-range handling for the documented 1-64 range, and the advertised-but-no-op bulk threshold on FTP/SFTP tables need to be resolved before merge.

# Runtime chain I rechecked
```text
FileSourceFactory.createSource() -> BaseFileSourceConfig.<init>(readonlyConfig)  BaseFileSourceConfig.java:~78
  -> resolves BaseSourceConfigOptions.UPDATE_COMPARE_PARALLELISM / UPDATE_COMPARE_BULK_THRESHOLD  BaseSourceConfigOptions.java (new options)
  -> AbstractReadStrategy.getFileNamesByPath(sourcePath)  AbstractReadStrategy.java (source enumeration, filters applied while listing)
    -> update-compare helper (sync_mode=update): groups candidates by target parent
      -> candidates >= update_compare_bulk_threshold (> 0, non-FTP/SFTP target): single listStatus(parent) via HadoopFileSystemProxy, compare len/mtime or checksum against listing
      -> otherwise: bounded point lookups getFileStatus(targetPath) on an executor sized by update_compare_parallelism, submitted-but-incomplete cap = 8 * parallelism
      -> FTP/SFTP target filesystems: always take the directory-listing branch regardless of threshold
    -> SKIP/COPY decision per compare_mode (len_mtime | checksum) feeds the split list returned to FileSourceSplitEnumerator
```

# Findings
**Issue 1: Unconditional directory listing for FTP/SFTP targets is a default-behavior change that can regress large flat directories on upgrade**
- Location: `docs/en/connectors/source/SftpFile.md:384`
- Why it matters: - The new docs explicitly state that FTP and SFTP targets "always use directory listing", regardless of `update_compare_bulk_threshold`, and the SFTP doc even concedes that "A flat SFTP directory must still return all entries because SFTP `READDIR` has no server-side modification-time filter."
- Root cause: the new comparison path in `UpdateFileMetadataLoader` hard-routes FTP/SFTP targets to the bulk-listing branch instead of the bounded point-lookup branch, so the option's documented default (`0` = disabled bulk listing) does not hold for these two connectors.
- Impact: an existing `sync_mode=update` job whose target is an SFTP/FTP server with a very large flat directory but only a handful of candidate files will, after upgrade, enumerate the entire remote directory on every discovery cycle — higher latency, higher memory on the enumerator, and more load on the remote server, with no config knob to opt out. This is exactly the "unexpectedly expensive target directory scan" the default `0` claims to avoid.
- Fix: either honor `update_compare_bulk_threshold=0` for FTP/SFTP targets too (fall back to per-file `getFileStatus` point lookups, which both `SeaTunnelFTPFileSystem` and `SFTPFileSystem` support), or gate the forced listing behind an explicit option and call out the upgrade impact in the docs and release notes.
- Evidence: `docs/en/connectors/source/FtpFile.md:535  For non-FTP/SFTP targets, a positive value switches comparison to one directory listing when the candidate count under a target parent reaches the threshold. The default `0` disables automatic bulk listing and uses bounded point lookups, avoiding an unexpectedly expensive target directory scan. FTP and SFTP targets always use directory listing.`
- Suggested fix: This documents a default-behavior change for FTP/SFTP targets: they "always use directory listing" even though `update_compare_bulk_threshold=0` is supposed to disable bulk listing. For an existing `sync_mode=update` job pointed at a large flat SFTP directory with few candidate files, upgrading will silently switch from per-file lookups to a full `READDIR` of the whole directory with no way to opt out — the exact "unexpectedly expensive target directory scan" the default `0` is meant to avoid. Please either honor the point-lookup path for FTP/SFTP targets when the threshold is 0, or gate the forced listing behind an explicit option and flag the upgrade impact prominently.
- Severity: High

**Issue 2: New concurrent lookup path and FTP/SFTP filesystem listing changes lack e2e coverage; only unit tests were added**
- Location: `docs/en/connectors/source/LocalFile.md:511`
- Why it matters: - This PR introduces a bounded concurrent lookup engine (`UpdateFileMetadataLoader`, +296 lines), a new streaming listing abstraction (`StreamingFileSystem`, `FileStatusListingSession`), and ~190 lines of changes to `SeaTunnelFTPFileSystem`/`SFTPFileSystem`, but the changed-files list contains only unit tests (`UpdateFileMetadataLoaderTest`, `FileDiscoveryScannerTest`, `SeaTunnelFTPFileSystemTest`) and no changes under `seatunnel-e2e`.
- The documented behavior — bounded concurrency with an in-flight cap of "eight times this value" — involves thread-pool lifecycle, exception propagation from worker threads back to the enumerator, and interruption/shutdown semantics that unit tests with mocked filesystems typically do not exercise end-to-end on Zeta.
- Impact: a leaked executor or swallowed lookup exception in the enumerator would only surface as hung or partially-completed `sync_mode=update` jobs in production; the FTP/SFTP `listStatus` changes also touch a Hadoop `FileSystem` implementation shared with other read paths, so a regression there affects normal (non-update) reads too.
- Fix: add an e2e case in `connector-file-local-e2e` (and ideally `connector-file-sftp-e2e`) running `sync_mode=update` with `update_compare_parallelism` set and with `update_compare_bulk_threshold` both 0 and >0, asserting the copied/skipped file sets; also add a test that a failing target lookup fails the job promptly rather than hanging.
- Evidence: `docs/en/connectors/source/LocalFile.md:511  For non-FTP/SFTP targets, a positive value switches comparison to one directory listing when the candidate count under a target parent reaches the threshold. The default `0` disables automatic bulk listing and uses bounded point lookups, avoiding an unexpectedly expensive target directory scan.`
- Suggested fix: The bulk-listing vs bounded point-lookup switch documented here is only covered by unit tests in this PR — there is no e2e case exercising `sync_mode=update` with `update_compare_bulk_threshold` > 0 versus 0 on a real filesystem. Given this path involves a concurrent executor and directory enumeration in the enumerator, please add an e2e test in connector-file-local-e2e that verifies identical copy/skip decisions under both settings and that a failing target lookup fails the job instead of hanging.
- Severity: Medium

**Issue 3: Docs promise a 1–64 valid range for update_compare_parallelism but do not define out-of-range behavior; validation must fail fast at config time**
- Location: `docs/en/connectors/source/FtpFile.md:531`
- Why it matters: - All four connector docs state "valid values are `1` through `64`" for `update_compare_parallelism`, but nothing in the docs (or the option table default column) says whether an out-of-range value is rejected, clamped, or silently accepted.
- Root cause: `Options.key(...).intType().defaultValue(8)` in `FileBaseSourceOptions` does not enforce ranges by itself, so unless the read strategy or `UpdateFileMetadataLoader` explicitly validates, a user setting `update_compare_parallelism = 0` or a very large value would only fail deep inside file discovery at runtime (or spin up an oversized thread pool), which differs per engine (Zeta enumerator vs Flink/Spark driver-side discovery).
- Impact: late, hard-to-diagnose failures during split enumeration instead of a clear `OptionValidationException`-style error at job submission; a value like `0` could also produce a zero-thread executor and hang or throw `IllegalArgumentException` from `ThreadPoolExecutor`.
- Fix: validate the range in the source factory / read strategy constructor with a descriptive `IllegalArgumentException` (following the pattern used for `compare_mode`/`update_strategy` validation), add a unit test for the boundary values in `UpdateCompareConfigTest`, and state the failure behavior explicitly in the docs.
- Evidence: `docs/en/connectors/source/HdfsFile.md:91  | update_compare_parallelism | int     | no       | 8                           | Maximum parallelism for sparse target metadata lookups. Valid range: `1-64`.                                                                                                                                                                                                                                                              |`
- Suggested fix: The doc declares a valid range of 1–64 but doesn't say what happens for out-of-range values. Please make sure the connector fails fast at config-validation time (factory/read-strategy constructor) with a clear error for values like `0` or `> 64` — a `0` here would otherwise reach `ThreadPoolExecutor` construction or produce a hang deep in file discovery — and document that behavior here. Also add boundary-value cases to `UpdateCompareConfigTest`.
- Severity: Medium

**Issue 4: `update_compare_bulk_threshold` is advertised in the FtpFile/SftpFile option tables but is a no-op for FTP/SFTP targets per the same docs**
- Location: `docs/en/connectors/source/FtpFile.md:88`
- Why it matters: - The option tables in `FtpFile.md` and `SftpFile.md` list `update_compare_bulk_threshold` with default `0`, yet the parameter description in the very same files says FTP and SFTP targets always use directory listing, meaning the threshold has no effect for the most common configuration of these connectors (FTP source → FTP-style target).
- Impact: users of the FTP/SFTP connectors will tune this option expecting to control bulk-listing behavior and see no change, leading to confusing support issues; the description "For non-FTP/SFTP targets..." is easy to miss below a table that presents the option as generally applicable.
- Fix: in `FtpFile.md` and `SftpFile.md` (both en and zh), annotate the option-table row (e.g., "only effective when the target filesystem is not FTP/SFTP") or restructure the description so the applicability constraint is the first sentence, mirroring how `compare_mode` documents its `update_strategy=strict` dependency.
- Evidence: `docs/en/connectors/source/FtpFile.md:535  For non-FTP/SFTP targets, a positive value switches comparison to one directory listing when the candidate count under a target parent reaches the threshold. The default `0` disables automatic bulk listing and uses bounded point lookups, avoiding an unexpectedly expensive target directory scan. FTP and SFTP targets always use directory listing.`
- Suggested fix: This table row presents `update_compare_bulk_threshold` as a generally applicable option, but the parameter section below (line 535) states "FTP and SFTP targets always use directory listing", i.e. the threshold is a no-op for FTP/SFTP targets. Please add a note in the table (e.g., "only effective for non-FTP/SFTP targets") so FtpFile users don't tune a parameter that has no effect for them.
- Severity: Medium

**Issue 5: Documented 1–64 validation range for update_compare_parallelism must match actual OptionRule/config validation, and out-of-range behavior is unspecified**
- Location: `docs/en/connectors/source/HdfsFile.md:298`
- Why it matters: - Root cause: the docs promise a hard valid range of `1` through `64` for `update_compare_parallelism` across all four connectors, but the docs do not say what happens on out-of-range input (fail-fast vs silent clamp), and this contract must be enforced in `FileBaseSourceOptions`/the source factories rather than only stated in Markdown.
- Impact: if the option definition in `FileBaseSourceOptions.java` does not validate the range, a user setting `update_compare_parallelism = 0` or a negative value could yield a zero-sized or unbounded executor at runtime deep inside `UpdateFileMetadataLoader`, producing an obscure `IllegalArgumentException` from `Executors`/`ThreadPoolExecutor` instead of a clear config error at job submission — inconsistent behavior across Zeta/Flink/Spark submission paths.
- Fix: enforce the 1–64 bound at option-validation time (fail fast with a `FileConnectorException` naming the option and the received value), cover it in `UpdateCompareConfigTest`, and state the fail-fast behavior explicitly in the docs ("values outside 1–64 are rejected at config validation").
- Evidence: `docs/en/connectors/source/LocalFile.md:507  Maximum parallelism for sparse target metadata lookups in `sync_mode=update`. The default is `8`; valid values are `1` through `64`.`
- Suggested fix: The docs commit to a hard `1–64` range but don't say what happens outside it. Please enforce this fail-fast in the option definition (`FileBaseSourceOptions`) / factory validation so an out-of-range value produces a clear config error at submission time (with a `UpdateCompareConfigTest` case for 0, -1, and 65), rather than surfacing as an opaque `ThreadPoolExecutor`/`Executors` `IllegalArgumentException` at runtime inside `UpdateFileMetadataLoader`. Then add one sentence here stating that out-of-range values are rejected at config validation.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Unconditional directory listing for FTP/SFTP targets is a default-behavior change that can regress large flat directories on upgrade

- Issue 2: New concurrent lookup path and FTP/SFTP filesystem listing changes lack e2e coverage; only unit tests were added
- Issue 3: Docs promise a 1–64 valid range for update_compare_parallelism but do not define out-of-range behavior; validation must fail fast at config time
- Issue 4: `update_compare_bulk_threshold` is advertised in the FtpFile/SftpFile option tables but is a no-op for FTP/SFTP targets per the same docs
- Issue 5: Documented 1–64 validation range for update_compare_parallelism must match actual OptionRule/config validation, and out-of-range behavior is unspecified

DanielLeens:
Thanks @SEZ9. I agree the update-comparison optimization still needs one more pass on the current head, especially around the unconditional FTP/SFTP listing behavior, the documented-but-not-enforced range semantics, and the missing coverage for the bounded concurrent path. From Daniel's side I would wait for a code update that resolves those points before another full re-review.

DanielLeens (approved):

I also recorded the base/head sync fact for the queue. This branch is still `diverged` with `behind_by=78`, but because the current required Build signal is green, I am not asking for a sync-only retry from Daniel's side here. If GitHub's merge gate later asks for an update branch step, please handle that as a merge-gate action.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=5`, `behind_by=86`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

DanielLeens:
Rechecked the same unchanged head `51cf5e76f1bb` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
File source startup
  -> AbstractReadStrategy.setPluginConfig(...)
      -> validateUpdateSyncConfig(...)
  -> AbstractReadStrategy.getFileNamesByPath(...)
      -> FileDiscoveryScanner.scan(...)
      -> UpdateFileMetadataLoader.load(...)
          -> bulk directory listing for dense groups
          -> bounded point lookups for sparse groups
      -> shouldSyncFileInUpdateMode(...)
```

## Merge conclusion

1. Blocking items

2. Current gate
- The current CI signal is now green: https://github.com/apache/seatunnel/actions/runs/28992887399/job/86036383037

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current diff and the relevant file update discovery/comparison runtime path.

The threshold and point-lookup blocker is resolved across FTP/SFTP, and `Build` is green. The remaining gate is repository review/maintainer approval.

davidzollo:
@SEZ9 Please help review this PR again,thx

davidzollo (approved):
LGTM

PR #11316, opened 2026-07-03
[Test][E2E] Wait for Doris multi-read sink visibility

DanielLeens:
## What
- wait for the Doris sink row count to match the filtered source result before asserting sink data
- avoid querying the multi-read Doris sink immediately after `executeJob(...)` returns

## Why
`DorisMultiReadIT` currently checks the sink table as soon as the SeaTunnel job exits. In CI, the Doris sink data can become visible slightly later, which makes the test intermittently fail at `assertHasData(sinkDB, table)` even though the job itself succeeds.

## Validation
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-doris-e2e spotless:apply -DskipTests`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-doris-e2e -DskipTests -DfailIfNoTests=false test-compile -Dspotless.check.skip=true -Dcheckstyle.skip=true`
- attempted `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-doris-e2e -Dtest=DorisMultiReadIT -DfailIfNoTests=false -DskipITs=false test -Dspotless.check.skip=true -Dcheckstyle.skip=true`, but local Testcontainers execution is blocked because this machine currently has no Docker socket (`/var/run/docker.sock` missing)

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-doris-e2e/src/test/java/org/apache/seatunnel/e2e/connector/doris/DorisMultiReadIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-doris-e2e/src/test/java/org/apache/seatunnel/e2e/connector/doris/DorisMultiReadIT.java
@@ -52,12 +52,21 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


davidzollo (approved):
+1

PR #11329, opened 2026-07-04
[Docs][Connector-V2] Improve HugeGraph Qdrant and Milvus docs

DanielCarter-stack:
## Summary

- Improve HugeGraph sink docs with field filtering guidance, schema_config details, supported type mapping, and write behavior notes.
- Clarify Qdrant sink collection/vector requirements and flush behavior.
- Update Milvus source feature support for column projection, parallel reads, and multi-table reads, and add Milvus sink notes for upsert and collection loading.
- Keep English and Chinese connector docs aligned.

## Validation

- ./mvnw spotless:apply


DanielLeens (commented):
Thanks for the documentation refresh here. I reviewed the latest head against the Milvus connector option definitions in the current source tree.

## What this PR tries to improve
- User pain: the HugeGraph / Qdrant / Milvus docs are sparse and miss practical usage notes.
- Fix approach: expand the connector docs with option tables, usage notes, and richer examples.
- One-line summary: the Milvus direction is helpful overall, but one of the newly documented compatibility claims is not backed by the current connector code.

## Runtime / option path I checked
```text
Milvus sink config
  -> MilvusBaseOptions
      -> URL / COLLECTION / TOKEN
  -> MilvusSinkOptions
      -> DATABASE / DATA_SAVE_MODE / ENABLE_UPSERT / ...
  -> MilvusSinkFactory.optionRule(...)
      -> uses the current Milvus option set
```

## Finding 1
- Location:
  - `docs/en/connectors/sink/Milvus.md:56`
  - `seatunnel-connectors-v2/connector-milvus/src/main/java/org/apache/seatunnel/connectors/seatunnel/milvus/config/MilvusBaseOptions.java:33-37`
- Problem:
  the new sink option table says that the deprecated `collection_name` key is accepted as an alias for `collection`, but the real option definitions only declare `Options.key("collection")`. I could not find any `collection_name` alias handling in the current Milvus connector path.
- Risk:
  - users following the new docs can write a config that looks documented but is rejected by the actual connector
  - because this is framed as a backward-compatibility alias, it is especially misleading for existing users trying to migrate old configs safely
- Best fix:
  - Option A, preferred: remove the alias claim from the docs and document only `collection`
  - Option B: if backward compatibility is intended, add the real alias handling in the connector before documenting it
- Severity: High

## Review conclusion
1. Blocking items
- Finding 1: the docs claim `collection_name` is an accepted alias, but the current Milvus connector only exposes `collection`.

2. Suggested follow-up
- The rest of the Milvus doc expansion looks directionally useful once that unsupported alias claim is removed or implemented.

Happy to re-review after the documented option contract matches the real Milvus sink configuration surface.

DanielCarter-stack:
Updated the PR branch to remove the unsupported documented option contract and keep the connector docs aligned with the current option surface.

DanielLeens (commented):
Sorry for missing this in the last review — that's on me. I re-reviewed the latest head from scratch, and I need to correct my earlier Milvus conclusion.

The latest docs revision removed the `collection_name` compatibility note, but the current runtime still accepts that alias on the real sink path:
```text
Milvus sink config
  -> MilvusBaseOptions.COLLECTION
      -> Options.key("collection").withFallbackKeys("collection_name")
  -> Milvus sink/source tests also still cover `collection_name`
```

So on the current head the new mismatch is the opposite of what I said earlier: the docs now understate the current compatibility contract instead of overstating it.

1. Blocking items
- Please either keep documenting `collection_name` as the current fallback alias, or remove the alias from code first and then update the docs in the same change.

2. Suggested follow-up
- The rest of the Milvus / HugeGraph / Qdrant doc refresh still looks directionally useful once the option contract is aligned again.

This is a carryover from the previous version that I should have caught earlier. Happy to re-review once the docs and runtime contract match again.

DanielCarter-stack:
Updated the branch to document `collection_name` as the backward-compatible alias for `collection`, matching `MilvusBaseOptions.COLLECTION.withFallbackKeys("collection_name")`.

DanielLeens (approved):
Thanks for the update. I checked the latest doc diff against the current HugeGraph / Qdrant / Milvus connector option surface.

What I verified
```text
Milvus option parsing
  -> MilvusBaseOptions.COLLECTION [MilvusBaseOptions.java:37]
      -> withFallbackKeys("collection_name")
  -> old configs using collection_name are still accepted
```

Findings
- The `collection_name` compatibility note is real and matches the current code.
- The updated docs make an important backward-compatible alias explicit instead of leaving users to guess.
- The latest Build is green.

1. Blocking items
- None from my side.

2. Non-blocking suggestions
- None.

This looks good to merge.

davidzollo (approved):

PR #10715, opened 2026-04-05
[Feature] [Connectors-v2] Add Doris sink redirect enhancement

yzeng1618:
### Purpose of this pull request

This PR enhances Doris sink redirect handling and adds an opt-in direct-to-BE write path for `connector-doris`.
https://github.com/apache/seatunnel/issues/10697

### Does this PR introduce _any_ user-facing change?
Yes.

Before this change, Doris sink always sent stream load requests to FE and relied on FE redirect. When users encountered a raw `307 Temporary Redirect`, the connector only exposed limited diagnostics, and there was no explicit way to opt in to direct BE writes.

After this change:

- users can configure `benodes` and `direct_to_be=true` to send stream load write requests directly to Doris BE
- when `direct_to_be=true` and `sink.enable-2pc=true`, data writes go to BE while commit/abort still go through FE
- the connector fails fast if `direct_to_be=true` but `benodes` is missing or blank
- redirect-related exceptions now include more context such as request path, redirect location, `direct_to_be` / `sink.enable-2pc` state, and troubleshooting guidance
- Doris connector docs now describe the new options and redirect troubleshooting guidance

### How was this patch tested?
Unit tests
E2E

### Check list

* [ * ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [ * ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config)

PR #11330, opened 2026-07-04
[Fix][Connector-V2] Fix Doris stream load waitForContinue timeout when FE redirect is slow

gnehil:

Fix a Doris sink stream load failure that occurs when FE is slow to return the 307 redirect.

The stream load client sends `Expect: 100-continue` and relies on FE's 307 redirect to reach BE. `HttpUtil` builds the HttpClient without a custom `HttpRequestExecutor`, so the default `waitForContinue` timeout of **3s** applies. When FE is busy (heavy load / FullGC / network jitter) and the 307 arrives later than 3s:

1. The client stops waiting and sends the non-repeatable `InputStreamEntity` body to FE;
2. When the 307 finally arrives, `RedirectExec` detects the consumed entity (`RequestEntityProxy.isRepeatable == false`), silently refuses to follow the redirect, and returns the raw 307 response to the caller;
3. The load fails with `STREAM_LOAD_FAILED ... HTTP/1.1 307 Temporary Redirect`.

This PR raises the `waitForContinue` timeout to **60s** by setting `.setRequestExecutor(new HttpRequestExecutor(60_000))`, aligned with the same fix in doris-flink-connector (apache/doris-flink-connector#522).

The longer timeout is scoped to the **stream-load upload path only** - the request that sends a non-repeatable `InputStreamEntity` body and actually needs the longer window. `HttpUtil` now exposes two clients: `getHttpClient()` (default timeout) for the empty-entity control requests - `DorisStreamLoad` pre-commit / abort and `DorisCommitter` 2PC commit / abort - and `getStreamLoadHttpClient()` (60s) used only by `DorisStreamLoad.startStreamLoad`. This keeps a slow FE from stretching checkpoint completion and transaction cleanup; only the upload waits longer for the redirect.

**Relationship to #10715**: that PR added an opt-in direct-to-BE write path (`direct_to_be=true` + `benodes`), which avoids the FE redirect entirely and is one way to work around this problem. However, direct BE access is not always feasible — e.g. in cloud / Kubernetes / cross-network deployments only FE is exposed to the client while BE addresses are internal and unreachable, or BE topology changes dynamically so a static `benodes` list is hard to maintain. For those cases the default FE-redirect path is still the only option, and this fix makes it reliable when FE is slow to return the 307.

No. No config option, public API, or dependency changes — only the stream-load upload path wait-for-continue timeout is raised from 3s to 60s, matching doris-flink-connector's default behavior.

Added `HttpUtilTest#testStreamLoadClientWaitsForSlowFeRedirect`, which reproduces the scenario end-to-end:

- A raw-socket FE stub reads the request headers, delays the 307 redirect by 4s (longer than the old 3s default, far shorter than the new 60s) without sending `100 Continue`. (A JDK `HttpServer` cannot simulate this — it auto-replies `100 Continue` before the handler runs.)
- A JDK `HttpServer` BE stub receives the redirected request and returns 200.
- The client sends an `HttpPut` built by `HttpPutBuilder` with a non-repeatable `InputStreamEntity`.
- Asserts the request follows the 307 to BE, returns 200, and BE receives the full body.

Before the fix this test fails with `expected: <200> but was: <307>` (instrumentation confirmed the body was prematurely sent to FE and BE was never hit); after the fix it passes.

Also ran the full connector-doris unit test suite (111 tests, all passing), including the existing `DorisStreamLoadTest` / `DorisCommitterTest` 307-handling regressions, plus `./mvnw spotless:apply` and `./mvnw -DskipTests verify`.

### Check list

  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md) — N/A, no new dependency
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs — N/A, no user-facing change
* [x] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR. — N/A, no incompatible change
* [ ] If you are contributing the connector code, please check that the following files are updated: — N/A, bug fix in existing connector

### DIFF
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisStreamLoad.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisStreamLoad.java
@@ -85,6 +85,10 @@ public class DorisStreamLoad implements Serializable {
     private final RecordStream recordStream;
     @Getter private Future<CloseableHttpResponse> pendingLoadFuture;
     private final CloseableHttpClient httpClient;
+    // Dedicated client for the stream-load upload path. Unlike the control requests that share
+    // httpClient, the upload sends a non-repeatable body and needs HttpUtil's longer
+    // waitForContinue window so a slow FE 307 redirect does not force the body to FE first.
+    private final CloseableHttpClient streamLoadHttpClient;
     private final ExecutorService executorService;
     private volatile boolean loadBatchFirstRecord;
     private volatile boolean loading = false;
@@ -113,6 +117,7 @@ public class DorisStreamLoad implements Serializable {
         this.streamLoadProp = dorisSinkConfig.getStreamLoadProps();
         this.enableDelete = dorisSinkConfig.getEnableDelete();
         this.httpClient = httpClient;
+        this.streamLoadHttpClient = new HttpUtil().getStreamLoadHttpClient();
         this.executorService =
                 new ThreadPoolExecutor(
                         1,
@@ -297,7 +302,7 @@ public class DorisStreamLoad implements Serializable {
                             () -> {
                                 log.info("start execute load");
                                 return HttpUtil.executeWithRedirectTracking(
-                                        httpClient,
+                                        streamLoadHttpClient,
                                         putBuilder.build(),
                                         loadUrlStr,
                                         directToBe,
@@ -356,6 +361,13 @@ public class DorisStreamLoad implements Serializable {
     }
 
     public void close() throws IOException {
+        if (null != streamLoadHttpClient) {
+            try {
+                streamLoadHttpClient.close();
+            } catch (IOException e) {
+                throw new IOException("Closing streamLoadHttpClient failed.", e);
+            }
+        }
         if (null != httpClient) {
             try {
                 httpClient.close();
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/util/HttpUtil.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/util/HttpUtil.java
@@ -27,6 +27,7 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.DefaultRedirectStrategy;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
+import org.apache.http.protocol.HttpRequestExecutor;
 import org.apache.http.protocol.RequestContent;
 
 import java.io.IOException;
@@ -35,8 +36,16 @@ import java.util.List;
 
 /** util to build http client. */
 public class HttpUtil {
+    // Stream load sends "Expect: 100-continue" and relies on FE's 307 redirect to reach BE.
+    // The default waitForContinue timeout of HttpRequestExecutor is only 3s: when FE is busy
+    // (heavy load / FullGC), the client stops waiting, sends the non-repeatable request body
+    // to FE, and can no longer follow the late 307 redirect. Wait 60s instead, aligned with
+    // doris-flink-connector.
+    private static final int WAIT_FOR_CONTINUE_TIMEOUT_MS = 60 * 1000;
+
     private final HttpClientBuilder httpClientBuilder =
             HttpClients.custom()
+                    .setRequestExecutor(new HttpRequestExecutor(WAIT_FOR_CONTINUE_TIMEOUT_MS))
                     .setRedirectStrategy(
                             new DefaultRedirectStrategy() {
                                 @Override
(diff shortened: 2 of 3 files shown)


DanielLeens (commented):
Thanks for working on this. I reviewed the latest head against the real Doris stream-load and 2PC commit path.

## What this PR solves
- User pain: when the FE is slow to return the `307` redirect for stream load, the default Apache HttpClient `waitForContinue` timeout is short enough that the client can start sending the non-repeatable request body to FE first, and then the late redirect cannot be followed.
- Fix approach: raise the `HttpRequestExecutor` `waitForContinue` timeout from the default 3s to 60s.
- One-line summary: this does protect the late-redirect stream-load write path, but the timeout is currently widened for every Doris HTTP request that shares `HttpUtil`, not only for the non-repeatable upload path that actually needs it.

## Runtime path I checked
```text
stream-load write path
  -> DorisStreamLoad [writer path]
      -> HttpPutBuilder.addCommonHeader()
      -> HttpUtil.getHttpClient()
      -> HttpUtil.executeWithRedirectTracking(..., requestStage="stream-load")
      -> FE 307 redirect -> BE upload

2PC / commit path
  -> DorisCommitter.commitTransaction(...)
      -> HttpPutBuilder.addCommonHeader()
      -> setEmptyEntity()
      -> same HttpUtil.getHttpClient()
      -> HttpUtil.executeWithRedirectTracking(..., requestStage="commit")
```

## Finding 1
- Location:
  - `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/util/HttpUtil.java:39-49`
  - `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java:59-64`
  - `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisStreamLoad.java:137-155`
- Problem:
  `WAIT_FOR_CONTINUE_TIMEOUT_MS = 60000` is configured once on the shared `HttpUtil` client, and that same client is used by both the non-repeatable stream-load upload path and the empty-entity 2PC / commit / abort requests.

  The late-307 failure mode is specific to the upload request body path, because once the body has been sent to FE it cannot be replayed safely to BE. That same risk does not apply in the same way to the empty-entity commit / abort requests, but this patch still makes them all wait up to 60s before continuing.
- Risk:
  - slow FE / network problems now stretch commit / abort retries as well, even though those requests do not need the long 100-continue window
  - that can make checkpoint completion / transaction cleanup noticeably slower on unhealthy FE nodes
  - the fix becomes broader than the actual bug surface, which makes later Doris HTTP behavior harder to reason about
- Best fix:
  - Option A, preferred: scope the longer `waitForContinue` executor to the stream-load upload client only
  - Option B: keep two clients in `HttpUtil`, one for non-repeatable upload traffic and one for commit / abort / pre-commit requests
- Severity: High

## Test note
- The new redirect test is useful and does exercise the delayed FE redirect case.
- The blocker is not the test itself; it is the widened timeout scope of the production client.

## Review conclusion
1. Blocking items
- Finding 1: the 60s `waitForContinue` timeout is currently applied to all Doris HTTP requests that share `HttpUtil`, not just the stream-load upload path that needs it.

2. Suggested follow-up
- None beyond narrowing the timeout scope to the upload path.

Happy to re-review once the longer wait is isolated to the stream-load request path.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest rebased head from scratch against the real Doris stream-load and 2PC commit path again.

The same blocker still stands on `6917ec643ec582923c762c7aee033ed5f78eeb28`: `WAIT_FOR_CONTINUE_TIMEOUT_MS = 60000` is still configured on the shared `HttpUtil` client, and that same client is still used by both the non-repeatable stream-load upload path and the empty-entity commit / abort requests.

Runtime path revisited:
```text
stream-load upload
  -> DorisStreamLoad -> HttpUtil.getHttpClient() -> FE 307 redirect -> BE upload

2PC commit / abort
  -> DorisCommitter -> same HttpUtil.getHttpClient() -> empty-entity control requests
```

1. Blocking items
- The 60s `waitForContinue` timeout is still widened for every Doris HTTP request that shares `HttpUtil`, not only for the non-repeatable upload path that needs it.

2. Suggested follow-up
- The new delayed-redirect test is still useful to keep once the longer wait is scoped to the upload path only.

Happy to re-review once the longer timeout is isolated to the stream-load request path.

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete diff at `13a13fb09773091e9d427bbe0a34d4cda2298cbb`, and the earlier source blocker is resolved: the 60-second `waitForContinue` executor is now isolated to the non-repeatable stream-load upload client, while pre-commit, commit, and abort control requests continue to use the original client and default timeout.

I also ran the focused `HttpUtilTest` locally. It followed the delayed FE 307 redirect, delivered the complete body to the BE stub, and passed (`1` test, `0` failures/errors; module build success).

I do not see a new source-level blocker on this head. The technical verdict is **clear, waiting for CI**: the required `Build` is still queued. This branch is 87 commits behind the current `dev` (`compare status: diverged`), but there is no unrelated failing check to diagnose yet, so I am recording that sync fact rather than asking for an additional rebase solely because it is behind.

DanielLeens (commented):

Test-stability/resource note (non-blocking): the test-owned cached thread pool should be shut down explicitly, and production cleanup should attempt all client/executor closes while preserving the first exception.

DanielLeens (commented):

Test-stability/resource note (non-blocking): the test-owned cached thread pool should be shut down explicitly, and production cleanup should attempt all client/executor closes while preserving the first exception.

davidzollo (approved):
+1
LGTM

PR #11333, opened 2026-07-04
[Docs][Connector-V2] Improve ActiveMQ Sls and Typesense docs

DanielCarter-stack:
## Summary
- Clarify that the ActiveMQ connector is sink-only and can consume rows from any upstream SeaTunnel source.
- Align the SLS sink documentation with the exposed sink option surface by removing the unsupported `log_group_size` setting from EN/ZH docs and examples.
- Expand the Typesense sink option table in EN/ZH docs, including `multi_table_sink_replica`, clearer option descriptions, and Chinese feature wording.

## Verification
- `./mvnw spotless:apply -DskipTests`
- `./mvnw -q -DskipTests -Dskip.spotless=true -pl seatunnel-connectors-v2/connector-activemq,seatunnel-connectors-v2/connector-sls,seatunnel-connectors-v2/connector-typesense -am verify`

Note: a full `./mvnw -q -DskipTests -Dskip.spotless=true verify` run was started but manually interrupted after it stayed running for several minutes after exporting connector metadata.

### DIFF
--- a/docs/en/connectors/sink/Activemq.md
+++ b/docs/en/connectors/sink/Activemq.md
@@ -6,7 +6,8 @@ import ChangeLog from '../changelog/connector-activemq.md';
 
 ## Description
 
-Write SeaTunnel rows to an ActiveMQ queue. Each row is serialized as a JSON text message.
+Write SeaTunnel rows to an ActiveMQ queue. Each row is serialized as a JSON text message. This is
+a sink-only connector; SeaTunnel does not provide an ActiveMQ source connector.
 
 ## Key features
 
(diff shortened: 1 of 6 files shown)


DanielLeens (approved):
Thanks for the docs cleanup here. I reviewed the latest head against the real ActiveMQ / SLS / Typesense connector option surface instead of only reading the markdown diff.

Runtime path revisited: docs -> connector option definitions -> factory option rules -> actual writer behavior. The sink-only ActiveMQ note matches the current module layout, `TypesenseSinkFactory` does expose `multi_table_sink_replica`, and the SLS writer path still sends one row per write request, so removing the old `log_group_size` promise is aligned with the runtime.

1. Blocking items

2. Suggested follow-up
- None from this round.

davidzollo (approved):
+1
LGTM

PR #11336, opened 2026-07-04
[Docs][Connector-V2] Improve Aerospike DynamoDB and SQS docs

DanielCarter-stack:
## Purpose

Improve connector documentation for Aerospike, Amazon DynamoDB, and Amazon SQS based on the current connector option surface and available integration examples.

## Changes

- Clarify Aerospike type mapping, required record key usage, and `schema.field` behavior for `map`, `string`, and `kv` formats.
- Expand Amazon DynamoDB source and sink option tables with descriptions and add scan/batch-write guidance.
- Clarify that Amazon SQS source and sink URLs can target AWS SQS or SQS-compatible local endpoints.
- Keep English and Chinese connector docs aligned.

## Verification

- `./mvnw spotless:apply`

### DIFF
--- a/docs/en/connectors/sink/Aerospike.md
+++ b/docs/en/connectors/sink/Aerospike.md
@@ -44,14 +44,13 @@ The connector writes to one fixed target set. It does not route records to diffe
 | DOUBLE              | DOUBLE              | 64-bit floating point                                                         |
 | BOOLEAN             | BOOLEAN             | Stored as true/false values                                                   |
 | ARRAY               | BYTEARRAY           | Only support byte array type                                                  |
-| LIST                | LIST                | Support generic list types                                                   |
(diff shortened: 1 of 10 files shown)


DanielLeens (approved):
Thanks for the docs refresh here. I reviewed the latest head against the real Aerospike / AmazonDynamoDB / AmazonSqs connector behavior rather than only reading the markdown delta.

Runtime path revisited: docs -> option definitions -> writer/reader behavior. The Aerospike notes match the current type conversion path, the DynamoDB option descriptions align with the current source/sink configs, and the AmazonSqs source/sink still build clients from the configured `url` / region pair so the SQS-compatible endpoint note is directionally correct.

1. Blocking items

2. Suggested follow-up
- None from this round.

davidzollo (approved):
+1
LGTM

Issue #7006, opened 2024-06-18
[Discuss] support python scripts?


WilliaJing:
Hi,I want to write a transform plugin that can support the execution of python scripts during data conversion.

Do you have any suggestions for me? 
If possible, I will contribute to the community when I finish it.
I look forward to your reply.

Carl-Zhou-CN:
Can you describe it in more detail? What do you want to do with python

PR #11338, opened 2026-07-04
[Feature][Transform-V2] Add Python transform

DanielLeens:

This PR revives the Python transform request discussed in [#7006](https://github.com/apache/seatunnel/issues/7006).

I also reviewed the old draft PR [#8854](https://github.com/apache/seatunnel/pull/8854) for reference. The previous draft was useful for option scope and E2E placement, but the runtime implementation in this PR is a fresh design based on a lightweight stdin/stdout Python worker instead of a `py4j` bridge.

Yes.

Users can now define a `Python` transform in SeaTunnel jobs and choose either inline `source_code` or runtime `source_code_path` to execute Python logic during row transformation.

This PR adds:

- a new `Python` transform plugin for row-level scripted enrichment
- `script_config`, typed output `columns`, and `row_error_handle_way` support
- unit tests and E2E test resources for inline and path-based scripts
- English and Chinese user documentation

Executed locally:

- `./mvnw -pl seatunnel-transforms-v2,seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-2 spotless:apply`
- `./mvnw -pl seatunnel-transforms-v2 -DskipITs -DskipUTs=false -Dtest=PythonTransformTest test`
- `./mvnw -pl seatunnel-transforms-v2 -DskipTests install`

E2E status on this machine:

- `./mvnw -pl seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-2 -DskipUTs=false -DskipITs=false -Dtest=TestPythonTransformIT -DfailIfNoTests=false test`
- The E2E suite could not complete locally because the required starter jars were not built in this checkout, and Docker was not available on this machine for Testcontainers.

Local verification note:

- On this machine, `seatunnel-transforms-v2` currently has an unrelated `ScalaClassParser` constructor mismatch in existing dynamic-compile code during compilation. I used a local-only compile compatibility workaround to run the Python transform unit test and install verification, and then reverted that unrelated file before creating this PR. That workaround is not part of this PR.

### Check list

  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [ ] Update the [`release-note`](https://github.com/apache/seatunnel/blob/dev/release-note.md).


DanielLeens (commented):
Thanks for working on this. I went through the full diff and traced the runtime path from factory creation all the way to `TransformFlowLifeCycle`. The overall direction makes sense, but I still see two code-side blockers before this is ready to merge.

Runtime path I checked:

```text
PythonTransformFactory.createTransform(...)
  -> PythonMultiCatalogTransform
      -> PythonTransform
          -> PythonProcessWorker.open/processRow
              -> worker_template.py process(row, context)
          -> MultipleFieldOutputTransform appends the returned fields

Schema-change path
  -> TransformFlowLifeCycle.received(SchemaChangeEvent)
      -> t.mapSchemaChangeEvent(...)
      -> downstream t.setInputCatalogTables(...)

Close path
  -> TransformFlowLifeCycle.close()
      -> t.close() only on the outer transform object
```

1. `PythonProcessWorker` keeps the old schema cache after schema changes  
   Files: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/python/PythonTransform.java:39`, `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/python/PythonProcessWorker.java:78-131`

   `AbstractCatalogSupportMapTransform` already refreshes `inputCatalogTable` on `SchemaChangeEvent`, and `TransformFlowLifeCycle` already pushes the updated produced catalog downstream. The problem is that `PythonTransform` keeps a long-lived `PythonProcessWorker`, and that worker captures `inputRowType`, `rowToJsonConverter`, `outputConverters`, and the Python-side `context` when it is first created. There is no override here to drop and rebuild that worker after the upstream schema changes.

   That means after an `ALTER TABLE` path, the next rows can still be serialized with the old field names/order/types, so the Python script may miss newly added columns or read the wrong values under the wrong names. I would treat this as a blocker.

   The most direct fix is to override `mapSchemaChangeEvent(...)` and/or `setInputCatalogTable(...)` in `PythonTransform`, close the existing worker, and lazily rebuild it from the new `CatalogTable` before the next row is processed.

2. The outer multi-table wrapper never forwards `close()`, so the Python subprocess is not released  
   Files: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/python/PythonTransformFactory.java:82-85`, `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/python/PythonMultiCatalogTransform.java:30-75`

   The factory always returns `PythonMultiCatalogTransform`, even for the single-table case. At shutdown, `TransformFlowLifeCycle.close()` only calls `close()` on that outer transform object. `PythonMultiCatalogTransform` does not override `close()`, so it inherits the default no-op from `SeaTunnelTransform`. As a result, `PythonTransform.close()` is never reached in the production path, which means the Python worker process and the temporary script files are not cleaned up.

   In the engine runtime, that can leave orphaned Python processes behind on the worker node after jobs finish or restart. I would also treat this as a blocker.

   The fix can be either:
   - add lifecycle delegation in `AbstractMultiCatalogTransform`, or
   - at least override `close()` in `PythonMultiCatalogTransform` and close every inner transform explicitly.

Test notes:

- Coverage is still missing for the real factory-returned wrapper path and for schema-change handling. `PythonTransformTest` directly instantiates `PythonTransform`, so it cannot catch the two lifecycle issues above.
- Test stability is not ideal yet either: `TestPythonTransformIT` installs `python3` on the fly through the distro package manager when the image does not already have it. That adds an external mirror/network dependency to CI. I do not consider that the main blocker compared with the two code issues above, but it is still a risk worth tightening up.

Merge conclusion: fix before merge.

Blocking items:
- Rebuild the cached Python worker when the upstream schema changes.
- Ensure the outer transform wrapper actually closes the inner `PythonTransform` instances.

Non-blocking follow-ups:
- Add regression coverage for the wrapper lifecycle and schema-change path.
- Prefer an E2E image/runtime path that already has Python instead of installing it during the test.

Once those two blockers are fixed, I’m happy to take another look at the updated head.

DanielLeens:
Rechecked the latest head `028df6b403` again. The stderr-collector cleanup added in this revision is a good improvement, but the two earlier Daniel blockers are still open on the current version.

## What changed in this new revision
- `PythonProcessWorker.close()` now waits for the stderr collector thread to stop.
- `PythonTransformTest` adds coverage for the direct `PythonTransform.close()` path.

That is directionally good, but it still does not close the two production-path lifecycle gaps from the previous round:

1. **`PythonTransform` still keeps the old `PythonProcessWorker` across schema changes.**  
   Files: `PythonTransform.java:38-39,108-113`, `PythonProcessWorker.java:76-135`

   The worker still captures `inputRowType`, `rowToJsonConverter`, `outputConverters`, and the Python-side context when it is first created. I still do not see a schema-change path that closes and rebuilds that worker after the upstream `CatalogTable` changes, so the transform can keep serializing rows with stale schema metadata on the next records.

2. **`PythonMultiCatalogTransform` still does not forward `close()` to the inner `PythonTransform` instances.**  
   Files: `PythonTransformFactory.java:82-85`, `PythonMultiCatalogTransform.java:29-75`

   The production factory path still returns `PythonMultiCatalogTransform`, and that wrapper still does not override `close()`. So even though the direct `PythonTransform.close()` path is better now, the real factory-returned runtime object still does not guarantee that the inner Python worker processes are released.

The new test only covers the direct inner transform close path. It still does not cover:
- the factory-returned wrapper close path
- the schema-change rebuild path

So from my side the merge conclusion is unchanged on the latest head:
- fix required before merge
- blocking items:
  1. rebuild the cached Python worker after schema changes
  2. ensure the outer multi-table wrapper actually closes all inner Python transforms

Once those two blockers are fixed on top of this revision, I’m happy to re-review again.

DanielLeens (commented):
Thanks for working on this. I reviewed the current diff from the transform factory path down to the Java/Python worker protocol. The overall direction is promising, but I found one production-path blocker around the subprocess communication boundary.

## What this PR adds
- A new `Python` transform plugin.
- A long-lived Python worker process managed by `PythonProcessWorker`.
- A JSON-line protocol: Java sends the runtime context once, then sends one row per request; Python returns one JSON response per row.
- Docs plus UT/E2E coverage for the happy path and Python-side errors.

## Runtime path checked
```text
Job config uses transform = Python
  -> PythonTransformFactory.createTransform(...)
      -> PythonMultiCatalogTransform
          -> buildTransform(...) per input catalog table
              -> new PythonTransform(inputCatalogTable, PythonTransformConfig)

Task open
  -> PythonTransform.open()
      -> PythonProcessWorker.open()
          -> write temporary worker_template.py
          -> resolve script path or materialize inline source_code
          -> ProcessBuilder(python, worker_template.py, user_script.py)
          -> start stderr collector
          -> initializeRemoteContext()
              -> write context JSON to stdin
              -> wait for one stdout response line

Per row
  -> PythonTransform.getOutputFieldValues(inputRow)
      -> PythonProcessWorker.processRow(inputRow)
          -> write {id,row} JSON to stdin
          -> wait for one stdout response line
          -> response.error -> ErrorDataTransformException
          -> response.result -> convert back to SeaTunnel field values

Close / schema change
  -> PythonTransform.close() or mapSchemaChangeEvent(...)
      -> PythonProcessWorker.close()
          -> close stdin
          -> wait/destroy Python process
          -> delete temporary scripts
```

## Issue 1: Python initialization and per-row transform calls can block forever without a timeout
- Location: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/python/PythonProcessWorker.java`, `initializeRemoteContext(...)` and `processRow(...)`
- Problem: both paths write a request and then call `stdoutReader.readLine()` directly. If the Python worker never responds during initialization, or if a user function hangs / waits on an external service / fails to emit a protocol response for one row, the Java transform thread blocks indefinitely. This happens on the main transform path, not in a background helper.
- Risk: a single stuck Python call can hang the whole task data flow without a clear SeaTunnel timeout error. It can also delay cancellation/checkpoint progress and make the failure hard to diagnose.
- Suggested fix: add explicit initialization and per-row response timeouts. For example, read worker responses through a dedicated thread or bounded queue, wait with a configured timeout, and destroy/restart the worker with the recent stderr tail in the thrown `TransformException` when the timeout is reached. Alternatively, make the protocol asynchronous and match responses by request id with a bounded wait.
- Severity: High

## Tests
The tests cover normal transformation, script-path execution, and Python-side error handling. They do not cover worker initialization hang, row-level function hang, invalid protocol output, or close/cancel while a row is blocked. Once the timeout boundary is added, those cases should be covered as regression tests.

### Merge conclusion: changes requested
1. Blocking items
- Add a clear timeout/cancellation boundary for Python worker initialization and per-row responses so `processRow` cannot block forever.

2. Suggested follow-up
- Add tests for no-response/hanging Python code, invalid JSON/protocol output, and close while processing.

This is a useful feature and the architecture is heading in the right direction. I think it can be mergeable once the subprocess timeout boundary is closed.

DanielLeens:
Thanks for the update. I re-reviewed the latest head `6070c76036` from the factory-returned multi-table wrapper down to the Java/Python worker protocol.

The current head does close two earlier gaps from my previous rounds:
- the schema-change path now invalidates and rebuilds the cached worker instead of keeping stale schema metadata alive
- the outer multi-table wrapper now closes the inner transforms, so the direct worker cleanup path is no longer lost behind the wrapper

Current runtime path rechecked:
```text
Job config uses transform = Python
  -> PythonTransformFactory.createTransform(...)
  -> PythonMultiCatalogTransform
     -> PythonTransform.open()
        -> PythonProcessWorker.open()
           -> initializeRemoteContext()
              -> write init payload
              -> stdoutReader.readLine()
  -> PythonTransform.getOutputFieldValues(...)
     -> PythonProcessWorker.processRow(...)
        -> write row payload
        -> stdoutReader.readLine()
```

I still see one production-path blocker on the current revision:

1. Python worker initialization and per-row execution can still block forever without a timeout.
- Location: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/python/PythonProcessWorker.java`
- Why this is still real: both `initializeRemoteContext(...)` and `processRow(...)` still perform a direct `stdoutReader.readLine()` after writing to the Python worker. If initialization hangs, or a user function never responds for one row, the transform thread blocks indefinitely on the main data path.
- Risk: a single stuck Python call can stall task progress, cancellation, and troubleshooting with no explicit SeaTunnel timeout boundary.
- Best fix: add explicit initialization and per-row response timeouts, ideally through a dedicated response-reading thread or bounded queue, and fail fast with recent stderr context when the timeout is hit.

The current tests still do not cover the no-response / hanging-worker path, which is the exact lifecycle boundary that remains open.

### Merge conclusion: fix before merge
Blocking items:
1. Add a clear timeout / cancellation boundary for Python worker initialization and per-row responses.

Once that timeout boundary is closed on the latest head, I’m happy to re-review again.

davidzollo:
Superseded by #11390 because CI capacity on the DanielLeens fork is exhausted. Continuing review and CI on the davidzollo fork.

PR #11357, opened 2026-07-05
[Test][E2E] Stabilize MySQL CDC schema change order assertions

DanielLeens:
## Why
- shared CI is still hitting `MysqlCDCWithSchemaChangeIT.testMysqlCdcWithSchemaEvolutionCase` with a schema-order mismatch after schema evolution
- the current failure is not covered by an open dedicated PR

## What's changed
- compare MySQL CDC schema descriptions by column name instead of raw `DESC` row order
- project sink reads with the source column order before asserting row equality
- keep the existing projection checks for the explicit add-column cases

## Validation
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e -am spotless:apply -nsu`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e -am -DskipTests -Dskip.ui=true package -nsu -Dmaven.gitcommitid.skip=true -T 3C` on the previous code-identical revision before the final Javadoc-only touch-up

## Notes
- I did not run local Docker-based E2E in this session because local Docker execution was explicitly disabled.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/MysqlCDCWithSchemaChangeIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/MysqlCDCWithSchemaChangeIT.java
@@ -50,10 +50,12 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


DanielLeens (commented):
Rechecked the latest head from Daniel's side. I reviewed the test logic again and I don't see a new source-side blocker here.

What this PR solves
- It removes a brittle assumption from the MySQL CDC schema-change E2E assertions: source and sink can be semantically aligned while still materializing columns in a different physical order on the sink side.

Runtime path I checked
- The schema-evolution verification path here is `MysqlCDCWithSchemaChangeIT` only.
- The old assertions compared raw `DESC` / `SELECT *` ordering directly.
- The new helpers normalize `DESC` rows by column name and build sink queries using the source column order, which matches the actual Zeta runtime behavior much better.
- The change still keeps Awaitility-based polling and does not add new timing-sensitive sleeps into the verification path.

Merge conclusion
- Conclusion: can merge from Daniel's technical review side.
- Blocking items: none.
- Suggestions: none.

davidzollo (approved):
LGTM

PR #11358, opened 2026-07-05
[Fix][Connector-V2] Exclude dropped columns in postgresql connector

NganWave:
When reading table schema from PostgreSQL, the dropped columns will still be included in the schema.This lead to physical columns not match actual field list in table and make errors in table reading and writing.

This PR fixes it by filtering out dropped columns during schema retrieval, ensuring the connector only works with valid columns.


DanielLeens (changes requested):
Thanks for the fix. The SQL change itself looks directionally right: filtering `a.attisdropped` in `PostgresCatalog.SELECT_COLUMNS_SQL_TEMPLATE` should stop dropped columns from leaking back through catalog introspection.

I do have one blocker before merge:

1. Missing regression coverage for the dropped-column case.
   - The production change is in `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalog.java:42-83`.
   - But the current PR does not add an automated test that actually creates a Postgres table, drops a column, and verifies `getTable()` / column listing no longer returns that dropped field. The existing `PostgresCatalogTest` is still disabled/manual-only.
   - Since this is a catalog metadata regression, I think we need a real regression test here. Otherwise the SQL text can silently drift again later.

Merge conclusion
- Conclusion: fix before merge.
- Blocking items: add an automated regression test for dropped columns in the Postgres catalog path.
- Suggestions: once that test is in place, the SQL change itself looks reasonable.

DanielLeens:
Thanks a lot for the contribution to Apache SeaTunnel.

In addition to the regression-test blocker from my earlier source review on this PR, the current `Build` gate is also still stuck in an `ACTION_REQUIRED` state because the CI workflows are not enabled for the fork yet.

Good job.

To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration:
https://github.com/apache/seatunnel/runs/85227819328

image

DanielLeens:
Thanks for the update. I checked the current head against the latest `dev` again.

- base branch: `dev`
- head SHA: `fab6d1158ffbd38f128c47b62adac4434194ffa8`
- compare status: `diverged`
- ahead_by: `1`
- behind_by: `8`
- current failing signal: `Build`

On this round, the visible failing CI signal still does not look directly tied to the current PR-owned diff from Daniel's side, and the branch is still behind upstream. The lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the same lane still fails on the updated head afterwards, please drop the new failing check link here and I can help narrow it down against the rebased revision.

DanielLeens (changes requested):
Thanks for the update! I re-reviewed the latest head from the PostgreSQL JDBC catalog schema-read path.

## What this PR solves
- User pain: when a PostgreSQL table has a dropped column, `pg_attribute` keeps a placeholder row. If SeaTunnel reads that placeholder as a normal column, catalog schema discovery can expose a non-existent column to downstream source/sink logic.
- Fix approach: `PostgresCatalog.SELECT_COLUMNS_SQL_TEMPLATE` now filters `AND NOT a.attisdropped`.
- One-line summary: the catalog query now matches PostgreSQL's visible column contract.

## Runtime path checked

```text
JDBC catalog table discovery
  -> PostgresCatalog.getTable(TablePath)
      -> SELECT_COLUMNS_SQL_TEMPLATE
          -> pg_class / pg_namespace / pg_attribute / pg_type
          -> a.attnum > 0
          -> a.attisdropped must be filtered
      -> AbstractJdbcCatalog builds SeaTunnel schema from returned rows
```

## Review finding

Issue 1: Missing regression coverage for dropped PostgreSQL columns
- Location: seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalog.java:81
- Problem: the SQL fix itself is correct and the normal catalog path will hit it, but the current PR diff still only changes production code. There is no unit/integration regression that creates a PostgreSQL table, drops a column, and verifies that `getTable(...)` / column discovery no longer returns the dropped placeholder.
- Risk: this exact PostgreSQL catalog edge case can regress silently later because the behavior depends on PostgreSQL's `pg_attribute.attisdropped` metadata rather than an obvious Java branch.
- Suggested fix: please add a focused regression test around the Postgres catalog column query or JDBC catalog table schema discovery. The key assertion should be that a dropped column is absent while the remaining visible columns keep their order.
- Severity: Medium

- Compatibility: fully compatible. This only removes PostgreSQL internal dropped-column placeholders from schema discovery.
- API/config/default/protocol/serialization: unchanged.
- Performance: negligible; one boolean predicate on the existing catalog query.
- Error handling/logging: unchanged.
- Test stability: no test code was added in this PR revision.

## Merge conclusion

1. Blocking items
- Issue 1: please add regression coverage for the dropped-column catalog path.

2. Suggested follow-up
- None beyond the regression test above.

Overall, the implementation direction is right and very small. Once the regression coverage is in place, I do not see a source-side blocker from Daniel's side.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest PostgreSQL / OpenGauss catalog change.

# What this PR fixes
The SQL-side fix is correct in direction: dropped PostgreSQL columns should be excluded from catalog schema discovery so SeaTunnel does not expose stale columns after `ALTER TABLE ... DROP COLUMN`.

```text
Postgres / OpenGauss catalog schema read path
  -> catalog.getTable(tablePath)
  -> PostgresCatalog.getSelectColumnsSql(tablePath)
  -> pg_attribute rows are joined into CatalogTable columns
  -> dropped PostgreSQL columns must be filtered with NOT a.attisdropped
```

# Current review findings
## Issue 1: The regression coverage is still disabled, so CI will not protect this fix
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalogTest.java:36`
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/opengauss/OpenGaussCatalogTest.java:35`
- What I found: the production SQL now contains `AND NOT a.attisdropped`, which is the right filter. However, the new/updated regression tests are annotated with `@Disabled("Please Test it in your local environment")` and depend on hard-coded external database endpoints. That means the PR still has no executable CI regression for the dropped-column path.
- Why this blocks merge: the previous blocker was specifically about making this behavior test-protected. A disabled local-only test documents the scenario, but it will not catch a future regression.
- Suggested fix: add an executable unit-level regression around the generated SQL / catalog query path, or convert this into a CI-runnable container-backed test. For a small fix like this, a unit test that asserts the generated select-column SQL contains `AND NOT a.attisdropped` would already protect the core behavior without requiring an external PostgreSQL instance.
- Severity: High

- Fetched the PR head into `seatunnel-review-11358` and reviewed the current full diff.
- Checked the current Build state: `Build` is still `IN_PROGRESS`.

# Merge conclusion
Blocking items:
1. Issue 1: add CI-executable regression coverage for the dropped-column filter.

The SQL fix itself looks right; this should be a small follow-up to make the regression coverage enforceable.

DanielLeens (changes requested):
Thanks for the update! The production SQL change is in the right direction, and it is great to see you trying to turn the previous review into a test. I re-reviewed the latest head from the PostgreSQL catalog path. There are still two blockers before merge, mostly around making the regression test both meaningful and CI-safe.

## What this PR solves
- User pain: PostgreSQL keeps placeholder rows in `pg_attribute` after `ALTER TABLE ... DROP COLUMN`. If SeaTunnel reads those placeholders as normal columns, catalog schema discovery can expose columns that no longer exist.
- Fix approach: `PostgresCatalog.SELECT_COLUMNS_SQL_TEMPLATE` now adds `AND NOT a.attisdropped`.
- One-line summary: the runtime fix is small and correct in direction, but the regression coverage still needs to be executable and behavior-level.

## Runtime path checked

```text
JDBC catalog schema discovery
  -> PostgresCatalog.getTable(TablePath)
  -> PostgresCatalog.getSelectColumnsSql(TablePath)
  -> SELECT from pg_class / pg_namespace / pg_attribute / pg_type
  -> a.attnum > 0 and NOT a.attisdropped
  -> AbstractJdbcCatalog builds CatalogTable columns from the returned ResultSet
```

The normal catalog path will hit the new SQL predicate. The remaining risk is the test path.

## Findings

Issue 1: The new test does not actually verify the dropped-column behavior
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalogTest.java:59`
- Problem: `testGetSelectColumnsSql()` creates a table and then only asserts that the generated SQL string contains `AND NOT a.attisdropped`. It never runs `ALTER TABLE ... DROP COLUMN`, never calls `getTable(...)` / column discovery after the drop, and never asserts that the dropped column is absent while visible columns keep their order.
- Risk: the regression can still fail in real catalog discovery even if this string assertion passes. For example, future refactoring could keep the string but break the actual `ResultSet -> Column -> CatalogTable` path.
- Suggested fix: make the test behavior-level. The core assertion should be: create a PostgreSQL table, drop one column, call the catalog schema-read path, and verify the dropped column is not returned.
- Severity: High
- Already raised by others: This is the same blocker Daniel raised earlier; the current update does not fully close it yet.

Issue 2: Enabling `PostgresCatalogTest` makes CI depend on local external databases
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalogTest.java:32`
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalogTest.java:38`
- Problem: the class is no longer disabled, but `before()` connects to `jdbc:postgresql://127.0.0.1:5432/postgres` with fixed credentials. The same class also contains existing methods that reference fixed MySQL/PostgreSQL databases such as `127.0.0.1:33062`, `mingdongtest`, `st_test`, and `zdykdb`.
- Risk: CI and contributor machines normally do not have those databases. This turns the regression test into a high-risk flaky/failing unit test instead of a reliable guard.
- Suggested fix: either keep this as a pure unit test around the SQL generation, or make it a proper container-backed integration test with all database lifecycle setup inside the test.
- Severity: High
- Already raised by others: Related to Daniel's earlier disabled/local-only coverage blocker.

Issue 3: The POM adds unrelated broad test/build changes for a one-line catalog query fix
- Location: `seatunnel-connectors-v2/connector-jdbc/pom.xml:278`
- Location: `seatunnel-connectors-v2/connector-jdbc/pom.xml:460`
- Problem: the PR adds `seatunnel-hadoop3-3.1.4-uber` and a module-level Surefire classpath exclusion for `net.postgis:postgis-jdbc`. These changes are not part of the dropped-column runtime fix, and they are not narrowly scoped to the new regression.
- Risk: connector-jdbc has many dialect tests; changing the module test classpath can hide driver conflicts or introduce unrelated dependency behavior.
- Suggested fix: remove the unrelated POM changes unless they are strictly required by a CI-runnable regression, and document the need if they remain.
- Severity: Medium

## Compatibility / side effects / tests
- Compatibility: the production SQL predicate is compatible; PostgreSQL dropped-column placeholders should not be user-visible columns.
- API/config/default/protocol/serialization: unchanged.
- Performance: negligible for the production query.
- Test stability rating: High risk, because the enabled test class depends on fixed local database endpoints and credentials.
- Local execution: I did not run Maven tests; this review is based on the full diff and call-chain inspection.
- CI: `Build` is currently in progress on this head.

## Merge conclusion

1. Blocking items
- Issue 1: the regression test still does not verify the real dropped-column catalog behavior.
- Issue 2: the enabled test class depends on local external databases and is not CI-safe.

2. Suggested follow-up
- Issue 3: remove or justify the unrelated POM/classpath changes.

The SQL fix itself is very close. A small, deterministic regression test will make this PR much easier to merge confidently.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch.

# What this PR fixes
- User pain: after `ALTER TABLE ... DROP COLUMN`, PostgreSQL keeps metadata placeholder rows in `pg_attribute`. If SeaTunnel reads those rows back as normal columns, catalog schema discovery can expose columns that no longer exist.
- Fix approach: `PostgresCatalog.SELECT_COLUMNS_SQL_TEMPLATE` now filters `NOT a.attisdropped`, and the PR adds a behavior-level OpenGauss container test that creates a table, drops a column, calls catalog discovery, and checks the visible column order.
- One-line summary: the runtime SQL fix was already directionally right; this revision finally turns it into a real regression guard on the actual catalog path.

# Runtime path checked
```text
Catalog schema discovery
  -> JdbcCatalogUtils / catalog callers request table metadata
  -> PostgresCatalog.getTable(TablePath)
      -> PostgresCatalog.getSelectColumnsSql(tablePath)
      -> query pg_class / pg_namespace / pg_attribute / pg_type
      -> a.attnum > 0
      -> NOT a.attisdropped   [new predicate]
      -> AbstractJdbcCatalog maps the ResultSet into CatalogTable columns

Coverage path in this PR
  -> OpenGaussCatalog extends PostgresCatalog directly
  -> JdbcOpenGaussIT creates a table
  -> ALTER TABLE ... DROP COLUMN
  -> catalog.getTable(tablePath)
  -> assert the dropped placeholder is absent and the remaining physical columns keep order
```

# Review result
The previous blocker from my last round is closed for me now.

The important missing piece before was not the SQL predicate itself, but the lack of a CI-safe behavior-level regression. The new `JdbcOpenGaussIT#testCatalogExcludesDroppedColumns` closes that gap in the right way:
- it exercises the real `getTable(...)` catalog path instead of only checking for a SQL string
- it runs on the existing container-backed OpenGauss test infrastructure instead of depending on a contributor's local database
- it validates the observable contract that matters to users: the dropped column disappears while the remaining column order stays correct

I also rechecked the reuse story here. `OpenGaussCatalog` inherits directly from `PostgresCatalog`, so this test is covering the same production SQL template that the PR changes.

# Merge conclusion
### Conclusion: can merge

1. Blocking items

2. Suggested follow-up
- None required for merge.

At this point the remaining gate is CI completion rather than a new source-side blocker from Daniel's side.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `fa096c0f0784` from scratch.

# What this PR fixes
- User pain: PostgreSQL/OpenGauss tables that ever had a column dropped return `........pg.dropped.N........` placeholder columns from catalog schema discovery, so the connector's physical column list doesn't match the real table and reads/writes fail.
- Fix approach: Add `AND NOT a.attisdropped` to the `pg_attribute` filter in `PostgresCatalog.SELECT_COLUMNS_SQL_TEMPLATE` (inherited by `OpenGaussCatalog`), plus an enabled OpenGauss e2e regression test and two `@Disabled` local catalog tests.
- One-line summary: The one-line SQL fix is correct and minimal, but the PR bundles unrelated build changes (hadoop uber-jar dependency, module-wide postgis surefire exclusion), leaves the actual PostgreSQL path without CI coverage, and regresses `PostgresCatalogTest` with a hardcoded private-LAN IP — these need to be cleaned up before merge.

# Runtime chain I rechecked
```text
AbstractJdbcCatalog.getTable(TablePath)  AbstractJdbcCatalog.java
  -> PostgresCatalog.getListColumnsSql(tablePath)  (PostgresCatalog overrides the column-listing hook)
    -> String.format(SELECT_COLUMNS_SQL_TEMPLATE, schema, table)  PostgresCatalog.java:60-84 (new filter `AND NOT a.attisdropped` added at line 81, alongside existing `a.attnum > 0`)
  -> AbstractJdbcCatalog executes the SQL and iterates the ResultSet
    -> AbstractJdbcCatalog.buildColumnsWithErrorCheck(...)  AbstractJdbcCatalog.java
      -> PostgresCatalog.buildColumn(resultSet)  PostgresCatalog.java (dropped-column placeholder rows like `........pg.dropped.3........` no longer reach here)
  -> TableSchema / CatalogTable returned to JdbcSource / JdbcSink table discovery

Inheritance path exercised by the enabled e2e test:
OpenGaussCatalog extends PostgresCatalog  OpenGaussCatalog.java (inherits SELECT_COLUMNS_SQL_TEMPLATE unchanged)
  -> JdbcOpenGaussIT.testCatalogExcludesDroppedColumns()  JdbcOpenGaussIT.java:172-196
    -> openGaussCatalog.getTable(tablePath) -> same chain as above -> asserts columns == [c1, c3, c4]
```

# Findings
**Issue 1: Unrelated `seatunnel-hadoop3-3.1.4-uber` dependency added to connector-jdbc main pom — has nothing to do with the dropped-column fix and pollutes the module's build/classpath**
- Location: `seatunnel-connectors-v2/connector-jdbc/pom.xml:280`
- Why it matters: - Root cause: the pom adds `seatunnel-hadoop3-3.1.4-uber` as a `provided` dependency to `connector-jdbc`, but nothing in the runtime code of this PR (a one-line SQL predicate in `PostgresCatalog`) uses any Hadoop class. This looks like a workaround for a local test-compile/classpath problem rather than a deliberate design decision.
- Impact: the uber jar is huge and drags a large transitive surface into the compile classpath of `connector-jdbc`. Even with `provided` scope it can shadow classes during compilation and unit tests, change dependency convergence for every downstream module that depends on `connector-jdbc`, and slow down CI. It also creates a hidden ordering dependency: if any code accidentally starts referencing a Hadoop class, it will compile here but fail at runtime on engines that do not ship the uber jar (Flink/Spark deployments without the SeaTunnel Hadoop shim).
- Fix: remove this dependency from the main pom. If a specific test class genuinely needs Hadoop classes, add the dependency with `<scope>test</scope>` and explain why in the PR description; otherwise the fix should ship with zero pom changes.
- Suggested fix: This `seatunnel-hadoop3-3.1.4-uber` dependency is unrelated to the dropped-column fix — the only runtime change in this PR is a SQL predicate in `PostgresCatalog`. Adding the Hadoop uber jar (even as `provided`) changes the compile/test classpath for the whole connector-jdbc module and for every module depending on it, and risks code compiling against Hadoop classes that won't exist at runtime on Flink/Spark clusters without the SeaTunnel shim. Please remove it; if a test actually needs it, scope it to `test` and document why.
- Severity: High

**Issue 2: The `attisdropped` filter is only added to the column-listing SQL — other PG-family catalog queries and Postgres-derived catalogs need the same audit to avoid dropped-column placeholders leaking through other paths**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalog.java:81`
- Why it matters: - Root cause: the fix adds `AND NOT a.attisdropped` to `SELECT_COLUMNS_SQL_TEMPLATE` in `PostgresCatalog` only. PostgreSQL keeps dropped columns as `........pg_dropped.N........` placeholder rows in `pg_attribute`, so *every* catalog query that joins on `pg_attribute` (constraint/primary-key discovery, and any subclass such as Redshift-style catalogs that override or add their own column SQL) is exposed to the same bug.
- Impact: if any other metadata query in the Postgres catalog family still returns placeholder attributes, users will keep hitting the same 'physical columns don't match actual field list' errors reported in this PR, but through a different code path (e.g., primary-key based split column selection or schema comparison in the sink's auto-create/schema-evolution flow), which is much harder to diagnose because this PR claims the issue is fixed.
- Fix: audit the remaining metadata SQL in `PostgresCatalog` and its subclasses (`OpenGaussCatalog` inherits this template so it is covered, but any subclass that defines its own `pg_attribute`-based query is not) and apply `AND NOT a.attisdropped` consistently; ideally add a short comment next to the predicate explaining why it is required so future refactors don't drop it.
- Suggested fix: The predicate itself is correct — `pg_attribute` keeps `........pg_dropped.N........` placeholder rows after `DROP COLUMN`, so filtering `attisdropped` here is the right fix. Two follow-ups: (1) please add a brief comment explaining why this predicate exists so a future SQL refactor doesn't drop it, and (2) please audit the other `pg_attribute`-based queries in this catalog family (constraint/PK discovery, and any Postgres-derived catalog that defines its own column SQL rather than inheriting this template) — they are exposed to exactly the same placeholder-row issue and would reintroduce the reported mismatch through a different path.
- Severity: Medium

**Issue 3: Module-wide surefire exclusion of net.postgis:postgis-jdbc changes the test classpath for ALL connector-jdbc unit tests, risking breakage of existing PostGIS-related tests**
- Location: `seatunnel-connectors-v2/connector-jdbc/pom.xml:467`
- Why it matters: - Root cause: the PR adds a `maven-surefire-plugin` configuration at the module level that excludes `net.postgis:postgis-jdbc` from the test classpath. This appears to be a local workaround (likely a JDBC driver registration conflict between the postgis and opengauss drivers when running the new `@Disabled` catalog tests), but it is applied globally, not scoped to the affected tests.
- Impact: connector-jdbc supports PostGIS geometry/geography types via the postgis-jdbc classes; any existing or future unit test in this module that touches PostGIS type conversion will now fail with `ClassNotFoundException` at test time, and the failure will be very hard to diagnose because the exclusion is buried in build config. It also diverges the test classpath from the runtime classpath, so unit tests no longer validate what actually ships.
- Fix: drop this global exclusion. If there is a genuine driver conflict for the OpenGauss/Postgres catalog tests, isolate those tests (e.g., a dedicated surefire execution/profile, or resolve the driver explicitly via `Class.forName` on the configured driver class) instead of removing a real dependency from every test's classpath.
- Suggested fix: Excluding `net.postgis:postgis-jdbc` from surefire at the module level affects every unit test in connector-jdbc, not just the new catalog tests — any test exercising PostGIS geometry type handling will now fail with missing classes, and the test classpath no longer matches the runtime classpath. If this works around a driver conflict for the new OpenGauss/Postgres catalog tests, please scope the workaround to those tests (dedicated surefire execution/profile, or explicit driver loading) instead of a global exclusion.
- Severity: Medium

**Issue 4: The actual PostgreSQL fix has no CI coverage — both new catalog tests are `@Disabled`, and the only enabled regression test runs against OpenGauss**
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/JdbcOpenGaussIT.java:172`
- Why it matters: - Root cause: the one-line fix lives in `PostgresCatalog.SELECT_COLUMNS_SQL_TEMPLATE`, but `PostgresCatalogTest.testGetTableSchemaAfterColsDropped` and the whole `OpenGaussCatalogTest` class are `@Disabled("Please Test it in your local environment")`. The only test that executes in CI is `JdbcOpenGaussIT#testCatalogExcludesDroppedColumns`, which exercises `OpenGaussCatalog` (inherited SQL) but never a real PostgreSQL server.
- Impact: PostgreSQL and OpenGauss diverge in catalog behavior (OpenGauss is a fork with its own `pg_attribute` semantics in newer versions), so a regression in the Postgres path — the connector this PR is titled for — would not be caught by CI. The `attisdropped` predicate is exactly the kind of catalog-query detail that can regress during future SQL refactors.
- Fix: add the same dropped-column assertion to the PostgreSQL e2e suite (`JdbcPostgresIT` in connector-jdbc-e2e, which already spins up a Postgres testcontainer), mirroring the new `testCatalogExcludesDroppedColumns` test, so the fix is verified against a real PostgreSQL instance on every CI run.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/opengauss/OpenGaussCatalogTest.java:36  @Disabled("Please Test it in your local environment")`; `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/JdbcOpenGaussIT.java:172  public void testCatalogExcludesDroppedColumns() {`
- Suggested fix: This is a good regression test, but it only covers `OpenGaussCatalog`; the two Postgres-side tests are `@Disabled` and never run in CI, so the actual PostgreSQL fix (`AND NOT a.attisdropped` in `PostgresCatalog`) has zero automated coverage against a real Postgres server. Please add an equivalent `testCatalogExcludesDroppedColumns` case to `JdbcPostgresIT` in the Postgres e2e suite, which already uses a Postgres testcontainer.
- Severity: Medium

**Issue 5: Developer-machine private IPs (192.168.1.20 / 192.168.1.10) hardcoded into catalog tests, including a regression of the existing PostgresCatalogTest baseline from 127.0.0.1**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/psql/PostgresCatalogTest.java:49`
- Why it matters: - Root cause: `PostgresCatalogTest.before()` was changed from `127.0.0.1` to `192.168.1.20`, and the new `OpenGaussCatalogTest` hardcodes `192.168.1.10:8000`. These are the author's LAN addresses leaking into the shared codebase.
- Impact: the Postgres change is a regression for every contributor — these `@Disabled` tests are documented as "run in your local environment", and the existing convention (`127.0.0.1`) worked out-of-the-box against a local container; now `before()` (which also feeds the pre-existing `testCatalog` in the same class) points at an unreachable private host. It also leaks internal environment topology into the repo, which is poor hygiene even in test code.
- Fix: revert the Postgres URL to `jdbc:postgresql://127.0.0.1:5432/postgres` and use `127.0.0.1` (or `localhost`) in `OpenGaussCatalogTest` as well. Longer term, prefer covering this via the Testcontainers-based e2e test you already added in `JdbcOpenGaussIT`, which needs no manual environment.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/opengauss/OpenGaussCatalogTest.java:49  JdbcUrlUtil.getUrlInfo("jdbc:opengauss://192.168.1.10:8000/postgres"),`
- Suggested fix: Please revert this to `jdbc:postgresql://127.0.0.1:5432/postgres`. `192.168.1.20` is your local LAN address — since `before()` also serves the pre-existing `testCatalog` in this class, this regresses the shared baseline that other contributors run against a local Postgres. Same applies to `192.168.1.10` in the new `OpenGaussCatalogTest`.
- Severity: Low

**Issue 6: OpenGaussCatalogTest duplicates PostgresCatalogTest logic and lacks teardown**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/opengauss/OpenGaussCatalogTest.java:56`
- Why it matters: - Root cause: `OpenGaussCatalogTest#testGetTableSchemaAfterColsDropped` is a near line-for-line copy of the same method added to `PostgresCatalogTest` (same DDL, same drop-column statements, same assertions), and the class opens the catalog in `@BeforeAll` but never closes it or drops the created table afterwards.
- Impact: duplicated test logic doubles maintenance cost when the assertion or DDL changes, and the missing `@AfterAll` teardown leaves the `opengauss_cols_drop_test` table and an open catalog connection behind when the test is run locally.
- Fix: extract the shared scenario into a helper (or rely on the e2e IT for OpenGauss, since `OpenGaussCatalog` inherits `PostgresCatalog`'s query), and add an `@AfterAll` that drops the test table and calls `catalog.close()`.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/opengauss/OpenGaussCatalogTest.java:80  String dropColsSql = "alter table " + tableName + " drop column c3," + " drop column c4";`
- Suggested fix: This method is essentially a copy of `PostgresCatalogTest#testGetTableSchemaAfterColsDropped` (same DDL, drops, and assertions). Since `OpenGaussCatalog` inherits the fixed query from `PostgresCatalog` and the OpenGauss path is already covered by the new e2e IT, consider either removing this duplicate or extracting the shared scenario into a helper. Also, please add an `@AfterAll` that drops `opengauss_cols_drop_test` and calls `catalog.close()` so local runs clean up after themselves.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Unrelated `seatunnel-hadoop3-3.1.4-uber` dependency added to connector-jdbc main pom — has nothing to do with the dropped-column fix and pollutes the module's build/classpath

- Issue 2: The `attisdropped` filter is only added to the column-listing SQL — other PG-family catalog queries and Postgres-derived catalogs need the same audit to avoid dropped-column placeholders leaking through other paths
- Issue 3: Module-wide surefire exclusion of net.postgis:postgis-jdbc changes the test classpath for ALL connector-jdbc unit tests, risking breakage of existing PostGIS-related tests
- Issue 4: The actual PostgreSQL fix has no CI coverage — both new catalog tests are `@Disabled`, and the only enabled regression test runs against OpenGauss
- Issue 5: Developer-machine private IPs (192.168.1.20 / 192.168.1.10) hardcoded into catalog tests, including a regression of the existing PostgresCatalogTest baseline from 127.0.0.1
- Issue 6: OpenGaussCatalogTest duplicates PostgresCatalogTest logic and lacks teardown

DanielLeens (commented):
Thanks @SEZ9 for the additional review. I noticed your `CHANGES_REQUESTED` review landed after my earlier approval, so the current merge gate is the new requested-change set, not my older approval alone.

No new commit has landed after that review yet. Please address those items first; I will re-review the updated head from scratch after the next push.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `8948ad5fa908` from the current PostgreSQL / OpenGauss catalog diff.

# What this PR fixes
- User pain: PostgreSQL-compatible catalogs keep placeholder rows in `pg_attribute` after `ALTER TABLE ... DROP COLUMN`. If SeaTunnel reads those placeholders as normal fields, catalog schema discovery returns columns that no longer exist.
- Fix approach: add `AND NOT a.attisdropped` to `PostgresCatalog.SELECT_COLUMNS_SQL_TEMPLATE`, and add behavior-level PostgreSQL and OpenGauss e2e coverage that creates a table, drops a column, calls catalog discovery, and verifies the remaining visible columns.
- One-line summary: the runtime SQL fix is correct, and the latest head now has executable behavior-level coverage for both PostgreSQL and OpenGauss.

# Runtime chain I rechecked
```text
JDBC catalog schema discovery
  -> AbstractJdbcCatalog.getTable(TablePath)
      -> PostgresCatalog.getSelectColumnsSql(tablePath)
          -> SELECT pg_class / pg_namespace / pg_attribute / pg_type
          -> a.attnum > 0
          -> NOT a.attisdropped
      -> PostgresCatalog.buildColumn(resultSet)
      -> TableSchema / CatalogTable returned to JdbcSource / JdbcSink metadata flow

Regression coverage
  -> JdbcPostgresIT.testCatalogExcludesDroppedColumns()
      -> create table (c1, c2, c3, c4)
      -> ALTER TABLE ... DROP COLUMN c2
      -> postgresCatalog.getTable(tablePath)
      -> assert columns == [c1, c3, c4]
  -> JdbcOpenGaussIT.testCatalogExcludesDroppedColumns()
      -> same visible-column contract through OpenGaussCatalog, which inherits PostgresCatalog
```

# Findings
The source-level blocker from Daniel's previous rounds is closed for me now.

I also rechecked the later concerns from @SEZ9's review on the previous head:
- The latest diff no longer includes the unrelated `connector-jdbc` POM dependency / surefire classpath changes.
- The latest diff no longer changes disabled local catalog tests or hardcodes private LAN database endpoints.
- The actual PostgreSQL path now has enabled e2e coverage in `JdbcPostgresIT`, not only OpenGauss coverage.

Test stability rating: stable. The added tests use the existing container-backed e2e infrastructure, avoid sleeps and timing assumptions, clean up the test table in `finally`, and assert the observable catalog behavior rather than only checking SQL text.

I did not run local tests in this round. This review is based on:
- `git diff --stat upstream/dev...8948ad5fa908`: 3 files changed, 103 insertions.
- `git diff upstream/dev...8948ad5fa908` for the production SQL and both e2e tests.
- `git show 8948ad5fa908:<path>` for `PostgresCatalog`.
- GitHub check metadata for the current `Build` check.

# Current CI / merge gate
The aggregate `Build` check is currently failing: https://github.com/apache/seatunnel/runs/86650295959

The check summary points to a fork workflow job, but that fork workflow/job returns `404` from Daniel's account, so I cannot confirm the failing lane details from here. This branch is also currently diverged from `dev` (`ahead_by=8`, `behind_by=8`). Since I do not see a source blocker in the current diff, the lowest-cost next step is to sync with the latest `dev` and rerun Build. If it still fails after the sync, please paste the fresh failing job link here and I can help narrow it down against the updated head.

# Merge conclusion

Blocking items:
- None from Daniel's source review on the latest head.
- Remaining gate: `Build` is red and must pass before approval / merge.

Suggested follow-ups:
- None required from the source review.

Overall, the current implementation is small, correct, and now covered by real PostgreSQL/OpenGauss catalog behavior tests. I am not approving yet only because the required Build gate is still red.

DanielLeens:
Rechecked the same unchanged head `1f5afa9d263b` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
JDBC catalog schema discovery
  -> AbstractJdbcCatalog.getTable(TablePath)
      -> PostgresCatalog.getSelectColumnsSql(tablePath)
          -> SELECT pg_class / pg_namespace / pg_attribute / pg_type
          -> a.attnum > 0
          -> NOT a.attisdropped
      -> PostgresCatalog.buildColumn(resultSet)
      -> TableSchema / CatalogTable returned to JdbcSource / JdbcSink metadata flow

Regression coverage
  -> JdbcPostgresIT.testCatalogExcludesDroppedColumns()
      -> create table (c1, c2, c3, c4)
      -> ALTER TABLE ... DROP COLUMN c2
      -> postgresCatalog.getTable(tablePath)
      -> assert columns == [c1, c3, c4]
  -> JdbcOpenGaussIT.testCatalogExcludesDroppedColumns()
      -> same visible-column contract through OpenGaussCatalog, which inherits PostgresCatalog
```

## Merge conclusion
### Conclusion: source blocker cleared from Daniel's side; waiting for the current gate

1. Blocking items

2. Current gate
- The current remaining gate is still pending: `PENDING gate: Build` (https://github.com/apache/seatunnel/runs/87058542669)
- Daniel only has `READ` permission on this repository, so a write-capable maintainer may still need to handle the final approval / dismissal step once the gate is clear.

NganWave:
Thank you for reviewing my PR and guiding me through the procedure. This PR has been approved and all CI checks have passed. Could you please help merge it? I’d prefer not to rebase again to avoid triggering an unnecessary CI run. Thanks!

DanielLeens (approved):
Rechecked the latest head from the PostgreSQL/OpenGauss catalog query path.

The fix remains focused and correct on the current revision: the catalog SQL now excludes dropped attributes instead of surfacing them back into the discovered schema, and the regression coverage hits both PostgreSQL and OpenGauss.

# Runtime path rechecked
```text
catalog schema discovery
  -> PostgresCatalog.getTable(...)
      -> query pg_attribute metadata
      -> filter out dropped columns with NOT a.attisdropped
  -> downstream JDBC catalog schema mapping
      -> only live columns remain visible to SeaTunnel
```

# Conclusion
### Conclusion: can merge

1. Blocking items

2. Notes
- The current `Build` signal is green, so I do not have a remaining Daniel-side code or CI blocker to add here.

This looks ready from Daniel's side.

davidzollo (approved):
+1
LGTM

PR #11359, opened 2026-07-05
[Improve][Connector-V2] Add row-level resume for Bigtable source chec…

programmerloverun:
…kpoint

Persist lastReadRowKey in split state so failover resumes inside the split instead of re-scanning from split start. Semantics remain at-least-once.

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceReader.java
+++ b/seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceReader.java
@@ -137,6 +137,8 @@ public class BigtableSourceReader implements SourceReader<SeaTunnelRow, Bigtable
                             SeaTunnelRow seaTunnelRow = convertRow(bigtableRow);
                             synchronized (output.getCheckpointLock()) {
                                 output.collect(seaTunnelRow);
+                                // 与 collect 同锁更新进度，保证 triggerBarrier 快照时能看到已 emit 的最后 rowkey
+                                split.setLastReadRowKey(bigtableRow.getKey().toStringUtf8());
                             }
                         });
     }
@@ -144,7 +146,8 @@ public class BigtableSourceReader implements SourceReader<SeaTunnelRow, Bigtable
     private Query buildQuery(BigtableSourceSplit split) {
         Query query = Query.create(parameters.getTable());
 
-        String startKey = split.getStartRowKey();
+        // 恢复场景使用 lastReadRowKey 续读，避免对整个 split 从头重扫
+        String startKey = split.getResumeStartRowKey();
         String endKey = split.getEndRowKey();
         if (!startKey.isEmpty() && !endKey.isEmpty()) {
             query.range(startKey, endKey);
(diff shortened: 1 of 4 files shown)


DanielLeens (changes requested):
Thanks for working on this. The idea of resuming inside a Bigtable split makes sense, but I found one blocker in checkpoint compatibility.

1. `BigtableSourceSplit` changes the serialized identity of checkpointed split state.
   - `seatunnel-connectors-v2/connector-google-bigtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/bigtable/source/BigtableSourceSplit.java:32` changes `serialVersionUID` from `1L` to `2L`.
   - This split is part of checkpoint/savepoint state. Bumping the `serialVersionUID` means existing checkpoints written by the current code will no longer deserialize on the upgraded code path.
   - The new `legacySplitWithoutProgressDeserializesWithNullLastReadRowKey` test does not actually prove backward compatibility, because it serializes and deserializes the current class on the same head. It never loads bytes produced by the old `serialVersionUID = 1L` class.

Please keep the serialized form backward-compatible here and rely on `lastReadRowKey == null` as the old-state default, or add a true legacy-bytes compatibility test if you intentionally need a format migration.

Merge conclusion
- Conclusion: fix before merge.
- Blocking items: preserve checkpoint/savepoint compatibility for `BigtableSourceSplit`.
- Suggestions: after that compatibility point is fixed, the in-split resume idea itself looks reasonable.

programmerloverun:
Thanks @DanielLeens for the review.

- Reverted `serialVersionUID` to `1L` so existing checkpoint/savepoint bytes stay deserializable.
- Replaced the self-roundtrip "legacy" test with `deserializesLegacyCheckpointBytesFromDevRelease()`, which deserializes bytes produced by the pre-change `BigtableSourceSplit` class from dev.
- Converted new comments to English for CI.

Please take another look when CI finishes.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head `b4ddf49043` from the Bigtable split-serialization path down to the checkpoint/recovery query path again.

## What this PR solves
- User pain: if a Bigtable source task fails in the middle of a split, the old checkpoint state only knew “this split is still in flight”, so recovery restarted the whole split from its original start key.
- Fix approach: persist the last emitted row key inside `BigtableSourceSplit`, update it under the same checkpoint lock as `collect()`, and use that row key as the next scan start after restore.
- One-line summary: this turns the failover path from “re-read the whole split from the beginning” into “resume from the last emitted row within the split”, while keeping old checkpoint bytes readable.

## Runtime chain I rechecked
```text
BigtableSourceReader.pollNext() [BigtableSourceReader.java:110-125]
  -> marks currentSplit as in-flight
  -> readSplit(split, output)

BigtableSourceReader.readSplit() [128-145]
  -> readRows(query).forEach(...)
  -> synchronized(checkpointLock)
      -> output.collect(row)
      -> split.setLastReadRowKey(rowKey)

BigtableSourceReader.buildQuery() [147-170]
  -> split.getResumeStartRowKey()
      -> use lastReadRowKey when present [BigtableSourceSplit.java:89-99]
      -> otherwise fall back to startRowKey
```

## Findings
- The previous Daniel blocker around checkpoint compatibility is closed for me now:
  - `BigtableSourceSplit` keeps `serialVersionUID = 1L` again [BigtableSourceSplit.java:33-34]
  - `BigtableSourceSplitTest.deserializesLegacyCheckpointBytesFromDevRelease()` proves old dev-release bytes still deserialize on the new class [BigtableSourceSplitTest.java:71-80]
- The in-flight progress is also captured on the real emit path:
  - `BigtableSourceReader.readSplit()` updates `lastReadRowKey` under the same lock as `output.collect(...)` [BigtableSourceReader.java:136-143]
  - `BigtableSourceReaderTest` covers both “snapshot during scan captures progress” and “restored split resumes from that key” [BigtableSourceReaderTest.java:146-209]

## Tests / stability
- Stability rating: Stable.
- The added tests are pure reader/split unit tests and serialization round-trip checks:
  - no sleeps
  - no ports
  - no external services

## CI / sync snapshot
- The latest CI checks are green on the current head.
- This branch is still `behind_by=1` against `dev`, so I am recording that merge-state fact in the queue, but I am not treating it as a source-side blocker here.

## Merge conclusion
### Conclusion: can merge

1. Blocking items
- None from the current source path.

- None from this round.

Overall, this is now a clean and focused fix for the Bigtable failover path. The checkpoint compatibility concern from the previous round is addressed, and the resume logic is covered by the right regression tests.

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

PR #11366, opened 2026-07-06
[Fix][Core] Fix Spark cluster config lookup

### LINKED ISSUE #9890: [Bug] [spark-3-starter] Failed to read config file while starting with cluster mode
### Search before asking

### What happened

When I start a SeaTunnel job using `start-seatunnel-spark-3-connector-v2.sh` **in cluster mode**, SPARK **FAILED** at the driver stage. The error message indicates that reading the configuration file failed because the file does not exist.

However, if I start the job in **client mode**, it runs correctly.

### Additional Investigation

I have reviewed the source code of the SeaTunnel starter and noticed the following:

* When starting a job in **cluster mode**, the starter adds the `--files` option in `spark-submit` to distribute the configuration file.
* Since the configuration file is loaded **before** the `SparkSession` is created, the code directly tries to read it by file name (assuming it exists in the **current working directory**), instead of resolving it through `SparkFiles`.
* At runtime, I found that the working directory of these code was `$SPARK_HOME/work/driver-xxxxx-xxx/`, while Spark distributed the file from `--files` into`/tmp/spark-xxxxx/`

This mismatch might be the root cause of the problem? I guess?

How can I resolve this issue? Thank you!

### SeaTunnel Version

2.3.12

### SeaTunnel Config

```conf
env {
  parallelism = 1
  job.mode = "BATCH"
}

source {
  FakeSource {
    plugin_output = "fake"
    row.num = 16
    schema = {
      fields {
        name = "string"
        age = "int"
      }
    }
  }
}

transform {
  FieldMapper {
    plugin_input = "fake"
    plugin_output = "fake1"
    field_mapper = {
      age = age
      name = new_name
    }
  }
}

sink {
  Console {
    plugin_input = "fake1"
  }
}
```

### Running Command

```shell
$SEATUNNEL_HOME/bin/start-seatunnel-spark-3-connector-v2.sh --master spark://**.**.**.**:7077 --deploy-mode cluster --config "$SEATUNNEL_HOME/config/v2.batch.config"
```

### Error Exception

```log
25/09/23 11:26:13 INFO SecurityManager: Changing view acls to: dwisparkx
25/09/23 11:26:13 INFO SecurityManager: Changing modify acls to: dwisparkx
25/09/23 11:26:13 INFO SecurityManager: Changing view acls groups to: 
25/09/23 11:26:13 INFO SecurityManager: Changing modify acls groups to: 
25/09/23 11:26:13 INFO SecurityManager: SecurityManager: authentication disabled; ui acls disabled; users  with view permissions: Set(dwisparkx); groups with view permissions: Set(); users  with modify permissions: Set(dwisparkx); groups with modify permissions: Set()
25/09/23 11:26:13 INFO Utils: Successfully started service 'Driver' on port 22106.
25/09/23 11:26:13 INFO DriverWrapper: Driver address: **.**.**.**:22106
25/09/23 11:26:13 INFO WorkerWatcher: Connecting to worker spark://Worker@**.**.**.**:7078
25/09/23 11:26:13 INFO TransportClientFactory: Successfully created connection to /**.**.**.**:7078 after 59 ms (0 ms spent in bootstraps)
25/09/23 11:26:13 INFO WorkerWatcher: Successfully connected to spark://Worker@**.**.**.**:7078
25/09/23 11:26:14 ERROR SeaTunnel: 

===============================================================================


25/09/23 11:26:14 ERROR SeaTunnel: Fatal Error, 

25/09/23 11:26:14 ERROR SeaTunnel: Please submit bug report in https://github.com/apache/seatunnel/issues

25/09/23 11:26:14 ERROR SeaTunnel: Reason:ErrorCode:[COMMON-22], ErrorDescription:[SeaTunnel read file 'v2.batch.config' failed, because it not existed.] 

25/09/23 11:26:14 ERROR SeaTunnel: Exception StackTrace:org.apache.seatunnel.common.exception.SeaTunnelRuntimeException: ErrorCode:[COMMON-22], ErrorDescription:[SeaTunnel read file 'v2.batch.config' failed, because it not existed.]
	at org.apache.seatunnel.common.exception.CommonError.fileNotExistFailed(CommonError.java:89)
	at org.apache.seatunnel.core.starter.utils.FileUtils.checkConfigExist(FileUtils.java:66)
	at org.apache.seatunnel.core.starter.spark.command.SparkTaskExecuteCommand.execute(SparkTaskExecuteCommand.java:37)
	at org.apache.seatunnel.core.starter.SeaTunnel.run(SeaTunnel.java:40)
	at org.apache.seatunnel.core.starter.spark.SeaTunnelSpark.main(SeaTunnelSpark.java:35)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.spark.deploy.worker.DriverWrapper$.main(DriverWrapper.scala:63)
	at org.apache.spark.deploy.worker.DriverWrapper.main(DriverWrapper.scala)
 
25/09/23 11:26:14 ERROR SeaTunnel: 
===============================================================================



25/09/23 11:26:14 INFO ShutdownHookManager: Shutdown hook called
25/09/23 11:26:14 INFO ShutdownHookManager: Deleting directory /tmp/spark-efa59809-250b-4c92-a7f6-a550b4500ed6
```

Spark 3.3.2
Deployed in StandAlone mode

### Java or Scala Version

openjdk version "1.8.0_462"

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
### Purpose

Fixes #9890.

Spark cluster mode ships the SeaTunnel job config through `spark-submit --files`. The runtime previously reduced the submitted config path to only its file name and checked the driver working directory directly. In Spark Standalone cluster mode the shipped file can live under Spark's files directory instead, so SeaTunnel failed before reading the job config.

### Changes

- Resolve the cluster-mode config file from `SparkFiles` when the local basename does not exist.
- Keep the existing path behavior for client mode and for cluster mode when the basename is already available locally.
- Add unit coverage for both behaviors.

### Tests

- `./mvnw -pl seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common -Dtest=SparkTaskExecuteCommandTest test`
- `./mvnw spotless:apply`


DanielLeens (changes requested):
Thanks for working on this. I went through the full diff and the adjacent Spark command path, not just the new helper method.

What this PR solves
- User pain: in Spark cluster mode, the executor can receive only the shipped config basename while the CLI still points at an absolute client-side path, so the job fails to find its config.
- Fix approach: `SparkTaskExecuteCommand` now falls back to `SparkFiles.get(...)` when cluster mode is using a shipped config file.
- One-line summary: this patch fixes the main Spark execute path, but it does not yet fix the matching config-lookup path everywhere.

Runtime path I checked
- Spark submit args
  -> `SparkCommandArgs.buildCommand()`
  -> `SparkTaskExecuteCommand.execute()`
  -> config path resolution
  -> `ConfigBuilder.of(...)`
  -> `SparkExecution.execute()`

Blocking item
1. `SparkConfValidateCommand` still uses the old cluster lookup path, so the fix is incomplete for the sibling Spark config-validation flow.
- Location: `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:53`, `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkConfValidateCommand.java:41`
- Why this is real: `SparkTaskExecuteCommand` now resolves shipped cluster configs through `SparkFiles`, but `SparkConfValidateCommand.execute()` still calls `FileUtils.getConfigPath(sparkCommandArgs)` directly. That leaves the parallel Spark command path on the old basename-only behavior in cluster mode, so the same config lookup problem can still surface there.
- Risk: users get inconsistent behavior across Spark command entry points: execute works after this patch, while the validation flow can still fail on the same shipped config layout.
- Best fix: reuse the same `resolveConfigPath(...)` logic from the validation command as well, so the Spark-side config lookup rule stays consistent.

CI / upstream note
- The current `Build` is red, but the visible failing lanes are broad unrelated jobs such as `engine-v2-it`, multiple `jdbc-connectors-it` partitions, `transform-v2-it-part-2`, `amazonSqs-connector-it`, `doris-connector-it`, and `oracle-cdc-connector-it`, while this PR only touches Spark starter config lookup. The branch is also `behind_by=4` / `compare status=diverged`.
- I would still fix the source-side blocker above first. After that, syncing the latest `dev` and rerunning CI is the lowest-cost next step for the unrelated red lanes.

Merge conclusion: fix before merge

SEZ9 (changes requested):
---
### 1. [exception-handling] (HIGH) SparkFiles.get can throw NPE when SparkEnv is not initialized; resolver call has no exception fallback
Root cause: `resolveConfigPath` invokes `sparkFilesResolver.apply(...)` (bound to `SparkFiles::get`) whenever deploy mode is CLUSTER and the basename is missing locally. `SparkFiles.get()` internally calls `SparkEnv.get().driverTmpDir...`, and `SparkEnv.get()` returns null until a SparkContext exists. Both call sites run BEFORE any SparkSession/SparkContext is created: `SparkTaskExecuteCommand.execute()` resolves the path before `new SparkExecution(config)`, and `SparkConfValidateCommand.execute()` (the `--check` path) never creates a Spark context at all. Impact: a user running `--check` with `--deploy-mode cluster`, or a cluster driver where SparkEnv is not yet initialized at that point, gets a raw `NullPointerException` instead of the intended fallback to the original path or a clear 'config not found' message — a regression from the previous behavior which never touched SparkFiles. Fix: wrap the resolver invocation in a try/catch (catching `NullPointerException`/`IllegalStateException`), log a debug message, and fall back to `configFile`; alternatively guard with `SparkEnv.get() != null` before calling `SparkFiles.get`.

Evidence:
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:88  String sparkFilePath = sparkFilesResolver.apply(fileName.toString());`
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkConfValidateCommand.java:42-43  Path configPath = SparkTaskExecuteCommand.resolveConfigPath(sparkCommandArgs, SparkFiles::get);`
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:56  Path configFile = resolveConfigPath(sparkCommandArgs, SparkFiles::get);`

Suggestion @ `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:88`: `SparkFiles.get()` dereferences `SparkEnv.get()`, which is null until a SparkContext exists. Both call sites (`SparkTaskExecuteCommand.execute()` before `new SparkExecution(config)`, and `SparkConfValidateCommand.execute()` in `--check` mode, which never starts Spark) run before that point, so this can surface as a raw `NullPointerException` for cluster-mode jobs whose basename is missing locally. Could we wrap this call in a try/catch (or guard on `SparkEnv.get() != null`) and fall back to `configFile` with a debug log, so users still get the meaningful `checkConfigExist` error instead of an NPE?

### 2. [Untrusted search path / config substitution] (MEDIUM) Cluster-mode config resolution silently prefers a basename-relative file in the driver working directory (and then a SparkFiles copy) without logging or verifying which file was actually loaded, enabling silent job-config substitution.
Root cause: in cluster mode `FileUtils.getConfigPath` reduces the submitted config path to its basename, and the new `resolveConfigPath` returns that relative path as soon as `Files.exists(configFile)` is true, i.e. any file with the matching basename in the driver's current working directory wins over the file legitimately shipped via `--files`; if the CWD file is absent it falls back to `SparkFiles.get(...)`, again with no logging or integrity check. The SeaTunnel job config is highly sensitive input: it contains connector credentials and fully controls sources/sinks, so an actor with write access to the driver work dir or the Spark files staging dir on a shared/multi-tenant Standalone cluster can substitute a config that exfiltrates data to an attacker-controlled JDBC/HTTP endpoint while the operator believes the submitted config ran. The substitution is invisible because neither branch logs the resolved absolute path. Fix: log the resolved absolute path (e.g. `log.info("Resolved SeaTunnel config file: {}", configFile.toAbsolutePath())`) in both branches of `resolveConfigPath`, prefer the SparkFiles-managed copy over an arbitrary CWD basename in cluster mode (SparkFiles is the authoritative distribution channel), and consider resolving via `toRealPath()` so audit logs reflect the actual file read.

Evidence:
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:78-81  if (!DeployMode.CLUSTER.equals(sparkCommandArgs.getDeployMode()) || Files.exists(configFile)) { return configFile; }`
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:93-96  Path shippedConfigFile = Paths.get(sparkFilePath); if (Files.exists(shippedConfigFile)) { return shippedConfigFile; }`

Suggestion @ `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:78`: In cluster mode this check makes any file in the driver's working directory whose name matches the config basename take precedence over the copy Spark actually shipped via `--files`, and neither branch logs which file was chosen. Since the job config carries connector credentials and fully controls where data flows, a same-basename file planted in the work dir (or a swapped SparkFiles copy) would run silently. Please log the resolved absolute path (`log.info("Resolved SeaTunnel config file: {}", ...)`) in all branches, and consider preferring the SparkFiles-managed copy as the authoritative source in cluster mode.

### 3. [boundary-conditions / test-coverage] (MEDIUM) Tests cover only happy paths; missing coverage for resolver returning null, resolver throwing, and SparkFiles path not existing
Root cause: the new test class exercises exactly two scenarios — the shipped file exists in the resolver directory, and client mode short-circuits — while `resolveConfigPath` has three additional branches: `fileName == null`, `sparkFilePath == null`, and `!Files.exists(shippedConfigFile)`. There is also a fourth untested real-world case: the local basename exists in CLUSTER mode (the `Files.exists(configFile)` short-circuit), which is the documented 'keep existing path behavior' guarantee of this PR. Impact: the fallback branches — precisely the failure-recovery logic — can regress silently, and the untested `SparkFiles::get`-throws scenario (see the NPE finding) is exactly the kind of edge that unit tests with a throwing lambda would have caught. Fix: add tests where the resolver lambda returns null, throws a RuntimeException, and returns a non-existent path, asserting the original `configFile` is returned (or a graceful error is raised) in each case, plus a CLUSTER-mode test where the basename exists in the working directory.

Evidence:
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/test/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommandTest.java:36-49  void resolveClusterConfigFromSparkFilesWhenLocalBasenameMissing() throws IOException { ...`
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:89-91  if (sparkFilePath == null) {
            return configFile;
        }`
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:84-86  if (fileName == null) {
            return configFile;
        }`

Suggestion @ `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/test/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommandTest.java:36`: These two tests only cover the happy paths. Could we add cases for the fallback branches of `resolveConfigPath`: resolver returning null, resolver throwing (simulating `SparkFiles.get` before SparkEnv init), resolver returning a non-existent path, and CLUSTER mode where the basename already exists locally? Those branches are the actual failure-recovery logic this PR introduces and are currently unverified.

### 4. [code-convention/dead-code] (MEDIUM) SparkConfValidateCommand computes the resolved config path but never uses it, leaving the validation command a no-op with an unresolved TODO
The change routes SparkConfValidateCommand.execute() through the new SparkTaskExecuteCommand.resolveConfigPath(), but the resulting `configPath` local variable is still never consumed — the method body ends with the pre-existing `// TODO: validate the config by new api`. The root cause is that the refactor only swapped the path-resolution call without adding any actual validation, so `--check`-style invocations still silently succeed even when the config file is missing in cluster mode, which is exactly the failure scenario #9890 describes. At minimum, please call `org.apache.seatunnel.core.starter.utils.FileUtils.checkConfigExist(configPath)` here (the same helper SparkTaskExecuteCommand already uses at line 57) so the validate command gives users an early, actionable error; otherwise the new cross-class dependency adds coupling with no behavioral benefit. If validation is intentionally deferred, please add a Javadoc note or log statement explaining that, so future readers do not assume the command validates anything.

Evidence:
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkConfValidateCommand.java:42-44  Path configPath = SparkTaskExecuteCommand.resolveConfigPath(sparkCommandArgs, SparkFiles::get);  // TODO: validate the config by new api`

Suggestion @ `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkConfValidateCommand.java:43`: Thanks for wiring the new resolution logic into the validate command as well. However, `configPath` is still unused after this line — the method is effectively a no-op with the `// TODO: validate the config by new api` left in place. Could we at least call `checkConfigExist(configPath)` (as `SparkTaskExecuteCommand.execute()` does) so validation surfaces the missing-config error early in cluster mode? Otherwise this adds a cross-class dependency without changing observable behavior.

### 5. [failure-recovery / diagnosability] (MEDIUM) When neither the local basename nor the SparkFiles copy exists, the error message hides the SparkFiles location that was probed
Root cause: when `Files.exists(shippedConfigFile)` is false, `resolveConfigPath` silently returns the original basename path, and the subsequent `checkConfigExist(configFile)` reports only that basename (e.g. `v2.batch.config`) as missing. Impact: for exactly the failure class this PR targets (Standalone cluster mode with `--files` shipping), a user whose file failed to ship — or shipped under a different name — sees an error mentioning only the working-directory basename, with no hint that `SparkFiles.get(...)` was also checked and what absolute path it resolved to, making the root cause hard to debug on a remote driver. Fix: log (at WARN/INFO) the resolved `sparkFilePath` that was probed and did not exist before falling back, e.g. `log.warn("Config {} not found in working dir nor in SparkFiles path {}", configFile, shippedConfigFile)`, so the driver log pinpoints both candidate locations.

Evidence:
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:93-97  Path shippedConfigFile = Paths.get(sparkFilePath);
        if (Files.exists(shippedConfigFile)) {
            return shippedConfigFile;
        }
        return configFile;`
- `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/command/SparkTaskExecuteCommand.java:57  checkConfigExist(configFile);`

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest Spark config path fix.

# What this PR fixes
The PR makes Spark cluster-mode config lookup prefer the file shipped through SparkFiles, and applies the same resolution path to config validation.

```text
Spark cluster submit / validate path
  -> SparkTaskExecuteCommand.resolveConfigPath(args, SparkFiles::get)
      -> cluster mode first checks shipped SparkFiles basename
      -> client mode keeps the user-provided config path
      -> fallback keeps the existing FileUtils path when SparkFiles is unavailable
  -> SparkConfValidateCommand uses the same resolver before checkConfigExist()
```

# Current review findings
- The previous blocker is resolved: `SparkConfValidateCommand` now uses the same `resolveConfigPath` helper as execution.
- The tests cover shipped SparkFiles, client mode, null/exception fallback, missing shipped file fallback, and preference for the shipped file when a local basename also exists.
- Compatibility: the fallback preserves the existing config path behavior when SparkFiles is unavailable.
- Test stability: the new unit tests use `@TempDir` and deterministic path assertions; I do not see a flaky-test pattern in this static review.
- CI: `Build` is currently `QUEUED`, so I cannot approve yet.
- Base sync fact recorded: compare status `diverged`, `behind_by=1`; no sync request unless the queued Build later fails from upstream drift.

# Merge conclusion
Blocking items: no source-level blocker from Daniel's current review.
Remaining gate: Build must complete successfully before approval.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `58793b7f93` against the Spark submit and validation paths again.

+1 to the earlier technical concerns from @SEZ9 — I retraced the same resolver path on the latest head, and the current revision closes the source-side gaps from that thread.

Current runtime path rechecked:
```text
Spark cluster submit / validate path
  -> SparkTaskExecuteCommand.resolveConfigPath(args, SparkFiles::get)
     -> cluster mode prefers the shipped SparkFiles path
     -> resolver exceptions fall back cleanly to the original config path
     -> resolved path is logged before use
  -> SparkConfValidateCommand uses the same resolver before checkConfigExist()
```

What is now covered on the latest head:
- `SparkConfValidateCommand` no longer stays on the old lookup path
- the resolver fallback covers null / exception / missing-shipped-file cases
- the new tests exercise shipped SparkFiles lookup, client mode short-circuit, local-basename fallback, null resolver, and throwing resolver

So from Daniel's source-review side, I do not see a reopened blocker on the latest revision.

Current live merge facts while posting this follow-up:
- base branch: `dev`
- compare status: `ahead`
- ahead_by: `7`
- behind_by: `0`
- current visible gate: the latest check is still queued on this head

### Merge conclusion: source blocker cleared, waiting for CI
Blocking items: no code-side blocker from Daniel's current review.
Remaining gate: let the current queued check finish on this head.

If that refreshed run turns red, please share the new failing check link and I can narrow it down from the current revision.

DanielLeens (approved):
Rechecked the current head after the queued CI gate finished.

My previous Daniel review had already cleared the Spark config lookup source path and was only waiting for the queued check to complete. The current `Build` check is now green on the same source revision:
https://github.com/apache/seatunnel/runs/85840019913

The source-level conclusion is unchanged: the shipped SparkFiles resolver path, validation path, fallback handling, and tests still look good to me. The branch is currently `compare=diverged`, `behind_by=11`; I am recording that as merge metadata, but I am not treating it as a Daniel-side blocker because this head has a direct green Build result.

1. Blocking items
- None from Daniel's current source review.

2. Suggested follow-up
- None from this round.

Approving this revision from Daniel's side.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=7`, `behind_by=19`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

DanielLeens:
Rechecked the same unchanged head `58793b7f93c3` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
Spark cluster submit / validate path
  -> SparkTaskExecuteCommand.resolveConfigPath(args, SparkFiles::get)
     -> cluster mode prefers the shipped SparkFiles path
     -> resolver exceptions fall back cleanly to the original config path
     -> resolved path is logged before use
  -> SparkConfValidateCommand uses the same resolver before checkConfigExist()
```

## Merge conclusion

1. Blocking items

2. Current gate
- The current CI signal is now green: https://github.com/apache/seatunnel/actions/runs/28934078890/job/85839994178

davidzollo (approved):
LGTM

PR #11371, opened 2026-07-07
[Docs][Connector-V2] Improve Google Bigtable connector docs

DanielCarter-stack:
## Summary
- Expand Google Bigtable source docs with row-key mapping, bounded scan behavior, timestamp/version filters, and additional examples.
- Expand Google Bigtable sink docs with save-mode limits, multi-table sink option, data type storage formats, column family behavior, and additional examples.
- Align Chinese docs with the English option coverage and translate feature labels consistently.

## Verification
- ./mvnw spotless:apply

## Notes
- Checked recent PRs from the last 7 days and found no GoogleBigtable documentation PR or direct GoogleBigtable.md change.
- No Google Bigtable e2e job config exists in the current tree, so the documentation was verified against connector option rules and existing unit tests.

### DIFF
--- a/docs/en/connectors/sink/GoogleBigtable.md
+++ b/docs/en/connectors/sink/GoogleBigtable.md
@@ -12,6 +12,7 @@ Writes data to Google Cloud Bigtable using the native Bigtable Data v2 Java clie
 
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
 - [x] [batch](../../introduction/concepts/connector-v2-features.md)
+- [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)
 
 ## Options
 
@@ -27,6 +28,9 @@ Writes data to Google Cloud Bigtable using the native Bigtable Data v2 Java clie
 | version_column      | string  | no       | -             |
 | null_mode           | string  | no       | skip          |
 | batch_mutation_size | int     | no       | 100           |
+| schema_save_mode    | enum    | no       | RECREATE_SCHEMA |
+| data_save_mode      | enum    | no       | APPEND_DATA   |
+| multi_table_sink_replica | int | no       | -             |
 | common-options      |         | no       | -             |
 
 ### project_id [string]
@@ -91,6 +95,22 @@ How to handle `null` field values. Supported: `skip` (default), `empty`.
 
 Number of row mutations to accumulate before sending a BulkMutation to Bigtable. Default is `100`. Increase for higher throughput at the cost of higher per-task memory usage.
 
+### schema_save_mode [enum]
+
+Schema save mode. Only `RECREATE_SCHEMA` is supported now.
+
+The connector does not create Bigtable tables or column families. Create the target table and all column families before the job starts.
+
+### data_save_mode [enum]
+
+Data save mode. Only `APPEND_DATA` is supported now.
+
+`DROP_DATA` and `ERROR_WHEN_DATA_EXISTS` are not implemented for this connector. If you need a clean target, truncate or recreate the Bigtable table before running the job.
+
+### multi_table_sink_replica [int]
+
+The number of sink replicas used for multi-table writing. For details, see [Sink Common Options](../common-options/sink-common-options.md).
+
 ### common options
 
 Sink plugin common parameters, please refer to [Sink Common Options](../common-options/sink-common-options.md) for details.
@@ -114,6 +134,12 @@ All SeaTunnel types are supported:
 | DATE                         | UTF-8 `yyyy-MM-dd`              |
 | TIME                         | UTF-8 `HH:mm:ss`                |
 | TIMESTAMP                    | UTF-8 `yyyy-MM-dd HH:mm:ss`     |
+
+:::tip
+
+Bigtable does not have relational columns. The sink writes every non-row-key field as a Bigtable cell. The target column family is selected by `column_family`; the Bigtable qualifier is the SeaTunnel field name.
+
+:::
 
 ## Example
 
@@ -166,6 +192,26 @@ sink {
       email       = "identity"
       age         = "stats"
       last_login  = "stats"
+    }
+  }
+}
+```
+
+### Use a version column and empty null values
+
+```hocon
+sink {
+  GoogleBigtable {
+    project_id       = "my-gcp-project"
+    instance_id      = "my-bigtable-instance"
+    table            = "events"
+    rowkey_column    = ["tenant_id", "event_id"]
+    rowkey_delimiter = "#"
+    version_column   = "event_ts"
+    null_mode        = "empty"
+    column_family {
+      all_columns = "data"
+      event_type  = "meta"
     }
   }
 }
--- a/docs/en/connectors/source/GoogleBigtable.md
+++ b/docs/en/connectors/source/GoogleBigtable.md
@@ -15,6 +15,12 @@ Reads data from Google Cloud Bigtable using the native Bigtable Data v2 Java cli
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
 - [ ] [parallelism](../../introduction/concepts/connector-v2-features.md)
 
+:::tip
+
+The source is bounded. It creates one split for the configured table or row-key range, so increasing job parallelism does not split one Bigtable scan into multiple tablet-range reads.
+
+:::
+
 ## Options
 
 | name             | type   | required | default value |
@@ -50,7 +56,7 @@ Path to the Google Cloud service account JSON key file. If omitted, Application 
 
 ### rowkey_column [list]
 
-Optional list of field names that should receive the row key value. Declare a field named `rowkey` in your schema to capture the raw row key bytes as a `BYTES` or `STRING` field.
+Optional list of field names that should receive the row key value. If this option is not set, the connector uses a schema field named `rowkey` as the row-key field. The row-key field can be declared as `BYTES` or `STRING`.
 
 ### start_rowkey [string]
 
@@ -82,13 +88,19 @@ Source plugin common parameters, please refer to [Source Common Options](../comm
 
 ## Schema Mapping
 
-Field names in the SeaTunnel schema must follow the pattern `familyName:qualifier`, for example `cf:name` or `stats:age`. The special field name `rowkey` maps to the Bigtable row key.
+Field names in the SeaTunnel schema must follow the pattern `familyName:qualifier`, for example `cf:name` or `stats:age`. The row-key field is controlled by `rowkey_column`; if it is not configured, the special field name `rowkey` maps to the Bigtable row key.
 
 | Schema field name | Mapped Bigtable cell        |
 |-------------------|-----------------------------|
 | `rowkey`          | Row key                     |
 | `cf:name`         | Column family `cf`, qualifier `name` |
 | `stats:age`       | Column family `stats`, qualifier `age` |
+
+:::tip
+
+The source reads the latest returned cell for each `family:qualifier` field. Use `start_timestamp`, `end_timestamp`, and `max_versions` to control the Bigtable scan filter. SeaTunnel field types must match the bytes stored in Bigtable. For example, numeric values written by this connector are binary big-endian values, while `STRING`, `DATE`, `TIME`, `TIMESTAMP`, and `DECIMAL` are UTF-8 text.
+
+:::
 
 ## Example
 
@@ -126,6 +138,26 @@ source {
     schema {
       fields {
         rowkey    = STRING
+        "cf:type" = STRING
+        "cf:data" = STRING
+      }
+    }
+  }
+}
+```
+
+### Use a custom row-key field name
+
+```hocon
+source {
+  GoogleBigtable {
+    project_id    = "my-gcp-project"
+    instance_id   = "my-bigtable-instance"
+    table         = "events"
+    rowkey_column = ["event_id"]
+    schema {
+      fields {
+        event_id  = STRING
         "cf:type" = STRING
         "cf:data" = STRING
       }
--- a/docs/zh/connectors/sink/GoogleBigtable.md
+++ b/docs/zh/connectors/sink/GoogleBigtable.md
@@ -11,7 +11,8 @@ import ChangeLog from '../changelog/connector-google-bigtable.md';
 ## 主要特性
 
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
-- [x] [batch](../../introduction/concepts/connector-v2-features.md)
+- [x] [批处理](../../introduction/concepts/connector-v2-features.md)
+- [x] [支持多表写入](../../introduction/concepts/connector-v2-features.md)
 
 ## 参数
 
@@ -27,6 +28,9 @@ import ChangeLog from '../changelog/connector-google-bigtable.md';
 | version_column     | string  | 否     | -    |
 | null_mode          | string  | 否     | skip |
 | batch_mutation_size| int     | 否     | 100  |
+| schema_save_mode   | enum    | 否     | RECREATE_SCHEMA |
+| data_save_mode     | enum    | 否     | APPEND_DATA |
+| multi_table_sink_replica | int | 否    | -    |
 | common-options     |         | 否     | -    |
 
 ### project_id [string]
@@ -82,9 +86,57 @@ Google Cloud 服务账号 JSON 密钥文件路径。未设置时使用应用默�
 
 ### batch_mutation_size [int]
 
-每次批量提交的行数，默认 `100`。
+每次批量提交的行数，默认 `100`。调大该值可以提高吞吐，但会增加每个任务本地缓存的数据量。
+
+### schema_save_mode [enum]
+
+Schema 保存模式。当前只支持 `RECREATE_SCHEMA`。
+
+连接器不会自动创建 Bigtable 表或列族。运行作业前，需要先在 Bigtable 中创建好目标表和所有会用到的列族。
+
+### data_save_mode [enum]
+
+数据保存模式。当前只支持 `APPEND_DATA`。
+
+本连接器暂不支持 `DROP_DATA` 和 `ERROR_WHEN_DATA_EXISTS`。如果需要干净的目标表，请在运行作业前手动清空或重建 Bigtable 表。
+
+### multi_table_sink_replica [int]
+
+多表写入时使用的 Sink 副本数。更多说明请参考 [Sink Common Options](../common-options/sink-common-options.md)。
+
+### common options
+
+Sink 插件通用参数，详见 [Sink Common Options](../common-options/sink-common-options.md)。
+
+## 数据类型
+
+Bigtable 没有关系型数据库那样的列类型。连接器会按如下格式把 SeaTunnel 字段写入 Bigtable Cell：
+
+| SeaTunnel 类型 | Bigtable 中的保存格式 |
+|----------------|----------------------|
+| TINYINT        | 1 字节二进制 |
+| SMALLINT       | 2 字节大端二进制 |
+| INT            | 4 字节大端二进制 |
+| BIGINT         | 8 字节大端二进制 |
+| FLOAT          | 4 字节 IEEE 754 大端二进制 |
+| DOUBLE         | 8 字节 IEEE 754 大端二进制 |
+| BOOLEAN        | 1 字节，`1` 表示 true，`0` 表示 false |
+| BYTES          | 原始字节 |
+| STRING         | UTF-8 文本 |
+| DECIMAL        | UTF-8 普通数字字符串 |
+| DATE           | UTF-8 `yyyy-MM-dd` |
+| TIME           | UTF-8 `HH:mm:ss` |
+| TIMESTAMP      | UTF-8 `yyyy-MM-dd HH:mm:ss` |
+
+:::tip
+
+Sink 会把非行键字段写成 Bigtable Cell。目标列族由 `column_family` 决定，Bigtable 的列限定符使用 SeaTunnel 字段名。
+
+:::
 
 ## 示例
+
+### 使用应用默认凭证写入
 
 ```hocon
 sink {
@@ -95,6 +147,64 @@ sink {
     rowkey_column = ["event_id"]
     column_family {
       all_columns = "cf"
+    }
+  }
+}
+```
+
+### 使用服务账号和复合行键写入
+
+```hocon
+sink {
+  GoogleBigtable {
+    project_id       = "my-gcp-project"
+    instance_id      = "my-bigtable-instance"
+    table            = "events"
+    credentials_path = "/secrets/sa-key.json"
+    rowkey_column    = ["tenant_id", "event_id"]
+    rowkey_delimiter = "#"
+    column_family {
+      all_columns = "data"
+    }
+    batch_mutation_size = 500
+  }
+}
+```
+
+### 写入多个列族
+
+```hocon
+sink {
+  GoogleBigtable {
+    project_id  = "my-gcp-project"
+    instance_id = "my-bigtable-instance"
+    table       = "user_profile"
+    rowkey_column = ["user_id"]
+    column_family {
+      name       = "identity"
+      email      = "identity"
+      age        = "stats"
+      last_login = "stats"
+    }
+  }
+}
+```
+
+### 使用版本列并把空值写成空字节
+
+```hocon
+sink {
+  GoogleBigtable {
+    project_id       = "my-gcp-project"
+    instance_id      = "my-bigtable-instance"
+    table            = "events"
+    rowkey_column    = ["tenant_id", "event_id"]
+    rowkey_delimiter = "#"
+    version_column   = "event_ts"
+    null_mode        = "empty"
+    column_family {
+      all_columns = "data"
+      event_type  = "meta"
     }
   }
 }
--- a/docs/zh/connectors/source/GoogleBigtable.md
+++ b/docs/zh/connectors/source/GoogleBigtable.md
@@ -10,9 +10,16 @@ import ChangeLog from '../changelog/connector-google-bigtable.md';
 
 ## 主要特性
 
-- [x] [batch](../../introduction/concepts/connector-v2-features.md)
-- [ ] [parallelism](../../introduction/concepts/connector-v2-features.md)
+- [x] [批处理](../../introduction/concepts/connector-v2-features.md)
+- [ ] [流处理](../../introduction/concepts/connector-v2-features.md)
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
+- [ ] [并行度](../../introduction/concepts/connector-v2-features.md)
+
+:::tip
+
+该 Source 是有界读取。当前只会为配置的表或行键范围生成一个切分，所以提高作业并行度不会把一次 Bigtable 扫描拆成多个 tablet 范围并发读取。
+
+:::
 
 ## 参数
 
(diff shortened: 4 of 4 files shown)


DanielLeens (approved):
Thanks for the documentation improvement! I checked this against the current GoogleBigtable connector implementation rather than only reading the Markdown diff.

## What this PR improves

- User pain: the existing GoogleBigtable docs did not explain several practical details clearly enough, especially row-key mapping, Bigtable cell type encoding, save-mode limitations, and multi-table sink configuration.
- Fix approach: this PR expands both English and Chinese source/sink docs with option details, examples, and connector-specific notes.
- One-line summary: the docs now describe the current connector contract much more clearly.

## Runtime / contract path I checked

```text
user HOCON config
  -> BigtableSourceFactory.optionRule() / BigtableSinkFactory.optionRule()
      -> validates the documented options
  -> BigtableSource / BigtableSink
      -> builds BigtableParameters
  -> Source reader / Sink writer
      -> reads Bigtable cells or writes mutations
```

I also checked the multi-table sink contract:

- `BigtableSink` implements `SupportMultiTableSink`
- `BigtableSinkWriter` implements `SupportMultiTableSinkWriter`
- `BigtableSinkFactory.optionRule()` includes `SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA`

The new `schema_save_mode` and `data_save_mode` docs also match the connector-side option registration and the current limitation that Bigtable tables / column families must already exist.

- Documentation-only change.
- No API, config default, protocol, serialization, or runtime behavior change.
- No runtime performance impact.

## Merge conclusion

### Conclusion: can merge

1. Blocking items
- None.

2. Suggested follow-up
- None from this round.

Overall, this is a helpful docs update and it lines up with the current GoogleBigtable source/sink implementation. LGTM from Daniel's side.

davidzollo (approved):
Docs-only connector documentation update. CI is green.

SEZ9 (changes requested):
🤖 **Automated review draft** (SeaTunnel Review Agent) — please treat as suggestions; verify against the code before acting.

**Recommendation:** REQUEST_CHANGES   **Risk:** HIGH

- This change touches: docs/en, docs/zh
- Risk signals: security-sensitive keywords: secret, credential; connector logic changes; docs/tests only
- Most important concern: Docs add `schema_save_mode`, `data_save_mode`, `multi_table_sink_replica` and a 'support multiple table write' feature flag with no accompanying code change — these must be verified against the actual GoogleBigtable sink factory/option rules or users will configure options the connector rejects or ignores
- Related historical context: PR#11369, PR#11369, PR#11123

---
### 1. [Docs/Option Compatibility] (HIGH) Docs add `schema_save_mode`, `data_save_mode`, `multi_table_sink_replica` and a 'support multiple table write' feature flag with no accompanying code change — these must be verified against the actual GoogleBigtable sink factory/option rules or users will configure options the connector rejects or ignores
(a) This is a docs-only PR (4 changed files, all under docs/), yet it adds three new options to the sink option table and flips the feature checklist to claim multi-table write support: `docs/en/connectors/sink/GoogleBigtable.md:15` adds `- [x] [support multiple table write]` and lines 31-33 add `schema_save_mode`/`data_save_mode`/`multi_table_sink_replica` to the option table. The PR description itself admits there is no Bigtable e2e config and verification was only 'against connector option rules and existing unit tests'. (b) If `GoogleBigtableSinkFactory.optionRule()` does not declare `SinkConnectorCommonOptions.SCHEMA_SAVE_MODE`/`DATA_SAVE_MODE`, and the sink does not implement `SupportSaveMode`, users copying these options will either hit option-rule validation failures on Zeta/Flink/Spark job submission or silently get no save-mode behavior; similarly, the multi-table-write checkbox is only true if the sink implements `SupportMultiTableSink` (and the writer `SupportMultiTableSinkWriter`) — otherwise multi-table pipelines (e.g. CDC fan-out) will fail at plan time with a misleading doc as the root cause. (c) Before merging, grep `seatunnel-connectors-v2/connector-google-bigtable` for `SupportSaveMode`, `SupportMultiTableSink`, and the factory `optionRule()`; remove any option row that is not actually declared, or land the connector change first and document it in the same PR per SeaTunnel convention that docs mirror the option rule exactly.

Evidence:
- `docs/en/connectors/sink/GoogleBigtable.md:15  + - [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)`
- `docs/en/connectors/sink/GoogleBigtable.md:31-33  + | schema_save_mode    | enum    | no       | RECREATE_SCHEMA | / + | data_save_mode      | enum    | no       | APPEND_DATA   | / + | multi_table_sink_replica | int | no       | -             |`
- `docs/zh/connectors/sink/GoogleBigtable.md:15  + - [x] [支持多表写入](../../introduction/concepts/connector-v2-features.md)`
- `PR description: 'No Google Bigtable e2e job config exists in the current tree, so the documentation was verified against connector option rules and existing unit tests.'`

Suggestion @ `docs/en/connectors/sink/GoogleBigtable.md:31`: These three options (`schema_save_mode`, `data_save_mode`, `multi_table_sink_replica`) and the new 'support multiple table write' feature checkbox are added without any connector code change in this PR. Could you point to where `GoogleBigtableSinkFactory.optionRule()` declares `SCHEMA_SAVE_MODE`/`DATA_SAVE_MODE`, and where the sink implements `SupportSaveMode`/`SupportMultiTableSink`? If the option rule does not declare them, users copying this table will get validation errors (or silently ignored config) on job submission, and multi-table pipelines will fail at plan time. Docs should mirror the actual option rule exactly.

### 2. [documentation-correctness / save-mode semantics] (HIGH) Documented default `schema_save_mode = RECREATE_SCHEMA` directly contradicts the statement that the connector never creates tables, and the option may not even exist in the connector's OptionRule
(a) Root cause: the new docs declare `schema_save_mode` with default `RECREATE_SCHEMA` (a mode whose defined semantic across SeaTunnel is 'drop and recreate the target schema'), yet two lines later state 'The connector does not create Bigtable tables or column families'. If the connector truly performs no DDL, the honest value would be `IGNORE`/`CREATE_SCHEMA_WHEN_NOT_EXIST`-style semantics or the option should not be documented at all. (b) Impact: users on Zeta/Flink/Spark will either (1) expect the target table to be recreated and lose that guarantee silently, leading to `NotFoundException: table not found` failures at write time when the table is missing, or (2) if `schema_save_mode`/`data_save_mode` are not actually declared in `GoogleBigtableSinkFactory`'s `optionRule()`, setting them per these docs will fail config validation or be silently dropped depending on engine strictness. The author admits verification was 'against connector option rules and existing unit tests' with no e2e config, so this is unverified. (c) Fix: check `GoogleBigtableSinkOptions`/`GoogleBigtableSinkFactory` for `SCHEMA_SAVE_MODE`/`DATA_SAVE_MODE` declarations and the actual `SaveModeHandler` behavior; if the connector performs no DDL, document the actually-supported enum value and explicitly state that a missing table causes a runtime failure (with the exception the user will see), rather than pairing `RECREATE_SCHEMA` with 'does not create tables'.

Evidence:
- `docs/en/connectors/sink/GoogleBigtable.md:31  | schema_save_mode    | enum    | no       | RECREATE_SCHEMA |`
- `docs/en/connectors/sink/GoogleBigtable.md:98  Schema save mode. Only `RECREATE_SCHEMA` is supported now.`
- `docs/en/connectors/sink/GoogleBigtable.md:100  The connector does not create Bigtable tables or column families. Create the target table and all column families before the job starts.`
- `docs/zh/connectors/sink/GoogleBigtable.md:91  Schema 保存模式。当前只支持 `RECREATE_SCHEMA`。`

Suggestion @ `docs/en/connectors/sink/GoogleBigtable.md:100`: This section says the only supported mode is `RECREATE_SCHEMA` but then states the connector 'does not create Bigtable tables or column families'. `RECREATE_SCHEMA` semantics in SeaTunnel imply drop-and-recreate, so this is contradictory and will mislead users into expecting table creation, only to hit a table-not-found failure at write time. Could you verify what `GoogleBigtableSinkFactory` actually declares in its `optionRule()` and what the SaveModeHandler does (if one exists)? If the connector performs no DDL, please document the real enum value/behavior and explicitly state the runtime error a user sees when the table or a column family is missing.

### 3. [Docs/Behavior Claim Verification] (MEDIUM) Rewritten `rowkey_column` semantics for the source ('overrides the default `rowkey` field name') and the new custom-row-key example are behavioral claims that must match GoogleBigtableSource's schema-mapping code
(a) `docs/en/connectors/source/GoogleBigtable.md:59` changes the description from 'Declare a field named `rowkey` in your schema' to 'If this option is not set, the connector uses a schema field named `rowkey` as the row-key field', and a new example at lines 149-166 configures `rowkey_column = ["event_id"]` with an `event_id` schema field. This asserts the source honors `rowkey_column` as a rename of the row-key field. Note the option is a `list` — the docs never say what happens if multiple field names are supplied on the source side (on the sink it means composite key with `rowkey_delimiter`, but a source cannot split a raw key without a delimiter option, and the source option table has no `rowkey_delimiter`). (b) If the source actually only recognizes the literal `rowkey` field name (as the old text implied) or ignores `rowkey_column`, users copying this example get a job where `event_id` is treated as a `family:qualifier` field, failing schema validation ('field name must follow familyName:qualifier') or producing null columns. (c) Verify in the source's schema-mapping code (`GoogleBigtableSource`/deserializer) that `rowkey_column` is read and mapped, and explicitly document the multi-element behavior (or state that only a single element is honored on the source side).

Evidence:
- `docs/en/connectors/source/GoogleBigtable.md:59  + Optional list of field names that should receive the row key value. If this option is not set, the connector uses a schema field named `rowkey` as the row-key field.`
- `docs/en/connectors/source/GoogleBigtable.md:149-158  + ### Use a custom row-key field name ... + rowkey_column = ["event_id"] ... + event_id  = STRING`
- `docs/zh/connectors/source/GoogleBigtable.md:56-58  + 用于接收 Bigtable 行键的字段名列表。未设置时，连接器默认把名为 `rowkey` 的字段当作行键字段。`

Suggestion @ `docs/en/connectors/source/GoogleBigtable.md:59`: This rewords `rowkey_column` from 'declare a field named `rowkey`' to a fallback semantic ('if not set, `rowkey` is used'). Please confirm the source code actually resolves `rowkey_column` to rename the row-key field — the new example (`rowkey_column = ["event_id"]`) will fail schema validation if the reader only special-cases the literal `rowkey` name. Also, since this is a list and the source has no `rowkey_delimiter`, please document what happens when multiple field names are provided (composite split is not possible on read).

### 4. [boundary-conditions / data-loss risk] (MEDIUM) `version_column` example is added without documenting the required type or timestamp unit — wrong units silently corrupt Bigtable cell versions and can trigger GC deletion
(a) Root cause: the new sink example sets `version_column = "event_ts"` but neither the example nor the option section states what type `event_ts` must be (BIGINT epoch? TIMESTAMP?) or the unit Bigtable expects (cell timestamps are microseconds since epoch). The same docs say TIMESTAMP fields are stored as UTF-8 `yyyy-MM-dd HH:mm:ss` strings, which cannot directly be a cell version, so the conversion path is ambiguous. (b) Impact: a user supplying epoch milliseconds where the connector expects microseconds (or vice versa) writes cells with versions off by 1000x; with an age-based GC policy on the column family, cells dated in 1970 can be garbage-collected almost immediately — silent data loss — and multi-version reads via the source's `start_timestamp`/`end_timestamp` filters (documented as microseconds in the zh docs at line ~69) will return nothing. Null/negative `event_ts` values are another undocumented edge: does the writer skip the row, throw, or fall back to server time? (c) Fix: read `BigtableSinkWriter`'s version handling and document the accepted field types, the unit, and the null-value behavior for `version_column`, plus a warning about GC policies interacting with client-supplied timestamps.

Evidence:
- `docs/en/connectors/sink/GoogleBigtable.md:210  version_column   = "event_ts"`
- `docs/en/connectors/sink/GoogleBigtable.md:135  | TIMESTAMP                    | UTF-8 `yyyy-MM-dd HH:mm:ss`     |`
- `docs/zh/connectors/source/GoogleBigtable.md:69  Cell 时间戳过滤的起始值，包含该时间戳，单位是微秒。`

Suggestion @ `docs/en/connectors/sink/GoogleBigtable.md:210`: The new example uses `version_column = "event_ts"` but the docs never say what type this field must be or what unit the connector converts it to (Bigtable cell versions are microseconds since epoch). If a user supplies epoch milliseconds, versions are 1000x off and an age-based GC policy can delete the cells immediately — silent data loss. Could you check the writer's conversion logic and document the accepted types, the unit, and what happens when `event_ts` is null/negative?

### 5. [documentation-correctness / multi-table sink] (MEDIUM) New 'support multiple table write' feature flag and `multi_table_sink_replica` option are documented without evidence the connector implements `SupportMultiTableSink`
(a) Root cause: the PR flips the sink feature checklist to `[x] [support multiple table write]` and documents `multi_table_sink_replica`, but the PR description says there is no e2e config and verification was docs-only; there is no shown code implementing `SupportMultiTableSink`/`SupportMultiTableSinkWriter` for GoogleBigtable. (b) Impact: if the sink class does not implement `SupportMultiTableSink`, users who wire this connector behind a multi-table source (e.g., CDC fan-out) per these docs will get a job-submission or runtime failure on Zeta ('sink does not support multi table'), or `multi_table_sink_replica` will be rejected/ignored — a happy-path doc claim with an unhandled failure scenario. Additionally, if multi-table writing were supported, concurrency matters: multiple writer replicas sharing a `BigtableDataClient`/`BulkMutation` batcher must be thread-safe, which the docs say nothing about. (c) Fix: verify `GoogleBigtableSink implements SupportMultiTableSink` (and the writer's thread-safety for shared clients) before merging; if unsupported, revert the checkbox to `[ ]` and remove the `multi_table_sink_replica` row from both en and zh option tables.

Evidence:
- `docs/en/connectors/sink/GoogleBigtable.md:15  - [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)`
- `docs/en/connectors/sink/GoogleBigtable.md:33  | multi_table_sink_replica | int | no       | -             |`
- `docs/zh/connectors/sink/GoogleBigtable.md:15  - [x] [支持多表写入](../../introduction/concepts/connector-v2-features.md)`
- `docs/zh/connectors/sink/GoogleBigtable.md:103  多表写入时使用的 Sink 副本数。更多说明请参考 [Sink Common Options](../common-options/sink-common-options.md)。`

Suggestion @ `docs/en/connectors/sink/GoogleBigtable.md:15`: This PR marks 'support multiple table write' as supported, but the notes say verification was documentation-only. Can you confirm `GoogleBigtableSink` actually implements `SupportMultiTableSink` (and that the writer is safe when multiple replicas run per `multi_table_sink_replica`)? If the interface is not implemented, a user following this doc will hit a job-submission/runtime failure on Zeta when placing this sink behind a multi-table source, so the checkbox and the `multi_table_sink_replica` row should be removed.

### 6. [Docs/Semantics Accuracy] (MEDIUM) `start_timestamp`/`end_timestamp` documented as microseconds in zh docs — a 1000x unit error if the connector actually accepts milliseconds
(a) `docs/zh/connectors/source/GoogleBigtable.md:66-72` newly asserts both timestamp filter options are '单位是微秒' (microseconds). The Bigtable Data API's `TimestampRangeFilter` natively uses microseconds, but many SeaTunnel connectors (e.g. HBase-style sources) accept milliseconds and convert internally — the unit depends on whether `GoogleBigtableSourceReader` multiplies by 1000 before building the filter. The EN options section for these fields is not shown in the diff, so the zh page may now diverge from EN or from the code. (b) If the actual unit is milliseconds, users following this doc will pass epoch-micros values and get an effectively unbounded or empty time filter (values 1000x too large), silently reading wrong data slices — a nasty correctness issue in incremental batch pipelines. (c) Trace the reader code where `FILTERS.timestamp().range()` (or equivalent) is built from these options and confirm whether a millis→micros conversion exists; then state the unit identically in both EN and zh option descriptions.

Evidence:
- `docs/zh/connectors/source/GoogleBigtable.md:66-68  + ### start_timestamp [long] + Cell 时间戳过滤的起始值，包含该时间戳，单位是微秒。`
- `docs/zh/connectors/source/GoogleBigtable.md:70-72  + ### end_timestamp [long] + Cell 时间戳过滤的结束值，不包含该时间戳，单位是微秒。`

Suggestion @ `docs/zh/connectors/source/GoogleBigtable.md:68`: “单位是微秒” is a strong claim — Bigtable's TimestampRangeFilter uses micros, but the connector may accept milliseconds and convert internally (common SeaTunnel pattern). If the code takes millis, this doc causes a 1000x unit error and users' timestamp filters will silently match nothing/everything. Could you cite the reader code that builds the timestamp filter, and mirror the confirmed unit in the EN option description too?

### 7. [docs-accuracy] (MEDIUM) New tip says the source "reads the latest returned cell" per field, which conflicts with documenting `max_versions` as a user-facing control
The added tip states "The source reads the latest returned cell for each `family:qualifier` field" while in the same sentence advising users to tune `max_versions` (documented elsewhere on the page with default `1`). If the deserializer always takes the latest cell per qualifier, setting `max_versions > 1` changes the scan filter but has no observable effect on the emitted row, which is confusing and looks like a knob that silently does nothing; if instead multiple versions do affect output, the tip is wrong. This ambiguity was propagated verbatim into the ZH page as well. Please verify the source's row-to-SeaTunnelRow conversion logic and either (a) state explicitly that `max_versions > 1` only widens the filter but the row still carries one value per field, or (b) document how multiple versions are surfaced, and adjust the `max_versions` option description accordingly on both EN and ZH pages.

Evidence:
- `docs/en/connectors/source/GoogleBigtable.md:101  The source reads the latest returned cell for each `family:qualifier` field. Use `start_timestamp`, `end_timestamp`, and `max_versions` to control the Bigtable scan filter.`
- `docs/zh/connectors/source/GoogleBigtable.md:~101  Source 会读取每个 `列族:列限定符` 字段返回的最新 Cell。可以通过 `start_timestamp`、`end_timestamp` 和 `max_versions` 控制 Bigtable 扫描过滤条件。`
- `docs/zh/connectors/source/GoogleBigtable.md:76  每个列限定符最多返回的 Cell 版本数。默认值 `1` 表示只读取最新版本。`

Suggestion @ `docs/en/connectors/source/GoogleBigtable.md:101`: This tip says the source reads "the latest returned cell" per field, but it also directs users to tune `max_versions`. If the reader always picks one cell per qualifier, `max_versions > 1` would only change the scan filter without affecting the emitted row — could you verify the deserialization behavior and clarify here what setting `max_versions > 1` actually does for the output row? The same clarification should go into the ZH tip and the `max_versions` option description.

### 8. [docs-consistency] (MEDIUM) EN/ZH feature lists diverge: ZH source page adds a `流处理` (streaming) row that does not exist in the EN source page
The ZH source page's feature list was rewritten to four items including a new unchecked `- [ ] [流处理]` row, while the EN source page (also touched in this PR at the same location, lines 15-17) keeps only exactly-once and parallelism plus batch and has no streaming row. SeaTunnel docs convention is that EN and ZH connector pages mirror each other structurally so translations can be diffed and kept in sync; introducing an extra row only on the ZH side guarantees future drift. The impact is user confusion for Chinese readers who may interpret the presence of a `流处理` checkbox as meaning streaming is planned/relevant for this connector while EN readers see nothing. Please either add the same `- [ ] [stream]` row to docs/en/connectors/source/GoogleBigtable.md or drop the extra row from the ZH page so both lists have identical items in identical order.

Evidence:
- `docs/zh/connectors/source/GoogleBigtable.md:13-16  - [x] [批处理](...) / - [ ] [流处理](...) / - [ ] [exactly-once](...) / - [ ] [并行度](...)`
- `docs/en/connectors/source/GoogleBigtable.md:15-16  - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
- [ ] [parallelism](../../introduction/concepts/connector-v2-features.md)`

Suggestion @ `docs/zh/connectors/source/GoogleBigtable.md:14`: This adds a `- [ ] [流处理]` row that has no counterpart in the EN source page (docs/en/connectors/source/GoogleBigtable.md keeps only batch/exactly-once/parallelism). To keep EN and ZH pages structurally in sync per docs convention, could you either add the matching `stream` row to the EN page or remove this row here so both feature lists contain the same items in the same order?

### 9. [Credential/Secret Handling] (LOW) New examples normalize long-lived service-account JSON key files on local disk without any secret-handling guidance
(a) The newly added English and Chinese examples introduce `credentials_path = "/secrets/sa-key.json"` as the canonical way to authenticate, but the surrounding option docs (`### credentials_path`) give no guidance on protecting that key file. (b) Google SA JSON keys are long-lived bearer credentials; on a shared Zeta/Flink/Spark cluster the path must be readable by every task-manager/worker user, so a world-readable key file grants any co-tenant job full Bigtable data-plane access for that service account (read/write/delete of all rows the SA can reach). Job configs are also frequently committed to git or exposed via the engine REST/UI, making the key location trivially discoverable. (c) Add a security note under `credentials_path` in both EN and ZH pages recommending Application Default Credentials / Workload Identity as the preferred mechanism (the docs already support ADC when the option is omitted), and, when a key file is unavoidable, requiring restrictive filesystem permissions (e.g. 0600, owned by the job runtime user), distribution via secret manager/K8s Secret volume rather than baked images, and a least-privilege Bigtable IAM role (e.g. `roles/bigtable.user` scoped to the instance).

Evidence:
- `docs/zh/connectors/sink/GoogleBigtable.md:~163  credentials_path = "/secrets/sa-key.json"`
- `docs/zh/connectors/source/GoogleBigtable.md:~134  credentials_path = "/secrets/sa-key.json"`
- `docs/zh/connectors/source/GoogleBigtable.md:~55  Google Cloud 服务账号 JSON 密钥文件路径。未设置时使用应用默认凭证（ADC）。`
- `docs/en/connectors/source/GoogleBigtable.md:~57  Path to the Google Cloud service account JSON key file. If omitted, Application ...`

Suggestion @ `docs/zh/connectors/sink/GoogleBigtable.md:163`: Since these examples establish `credentials_path = "/secrets/sa-key.json"` as the recommended pattern, could we add a short security note under `### credentials_path` (EN + ZH)? Suggest: prefer Application Default Credentials / Workload Identity where possible; if a key file must be used, mount it via a secret manager or K8s Secret with 0600 permissions owned by the job runtime user, grant the service account least-privilege Bigtable roles, and never commit the key alongside job configs. SA JSON keys are long-lived credentials, and on shared Zeta/Flink/Spark clusters an over-permissive key file is readable by any co-located job.

### 10. [Docs/Engine Semantics] (LOW) New 'single split / bounded' tips and the zh feature-checklist rewrite hard-code enumerator behavior that should be verified against the split enumerator, and will go stale if range splitting is added
(a) Both pages now assert the source 'creates one split for the configured table or row-key range' (docs/en/connectors/source/GoogleBigtable.md:18-22, docs/zh/connectors/source/GoogleBigtable.md:17-21), and the zh feature list is restructured — `parallelism` is replaced by a new unchecked `流处理` line plus `并行度` (docs/zh/connectors/source/GoogleBigtable.md:13-16). (b) The single-split claim is plausible given `- [ ] [parallelism]` was already unchecked, but it is an implementation detail of the SourceSplitEnumerator; if the enumerator actually emits one split per configured range or per tablet, the tip actively misleads users tuning parallelism on Zeta/Flink/Spark (extra readers would sit idle vs. actually scaling). The zh checklist now also includes a `流处理` item the EN page does not have, so EN/zh feature lists are no longer symmetric. (c) Verify against `GoogleBigtableSourceSplitEnumerator` that exactly one split is produced, soften the wording to 'currently' (the zh text already says 当前, EN does not), and mirror the same feature-checklist rows in both languages.

Evidence:
- `docs/en/connectors/source/GoogleBigtable.md:18-22  + The source is bounded. It creates one split for the configured table or row-key range, so increasing job parallelism does not split one Bigtable scan into multiple tablet-range reads.`
- `docs/zh/connectors/source/GoogleBigtable.md:13-16  + - [x] [批处理](...) + - [ ] [流处理](...) ... + - [ ] [并行度](...)`
- `docs/zh/connectors/source/GoogleBigtable.md:17-21  + 该 Source 是有界读取。当前只会为配置的表或行键范围生成一个切分`

Suggestion @ `docs/en/connectors/source/GoogleBigtable.md:20`: Please double-check this against the source's SplitEnumerator — 'creates one split' is an implementation detail, and if the enumerator emits one split per row-key range (or per tablet) this tip would mislead users tuning parallelism. Also consider adding 'currently' (as the zh page does with 当前) so the statement doesn't go stale, and mirror the same feature-checklist rows on both EN and zh pages (zh now has a 流处理 row EN lacks).

### 11. [Sensitive Data Exposure via Job Configs] (LOW) Docs don't note that HOCON job configs (including GCP project/instance identifiers and credential paths) may be surfaced through engine logs, REST API, or checkpoint metadata
(a) The expanded examples encourage embedding `project_id`, `instance_id`, `table`, and `credentials_path` directly in job HOCON files, with no note about how these configs propagate. (b) In SeaTunnel Zeta the submitted job config is available via the REST API and often echoed in job-submission logs; while `credentials_path` is a pointer rather than a secret, exposing the exact key-file location plus GCP project/instance identifiers materially aids lateral movement for anyone with read access to logs or the cluster REST endpoint (they know exactly which file to exfiltrate and which Bigtable instance it unlocks). (c) Since these connector pages are being substantially expanded anyway, add a one-line note (or link to the existing variable-substitution docs) recommending parameterizing environment-specific values via `-i`/variable substitution and restricting access to the engine REST/UI in multi-tenant deployments; this is consistent with how other connector docs handle credential-adjacent options.

Evidence:
- `docs/zh/connectors/sink/GoogleBigtable.md:~158-163  project_id       = "my-gcp-project" / instance_id      = "my-bigtable-instance" / ... / credentials_path = "/secrets/sa-key.json"`
- `docs/zh/connectors/source/GoogleBigtable.md:~129-134  project_id       = "my-gcp-project" / ... / credentials_path = "/secrets/sa-key.json"`

Suggestion @ `docs/zh/connectors/source/GoogleBigtable.md:134`: Minor hardening suggestion: since job configs can be surfaced via engine logs and the Zeta REST API, exposing the exact `credentials_path` plus `project_id`/`instance_id` gives an observer everything needed to target the key file. Worth adding a brief note recommending variable substitution (`-i`) for environment-specific values and restricting REST/UI access in shared clusters.

### 12. [style] (LOW) ZH pages mix heading levels and English link labels/headings inconsistently with the EN pages
The ZH source page keeps `### Schema 映射` as a level-3 heading nested under `## 参数`, while the EN page uses a top-level `## Schema Mapping` section — after this PR the ZH schema-mapping section gained a full table and tip, making the nesting mismatch more visible in the sidebar TOC. Additionally, the ZH sink page introduces an English-language heading `### common options` and English link text `[Sink Common Options](...)` / `[Source Common Options](...)` inside otherwise fully translated pages, which is inconsistent with the PR's stated goal of translating labels consistently. This is purely presentational with no functional impact, but aligning heading levels and using translated link text (e.g., `[Sink 插件通用参数](../common-options/sink-common-options.md)`) would improve reader experience and match the EN structure. Please also verify the relative link `../common-options/source-common-options.md` resolves in the docs/zh tree.

Evidence:
- `docs/zh/connectors/source/GoogleBigtable.md:~100  ### Schema 映射  (EN equivalent is `## Schema Mapping` at docs/en/connectors/source/GoogleBigtable.md)`
- `docs/zh/connectors/sink/GoogleBigtable.md:~107  ### common options`
- `docs/zh/connectors/sink/GoogleBigtable.md:~105  多表写入时使用的 Sink 副本数。更多说明请参考 [Sink Common Options](../common-options/sink-common-options.md)。`
- `docs/zh/connectors/source/GoogleBigtable.md:~82  Source 插件通用参数，详见 [Source Common Options](../common-options/source-common-options.md)。`

Suggestion @ `docs/zh/connectors/sink/GoogleBigtable.md:107`: Minor style nit: since this PR aims to translate labels consistently, consider translating this heading (e.g., `### 通用参数`) and the `[Sink Common Options]` link text on this page, and promoting the ZH source page's `### Schema 映射` to `##` so it mirrors the EN `## Schema Mapping` section level. Also please double-check that `../common-options/sink-common-options.md` resolves under the docs/zh tree.

Suggested reviewers: @goutamadwant

PR #11373, opened 2026-07-07
[Docs][Connector-V2] Improve Feishu sink docs

DanielCarter-stack:
## Summary

This PR improves the Feishu sink connector documentation in both English and Chinese.

Changes include:
- document all Feishu sink options inherited from the HTTP sink option rule, including retry, batching, request interval, params, and multi-table sink replica
- clarify POST JSON request behavior, batching behavior, and non-exactly-once delivery caveat
- add practical examples for simple usage, headers with retry, JSON array batching, and multi-table webhook paths
- align the Chinese feature list wording with the connector documentation style

## Verification

- `./mvnw spotless:apply`

### DIFF
--- a/docs/en/connectors/sink/Feishu.md
+++ b/docs/en/connectors/sink/Feishu.md
@@ -52,24 +52,25 @@ Feishu webhook URLs and custom authentication headers are sensitive. Do not prin
 
 ## Sink Options
 
-| Name                        | Type    | Required | Default | Description                                                                                                 |
-|-----------------------------|---------|----------|---------|-------------------------------------------------------------------------------------------------------------|
-| url                         | String  | Yes      | -       | Feishu webhook URL. It can include `${database_name}`, `${schema_name}`, and `${table_name}` placeholders in multi-table jobs. |
-| headers                     | Map     | No       | -       | HTTP request headers. Use it when the webhook gateway requires extra headers.                               |
-| params                      | Map     | No       | -       | Accepted by the option rule. For the current sink writer, put query parameters directly in `url`; rows are posted to the final URL as the request body. |
-| retry                       | Int     | No       | -       | The maximum retry times when the HTTP request fails with an `IOException`.                                  |
-| retry_backoff_multiplier_ms | Int     | No       | 100     | Retry backoff multiplier in milliseconds.                                                                   |
-| retry_backoff_max_ms        | Int     | No       | 10000   | Maximum retry backoff in milliseconds.                                                                      |
-| array_mode                  | Boolean | No       | false   | Send rows as a JSON array when true, or as one JSON object per request when false.                          |
-| batch_size                  | Int     | No       | 1       | The maximum number of rows sent in one request. Only works when `array_mode` is true.                       |
-| request_interval_ms         | Int     | No       | 0       | Interval in milliseconds between two HTTP requests, used to avoid sending requests too frequently.          |
+| Name                        | Type    | Required | Default | Description                                                                                                             |
+|-----------------------------|---------|----------|---------|-------------------------------------------------------------------------------------------------------------------------|
+| url                         | String  | Yes      | -       | Feishu webhook URL. The current sink writer sends requests to this fixed URL and does not replace table-name placeholders. |
+| headers                     | Map     | No       | -       | HTTP request headers. Use it when the webhook gateway requires extra headers.                                           |
+| params                      | Map     | No       | -       | Accepted by the option rule, but the current sink writer does not pass it to requests. Put non-sensitive query parameters directly in `url` when needed. |
+| retry                       | Int     | No       | -       | The maximum retry times when the HTTP request fails with an `IOException`.                                              |
+| retry_backoff_multiplier_ms | Int     | No       | 100     | Retry backoff multiplier in milliseconds.                                                                               |
+| retry_backoff_max_ms        | Int     | No       | 10000   | Maximum retry backoff in milliseconds.                                                                                  |
+| array_mode                  | Boolean | No       | false   | Send rows as a JSON array when true, or as one JSON object per request when false.                                      |
+| batch_size                  | Int     | No       | 1       | The maximum number of rows sent in one request. Only works when `array_mode` is true.                                   |
+| request_interval_ms         | Int     | No       | 0       | Interval in milliseconds between two HTTP requests, used to avoid sending requests too frequently.                      |
 | multi_table_sink_replica    | Int     | No       | 1       | Number of sink replicas used for multi-table write. See [Sink Common Options](../common-options/sink-common-options.md). |
 | common-options              |         | No       | -       | Sink plugin common parameters, please refer to [Sink Common Options](../common-options/sink-common-options.md) for details. |
 
 ## Usage Notes
 
 - The sink always sends `POST` JSON requests; it does not expose a `method` option.
-- If the webhook URL needs query parameters, include them directly in `url`.
+- If the webhook URL needs query parameters, include non-sensitive parameters directly in `url`. Prefer `headers` for authentication material when the gateway supports it, because the full URL, including its query string, may appear in logs or job metadata.
+- Multi-table jobs can use `multi_table_sink_replica`, but the Feishu sink sends every row to the configured fixed `url`. It does not replace `${database_name}`, `${schema_name}`, or `${table_name}` in the URL.
 - `array_mode` is useful when the receiver accepts a JSON array and you want fewer HTTP requests.
 - Feishu webhook delivery is not exactly-once. If a retry happens after the remote service has already handled a request, the receiver may see duplicate messages.
 
@@ -103,7 +104,7 @@ source {
 
 sink {
   Feishu {
-    url = "https://open.feishu.cn/open-apis/bot/v2/hook/xxxx"
+    url = "https://open.feishu.cn/open-apis/bot/v2/hook/<your-hook-token>"
   }
 }
 ```
@@ -112,7 +113,7 @@ sink {
 
 ```hocon
 Feishu {
-  url = "https://open.feishu.cn/open-apis/bot/v2/hook/xxxx"
+  url = "https://open.feishu.cn/open-apis/bot/v2/hook/<your-hook-token>"
   headers {
     Content-Type = "application/json"
   }
@@ -126,18 +127,18 @@ Feishu {
 
 ```hocon
 Feishu {
-  url = "https://open.feishu.cn/open-apis/bot/v2/hook/xxxx"
+  url = "https://open.feishu.cn/open-apis/bot/v2/hook/<your-hook-token>"
   array_mode = true
   batch_size = 20
   request_interval_ms = 500
 }
 ```
 
-### Multiple table webhook path
+### Multi-table sink replica
 
 ```hocon
 Feishu {
-  url = "https://example.com/feishu/${database_name}/${table_name}"
+  url = "https://open.feishu.cn/open-apis/bot/v2/hook/<your-hook-token>"
   multi_table_sink_replica = 2
 }
 ```
--- a/docs/en/connectors/sink/Feishu.md
+++ b/docs/en/connectors/sink/Feishu.md
@@ -14,19 +14,26 @@ import ChangeLog from '../changelog/connector-http-feishu.md';
 
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
 - [ ] [cdc](../../introduction/concepts/connector-v2-features.md)
+- [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)
 - [ ] [timer flush](../../introduction/concepts/connector-v2-features.md)
 
 ## Description
 
-Used to launch Feishu web hooks using data.
+Used to launch Feishu webhooks using upstream data.
 
 > For example, if the data from upstream is [`age: 12, name: tyrantlucifer`], the body content is the following: `{"age": 12, "name": "tyrantlucifer"}`
 
-**Tips: Feishu sink only support `post json` webhook and the data from source will be treated as body content in web hook.**
+The Feishu sink sends `POST` requests. Each upstream row is converted to JSON and used as the request body. When `array_mode = true`, multiple rows are accumulated into one JSON array before sending.
+
+:::tip
+
+Feishu webhook URLs and custom authentication headers are sensitive. Do not print real tokens in logs or examples.
+
+:::
 
 ## Data Type Mapping
 
-|     Seatunnel Data Type     | Feishu Data Type |
+|     SeaTunnel Data Type     | Feishu Data Type |
 |-----------------------------|------------------|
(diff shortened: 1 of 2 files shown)


DanielCarter-stack:
Synced this branch with the latest dev branch and marked it ready for review because the docs-only CI checks passed.

SEZ9 (commented):
Thanks for the contribution, this is a valuable change! I went through the diff carefully — nice work overall. Before it merges, there are a few things I'd like to confirm/address, listed by priority (0 high, 3 medium):

### 1. 🟠 Medium · Docs — Multi-table write claim and `${database_name}`/`${table_name}` URL placeholders are not backed by the Feishu sink implementation
- The PR flips the feature checkbox to `[x] support multiple table write` and documents that `url` supports `${database_name}`, `${schema_name}`, and `${table_name}` placeholders, plus adds a whole 'Multiple table webhook path' example.
- The Feishu sink is a thin wrapper over `connector-http-base`, and `HttpSinkWriter` does not perform per-table placeholder substitution in the URL; that substitution pattern belongs to sinks like JDBC/File that explicitly implement it via `ReplaceName`/table-path replacement. Users following this example would send literal `${database_name}` strings to their webhook gateway.
- Impact: users configuring multi-table jobs against Feishu will silently get wrong URLs on all engines (Zeta/Flink/Spark), and the feature matrix would advertise a capability the connector does not implement.
- Fix: only check the multi-table box if `FeishuSink` (or its `HttpSink` base) actually implements `SupportMultiTableSink`, and remove the placeholder claim and the multi-table example unless the writer resolves table placeholders. Keep the docs strictly aligned with `FeishuHttpSinkOptions`/`HttpSinkWriter` behavior in the current code base.

**Where (this PR):**
- `docs/en/connectors/sink/Feishu.md:16  - [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)`
- `docs/en/connectors/sink/Feishu.md:56  | url | String | Yes | - | Feishu webhook URL. It can include `${database_name}`, `${schema_name}`, and `${table_name}` placeholders in multi-table jobs. |`
- `docs/en/connectors/sink/Feishu.md:139  url = "https://example.com/feishu/${database_name}/${table_name}"`
- `docs/zh/connectors/sink/Feishu.md:16  - [x] [支持多表写入](../../introduction/concepts/connector-v2-features.md)`

**Suggested fix** (`docs/en/connectors/sink/Feishu.md:56`): The Feishu sink delegates to `HttpSinkWriter` in `connector-http-base`, which does not substitute `${database_name}`/`${schema_name}`/`${table_name}` placeholders in the URL. As documented here (and in the 'Multiple table webhook path' example below), users would end up POSTing to a literal `.../feishu/${database_name}/${table_name}` path. Please remove the placeholder claim and the multi-table example (and revert the `[x] support multiple table write` checkbox) unless the connector code actually implements this substitution and `SupportMultiTableSink`.

### 2. 🟠 Medium · Security — Docs steer users to place webhook query parameters (often auth tokens/signatures) directly in `url`, where they are prone to leaking via logs and error messages
- Root cause: the new Usage Notes and the `params` option description explicitly tell users "If the webhook URL needs query parameters, include them directly in `url`". For Feishu-style webhooks and gateway fronts, query parameters frequently carry signatures, tokens, or API keys.
- Impact: the full request URL is the value most likely to surface in SeaTunnel task logs, retry/`IOException` stack traces, job config dumps, and REST job-info endpoints on Zeta/Flink/Spark. Encouraging secrets in the URL string materially increases the chance of credential exposure across all engines, especially when `retry` is enabled and failures are logged repeatedly.
- Fix: keep the guidance for functional query parameters, but add an explicit warning right next to it that authentication material should be passed via `headers` when the gateway supports it, and that the URL (including its query string) may appear in logs and job metadata. Apply the same wording to the Chinese page (`docs/zh/connectors/sink/Feishu.md`, 使用说明 section), which currently gives the same advice without a caveat.

**Where (this PR):**
- `docs/en/connectors/sink/Feishu.md:71  - If the webhook URL needs query parameters, include them directly in `url`.`
- `docs/en/connectors/sink/Feishu.md:58  | params                      | Map     | No       | -       | Accepted by the option rule. For the current sink writer, put query parameters directly in `url`; rows are posted to the final URL as the request body. |`
- `docs/zh/connectors/sink/Feishu.md:70  - 如果 Webhook URL 需要查询参数，请直接写在 `url` 中。`

**Suggested fix** (`docs/en/connectors/sink/Feishu.md:71`): Since Feishu/gateway webhook query parameters often carry tokens or signatures, and the full URL can end up in task logs, retry stack traces, and job config dumps, please add a warning here: prefer `headers` for authentication material when the gateway supports it, and be aware the URL query string may be exposed in logs. The nice tip block at the top mentions the URL is sensitive, but this note actively tells users to embed parameters in it without that caveat. Same applies to the zh page.

### 3. 🟠 Medium · Docs — Documented default for `batch_size` (1) does not match the HTTP sink base option default (1024)
- The new options table states `batch_size` defaults to `1`, but the base HTTP sink option rule and the existing `docs/en/connectors/sink/Http.md` document the default as `1024` (only effective when `array_mode = true`).
- Since the Feishu sink inherits its option rule from the HTTP sink, publishing a different default here will confuse users tuning batching behavior — they may assume they must set `batch_size` explicitly to get batching, or misjudge memory/latency behavior when they enable `array_mode`.
- Fix: copy the default straight from `HttpSinkOptions.BATCH_SIZE` (and cross-check `request_interval_ms` at line 64 the same way) so the Feishu table matches the base HTTP sink documentation, in both the English and Chinese files (docs/zh/connectors/sink/Feishu.md:63).

**Where (this PR):**
- `docs/en/connectors/sink/Feishu.md:63  | batch_size | Int | No | 1 | The maximum number of rows sent in one request. Only works when `array_mode` is true. |`
- `docs/zh/connectors/sink/Feishu.md:63  | batch_size | Int | 否 | 1 | 单次请求最多发送的数据条数，仅在 `array_mode` 为 true 时生效。 |`

**Suggested fix** (`docs/en/connectors/sink/Feishu.md:63`): The default here is documented as `1`, but the HTTP sink base (`HttpSinkOptions.BATCH_SIZE`, also documented in `docs/en/connectors/sink/Http.md`) uses `1024`. Since Feishu inherits the HTTP sink option rule, please take the default value from the code rather than restating it, and apply the same correction to the Chinese table (docs/zh/connectors/sink/Feishu.md:63). It would be good to double-check `request_interval_ms` on line 64 against the code default as well.

### 4. 🟢 Low · Security — Multi-table URL placeholder example should warn that `${database_name}`/`${table_name}` values are interpolated into the request path and must come from trusted catalogs
- Root cause: the new multi-table example builds the webhook destination from runtime table identity: `url = "https://example.com/feishu/${database_name}/${table_name}"`, with no note about where those values come from or how they are encoded.
- Impact: in multi-table jobs the placeholder values originate from upstream catalogs (e.g., CDC sources against multi-tenant databases). Table/database names containing `/`, `..`, `?`, or `#` characters would change the effective request path or query, redirecting deliveries to unintended endpoints on the same host — a request-routing/SSRF-adjacent concern the docs currently gloss over.
- Fix: add a sentence under this example (and the matching zh example) stating that placeholder values are taken from upstream table identifiers, should be treated as untrusted if the source is multi-tenant, and that operators should verify the sink URL-encodes or validates these segments. If the connector does not sanitize placeholder substitution today, that is worth a follow-up issue against the HTTP sink base.

**Where (this PR):**
- `docs/en/connectors/sink/Feishu.md:139    url = "https://example.com/feishu/${database_name}/${table_name}"`
- `docs/en/connectors/sink/Feishu.md:56  | url                         | String  | Yes      | -       | Feishu webhook URL. It can include `${database_name}`, `${schema_name}`, and `${table_name}` placeholders in multi-table jobs. |`
- `docs/zh/connectors/sink/Feishu.md:140    url = "https://example.com/feishu/${database_name}/${table_name}"`

**Suggested fix** (`docs/en/connectors/sink/Feishu.md:139`): Since `${database_name}`/`${table_name}` come from upstream table identifiers, names containing `/`, `..`, or `?` would change the effective request path/query and could route deliveries to unintended endpoints. Please add a note that these placeholder values must come from trusted catalogs and that operators should validate/encode them — and if the HTTP sink base doesn't sanitize the substituted segments today, let's track that as a follow-up.

### 5. 🟢 Low · Style — Markdown table column alignment is inconsistent between the English and Chinese option tables
- The English table pads columns to a fixed width (e.g. `| retry                       | Int     | No       | -       |`) while the Chinese table uses narrower, uneven padding (e.g. `| retry_backoff_multiplier_ms | Int     | 否    | 100   |`), and the separator rows have mismatched widths relative to the headers in both files.
- This is cosmetic but the connector docs across the repo (see the recently reviewed Jira/Klaviyo/Console doc PRs) keep table pipes aligned, which also makes future diffs cleaner.
- Fix: re-align the pipes in both tables so header, separator, and data rows share the same column widths.

**Where (this PR):**
- `docs/en/connectors/sink/Feishu.md:55  |-----------------------------|---------|----------|---------|-------------------------------------------------------------------------------------------------------------|`
- `docs/zh/connectors/sink/Feishu.md:60  | retry_backoff_multiplier_ms | Int     | 否    | 100   | 重试退避时间倍数，单位毫秒。 |`

**Suggested fix** (`docs/zh/connectors/sink/Feishu.md:60`): Nit: the column padding in this table is uneven compared to the English version (e.g. the `是否必需`/`默认值` columns). Aligning the pipes consistently across header, separator, and data rows keeps the docs style consistent with the other connector pages and makes future diffs easier to read.

### 6. 🟢 Low · Docs — English tip claims examples must not contain real tokens, but the simple example URL pattern should use an obvious placeholder marker
- The new tip warns users not to print real tokens in logs or examples, which is good, and the examples use `https://open.feishu.cn/open-apis/bot/v2/hook/xxxx`.
- To make the placeholder unmistakable and consistent with other connector docs, the hook path segment is better written as something like `your-hook-token` or `<hook-token>` rather than `xxxx`, which can be mistaken for a literal short token format.
- Fix: replace `xxxx` with an explicit placeholder in all four example blocks in both language files (e.g. docs/en/connectors/sink/Feishu.md lines 105, 114, 128 and the corresponding Chinese lines).

**Where (this PR):**
- `docs/en/connectors/sink/Feishu.md:105  url = "https://open.feishu.cn/open-apis/bot/v2/hook/xxxx"`
- `docs/en/connectors/sink/Feishu.md:29  Feishu webhook URLs and custom authentication headers are sensitive. Do not print real tokens in logs or examples.`

**Suggested fix** (`docs/en/connectors/sink/Feishu.md:105`): Nit: since the tip above emphasizes token sensitivity, consider using an explicit placeholder like `.../hook/<your-hook-token>` instead of `xxxx` here and in the other three examples (both EN and ZH), so readers immediately recognize it as a value to replace.

_Happy to discuss any of these — thanks again!_

DanielLeens (approved):
Thanks for the contribution! I went through the current Feishu docs diff and checked the connector contract behind the new documentation claims.

# What this PR fixes
- User pain point: the Feishu Sink page was too thin, so users could not reliably discover the current headers, retry, array mode, and multi-table behavior.
- Fix approach: update both English and Chinese docs with the option table, safer examples, and usage notes.
- In one sentence: this brings the Feishu Sink docs closer to the actual HTTP sink implementation.

# Code / contract review
The PR only changes `docs/en/connectors/sink/Feishu.md` and `docs/zh/connectors/sink/Feishu.md`.

Key path checked:
```text
User configures Feishu Sink
  -> FeishuSinkFactory.optionRule() inherits HttpSinkFactory.optionRule()
      -> optional(SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA)
  -> FeishuSink extends HttpSink
      -> HttpSink implements SupportMultiTableSink
  -> HttpSink.createWriter()
      -> HttpSinkWriter implements SupportMultiTableSinkWriter
```

The new multi-table docs are backed by the inherited HTTP sink contract: `HttpSink` implements `SupportMultiTableSink`, `HttpSinkWriter` implements `SupportMultiTableSinkWriter`, and `HttpSinkFactory` exposes `multi_table_sink_replica` through the option rule. I also checked that the English and Chinese pages were updated together and that the examples do not expose real webhook tokens.

This is documentation-only. It does not change APIs, option defaults, protocol, serialization, retry behavior, resource usage, or runtime logging.

- Checked the Feishu docs diff: scope is the two Feishu docs files.
- Checked Feishu and HTTP sink classes: confirmed the multi-table sink contract described above.
- I did not run a local build or tests; this review is based on the docs diff plus source-contract verification.

# Issues
No blocking issues found.

# Merge conclusion
### Conclusion: can merge
Blocking items: none.
Non-blocking suggestions: none.

Overall, this is a clean documentation update. Current CI is green. I recorded that the branch is currently `behind_by=1` against `dev`, but that alone is not a source-level blocker from my side.

DanielCarter-stack:
Updated the Feishu docs branch to address the review feedback:

- Removed the claim that the Feishu URL supports `${database_name}`, `${schema_name}`, and `${table_name}` replacement, and replaced the multi-table URL placeholder example with a `multi_table_sink_replica` example using a fixed webhook URL.
- Added a note that query parameters should only be non-sensitive, and that authentication material should prefer headers when the gateway supports it because the full URL may appear in logs or job metadata.
- Replaced `xxxx` webhook tokens with an explicit `<your-hook-token>` placeholder in both EN and ZH examples.
- Re-aligned the option tables in both files.

I kept `batch_size` default as `1` after checking `HttpSinkOptions.BATCH_SIZE` and `HttpParameter`, which both define the sink default as `1` on the current branch.

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head `f06e8a7070` after the follow-up docs commit.

+1 to @SEZ9's earlier source-backed points here: I rechecked the current Feishu docs against the actual HTTP sink contract, and the latest head closes the remaining doc-side gaps from that thread.

Current contract path rechecked:
```text
User configures Feishu Sink
  -> FeishuSinkFactory.optionRule() inherits HttpSinkFactory.optionRule()
     -> optional(SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA)
  -> FeishuSink extends HttpSink
     -> HttpSink implements SupportMultiTableSink
  -> HttpSink.createWriter()
     -> HttpSinkWriter implements SupportMultiTableSinkWriter
```

What changed on the latest head matches that contract:
- the unsupported `${database_name}` / `${table_name}` URL-placeholder claim is gone
- the docs now steer users toward fixed webhook URLs plus `multi_table_sink_replica`
- the examples use explicit token placeholders instead of real-looking hook values
- the EN/ZH pages both carry the same security caveat about putting only non-sensitive query params in the URL
- keeping `batch_size = 1` is correct on the current branch because the effective HTTP sink parameter default comes from `HttpParameter.batchSize = 1`

I do not see a reopened source-level blocker on the current docs revision, and the visible CI checks on this head are green.

Blocking items: none.

Overall this looks ready from Daniel's side.

davidzollo (approved):

PR #11374, opened 2026-07-07
[Fix][Connector-V2][JDBC] Fix duplicate column error when fetching SQL Server metadata with multiple extended properties

zhangweiwen:
#### **Background**

When using the SQL Server JDBC connector to synchronize data, the job fails with a "duplicate column" or "duplicate field" error, even when testing with a single table.

#### **Root Cause**

The issue stems from the SQL query used in `SqlServerCatalog` (or the SQL Server metadata dialect) to fetch table schema information.

The query performs a `LEFT JOIN` on `sys.extended_properties` to retrieve column comments (`MS_Description`). However, it only joins on `major_id` and `minor_id` without specifying `ext.name = 'MS_Description'` in the join condition:

```sql
LEFT JOIN sys.extended_properties ext ON ext.major_id = col.object_id AND ext.minor_id = col.column_id

```

In SQL Server, a single column can have **multiple extended properties** (e.g., custom metadata, system tracking, diagrams, or third-party tool configs) besides just the column description. Because of the missing name constraint, the `LEFT JOIN` multiplies the result rows for a single column, leading SeaTunnel to believe the table contains duplicate fields.

#### **Example Symptom**

If a column `ALARM_ID` has multiple extended properties (e.g., `-1`, `false`, `2370`), the query returns duplicate rows for the same `column_id`:

| table_name | column_name | comment | column_id | type | ... |
| --- | --- | --- | --- | --- | --- |
| DEF_ALARM_MSG | ALARM_ID | -1 | 1 | nvarchar | ... |
| DEF_ALARM_MSG | ALARM_ID | false | 1 | nvarchar | ... |
| DEF_ALARM_MSG | ALARM_ID | 2370 | 1 | nvarchar | ... |

#### **Changes**

* Moved and explicitly applied the `ext.name = 'MS_Description'` constraint directly into the `LEFT JOIN sys.extended_properties` clause.
* Corrected the subsequent `sys.default_constraints` join logic to ensure it maps properly based on the column's default object ID rather than leaking the `ext.name` filter into the wrong join condition.

**Updated SQL Logic:**

```sql
LEFT JOIN sys.extended_properties ext 
    ON ext.major_id = col.object_id 
   AND ext.minor_id = col.column_id 
   AND ext.name = 'MS_Description'
LEFT JOIN sys.default_constraints def 
    ON col.default_object_id = def.object_id

```

#### **Test Record**

* **Manual Verification:** Verified the updated query against a SQL Server instance containing tables with multiple extended properties. The query now correctly returns exactly one row per column.
* **Integration Test:** Executed the SeaTunnel sync job using the modified `connector-jdbc-sqlserver` JAR; the metadata parsing succeeded without throwing duplicate field errors.


davidzollo:

To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration https://github.com/apache/seatunnel/pull/11374/checks?check_run_id=85661325875

image

DanielLeens (changes requested):
Thanks for the contribution, and welcome! This is a good direction: moving the `MS_Description` filter onto the `extended_properties` join is the right way to avoid duplicate rows from unrelated extended properties. I found one remaining correctness issue that should be fixed before merge.

# What this PR fixes
- User pain point: SQL Server catalog metadata can become duplicated or incorrect when column comments/default constraints are read from system tables.
- Fix approach: tighten the joins against `sys.extended_properties` and `sys.default_constraints`.
- In one sentence: the PR is trying to make SQL Server column metadata discovery deterministic, but the default-value join is still tied to column comments.

# Code path checked
```text
JDBC catalog reads SQL Server table schema
  -> SqlServerCatalog.getSelectColumnsSql(tablePath)
      -> SELECT_COLUMNS_SQL_TEMPLATE
          -> LEFT JOIN sys.extended_properties ext ... ext.name = 'MS_Description'
          -> LEFT JOIN sys.default_constraints def ... AND ext.minor_id = col.column_id
  -> ResultSet.default_value
  -> SqlServerCatalog.buildColumn()
  -> BasicTypeDefine.defaultValue(defaultValue)
```

# Issue 1: default constraints are still joined through `ext`
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/sqlserver/SqlServerCatalog.java:58`
- Problem: after this change, `ext` only exists when the column has an `MS_Description` extended property. If a column has a default constraint but no column comment, `ext.minor_id` is `NULL`, so `AND ext.minor_id = col.column_id` prevents `def` from joining. The result is still `default_value = null` for a valid SQL Server default constraint.
- Risk: SeaTunnel will keep building incomplete SQL Server catalog metadata. SaveMode, auto-create, or metadata display paths can treat a defaulted column as if it has no default.
- Suggested fix: decouple `def` from `ext`, for example:
  - `LEFT JOIN sys.default_constraints def ON col.default_object_id = def.object_id`, or
  - `LEFT JOIN sys.default_constraints def ON def.parent_object_id = col.object_id AND def.parent_column_id = col.column_id`.
- Severity: medium, merge-blocking for this metadata fix.

# Testing / CI
I did not run local tests; this review is based on the local diff and the SQL Server catalog call chain. Please also add or update a regression case for a SQL Server column that has a default value but no `MS_Description`, plus the duplicate extended-property case this PR is fixing.

Current CI is `ACTION_REQUIRED`: https://github.com/apache/seatunnel/runs/85685411920. Since this PR is from a first-time contributor and the label is already present, please enable/approve the workflow in your fork or PR checks first, then rerun CI after the code update.

# Merge conclusion
### Conclusion: can merge after changes
Blocking items:
1. Fix Issue 1 so default constraints do not depend on the presence of column comments.

Non-blocking suggestions:
- Add the SQL Server metadata regression cases described above.

Overall, this is very close. Once the default-constraint join is corrected and CI is enabled, I’ll be happy to re-review.

DanielLeens (changes requested):
Thanks for the update, and welcome again. I rechecked the latest head and one source-level blocker from my previous round is still present. One CI fact has also changed since my last review, so I want to update that explicitly as well.

# What this PR fixes
- User pain point: SQL Server catalog metadata can return duplicated or incorrect column rows when `sys.extended_properties` contains multiple entries for the same column.
- Fix approach: narrow the `extended_properties` join to `MS_Description` so comment rows are filtered earlier, while still reading the column default constraint.
- In one sentence: this PR is moving the SQL Server column-metadata query in the right direction, but the default-value lookup is still accidentally tied to whether a column has an `MS_Description` comment.

# Full call path I checked
```text
JDBC catalog reads a SQL Server table
  -> AbstractJdbcCatalog.getTable(tablePath)
      -> buildColumnsReturnTablaSchemaBuilder(tablePath, conn)
          -> PreparedStatement(getSelectColumnsSql(tablePath))
              -> SqlServerCatalog.getSelectColumnsSql()
                  -> SELECT sys.columns
                     LEFT JOIN sys.extended_properties ext
                     LEFT JOIN sys.default_constraints def
          -> buildColumnsWithErrorCheck(...)
              -> SqlServerCatalog.buildColumn(resultSet)
                  -> resultSet.getObject("default_value")
                  -> BasicTypeDefine.builder().defaultValue(defaultValue)
```

# Core logic analysis
The key change is in `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/sqlserver/SqlServerCatalog.java:57-58`.

Before:
```java
LEFT JOIN sys.extended_properties ext
  ON ext.major_id = col.object_id AND ext.minor_id = col.column_id
LEFT JOIN sys.default_constraints def
  ON col.default_object_id = def.object_id
 AND ext.minor_id = col.column_id
 AND ext.name = 'MS_Description'
```

After:
```java
LEFT JOIN sys.extended_properties ext
  ON ext.major_id = col.object_id
 AND ext.minor_id = col.column_id
 AND ext.name = 'MS_Description'
LEFT JOIN sys.default_constraints def
  ON col.default_object_id = def.object_id
 AND ext.minor_id = col.column_id
```

Filtering `ext` to `MS_Description` earlier is the right part of the fix, because it prevents unrelated extended-property rows from duplicating the column metadata row. The remaining problem is that `def` still depends on `ext.minor_id = col.column_id`.

# Issue 1: default-constraint lookup still depends on the presence of `MS_Description`
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/sqlserver/SqlServerCatalog.java:57-58`
- Problem: after this change, `ext` is only non-null when the column has an `MS_Description` entry. For a column that has a valid default constraint but no comment, `ext` is `NULL`, so `AND ext.minor_id = col.column_id` still filters out the `def` row. As a result, `buildColumn()` continues to receive `default_value = null` for a real SQL Server default constraint.
- Risk: SeaTunnel will still build incomplete SQL Server catalog metadata on the normal catalog-read path. Anything that consumes the catalog column definition downstream can incorrectly treat a defaulted column as if it had no default.
- Suggested fix:
  - Option A: `LEFT JOIN sys.default_constraints def ON col.default_object_id = def.object_id`
  - Option B: `LEFT JOIN sys.default_constraints def ON def.parent_object_id = col.object_id AND def.parent_column_id = col.column_id`
  In both cases, the important part is to decouple the default-constraint join from `ext`.
- Severity: blocker
- Previously raised by others: same blocker as my earlier review; it still exists on the latest head.

- Compatibility: fully backward compatible. No API, config, protocol, serialization, or default-value contract changes are introduced by the patch itself.
- Performance / side effects: no material CPU, memory, GC, network, or concurrency regression from the current diff. The concern here is correctness of the metadata row, not runtime overhead.
- Error handling / logging: unchanged in this PR.

# Tests
I did not run local tests; this review is based on the full diff plus the live catalog call chain above.

I still recommend adding a regression case for both scenarios below:
- multiple unrelated `extended_properties` entries on the same column should not duplicate the column row
- a column with a default constraint but without `MS_Description` should still return `default_value`

The current PR does not add automated coverage for either case.

# Current CI status
The CI fact changed from my previous round, so I want to update it clearly: the current `Build` failure is not pointing to a new JDBC-source failure caused by this diff. The failing fork run is `https://github.com/zhangweiwen/seatunnel/actions/runs/28953450560`, and the failed jobs (`changes`, `Dead links`, `License header`, `Code style`) were not actually started because GitHub reports: `The job was not started because your account is locked due to a billing issue.`

So at this point:
- the source-level blocker is still Issue 1 above
- the current red CI signal is a separate fork Actions/account issue, not evidence that this JDBC change itself failed in those lanes

Once Issue 1 is fixed, please also resolve the fork Actions/account problem and rerun CI on the updated head.

# Merge conclusion
### Conclusion: can merge after changes

Blocking items:
1. Fix Issue 1 so the default-constraint join no longer depends on whether `MS_Description` exists for the column.

Non-blocking suggestions:
- Add regression coverage for the two SQL Server metadata scenarios described above.
- After the code fix, rerun CI once the fork Actions/account issue is resolved.

Overall, the direction is good and the duplicate-row part is closer, but the current head still leaves the default-value metadata path incorrect. Once that join is decoupled and CI is rerun, I’m happy to take another look.

zhangweiwen:
@DanielLeens I changed the code use your option A fix.

DanielLeens (approved):
Thanks for applying the fix! I re-reviewed the latest head `58cada0a8b` from scratch.

# What changed since my last review
You applied Option A from my previous review — the `default_constraints` join is now decoupled from `ext`:

```sql
LEFT JOIN sys.default_constraints def ON col.default_object_id = def.object_id
```

This is exactly right. A column with a default constraint but no `MS_Description` comment will now correctly return its `default_value` in the catalog metadata. The source-level blocker from my previous round is resolved.

# Full re-review of the current head
The diff is a 2-line change to the SQL Server column metadata query in `SqlServerCatalog.java:57-58`. The `extended_properties` join is filtered to `MS_Description` early (preventing duplicate rows from unrelated properties), and the `default_constraints` join now depends only on `col.default_object_id = def.object_id` (independent of whether a comment exists).

No new source-level blockers found on this head.

# CI status
The Build is still failing, but the failure links to your fork run (`zhangweiwen/seatunnel/actions/runs/29076205230`) which appears to be the same fork Actions/billing issue from the previous round. This is not a source code failure. Once the fork Actions issue is resolved and CI is rerun, the gate should clear.

The branch is also `behind_by=1` from `dev`, so a quick sync before the final CI run would be good.

# Merge conclusion
Source blocker cleared from my side. The remaining gate is CI (fork Actions issue) + sync with latest `dev`. Once those are resolved, this is ready to merge.

My earlier `changes requested` entry may still appear in GitHub's review summary because I do not have write-level review controls on this repository — a maintainer with write access may still need to issue the final approval step on top of this technical recheck.

DanielLeens:
Rechecked the current head again: the source-level blocker from my previous round is still closed for me, and I do not see a new code-side blocker from Daniels side on this same revision.

The part that changed since my last comment is the CI fact. The current remaining gate is now this `Build` failure: https://github.com/apache/seatunnel/runs/86385064237

The linked test-results summary currently points to the contributor fork run here: https://github.com/zhangweiwen/seatunnel/actions/runs/29099485107/job/86384380982

Since this branch is not behind the current `dev` base in the live compare metadata, I am not asking for an upstream sync just for mergeability. From Daniels side the source blocker remains cleared; the remaining gate is the current `Build` result above.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `9233154180a3` from scratch.

# What this PR fixes
- User pain: SQL Server sync jobs fail with a "duplicate column/field" error whenever any column carries more than one extended property, because the `sys.extended_properties` LEFT JOIN was unconstrained by property name and fanned out one row per property per column.
- Fix approach: Move the `ext.name = 'MS_Description'` predicate into the `sys.extended_properties` join condition where it belongs, and strip the stray `ext.minor_id`/`ext.name` predicates that had leaked into the `sys.default_constraints` join in `SqlServerCatalog.SELECT_COLUMNS_SQL_TEMPLATE`.
- One-line summary: The two-line SQL change is correct and minimal — the join now returns exactly one row per column with only the MS_Description comment — with one behavioral side effect worth flagging: default values are now populated for columns that lack a comment, where the old (buggy) join predicates would have nulled them out.

# Runtime chain I rechecked
```text
AbstractJdbcCatalog.getTable(TablePath)  AbstractJdbcCatalog.java:~180
  -> SqlServerCatalog.getListColumnsSql(tablePath)  SqlServerCatalog.java:~100 (formats SELECT_COLUMNS_SQL_TEMPLATE, SqlServerCatalog.java:46-60 — the changed lines)
  -> AbstractJdbcCatalog.buildTable(resultSet, tablePath)  AbstractJdbcCatalog.java:~200 (iterates one ResultSet row per column; pre-fix, fan-out from the unfiltered ext join yielded repeated column_id rows)
  -> SqlServerCatalog.buildColumn(resultSet)  SqlServerCatalog.java:~110 (reads column_name, comment, default_value per row)
  -> TableSchema.Builder.column(...) / SeaTunnelRowType construction — where the duplicate-field check fires when the same column name is added twice
```

# Findings
**Issue 1: Upgrade behavior change: column default values are now returned for columns without an MS_Description comment, which previously came back NULL due to the ext predicates leaking into the def join**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/sqlserver/SqlServerCatalog.java:58`
- Why it matters: - Root cause: before this PR, the `sys.default_constraints` LEFT JOIN carried `ext.minor_id = col.column_id AND ext.name = 'MS_Description'` in its ON clause, so for any column lacking an `MS_Description` extended property (`ext` columns NULL) the join predicate evaluated to UNKNOWN and `def` was always NULL. The corrected join `ON col.default_object_id = def.object_id` now resolves defaults for all columns.
- Impact: this is the correct behavior, but it is a visible metadata change on upgrade. Pipelines using SQL Server as a source with sink `save_mode` auto table creation will start seeing `defaultValue` populated where it was previously NULL, changing the generated CREATE TABLE DDL. SQL Server also stores default definitions wrapped in parentheses (e.g. `((0))`, `('N/A')`), and because these values were effectively never surfaced before, downstream default-value handling for this dialect has had little real-world exercise.
- Fix: call out this behavior change in the PR description/release notes, and verify (with a test) that `((0))`-style definitions from `sys.default_constraints.definition` are normalized correctly when SeaTunnel builds sink DDL rather than being emitted verbatim.
- Suggested fix: This correction is right, but note it is also a behavior change on upgrade: previously the leaked `ext.*` predicates made this join return NULL for every column without an `MS_Description` property, so defaults were effectively never populated. Now `defaultValue` will be filled for all columns with a default constraint, which affects sink auto-created DDL under `save_mode`. SQL Server returns definitions like `((0))` / `('N/A')` — please confirm these are normalized when building sink CREATE TABLE statements and mention this behavior change in the changelog.
- Severity: Medium

# Review conclusion

- Issue 1: Upgrade behavior change: column default values are now returned for columns without an MS_Description comment, which previously came back NULL due to the ext predicates leaking into the def join

DanielLeens:
Thanks @SEZ9. I rechecked the same head `9233154180a3` after your follow-up review.

I agree with your point that the now-correct default-value lookup is a real behavior change worth documenting, because columns without `MS_Description` will now surface `default_value` where the old buggy join path silently returned `NULL`. I would still keep that as non-blocking on this PR, though, because the current two-line SQL fix is the right source-level correction for the duplicate-row bug and Daniel's earlier blocker on the `default_constraints` join is already closed on this head.

So from Daniel's side:
- source blocker remains cleared on the current head
- your upgrade-note or release-note point is valid, but I would treat it as a follow-up rather than a new merge blocker
- the current remaining gate is still the latest `Build` failure here: https://github.com/apache/seatunnel/runs/86385064237

Happy to recheck again if a new head is pushed, but on this unchanged revision I am not reopening a new source-side blocker.

zhangweiwen:
Thanks @DanielLeens , agreed on all points above.

Regarding the remaining Build failure on this head: it's actually caused by a billing issue on my GitHub account (my fork's Actions runs are being blocked with "account is locked due to a billing issue"), not by the source change itself. I've already checked my billing details and there's no actual outstanding balance, so this looks like a stale/incorrect lock on GitHub's side. I've opened a support ticket with GitHub to get it resolved.

I'll push again (or just re-run the workflow) once my account is unlocked so the Build job can go green. Will update this thread as soon as I hear back from GitHub Support.
image

DanielLeens:
Thanks for the clear update, and sorry you are having to deal with that GitHub Actions account lock. From Daniel's side I still do not see a new source-level blocker on this head, and I would not patch this PR with unrelated dependency or workflow changes just to work around an external billing/account state.

The current failing gate is `Build` at https://github.com/apache/seatunnel/runs/86385064237. The lowest-cost next step is exactly what you described: wait for GitHub Support / account unlock, then rerun the Build job or push a no-op source refresh if rerun is unavailable. After it reruns, please drop the new check link here if it still fails and I can help narrow it down against the fresh result.

One small queue note: this head is `diverged` and `behind_by=1`, but since the visible blocker is the fork Actions account state, I am not treating that as a request to make code changes in this PR.

DanielLeens:
Rechecked the latest head `cfa8c8ebb9` and the current merge gate.

Source-side conclusion remains clear from Daniel's side. The SQL Server catalog query now does both pieces correctly:
- `sys.extended_properties` is filtered to `MS_Description` in the `ext` join, so unrelated extended properties no longer fan out duplicate column rows.
- `sys.default_constraints` is joined independently via `col.default_object_id = def.object_id`, so default values no longer depend on a column comment being present.

+1 to @SEZ9's note that surfacing `default_value` for columns without `MS_Description` is a visible metadata behavior change worth mentioning. I still would not treat it as a new merge blocker for this PR, because it is the correct result of fixing the old join bug.

The CI fact has changed since my earlier follow-up: this branch is now up to date with the live `dev` compare (`ahead_by=6`, `behind_by=0`), and the previous fork-account `Build` failure is no longer the current gate. The current remaining gate is the in-progress `Build` check here: https://github.com/apache/seatunnel/runs/86782380629

One GitHub-side note: the PR still shows `reviewDecision=REVIEW_REQUIRED`, so even though Daniel's technical source blocker is cleared, a write-capable maintainer may still need to provide the final approval/review gate once Build completes.

DanielLeens:
Rechecked the current head `cfa8c8ebb9f9` against the latest `dev` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
JDBC catalog reads a SQL Server table
  -> AbstractJdbcCatalog.getTable(tablePath)
      -> buildColumnsReturnTablaSchemaBuilder(tablePath, conn)
          -> PreparedStatement(getSelectColumnsSql(tablePath))
              -> SqlServerCatalog.getSelectColumnsSql()
                  -> SELECT sys.columns
                     LEFT JOIN sys.extended_properties ext
                     LEFT JOIN sys.default_constraints def
          -> buildColumnsWithErrorCheck(...)
              -> SqlServerCatalog.buildColumn(resultSet)
                  -> resultSet.getObject("default_value")
                  -> BasicTypeDefine.builder().defaultValue(defaultValue)
```

## Merge conclusion

1. Blocking items
- This branch is behind upstream (`compare status: diverged`, `ahead_by=6`, `behind_by=1`), and the current failing gate is `FAILURE gate: Build`: https://github.com/apache/seatunnel/runs/86782380629

2. Lowest-cost next step

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `40700ad6241a` against the SQL Server catalog metadata path.

# What this PR fixes
- User pain: SQL Server column metadata can be duplicated when a column has multiple extended properties, because the old `sys.extended_properties` join could fan out one row per property.
- Fix approach: keep `ext.name = 'MS_Description'` on the extended-properties join, and keep the default-constraint join independent via `col.default_object_id = def.object_id`.
- In one sentence: the catalog query now returns one row per column while still preserving default values for columns without comments.

# Runtime path checked
```text
SQL Server catalog schema discovery
  -> AbstractJdbcCatalog.getTable(tablePath)
      -> buildColumnsReturnTablaSchemaBuilder(...)
          -> SqlServerCatalog.getSelectColumnsSql(tablePath)
              -> sys.columns
              -> LEFT JOIN sys.extended_properties ... ext.name = 'MS_Description'
              -> LEFT JOIN sys.default_constraints ... col.default_object_id = def.object_id
          -> ResultSet.default_value / column_comment
          -> build Column metadata
```

This normal JDBC catalog path does hit the changed SQL. The previous Daniel-side blocker about default values being tied to column comments remains fixed on this head.

Fully compatible. No API, option, default, protocol, serialization, or historical behavior contract changes. The query should reduce duplicate-row risk rather than add resource pressure.

# Test coverage and stability
The effective current diff against latest `dev` is the SQL template change only. I did not run local tests; per the SeaTunnel PR review rule, this review relies on source analysis plus the current GitHub CI for execution proof. Test stability analysis is not applicable to this effective diff.

# Merge conclusion
### Conclusion: can merge after CI completes

1. Blocking items that must be fixed
- No source-level blocker found in this review.
- The current `Build` check is still queued, so merge should wait for the required CI result.

2. Non-blocking suggestions
- None.

Overall, I agree with the existing maintainer direction here. I am leaving this as `COMMENTED` rather than `APPROVED` only because the latest required Build has not completed yet.

DanielLeens (approved):
Thanks for the contribution and for keeping this SQL Server metadata fix focused.

I rechecked the current gate state after my previous Daniel review, where the source path was clear and the remaining item was the queued `Build` check. That CI fact has now changed: the current required checks are green on this head.

Because the previous Daniel conclusion had no source-side blocker and only the CI / merge gate was still pending, I am updating my review conclusion to approve this revision from Daniel's side.

## Current conclusion

### Conclusion: can merge

1. Blocking items that must be fixed
   - None from Daniel's current source review.

2. Suggested follow-up, non-blocking
   - None from this review.

The runtime path remains the same as the previous review:

```text
SQL Server catalog schema discovery
  -> AbstractJdbcCatalog.getTable(tablePath)
      -> buildColumnsReturnTablaSchemaBuilder(...)
          -> SqlServerCatalog.getSelectColumnsSql(tablePath)
              -> sys.columns
              -> LEFT JOIN sys.extended_properties ... ext.name = 'MS_Description'
              -> LEFT JOIN sys.default_constraints ... col.default_object_id = def.object_id
          -> ResultSet.default_value / column_comment
          -> build Column metadata
```

The fix still looks compatible and narrowly scoped: it prevents duplicate column metadata rows caused by multiple SQL Server extended properties while preserving default-value discovery and the existing JDBC catalog flow.

davidzollo (approved):
+1
LGTM

PR #11377, opened 2026-07-08
[Improve][Connector-V2][SftpFile] Defer binary file discovery

### LINKED ISSUE #11306: [Bug] SftpFile REST submit may block for a long time before returning `jobId` when source directory contains many files
### Search before asking

## Describe the bug

When using `SftpFile -> S3File` and submitting the job through SeaTunnel REST API, the request may stay in the submit phase for a very long time before returning `jobId`.

In our case, the submit request stayed in source preparation for about 15 minutes before `jobId` was returned.

This happened in both:

- `sync_mode=update` with `file_filter_modified_start/end`
- `sync_mode=full` with `file_filter_modified_start/end`

Because `full` mode shows nearly the same delay as `update`, we believe the main bottleneck is not target-side compare, but source-side pre-scan / file enumeration / local filtering during submit.

This behavior is especially problematic when REST is exposed through ingress or proxy, because `submit-job` can hit HTTP timeout first, while the job may still continue in the backend.

## Deployment
- SeaTunnel deployed on Kubernetes
- Job submitted through REST API
- REST API exposed through ingress/proxy
- Source SFTP server deployed on a separate Windows VM
- Target object storage is MinIO in the same Kubernetes cluster

### SeaTunnel Version

`2.3.13`

### SeaTunnel Config

```conf
Reproduction config
The real config contains secrets, so below is a masked and simplified example

Case 1: `update` mode with time window
{
  "env": {
    "parallelism": 2,
    "job.mode": "BATCH",
    "job.name": "st_blob_sftp_salecontract_files_to_minio_incr"
  },
  "source": [
    {
      "plugin_name": "SftpFile",
      "host": "xxx",
      "port": "22",
      "user": "xxx",
      "password": "xxx",
      "path": "/SaleContractFiles/202606",
      "file_filter_pattern": ".*\\.(pdf|PDF|doc|docx|xls|xlsx)$",
      "file_format_type": "binary",
      "binary_complete_file_mode": "true",
      "target_path": "s3a://lakehouse-dev/data/bronze/blobs/app0043/SaleContractFiles/202606",
      "sync_mode": "update",
      "update_strategy": "distcp",
      "compare_mode": "len_mtime",
      "file_filter_modified_start": "2026-06-30 00:00:00",
      "file_filter_modified_end": "2026-07-01 00:00:00",
      "target_hadoop_conf": {
        "fs.defaultFS": "s3a://lakehouse-dev",
        "fs.s3a.endpoint": "http://minio-svc:9000",
        "fs.s3a.path.style.access": "true",
        "fs.s3a.connection.ssl.enabled": "false"
      }
    }
  ],
  "transform": [],
  "sink": [
    {
      "plugin_name": "S3File",
      "bucket": "s3a://lakehouse-dev",
      "path": "/data/bronze/blobs/app0043/SaleContractFiles/202606",
      "file_format_type": "binary",
      "is_enable_transaction": "false"
    }
  ]
}
Case 2: `full` mode with time window
{
  "env": {
    "parallelism": 2,
    "job.mode": "BATCH",
    "job.name": "st_blob_sftp_salecontract_files_to_minio_incr"
  },
  "source": [
    {
      "plugin_name": "SftpFile",
      "host": "xxx",
      "port": "22",
      "user": "xxx",
      "password": "xxx",
      "path": "/SaleContractFiles/202606",
      "file_filter_pattern": ".*\\.(pdf|PDF|doc|docx|xls|xlsx)$",
      "file_format_type": "binary",
      "binary_complete_file_mode": "true",
      "file_filter_modified_start": "2026-06-29 00:00:00",
      "file_filter_modified_end": "2026-06-30 00:00:00"
    }
  ],
  "transform": [],
  "sink": [
    {
      "plugin_name": "S3File",
      "bucket": "s3a://lakehouse-dev",
      "path": "/data/bronze/blobs/app0043/SaleContractFiles/202606",
      "file_format_type": "binary",
      "is_enable_transaction": "false"
    }
  ]
}
```

### Running Command

```shell
restapi submit
```

### Error Exception

```log
## Source directory characteristics

- The configured path is already the deepest business directory, and it is a flat directory
- The directory contains about `3400+` files
- The time window usually matches only about `80 ~ 200` files
- About `2800` files may already exist on target in `update` scenarios

## Steps to reproduce

1. Prepare an SFTP directory with `3000+` files under one flat directory.
2. Configure `SftpFile` source with `file_filter_modified_start/end`.
3. Submit the job through SeaTunnel REST API.
4. Observe that the submit request stays for a long time before returning `jobId`.

## Actual behavior

- `submit-job` spends a long time in source preparation before `jobId` is returned
- The long delay happens before real job execution starts
- `parallelism=2` does not help this stage
- `full` mode and `update` mode both show similar submit latency
- If REST is behind ingress/proxy, client may get `504 Gateway Time-out`, but the job can still continue in backend and eventually succeed

## Expected behavior

- `submit-job` should return `jobId` much earlier
- Expensive source file enumeration should ideally not block the REST submit request for such a long time
- Time-window filtering should reduce submit cost more effectively, or at least there should be better visibility in logs about where time is spent

## Log timeline

### `Job_test_202606_update_window`


18:35:48.630 start generating all sources
18:35:48.633 Update sync mode enabled
18:50:46.326 Update sync mode statistics: scanned=101, skipped=101, to_sync=0
18:50:46.329 start generating all transforms
18:50:46.390 Init JobMaster
18:50:47.689 EnumeratorOpenEvent


Observed submit-side delay before `jobId`: about `14m 58s`

### `Job_test_202606_full_window`


18:44:31.369 start generating all sources
18:59:17.480 start generating all transforms
18:59:17.552 Init JobMaster
18:59:18.809 EnumeratorOpenEvent


Observed submit-side delay before `jobId`: about `14m 46s`

The important point is that `full` mode does not do target compare, but it still spends almost the same amount of time before job creation.

## Additional observations

### 1. The bottleneck seems to be before job creation

From logs, the long delay happens between:

- `start generating all sources`
- `start generating all transforms` / `Init JobMaster`

So the delay happens during source preparation in the submit thread, not during normal reader/sink runtime.

### 2. `file_filter_modified_start/end` does not appear to reduce source scan cost enough

Even though the time window matches only a small subset of files, SeaTunnel still appears to need full source metadata enumeration before filtering.

### 3. One manual SFTP directory listing is much faster than the end-to-end submit delay

From a Kubernetes node, a manual SFTP `ls` test on the same directory returned about `3654` entries in about `2.3s`.

So the observed `~15 minutes` delay seems much larger than a single directory listing, which suggests the expensive part is likely inside connector-side preparation logic rather than just one simple remote `ls`.

### 4. Master health logs did not show obvious JVM pressure

During the submit delay, we did not observe clear evidence of:

- high CPU saturation
- Hazelcast queue backlog
- Full GC pressure

So this does not look like a generic master resource exhaustion issue.

## Source-code-level analysis

Based on local source analysis, the submit path seems to be:

1. `MultipleTableJobConfigParser.parseSource()`
2. `FactoryUtil.createAndPrepareSource()`
3. `BaseFileSourceConfig.parseFilePaths()`
4. `ReadStrategy.getFileNamesByPath()`
5. `AbstractReadStrategy.collectFileNamesByPath()`

Relevant files:

- `seatunnel-engine/seatunnel-engine-core/.../MultipleTableJobConfigParser.java`
- `seatunnel-connectors-v2/connector-file/connector-file-base/.../BaseFileSourceConfig.java`
- `seatunnel-connectors-v2/connector-file/connector-file-base/.../AbstractReadStrategy.java`
- `seatunnel-connectors-v2/connector-file/connector-file-base/.../BinaryReadStrategy.java`
- `seatunnel-connectors-v2/connector-file/connector-file-sftp/.../SFTPFileSystem.java`

### Why submit is blocked

`BaseFileSourceConfig` initializes `readStrategy` and immediately calls `parseFilePaths()`, which calls:


return readStrategy.getFileNamesByPath(rootPath);


This means file path enumeration is done during source config creation, before the job is fully created and before `jobId` is returned.

### How time-window filtering works

In `AbstractReadStrategy.collectFileNamesByPath()`:

- `hadoopFileSystemProxy.listStatus(path)` is called first
- then each `FileStatus` is filtered locally
- `filterFileByModificationDate(fileStatus)` is applied after metadata is already fetched

So `file_filter_modified_start/end` is local filtering on `FileStatus`, not a server-side SFTP pushdown.

### Why `update` mode is not the only problem

In `update` mode, `shouldSyncFileInUpdateMode()` is called after source-side filtering.

For `distcp + len_mtime`, the logic is roughly:

1. Build target file path from source relative path
2. Call `targetHadoopFileSystemProxy.getFileStatus(targetFilePath)`
3. If target file does not exist, sync it
4. If target length differs, sync it
5. If length is equal and source `mtime > target mtime`, sync it
6. Otherwise skip it

This target compare adds extra cost in `update` mode, but because `full` mode still has almost the same `~15 min` submit delay, target compare does not appear to be the dominant cause in this case.

### SFTP filesystem implementation detail

In `SFTPFileSystem.listStatus(Path)`:

- it first calls internal `getFileStatus(client, absolute)`
- then for directories it calls `client.ls(...)`

Also, `BinaryReadStrategy.init()` first calls:


basePathIsFile = hadoopFileSystemProxy.isFile(basePath);


So some extra metadata round trips happen even before actual enumeration finishes.

## Suspected root cause

The main issue appears to be that `SftpFile` performs synchronous, submit-time source enumeration and filtering in the REST submit thread.

For directories with many files:

- the full source directory metadata still needs to be traversed first
- `file_filter_modified_start/end` only filters after metadata is already available locally
- `parallelism` does not help, because the delay happens before job creation

This can make REST submit very slow and can even cause HTTP timeout at the gateway layer.

## Impact

- Very slow `submit-job` response for `SftpFile`
- Hard to use behind ingress/proxy because `submit-job` may hit timeout
- Client may fail to get `jobId`, while backend job is still running
- Time-window filtering is not enough to make large flat directories practical

## Suggested improvement directions

1. Avoid doing full source file enumeration synchronously in the REST submit path.
2. Consider moving expensive source preparation to a later async stage after `jobId` creation.
3. Add finer-grained timing logs for:
   - source `listStatus`
   - local file filtering
   - update-mode target compare
4. Consider whether some filtering or optimization can be pushed down or cached for SFTP sources.
5. Consider improvements for large flat-directory scenarios, because they are common in file-drop integrations.

## Question for maintainers

Is this current behavior expected for `SftpFile`, or could it be considered a bug / performance issue in submit-time source preparation?

If needed, I can also provide a sanitized full config and more complete submit logs.
```

SeaTunnel Zeta

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

goutamadwant:

Fixes #11306.

For SftpFile binary sources in once discovery mode, file discovery was done while the source config was being created. In the REST submit path that happens before `jobId` is returned, so a large SFTP directory can keep `submit-job` waiting while the connector lists and filters remote files.

This change keeps schema/config construction lightweight for SftpFile binary sources and defers full file discovery until the once-mode split enumerator opens. The deferred lookup uses a temporary read strategy so closing discovery resources does not close the strategy retained by the source reader. Other file formats keep the existing eager behavior because some formats still infer schema from discovered files.

This also adds a timing log for file discovery with plugin name, path, discovered file count, and elapsed time.

Yes. For SftpFile binary sources, REST submit can return the `jobId` before full source file discovery runs. If deferred discovery fails, the job fails during split enumeration instead of failing during submit.

Non-binary file formats keep the existing behavior.

Added unit coverage for the deferred discovery path:

- deferred file discovery does not call `listStatus` during source config creation
- eager file discovery keeps the existing config-time discovery behavior
- the split enumerator discovers deferred file paths when `open()` runs
- deferred discovery closes its temporary resources without closing the read strategy retained by the source reader

Verified with:

```shell
./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-base -Dtest=MultipleTableFileSourceSplitEnumeratorTest test
./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-base -DfailIfNoTests=false package
./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-sftp -am -DskipTests package
```

### Check list


DanielLeens (commented):
Thanks for the contribution! I reviewed the current head and the lifecycle change looks well scoped. I do not see a source-level blocker from Daniel’s side; the remaining gate is the queued Build check.

# What this PR fixes
- User pain point: SFTP binary reads can enumerate remote files too early while the source config is being constructed.
- Fix approach: defer file discovery for SFTP `binary` + `ONCE` mode and move the actual listing to the split enumerator open phase.
- In one sentence: this moves remote file discovery to the source split lifecycle stage where it belongs.

# Code path checked
```text
SFTP binary source initialization
  -> SFTPFileSourceConfig.shouldDeferFileDiscovery() [47]
      -> file_format_type == BINARY && discovery_mode == ONCE
  -> BaseFileSourceConfig.parseFilePaths()
      -> deferred=true, filePaths=empty
  -> MultipleTableFileSourceSplitEnumerator.open() [85]
      -> getFilePathsForSplitEnumerator() [91]
          -> discoverFilePaths() lists SFTP files
      -> FileSplitStrategy.split(tableId, filePath)
  -> assignSplit() distributes splits by parallelism
```

# Correctness / compatibility
The normal SFTP binary once path hits this change. Other file connectors keep the old eager behavior because the base hook defaults to `false`. I also checked production call sites for `getFilePaths`; the split enumerator now uses `getFilePathsForSplitEnumerator()`, and the remaining direct getter usages are test-side checks.

This is compatible with existing configs and does not change option names, defaults, protocol, serialization, or checkpoint state. The sensitive part of the path logging is masked through URI user-info handling.

# Testing and stability
The added unit coverage checks deferred construction, eager legacy behavior, and enumerator-open discovery. The test uses an in-memory counting filesystem, with no fixed ports, sleeps, external services, or timing-sensitive assertions. Stability rating: stable.

I did not run local tests; this review is based on the local diff, call-chain inspection, and test stability analysis.

# Issues
No blocking source issues found.

# Merge conclusion
Blocking items: none from the source review.
Non-blocking suggestions: none.

Overall this is a precise lifecycle fix. Current Build is still queued: https://github.com/apache/seatunnel/runs/85806781950. Once the required checks finish successfully, I’m happy to approve.

DanielLeens (commented):
Thanks for the follow-up. I re-reviewed the latest head `e9763a2eee49` from scratch and traced the SFTP binary-source path again.

# What this PR fixes
- User pain: in SFTP `binary` + `ONCE` mode, file discovery currently happens too early during source-config construction, before the split enumerator lifecycle owns that work.
- Fix approach: defer binary file discovery for that specific mode, keep eager discovery for the existing paths, and let the split enumerator resolve file paths on `open()`.
- In one sentence: this moves remote file listing into the split-enumerator stage where it belongs.

# Runtime path checked
```text
SFTP binary source initialization
  -> SFTPFileSourceConfig.shouldDeferFileDiscovery()
      -> file_format_type == BINARY && discovery_mode == ONCE
  -> BaseFileSourceConfig.parseFilePaths()
      -> deferred=true, config-time file list stays empty
  -> MultipleTableFileSourceSplitEnumerator.open()
      -> getFilePathsForSplitEnumerator()
          -> discoverFilePaths()
      -> split file paths into FileSourceSplit entries
```

# Current findings
- The deferred/eager split is still scoped correctly on the latest head.
- The new tests cover both modes directly and avoid sleep-based timing, so the regression coverage looks stable.
- I do not see a reopened source-level blocker on this head.

# Merge conclusion
Source-level: no blocker from Daniel's side on the current head.

Current gate snapshot:
- `Build`: FAILURE https://github.com/apache/seatunnel/runs/86273422264
- linked test-results summary currently points to the contributor fork run: https://github.com/goutamadwant/seatunnel/actions/runs/29064610309/job/86273404798
- branch/base relation: `compare=diverged`, `behind_by=1`

Given the branch is still behind the latest `dev` and the current failure signal is not clearly attributable to this small SFTP diff yet, the lowest-cost next step is to sync with the latest `dev` and rerun `Build` first. If the same lane still fails after the sync, please drop the refreshed run link here and I can narrow it down on top of the updated head.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `e60a27f84baa` from the SFTP/file-source lifecycle again.

The new commit after my previous review is a merge from latest `dev`; the extra diff since the last reviewed head is upstream Hudi/Tablestore docs, so the SFTP/file-source code itself is unchanged. I still checked the full current PR-owned diff and test path again.

## What this PR fixes
- User pain: SFTP `binary` + `ONCE` mode can enumerate remote files during source-config construction, before the split enumerator lifecycle owns that work.
- Fix approach: defer file discovery only for SFTP `binary` + `ONCE`, and resolve file paths inside `MultipleTableFileSourceSplitEnumerator.open()`.
- One-line summary: remote file listing moves to the split-enumerator stage while all other file-source paths keep the old eager behavior.

## Runtime path checked
```text
SFTP binary source initialization
  -> SFTPFileSourceConfig.shouldDeferFileDiscovery()
      -> file_format_type == BINARY && discovery_mode == ONCE
  -> BaseFileSourceConfig.parseFilePaths()
      -> deferred=true, config-time file list stays empty
  -> MultipleTableFileSourceSplitEnumerator.open()
      -> getFilePathsForSplitEnumerator()
          -> discoverFilePaths()
      -> FileSplitStrategy.split(tableId, filePath)
  -> run()/assignSplit(taskId)
      -> context.assignSplit()
      -> context.signalNoMoreSplits()
```

## Findings
No source-level blockers from my side on the current head.

Compatibility: config names/defaults, checkpoint state, split assignment, and non-SFTP connector behavior stay compatible. The only intended behavior change is the discovery timing for SFTP `binary` + `ONCE`.

Testing/stability: the added unit tests cover deferred construction, existing eager behavior, and enumerator-open discovery through an in-memory counting filesystem. No fixed sleeps, ports, external services, or timing-sensitive assertions. Stability rating: stable.

I did not run local tests; this review is based on local diff inspection, call-chain tracing, existing review context, and the current GitHub check state.

## Merge conclusion

1. Blocking items
- None from source review.

2. Non-blocking suggestions
- None.

Current gate: `Build` is still queued: https://github.com/apache/seatunnel/runs/86567859771. I am not approving before the required check finishes successfully, but the source side looks clear to me.

DanielLeens (commented):
Rechecked the latest head: I still do not see a new Daniel-side source blocker, but the current remaining gate is a failed `Build` check.

The failed fork workflow currently shows:
- `Dead links`
- `connector-file-sftp-it (11, ubuntu-latest)`

Because this PR touches SFTP file discovery, the `connector-file-sftp-it` failure is worth treating as relevant until a rerun proves otherwise. This branch is also diverged from latest `dev` (`behind_by=1`), so I would sync `dev` and rerun Build first; if `connector-file-sftp-it` still fails after that, please focus the next fix on that lane rather than unrelated CI changes.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `9b93e6b0604a` from the current `dev` merge-base and traced the SFTP/file-source lifecycle again.

The new commit is a `dev` merge; the PR-owned file-source diff is unchanged in substance from the version I previously reviewed.

Runtime chain checked:
```text
SFTP source config creation
  -> SFTPFileSourceConfig.shouldDeferFileDiscovery(...)
      -> true only for binary + discovery_mode=ONCE
  -> BaseFileSourceConfig constructor
      -> parseFilePaths(...) returns empty when discovery is deferred
      -> binary catalog schema still uses getSchemaForEmptyFilePath(...)

Split enumeration
  -> MultipleTableFileSourceSplitEnumerator.open()
      -> iterates BaseFileSourceConfig objects
      -> getFilePathsForSplitEnumerator()
          -> deferred SFTP binary path discovers remote files here
          -> eager/continuous paths keep existing behavior
      -> FileSplitStrategy.split(tableId, filePath)
```

Key findings:
- The normal SFTP binary `ONCE` path now avoids remote file listing during config construction and performs discovery in enumerator `open()`, which is the intended lifecycle shift.
- Non-deferred file sources keep the existing eager file list behavior.
- The new tests use a deterministic in-memory counting filesystem and direct assertions; I do not see fixed sleeps, external network resources, or order-sensitive assertions.
- I do not see a new Daniel-side source blocker on this head.

### Merge conclusion: source clear, waiting for CI
Suggested follow-ups: wait for the queued `Build` to finish: https://github.com/apache/seatunnel/runs/86625884456.

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current diff and the relevant deferred SFTP binary discovery runtime path.

The dedicated discovery strategy is closed in `finally`, tests cover the lifecycle, and `Build` is green. The remaining gate is repository review/maintainer approval.

goutamadwant:
@DanielLeens @davidzollo does this look good to be merged as builds look green too?

DanielLeens:
Thanks for checking. I re-read the live gate on the current head: the required `Build` is green, and I do not see a reopened source-level blocker from Daniel's review.

The remaining step is repository reviewer/maintainer approval. Daniel has READ permission here, so a write-capable maintainer still needs to complete that formal gate before merge.

davidzollo (approved):
+1
LGTM

PR #11380, opened 2026-07-08
[Feature][Connector-V2][PostgreSQL-CDC] Add committed-offset and snapshot-only startup modes

### LINKED ISSUE #11039: [Feature][Connector-V2][PostgreSQL-CDC] Add committed-offset and snapshot-only startup modes
## Background
SeaTunnel's `PostgreSQL-CDC` startup-mode surface is still missing two important operational modes:
1. **committed-offset startup**: begin from the replication slot's committed LSN position
2. **snapshot-only startup**: read snapshot data and finish as a bounded job without entering WAL streaming

These two modes belong together because they affect the same connector-owned startup option surface, the same offset / restore path, and the same bounded-vs-unbounded runtime contract.

## Scope
This issue covers exactly these two additions for `PostgreSQL-CDC`:
- `committed-offset` startup
- `snapshot-only` startup

It does **not** include Oracle startup work, dynamic newly-added table discovery, schema evolution policy work, or metadata enrichment.

## First delivery boundary
The first delivery should keep the scope narrow:
- add the two startup modes to PostgreSQL only
- preserve existing snapshot and WAL reader logic as much as possible
- do not combine this with broader PG base refactors unless they are strictly required

If PG base abstraction work is in progress elsewhere, this issue should still be implementable either on top of that work or as a small delta against the current PostgreSQL connector.

### 1. Extend connector-owned startup option parsing
Primary touch points are expected to include:
- `PostgresSourceOptions`
- `PostgresIncrementalSourceFactory`
- `PostgresIncrementalSource`
- possibly the shared `StartupMode` enum if the current shared model is too narrow

The new modes should be explicit SeaTunnel connector options, not implicit Debezium passthrough behavior.

### 2. Committed-offset startup
For committed-offset mode:
- resolve the initial LSN from the configured replication slot's committed position
- skip snapshot entirely
- require an explicit replication slot name in the first delivery
- fail fast if the slot does not exist, has no usable committed LSN, or the runtime assumptions do not allow a safe committed-offset start

Expected implementation touch points likely include:
- `PostgresSourceConfigFactory`
- `LsnOffsetFactory`
- PostgreSQL source startup initialization path in `PostgresIncrementalSource`

### 3. Snapshot-only startup
For snapshot-only mode:
- reuse the current snapshot planning and snapshot reader path
- do **not** enter WAL streaming after the last snapshot split completes
- expose bounded completion semantics so the job ends naturally instead of remaining in an idle streaming state

Expected implementation touch points likely include:
- `PostgresIncrementalSource`
- enumerator / reader completion behavior
- boundedness reporting inherited from the base source path

### 4. Recovery and serialization
Both new modes must participate in the normal checkpoint / restore path:
- startup mode must survive serialization and restore
- committed-offset startup must restore into the correct LSN-based position semantics
- snapshot-only jobs must still finish cleanly after restore when snapshot work is complete

### 5. Validation contract
Validation should be explicit for contributors and users:
- `committed-offset` requires a configured slot name
- if the slot cannot provide a usable committed LSN, startup fails instead of silently falling back to another mode
- `snapshot-only` must reject incompatible combinations with streaming stop-mode options if those combinations would make semantics ambiguous

### 6. Documentation and tests
Suggested test layers:
- option parsing and validation tests
- offset serialization / restore tests for committed-offset mode
- bounded completion tests for snapshot-only mode
- at least one runnable validation path for PostgreSQL startup behavior

## Suggested config surface
Illustrative target shape:

```hocon
startup.mode = "committed-offset"
startup.mode = "snapshot"
```

Exact naming can be refined during implementation, but semantics must be explicit and documented.

## Acceptance criteria
- `PostgreSQL-CDC` supports committed-offset startup.
- `PostgreSQL-CDC` supports snapshot-only startup.
- `committed-offset` has explicit slot-based validation rules.
- Validation fails fast for unsupported or incompatible configuration.
- Checkpoint / restore behavior remains correct for both modes.
- Snapshot-only mode finishes as a bounded job.
- Add focused tests and at least one runnable validation path.

## Non-goals
- Oracle SCN startup.
- Schema evolution policy modes.
- New CDC metadata fields.

CloverDew:

Close: https://github.com/apache/seatunnel/issues/11039

This PR adds two PostgreSQL CDC startup modes:
- `committed-offset`: starts WAL streaming from the configured replication slot's committed LSN and skips snapshot reading.
- `snapshot-only`: reads the startup snapshot and finishes as a bounded job without entering WAL streaming.

Yes.
PostgreSQL CDC now accepts two additional `startup.mode` values:
- `snapshot-only`
  - Previous behavior: PostgreSQL CDC snapshot startup always transitioned into WAL streaming.
  - New behavior: the job reads snapshot data and then finishes naturally as a bounded job.
- `committed-offset`
  - Previous behavior: PostgreSQL CDC supported `initial`, `earliest`, and `latest`, but could not explicitly start from the committed position of an existing logical replication slot.
  - New behavior: the job starts from the configured replication slot's `confirmed_flush_lsn`, skips snapshot data, and fails fast if `slot.name` is missing, the slot does not exist, or the slot has no usable committed LSN.

Added PostgreSQL CDC E2E coverage for the new startup modes:
- `snapshot-only` verifies that snapshot data is written and later source changes are not streamed.
- `committed-offset` verifies that the connector attaches to an existing logical replication slot and does not snapshot existing rows.

### Check list


DanielLeens (approved):
Thanks for the contribution! I re-reviewed the latest head `f714fc533e89` from scratch and traced the new startup modes through the real PostgreSQL CDC code path. The implementation looks solid to me overall, and I did not find a new source-side blocker in the current version.

# What this PR fixes
- User pain: PostgreSQL-CDC was still missing two practical startup modes that people often need operationally: a bounded snapshot-only run, and a slot-based startup from an already committed replication position.
- Fix approach: extend the PostgreSQL CDC startup-mode surface with `snapshot-only` and `committed-offset`, wire them into startup offset resolution / boundedness / split assignment, and add E2E + docs coverage.
- One-line summary: the new modes are implemented as explicit SeaTunnel behaviors rather than hidden Debezium passthrough, and the main source/enumerator/reader path is aligned with the documented semantics.

# Full call path I checked
```text
snapshot-only
  -> startup.mode = snapshot-only
  -> PostgresSourceOptions.STARTUP_MODE accepts SNAPSHOT_ONLY
  -> StartupConfig.getStartupOffset() returns null
  -> IncrementalSource.getBoundedness() => BOUNDED
  -> IncrementalSource.createEnumerator()
      -> SnapshotOnlySplitAssigner
      -> snapshot splits only
  -> PostgresSourceFetchTaskContext.configure()
      -> snapshotter does not enter WAL streaming
  -> job finishes naturally after the snapshot phase

committed-offset
  -> startup.mode = committed-offset
  -> PostgresIncrementalSource.validateStartupOptions()
      -> require explicit slot.name
  -> StartupConfig.getStartupOffset()
      -> offsetFactory.committedOffset()
  -> LsnOffsetFactory.committedOffset()
      -> SELECT confirmed_flush_lsn::text FROM pg_replication_slots WHERE slot_name = ?
      -> build LsnOffset from the slot's committed LSN
  -> IncrementalSplitAssigner.createIncrementalSplit()
      -> use that LSN as the incremental split startup offset
  -> PostgresSourceFetchTaskContext.loadStartingOffsetState()
      -> rebuild PostgresOffsetContext from the startup offset
  -> PostgresWalFetchTask.execute()
      -> Debezium streaming starts from that WAL position
```

# Findings
I did not find a new blocker in the current version.

The key points I explicitly cross-checked were:
- `snapshot-only` really becomes a bounded source (`IncrementalSource.getBoundedness()` + `SnapshotOnlySplitAssigner`)
- `committed-offset` really comes from the replication slot's committed LSN instead of a snapshot-side workaround
- the fetch context no longer tries to create a slot when the slot already exists and the job is supposed to reuse it
- both EN and ZH docs were updated to describe the new modes consistently

# Test coverage / stability note
For the test-stability pass, I would rate the new PostgreSQL CDC E2E coverage as **stable**. The new test structure does not depend on fixed ports, shared mutable static test state, or timing-only assertions for job completion. The replication-slot setup/cleanup is also handled deliberately, which reduces cross-test contamination in the shared Postgres container.

One small non-blocking follow-up I would still consider: the `committed-offset` E2E currently proves that existing rows are not snapshotted again, which is good, but it could be even stronger if it also inserted one new row after the job starts and asserted that the new WAL event reaches the sink. That would prove both halves of the contract in one place.

# Review conclusion
### Conclusion: can merge

1. Blocking items
- None from my side.

2. Non-blocking suggestions
- Consider extending the `committed-offset` E2E with one post-start insert so the test verifies both "skip historical rows" and "continue streaming new WAL changes".

Overall this is a clean, focused PostgreSQL-CDC feature addition. The startup-mode parsing, offset path, boundedness behavior, docs, and E2E coverage are aligned well enough for me to approve.

DanielLeens (approved):

I also recorded the base/head sync fact for the queue. This branch is still `diverged` with `behind_by=11`, but because the current required Build signal is green, I am not asking for a sync-only retry from Daniel's side here. If GitHub's merge gate later asks for an update branch step, please handle that as a merge-gate action.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=2`, `behind_by=19`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

DanielLeens (changes requested):
Sorry for missing these points in my earlier reviews - that's on me. I re-reviewed the full latest head `f714fc533e89` from scratch, including the SeaTunnel enumerator/reader lifecycle, Debezium 1.9.8 snapshotter behavior, replication-slot ownership, and the updated E2E. My earlier approval and today's source-cleared comment are not the right technical conclusion for this revision.

Thank you for taking on these two useful PostgreSQL CDC startup modes. The overall direction is good: `snapshot-only` is bounded at the SeaTunnel enumerator layer, and `committed-offset` resolves the slot's `confirmed_flush_lsn`. However, I found two source-level blockers that need to be closed before merging.

## Full runtime path checked

```text
snapshot-only
  -> PostgresSourceOptions.STARTUP_MODE
  -> IncrementalSource.getBoundedness() returns BOUNDED
  -> SnapshotOnlySplitAssigner emits snapshot splits only
  -> reader receives a SnapshotSplit
  -> PostgresSourceConfigFactory.create()
      -> does not map SeaTunnel snapshot-only to Debezium snapshot.mode=initial_only
      -> Debezium defaults to snapshot.mode=initial
  -> PostgresSourceFetchTaskContext.configure()
      -> InitialSnapshotter.shouldStream() returns true
      -> creates a replication connection
      -> creates the logical replication slot when it does not exist
  -> snapshot finishes and the enumerator signals NoMoreSplits
  -> bounded job exits
  -> slot.drop.on.stop defaults to false
  -> the slot remains without a WAL consumer advancing confirmed_flush_lsn

committed-offset
  -> validates that slot.name is explicitly configured
  -> LsnOffsetFactory.committedOffset()
      -> reads pg_replication_slots.confirmed_flush_lsn
      -> builds the Debezium PostgreSQL offset
  -> IncrementalSplitAssigner creates the incremental split
  -> PostgresSourceFetchTaskContext rebuilds PostgresOffsetContext
  -> PostgresWalFetchTask
  -> Debezium PostgresStreamingChangeEventSource
      -> starts from the committed WAL position
      -> locates the transaction boundary
      -> reads WAL events
      -> advances the committed offset after checkpoint completion
```

## Blocking issue 1: `snapshot-only` still creates and retains a logical replication slot

- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/config/PostgresSourceConfigFactory.java:58`, and `PostgresSourceFetchTaskContext.java:133,197-210`
- What happens: the new mode stops incremental split assignment only at the SeaTunnel layer. The reader still uses Debezium's default `snapshot.mode=initial`; `InitialSnapshotter.shouldStream()` is `true`, so `configure()` enters replication setup and creates the slot if it is absent.
- Risk: after the bounded job exits, the slot is not dropped by default and no reader advances it. PostgreSQL can keep retaining WAL for that slot until disk usage becomes operationally dangerous. It can also collide with the default slot name and unnecessarily require replication privileges for a snapshot-only job.
- Recommended fix: preferably make `PostgresSourceFetchTaskContext` explicitly skip replication setup for `StartupMode.SNAPSHOT_ONLY`. Alternatively, force Debezium `snapshot.mode=initial_only` as an internal contract after user Debezium properties are applied. Please add a regression assertion proving that the job creates no active or retained slot after completion.

## Blocking issue 2: the latest E2E can pass even when committed-offset WAL streaming is broken

- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-postgres-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/PostgresCDCIT.java:419,449-475`
- What happens: the latest `fix ci` revision removed the earlier positive WAL row/assertion. The test now checks only that historical rows `10` and `11` are absent. `untilAsserted(isEmpty())` can succeed on its first empty observation, and the `CompletableFuture` is neither retained nor joined, so a background job failure does not fail the test.
- Risk: CI stays green even if offset reconstruction, WAL startup, incremental delivery, or the async job itself is broken. The new `SnapshotPhaseState` restore branch is also not exercised by checkpoint/savepoint recovery.
- Recommended fix: create the slot and committed position, write at least one event after that position, start the job, then assert post-start insert/update/delete delivery. Please retain and join the future so async failures propagate. Also add focused slot-missing/null-LSN coverage and a checkpoint/restore test for the new snapshot-only state path.

Test stability rating: **high risk**, because the current committed-offset E2E has a false-positive path rather than merely a timing-related flaky risk.

## Review conclusion

### Conclusion: changes are required before merge

1. Blocking items
- Make `snapshot-only` avoid creating or retaining a PostgreSQL replication slot, and verify the external resource lifecycle.
- Restore a positive committed-offset WAL assertion, propagate async failures, and cover the newly introduced checkpoint/restore path.

2. Non-blocking suggestions
- Replace the fixed five-second sleep in the snapshot-only E2E with a condition tied to job completion and slot state.
- Add a compatibility test for the shared `StartupMode.toString()` lowercase/hyphen behavior, or keep that conversion scoped to configuration parsing.

The current CI checks are green, which is helpful, but issue 2 is precisely why that green result does not clear the source blockers yet. The branch is also currently diverged from `dev` (`ahead_by=2`, `behind_by=20`); I am recording that as a merge-state fact, not asking for a sync-only retry, because it is not the cause of the two issues above.

The feature is worth continuing, and most of the structure is already in place. Once these two lifecycle and verification gaps are fixed, I will be happy to re-review the complete updated head promptly. Thanks again for the work and for your patience with the correction.

DanielLeens:
Rechecked the current head `751dec8ca244` against the latest `dev` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
snapshot-only
  -> PostgresSourceOptions.STARTUP_MODE
  -> IncrementalSource.getBoundedness() returns BOUNDED
  -> SnapshotOnlySplitAssigner emits snapshot splits only
  -> reader receives a SnapshotSplit
  -> PostgresSourceConfigFactory.create()
      -> does not map SeaTunnel snapshot-only to Debezium snapshot.mode=initial_only
      -> Debezium defaults to snapshot.mode=initial
  -> PostgresSourceFetchTaskContext.configure()
      -> InitialSnapshotter.shouldStream() returns true
      -> creates a replication connection
      -> creates the logical replication slot when it does not exist
  -> snapshot finishes and the enumerator signals NoMoreSplits
  -> bounded job exits
  -> slot.drop.on.stop defaults to false
  -> the slot remains without a WAL consumer advancing confirmed_flush_lsn

committed-offset
  -> validates that slot.name is explicitly configured
  -> LsnOffsetFactory.committedOffset()
      -> reads pg_replication_slots.confirmed_flush_lsn
      -> builds the Debezium PostgreSQL offset
  -> IncrementalSplitAssigner creates the incremental split
  -> PostgresSourceFetchTaskContext rebuilds PostgresOffsetContext
  -> PostgresWalFetchTask
  -> Debezium PostgresStreamingChangeEventSource
      -> starts from the committed WAL position
      -> locates the transaction boundary
      -> reads WAL events
      -> advances the committed offset after checkpoint completion
```

## Merge conclusion

1. Blocking items
- Separately, this branch is behind upstream (`compare status: diverged`, `ahead_by=3`, `behind_by=20`), and the current failing gate is `FAILURE gate: Build`: https://github.com/apache/seatunnel/runs/86998185983

2. Lowest-cost next step

DanielLeens:
Rechecked the latest head again from the PostgreSQL-CDC startup-mode path.

The earlier Daniel-side source concerns I had about `snapshot-only` and the regression proof are materially addressed for me on this revision. At this point I do not see a reopened source blocker in the owned startup-mode logic itself.

This branch is still quite stale against `dev`, though (`behind_by=22`), and the current failing lanes are not clean enough yet to separate "old branch drift" from a real new PR regression. So the lowest-cost next step here is still to sync the latest `dev` first and rerun `Build` on top of the refreshed head. If the same lanes stay red after that sync, please paste the updated run link and I can narrow down the remaining failure against the newer base.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `eb91517193c0` from scratch.

# What this PR fixes
- User pain: PostgreSQL CDC users could not attach to an existing replication slot's committed position (forcing re-snapshots or manual LSN juggling) and had no way to run a bounded snapshot-only job, unlike what MySQL CDC already offers (#11039).
- Fix approach: extend `startup.mode` with `committed-offset` (resolve `confirmed_flush_lsn` of the configured `slot.name` via `LsnOffsetFactory` and skip the snapshot phase) and `snapshot-only` (run the snapshot assigner to completion and finish as a bounded job), with fail-fast validation, docs (en/zh) and E2E coverage for both modes.
- One-line summary: the feature works end-to-end and is well tested, but before merging I want the committed-offset resolution pinned at first startup only (not re-derived from `confirmed_flush_lsn` on checkpoint restore), the `LsnOffset` signature change checked for binary/state compatibility, NULL-LSN and slot-dropped races plus JDBC resource cleanup hardened in `LsnOffsetFactory`, and the zh FAQ brought to parity with the en note about the slot needing to pre-exist.

# Runtime chain I rechecked
```text
PostgresIncrementalSourceFactory.createSource() -> PostgresIncrementalSource.getStartupConfig()  PostgresIncrementalSource.java:78
  PostgresIncrementalSource.getStartupConfig() validates COMMITTED_OFFSET requires explicit slot.name  PostgresIncrementalSource.java:96
IncrementalSource.createEnumerator() -> createSplitAssigner()  IncrementalSource.java:198
  startupConfig.getStartupOffset(offsetFactory)  StartupConfig.java:57
    case COMMITTED_OFFSET -> LsnOffsetFactory.committedOffset()  LsnOffsetFactory.java:74
      LsnOffsetFactory.committedOffset() -> PostgresConnection.query("SELECT confirmed_flush_lsn FROM pg_replication_slots WHERE slot_name = ?")  LsnOffsetFactory.java:88   <-- must fail fast on missing slot / NULL LSN, and must NOT be re-invoked on checkpoint restore
    case SNAPSHOT_ONLY -> falls through to snapshot assignment with a bounded stop config  StartupConfig.java:63
  IncrementalSplitAssigner.createIncrementalSplits() embeds resolved LsnOffset as startupOffset  IncrementalSplitAssigner.java:212
snapshot-only path:
  IncrementalSource.getBoundedness() -> BOUNDED when StopMode == SNAPSHOT_ONLY  IncrementalSource.java:142
  HybridSplitAssigner.noMoreSplits() returns true after SnapshotSplitAssigner completes, skipping WAL streaming splits  HybridSplitAssigner.java:131
restore path (the one I re-checked for the HIGH finding):
  IncrementalSourceEnumerator restored from PendingSplitsState -> IncrementalSplit.getStartupOffset() from checkpoint state, bypassing StartupConfig.getStartupOffset()  IncrementalSourceEnumerator.java:105
  PostgresSourceFetchTaskContext.getStreamOffset() deserializes LsnOffset from split state  PostgresSourceFetchTaskContext.java:167   <-- LsnOffset field/ctor change must round-trip against pre-existing serialized state
```

# Findings
**Issue 1: committed-offset startup offset must not be re-resolved from confirmed_flush_lsn on checkpoint restore — risk of data loss on recovery**
- Location: `docs/en/connectors/source/PostgreSQL-CDC.md:254`
- Why it matters: - Root cause: the new `committed-offset` mode documents that SeaTunnel "uses its `confirmed_flush_lsn` as the startup offset". `confirmed_flush_lsn` is a moving target: Debezium confirms flushed LSNs to the server while streaming, so it advances during the job's lifetime.
- Impact: if the offset resolution path added in `IncrementalSource`/`LsnOffsetFactory` is executed again when the enumerator is reconstructed from a checkpoint (Zeta job restart, Flink failover), the job would jump forward to the *current* `confirmed_flush_lsn` instead of the checkpointed offset, silently skipping every WAL event between the last checkpoint and the last server-side flush confirmation. That is data loss on recovery, the exact scenario checkpointing exists to prevent.
- Fix: ensure the committed-offset lookup runs only on a fresh start (i.e., in `IncrementalSource#createEnumerator` when there is no restored `PendingSplitsState`), and that `restoreEnumerator` always uses the checkpointed stream split offsets. Add an E2E case that kills and restores the committed-offset job mid-stream and asserts no rows are missing.
- Evidence: `docs/en/connectors/source/PostgreSQL-CDC.md:254  uses its `confirmed_flush_lsn` as the startup offset.`
- Suggested fix: Since `confirmed_flush_lsn` advances while the job runs (Debezium confirms flushed LSNs back to the server), the committed-offset lookup must only happen on a *fresh* start. Please confirm the resolution path in `IncrementalSource`/`LsnOffsetFactory` is skipped when the enumerator is restored from checkpoint state — re-querying `confirmed_flush_lsn` on failover would fast-forward past unread WAL and silently drop events. An E2E test that restarts the committed-offset job from a checkpoint and verifies no missing rows would lock this in.
- Severity: High

**Issue 2: Possible public API change in seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/source/offset/LsnOffset.java**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/source/offset/LsnOffset.java`
- Severity: High

**Issue 3: committed-offset validation must handle NULL confirmed_flush_lsn, slot-dropped races, and close JDBC resources in LsnOffsetFactory**
- Location: `docs/en/connectors/source/PostgreSQL-CDC.md:100`
- Why it matters: - Root cause: the docs state the mode "fails if the slot does not exist or has no usable committed LSN". On PostgreSQL, a freshly created logical slot (or a slot on a standby) can have `confirmed_flush_lsn = NULL` in `pg_replication_slots`, so the query result can be a row with a null column, not just an empty result set — a naive `resultSet.getString(...)`/`LogSequenceNumber.valueOf(null)` path throws an unhelpful NPE instead of the documented fail-fast error.
- There is also a TOCTOU window: the slot can be dropped (or grabbed by another consumer, since logical slots are single-consumer) between the validation query in `LsnOffsetFactory` and Debezium attaching to it, so the startup error surfaces later from inside the Debezium engine with a much less actionable message.
- Impact: users get NPEs or opaque Debezium replication errors instead of the clear "slot has no usable committed LSN" failure the docs promise.
- Fix: in the new `LsnOffsetFactory` lookup, use try-with-resources for the `Connection`/`Statement`/`ResultSet`, explicitly check `rs.wasNull()` / empty result and throw a `SeaTunnelRuntimeException` with the slot name and remediation hint, and add a unit test for the NULL-`confirmed_flush_lsn` case alongside `LsnOffsetFactoryTest`.
- Suggested fix: The documented fail-fast behavior needs to cover the case where the slot row exists but `confirmed_flush_lsn` is NULL (a freshly created slot never flushed to). In `LsnOffsetFactory`, please make sure the lookup uses try-with-resources for the JDBC `Connection`/`Statement`/`ResultSet`, checks `rs.wasNull()`, and throws a descriptive `SeaTunnelRuntimeException` naming the slot — otherwise users hit an NPE or a late, opaque Debezium error instead of this documented failure.
- Severity: Medium

**Issue 4: Chinese docs update the `startup.mode` option table but I don't see the corresponding FAQ note about `committed-offset` requiring a pre-existing replication slot; en/zh parity should be kept.**
- Location: `docs/zh/connectors/source/PostgreSQL-CDC.md:98`
- Why it matters: - Root cause: the English docs add a dedicated FAQ paragraph under "How are replication slots managed?" explaining that `committed-offset` requires the slot to already exist, while the visible zh change is confined to the option table row for `startup.mode`.
- Impact: Chinese-speaking users reading only the zh page get the enum values but miss the operational precondition (slot must pre-exist, job fails otherwise), which is exactly the kind of detail that generates avoidable support issues and failed first runs.
- Fix: mirror the two-line FAQ addition from `docs/en/connectors/source/PostgreSQL-CDC.md` into the corresponding replication-slot FAQ section of `docs/zh/connectors/source/PostgreSQL-CDC.md`; if it is already among the remaining zh additions not shown in this hunk, then this is covered.
- Evidence: `docs/en/connectors/source/PostgreSQL-CDC.md:253  When `startup.mode` is `committed-offset`, the replication slot must already exist because SeaTunnel`
- Suggested fix: The English page adds a FAQ note under "How are replication slots managed?" stating that `committed-offset` requires the replication slot to already exist (SeaTunnel reads its `confirmed_flush_lsn` as the startup offset). Please add the equivalent paragraph to the zh replication-slot FAQ section as well so both languages document this operational precondition consistently.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: committed-offset startup offset must not be re-resolved from confirmed_flush_lsn on checkpoint restore — risk of data loss on recovery
- Issue 2: Possible public API change in seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/source/offset/LsnOffset.java

- Issue 3: committed-offset validation must handle NULL confirmed_flush_lsn, slot-dropped races, and close JDBC resources in LsnOffsetFactory
- Issue 4: Chinese docs update the `startup.mode` option table but I don't see the corresponding FAQ note about `committed-offset` requiring a pre-existing replication slot; en/zh parity should be kept.

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current diff and the relevant PostgreSQL CDC committed-offset and snapshot-only paths runtime path.

The prior restore/NULL/resource/documentation findings are resolved or rejected with same-proposition evidence. The aggregate build was cancelled after a unit-test lane failed, so CI remains outstanding.

DanielLeens (commented):



The current Build is still pending; CI remains a separate gate.

DanielLeens (commented):
The CI fact from my previous review has changed: the required `Build` is now green on the unchanged head `e559b22c4e17`.

The source-level conclusion remains clear from Daniel's side; the previous restore, NULL-LSN, resource-lifecycle, and documentation findings are resolved or dispositioned on this head. The remaining repository gate is reviewer/maintainer approval. Daniel has READ permission, so this follow-up records the current technical conclusion but does not replace a write-capable maintainer's approval step.

This branch is currently `diverged` from `dev` (ahead_by=5, behind_by=4). I am recording that merge-state fact, but I am not requesting a sync-only rerun because the current required Build is green.

DanielLeens (commented):
Rechecked the latest head: the source-side changes are still clear from my side, and the required checks are now green. The remaining gate is that GitHub is currently reporting merge conflicts with the latest `dev` (`compare status: diverged`, `behind_by=23`, `mergeable_state=dirty`).

Could you please sync this branch with the latest `dev`, resolve the conflicts, and rerun the checks once that is done? Since I only have read-level review controls here, a write-capable maintainer may still need to complete the final approval step after the branch is back in sync.

DanielLeens (commented):
Thanks for the update. I re-reviewed the current head from the PostgreSQL startup-mode option through offset resolution and split-assignment restore.

The earlier findings are closed: committed-offset handles a missing/NULL LSN explicitly and closes JDBC resources, snapshot-only completion is preserved in enumerator state across restore, and the tests cover restored assignment. The current branch is conflict-free, and the required Build is green.

I do not see a source-level blocker. The remaining gate is one maintainer review. Daniel has read-only permission, so this comment records the technical conclusion without changing GitHub's approval state.

davidzollo (approved):
+1
LGTM

PR #11389, opened 2026-07-09
[Fix][Zeta] Refresh job logs when job id changes

zhaoysg:
### Purpose
Refresh job log data when the selected job changes.

### Changes
- Watch `jobId` and reload logs immediately and on later prop changes.
- Remove the unused `NSpace` import.

### Tests
- `git diff --check`
- `npm.cmd run type-check` was attempted locally, but `vue-tsc` is unavailable because frontend dependencies are not installed in `seatunnel-engine-ui/node_modules`.


DanielLeens (commented):
Thanks for working on this. I re-reviewed the latest head `f7767c105fbd` from scratch.

# What this PR fixes
- User pain: on the Zeta job detail page, the `JobLog` panel can miss the real job logs because the component tries to load logs before the parent page has finished populating `job.jobId`.
- Fix approach: replace the one-shot `getJobLogs(props.jobId)` call with a `watch(() => props.jobId, ..., { immediate: true })`, so the log request runs again once the prop becomes available.
- One-line summary: this is a small but accurate lifecycle fix for the job log panel, and I do not see a new source-side blocker in the current head.

# Full call path I checked
```text
Route /jobs/:jobId
  -> views/jobs/detail.tsx
      -> getJobInfo(jobId) [detail.tsx:61-63]
      -> Object.assign(job, res)
      -> <JobLog jobId={job.jobId}> [detail.tsx:574]

JobLog component
  -> components/job-log/index.tsx
      -> old code: getJobLogs(props.jobId) once during setup
      -> new code: watch(() => props.jobId, ...) with immediate=true
      -> getJobLogs(`/logs/${jobId}?format=json`)
      -> render NCollapse items + iframe log links
```

I also checked the route lifecycle: the main layout renders `<router-view key={route.fullPath}>` in `layouts/main/index.tsx:66`, so a full path change remounts the page. That means the real bug here is the async "prop becomes valid after mount" path inside the same detail page, and this patch addresses exactly that path.

# Findings
I did not find a new code-side blocker in the current version.

The only follow-up I would suggest is adding a small frontend regression test for the case where `JobLog` mounts with an empty/placeholder `jobId` and then receives the real value later, so this lifecycle bug stays covered in the future. I would treat that as a non-blocking improvement rather than a merge blocker for this patch.

# Review conclusion

1. Blocking items
- None from my side in the current source diff.

2. Non-blocking suggestion
- Add a focused regression test for the delayed-`jobId` update path in `JobLog`.

Overall, the fix looks correct to me on the actual parent/child lifecycle path. The remaining gate right now is CI: the required `Build` check is still queued at `https://github.com/apache/seatunnel/runs/86071083921`, so I am not turning this into an approval yet. Once the required checks are green, this looks good from the source side.

DanielLeens:
Rechecked the current unchanged head `f7767c105fbd` again.

My earlier source-side blocker on this same revision still stands from Daniel's side. Separately, this branch is also behind the latest `dev` (`compare status: diverged`, `behind_by=3`), and the current red `Build` signal does not look actionable on top of this stale head yet.

https://github.com/apache/seatunnel/actions/runs/29003968170/job/86070979674

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `f7767c105fbd` from scratch.

# What this PR fixes
- User pain: The job log panel fetched logs only once on component setup, so switching to a different job left stale logs from the previously selected job on screen.
- Fix approach: Replace the one-shot `getJobLogs(props.jobId)` call with a `watch` on `props.jobId` (with `immediate: true` and a falsy-guard) so logs reload both on mount and on every subsequent jobId change; also drops the now-unused `NSpace` import.
- One-line summary: The fix is correct and idiomatic Vue, but the async fetch inside the watcher has no stale-response guard, so rapid jobId switches can race and render the wrong job's logs, and the change ships without any test or completed type-check.

# Runtime chain I rechecked
```text
JobLog component setup(props)  seatunnel-engine-ui/src/components/job-log/index.tsx:30
  watch(() => props.jobId, handler, { immediate: true })  index.tsx:32-39
    handler fires immediately on mount and again on each jobId prop change  index.tsx:34
      guard: if (!jobId) return  index.tsx:35
      getJobLogs(jobId)  index.tsx:36 -> seatunnel-engine-ui/src/service/job-log/index.ts (HTTP fetch of job logs)
        .then((res) => (logList.value = res))  index.tsx:36  <-- assignment is unconditional; a slow response for an old jobId can overwrite logs for the currently selected job
  render: logList.value mapped into NCollapse/NCollapseItem  index.tsx:41-48
```

# Findings
**Issue 1: Race condition: out-of-order responses when jobId changes quickly can display the wrong job's logs**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:36`
- Why it matters: - Root cause: the watcher fires `getJobLogs(jobId)` on every jobId change but never cancels or guards the previous in-flight request. If the user switches from job A to job B while A's request is still pending, and A's response arrives after B's, `logList.value = res` at seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:36 overwrites the correct data with stale logs.
- Impact: operators on the Zeta engine UI can see logs labeled for job B that actually belong to job A — a misleading result that is worse than an empty view, especially on slow REST endpoints or large log payloads.
- Fix: capture the current jobId in the closure and compare before assigning, e.g. `getJobLogs(jobId).then((res) => { if (jobId === props.jobId) logList.value = res })`, or use an incrementing request token / AbortController to invalidate superseded requests.
- Evidence: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:36  getJobLogs(jobId).then((res) => (logList.value = res))`
- Suggested fix: There is a stale-response race here: if `jobId` changes while a previous `getJobLogs` call is still in flight, the older response can resolve last and overwrite `logList` with the wrong job's logs. Please guard the assignment, e.g. `getJobLogs(jobId).then((res) => { if (jobId === props.jobId) logList.value = res })`, or track a request token so only the latest request wins.
- Severity: Medium

**Issue 2: No test or verified type-check for the reworked watcher logic; the fixed behavior is unguarded against regression**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:32`
- Why it matters: - The PR description states `vue-tsc` could not be run locally, so the only changed file has not even been type-checked, let alone covered by a component test.
- The exact bug being fixed (logs not refreshing on jobId change) is behavioral and easy to regress if someone later refactors the setup function back to a one-shot fetch.
- Fix: add a Vitest component test (the seatunnel-engine-ui module already uses Vue 3 tooling) that mounts the component, updates the `jobId` prop, and asserts `getJobLogs` is called again with the new id and that `{ immediate: true }` triggers the initial load; at minimum confirm `npm run type-check` and `npm run build` pass in CI before merge.
- Suggested fix: Since the PR notes `vue-tsc` could not be run locally, please make sure `npm run type-check`/`build` pass in CI, and ideally add a small Vitest test that mounts this component, changes the `jobId` prop, and asserts `getJobLogs` is re-invoked — that would lock in the refresh behavior this PR fixes.
- Severity: Low

# Review conclusion

- Issue 1: Race condition: out-of-order responses when jobId changes quickly can display the wrong job's logs
- Issue 2: No test or verified type-check for the reworked watcher logic; the fixed behavior is unguarded against regression

DanielLeens (commented):
Thanks @SEZ9 for the fresh review. I agree the stale-response race in the job-log watcher is the main code-side risk to address before this can move forward; it is aligned with the concern that the UI must not render logs from an older `jobId` after a fast switch.

No new commit has landed after that review yet, so I am not doing another full Daniel re-review in this round. Once the branch is updated, I will recheck the latest head.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `01945d6ef326` from scratch.

# What this PR fixes
- User pain: In the Zeta engine UI, the job log panel fetched logs only once at component setup, so switching to a different job kept showing the previous job's stale logs.
- Fix approach: Replace the one-shot `getJobLogs(props.jobId)` call in `setup()` with a `watch` on `() => props.jobId` (with `{ immediate: true }` and a falsy-id guard) that refetches and reassigns `logList`, plus removal of the now-unused `NSpace` import.
- One-line summary: The fix is small, correct for the reported symptom, and keeps the initial-load behavior via `immediate: true`, but the watcher has no request cancellation/staleness guard, so rapid jobId switches can race and render the wrong job's logs, and neither a component test nor a successful `vue-tsc` run backs the change.

# Findings
**Issue 1: Out-of-order responses can display the wrong job's logs when jobId changes rapidly**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:36`
- Why it matters: - Root cause: the `watch` callback fires a new `getJobLogs(jobId)` request on every jobId change, but the `.then` unconditionally writes into `logList.value`. If the user switches from job A to job B and A's (slower) response resolves after B's, the UI will show job A's logs under job B's view.
- Impact: operators inspecting Zeta job logs through the engine UI can be shown log content belonging to a different job than the one selected, which is misleading during incident triage and is a mild cross-job information-exposure in shared-console scenarios.
- Fix: guard the assignment with a staleness check, e.g. capture the requested id and only assign when it still matches the current prop: `getJobLogs(jobId).then((res) => { if (jobId === props.jobId) logList.value = res })`, or use an incrementing request token / AbortController to cancel superseded requests.
- Suggested fix: This `.then` assigns to `logList.value` unconditionally, so if the user switches jobs quickly, a slower response for the previous job can overwrite the logs of the currently selected job. Please add a staleness guard, e.g. `getJobLogs(jobId).then((res) => { if (jobId === props.jobId) logList.value = res })`, or track a request token/AbortController so only the latest request wins.
- Severity: Medium

**Issue 2: No component test for the new watch behavior; type-check was not actually run**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:32`
- Why it matters: - Root cause: the PR adds reactive re-fetch logic but ships no Vitest/`@vue/test-utils` coverage, and the description states `vue-tsc` could not be executed locally because `seatunnel-engine-ui/node_modules` was not installed, so even static type verification of the new `watch` signature is unconfirmed.
- Impact: regressions in this component (e.g. the race and stale-state issues flagged above, or a future refactor of `getJobLogs`) would go undetected; the engine-ui module already has a test setup that this change bypasses.
- Fix: add a component test that mocks `getJobLogs`, mounts the component, updates the `jobId` prop, and asserts a second fetch occurs and `logList` is replaced; also run `npm run type-check` in CI-equivalent conditions before merge.
- Evidence: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:33        () => props.jobId,`
- Suggested fix: Please add a small component test (mock `getJobLogs`, mount, change the `jobId` prop, assert a second fetch and list refresh) — this new `watch` is exactly the kind of reactive logic that regresses silently. Also please run `npm run type-check` with dependencies installed before merge, since it was skipped per the PR description.
- Severity: Low

**Issue 3: Consider watchEffect or explicit oldValue check to keep the watcher minimal**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:32  watch(`
- Why it matters: - Root cause: the watcher only depends on `props.jobId` and immediately uses it, which is exactly the pattern `watchEffect` covers with less boilerplate; alternatively the current `watch` is fine but should stay a single-source watcher.
- Impact: purely stylistic — no functional difference — but consolidating on one idiom keeps the seatunnel-engine-ui components consistent.
- Fix: either keep the current `watch(() => props.jobId, ..., { immediate: true })` (acceptable) or switch to `watchEffect(() => { const id = props.jobId; if (!id) return; ... })`; whichever is chosen, pair it with the staleness guard from the other comment.
- Evidence: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:38  { immediate: true }`
- Severity: Low

# Review conclusion

- Issue 1: Out-of-order responses can display the wrong job's logs when jobId changes rapidly
- Issue 2: No component test for the new watch behavior; type-check was not actually run
- Issue 3: Consider watchEffect or explicit oldValue check to keep the watcher minimal

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `01945d6ef326` against the actual Zeta UI job-detail lifecycle.

# What this PR fixes
- User pain: the JobLog component previously loaded logs only once in `setup()`, so when the parent detail page populated or changed `job.jobId` after mount, the log panel could keep the old or empty result.
- Fix approach: replace the one-shot `getJobLogs(props.jobId)` call with a `watch(() => props.jobId, ..., { immediate: true })`, so initial load and later `jobId` updates both trigger a log fetch.
- One-line summary: this is the correct Vue lifecycle fix for the delayed-`jobId` path.

# Runtime chain I rechecked
```text
Job detail route
  -> layouts/main/index.tsx renders router-view with key={route.fullPath}
      -> navigating to another job detail remounts the detail page
  -> views/jobs/detail.tsx
      -> getJobInfo(route job id)
      -> Object.assign(job, res)
      -> <JobLog jobId={job.jobId}>

JobLog component
  -> setup(props)
      -> watch(() => props.jobId, handler, { immediate: true })
      -> handler ignores empty jobId
      -> getJobLogs(jobId)
      -> logList.value = response
      -> render NCollapseItem / iframe log links
```

# Findings
I do not see a remaining source-level blocker on the latest head.

+1 to @SEZ9's point that guarding stale async responses would make this component more robust. I rechecked the current parent/route lifecycle, though, and I would not keep that as a merge blocker for this PR: the normal cross-job navigation path remounts the page because `router-view` is keyed by `route.fullPath`, while this PR's real bug is the same component receiving the real `jobId` after initial mount. The new watcher handles that path, and the falsy guard avoids the initial empty request.

I would still keep the stale-response guard as a nice follow-up, for example:
```ts
getJobLogs(jobId).then((res) => {
  if (jobId === props.jobId) logList.value = res
})
```

I did not run local frontend tests in this round. This review is based on:
- `git diff --stat upstream/dev...01945d6ef326`: 1 file changed, 10 insertions and 3 deletions.
- `git diff upstream/dev...01945d6ef326` for `components/job-log/index.tsx`.
- Parent lifecycle lookup in `layouts/main/index.tsx` and `views/jobs/detail.tsx`.
- GitHub check metadata for the current Build run.

# Current CI / merge gate
The aggregate `Build` check is red: https://github.com/apache/seatunnel/runs/86718101675

The linked fork workflow run is accessible and does not show a source failure from this UI diff: it has 10 successful jobs, 19 cancelled jobs, and 1 skipped job. I do not see a failing job that points back to `components/job-log/index.tsx`.

So from Daniel's side, the source blocker is cleared. The remaining gate is to rerun/retrigger the Build workflow so the cancelled aggregate result can be replaced with a completed green run.

# Merge conclusion

Blocking items:
- None from Daniel's source review on the latest head.
- Remaining gate: rerun Build; the current aggregate result is red because jobs were cancelled/skipped.

Suggested follow-ups, non-blocking:
- Add the stale-response guard above for extra robustness.
- Add a small component test for the delayed-`jobId` update path when the UI test setup is available.

Overall, this is a small and correct Zeta UI lifecycle fix. I am not approving yet only because the required Build gate is not green.

DanielLeens:
Rechecked the current head `01945d6ef326` against the latest `dev` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
Job detail route
  -> layouts/main/index.tsx renders router-view with key={route.fullPath}
      -> navigating to another job detail remounts the detail page
  -> views/jobs/detail.tsx
      -> getJobInfo(route job id)
      -> Object.assign(job, res)
      -> <JobLog jobId={job.jobId}>

JobLog component
  -> setup(props)
      -> watch(() => props.jobId, handler, { immediate: true })
      -> handler ignores empty jobId
      -> getJobLogs(jobId)
      -> logList.value = response
      -> render NCollapseItem / iframe log links
```

## Merge conclusion

1. Blocking items
- This branch is behind upstream (`compare status: diverged`, `ahead_by=3`, `behind_by=1`), and the current failing gate is `FAILURE gate: Build`: https://github.com/apache/seatunnel/runs/86718101675

2. Lowest-cost next step

DanielLeens (commented):
Thanks for keeping this moving. I re-reviewed the current head `900f6f99c4f0` from scratch, independently retracing the full runtime chain rather than relying on my prior conclusions.

**Important finding first**: the only commit added since my last full review (at `01945d6ef326`) is a merge commit (`900f6f99c4f0`, "Merge remote-tracking branch 'upstream/dev' into pr11389-head") that pulls in ~200 unrelated `dev` commits to refresh the branch. I diffed the actual PR content at both heads and it is byte-for-byte identical:

```
git diff 01945d6ef326 900f6f99c4f0 -- seatunnel-engine/seatunnel-engine-ui
(empty)
```

So the source-level conclusions from my last review still apply verbatim; this round is confirmation, not a new finding.

In the Zeta engine UI job-detail page, `JobLog` (`seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx`) fetched logs exactly once inside `setup()`, using whatever `props.jobId` happened to be at that instant. The single call site, `views/jobs/detail.tsx:616` (`<JobLog jobId={job.jobId}></JobLog>`), passes `job.jobId` from a `reactive({} as Job)` object that starts empty and is populated asynchronously by `Object.assign(job, res)` inside `fetch()` after `getJobInfo(jobId)` resolves. That means `JobLog` mounts with `props.jobId === undefined`, fires `getJobLogs(undefined)` once, and never refetches once the real id lands — logs silently never load. This PR replaces the one-shot call with a `watch(() => props.jobId, handler, { immediate: true })`, with a falsy-id guard, so the initial `undefined` request is skipped and the watcher fires again once the id resolves to a real value.

# 1. Code Change Review
## 1.1 Core Logic Analysis
Diff (`seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx`, 1 file, +10/-3):
```diff
-import { NCollapse, NCollapseItem, NSpace } from 'naive-ui'
-import { defineComponent, ref } from 'vue'
+import { NCollapse, NCollapseItem } from 'naive-ui'
+import { defineComponent, ref, watch } from 'vue'
...
-    getJobLogs(props.jobId).then((res) => (logList.value = res))
+    watch(
+      () => props.jobId,
+      (jobId) => {
+        if (!jobId) return
+        getJobLogs(jobId).then((res) => (logList.value = res))
+      },
+      { immediate: true }
+    )
```
I independently retraced the full call chain rather than trusting the PR description:
```
layouts/main/index.tsx:66  <router-view key={this.routeKey} .../>   (routeKey = route.fullPath, line 27/37)
  -> navigating to a different job's URL remounts the whole detail page (new component instance)
views/jobs/detail.tsx
  -> jobId = route.params.jobId (fixed for the page's lifetime)
  -> job = reactive({} as Job)               (job.jobId starts undefined)
  -> fetch() -> getJobInfo(jobId) -> Object.assign(job, res)   (job.jobId set exactly once to a stable string)
  -> fetchTimer = setTimeout(fetch, 5000)     (re-polls every 5s, but returns the SAME jobId each time)
  -> <JobLog jobId={job.jobId}>
JobLog.setup(props)
  -> watch(() => props.jobId, handler, { immediate: true })
  -> handler(undefined) on mount -> guarded, no request
  -> handler(realJobId) once job info resolves -> getJobLogs(realJobId) -> logList.value = res
  -> render: logList.value mapped into NCollapse/NCollapseItem + iframe per log
```
This confirms: (a) the fix targets the correct root cause — the async `undefined -> real jobId` transition on initial load, not the "switch between two different jobs while mounted" scenario SEZ9 and I both flagged in earlier rounds; and (b) there is only one call site for `JobLog` in the whole codebase (verified via `git grep -n "JobLog"`), so there is no other code path that could change `props.jobId` a second time on an already-mounted instance — `job.jobId` is assigned once and every subsequent 5s poll returns the same string, which does not re-trigger the watcher (Vue only re-fires `watch` on an actual value change). The fix is correct and complete for every reachable path that can change the active job id in this codebase today.

## 1.2 Compatibility Impact
No config, protocol, checkpoint/serialization, or public-API surface touched — this is a pure frontend component internal change (a `setup()`-scoped `ref`/`watch`, no prop/type/route changes). No backward-compatibility risk for existing tasks, saved jobs, or upgrade paths.

`{ immediate: true }` plus the falsy guard means the watcher fires at most twice per component instance in the realistic path (once on mount with `undefined` → skipped, once when the real id resolves → one HTTP call), i.e. no extra network traffic versus the old one-shot call, and it now actually succeeds instead of firing a doomed `getJobLogs(undefined)` request. No memory leak: `watch` inside `setup()` is automatically torn down on unmount by Vue; no explicit stop handle is needed here since the component's lifetime bounds it.

No behavior change here versus the prior code — neither version has a `.catch` on the `getJobLogs` promise. This is a pre-existing gap, not something this PR should be scoped to fix, but worth a non-blocking mention below (Issue 3).

# 2. Code Quality Assessment
## 2.1 Coding Standards
Idiomatic Vue 3 Composition API; the dangling `NSpace` import removal is correct — `git grep NSpace` under `components/job-log` returns nothing post-change, confirming it's genuinely unused and not a silent behavior change.

No Vitest/`@vue/test-utils` coverage was added for the new watcher behavior, and the PR description states `vue-tsc` type-check could not be run locally (`node_modules` not installed). This was already raised by SEZ9 (Low) and myself; I'm not escalating it to a blocker for the same reason as last round — it's a one-line, low-risk reactive fix with an obvious runtime chain, and the module's existing test setup is not exercised by this change either way, so it is not a regression in coverage discipline for this file specifically. Still a legitimate non-blocking follow-up.

## 2.3 Documentation Updates
None required; no user-facing config/behavior contract changed.

# 3. Architectural Soundness
Minimal, idiomatic, single-file fix. `watchEffect` was suggested by SEZ9 as a stylistic alternative (Low, non-blocking) — the current explicit `watch` is equally acceptable and arguably clearer about the dependency being tracked.

## 3.2 Maintainability
Small enough that maintenance risk is low. The guard `if (!jobId) return` is easy to miss the intent of without a comment — a one-line comment noting "skip until job info has resolved a real id" would help future readers, but I'm not blocking on it.

## 3.3 Extensibility
Fine as-is; if `JobLog` ever gains a second call site where `jobId` can genuinely change more than once post-mount (e.g., a future in-page job switcher), the stale-response race SEZ9 and I both flagged becomes real and must be fixed then. Worth flagging explicitly in case this component is reused later.

Not applicable — pure frontend component, no serialization/checkpoint/upgrade-path involvement, no released config or API being touched.

# 4. Issue Summary

**Issue 1 (Medium, non-blocking): Stale-response race is theoretically possible if `JobLog` ever gets a second call site with a truly dynamic `jobId`**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:36`
- `getJobLogs(jobId).then((res) => (logList.value = res))` has no staleness guard; if `jobId` could change twice on one mounted instance with out-of-order network responses, an older response could overwrite newer logs.
- I verified this is NOT currently reachable: the sole call site (`views/jobs/detail.tsx:616`) only ever transitions `jobId` from `undefined` to one stable value per component instance (cross-job navigation remounts the page via the keyed `router-view`), so the watcher fires the real request at most once per instance.
- Suggested fix (future-proofing, not required for this PR): `getJobLogs(jobId).then((res) => { if (jobId === props.jobId) logList.value = res })`.
- Raised by another reviewer: Yes (SEZ9, both rounds; also raised by me in the prior round).

**Issue 2 (Low, non-blocking): No component test and no confirmed `vue-tsc` type-check for the new watcher**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:32`
- Suggested fix: a small Vitest test mounting the component, updating the `jobId` prop, and asserting `getJobLogs` is invoked once the id becomes truthy; confirm `npm run type-check` passes in CI.
- Raised by another reviewer: Yes (SEZ9).

**Issue 3 (Low, non-blocking): `getJobLogs(...).then(...)` has no `.catch`**
- Location: `seatunnel-engine/seatunnel-engine-ui/src/components/job-log/index.tsx:36`
- Pre-existing gap (unchanged from before this PR), so not a regression, but since this code is being touched anyway it would be a good time to add a rejection handler so a failed log fetch doesn't produce an unhandled promise rejection with no user-visible feedback.

# 5. Merge Recommendation
### Conclusion: source blocker cleared on re-review; CI failure on this head is unrelated to the PR diff, confirmed via direct log inspection

This is a correct, minimal, single-file Vue lifecycle fix. Confirmed no code change since my last full review — the new commit is a no-op merge from `dev`. All outstanding items are non-blocking follow-ups already on record from SEZ9's review.

1. Blockers — must be fixed: None.
   - Add the stale-response guard (Issue 1) as defensive future-proofing.
   - Add a small Vitest test + confirm `vue-tsc` type-check (Issue 2).
   - Add a `.catch` on the `getJobLogs` promise while this file is already open (Issue 3).

**CI status**: the aggregate apache-side `Build` check is red, but per SeaTunnel's fork-CI model the apache-side check is only a pointer to the real run in the contributor's fork (`zhaoysg/seatunnel` run `31950045096`). I inspected that run directly:
- The actual failure is a single real (non-cancelled) job: `unit-test (11, windows-latest)`, step "run all modules unit test (Windows)", failing in `seatunnel-engine-server` with `CheckpointTimeOutTest>AbstractSeaTunnelServerTest.before:70 » IllegalState Node...` — this is the known Windows Hazelcast "Node failed to start!" environmental flake (different test class hit each time; unrelated to any Java/backend code, and this PR touches zero Java files).
- The other 3 `unit-test` matrix jobs (8/windows, 8/ubuntu, 11/ubuntu) show `conclusion: cancelled`, not `failure` — they were cancelled by the fail-fast matrix once the windows/JDK11 lane failed, not independently failing.
- `kudu-connector-it`, `rocketmq-connector-it`, and `doris-connector-it` also show as failed/cancelled in the run, but these are Java connector integration tests entirely unrelated to a Vue component watcher change — this PR's diff touches only `seatunnel-engine-ui/src/components/job-log/index.tsx`.
- Net: the red CI on this head is upstream/environmental noise (Windows Hazelcast flake plus its fail-fast fallout), not caused by this diff.

Overall assessment: no source-level blocker exists on the current head. This is ready to merge from a code-review standpoint once the branch gets a green (or PR-scoped) CI run — I'd suggest rerunning just the failed jobs rather than the whole matrix, since the failure is a known infra flake.

PR #11393, opened 2026-07-10
[Fix][Core] Exclude Scala classes from Spark starter

### LINKED ISSUE #10881: [Bug] Spark 3.3.1 on YARN cluster mode fails with Scala signature error (expected 5.0 found 5.2)
### Search before asking

### What happened

SeaTunnel 2.3.12，Spark 3.3.1，Scala 2.12，YARN Cluster

Error reading Scala signature of scala.collection.immutable.package: expected: 5.0 found: 5.2

### SeaTunnel Version

2.3.12

### SeaTunnel Config

```conf
env {
        spark.yarn.submit.waitAppCompletion=true
    spark.executor.instances = 1
    spark.executor.cores = 1
    spark.executor.memory = "2g"
        parallelism = 1
        job.mode = "BATCH"
        job.name = "mysql2hive2"


}

source {
   jdbc {
        url = "jdbc:mysql://:3307/dolphinscheduler341",
        driver = "com.mysql.cj.jdbc.Driver",
        user = "",
        password = "",
        query = "SELECT *,'20260508' as dt from dolphinscheduler341.t_ds_workflow_instance WHERE start_time <'2026-05-08 00:00:00' "
        }
}


sink {
  Hive {
    table_name = "default.t_ds341_workflow_instance_bak_from_1783"
    metastore_uri ="
        overwrite = "true"

  }
}
```

### Running Command

```shell
./bin/start-seatunnel-spark-3-connector-v2.sh   --master yarn   --deploy-mode cluster   --config ./config/mysql2hive.conf
```

### Error Exception

```log
Error reading Scala signature of scala.collection.immutable.package: expected: 5.0 found: 5.2
```

MRS 850, Spark 3.3.1 (Scala 2.12)

### Java or Scala Version

openjdk version "1.8.0_452"
scala-library-2.12.15

### Screenshots

Temporary solution:
1. Using Client mode can run normally.

2. Manually delete with the command: zip -d seatunnel-spark-3-starter.jar "scala/*".

### Code of Conduct

zhangshenghang:
## What does this PR do?

Fixes #10881.

The Spark starter shaded Scala classes from both the direct Scala dependency and already-shaded SeaTunnel dependencies. In YARN cluster mode, those classes can conflict with the Scala runtime provided by the Spark distribution and cause Scala signature errors.

This change excludes Scala classes and stale signature metadata from Spark starter artifacts. It also adds a packaging integration test that verifies the Spark 3 starter jar does not contain `scala/` entries.

## Validation

- `./mvnw -pl seatunnel-core/seatunnel-spark-starter/seatunnel-spark-3-starter -am verify -DskipUT=true -DskipIT=false -Dit.test=SparkStarterJarContentIT -DfailIfNoTests=false`

### DIFF
--- a/docs/en/connectors/sink/Hudi.md
+++ b/docs/en/connectors/sink/Hudi.md
@@ -15,6 +15,12 @@ Used to write data to Hudi.
 - [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)
 - [x] [timer flush](../../introduction/concepts/connector-v2-features.md)
 
+:::caution Hive Metastore synchronization
+
+The SeaTunnel Hudi sink writes Hudi data files and `.hoodie` metadata, but it does not register the table in or synchronize the table with Hive Metastore. Options matching `hoodie.datasource.hive_sync.*` are not supported sink options and are not passed to the Hudi write client. Run Apache Hudi's `HiveSyncTool` or another table registration process separately when Hive Metastore registration is required.
+
+:::
+
 ## Options
 
 Base configuration:
--- a/docs/en/connectors/sink/Tablestore.md
+++ b/docs/en/connectors/sink/Tablestore.md
@@ -6,65 +6,98 @@ import ChangeLog from '../changelog/connector-tablestore.md';
 
 ## Description
 
-Write data to `Tablestore`
+Write SeaTunnel rows to Alibaba Cloud Tablestore.
 
 ## Key features
 
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
+- [ ] [cdc](../../introduction/concepts/connector-v2-features.md)
+- [ ] [support multiple table write](../../introduction/concepts/connector-v2-features.md)
+- [ ] [timer flush](../../introduction/concepts/connector-v2-features.md)
 
 ## Options
 
-|       name        |  type  | required | default value |
-|-------------------|--------|----------|---------------|
-| end_point         | string | yes      | -             |
-| instance_name     | string | yes      | -             |
-| access_key_id     | string | yes      | -             |
-| access_key_secret | string | yes      | -             |
-| table             | string | yes      | -             |
-| primary_keys      | array  | yes      | -             |
-| batch_size        | string | no       | 25            |
-| common-options    | config | no       | -             |
+| name              | type   | required | default value | description                                      |
+|-------------------|--------|----------|---------------|--------------------------------------------------|
+| end_point         | string | yes      | -             | Tablestore endpoint.                             |
+| instance_name     | string | yes      | -             | Tablestore instance name.                        |
+| access_key_id     | string | yes      | -             | AccessKey ID used to access Tablestore.          |
+| access_key_secret | string | yes      | -             | AccessKey secret used to access Tablestore.      |
+| table             | string | yes      | -             | Target Tablestore table name.                    |
+| primary_keys      | array  | yes      | -             | Primary key field names in the target table.     |
+| schema            | config | yes      | -             | Input schema. Primary key fields must also exist in `schema.fields`. |
+| batch_size        | int    | no       | 25            | Maximum number of rows written in one batch.     |
+| common-options    | config | no       | -             | Sink common options.                             |
 
-### end_point [string]
+## Usage notes
 
-endPoint to write to Tablestore.
+- `primary_keys` can contain one or more primary key fields. These fields are written as Tablestore primary key columns; all other schema fields are written as normal attribute columns.
+- The sink writes rows with Tablestore `RowPutChange` and `RowExistenceExpectation.IGNORE`. It does not delete rows when upstream sends `DELETE` row kinds.
+- `batch_size` controls when buffered rows are flushed. The writer also flushes remaining rows when the job closes.
+- Keep `access_key_id` and `access_key_secret` out of committed job files. Prefer runtime variable substitution or a secret manager supported by your deployment environment.
 
-### instanceName [string]
-
-The instanceName of Tablestore.
-
-### access_key_id [string]
-
-The access id of Tablestore.
-
-### access_key_secret [string]
-
-The access secret of Tablestore.
-
-### table [string]
-
-The table of Tablestore.
-
-### primaryKeys [array]
-
-The primaryKeys of Tablestore.
-
-### common options [ config ]
+### common options [config]
 
 Sink plugin common parameters, please refer to [Sink Common Options](../common-options/sink-common-options.md) for details.
 
 ## Example
 
-```bash
-Tablestore {
-    end_point = "xxxx"
-    instance_name = "xxxx"
-    access_key_id = "xxxx"
-    access_key_secret = "xxxx"
-    table = "sink"
-    primary_keys = ["pk_1","pk_2","pk_3","pk_4"]
+```hocon
+env {
+  parallelism = 1
+  job.mode = "BATCH"
+}
+
+source {
+  FakeSource {
+    row.num = 2
+    schema = {
+      fields {
+        order_id = string
+        user_id = string
+        amount = double
+      }
+    }
+    rows = [
+      {
+        fields = ["order-1", "user-1", 99.5]
+      },
+      {
+        fields = ["order-2", "user-2", 20.0]
+      }
+    ]
   }
+}
+
+sink {
+  Tablestore {
+    end_point = "https://<instance>.<region>.ots.aliyuncs.com"
+    instance_name = "<instance-name>"
+    access_key_id = "${ACCESS_KEY_ID}"
+    access_key_secret = "${ACCESS_KEY_SECRET}"
+    table = "orders"
+    primary_keys = ["order_id"]
+    batch_size = 25
+    schema = {
+      fields {
+        order_id = string
+        user_id = string
+        amount = double
+      }
+    }
+  }
+}
 ```
+
+## Type mapping
+
+| SeaTunnel type                         | Tablestore attribute column type | Tablestore primary key type |
+|----------------------------------------|----------------------------------|-----------------------------|
+| `INT`, `TINYINT`, `SMALLINT`, `BIGINT` | `INTEGER`                        | `INTEGER`                   |
+| `FLOAT`, `DOUBLE`, `DECIMAL`           | `DOUBLE`                         | `STRING`                    |
+| `STRING`, `DATE`, `TIME`, `TIMESTAMP`  | `STRING`                         | `STRING`                    |
+| `BOOLEAN`                              | `BOOLEAN`                        | `STRING`                    |
+| `BYTES`                                | `BINARY`                         | `BINARY`                    |
 
 ## Changelog
 
--- a/docs/en/connectors/source/Tablestore.md
+++ b/docs/en/connectors/source/Tablestore.md
@@ -6,99 +6,98 @@ import ChangeLog from '../changelog/connector-tablestore.md';
 
 ## Description
 
-Read data from Alicloud Tablestore，support full and CDC.
-
+Read full and incremental data from Alibaba Cloud Tablestore. The source uses Tablestore Tunnel in `BaseAndStream` mode, so it can read existing data first and then consume later changes.
 
 ## Key features
 
-- [ ] [batch](../../introduction/concepts/connector-v2-features.md)
-- [X] [stream](../../introduction/concepts/connector-v2-features.md)
+- [x] [batch](../../introduction/concepts/connector-v2-features.md)
+- [x] [stream](../../introduction/concepts/connector-v2-features.md)
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
-- [ ] [column projection](../../introduction/concepts/connector-v2-features.md)
-- [ ] [parallelism](../../introduction/concepts/connector-v2-features.md)
+- [x] [cdc](../../introduction/concepts/connector-v2-features.md)
+- [x] [column projection](../../introduction/concepts/connector-v2-features.md)
+- [x] [parallelism](../../introduction/concepts/connector-v2-features.md)
 - [ ] [support user-defined split](../../introduction/concepts/connector-v2-features.md)
 
 ## Options
 
-| name                  | type   | required | default value |
-|-----------------------|--------|----------|---------------|
-| end_point             | string | yes      | -             |
-| instance_name         | string | yes      | -             |
-| access_key_id         | string | yes      | -             |
-| access_key_secret     | string | yes      | -             |
-| table                 | string | yes      | -             |
-| primary_keys          | array  | yes      | -             |
-| schema                | config | yes      | -             |
+| name              | type   | required | default value | description                                                                 |
+|-------------------|--------|----------|---------------|-----------------------------------------------------------------------------|
+| end_point         | string | yes      | -             | Tablestore endpoint, for example `https://<instance>.<region>.ots.aliyuncs.com`. |
+| instance_name     | string | yes      | -             | Tablestore instance name.                                                    |
+| access_key_id     | string | yes      | -             | AccessKey ID used to access Tablestore.                                      |
+| access_key_secret | string | yes      | -             | AccessKey secret used to access Tablestore.                                  |
+| table             | string | yes      | -             | Tablestore table name. Multiple tables can be separated by commas.           |
+| primary_keys      | array  | yes      | -             | Primary key names. For multiple source tables, configure one primary key name for each table in the same order as `table`. |
+| schema            | config | yes      | -             | Output schema. For details, see [Schema Feature](../../introduction/concepts/schema-feature.md). |
 
+## Usage notes
 
-### end_point [string]
-
-The endpoint of Tablestore.
-
-### instance_name [string]
-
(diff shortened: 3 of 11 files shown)


DanielLeens (approved):
Thanks for the fix. I reviewed the full diff and traced the packaging path through the Spark starter parent POM into the shaded starter artifacts.

# What this PR fixes
- User pain: in Spark cluster mode, shipping SeaTunnel starter jars that still embed Scala runtime classes can collide with the Scala runtime already provided by the Spark cluster image / distribution.
- Fix approach: exclude the direct `scala-library` artifact from the shaded Spark starter packaging and add a jar-content regression test that checks the produced starter jar for embedded `scala/` classes.
- In one sentence: this keeps the Spark starter aligned with the cluster-provided Scala runtime instead of bundling another copy inside the shipped jar.

# Runtime path checked
```text
Build Spark starter module
  -> parent spark-starter pom drives maven-shade-plugin packaging
      -> scala-library is excluded from the shaded artifact set
      -> scala/** entries are filtered out from shaded contents
  -> produced starter jar is written to target/
      -> SparkStarterJarContentIT opens the built jar
      -> asserts scala/ classes are not embedded in the final artifact
```

# Review findings
- I do not see a source-level blocker in the current implementation.
- The regression test is useful because it checks the produced jar content rather than only the dependency tree.
- Small note only: the parent POM change applies to both Spark starter variants, while the regression test currently covers the Spark 3 starter path. That is reasonable for now, but if we later see Spark 2 packaging drift independently, adding the same jar-content check there would make the coverage symmetric.

# Merge conclusion
### Conclusion: can merge

1. Blocking items
- None from my side.

2. Suggested follow-up
- Optional only: consider mirroring the jar-content assertion in the Spark 2 starter module later for symmetric regression coverage.

The remaining gate on the PR is CI completion rather than a new code-side blocker from Daniel's side.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `5510f2ce353f` from scratch.

# Findings
**Issue 1: Blanket `scala/**` filter on `*:*` strips Scala classes from every shaded dependency; safe only if nothing on the starter's pre-spark-submit classpath needs them and Spark's provided Scala is binary compatible**
- Location: `seatunnel-core/seatunnel-spark-starter/pom.xml:112`
- Why it matters: - Root cause: the new filter applies `scala/**` to `<artifact>*:*</artifact>`, so Scala classes embedded in already-shaded SeaTunnel dependencies (e.g., translation/uber jars) are removed from the starter jar, not just the direct `org.scala-lang:scala-library` artifact excluded at line 104.
- Impact: any code executed by the starter JVM itself (before `spark-submit` hands off to the Spark-provided runtime) that touches a Scala class will now throw `NoClassDefFoundError`; additionally, if a shaded SeaTunnel dependency was compiled against a Scala minor version different from the Spark distribution's (e.g., 2.12.x patch differences are fine, but 2.12 vs 2.13 is not), users can hit `ScalaSignature`/binary incompatibility at job runtime instead of the previous duplicate-class conflict.
- Fix: document in the pom comment which dependencies were found to embed Scala and why relying on Spark's runtime is safe for them, and add a smoke IT (even a local-mode `SparkStarter` invocation) confirming the starter launches end-to-end, not just that `scala/` entries are absent. Also consider widening the artifactSet exclusion to `org.scala-lang:*` so `scala-reflect`/`scala-compiler` are excluded at the dependency level rather than only via the content filter.
- Evidence: `seatunnel-core/seatunnel-spark-starter/pom.xml:111  <!-- Some SeaTunnel dependencies embed Scala classes. -->`; `seatunnel-core/seatunnel-spark-starter/pom.xml:112  <exclude>scala/**</exclude>`
- Suggested fix: Applying `scala/**` under `<artifact>*:*</artifact>` strips Scala classes out of every shaded dependency, not just `scala-library`. That is the right call for YARN cluster mode, but it means (a) any starter-side code path executed before `spark-submit` must be pure Java, and (b) shaded SeaTunnel deps that embedded Scala must be binary compatible with the Spark distribution's Scala version. Please note in this comment which dependencies were found embedding Scala, and consider also excluding `org.scala-lang:*` (covers `scala-reflect`) at the artifactSet level. A local-mode launch smoke test would give stronger assurance than the jar-content check alone.
- Severity: Medium

**Issue 2: The shade fix lives in the shared parent pom (covers both Spark 2 and Spark 3 starters) but the packaging IT only guards the Spark 3 starter**
- Location: `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-3-starter/src/test/java/org/apache/seatunnel/core/starter/spark/SparkStarterJarContentIT.java:31`
- Why it matters: - Root cause: the `<excludes>` and `<filters>` were added to `seatunnel-core/seatunnel-spark-starter/pom.xml`, which is inherited by both `seatunnel-spark-2-starter` and `seatunnel-spark-3-starter`, but the new `SparkStarterJarContentIT` was only added under `seatunnel-spark-3-starter`.
- Impact: a future change that overrides the shade configuration in the Spark 2 starter module (a common pattern when modules customize `artifactSet`) would reintroduce embedded Scala classes there with zero test coverage, and #10881-style Scala signature conflicts would resurface only for Spark 2 / YARN users.
- Fix: add an equivalent jar-content IT to the Spark 2 starter module, or move the test into a shared test source (e.g., a small abstract base in a common test module parameterized by artifact name). Also make sure failsafe is actually bound in these modules so `*IT` classes run in the default CI build, not only with the manual `-DskipIT=false -Dit.test=...` invocation from the PR description.
- Evidence: `seatunnel-core/seatunnel-spark-starter/pom.xml:112  <exclude>scala/**</exclude>`; `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-3-starter/src/test/java/org/apache/seatunnel/core/starter/spark/SparkStarterJarContentIT.java:31  public class SparkStarterJarContentIT {`
- Suggested fix: Since the shade exclusions were added to the shared `seatunnel-spark-starter/pom.xml`, they also affect the Spark 2 starter, but only the Spark 3 module gets this guard. Please add the same jar-content check for `seatunnel-spark-2-starter` (or share the test via a parameterized base class), and confirm failsafe is bound in this module so `*IT` runs in the standard CI pipeline rather than only with `-Dit.test=SparkStarterJarContentIT`.
- Severity: Medium

**Issue 3: SparkStarterJarContentIT picks an arbitrary jar from target/, which can silently validate the wrong artifact (sources/javadoc/tests jar) instead of the shaded starter jar**
- Location: `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-3-starter/src/test/java/org/apache/seatunnel/core/starter/spark/SparkStarterJarContentIT.java:39`
- Why it matters: - Root cause: `Files.list(Paths.get("target"))` returns entries in no guaranteed order, and the only filters applied are `.endsWith(".jar")` and `!startsWith("original-")`. When the build also produces `*-sources.jar`, `*-javadoc.jar`, or `*-tests.jar` (all common in release/deploy profiles), `findFirst()` may return one of those instead of the shaded starter jar.
- Impact: the assertion `!entry.getName().startsWith("scala/")` trivially passes on a sources/javadoc jar, so the test can go green while the shaded jar regresses and re-embeds Scala classes — exactly the failure mode #10881 describes would go undetected in CI.
- Fix: filter on the expected artifact name, e.g. `path.getFileName().toString().matches("seatunnel-spark-3-starter-.*\\.jar")` and additionally exclude `-sources`, `-javadoc`, and `-tests` suffixes; alternatively resolve the jar name from `project.build.finalName` passed in via a system property from the failsafe configuration.
- Evidence: `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-3-starter/src/test/java/org/apache/seatunnel/core/starter/spark/SparkStarterJarContentIT.java:38  .filter(path -> !path.getFileName().toString().startsWith("original-"))`; `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-3-starter/src/test/java/org/apache/seatunnel/core/starter/spark/SparkStarterJarContentIT.java:39  .findFirst()`
- Suggested fix: `findFirst()` over `Files.list(target)` is non-deterministic and the only exclusion is the `original-` prefix. If a `-sources.jar`, `-javadoc.jar`, or `-tests.jar` is produced (release/deploy profiles), this test can validate that jar and pass trivially even if the shaded starter jar regresses. Please match the shaded artifact explicitly, e.g. filter on `seatunnel-spark-3-starter-` prefix and exclude `-sources`/`-javadoc`/`-tests` suffixes, or pass `${project.build.finalName}` in via a failsafe system property.
- Severity: Medium

**Issue 4: User-facing packaging change (Scala no longer bundled in Spark starter jars) has no release-note/changelog or deployment-doc update**
- Location: `seatunnel-core/seatunnel-spark-starter/pom.xml:104`
- Why it matters: - Root cause: this PR changes what ships inside the Spark starter fat jars — Scala classes and signature files are removed — but touches no files under `docs/` and adds no release-note entry.
- Impact: users who currently run the starter jar in environments without a full Spark distribution on the classpath (custom launchers, embedded testing setups) will see `NoClassDefFoundError: scala/...` after upgrading with no documented explanation; conversely, YARN cluster users won't know this release fixes #10881 unless it's called out.
- Fix: add a brief entry to the release notes / upgrade guide stating that starting from this version the Spark starter jars no longer embed the Scala runtime and rely on the Scala provided by the Spark distribution, and reference #10881. If the Spark deployment docs (e.g., `docs/en/start-v2/` Spark engine pages) mention jar contents or classpath requirements, update them accordingly.
- Evidence: `seatunnel-core/seatunnel-spark-starter/pom.xml:104  <exclude>org.scala-lang:scala-library</exclude>`
- Suggested fix: Since this removes the embedded Scala runtime from the shipped starter jars, please add a short note to the release notes / upgrade guide (referencing #10881) that the Spark starter now relies on the Scala runtime provided by the Spark distribution. Users launching the starter jar outside a standard Spark distribution need to know this to avoid surprise `NoClassDefFoundError: scala/...` failures.
- Severity: Medium

# Review conclusion

- Issue 1: Blanket `scala/**` filter on `*:*` strips Scala classes from every shaded dependency; safe only if nothing on the starter's pre-spark-submit classpath needs them and Spark's provided Scala is binary compatible
- Issue 2: The shade fix lives in the shared parent pom (covers both Spark 2 and Spark 3 starters) but the packaging IT only guards the Spark 3 starter
- Issue 3: SparkStarterJarContentIT picks an arbitrary jar from target/, which can silently validate the wrong artifact (sources/javadoc/tests jar) instead of the shaded starter jar
- Issue 4: User-facing packaging change (Scala no longer bundled in Spark starter jars) has no release-note/changelog or deployment-doc update

DanielLeens (changes requested):
Thanks @SEZ9. I rechecked the current unchanged head `5510f2ce353f` against the packaging path again, and I agree this is still a blocker on top of the current diff.

The jar-content assertion added in `SparkStarterJarContentIT` proves that `scala/` entries are removed from the built jar, but it still does not prove that the starter remains runnable after applying that wildcard filter to every shaded artifact. The critical point for me is that the starter-side code path still references Scala classes directly before we even get to a Spark job runtime. For example, `seatunnel-core/seatunnel-spark-starter/seatunnel-spark-starter-common/src/main/java/org/apache/seatunnel/core/starter/spark/execution/SourceExecuteProcessor.java:43` still imports `scala.Tuple2`, while the new parent-POM filter applies `scala/**` under `<artifact>*:*</artifact>` in `seatunnel-core/seatunnel-spark-starter/pom.xml:112`.

So from Daniel's side I need one more change before merge:

1. either narrow the exclusion to the actual Scala artifacts/classes we intentionally want to drop,
2. or add a real starter smoke path that proves the launcher still works end-to-end with the filtered jar, not just that the jar no longer contains `scala/` entries.

Until that is covered, I do not think the current head is safe to merge yet.

DanielLeens:
Rechecked the current unchanged head `5510f2ce353f` again.

My earlier source-side blocker on this same revision still stands from Daniel's side. Separately, this branch is also behind the latest `dev` (`compare status: diverged`, `behind_by=1`), and the current red `Build` signal does not look actionable on top of this stale head yet.

https://github.com/apache/seatunnel/actions/runs/29081335402/job/86324629152

DanielLeens (approved):
Thanks for the update. I re-reviewed the latest head from scratch.

# What this PR fixes
- User pain: the Spark starter fat jar used to carry its own Scala classes, which can collide with the Scala runtime already provided by the target Spark distribution in cluster deployments.
- Fix approach: the shared Spark starter shade config now removes embedded Scala classes, and the PR adds regression coverage for both Spark 2 and Spark 3 starter artifacts.
- One-line summary: the starter jar now stays aligned with the Scala runtime that Spark already ships, and the new tests prove the shaded jar still behaves like a starter rather than only getting smaller.

# Runtime path checked
```text
Starter packaging
  -> seatunnel-core/seatunnel-spark-starter/pom.xml
      -> exclude scala-library from the shaded artifact set
      -> filter scala/** out of shaded contents for inherited Spark starter modules

Starter CLI path
  -> TestSparkStarterJarContentIT
      -> open the produced shaded starter jar
      -> assert scala/ entries are absent
      -> launch SparkStarter from that shaded jar with -h
      -> launch SparkStarter from that shaded jar with --config ... --deploy-mode client
      -> verify the starter still prints usage / builds the spark-submit command successfully

Spark runtime path
  -> SparkStarter only builds the spark-submit command in the local JVM
  -> runtime-side classes such as SourceExecuteProcessor run after handoff under Spark's own classpath
  -> those Scala references are therefore resolved from the Spark-provided runtime instead of the shaded starter jar
```

# Review result
The previous blocker from my last round is closed for me now.

The important gap before was that we only knew the jar no longer contained `scala/` entries, but we did not yet know whether the shaded starter itself still launched correctly. The new Spark 2 / Spark 3 integration tests close that gap in the right place:
- they explicitly select the real shaded starter jar instead of an arbitrary jar under `target/`
- they execute the starter directly from that shaded jar
- they cover both the usage path and the command-building path

I also rechecked the earlier concern around `SourceExecuteProcessor` still importing `scala.Tuple2`. That import sits on the Spark runtime side after `spark-submit` hands execution off to Spark, not on the lightweight starter CLI path that these tests exercise. Given that call chain, relying on Spark's provided Scala runtime is consistent with the current design.

I noticed @SEZ9's earlier comments as well. With the new Spark 2 coverage and the direct starter smoke path added in this revision, I think the practical source-side gap from that thread is addressed for this PR.

# Merge conclusion
### Conclusion: can merge

1. Blocking items

2. Suggested follow-up
- None required for merge.

At this point the remaining gate is CI completion rather than a new source-side blocker from Daniel's side.

DanielLeens:
Rechecked the current merge gate after my approval.

My source-side conclusion is unchanged: I still do not see a reopened Daniel-side blocker in the Spark starter shading diff. The CI fact has changed, though: the aggregate `Build` is no longer just pending/ongoing on this head; the current check run is now cancelled.

Current gate details:
```text
Apache aggregate check
  -> Build: cancelled
  -> https://github.com/apache/seatunnel/runs/86597995417
      -> fork workflow run: zhangshenghang/seatunnel/actions/runs/29173343906
          -> failed job: Run / all-connectors-it-1 (8, ubuntu-latest)
          -> failed job: Run / kafka-connector-it (11, ubuntu-latest)
          -> cancelled siblings: Run / kudu-connector-it (8, ubuntu-latest), Run / kafka-connector-it (8, ubuntu-latest), Run / paimon-connector-it (11, ubuntu-latest)
```

These lanes do not line up with this PR's starter-jar shading/test surface, and this branch is now behind the latest `dev` by 3 commits (`compare status: diverged`, `ahead_by=4`, `behind_by=3`). I also checked the current `dev` head, but its aggregate Build is still pending right now, so I cannot point to a confirmed upstream fix yet. The lowest-cost next step is to sync with the latest `dev` and rerun `Build` first.

My earlier approval still reflects the current source diff. If the same unrelated lanes keep failing after the sync, please paste the rerun links here and I can help narrow them down further.

### Merge conclusion: source approved, waiting for CI
Current merge gate: cancelled unrelated CI lanes above; please sync latest `dev` and rerun first.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=4`, `behind_by=8`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

DanielLeens:
Rechecked the same unchanged head `f7840be7d976` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
Apache aggregate check
  -> Build: cancelled
  -> https://github.com/apache/seatunnel/runs/86597995417
      -> fork workflow run: zhangshenghang/seatunnel/actions/runs/29173343906
          -> failed job: Run / all-connectors-it-1 (8, ubuntu-latest)
          -> failed job: Run / kafka-connector-it (11, ubuntu-latest)
          -> cancelled siblings: Run / kudu-connector-it (8, ubuntu-latest), Run / kafka-connector-it (8, ubuntu-latest), Run / paimon-connector-it (11, ubuntu-latest)
```

## Merge conclusion

1. Blocking items

2. Current gate
- The current CI signal is now green: https://github.com/apache/seatunnel/actions/runs/29173344457/job/86670647621

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

Issue #11364, opened 2026-07-06
[STIP-30][Feature][CDC] : Unified CDC lag and progress observability model


goutamadwant:
### Search before asking

- [x] I searched existing feature and CDC issues and found no equivalent proposal.

### Description

This is STIP-30, a design proposal for #11354.

SeaTunnel CDC connectors expose different source positions and lifecycle state: MySQL binlog files and GTIDs, PostgreSQL LSNs, Oracle SCNs, TiDB timestamps, snapshot splits, and checkpointed reader state. Operators currently need connector logs or checkpoint internals to determine whether a job is snapshotting, catching up, consuming incrementally, recovering, idle, or no longer progressing.

The proposal defines a normalized latest-progress view while preserving connector-native positions and without claiming that every connector supports the same lag semantics.

## Goals

- expose current snapshot, incremental, checkpoint, and restore progress
- keep connector-native position details available for diagnosis
- distinguish CDC lifecycle, recovery state, and runtime health
- represent unsupported or approximate values explicitly
- keep v1 additive and latest-snapshot only
- validate the model through complete MySQL and PostgreSQL runtime paths before stabilizing the public contract

## Non-goals for v1

- no historical timeline or retention store
- no UI dependency
- no universal numeric lag across connectors
- no inference that an unchanged position means backpressure or a stalled source
- no unbounded list of completed split details in job detail
- no change to CDC checkpoint or restore behavior

## Design principles

1. Connectors report only facts they own.
2. The engine adds job, pipeline, vertex, task, observation, and aggregation context.
3. Snapshot assignment and reader consumption remain separate report sources.
4. `EXACT` is used only when current runtime state proves a value.
5. Missing, stale, unsupported, and best-effort values are not replaced with zero or guessed values.
6. The implementation is validated internally before the public `seatunnel-api` contract is stabilized.

## Contract layers

| Layer | Responsibility |
| --- | --- |
| CDC reader report | Current consumed position, active split, source-event time, checkpoint candidate/completion state, and restore state owned by that reader |
| CDC enumerator report | Snapshot assignment, completed split watermarks, prepared remaining splits, remaining unchunked tables, and source-level lifecycle information |
| Engine snapshot | Runtime identity, report observation time, stale/missing report state, task aggregation, and latest-snapshot storage |
| REST response | Additive `cdcProgress` section built from the latest engine snapshot |

[...]

PR #11395, opened 2026-07-10
[Feature] [CDC] Add CDC progress snapshot model

goutamadwant:

Part of #11364.

This adds the first narrow CDC progress implementation slice:

- add a shared CDC progress snapshot model in `seatunnel-api`
- add CDC base helpers for building snapshot and incremental progress from CDC state
- add MySQL CDC progress mapping for binlog offsets
- add focused unit coverage for snapshot counts, incremental checkpoint position, raw offset copying, and MySQL binlog position preservation

This PR intentionally does not expose `cdcProgress` in job detail yet. It keeps the first change focused on the shared model and MySQL CDC helper path.

Yes, for the CDC progress contract.

This PR adds the shared progress snapshot model and MySQL CDC mapping helper. It does not change job execution behavior or expose `cdcProgress` in REST job detail yet.

Added unit tests:

- `CdcProgressSnapshotsTest`
- `MySqlCdcProgressTest`

Verified with Java 11:

`./mvnw -pl seatunnel-api,seatunnel-connectors-v2/connector-cdc/connector-cdc-base,seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql -DskipIT=true -Dlicense.skipAddThirdParty=true -Dskip.ui=true -DfailIfNoTests=false test -Dtest=CdcProgressSnapshotsTest,MySqlCdcProgressTest`

Also verified `seatunnel-api` clean compile with Java 11:

`./mvnw -pl seatunnel-api -DskipTests -DskipIT=true -Dlicense.skipAddThirdParty=true -Dskip.ui=true clean compile`

### Check list


DanielLeens (commented):
Thanks for this well-structured first slice of the CDC progress model! The design is clean — immutable model classes with `@Builder`, defensive copies in `CdcProgressPosition`, and `Serializable` throughout for Hazelcast compatibility. The separation between snapshot/incremental/checkpoint progress phases maps well to the actual CDC lifecycle.

A few observations:

**Issue 1 (Low): `CdcProgressPosition` constructor is public but `of()` factory duplicates the logic**
- `CdcProgressPosition` has both a public constructor and a static `of()` factory that does the same thing. Consider making the constructor package-private and using `of()` as the canonical entry point, or removing `of()` if the constructor is intended to be public. Minor, not a blocker.

**Issue 2 (Low): `CdcSnapshotProgress` does not default `snapshotProgress` to empty in `CdcProgressSnapshot`**
- In `CdcProgressSnapshot`, `snapshotProgress`, `incrementalProgress`, and `checkpointProgress` are nullable (no null-defaulting in the builder). This is fine since the `phase` enum tells the caller which one is populated, but a brief Javadoc on `CdcProgressSnapshot` clarifying that only the progress object matching the current `phase` is guaranteed to be non-null would help consumers.

**Test coverage** looks solid for a model PR — `CdcProgressSnapshotsTest` covers snapshot phase (assigned/completed/running/remaining counts, watermark extraction), incremental phase (consumed position, checkpoint position), and raw offset copy independence. `MySqlCdcProgressTest` verifies binlog field preservation (file, pos, gtids, server_id). The `TestOffset` test double directly sets the package-private `offset` field, which is fine for a test fixture.

No source-level blockers. Build is still pending — will check CI once it completes. This is a good foundation for the CDC progress feature.

DanielLeens:
Rechecked the current unchanged head `7f714aedf386` again.

My earlier source-side blocker on this same revision still stands from Daniel's side. Separately, this branch is also behind the latest `dev` (`compare status: diverged`, `behind_by=1`), and the current red `Build` signal does not look actionable on top of this stale head yet.

https://github.com/apache/seatunnel/actions/runs/29073063779/job/86298527028

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `afd0a58e1df8` against the full PR-owned CDC progress diff again.

One useful context note: the new commit after my previous review is a merge from latest `dev`; the extra diff since my last reviewed commit is only upstream Hudi/Tablestore docs, so the CDC progress model code itself is unchanged. I still rechecked the full current head rather than relying on the old review.

## What this PR does
- User pain: CDC jobs do not yet have a common model for snapshot, incremental, checkpoint, raw-position, and stalled progress reporting.
- Fix approach: add immutable CDC progress model classes in `seatunnel-api`, shared helper builders in CDC base, and a MySQL binlog-position helper.
- One-line summary: this is a clean model-layer foundation for future CDC progress observability; it does not yet wire progress reporting into the running source lifecycle.

## Runtime path checked
```text
Future CDC progress caller
  -> CdcProgressSnapshots.forSnapshotPhase(connectorType, SnapshotPhaseState, lastProgressTime)
      -> currentSnapshotSplit()/currentCompletedWatermark()
      -> CdcSnapshotProgress.builder()
      -> CdcProgressSnapshot.builder()

MySQL incremental progress helper
  -> MySqlCdcProgress.forIncrementalSplit(split, lastProgressTime)
      -> CdcProgressSnapshots.forIncrementalSplit()
      -> CdcProgressSnapshots.toPosition(BinlogOffset)
      -> immutable CdcProgressPosition values
```

## Findings
No source-level blockers from my side on the current head.

Testing/stability: the added unit tests cover snapshot counts/watermarks, incremental checkpoint/raw position mapping, defensive copy behavior, and MySQL binlog fields. I do not see sleep-based waits, external services, hard-coded ports, or timing-sensitive assertions. Stability rating: stable.

I did not run local tests; this review is based on local diff inspection, call-chain tracing, existing review context, and the current GitHub check state.

## Merge conclusion

1. Blocking items
- None from source review.

2. Non-blocking suggestions
- When this model becomes user-facing through metrics/REST/docs, please add field-level contract documentation and an end-to-end reporting test on top.

Current gate: `Build` is still queued: https://github.com/apache/seatunnel/runs/86567837028. I am not approving before the required check finishes successfully, but the source side looks clear to me.

DanielLeens (commented):
Rechecked the latest head: my source-side conclusion has not changed, but the current remaining gate is now a failed `Build` check.

The failed fork workflow currently shows:
- `Dead links`
- `all-connectors-it-1 (8, ubuntu-latest)`
- `kafka-connector-it (11, ubuntu-latest)`

This branch is diverged from latest `dev` (`behind_by=1`). These failed lanes do not appear to be directly tied to the CDC progress snapshot model itself yet, so the lowest-cost next step is to sync with latest `dev` and rerun Build. If the same lanes still fail afterward, the updated job log should give us a cleaner signal.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `16ce620fd282` from the current `dev` merge-base and walked the full PR-owned CDC progress diff again.

The new commit is a `dev` merge; the PR-owned source surface is still the CDC progress model/helper slice I reviewed before. I rechecked the API model classes, `CdcProgressSnapshots`, the MySQL helper, and the unit tests against the current base.

Runtime chain checked:
```text
CDC enumerator/source state snapshot
  -> SnapshotPhaseState / IncrementalSplit carries split, table, watermark and offset state
  -> CdcProgressSnapshots.forSnapshotPhase(...) or forIncrementalSplit(...)
      -> normalizes connector state into CdcProgressSnapshot
      -> copies Offset#getOffset() into immutable CdcProgressPosition
      -> records support-level metadata for snapshot/incremental/checkpoint/raw-position groups
  -> MySqlCdcProgress delegates MySQL binlog offsets into the same normalized model
```

Key findings:
- This PR still introduces model/helper APIs only; `git grep` shows no existing runtime path outside the new helper/tests is wired to publish these snapshots yet.
- The normal helper paths handle null/empty offsets defensively and copy offset maps before exposing them.
- The added tests are pure object-construction assertions; I do not see sleep, fixed ports, container startup, wall-clock assertions, or other flaky-test anti-patterns.
- I do not see a new Daniel-side source blocker on the current head.

### Merge conclusion: source clear, waiting for CI
Suggested follow-ups: once the queued `Build` finishes, this can move forward if CI stays green. The current check is https://github.com/apache/seatunnel/runs/86625906152.

DanielLeens (commented):
Hi @goutamadwant, thank you again for putting together this first CDC progress slice. A shared progress contract is valuable, and the overall layering idea—common model in `seatunnel-api`, connector-neutral mapping in CDC base, and connector-specific position adaptation in MySQL—is a promising direction.

I owe you an apology: after a deeper pass through the real `SnapshotSplitAssigner` and `IncrementalSourceReader` lifecycles, I found several correctness blockers that I missed in my earlier reviews on this revision. Sorry for the extra review round. These are not style preferences; the current helpers can publish incorrect values while marking them as `EXACT`, so I do not think this revision is ready to merge yet.

## 1. Snapshot split and watermark can be combined from different splits

Location:

- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/progress/CdcProgressSnapshots.java:44-71,128-175`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/enumerator/SnapshotSplitAssigner.java:172-188,200-203,237-251`

`currentSnapshotSplit()` takes the first value from `assignedSplits`, while `currentCompletedWatermark()` independently takes the first value from `splitCompletedOffsets`. The production maps are unordered, and `assignedSplits` retains both running and completed splits. There is no `splitId` join between the selected split and watermark, so table/split A can be reported with table/split B's watermark.

`remainingSplitCount` also counts only `remainingSplits`. It excludes assigned-but-not-completed splits and tables in `remainingTables` that have not been chunked yet, but the snapshot group is still marked `EXACT`.

Could we please change this so that:

- watermark data is associated through the same `splitId`;
- multi-parallel progress is represented as per-split progress, or the aggregate omits a singular `currentSplit/currentTable`;
- aggregate counts distinguish assigned, completed, running, and not-yet-chunked tables;
- any value that cannot be derived exactly is `BEST_EFFORT`/unknown rather than `EXACT`.

## 2. `startupOffset` is not the last completed checkpoint position

Location:

- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/progress/CdcProgressSnapshots.java:74-105`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/reader/IncrementalSourceReader.java:123-126,237-252`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/reader/IncrementalSourceRecordEmitter.java:144-152`

The helper uses the same `split.getStartupOffset()` for `currentConsumedPosition`, `lastCheckpointedPosition`, and `rawPosition`, and marks all related groups as `EXACT`.

In the real lifecycle, the offset advances as records/heartbeats are emitted. `snapshotState(checkpointId)` captures a candidate state at the checkpoint barrier, while `notifyCheckpointComplete(checkpointId)` happens later and the checkpoint may fail. Therefore the helper cannot know the last completed checkpoint position from an `IncrementalSplit` alone.

The safer lifecycle would be:

```text
record/heartbeat emitted
  -> update currentConsumedPosition

snapshotState(checkpointId)
  -> retain checkpoint candidate for that id

notifyCheckpointComplete(checkpointId)
  -> promote the candidate to lastCheckpointedPosition

restore/addSplits
  -> record restoredFromPosition separately
```

Until those hooks are wired, `CHECKPOINT_PROGRESS` should be `UNSUPPORTED`, and the startup offset should not be presented as the last completed checkpoint.

## 3. The public API contract is being frozen before the runtime contract is validated

Location:

- `seatunnel-api/src/main/java/org/apache/seatunnel/api/cdc/CdcProgressSnapshot.java:34-50`
- `seatunnel-api/src/main/java/org/apache/seatunnel/api/cdc/CdcSnapshotProgress.java:34-42`
- `seatunnel-api/src/main/java/org/apache/seatunnel/api/cdc/CdcIncrementalProgress.java:34-41`
- `seatunnel-api/src/main/java/org/apache/seatunnel/api/cdc/CdcCheckpointProgress.java:34-39`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/progress/MySqlCdcProgress.java:33-39`

At the current head, there is no production reader/enumerator/engine/job-detail path publishing these snapshots; outside the new tests, the MySQL wrapper is not called by the running source lifecycle. This means the public API has not yet been validated against checkpoint completion, restore, runtime identity enrichment, or cross-task aggregation.

There are also conflicting sources of truth: support level exists both in the top-level `supportLevels` map and inside the nested progress objects, so callers can construct contradictory states. `jobId/pipelineId/vertexId` enrichment is not defined, and `CdcProgressPosition` currently exposes connector-internal offset keys without a position type or schema version.

My preferred direction is to separate:

1. a connector/reader-side report containing only facts the connector actually owns; and
2. an engine/API snapshot that adds runtime identity, observation time, and completed-checkpoint facts.

Then keep one canonical support-level source and explicitly map stable external position fields per connector. If this must land in slices, please keep the model internal or mark the public types `@Experimental` until at least one real reader + checkpoint-complete + engine aggregation path validates the contract.

## 4. The tests currently encode states that the production path does not create

Location:

- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/test/java/org/apache/seatunnel/connectors/cdc/base/source/progress/CdcProgressSnapshotsTest.java:46-104,107-142`

The snapshot test uses `LinkedHashMap`, which hides the unordered production behavior, and the completed watermark key does not belong to the assigned split keys. In the real assigner, a completed watermark comes from an assigned split. The incremental test also asserts that the same offset is both current and last-checkpointed, which locks in the lifecycle issue above.

Please add coverage driven by the real lifecycle for:

- multiple running/completed splits and splitId-to-watermark association;
- `remainingTables` and lazy chunk creation;
- consumption advancing past checkpoint N;
- checkpoint candidate, failure/completion, and checkpoint N+1;
- restore and `restoredFromPosition`;
- `CATCH_UP -> INCREMENTAL`;
- serialization round trips and invalid phase/payload/support combinations.

I ran the focused Java 8 tests locally: the 3 CDC base tests and 1 MySQL test all pass. They prove the DTO construction works, but they do not exercise the lifecycle conditions above. Full Spotless also completed successfully with no source changes. The current GitHub `Build` check is still failing; I could only confirm that as a separate merge gate and do not want to guess at its root cause without usable job logs.

## 5. Please document the public contract before exposing it

The new public state fields and the non-trivial helper methods need field/method-level Javadocs covering units, nullability, phase/payload invariants, support-level precedence, raw-position stability, and the exact meaning of `EXACT`. This is especially important because these classes live in `seatunnel-api` and will be difficult to change once consumers adopt them.

The high-level goal is still worth pursuing, and I appreciate the effort to make the model immutable and connector-neutral. The main adjustment I am asking for is to make the API follow verified runtime facts, rather than deriving lifecycle state from snapshots that do not contain enough information. Once these points are addressed, please ping me—I will gladly re-review the full updated head promptly.

DanielLeens:
Rechecked the same unchanged head `a1f709208e59` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
record/heartbeat emitted
  -> update currentConsumedPosition

snapshotState(checkpointId)
  -> retain checkpoint candidate for that id

notifyCheckpointComplete(checkpointId)
  -> promote the candidate to lastCheckpointedPosition

restore/addSplits
  -> record restoredFromPosition separately
```

## Merge conclusion

1. Blocking items

2. Current gate update
- The CI / merge-gate fact I referenced earlier has changed. The current status is `PENDING gate: Build`: https://github.com/apache/seatunnel/runs/86985967620

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head again against the actual CDC enumerator/source lifecycle rather than only the DTO surface.

I also want to correct my own earlier positive signal on older rounds: after walking the real `SnapshotSplitAssigner` / incremental reader lifecycle more carefully, I do think the current helper layer still over-claims precision and freezes the public API too early.

# What this PR solves
- User pain: CDC jobs do not yet have a shared model for snapshot/incremental/checkpoint/raw-position progress reporting.
- Fix approach: add CDC progress model classes in `seatunnel-api`, helper builders in CDC base, and a MySQL adapter.
- One-line summary: the model layer is tidy, but the current helpers can still publish lifecycle-inaccurate values while marking them as `EXACT`, so I do not think this revision is ready to merge.

# Runtime path I checked
```text
snapshot phase helper
  -> CdcProgressSnapshots.forSnapshotPhase(...)
      -> currentSnapshotSplit()
      -> currentCompletedWatermark()
      -> assemble one singular current split/table/watermark view

incremental phase helper
  -> CdcProgressSnapshots.forIncrementalSplit(...)
      -> split.getStartupOffset()
      -> reuse it as currentConsumedPosition
      -> reuse it again as lastCheckpointedPosition
      -> reuse it again as rawPosition
```

# Findings
Issue 1: snapshot helper can combine one split identity with another split's watermark while still marking the result `EXACT`
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/progress/CdcProgressSnapshots.java:44-71,135-175`
- Why this is a blocker:
  `currentSnapshotSplit()` picks the first assigned split, while `currentCompletedWatermark()` independently picks the first completed watermark entry. There is no `splitId` join between them. In the real multi-split lifecycle, that can publish table/split A together with watermark data from split B, but the result is still labeled `EXACT`.
- Risk:
  user-facing progress can be wrong even when the API claims exactness.
- Severity: High

Issue 2: incremental helper uses `startupOffset` as current consumed position and last completed checkpoint position, which does not match the real reader/checkpoint lifecycle
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/progress/CdcProgressSnapshots.java:74-105`
- Why this is a blocker:
  a running reader advances beyond `startupOffset`, and a completed checkpoint is not the same thing as the initial startup anchor. The helper currently exposes the same offset for all three concepts and marks them `EXACT`.
- Risk:
  callers will consume a false checkpoint/progress contract from a public API type.
- Severity: High

Issue 3: this is being frozen in `seatunnel-api` before the runtime publishing contract is actually validated end to end
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/cdc/CdcProgressSnapshot.java:32-45` and related new CDC API classes
- Why this is a blocker:
  the public API contract is additive and attractive to downstream consumers, but the current PR still does not wire a real engine/runtime publishing path that validates checkpoint-complete, restore, and cross-task aggregation semantics.
- Risk:
  once consumers adopt these API types, correcting the contract later becomes much harder.
- Severity: High

Issue 4: the tests mostly prove DTO assembly, not real lifecycle correctness
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/test/java/org/apache/seatunnel/connectors/cdc/base/source/progress/CdcProgressSnapshotsTest.java:46-142`
- Why this matters:
  the test uses ordered maps and synthetic combinations that do not prove the real split-to-watermark association or checkpoint-complete lifecycle.
- Severity: Medium

# Merge conclusion

1. Blocking items
- Issue 1: join snapshot identity and watermark by the same split lifecycle instead of selecting them independently.
- Issue 2: stop exposing `startupOffset` as the last completed checkpoint position.
- Issue 3: either validate the runtime contract before freezing this in `seatunnel-api`, or reduce the public/stability claim for the current slice.

2. Suggested follow-up
- Issue 4: add lifecycle-driven tests that match the real multi-split and checkpoint-complete behavior.

The overall direction is still valuable, but the current revision is not yet safe enough for an API-level merge from Daniel's side.

goutamadwant:
I reworked this based on the review feedback and moved the updated implementation to #11512.

The new PR wires the model into the runtime and corrects the split watermark checkpoint and execution attempt handling discussed here.

Closing this one to keep the review on a single implementation.

PR #6917, opened 2024-05-28
[FIX][Connector-V2]Fix Date partition column error

dik111:
### Purpose of this pull request

to fix #6916

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update change log that in connector document. For more details you can refer to [connector-v2](https://github.com/apache/seatunnel/tree/dev/docs/en/connector-v2)
  2. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  3. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
* [ ] Update the [`release-note`](https://github.com/apache/seatunnel/blob/dev/release-note.md).

PR #11396, opened 2026-07-10
[Fix][Connector-V2][Jdbc] Fix DATE partition column splitting

### LINKED ISSUE #6916: [Bug] [Connector-V2] JDBC-SOURCE: SAP DBTech JDBC: Cannot convert Java type java.math.BigDecimal to SQL type DATE
### Search before asking

### What happened

When I try to use seatunnel to synchronize SAP hana data to doris, when partition_column is of date type, an error will be reported. 
I checked the code and found that the split value will be converted into timestamp, for example: `2024-02-06 07:12:00` is converted to `1707174720003`, then the (param instanceof BigDecimal) judgment will be entered in the `createNumberColumnSplitStatement` method. I think we should first determine whether the value is a timestamp. If it is a timestamp, then `statement.setTimestamp`

``` java
// here is the source code
else if (param instanceof BigDecimal) {
                statement.setBigDecimal(i + 1, (BigDecimal) param);
            }
```

### SeaTunnel Version

2.3.5

### SeaTunnel Config

```conf
env {
    parallelism = 8
    job.mode = "BATCH"
    flink.taskmanager.memory.process.size= "8G"
    flink.taskmanager.memory.framework.off-heap.size = "1G"
  }
  source {
      Jdbc {
          url = "jdbc:sap://xxx?reconnect=true"
          driver = "com.sap.db.jdbc.Driver"
          user = "xx"
          password = "xx"
          fetch_size = 10000
          query = "SELECT ACCOUNT_NUMBER,CREATION_DATE,ORG_ID FROM xxx WHERE CREATION_DATE between  TO_DATE('2024-01') and TO_DATE('2024-05')"
          partition_column = "CREATION_DATE"
          
      }
  }

  sink {
    Doris {
            fenodes = "xx:8030"
            username = root
            password = "xx"
            database = "model"
            table = "xx"
            data_save_mode = "DROP_DATA"
            doris.config ={
                format = "json"                                
         read_json_by_line = "true"  
         columns = "ACCOUNT_NUMBER,CREATION_DATE,ORG_ID" 
            }
        }
  }
```

### Running Command

```shell
/data/software/seatunnel/seatunnel-2.3.5/bin/start-seatunnel-flink-15-connector-v2.sh \
--master yarn-per-job \
--config /data/software/seatunnel/seatunnel-2.3.5/config/hana-doris-flink.cnf
```

### Error Exception

```log
org.apache.seatunnel.connectors.seatunnel.jdbc.exception.JdbcConnectorException: ErrorCode:[JDBC-04], ErrorDescription:[Connector database failed] - open() failed.SAP DBTech JDBC: Cannot convert Java type java.math.BigDecimal to SQL type DATE.
	at org.apache.seatunnel.connectors.seatunnel.jdbc.internal.JdbcInputFormat.open(JdbcInputFormat.java:104) ~[connector-jdbc-2.3.5.jar:2.3.5]
	at org.apache.seatunnel.connectors.seatunnel.jdbc.source.JdbcSourceReader.pollNext(JdbcSourceReader.java:67) ~[connector-jdbc-2.3.5.jar:2.3.5]
	at org.apache.seatunnel.translation.flink.source.FlinkSourceReader.pollNext(FlinkSourceReader.java:81) ~[seatunnel-flink-15-starter.jar:2.3.5]
	at org.apache.flink.streaming.api.operators.SourceOperator.emitNext(SourceOperator.java:385) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.streaming.runtime.io.StreamTaskSourceInput.emitNext(StreamTaskSourceInput.java:68) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.streaming.runtime.io.StreamOneInputProcessor.processInput(StreamOneInputProcessor.java:65) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.streaming.runtime.tasks.StreamTask.processInput(StreamTask.java:542) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.runMailboxLoop(MailboxProcessor.java:231) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.streaming.runtime.tasks.StreamTask.runMailboxLoop(StreamTask.java:831) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:780) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.runtime.taskmanager.Task.runWithSystemExitMonitoring(Task.java:935) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.runtime.taskmanager.Task.restoreAndInvoke(Task.java:914) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:728) ~[flink-dist-1.16.1.jar:1.16.1]
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:550) ~[flink-dist-1.16.1.jar:1.16.1]
	at java.lang.Thread.run(Thread.java:750) ~[?:1.8.0_381]
Caused by: com.sap.db.jdbc.exceptions.JDBCDriverException: SAP DBTech JDBC: Cannot convert Java type java.math.BigDecimal to SQL type DATE.
	at com.sap.db.jdbc.exceptions.SQLExceptionSapDB._newInstance(SQLExceptionSapDB.java:215) ~[ngdbc-2.20.11.jar:2.20.11-354e45de0dcbabd02f58c506ecf1138161ee4b1e]
	at com.sap.db.jdbc.exceptions.SQLExceptionSapDB.newInstance(SQLExceptionSapDB.java:26) ~[ngdbc-2.20.11.jar:2.20.11-354e45de0dcbabd02f58c506ecf1138161ee4b1e]
	at com.sap.db.jdbc.converters.AbstractConverter._newSetException(AbstractConverter.java:868) ~[ngdbc-2.20.11.jar:2.20.11-354e45de0dcbabd02f58c506ecf1138161ee4b1e]
	at com.sap.db.jdbc.converters.AbstractConverter.setBigDecimal(AbstractConverter.java:644) ~[ngdbc-2.20.11.jar:2.20.11-354e45de0dcbabd02f58c506ecf1138161ee4b1e]
	at com.sap.db.jdbc.PreparedStatementSapDB._setBigDecimal(PreparedStatementSapDB.java:2958) ~[ngdbc-2.20.11.jar:2.20.11-354e45de0dcbabd02f58c506ecf1138161ee4b1e]
	at com.sap.db.jdbc.PreparedStatementSapDB.setBigDecimal(PreparedStatementSapDB.java:1118) ~[ngdbc-2.20.11.jar:2.20.11-354e45de0dcbabd02f58c506ecf1138161ee4b1e]
	at org.apache.seatunnel.connectors.seatunnel.jdbc.source.FixedChunkSplitter.createNumberColumnSplitStatement(FixedChunkSplitter.java:207) ~[connector-jdbc-2.3.5.jar:2.3.5]
	at org.apache.seatunnel.connectors.seatunnel.jdbc.source.FixedChunkSplitter.createSplitStatement(FixedChunkSplitter.java:100) ~[connector-jdbc-2.3.5.jar:2.3.5]
	at org.apache.seatunnel.connectors.seatunnel.jdbc.source.ChunkSplitter.generateSplitStatement(ChunkSplitter.java:120) ~[connector-jdbc-2.3.5.jar:2.3.5]
	at org.apache.seatunnel.connectors.seatunnel.jdbc.internal.JdbcInputFormat.open(JdbcInputFormat.java:98) ~[connector-jdbc-2.3.5.jar:2.3.5]
	... 14 more
```

FLINK 1.16.1

### Java or Scala Version

java 8
scala 2.12

### Screenshots

_No response_

### Code of Conduct

JacobZheng0927:

Fixes #6916.

The JDBC fixed chunk splitter converted `DATE` partition bounds to numeric values and later bound those values as `BigDecimal`. JDBC drivers either reject that conversion or compare the `DATE` column with a numeric parameter incorrectly, which can make every generated split return zero rows.

This patch performs `DATE` range arithmetic in epoch days and converts generated split boundaries back to `java.sql.Date` before binding them. Numeric partition columns keep the existing `BigDecimal` behavior.

Unlike the closed approach in #6917, this change does not convert every integral `BigDecimal` parameter to a timestamp, so numeric partition columns are unaffected.

Yes. JDBC source fixed splitting now binds `DATE` partition boundaries with `PreparedStatement#setDate`, allowing `DATE` partition columns to return the expected rows.

Added unit coverage for generated `DATE` split boundaries and JDBC parameter binding.

```shell
./mvnw -pl seatunnel-connectors-v2/connector-jdbc \
  -DfailIfNoTests=false \
  -Dtest=FixedChunkSplitterTest test
```

Result: 6 tests run, 0 failures, 0 errors, 0 skipped.

### Check list

* [x] No documentation update is required; this fixes existing documented `DATE` partition-column support.
* [x] No new connector is added, so plugin mapping, distribution POM, labels, plugin config, and new-connector E2E updates are not applicable.


DanielLeens (commented):
Thanks for your first contribution to Apache SeaTunnel! This is a well-targeted fix for a real problem — binding `BigDecimal` to a `DATE` partition column causes JDBC drivers to either reject the parameter or compare incorrectly, leading to splits that return zero rows.

The approach of doing date range arithmetic in epoch days (as `BigDecimal`) and then converting back to `java.sql.Date` before binding the split boundaries is clean and correct. I like that you also handled `TIME` and `TIMESTAMP` in `convertSplitBoundary`, and the `parsePartitionBound` method with `DateTimeParseException` fallback is a nice defensive touch.

The unit tests are good — `testCreateDateColumnSplitsUseSqlDateBoundaries` verifies the split boundaries are `java.sql.Date` objects (not `BigDecimal`), and `testCreateDateSplitStatementBindsSqlDateParameters` verifies the actual `PreparedStatement.setDate` is called instead of `setBigDecimal`.

No source-level blockers from my side. The Build check currently shows `action_required` — this is expected for first-time contributors. Once a maintainer approves the workflow run, CI will proceed.

Looking forward to seeing this merged!

dybyte:
Please enable CI following [the instruction](https://github.com/apache/seatunnel/pull/11396/checks?check_run_id=86301794072)

DanielLeens:
Good job.

From Daniel's source-review side, I do not see a new code-side blocker on the current head. The current remaining gate is that the PR workflow is still `action_required`, so could you please enable CI workflows in your forked repository first? You can use this check link as the exact reference for the workflow that needs to be enabled:
https://github.com/apache/seatunnel/runs/86301794072

If you run into any trouble enabling CI or hit a follow-up workflow error after that, feel free to leave a comment here. Happy to take another look once the checks are running.

DanielLeens:
Rechecked the current head `9dd184d69e7c` against the latest `dev` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
same runtime path as the previous Daniel review on this unchanged head
```

## Merge conclusion

1. Blocking items
- This branch is behind upstream (`compare status: diverged`, `ahead_by=1`, `behind_by=11`), and the current failing gate is `FAILURE gate: Build`: https://github.com/apache/seatunnel/runs/86301794072

2. Lowest-cost next step

DanielLeens (commented):
Rechecked the unchanged head. The previous failing-Build instruction is outdated: the current `Build` is `ACTION_REQUIRED`, so workflow approval/retrigger is the immediate CI step. I do not see a reopened source-side blocker; after the workflow runs, repository review is still required.

DanielLeens:

davidzollo:

```

```

DanielLeens:
Checking in on the CI signal here since it looks like it flipped state. On this unchanged head (`9dd184d69e7c`), the `Build` check is currently showing `ACTION_REQUIRED` again (https://github.com/apache/seatunnel/runs/86301794072), rather than the `FAILURE` I recorded on July 14. This state usually means the workflow run needs a maintainer to approve/retrigger it before it can actually execute, so it is not something resolvable from the fork side alone.

From Daniel's source-review side there is still no reopened code blocker on this head. Once a maintainer approves the pending workflow run and it completes, I'll take another look at the real result.

@davidzollo (or another maintainer with the right permissions) — could someone approve the pending workflow run when convenient so this can move forward?

DanielLeens (changes requested):
Sorry for missing this in earlier rounds — that's on me. This head (`19830f75b`) looks like a rebase of the exact same `parsePartitionBound` / `convertSplitBoundary` design I already praised on July 10 without raising any blocker (the old head is no longer fetchable after the rebase, so I can't produce a literal diff, but the method names, structure, and even the defensive `DateTimeParseException` fallback match my July 10 description word-for-word). Since I can't fully rule out small changes, I re-reviewed the whole file from scratch rather than assuming the earlier "no blockers" conclusion still holds — and on this closer pass I found two real edge-case gaps in the new conversion helpers, plus a test-coverage gap, that I should have caught the first time. Treat these as carryover items, not regressions newly introduced by this rebase.

**User pain point:** When a JDBC source is configured with `split.size`/`partition_num` parallel splitting and the split key (`partition_column`) is a `DATE` (or `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`) column, `FixedChunkSplitter` computed split boundaries as `BigDecimal` and bound them to the split query with `PreparedStatement.setBigDecimal(...)`. Most JDBC drivers either reject a `BigDecimal` bound against a `DATE`/`TIME`/`TIMESTAMP` column or silently mis-compare it, so the generated splits return zero rows and the job silently loses data instead of failing loudly.

**Fix approach:** The splitter now tracks the split key's `SqlType` end-to-end. Partition bounds (whether user-configured via `partition_lower_bound`/`partition_upper_bound` or auto-discovered via `MIN()/MAX()`) are parsed into a type-aware `BigDecimal` (epoch-day for `DATE`, epoch-millis for `TIME`/`TIMESTAMP`, unchanged for everything else), the existing `JdbcNumericBetweenParametersProvider` range math still runs purely on that `BigDecimal`, and the two output boundaries are converted back to `java.sql.Date`/`Time`/`Timestamp` right before being stored on `JdbcSourceSplit`, so the eventual `PreparedStatement.setDate/setTime/setTimestamp` calls receive the correct JDBC type.

**One-sentence summary:** This turns "DATE/TIME/TIMESTAMP partition splitting silently returns 0 rows" into "DATE/TIME/TIMESTAMP partition splitting works correctly," by keeping the split-range arithmetic in `BigDecimal` but making the *binding* type-aware again.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Core changes:**
- `FixedChunkSplitter.java`
  - `getJdbcSourceSplits()` (L201-219): now calls the new `parsePartitionBound(...)` instead of `new BigDecimal(table.getPartitionStart())` directly, and passes `splitKeyType` into `findSplitColumnRange(...)`.
  - `createNumberColumnSplits()` (L398-429): wraps `parameterValues[i][0]`/`[1]` with the new `convertSplitBoundary(splitKeyType, ...)` before constructing each `JdbcSourceSplit`.
  - New `parsePartitionBound(String, SeaTunnelDataType)` (L536-548): for `DATE`, tries `LocalDate.parse(bound).toEpochDay()`; on `DateTimeParseException`, falls back to `new BigDecimal(bound)` (preserves the old raw-epoch-day workaround for backward compatibility). All other types are unchanged.
  - New `convertSplitBoundary(SeaTunnelDataType, Serializable)` (L550-565): converts a `BigDecimal` epoch value back to `Date`/`Time`/`Timestamp` via `longValueExact()`, for `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` only.
  - `convertToBigDecimal(Object, SeaTunnelDataType)` (L567-577): new overload used by the auto min/max path; adds a `DATE`-specific branch that also tolerates the min/max query returning a `Timestamp` for a `DATE` column (happens with some dialects/drivers), truncating to `LocalDate` first.

**Before / after (binding path):**
```java
// before
Serializable[][] parameterValues = jdbcNumericBetweenParametersProvider.getParameterValues();
...
split.getSplitStart() -> parameterValues[i][0]  // always BigDecimal
...
// createNumberColumnSplitStatement(): BigDecimal branch always wins for a DATE column
} else if (param instanceof BigDecimal) {
    statement.setBigDecimal(i + 1, (BigDecimal) param);   // wrong JDBC type for DATE/TIME/TIMESTAMP
```
```java
// after
convertSplitBoundary(splitKeyType, parameterValues[i][0])  // now Date/Time/Timestamp for temporal keys
...
} else if (param instanceof Date) {
    statement.setDate(i + 1, (Date) param);               // now hits this branch instead
```

**Key findings:**
1. The normal path (no bounds configured, or valid ISO-8601 bounds) is genuinely fixed: `DATE` splits now bind through `setDate`, not `setBigDecimal` — verified by the new `testCreateDateSplitStatementBindsSqlDateParameters` which asserts `bigDecimalParameters` is empty and `dateParameters` holds the right values.
2. The fix's scope is actually broader than the PR title suggests: `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` split keys had exactly the same `setBigDecimal`-on-a-temporal-column bug before this PR (nothing converted their boundaries either), and `convertSplitBoundary`/`convertToBigDecimal` fix all of them uniformly. That's a nice bonus, but only `DATE` is covered by the new unit tests (see Issue 3).
3. It's a precise fix, not a workaround: the split-range arithmetic itself (`JdbcNumericBetweenParametersProvider`) is untouched and still operates purely on integral `BigDecimal` values (epoch days for `DATE`), so the existing inclusive `>=`/`<=` range semantics are preserved exactly — no off-by-one is introduced by this change, since `LocalDate.parse(...).toEpochDay()` and `Date.valueOf(LocalDate.ofEpochDay(...))` are exact, whole-number round trips and the provider only ever adds/subtracts whole-number `BigDecimal.valueOf(long)` amounts.
4. Two genuine new edge-case gaps exist, but only on the *misconfiguration* path (Issues 1–2 below) — not on the normal auto-discovery or valid-ISO-date path.
5. Test coverage proves the `DATE` happy path end-to-end (range computation + statement binding) but leaves the `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` conversion branches and both `parsePartitionBound` fallback branches unverified (Issue 3).

**In-depth correctness analysis (call chain):**
```text
JdbcSourceSplitEnumerator -> ChunkSplitter.generateSplits(table)
  -> FixedChunkSplitter.createSplits(table, splitKey)                       [L64-119]
      -> splitKeyType = splitKey.getFieldType(0)   // e.g. LocalTimeType.LOCAL_DATE_TYPE -> SqlType.DATE
      -> getJdbcSourceSplits(table, splitKeyName, splitKeyType)             [L201-219]
          -> partitionStart = parsePartitionBound(table.getPartitionStart(), splitKeyType)  [L536]
              -- user configured "2024-01-01" -> LocalDate.parse succeeds -> epoch day BigDecimal
              -- user configured "19000" (legacy raw epoch day) -> LocalDate.parse fails -> BigDecimal("19000") fallback
              -- user configured "2024-01-01 00:00:00" or garbage -> LocalDate.parse fails -> new BigDecimal(...) throws
                 uncaught NumberFormatException (Issue 1)
          -- if either bound is null -> findSplitColumnRange(table, columnName, splitKeyType)  [L521-534]
              -> queryMinMax() -> ResultSet.getObject(1/2)   // driver-dependent: Date or (some dialects) Timestamp
              -> convertToBigDecimal(min/max, splitKeyType)  [L567-577] -- epoch-day for Date/Timestamp-as-DATE
          -> createNumberColumnSplits(table, splitKeyName, splitKeyType, partitionStart, partitionEnd)  [L398-429]
              -> JdbcNumericBetweenParametersProvider(partitionStart, partitionEnd).ofBatchNum(partitionNumber)
                  -> getParameterValues(): pure whole-number BigDecimal arithmetic, inclusive [start,end] pairs
              -> convertSplitBoundary(splitKeyType, parameterValues[i][0/1])  [L550-565]
                  -- normal case: boundary is a whole-number BigDecimal -> longValueExact() succeeds
                  -- edge case: boundary carries a fractional remainder inherited from the parsePartitionBound
                     numeric fallback above -> longValueExact() throws uncaught ArithmeticException (Issue 2)
              -> new JdbcSourceSplit(..., splitStart=Date/Time/Timestamp, splitEnd=Date/Time/Timestamp)

Read phase
  -> FixedChunkSplitter.createSplitStatement(split, schema)                 [L221-235]
      -> createNumberColumnSplitStatement(split)                           [L431-505]
          -> parameterValues = {split.getSplitStart(), split.getSplitEnd()}
          -> instanceof Date/Time/Timestamp -> statement.setDate/setTime/setTimestamp(...)   // fixed by this PR
```
Both new edge cases above require the user to explicitly set `partition_lower_bound`/`partition_upper_bound` to a value that is neither a valid ISO-8601 date nor a clean integer for a temporal split key — the default auto-discovery path and the documented `partition_lower_bound = "2024-01-01"` usage are unaffected and correct.

## 1.2 Compatibility Impact

**Fully compatible.**
- No config option is renamed, removed, or defaulted differently. `partition_lower_bound`/`partition_upper_bound` keep accepting the historical raw-numeric string for `DATE` columns via the `parsePartitionBound` fallback (`new BigDecimal(bound)` when `LocalDate.parse` fails), so any existing pipeline that was working around the old bug by passing a raw epoch-day number continues to work identically.
- `JdbcSourceSplit.splitStart`/`splitEnd` are typed as `Object`/`Serializable`, so storing `java.sql.Date`/`Time`/`Timestamp` there instead of `BigDecimal` for temporal split keys does not change the class's field/serialization contract (`java.sql.Date/Time/Timestamp` are all `Serializable`).
- The behavior change itself — DATE/TIME/TIMESTAMP splits going from "silently returns 0 rows" to "returns the correct rows" — is a bug fix, not an API/protocol change, so no entry in `incompatible-changes.md` is needed.

- No new I/O, no new locks, no new threads. `parsePartitionBound`/`convertSplitBoundary`/`convertToBigDecimal` are cheap, allocation-light, called only once per table during split enumeration (not per-row, not on any hot read path).
- `LocalDate.parse` + `toEpochDay()` and `LocalDate.ofEpochDay` + `Date.valueOf` are all standard-library, no external dependency added.
- No retries, no idempotency concerns introduced — split enumeration is deterministic given the same `MIN()/MAX()` result.

- **Issue 1 (Medium):** `parsePartitionBound` swallows `DateTimeParseException` for `DATE` columns and blindly retries `new BigDecimal(bound)` (`FixedChunkSplitter.java:536-548`). If the configured `partition_lower_bound`/`partition_upper_bound` is neither a valid ISO-8601 date string nor a valid `BigDecimal` string (e.g. `"2024-01-01 00:00:00"`, or any other common-but-non-ISO date format a user might reasonably try), the fallback throws an **uncaught `NumberFormatException`** with a message like `Character 2 is neither a decimal digit number...` — confusing for a user who just wants to know "your date format is wrong." *Fix:* catch the fallback's `NumberFormatException` too and rethrow a `JdbcConnectorException` (`ILLEGAL_ARGUMENT`) naming the offending config key/value and the expected `yyyy-MM-dd` format.
- **Issue 2 (Medium):** `convertSplitBoundary` calls `boundary.longValueExact()` for `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` (`FixedChunkSplitter.java:550-565`). This is new code (the pre-fix path never called `longValueExact()` on these boundaries — it bound the raw `BigDecimal` directly). If a temporal `partition_lower_bound`/`partition_upper_bound` string parses via the `parsePartitionBound` numeric fallback into a *fractional* `BigDecimal` (e.g. a stray `"19000.5"`), that fractional remainder survives the provider's whole-number arithmetic unchanged and propagates into every generated split boundary, so `longValueExact()` throws an **uncaught `ArithmeticException: Rounding necessary`** deep inside split planning rather than a clear config-validation error at startup. It fails safe/loud (no data corruption, no silent zero-row splits — this isn't a regression versus the pre-fix behavior, just a rough edge on the new code), but the diagnostic is poor. *Fix:* validate in `parsePartitionBound` that the numeric fallback for a temporal split key is a whole number (`stripTrailingZeros().scale() <= 0`), and throw a `JdbcConnectorException` immediately if not, instead of letting the exception surface later out of `longValueExact()`.
- No sensitive data is logged; `log.info("Number column split - params: ...")` logging the split boundaries is fine and unchanged in spirit from before.

**2.1 Coding Standards / documentation of new methods:** `parsePartitionBound`, `convertSplitBoundary`, and the new `convertToBigDecimal(Object, SeaTunnelDataType)` overload are all new private methods with real branching logic and no Javadoc/explanatory comment at all. Per the project's documentation-completeness rule, please add a short comment on each explaining the epoch-day/epoch-millis unit convention and the backward-compatible fallback behavior — this is exactly the kind of "why" (unit convention, compatibility reason) that isn't obvious from the code alone and will save the next person a repeat of this same review.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Naming and structure follow the existing file's conventions well (mirrors `convertToBigDecimal(Object)` naming pattern, keeps the private-method-at-bottom layout). See the documentation gap called out above under 1.4 — that's the only standards issue found.

**Issue 3 (Medium):** The two new tests (`testCreateDateColumnSplitsUseSqlDateBoundaries`, `testCreateDateSplitStatementBindsSqlDateParameters`) only exercise the `DATE`/`LocalTimeType.LOCAL_DATE_TYPE` happy path. Not covered by any test in this PR:
- `convertSplitBoundary`'s `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` branches (`FixedChunkSplitter.java:558-562`), even though this PR is what makes those branches reachable/correct for the first time.
- `convertToBigDecimal(Object, SeaTunnelDataType)`'s `Timestamp`-returned-for-a-`DATE`-column branch (`FixedChunkSplitter.java:571-573`).
- `parsePartitionBound`'s ISO-parse-failure fallback branch (raw epoch-day string) and its unhandled-garbage-input failure mode (Issue 1).
- The `longValueExact()` fractional-boundary failure mode (Issue 2).

None of these are exotic — they're the exact new branches this PR adds — so I'd ask for at least one test per branch (a `TIME`/`TIMESTAMP` variant of the existing DATE test, plus one asserting the improved error from Issue 1's fix) before merge.

**Test-stability analysis (mandatory since this PR touches UT code):** Both new tests are self-contained and deterministic — `testCreateDateColumnSplitsUseSqlDateBoundaries` invokes the private `createNumberColumnSplits` via reflection with fixed `BigDecimal` inputs and asserts exact `java.sql.Date` equality; `testCreateDateSplitStatementBindsSqlDateParameters` uses an in-memory dynamic `Proxy` for `PreparedStatement` (no real DB, no network, no `Thread.sleep`, no shared static/mutable state, no timezone/clock dependency in the assertions themselves). **Stability rating: Stable.** No flaky-test anti-patterns found in the added tests.

## 2.3 Documentation Updates

`docs/en/connectors/source/Jdbc.md:166-167` and `docs/zh/connectors/source/Jdbc.md:166-167` already type `partition_lower_bound`/`partition_upper_bound` as generic `String` (not `BigDecimal`), so nothing is factually wrong — but neither doc mentions that a `DATE`-typed `partition_column` now accepts an ISO-8601 date string (e.g. `"2024-01-01"`) in addition to the legacy raw epoch-day number. This is a genuinely new, user-facing capability introduced by this PR and is worth one sentence in both `docs/en` and `docs/zh` so users don't have to read the source to discover the ISO-date-string support. (Per-connector docs like `Mysql.md`/`Oracle.md` still say `BigDecimal` and "only support numeric type" for `partition_column`, but that's pre-existing staleness unrelated to this diff — not something this PR needs to fix.) This is Low severity / non-blocking, but a quick win to add.

# 3. Architectural Soundness

**Precise fix.** Keeping the range-arithmetic in `JdbcNumericBetweenParametersProvider` untouched (still pure `BigDecimal`) and only adding a thin type-aware parse/convert layer at the two boundaries (config-string → `BigDecimal`, `BigDecimal` → JDBC bind type) is the right level of surgical change — it doesn't touch the shared numeric-splitting algorithm used by every other numeric type, so the blast radius is exactly the temporal-column case this PR targets.

## 3.2 Maintainability

The four new private methods are short and single-purpose; the main gap is the missing "why" comments discussed in 1.4/2.1 (unit convention: epoch-day vs epoch-millis, and why the numeric fallback exists for backward compatibility). Adding those comments would make this code trivially maintainable for the next contributor.

## 3.3 Extensibility

The `SqlType`-dispatch pattern in `convertSplitBoundary`/`convertToBigDecimal(Object, SeaTunnelDataType)` is a reasonable, reusable pattern if another temporal-like type needs the same treatment later — it's not hard-coded to `DATE` alone.

Confirmed compatible with historical versions: no restore/checkpoint-state format changes (split enumeration state isn't touched by this diff), and the legacy raw-numeric `partition_lower_bound`/`partition_upper_bound` workaround for `DATE` columns keeps working via the fallback branch. No upgrade action required.

# 4. Issue Summary

| 1 | Malformed (non-ISO, non-numeric) `partition_lower_bound`/`partition_upper_bound` for a `DATE` column throws an uncaught `NumberFormatException` instead of a clear config error | `FixedChunkSplitter.java:536-548` | Medium |
| 2 | `convertSplitBoundary`'s `longValueExact()` throws an uncaught `ArithmeticException` if a temporal partition bound resolves to a fractional `BigDecimal` via the numeric fallback | `FixedChunkSplitter.java:550-565` | Medium |
| 3 | New `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` conversion branches and both `parsePartitionBound` fallback branches are untested | `FixedChunkSplitter.java:558-577`, `FixedChunkSplitterTest.java` | Medium |
| 4 | `docs/en`/`docs/zh` `Jdbc.md` don't mention the new ISO-8601 date-string support for `DATE`-typed partition bounds | `docs/en/connectors/source/Jdbc.md:166-167`, `docs/zh/connectors/source/Jdbc.md:166-167` | Low |

# 5. Merge Recommendation

   - Issue 1: unclear error on malformed temporal partition-bound config — should fail with an actionable `JdbcConnectorException`, not a raw `NumberFormatException`.
   - Issue 2: unclear error (and, more importantly, un-tested behavior) on a fractional temporal boundary — should be validated with a clear error rather than surfacing `ArithmeticException` from deep inside split planning.
   - Issue 3: please add test coverage for the `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` conversion branches this PR newly makes reachable, and for the two `parsePartitionBound` fallback branches (including the improved-error case from fixing Issue 1).

   - Issue 4 (docs): add one sentence to `docs/en`/`docs/zh` `Jdbc.md` noting that `DATE`-typed partition bounds now accept ISO-8601 date strings. Doesn't block merge, but is a quick, valuable addition since it's a new user-facing capability.
   - Add short Javadoc/comments to the four new private methods explaining the epoch-day/epoch-millis unit convention and the backward-compatibility rationale for the numeric fallback (noted under 1.4/2.1).

**Overall assessment:** The core design is correct and I stand by my July review's assessment of the approach — doing range arithmetic in epoch-day `BigDecimal` and converting only at the parse/bind boundaries is the right fix, it doesn't perturb the shared numeric-splitting algorithm, and it's fully backward compatible. What's missing is hardening of the two new boundary-parsing/conversion helpers against malformed config input, and test coverage matching the actual scope of the fix (which quietly extends to `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`, not just `DATE`). No alternative implementation approach is needed — this is a matter of tightening the edges of the current design, not restructuring it.

**CI status:** The `Build` check is currently `pending` (actively running on this new commit, not blocked on workflow approval anymore) — no action needed from the contributor right now, just waiting on results.

Thanks again for the contribution and for your patience through the review cycles on this one — looking forward to getting this merged once the fixes above are in!

DanielLeens (changes requested):
Thanks for pushing the update so quickly! I re-reviewed the whole file from scratch against the new head (`f83cf6968f7`) rather than just diffing the new commit, and I'm happy to confirm all three code-side blockers and the docs suggestion from my last round are addressed. I did find one small, narrowly-scoped gap that only became reachable because of the fix itself — details below, but it's a one-line-pattern fix, not a redesign.

**User pain point:** When a JDBC source uses `split.size`/`partition_num` parallel splitting and the split key is a `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` column, `FixedChunkSplitter` used to bind split boundaries with `PreparedStatement.setBigDecimal(...)`. Most drivers either reject that against a temporal column or silently mis-compare it, so splits return zero rows and the job silently drops data instead of failing loudly.

**Fix approach:** The splitter now carries the split key's `SqlType` end to end. Partition bounds (whether user-configured via `partition_lower_bound`/`partition_upper_bound` or auto-discovered via `MIN()/MAX()`) are parsed into a type-aware `BigDecimal` (epoch-day for `DATE`, epoch-millis for `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`), the existing `JdbcNumericBetweenParametersProvider` range math still runs purely on that `BigDecimal`, and the two output boundaries are converted back to `java.sql.Date`/`Time`/`Timestamp` right before being stored on `JdbcSourceSplit`, so `PreparedStatement.setDate/setTime/setTimestamp` gets the correct JDBC type.

**One-sentence summary:** Turns "DATE/TIME/TIMESTAMP partition splitting silently returns 0 rows" into "DATE/TIME/TIMESTAMP partition splitting works correctly, and now fails loudly with a clear message on most misconfigured bounds."

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Precise description of this commit's changes** (all in `FixedChunkSplitter.java` unless noted):
- `parsePartitionBound(String, SeaTunnelDataType)` (L537-578): now computes a `temporalSplitKey` flag covering `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`; after the numeric fallback parse, it rejects any temporal bound whose `BigDecimal` has a non-zero fractional scale with a `JdbcConnectorException`, and wraps a `NumberFormatException` on a malformed `DATE` bound into a `JdbcConnectorException` naming the expected `yyyy-MM-dd` format.
- `FixedChunkSplitterTest.java`: five new tests covering `convertSplitBoundary`'s `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` branches, `convertToBigDecimal`'s `Timestamp`-for-`DATE` branch, `parsePartitionBound`'s ISO/legacy-epoch-day fallback, the improved `DATE` malformed-bound error, and the new fractional-bound rejection (for both `DATE` and `TIMESTAMP`).
- `docs/en` and `docs/zh` `Jdbc.md`: `partition_upper_bound`/`partition_lower_bound` rows now document ISO-8601 date-string support for `DATE` columns, consistently in both languages.

**Before / after (`parsePartitionBound` malformed-input path):**
```java
// before (the head I reviewed on Aug 20)
if (SqlType.DATE.equals(sqlType)) {
    try {
        return BigDecimal.valueOf(LocalDate.parse(bound).toEpochDay());
    } catch (DateTimeParseException ignored) {
        return new BigDecimal(bound);   // uncaught NumberFormatException on garbage input
    }
}
return new BigDecimal(bound);
```
```java
// after (current head)
try {
    BigDecimal numericBound = new BigDecimal(bound);
    if (temporalSplitKey && numericBound.stripTrailingZeros().scale() > 0) {
        throw new JdbcConnectorException(ILLEGAL_ARGUMENT, "Temporal partition bounds must be whole numbers: " + bound);
    }
    return numericBound;
} catch (NumberFormatException e) {
    if (SqlType.DATE.equals(sqlType)) {
        throw new JdbcConnectorException(ILLEGAL_ARGUMENT, "Invalid DATE partition bound '" + bound + "'. Expected yyyy-MM-dd or a whole epoch-day number.", e);
    }
    throw e;   // TIME/TIMESTAMP/TIMESTAMP_TZ still fall through here — see Issue 1
}
```

**Key findings:**
1. Confirmed fix, with evidence, for all three code-side items from my last round:
   - The `DATE` malformed-bound `NumberFormatException` is now wrapped in a `JdbcConnectorException` (line 569-574), verified by the new `testRejectInvalidDatePartitionBound`.
   - The fractional-boundary `ArithmeticException` in `convertSplitBoundary`'s `longValueExact()` can no longer be reached: `parsePartitionBound` now rejects any temporal bound with `scale() > 0` before it ever becomes a split boundary (line 561-565), and `JdbcNumericBetweenParametersProvider.getParameterValues()` (unchanged, `internal/split/JdbcNumericBetweenParametersProvider.java`) only ever adds/subtracts whole-number `BigDecimal.valueOf(long)` amounts on top of that, so the arithmetic itself cannot reintroduce a fraction. Verified by tracing both callers of `convertSplitBoundary` (only `createNumberColumnSplits`, `FixedChunkSplitter.java:424-425`) and by the new `testRejectFractionalTemporalPartitionBound`.
   - Test coverage now includes the `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` conversion branches, the `Timestamp`-for-`DATE` branch, and both `parsePartitionBound` fallback branches — exactly what I asked for.
2. **New finding (Issue 1 below):** in hardening `parsePartitionBound`, this commit introduces a `temporalSplitKey` concept that explicitly covers `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` for the *fractional-bound* check, but the malformed-string wrapping right below it (`catch (NumberFormatException e)`) still only special-cases `DATE` — `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` fall through to `throw e;` (line 576) and surface a raw, unwrapped `NumberFormatException`. This is a new inconsistency introduced by this very commit, not a carryover from my last review (which was scoped to the `DATE` case only).
3. It remains a precise fix, not a workaround — the shared numeric-splitting algorithm in `JdbcNumericBetweenParametersProvider` is still untouched, so the blast radius stays limited to the parse/bind boundary layer.

**Complete runtime path (misconfiguration path for Issue 1):**
```text
JdbcSourceSplitEnumerator -> ChunkSplitter.generateSplits(table)
  -> FixedChunkSplitter.createSplits(table, splitKey)
      -> splitKeyType = SqlType.TIMESTAMP (or TIME / TIMESTAMP_TZ)
      -> getJdbcSourceSplits(...) [L201-219]
          -> partitionStart = parsePartitionBound(table.getPartitionStart(), splitKeyType) [L543-578]
              -- user configures e.g. "2024-01-01T00:00:00" or any non-numeric string for a TIMESTAMP bound
              -- SqlType.DATE check at L553 is false -> skip LocalDate.parse -> straight to numeric parse [L560]
              -- new BigDecimal(bound) throws NumberFormatException
              -- catch block [L567]: SqlType.DATE.equals(sqlType) is false -> `throw e;` [L576]
              -- raw java.lang.NumberFormatException propagates out of split enumeration, uncaught/unwrapped
```
This is a misconfiguration-only path (same category as the original Issue 1/2 — auto-discovery and valid-input paths are unaffected), and it does fail loudly rather than losing data, so it's not a regression versus pre-PR behavior for these types. But it's a real, verifiable inconsistency in the very hardening this commit set out to do.

## 1.2 Compatibility Impact

**Fully compatible.** No config option renamed/removed/re-defaulted. The legacy raw-epoch-day fallback for `DATE` still works. `JdbcSourceSplit.splitStart`/`splitEnd` typing is unchanged (`Serializable`). No checkpoint/state format touched. This is a pure bug-fix/hardening change; no entry needed in `incompatible-changes.md`.

No new I/O, threads, or locks. The added `stripTrailingZeros()`/`scale()` check and the extra `JdbcConnectorException` construction are cheap, called once per table during split enumeration only — not on the read hot path. No retry/idempotency concerns introduced.

**Issue 1 (Medium):** `parsePartitionBound`'s malformed-input handling is inconsistent across the temporal types it now explicitly recognizes. `FixedChunkSplitter.java:567-577`: the `catch (NumberFormatException e)` block only wraps the exception into a `JdbcConnectorException` when `sqlType == DATE`; for `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` it re-throws the raw `NumberFormatException` at line 576. Since this same commit just taught `parsePartitionBound` to treat all four types as `temporalSplitKey` for the fractional-value check three lines above, the malformed-string case should get the same treatment for consistency and a good user-facing error message.
- **Potential risk:** Low — this only affects a hand-configured `partition_lower_bound`/`partition_upper_bound` on a `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` split key that's neither a whole number nor blank; it fails the job at split-enumeration time either way, so there's no data-loss or silent-failure risk, just a less actionable error message than the `DATE` case now gets.
- **Best improvement:** Extend the existing `if (SqlType.DATE.equals(sqlType))` check at line 568 to cover `temporalSplitKey` instead, with a message that doesn't assume the `yyyy-MM-dd` format (e.g. "Invalid temporal partition bound '<bound>'. Expected a whole epoch-day/epoch-millis number." for non-`DATE` types, keeping the more specific `yyyy-MM-dd` message for `DATE`). This is a one-branch change, no new tests needed beyond one more `assertThrows` case mirroring the existing `testRejectInvalidDatePartitionBound`.
- **Severity:** Medium

No sensitive data is logged or leaked anywhere in this diff.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The four private methods this PR touches now all carry explanatory comments (`parsePartitionBound`, `findSplitColumnRange`, `convertSplitBoundary`, `convertToBigDecimal`) — this directly addresses the documentation gap I raised last round. No other standards issues found.

Test coverage now matches the scope of the fix: `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` conversion, the `Timestamp`-for-`DATE` min/max branch, both `parsePartitionBound` fallback branches, the improved `DATE` error message, and fractional-bound rejection for both `DATE` and `TIMESTAMP` are all now exercised. The only residual gap is the one covered by Issue 1 above (no test for a malformed, non-numeric `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` bound) — expected, since the source code doesn't yet produce a distinct behavior to assert there.

**Test-stability analysis (mandatory, this PR touches UT code):** All five new tests invoke private methods via reflection with fixed, literal inputs (`BigDecimal.valueOf(...)`, `Date.valueOf("2024-01-03")`, `Timestamp.valueOf("2024-01-03 12:34:56")`) and assert exact equality or exception type/message content. No `Thread.sleep`, no real DB/network/container, no shared mutable/static state, no reliance on execution order, no timezone dependency (`java.sql.Date.toLocalDate()`/`Timestamp.valueOf` operate on the literal calendar fields, not epoch-millis-through-TimeZone conversion). **Stability rating: Stable.** No flaky-test anti-patterns found in the new tests, consistent with my rating of the original `DATE`-only tests last round.

## 2.3 Documentation Updates

`docs/en/connectors/source/Jdbc.md` and `docs/zh/connectors/source/Jdbc.md` now both document ISO-8601 date-string support for `DATE`-typed partition bounds, with matching wording in English and Chinese. This closes the doc gap from my last round. (Not blocking, but worth a future follow-up: since this commit also formally hardens `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` bound handling — even though it always worked internally — a one-line mention that those types expect a whole epoch-millis number would be a nice-to-have, not required for this PR.)

# 3. Architectural Soundness

**Precise fix.** The hardening stays exactly at the parse/bind boundary layer; the shared `JdbcNumericBetweenParametersProvider` range arithmetic is untouched.

## 3.2 Maintainability

Good — the new Javadoc comments explain the epoch-day/epoch-millis unit convention and the backward-compatibility rationale, which was the main maintainability gap last round.

## 3.3 Extensibility

The `SqlType`-dispatch pattern remains reusable if another temporal-like type needs the same treatment later.

Confirmed compatible: no checkpoint/state format changes, legacy raw-epoch-day `DATE` bounds still work via the fallback branch. No upgrade action required.

# 4. Issue Summary

| 1 | `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` malformed (non-numeric) partition bound still throws a raw, unwrapped `NumberFormatException` instead of a `JdbcConnectorException`, inconsistent with the `DATE` handling this same commit introduced | `FixedChunkSplitter.java:567-577` | Medium |

All four issues from my previous review (malformed `DATE` bound, fractional-boundary `ArithmeticException`, missing `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`/fallback-branch test coverage, missing docs) are confirmed fixed on this head — see 1.1/2.2/2.3 above for the evidence trail.

# 5. Merge Recommendation

   - Issue 1: extend the existing `DATE`-only malformed-bound wrapping to also cover `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`, for consistency with the fractional-bound check three lines above that already treats all four types uniformly. This is genuinely a small, mechanical change — one more `if`/message branch plus one test case mirroring `testRejectInvalidDatePartitionBound`.

   - None beyond the optional docs follow-up noted in 2.3 (not required for this PR).

**Overall assessment:** This update closes out all three code-side blockers and the docs suggestion from my last round with real, verifiable fixes — not just tests/docs bolted on top. I traced the fractional-boundary fix all the way through `JdbcNumericBetweenParametersProvider`'s arithmetic to confirm `longValueExact()` genuinely can't be reached with a non-integer value anymore, and the new tests match the branches they claim to cover. The one remaining item is narrow and low-risk (fails loud either way, just a less friendly message for three specific types), so I'd expect this to be a very quick turnaround before merge. No alternative implementation is needed here — just finishing the consistency pass this commit already started.

**CI status:** `Build` is currently `IN_PROGRESS` on this head (`f83cf6968f7`) — no action needed from the contributor right now, just waiting on results.

Thanks again for turning this around so fast — really appreciate the thoroughness on the test additions. Happy to take another look as soon as Issue 1 is in!

JacobZheng0927:
Thanks for the detailed review.

I addressed the requested items in the previous fix commit and applied the Spotless formatting corrections reported by CI in commit eb5195ae1:
- Added clear validation for malformed and fractional temporal partition bounds.
- Added coverage for DATE, TIME, TIMESTAMP, TIMESTAMP_TZ, fallback parsing, and invalid inputs.
- Added the requested comments and updated the English and Chinese JDBC documentation.

The previous Build run failed only because of the reported formatting violations; no code or test failure was reported. A new CI run is now pending. Could you please take another look once CI completes?

DanielLeens (changes requested):
Thanks @JacobZheng0927 for pushing `eb5195ae1a` so quickly. Per protocol for re-reviews I re-pulled the current head and re-read the whole diff from scratch rather than only diffing the new commit — but I can report that up front, the diff between the head I reviewed last round (`f83cf6968f`) and today's head (`eb5195ae1a`) is exactly what your comment says it is: a Spotless formatting pass only (a comment reflow in `convertSplitBoundary`'s Javadoc and reformatted method-call argument wrapping in `FixedChunkSplitterTest.java`), confirmed via `git diff f83cf6968f eb5195ae1a` — no logic changed. So this is not a new round of fixes; it's the same code I reviewed on 2026-08-20T02:30:25Z, now spotless-clean.

That means my one outstanding item from last round — Issue 1 below — is still present and unaddressed on the current head. Re-stating it here (not a new finding) so the issue tracker in this review stays self-contained, and re-confirming everything else that was already fixed stays fixed.

- **User pain point:** when a JDBC source uses `split.size`/`partition_num` parallel splitting and the split key is a `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` column, `FixedChunkSplitter` used to bind split boundaries with `PreparedStatement.setBigDecimal(...)`. Most drivers either reject that against a temporal column or silently mis-compare it, so the generated splits return zero rows and the job silently drops data instead of failing loudly.
- **Fix approach:** the splitter carries the split key's `SqlType` end to end. Partition bounds (user-configured via `partition_lower_bound`/`partition_upper_bound`, or auto-discovered via `MIN()/MAX()`) are parsed into a type-aware `BigDecimal` (epoch-day for `DATE`, epoch-millis for `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`); the existing `JdbcNumericBetweenParametersProvider` range math still runs purely on that `BigDecimal`; the two output boundaries are converted back to `java.sql.Date`/`Time`/`Timestamp` right before being stored on `JdbcSourceSplit`, so `PreparedStatement.setDate/setTime/setTimestamp` gets the correct JDBC type.
- **One-sentence summary:** turns "DATE/TIME/TIMESTAMP partition splitting silently returns 0 rows" into "DATE/TIME/TIMESTAMP partition splitting works correctly, and fails loudly with a clear message on most misconfigured bounds" — with one small consistency gap left in the "fails loudly" part.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Confirmed unchanged from my last round (`FixedChunkSplitter.java`, current head):
- `parsePartitionBound(String, SeaTunnelDataType)` (L543-578): computes `temporalSplitKey` for `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`; for `DATE` tries `LocalDate.parse(bound).toEpochDay()` first; falls through to a numeric parse that rejects any `temporalSplitKey` bound with a non-zero fractional scale (`JdbcConnectorException`, L561-565); wraps a `NumberFormatException` into a `JdbcConnectorException` only for `DATE` (L568-575) — **still re-throws the raw `NumberFormatException` unwrapped for `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`** (L576, `throw e;`).
- `convertSplitBoundary(SeaTunnelDataType, Serializable)` (L584-598) and `convertToBigDecimal(Object, SeaTunnelDataType)` (L601-612): unchanged, correct, verified again this round — `longValueExact()` can no longer throw, since `parsePartitionBound` now rejects fractional temporal bounds before they can reach split-boundary arithmetic, and `JdbcNumericBetweenParametersProvider` only adds/subtracts whole-number amounts.

**Runtime path for the still-open gap** (unchanged from my last round, re-verified against the current head):
```text
JdbcSourceSplitEnumerator -> ChunkSplitter.generateSplits(table)
  -> FixedChunkSplitter.createSplits(table, splitKey)
      -> splitKeyType = SqlType.TIMESTAMP (or TIME / TIMESTAMP_TZ)
      -> getJdbcSourceSplits(...) [L201-219]
          -> partitionStart = parsePartitionBound(table.getPartitionStart(), splitKeyType) [L543-578]
              -- user configures a non-numeric, non-ISO string for a TIME/TIMESTAMP/TIMESTAMP_TZ bound
              -- SqlType.DATE check at L553 is false -> skip LocalDate.parse -> numeric parse [L560]
              -- new BigDecimal(bound) throws NumberFormatException
              -- catch block [L567]: SqlType.DATE.equals(sqlType) is false -> `throw e;` [L576]
              -- raw java.lang.NumberFormatException propagates out of split enumeration, unwrapped
```
This remains a misconfiguration-only path — the auto-discovery and valid-input paths are unaffected — and it still fails loudly rather than losing data, so there's no data-loss/regression risk. It's just an inconsistency in the very hardening this PR introduced: the fractional-value check three lines above already treats all four temporal types uniformly, but the malformed-string wrapping still only special-cases `DATE`.

**Key findings**
1. All items from my 2026-08-20T02:30:25Z review that I flagged as fixed remain fixed on this head: the `DATE` malformed-bound wrapping, the fractional-boundary rejection (both `DATE` and `TIMESTAMP`), the `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` conversion test coverage, and the `docs/en`/`docs/zh` `Jdbc.md` ISO-8601 date-string documentation are all still present and correct — confirmed by re-reading the file and `git diff f83cf6968f eb5195ae1a` showing no logic touched them.
2. The one item I flagged as the sole remaining blocker last round — Issue 1 below — is still open. Today's commit was a Spotless pass triggered by a CI formatting failure on the previous head, not a follow-up to my review comment; that's a reasonable thing to land on its own, but it means the code-side blocker is still there.

## 1.2 Compatibility Impact
**Fully compatible**, unchanged from my last round: no config option renamed/removed/re-defaulted, the legacy raw-epoch-day `DATE` fallback still works, `JdbcSourceSplit.splitStart`/`splitEnd` typing (`Serializable`) is unchanged, no checkpoint/state format touched. No entry needed in `incompatible-changes.md`.

No change from last round: no new I/O, threads, or locks; the validation/conversion helpers are cheap and run once per table during split enumeration, not on the read hot path.

**Issue 1 (Medium) — carryover, not newly introduced by today's commit:** `parsePartitionBound`'s malformed-input handling remains inconsistent across the temporal types it explicitly recognizes via `temporalSplitKey`.
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/source/FixedChunkSplitter.java:567-577`
- Problem: the `catch (NumberFormatException e)` block only wraps the exception into a `JdbcConnectorException` when `sqlType == DATE` (L568-575); for `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` it re-throws the raw `NumberFormatException` at L576. The three lines directly above (L561-565) already treat all four types identically for the fractional-value check, so this is a real, narrow inconsistency in the same commit that introduced the fractional check.
- Potential risk: Low — only affects a hand-configured `partition_lower_bound`/`partition_upper_bound` on a `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` split key that's neither a whole number nor blank; fails the job at split-enumeration time either way (no data loss, no silent zero-row splits), just a less actionable error message for three of the four temporal types than for `DATE`.
- Best improvement: extend the `if (SqlType.DATE.equals(sqlType))` check at L568 to `temporalSplitKey`, with a message that doesn't assume `yyyy-MM-dd` for the non-`DATE` case (e.g. "Invalid temporal partition bound '<bound>'. Expected a whole epoch-millis number." for `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`, keeping the more specific `yyyy-MM-dd` message for `DATE`). One branch, one mirrored test case (`testRejectInvalidDatePartitionBound` already shows the pattern to follow for `DATE`).
- Severity: Medium
- Raised by another reviewer: No (this is my own finding from 2026-08-20, restated because it's still unresolved on the current head — not something new that another reviewer caught in the interim)

No sensitive data logged or leaked anywhere in this diff.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Unchanged from last round — the four touched private methods (`parsePartitionBound`, `findSplitColumnRange`, `convertSplitBoundary`, `convertToBigDecimal`) all carry explanatory comments covering the epoch-day/epoch-millis unit convention and the backward-compatibility rationale, closing the documentation gap I raised on 2026-08-20T01:58:24Z. Today's commit only reflows one of those comments and reformats test call-site wrapping to satisfy Spotless — no standards regressions introduced.

Test coverage is unchanged in substance from last round (only argument-wrapping was reformatted) and still matches the scope of the fix: `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` conversion branches, the `Timestamp`-for-`DATE` min/max branch, both `parsePartitionBound` fallback branches, the `DATE` malformed-bound error, and fractional-bound rejection for both `DATE` and `TIMESTAMP` are all exercised. The only residual gap is still the one covered by Issue 1 (no test for a malformed, non-numeric `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` bound) — expected, since the source doesn't yet produce a distinct behavior to assert there.

**Test-stability analysis (mandatory, this PR touches UT code):** re-verified against the current head — all tests invoke private methods via reflection with fixed literal inputs and assert exact equality or exception type/message. No `Thread.sleep`, no real DB/network/container, no shared mutable/static state, no execution-order dependency, no timezone dependency in the assertions. **Stability rating: Stable.** Unchanged from last round; today's commit only touched formatting of the call sites, not the assertions or fixtures.

## 2.3 Documentation Updates
Unchanged from last round: `docs/en/connectors/source/Jdbc.md` and `docs/zh/connectors/source/Jdbc.md` both document ISO-8601 date-string support for `DATE`-typed partition bounds, in matching English/Chinese wording. No further action needed for this PR.

# 3. Architectural Soundness

**Precise fix.** Unchanged assessment — the hardening stays exactly at the parse/bind boundary layer; the shared `JdbcNumericBetweenParametersProvider` range arithmetic is untouched, keeping the blast radius limited to the temporal-column case this PR targets.

## 3.2 Maintainability
Good, unchanged from last round — the private methods are short, single-purpose, and now properly commented.

## 3.3 Extensibility
Unchanged — the `SqlType`-dispatch pattern is reusable if another temporal-like type needs the same treatment later.

Confirmed compatible, unchanged: no checkpoint/state format changes, legacy raw-epoch-day `DATE` bounds keep working via the fallback branch. No upgrade action required.

# 4. Issue Summary

| 1 | `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` malformed (non-numeric) partition bound still throws a raw, unwrapped `NumberFormatException` instead of a `JdbcConnectorException`, inconsistent with the `DATE` handling and the fractional-bound check three lines above (both already treat all four temporal types uniformly) | `FixedChunkSplitter.java:567-577` | Medium |

All other issues from my two previous rounds (malformed `DATE` bound, fractional-boundary `ArithmeticException`, missing `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`/fallback-branch test coverage, missing docs, missing method comments) remain confirmed fixed on this head.

# 5. Merge Recommendation

   - Issue 1: extend the existing `DATE`-only malformed-bound wrapping (`FixedChunkSplitter.java:568-575`) to also cover `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` via the already-computed `temporalSplitKey` flag, for consistency with the fractional-bound check three lines above. This is a small, mechanical change — one more `if`/message branch plus one test case mirroring `testRejectInvalidDatePartitionBound`.

   - None new this round.

**Overall assessment:** unchanged from my last round — the core design is correct, precise, and backward compatible; the only remaining gap is a narrow, low-risk consistency issue in error-message wrapping for three of the four temporal types this PR hardens. Today's commit was a Spotless formatting fix (confirmed via direct diff against the head I last reviewed), not a functional change, so Issue 1 is still open and is the only thing standing between this PR and merge. This should be a very quick turnaround — one more `if` branch and a matching test case.

**CI status:** `Build` is currently `pending` on the current head (`eb5195ae1a`) — this is a fresh commit, CI is still queued/running. No prior CI failure was code-related; the previous `Build` failure was the Spotless formatting violation that today's commit fixes.

Thanks again for the quick turnaround on formatting — just the one small consistency fix left and I think this is ready.

JacobZheng0927:
Addressed the remaining review item in commit 31b49e6b7.

Malformed non-numeric bounds for DATE, TIME, TIMESTAMP, and TIMESTAMP_TZ are now wrapped in JdbcConnectorException with type-appropriate messages, and a TIMESTAMP regression test was added.

The new Build check is pending; please take another look once CI completes.

DanielLeens (commented):
Thanks for turning this around so quickly. I re-pulled the current head (`31b49e6b7`) and re-read the whole file from scratch rather than only diffing the new commit, per my usual re-review process. Good news: the one remaining item from my last round is cleanly fixed, and I don't see anything new to raise.

- User pain: when a JDBC source uses `split.size`/`partition_num` parallel splitting and the split key is a `DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` column, `FixedChunkSplitter` used to bind split boundaries with `PreparedStatement.setBigDecimal(...)`. Most drivers either reject that against a temporal column or silently mis-compare it, so the generated splits return zero rows and the job silently drops data instead of failing loudly.
- Fix approach: the splitter carries the split key's `SqlType` end to end — partition bounds are parsed into a type-aware `BigDecimal` (epoch-day for `DATE`, epoch-millis for `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`), the existing `JdbcNumericBetweenParametersProvider` range math still runs purely on that `BigDecimal`, and the two output boundaries are converted back to `java.sql.Date`/`Time`/`Timestamp` right before being stored on `JdbcSourceSplit`.
- One-sentence summary: DATE/TIME/TIMESTAMP partition splitting now works correctly, and every one of those four types now fails loudly with a clear, type-specific message on a malformed bound.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Exact change since my last review (`eb5195ae1a` → `31b49e6b7`)**, `FixedChunkSplitter.java:567-582`:

```java
// before (eb5195ae1a)
} catch (NumberFormatException e) {
    if (SqlType.DATE.equals(sqlType)) {
        throw new JdbcConnectorException(
                CommonErrorCodeDeprecated.ILLEGAL_ARGUMENT,
                "Invalid DATE partition bound '" + bound
                        + "'. Expected yyyy-MM-dd or a whole epoch-day number.",
                e);
    }
    throw e;   // TIME/TIMESTAMP/TIMESTAMP_TZ still fell through here — my last-round blocker
}
```
```java
// after (31b49e6b7)
} catch (NumberFormatException e) {
    if (temporalSplitKey) {
        String expectedFormat =
                SqlType.DATE.equals(sqlType)
                        ? "yyyy-MM-dd or a whole epoch-day number."
                        : "a whole epoch-millisecond number.";
        throw new JdbcConnectorException(
                CommonErrorCodeDeprecated.ILLEGAL_ARGUMENT,
                "Invalid " + sqlType + " partition bound '" + bound + "'. Expected " + expectedFormat,
                e);
    }
    throw e;
}
```

**Key findings:**
1. This is exactly the fix I asked for: the malformed-bound wrapping now keys off the same `temporalSplitKey` flag (`DATE`/`TIME`/`TIMESTAMP`/`TIMESTAMP_TZ`) that the fractional-value check three lines above already used, closing the inconsistency between the two checks.
2. The message is correctly type-specific — `DATE` keeps the `yyyy-MM-dd or a whole epoch-day number` wording (matching its epoch-day unit), while the other three temporal types get `a whole epoch-millisecond number` (matching their epoch-millis unit from `convertSplitBoundary`/`convertToBigDecimal`). No copy-paste mismatch between the message and the actual unit convention.
3. New test `testRejectInvalidNonDateTemporalPartitionBound` (`FixedChunkSplitterTest.java`) exercises exactly this branch with `LocalTimeType.LOCAL_DATE_TIME_TYPE` and a non-numeric bound (`"2024-01-03 00:00:00"`), asserting the exception is `JdbcConnectorException` and the message contains both `TIMESTAMP` and `epoch-millisecond`. This closes the last coverage gap.
4. I re-traced the runtime path once more end to end (`getJdbcSourceSplits` → `parsePartitionBound` → `createNumberColumnSplits` → `convertSplitBoundary` → JDBC bind) and confirm nothing else in the shared `JdbcNumericBetweenParametersProvider` range arithmetic was touched — the change stays exactly at the parse/error-wrapping boundary, same scope as every prior round.

I don't see any new gap introduced by this small, mechanical change.

## 1.2 Compatibility Impact

**Fully compatible**, unchanged from my last two rounds: no config option renamed/removed/re-defaulted, the legacy raw-epoch-day `DATE` fallback still works, `JdbcSourceSplit.splitStart`/`splitEnd` typing (`Serializable`) is unchanged, no checkpoint/state format touched.

No change — this is purely an exception-message branch, evaluated only on the misconfiguration path during split enumeration, not on any hot path.

No open items. All four temporal types now get a wrapped, actionable `JdbcConnectorException` on a malformed bound instead of a raw `NumberFormatException`.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Clean — the four touched private methods remain properly commented from the previous round, and this change doesn't introduce anything new needing a comment beyond what's already there.

`testRejectInvalidNonDateTemporalPartitionBound` mirrors the existing `testRejectInvalidDatePartitionBound` pattern exactly (reflection-invoked private method, `InvocationTargetException` unwrap, exact message-content assertions). No `Thread.sleep`, no real DB/network/container, no shared mutable/static state, no timezone dependency. **Stability rating: Stable**, consistent with every other test in this class.

## 2.3 Documentation Updates

Unchanged from last round — already in good shape.

# 3. Architectural Soundness

Unchanged from my last two rounds: precise fix, good maintainability, reusable `SqlType`-dispatch pattern, fully backward compatible historically. Nothing to add.

# 4. Issue Summary

No open issues. All four items from my previous rounds (malformed `DATE` bound, fractional-boundary `ArithmeticException`, missing `TIME`/`TIMESTAMP`/`TIMESTAMP_TZ` test coverage, missing docs, and the malformed-bound wrapping inconsistency for non-`DATE` temporal types) are now confirmed fixed.

# 5. Merge Recommendation

I have no further source-side blockers or recommended fixes. CI on this exact commit (`31b49e6b7`) is currently queued on the fork (the previous head `eb5195ae1a` built green), so there's nothing to diagnose yet — just waiting on that run to complete. This PR is also `behind_by 48` commits from `dev`, which is a bit stale but the diff is narrowly scoped to `FixedChunkSplitter`/its test, so I wouldn't expect a rebase to change anything material here.

Thanks for the quick, precise turnaround across all these rounds — nice work closing out every item cleanly.

DanielLeens:
Updating the CI status from my last review, where I noted the `Build` check on this exact commit (`31b49e6b7`) was "currently queued on the fork... nothing to diagnose yet." That run (`32680436515`) has since completed with `failure`, so there's now something concrete to report.

I pulled the job logs rather than just reading the check name: the failures are five Doris E2E test classes (`DorisTimerFlushIT`, `DorisCDCSinkIT`, `DorisErrorIT`, `DorisCatalogIT`, `DorisMultiReadIT`), and every one of them fails the same way -- `org.testcontainers.containers.ContainerLaunchException: Timed out waiting for container port to open` while starting the `apache/doris:doris-all-in-one-2.1.0` testcontainer (one even failed one step earlier, on the image pull itself, with an `InterruptedException`). That's a container/runner-resource bootstrap issue in the Doris connector-it module, not a test assertion failure, and it has nothing to do with this PR's diff (`FixedChunkSplitter` / JDBC partition-column splitting) -- the Doris connector code path isn't touched here.

From my source-review side there's still no open code blocker on this head. Given the failure signature looks like transient container/runner contention rather than a deterministic break, the lowest-cost next step is a plain rerun of the failed jobs; if it reproduces identically on a clean rerun, that would point to something more systemic on the runner side worth a separate look, but I wouldn't hold this PR on it either way since the failing module is outside this diff's blast radius.

PR #11029, opened 2026-06-09
[Feature][Connector-V2] Introduce PG base CDC abstraction for Postgres and OpenGauss

davidzollo:
## What's changed
- add a new `connector-cdc-pg-base` module to centralize the shared PostgreSQL-compatible Debezium property assembly and row deserialization logic
- migrate the Postgres CDC source config and incremental source to the shared PG base implementation without changing the existing connector identifier or slot/plugin options
- wire OpenGauss CDC through its own source entry point so it can reuse the PG base behavior while keeping the current OpenGauss plugin identity
- give OpenGauss its own `startup.mode` option so it no longer inherits PostgreSQL-only startup modes
- add regression tests for Postgres table identifier normalization (`schema.table` and `db.schema.table`), the PG base source path, the OpenGauss entry path, and the replica-identity wiring described below

## Why
- Postgres and OpenGauss currently duplicate the same PG-compatible CDC assembly logic, which makes follow-up maintenance and future connector additions much harder
- this change keeps the refactor in a phase-1 shape: common config/deserialization logic moves to the PG base module first, while dialect-specific runtime behavior stays in the existing connector modules to reduce migration risk

## Behavior change: `require-replica-identity-full` enforcement is restored

This PR is behavior-preserving on the main path with one deliberate exception, called out here because a reader would otherwise not know to check for it.

`PostgresIncrementalSource` used to read `require-replica-identity-full` into a `final` field assigned *after* `super(options, catalogTables)` returned. `createDataSourceDialect()` is invoked from *inside* that constructor, so the dialect was always built with the Java default `false`, and the check has been silently a no-op on `dev` since #10334. This PR reads the option inside `createDataSourceDialect(config)`, which restores the documented default of `true`.

Impact: a `Postgres-CDC` or `Opengauss-CDC` job reading tables without `REPLICA IDENTITY FULL`, which does not set `require-replica-identity-full = false`, will now fail fast at submit, enumerator restore, and WAL fetch-task start — where a current `dev` build silently skips the check. This matches every released version and the existing documentation (`docs/en/connectors/source/PostgreSQL-CDC.md`), so it is a regression fix rather than a new restriction, but anyone running a recent `dev` snapshot with non-`FULL` tables should set `require-replica-identity-full = false` before upgrading onto this change.

[...]

PR #11399, opened 2026-07-10
[Fix][Connector-V2] Filter Postgres CDC database discovery

### LINKED ISSUE #8184: [Bug] [Connector-V2][Postgres-CDC] Scanning unwanted database and tables which are not mentioned in Source
### Search before asking

### What happened

When we provide a hostname for Postgres-CDC connection and try to run the job we are seeing unwanted WARN logs which says TableDiscoveryUtils class is trying to scan for all the databases under the specified host. Even though we specify the database name in the source config we are still getting these unwanted warnings. Is there any way to avoid other databases and tables which are not part of the connection ?

### SeaTunnel Version

seatunnel-2.3.8

### SeaTunnel Config

```conf
env {
  execution.parallelism = 1
  job.mode = "STREAMING"
  checkpoint.interval = 5000
  read_limit.bytes_per_second=7000000
  read_limit.rows_per_second=400
}

source {
  
  Postgres-CDC {
    result_table_name = "employees"
    username = "postgres"
    password = "password"
    database-names = ["postgres"]
    schema-names = ["public"]
	slot.name = "employeeslot"
    table-names = ["postgres.public.employees"]
    base-url = "jdbc:postgresql://localhost:5432/postgres"
  }
  
}

transform {
}

sink {
	
	Console {
      source_table_name = "employees"
  }
 
}
```

### Running Command

```shell
./bin/seatunnel.sh --config config/postgres-cdc.template --async -n postgrescdcsync
```

### Error Exception

```log
[915470173375299585] 2024-11-30 23:11:41,544 INFO  [.c.s.c.p.u.TableDiscoveryUtils] [hz.main.seaTunnel.task.thread-128] -                                                                              list of available databases is: [postgres, database2, database3]
[915470173375299585] 2024-11-30 23:11:41,544 INFO  [.c.s.c.p.u.TableDiscoveryUtils] [hz.main.seaTunnel.task.thread-128] - Read list of available tables in each database
[915470173375299585] 2024-11-30 23:11:41,606 WARN  [.c.s.c.p.u.TableDiscoveryUtils] [hz.main.seaTunnel.task.thread-128] -                                                                              skipping database 'database2' due to error reading tables: ERROR: cross-database references are not implemented: "database2.information_schema.tables"
  Position: 15
[915470173375299585] 2024-11-30 23:11:41,657 WARN  [.c.s.c.p.u.TableDiscoveryUtils] [hz.main.seaTunnel.task.thread-128] -                                                                              skipping database 'database3' due to error reading tables: ERROR: cross-database references are not implemented: "database3.information_schema.tables"
  Position: 15
```

_No response_

### Java or Scala Version

Jdk 1.8

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
## Purpose

Fixes #8184.

Postgres CDC now applies Debezium database filters before querying each database metadata schema. This prevents configured sources from probing unrelated PostgreSQL databases and emitting cross-database warnings.

## Tests

- Added a regression test that verifies only configured databases are queried.
- `./mvnw -o -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres -Dtest=TableDiscoveryUtilsTest test`
- `./mvnw -o -q -DskipTests verify`


DanielLeens (commented):
Thanks for the contribution! I reviewed the latest head `a8dd9ea6a3` against the full current diff and traced the Postgres table-discovery path from the connector entry down to the per-database metadata query.

# What this PR fixes
- User pain: when users scope Postgres CDC to a subset of databases, the current discovery path still queries every database returned by `pg_database`, including databases that are already excluded by the configured filter.
- Fix approach: apply `tableFilters.databaseFilter()` before issuing the per-database `INFORMATION_SCHEMA.TABLES` query, and add a regression test that asserts only the allowed database is queried.
- In one sentence: this turns the database filter into an actual pre-query gate instead of only filtering after metadata has already been fetched.

# Runtime path checked
```text
Postgres CDC source creation
  -> PostgresDialect.discoverDataCollections(...)
      -> TableDiscoveryUtils.listTables(...)
          -> query pg_database
          -> for each dbName
              -> databaseFilter().test(dbName)
              -> if included, query "db".INFORMATION_SCHEMA.TABLES
              -> dataCollectionFilter().isIncluded(tableId)
```

# Current findings
- `TableDiscoveryUtils.java:62-66` now skips filtered-out databases before the metadata query is sent.
- The new test `TableDiscoveryUtilsTest.shouldOnlyQueryDatabasesAllowedByConfiguredFilter()` covers the regression directly by asserting the unwanted database is never queried.
- I do not see a reopened source-level blocker on the latest head.
- Compatibility-wise this is safe: it tightens behavior to match the already-configured filter contract and does not change user-facing option names or defaults.
- Test stability looks good here: the regression test is fully mocked, deterministic, and does not rely on timing or external services.

# Merge conclusion
Source-level: no blocker from Daniel's side on the latest head.

The current remaining gate is CI: `Build` is still in progress here: https://github.com/apache/seatunnel/runs/86385021571

Once that finishes green, this looks good from the source-review side.

DanielLeens (approved):
Rechecked the latest gate after my previous review. My last Daniel-side conclusion on this PR had no remaining source blocker and was only waiting for `Build`.

That has now changed: `Build` is successful on the same head (`a8dd9ea6a3`): https://github.com/apache/seatunnel/runs/86600911950

I still do not see a reopened code-side blocker. The current compare metadata shows the branch is `diverged` from `dev` with `behind_by=1`, which I am recording as a mergeability fact, but not turning into a new contributor action while the source review and current Build are both green.

Approving this from Daniel's technical review state.

DanielLeens:

At this point I do not see a new code-side blocker from Daniel on this revision. For queue bookkeeping, this branch is still `diverged` against the latest `dev` (`ahead_by=1`, `behind_by=9`), but because the required `Build` signal is green now, I am not asking for a sync-only retry from Daniel's side here.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `a8dd9ea6a3bc` from scratch.

# Findings
**Issue 1: The fix is only effective if Debezium's `database.include.list`/`database.exclude.list` is actually populated by the SeaTunnel Postgres CDC config factory; otherwise `databaseFilter()` accepts everything and #8184 is not fixed**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtils.java:63`
- Why it matters: - Root cause: the new guard delegates to `RelationalTableFilters.databaseFilter()`, which in Debezium is built from `DATABASE_INCLUDE_LIST` / `DATABASE_EXCLUDE_LIST`. When neither property is set, the predicate degenerates to accept-all, and the added `continue` branch is never taken.
- The SeaTunnel Postgres CDC config path (now centralized in `connector-cdc-pg-base` per PR #11029) historically sets `database.dbname` for the connected database; whether it also sets `database.include.list` from `database-names` determines if this patch actually stops the cross-database probing and warnings described in the issue.
- Impact: if the include list is not set, existing pipelines will see zero behavior change — the connector will keep issuing `SELECT * FROM "otherdb".INFORMATION_SCHEMA.TABLES` against unrelated databases and logging warnings, i.e. the reported bug survives the release note claiming it is fixed.
- Fix: trace `PostgresSourceConfigFactory` (pg-base) and confirm it populates `database.include.list` from the configured `database-names`; if it does not, add that property assembly in the same PR, and cover the end-to-end path with a test that builds `RelationalTableFilters` from the real `PostgresConnectorConfig` instead of a mock.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtils.java:65  continue;`
- Suggested fix: `RelationalTableFilters.databaseFilter()` only filters when Debezium's `database.include.list`/`database.exclude.list` is set. Can you confirm `PostgresSourceConfigFactory` (pg-base) actually populates `database.include.list` from the user's `database-names`? If it doesn't, this predicate is accept-all in production and the cross-database probing from #8184 still happens. If it's missing, please add that property assembly here as well and add a test that constructs `RelationalTableFilters` from the real `PostgresConnectorConfig` rather than a Mockito mock.
- Severity: Medium

**Issue 2: Regression test mocks RelationalTableFilters, so it only proves the utility honors an arbitrary predicate — not that the SeaTunnel-configured filter actually excludes unrelated databases**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:44`
- Why it matters: - Root cause: `TableDiscoveryUtilsTest` stubs `tableFilters.databaseFilter()` with `"selected"::equals`, bypassing the entire Debezium filter-construction pipeline (`database.include.list`, regex compilation, case-insensitive matching) that a real pipeline goes through.
- Impact: the test would still pass even if SeaTunnel never sets `database.include.list` in the Debezium properties, in which case the production predicate is accept-all and the bug in #8184 is not fixed. The test therefore cannot regress on the actual failure mode reported by users.
- Fix: add a second test that builds `RelationalTableFilters` from a real `io.debezium.config.Configuration` containing `database.include.list=selected` (ideally the exact `Properties` emitted by `PostgresSourceConfigFactory` for a `database-names=[selected]` config) and asserts the `unwanted` database is skipped end-to-end.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:43  RelationalTableFilters tableFilters = Mockito.mock(RelationalTableFilters.class);`
- Suggested fix: Mocking `databaseFilter()` with `"selected"::equals` only verifies the utility respects some predicate; it doesn't verify that the filter SeaTunnel actually produces (via Debezium `database.include.list`) excludes unrelated databases — which is the actual bug in #8184. Please add a variant that builds a real `RelationalTableFilters` from a `Configuration` containing `database.include.list` (ideally derived from `PostgresSourceConfigFactory` output) so the regression test covers the production filter construction path.
- Severity: Medium

**Issue 3: MockJdbcConnection dispatches on exact hardcoded SQL strings, making the test brittle to harmless query refactors**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:74`
- Why it matters: - Root cause: the mock branches on `query.equals("select datname from pg_database")` and `query.contains("\"selected\"")`, and throws `AssertionError` for anything else, coupling the test to the exact SQL text and quoting produced by `listTables`.
- Impact: any legitimate change to the discovery SQL (adding a column list, whitespace, `WHERE datallowconn` filtering, or a different quote character) will break this test with an opaque `Unexpected database query` failure even though filtering behavior is unchanged, discouraging future improvements to the discovery query.
- Fix: dispatch on stable substrings (e.g. `contains("pg_database")` / `contains("INFORMATION_SCHEMA.TABLES")`) and keep the strict assertion focused on the essential property — that no query targets the `unwanted` database — rather than exact full-string equality of both queries.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:83  throw new AssertionError("Unexpected database query: " + query);`
- Suggested fix: Branching on the exact SQL string couples this test to the precise query text in `TableDiscoveryUtils`; any harmless SQL refactor (column list, whitespace, `datallowconn` filter) will fail with `Unexpected database query`. Consider matching on stable substrings (`pg_database`, `INFORMATION_SCHEMA.TABLES`) and asserting the key invariant — no query mentions the `unwanted` database — instead of exact full-string equality.
- Severity: Low

**Issue 4: Config/option changes without doc updates**
- Severity: Low

# Review conclusion

- Issue 1: The fix is only effective if Debezium's `database.include.list`/`database.exclude.list` is actually populated by the SeaTunnel Postgres CDC config factory; otherwise `databaseFilter()` accepts everything and #8184 is not fixed
- Issue 2: Regression test mocks RelationalTableFilters, so it only proves the utility honors an arbitrary predicate — not that the SeaTunnel-configured filter actually excludes unrelated databases
- Issue 3: MockJdbcConnection dispatches on exact hardcoded SQL strings, making the test brittle to harmless query refactors
- Issue 4: Config/option changes without doc updates

DanielLeens:
Thanks @SEZ9 for tracing this path in detail. I rechecked the production config assembly and I agree Issues 1 and 2 are real on the current head.

`PostgresSourceConfigFactory.create()` still sets `database.dbname`, but it does not populate Debezium `database.include.list`. That means `RelationalTableFilters.databaseFilter()` can indeed degenerate to accept-all here, so the new `continue` branch in `TableDiscoveryUtils` by itself is not enough to prove `#8184` is fixed end-to-end.

From Daniel's side I would treat the include-list wiring plus a real-filter regression test as required before merge.

DanielLeens (changes requested):
Thanks for the contribution, and my apologies up front: I reviewed this PR twice in July (COMMENTED on `a8dd9ea6a3` on Jul 10, then APPROVED the same head on Jul 12), and my approval missed a gap that @SEZ9 raised on Jul 14 as a Medium-severity open question ("is `database.include.list` actually populated?"). I hadn't re-reviewed since. I did a complete re-review from scratch of the current head `d171887c314` (which, per `git diff a8dd9ea6a3bc..d171887c314` on the two files this PR touches, is byte-identical to what @SEZ9 reviewed — the head only merged in unrelated upstream `dev` commits, no new content), and this time I traced the actual property assembly all the way through `PostgresSourceConfigFactory`. I can now confirm @SEZ9's concern is not a hypothetical — it's a verified, reproducible defect that means this PR does not fix the issue it claims to fix for the reporter's own configuration. Sorry for the earlier approval; escalating properly now.

- User pain (per the linked issue #8184): a Postgres CDC source configured with `database-names = ["postgres"]` still triggers `TableDiscoveryUtils` to scan every database visible to the connection (`postgres`, `database2`, `database3`, ...), logging `WARN ... skipping database 'database2' due to error reading tables: ERROR: cross-database references are not implemented` for every database outside the one actually configured.
- Fix approach as implemented: add `if (!tableFilters.databaseFilter().test(dbName)) { continue; }` before the per-database `INFORMATION_SCHEMA.TABLES` query in `TableDiscoveryUtils.listTables`, so a database excluded by Debezium's `RelationalTableFilters` is skipped before the query (and therefore before the warning) is ever issued.
- One-sentence summary: the guard itself is correctly placed and would work if `tableFilters.databaseFilter()` were ever configured to exclude anything, but nothing in the SeaTunnel Postgres CDC config path ever populates Debezium's `database.include.list`/`database.exclude.list`, so `databaseFilter()` is always accept-all in production — for the exact reproduction config in the linked issue, this PR changes nothing observable.

# 1. Code Change Review

## 1.1 Core Logic Analysis

### Exact change

`seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtils.java:63-66`:

```java
if (!tableFilters.databaseFilter().test(dbName)) {
    LOG.info("\t database '{}' is filtered out of capturing", dbName);
    continue;
}
```

This is the entire production-code diff (4 lines added, nothing removed).

### Runtime path traced end to end

```text
PostgresDialect.discoverDataCollections(sourceConfig)                     [PostgresDialect.java:112-121]
  -> postgresSourceConfig.getTableFilters()                               [PostgresSourceConfig.java:91-92]
      -> getDbzConnectorConfig().getTableFilters()                        (io.debezium.relational.RelationalDatabaseConnectorConfig)
          -> built from the debezium Configuration, which is constructed from the Properties
             assembled in PostgresSourceConfigFactory.create()            [PostgresSourceConfigFactory.java:60-129]
  -> TableDiscoveryUtils.listTables(jdbcConnection, tableFilters)         [PostgresDialect.java:117-118]
      -> query pg_database
      -> for each dbName: tableFilters.databaseFilter().test(dbName)     [TableDiscoveryUtils.java:63]  <- new guard
```

I read `PostgresSourceConfigFactory.create()` (`PostgresSourceConfigFactory.java:60-129`) in full. It sets `database.server.name`, `database.hostname`, `database.user`, `database.password`, `database.port`, `database.dbname` (line 72, from `databaseList.get(0)` — note only the *first* configured database is even used here), `plugin.name`, `slot.name`, database-history properties, `schema.include.list` (if `schemaList` is set), `table.include.list` (if `tableList` is set), any raw passthrough `dbzProperties`, and `snapshot.mode`. At no point does it set `database.include.list` or `database.exclude.list` from `databaseList` (the field populated from the user's `database-names` config option, `JdbcSourceConfigFactory.java:272`). I grepped the entire `connector-cdc-postgres` and `connector-cdc-base` modules for `database.include.list`/`database.exclude.list`/`DATABASE_INCLUDE_LIST`/`DATABASE_EXCLUDE_LIST` and found zero matches anywhere in the Postgres CDC source tree.

Debezium's `RelationalDatabaseConnectorConfig.getTableFilters()` builds `databaseFilter()` from exactly those two config keys via `Selectors.databaseSelector()`, defaulting to include-all when neither is set. Since SeaTunnel never populates them for Postgres, `tableFilters.databaseFilter().test(dbName)` returns `true` for every database, for every Postgres CDC job configured through the normal `database-names` option — which is precisely the option the linked issue's reproduction config uses (`database-names = ["postgres"]`).

### Key findings

1. The new guard at `TableDiscoveryUtils.java:63` is correctly shaped and would work if the filter were populated.
2. The filter is never populated for the standard `database-names` config path, so the guard is a no-op for the reported bug's own reproduction steps. Running the exact config from issue #8184 against this PR would still print the same `WARN ... skipping database 'database2' due to error reading tables: cross-database references are not implemented` lines — nothing observable changes for the reporter.
3. The regression test added in this PR (`TableDiscoveryUtilsTest.shouldOnlyQueryDatabasesAllowedByConfiguredFilter`) mocks `RelationalTableFilters` directly and stubs `databaseFilter()` to `"selected"::equals` — it never goes through `PostgresSourceConfigFactory` or real Debezium `Configuration`/`Selectors` construction, so it cannot catch the gap in finding 2. The test proves `TableDiscoveryUtils.listTables` correctly *honors* whatever predicate it's handed; it does not prove SeaTunnel ever hands it a meaningful predicate.

### Logic correctness deep dive

Where it works: if a caller manually sets `database.include.list`/`database.exclude.list` via the raw `dbzProperties` passthrough (an advanced, undocumented-for-this-purpose escape hatch), the new guard does filter correctly, and cross-database queries are skipped before they're issued.

Where it still fails: for the documented, standard `database-names` config option — the one in the linked issue and the one most users would reach for — `databaseFilter()` remains accept-all, and the underlying bug is unfixed.

## 1.2 Compatibility Impact

Judgment: Safe, but ineffective. This PR does not remove or rename any config option, does not change any default, and does not touch the wire protocol or checkpoint/restore state. There is no compatibility risk from merging it — the risk is purely that it ships as a claimed fix for #8184 without actually fixing the case the issue reports.

- CPU/memory/network: the guard, when it does trigger (the raw-`dbzProperties` case), correctly avoids an unnecessary per-database query — a small, real improvement in that narrow path.
- No new resource lifecycle, retry, or checkpoint interaction introduced.

**Issue 1: The fix has no observable effect for the standard `database-names` config path — `database.include.list` is never populated, so `databaseFilter()` remains accept-all and issue #8184's own reproduction steps still reproduce**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/config/PostgresSourceConfigFactory.java:60-129` (missing `database.include.list`/`database.exclude.list` assembly); guard site `TableDiscoveryUtils.java:63`
- Problem description: this is @SEZ9's Jul 14 Issue 1, which I did not verify when I approved this PR on Jul 12 (I had reviewed an earlier head before that question was raised, and didn't return to check it). I now confirm via direct code trace that `PostgresSourceConfigFactory.create()` never sets `database.include.list` or `database.exclude.list` from `databaseList` (the field backing the `database-names` option) — only `database.dbname` is set, and only from `databaseList.get(0)`. Debezium's table-filter construction defaults to include-all when those two keys are absent.
- Potential risk: users following the exact config pattern in issue #8184 (`database-names = ["postgres"]`) will see identical behavior before and after this PR — the same cross-database `INFORMATION_SCHEMA.TABLES` queries against every database visible to the connection, and the same WARN-level "cross-database references are not implemented" log spam the issue was filed about. The PR's own regression test cannot catch this because it bypasses `PostgresSourceConfigFactory` entirely (see Issue 2).
- Best improvement: populate `database.include.list` in `PostgresSourceConfigFactory.create()` from `databaseList` (e.g. `props.setProperty("database.include.list", String.join(",", databaseList))`), consistent with how `table.include.list` is already assembled from `tableList` a few lines below. Then add an integration-level test (or extend the existing one) that builds `PostgresSourceConfig`/`PostgresConnectorConfig` from a `database-names = [...]` config and asserts the resulting `RelationalTableFilters.databaseFilter()` actually excludes an unconfigured database name — this is the only way to prove the fix reaches the field the reporter's config uses.
- Severity: Critical / Blocker
- Raised by another reviewer: Yes — @SEZ9 raised this as Medium on Jul 14 ("The fix is only effective if Debezium's `database.include.list`/`database.exclude.list` is actually populated ... otherwise `databaseFilter()` accepts everything and #8184 is not fixed"). I'm escalating to Critical/Blocker because I have now confirmed via direct source trace (not just a question) that this is the case — it is not a possible gap, it is a definite one, verified against the exact reproduction config in the linked issue. I should have verified this myself before approving on Jul 12; apologies for the earlier miss.

**Issue 2 (carryover, unresolved from @SEZ9's Jul 14 Issue 2): the regression test mocks `RelationalTableFilters` directly, so it cannot detect Issue 1**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:43-44`
- Problem description: unchanged from @SEZ9's review — `RelationalTableFilters tableFilters = Mockito.mock(RelationalTableFilters.class); when(tableFilters.databaseFilter()).thenReturn("selected"::equals);` bypasses the entire Debezium filter-construction pipeline (`database.include.list` parsing, regex compilation) that a real job goes through. This is exactly why CI stays green despite Issue 1: the test only proves the utility method honors whatever predicate it's handed, not that SeaTunnel ever constructs a meaningful one.
- Potential risk: false confidence — this test will keep passing even if `PostgresSourceConfigFactory` is fixed incorrectly, or never fixed at all, because it never exercises the real config-to-filter path.
- Best improvement: add a second test that builds a real `RelationalTableFilters` from a `io.debezium.config.Configuration` containing `database.include.list=selected` (ideally the exact `Properties` produced by `PostgresSourceConfigFactory.create()` for a `database-names=["selected"]` config), and asserts the `unwanted` database is skipped end to end. This is the test that would have caught Issue 1.
- Severity: Medium (downstream of Issue 1; would become moot once Issue 1's fix ships with matching coverage)
- Raised by another reviewer: Yes — @SEZ9, Jul 14, unresolved since (no new commits touch this file).

**Issue 3 (carryover, unresolved from @SEZ9's Jul 14 Issue 3): `MockJdbcConnection` dispatches on exact hardcoded SQL strings, making the test brittle to harmless query refactors**
- Location: `TableDiscoveryUtilsTest.java:76` (`query.equals("select datname from pg_database")`), `79` (`query.contains("\"selected\"")`), `85` (`throw new AssertionError("Unexpected database query: " + query)`)
- Problem description: unchanged — confirmed byte-identical to the head @SEZ9 reviewed. Any legitimate change to the discovery SQL (added column, whitespace, a `datallowconn` filter) breaks this test with an opaque `Unexpected database query` failure even though filtering behavior would be unaffected.
- Best improvement: dispatch on stable substrings (`contains("pg_database")` / `contains("INFORMATION_SCHEMA.TABLES")`) and keep the strict assertion focused on the essential invariant — no query targets the `unwanted` database.
- Severity: Low
- Raised by another reviewer: Yes — @SEZ9, Jul 14, unresolved since.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The production change is minimal, readable, and correctly placed within the existing loop structure. No style concerns.

The single new test is deterministic and fully mocked (no timing/external-service dependency, so it won't flake), but as covered in Issues 1-2, it verifies the wrong layer: it proves `TableDiscoveryUtils.listTables` respects an arbitrary predicate, not that SeaTunnel's Postgres CDC config path ever produces a predicate that excludes anything for a standard `database-names` configuration. Test stability itself is fine; test *validity* against the actual reported bug is not.

## 2.3 Documentation Updates

No new user-facing config option is introduced by this PR (it changes internal discovery behavior only), so no doc update is strictly required for this diff as currently scoped. If Issue 1 is fixed by wiring `database-names` into `database.include.list`, that's an internal implementation detail of an existing option and likely still doesn't need new docs — but worth a one-line mention in the Postgres CDC connector doc's discovery/filtering notes if one exists, since the current behavior (silently scanning every database) is a common enough point of confusion that it produced this GitHub issue.

# 3. Architectural Soundness

The guard-before-query pattern is the right shape and reuses Debezium's own filter abstraction rather than reinventing one — a sound direction. The gap is entirely upstream of this file: the filter that `TableDiscoveryUtils` consults is never populated by the connector's own config factory.

## 3.2 Maintainability

Fine as far as it goes; the real maintainability risk is that a reader of `TableDiscoveryUtils.java:63` alone would reasonably assume the `database-names` option already drives this filter (since that's the intuitive design), when in fact it does not. That mismatch between the option's apparent purpose (`database-names` = "connect and restrict to these databases") and its actual effect (only sets the JDBC connection's single `database.dbname`, does not restrict discovery) is the root cause of the linked issue and remains unaddressed by this diff.

## 3.3 Extensibility

Once `database.include.list` is correctly populated, the same pattern extends cleanly to `database.exclude.list` or multi-database `database-names` lists (currently only `databaseList.get(0)` is used for the connection itself, which is a separate, pre-existing scoping question for Postgres CDC not introduced by this PR).

No compatibility risk — this PR only adds a `continue` skip, it doesn't change any existing option's meaning, default, or wire format. Safe to merge from a compatibility standpoint independent of whether Issue 1 is fixed; the concern is effectiveness, not safety.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | `database.include.list`/`database.exclude.list` never populated from `database-names`, so `databaseFilter()` stays accept-all and issue #8184's reproduction config is unaffected by this PR | `PostgresSourceConfigFactory.java:60-129`; `TableDiscoveryUtils.java:63` | Critical/Blocker |
| 2 | Regression test mocks `RelationalTableFilters` directly, cannot detect Issue 1 | `TableDiscoveryUtilsTest.java:43-44` | Medium |
| 3 | Mock SQL dispatch on exact hardcoded strings, brittle to harmless query refactors | `TableDiscoveryUtilsTest.java:76-85` | Low |

# 5. Merge Recommendation

CI on the current head `d171887c314`: `Build` is green (https://github.com/apache/seatunnel/runs/95172895041). I'm not treating that as dispositive here, since the one new test in this PR is constructed in a way that cannot exercise the actual defect (Issue 2) — a green CI run does not mean the reported bug is fixed.

1. Blockers — must be fixed:
   1. Issue 1 — populate `database.include.list` (and/or `database.exclude.list`) in `PostgresSourceConfigFactory.create()` from the user's `database-names` config, so `databaseFilter()` is actually restrictive for the standard config path, not just for callers who manually inject raw Debezium properties.
   2. Issue 2 — once Issue 1 is fixed, add or extend a test that builds `RelationalTableFilters` from a real `Configuration`/`PostgresConnectorConfig` (not a Mockito mock) so the regression suite actually proves the fix reaches production behavior. Ideally reproduce the exact config shape from issue #8184 (`database-names = ["postgres"]` with other databases present) and assert no cross-database query is issued.

   - Issue 3 — match on stable SQL substrings in the mock connection instead of exact strings.

Overall assessment: I owe you and the project an apology here — I approved this PR on Jul 12 without verifying the exact gap @SEZ9 raised two days later, and nobody (including me) followed up on it since. Having now traced it end to end, I'm confident the current diff, while safe to merge, does not fix the linked issue for the configuration pattern the issue itself uses. I'd rather flag this clearly now than let it merge as a claimed fix that doesn't change observable behavior for the reporter. The fix itself should be small — a couple of lines in `PostgresSourceConfigFactory` plus a non-mocked test — and I'm glad to re-review quickly once pushed. As always, I review with read-level permission, so a write-capable maintainer performs final approval and merge.

SEZ9 (commented):
Thanks for working on this. I reviewed the latest head `d171887c3145`.

# What this PR fixes
- User pain: Postgres CDC sources scoped to specific databases still probed every database returned by `pg_database`, issuing cross-database metadata queries and emitting warnings for each unrelated database (#8184).
- Fix approach: `TableDiscoveryUtils.listTables` now consults Debezium's `RelationalTableFilters.databaseFilter()` and `continue`s past non-matching databases before issuing the per-database `INFORMATION_SCHEMA.TABLES` query, backed by a new mock-based `TableDiscoveryUtilsTest`.
- One-line summary: The guard is correctly placed but is a no-op on the real configuration path — `PostgresSourceConfigFactory` never populates `database.include.list`, so `databaseFilter()` degenerates to accept-all, #8184 remains unfixed for `database-names` users, the mocked test cannot catch that gap, and the change also activates previously-ignored `debezium.database.include.list`/`exclude.list` passthrough settings for users who already set them.

# Runtime chain I rechecked
```text
PostgresIncrementalSource -> PostgresSourceConfigFactory.create()  PostgresSourceConfigFactory.java (sets database.dbname, does NOT set database.include.list)
  -> PostgresSourceConfig.getDbzConnectorConfig() -> PostgresConnectorConfig.getTableFilters()  (databaseFilter() == accept-all when include/exclude lists absent)
  -> PostgresDialect.discoverDataCollections() -> TableDiscoveryUtils.listTables(jdbc, tableFilters)  TableDiscoveryUtils.java:52
    -> jdbc.query("select datname from pg_database", ...)  TableDiscoveryUtils.java:~55
    -> NEW: tableFilters.databaseFilter().test(dbName) guard  TableDiscoveryUtils.java:63-66 (continue if filtered)
    -> jdbc.query("SELECT * FROM \"<dbName>\".INFORMATION_SCHEMA.TABLES ...")  TableDiscoveryUtils.java:68+ (only reached for accepted databases; still reached for ALL databases under default config because the filter is accept-all)
```

# Findings
**Issue 1: The new database filter guard is a no-op under default SeaTunnel Postgres CDC configuration because `PostgresSourceConfigFactory` never populates Debezium's `database.include.list`, so #8184 is not actually fixed for users who set `database-names`.**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtils.java:63`
- Why it matters: - Root cause: the added guard delegates entirely to `tableFilters.databaseFilter()`. In Debezium, `RelationalTableFilters` builds that predicate from `database.include.list`/`database.exclude.list`. As confirmed in the review discussion on this PR, `PostgresSourceConfigFactory.create()` only sets `database.dbname` from the configured database list and never sets `database.include.list`. With neither list set, `databaseFilter()` degenerates to accept-all, and `test(dbName)` returns true for every database in `pg_database`.
- Impact: users who configured `database-names` to scope the source still see discovery probe every database and emit the exact cross-database warnings that issue #8184 reports. The guard only takes effect for users who manually pass `debezium.database.include.list`, which is not the connector's canonical option.
- Fix: either populate the Debezium property in `PostgresSourceConfigFactory.create()` (e.g. `props.setProperty("database.include.list", String.join(",", databaseList))` when `databaseList` is non-empty), or pass the source config's database list into `listTables(...)` and intersect it with `databaseFilter()`. Add coverage that exercises the real config path (see the related test finding).
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtils.java:65  continue;`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtils.java:64  LOG.info("\t database '{}' is filtered out of capturing", dbName);`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtils.java:66  }`
- Suggested fix: This guard is correct in isolation, but it only fires if Debezium's `database.include.list`/`database.exclude.list` are set — and, as established in the review discussion, `PostgresSourceConfigFactory.create()` only sets `database.dbname`, never `database.include.list`. With the connector's canonical `database-names` option, `tableFilters.databaseFilter()` is accept-all and this branch is dead code, so #8184 is not fixed for the default configuration. Please either populate `database.include.list` from `databaseList` in `PostgresSourceConfigFactory`, or intersect this check with the source config's configured database list.
- Severity: High

**Issue 2: The regression test stubs `databaseFilter()` with a hand-written predicate, so it only verifies the guard consults the mock — it does not prove that a filter derived from real SeaTunnel/Debezium configuration filters anything.**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:44`
- Why it matters: - Root cause: the test replaces `RelationalTableFilters` with a Mockito mock whose `databaseFilter()` returns the hand-written predicate `"selected"::equals`. The predicate under test is invented by the test itself rather than derived from connector configuration.
- Impact: the test proves only that `listTables` consults whatever predicate the filter object exposes. It stays green even though the predicate the connector actually builds at runtime is accept-all, so it cannot detect the production gap where `database-names` never reaches `database.include.list`.
- Fix: add a variant that builds `RelationalTableFilters` from a real `PostgresConnectorConfig` constructed from the Debezium properties `PostgresSourceConfigFactory.create()` emits for a `database-names` configuration, and assert that non-matching databases are skipped. Such a test would fail while the production gap exists and would lock in the end-to-end fix.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:43  RelationalTableFilters tableFilters = Mockito.mock(RelationalTableFilters.class);`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:45  when(tableFilters.databaseFilter()).thenReturn("selected"::equals);`; `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:46  when(tableFilters.dataCollectionFilter()).thenReturn(tableId -> true);`
- Suggested fix: Stubbing `databaseFilter()` with `"selected"::equals` only proves `listTables` consults the predicate; it doesn't prove the fix works with the filter the connector actually builds at runtime. Since `PostgresSourceConfigFactory` doesn't set `database.include.list`, the real predicate is accept-all and this test would still pass. Please add a variant that constructs `RelationalTableFilters` from a real `PostgresConnectorConfig` built from the factory's Debezium properties, so the test fails while the production gap exists.
- Severity: Medium

**Issue 3: MockJdbcConnection extends the real Debezium `JdbcConnection` with a null-returning connection factory and is never closed, which is fragile and can leak Debezium-managed resources in the test JVM.**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:66`
- Why it matters: - Root cause: the mock calls `super(..., config -> null, "\"", "\"")` and the test instantiates it without try-with-resources; `JdbcConnection` is `AutoCloseable` and manages internal state for the connection lifecycle.
- Impact: any future change in `TableDiscoveryUtils.listTables` that calls a base `JdbcConnection` method other than the overridden `query(String, ResultSetConsumer)` (e.g. `connection()`, `prepareQuery`, quoting helpers) will NPE inside Debezium with a confusing stack trace rather than a clear test failure, and unclosed connections set a poor pattern for future tests that plug in a real connection.
- Fix: wrap the mock in try-with-resources in the test (or call `jdbc.close()` in a finally block), and consider overriding `connection()` to throw an explicit `AssertionError("unexpected connection access")` so failures are self-explanatory.
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:48  MockJdbcConnection jdbc = new MockJdbcConnection();`
- Suggested fix: `config -> null` means any inherited `JdbcConnection` method that touches the underlying connection will NPE deep inside Debezium, and the instance is never closed even though `JdbcConnection` is `AutoCloseable`. Please use try-with-resources around `MockJdbcConnection` in the test and override `connection()` to throw a descriptive `AssertionError`, so future changes to `listTables` fail with a clear message instead of an opaque NPE.
- Severity: Low

**Issue 4: Asserting the exact SQL text (quote characters, trailing semicolon) makes the test brittle**
- Location: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:57`
- Why it matters: - Root cause: the test asserts the recorded queries against full literal SQL strings, including the `\"` quoting characters and the trailing `;`, which are incidental formatting details of `TableDiscoveryUtils.listTables`.
- Impact: any harmless change to the SQL construction (different quote style, removing the semicolon, whitespace) will break this test even though the filtering behavior under test is unchanged, adding maintenance friction.
- Fix: assert the essential behavior instead — e.g. that the recorded queries contain exactly one per-database metadata query, that it references `selected`, and that no query mentions `unwanted` (`assertTrue(queries.stream().noneMatch(q -> q.contains("unwanted")))`).
- Evidence: `seatunnel-connectors-v2/connector-cdc/connector-cdc-postgres/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/postgres/utils/TableDiscoveryUtilsTest.java:56  "select datname from pg_database",`
- Suggested fix: This assertion compares the recorded queries against full literal SQL strings (quote chars, trailing `;`), which couples the test to incidental formatting of `listTables`. Consider asserting the intent instead: exactly one metadata query was issued, it references `selected`, and no query mentions `unwanted`. That keeps the regression guarantee while surviving harmless SQL formatting changes.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: The new database filter guard is a no-op under default SeaTunnel Postgres CDC configuration because `PostgresSourceConfigFactory` never populates Debezium's `database.include.list`, so #8184 is not actually fixed for users who set `database-names`.

**2. Remaining follow-ups**
- Issue 2: The regression test stubs `databaseFilter()` with a hand-written predicate, so it only verifies the guard consults the mock — it does not prove that a filter derived from real SeaTunnel/Debezium configuration filters anything.
- Issue 3: MockJdbcConnection extends the real Debezium `JdbcConnection` with a null-returning connection factory and is never closed, which is fragile and can leak Debezium-managed resources in the test JVM.
- Issue 4: Asserting the exact SQL text (quote characters, trailing semicolon) makes the test brittle

DanielLeens:
Thanks @SEZ9 for going back through this one with even more care — the additional test-hygiene points (the `MockJdbcConnection` never being closed, and the exact-SQL-string assertions being brittle) are good catches on top of Issue 1.

To confirm we're aligned: my read matches yours. Issue 1 (`database.include.list` never actually being populated from `database-names`) is still the real blocker, since that's what keeps #8184 from being fixed for the standard config path — the guard in `TableDiscoveryUtils` is correct in isolation but never fires in production today. No new commit has landed on this head since my last review, so I'm not starting another full pass yet.

Once a fix for Issue 1 lands (ideally with a non-mocked regression test per Issue 2, ideally reproducing the exact `database-names` config from the linked issue), I'll re-review from scratch and fold in your Issues 2-4 as well. Thanks again for staying on top of this.

DanielLeens (commented):
Thanks for the update, and thanks for sticking with this one through several rounds. I did a complete fresh re-review of the current head `74db0a10ae26` from scratch (full diff against `dev`, not incremental), since a new commit landed after my last comment on 2026-08-24.

- User pain (issue #8184): a Postgres CDC source configured with `database-names = ["postgres"]` still had `TableDiscoveryUtils` probe every database returned by `pg_database` (e.g. `postgres`, `database2`, `database3`, ...), issuing a per-database `INFORMATION_SCHEMA.TABLES` query that fails for every database other than the connected one (Postgres does not support cross-database queries on a single connection) and logging `WARN ... skipping database 'X' due to error reading tables: cross-database references are not implemented` for each one.
- Fix approach as of this head: two coordinated changes. (1) `TableDiscoveryUtils.listTables` (added in an earlier commit, unchanged here) skips a database before querying it if `tableFilters.databaseFilter().test(dbName)` returns false. (2) **New in this commit**: `PostgresSourceConfigFactory.create()` now actually populates Debezium's `database.include.list` from the user's configured `databaseList`, which is what the guard in (1) needed all along to be anything other than a no-op.
- One-sentence summary: this commit closes the exact gap that blocked merge in my 2026-08-17 review and SEZ9's 2026-08-23 review — the discovery guard was always correctly placed, but the filter it consulted was never populated for the standard `database-names` config path; it now is.

# What changed since my last review (2026-08-24) and does it address what was said then?

My CHANGES_REQUESTED review on 2026-08-17 (escalating SEZ9's 2026-07-14 Medium finding to Critical/Blocker) and SEZ9's 2026-08-23 High-severity Issue 1 both said the same thing: `PostgresSourceConfigFactory.create()` never sets `database.include.list`/`database.exclude.list` from `databaseList`, so `RelationalTableFilters.databaseFilter()` degenerates to accept-all in production, and the guard added to `TableDiscoveryUtils` is dead code for the issue's own reproduction config.

I diffed the previously-reviewed head (`d171887c3145`) against the current head (`74db0a10ae26`) directly and confirmed the *only* change is:

```java
// PostgresSourceConfigFactory.java:73 (new line)
props.setProperty("database.include.list", String.join(",", databaseList));
```

plus one new assertion in the existing `PostgresSourceConfigFactoryTest`:

```java
Assertions.assertEquals(
        "database",
        configFactory.create(0).getDbzConfiguration().getString("database.include.list"));
```

`TableDiscoveryUtils.java` and `TableDiscoveryUtilsTest.java` are byte-identical to the head I already reviewed. This is exactly the fix I asked for in the "Best improvement" of my Issue 1 ("populate `database.include.list` in `PostgresSourceConfigFactory.create()` from `databaseList`"). **My Issue 1 (Critical/Blocker) and SEZ9's equivalent Issue 1 (High) are now resolved** — see the independent verification below, which traces the mechanism all the way into the Debezium library source rather than relying on the diff alone.

# 1. Code Change Review

## 1.1 Core Logic Analysis

### Exact change (this commit)

`PostgresSourceConfigFactory.java:73`:
```java
props.setProperty("database.dbname", checkNotNull(databaseList.get(0)));
props.setProperty("database.include.list", String.join(",", databaseList));
```

### Runtime path, traced end to end (including into Debezium 1.9.8.Final library source, not just SeaTunnel code)

```text
PostgresSourceConfigFactory.create(subtask)                              [PostgresSourceConfigFactory.java:73]
  -> props["database.include.list"] = join(",", databaseList)            <- NEW
  -> new PostgresSourceConfig(..., props, ...)
       -> PostgresSourceConfig.getDbzConnectorConfig()                   [PostgresSourceConfig.java:88-90]
            -> new PostgresConnectorConfig(getDbzConfiguration())
                 -> RelationalDatabaseConnectorConfig(...)                (debezium-core, inherited — Postgres
                                                                            connector does not override table-filter
                                                                            construction)
                      -> this.tableFilters = new RelationalTableFilters(config, ...)
                           -> databaseFilter = Selectors.databaseSelector()
                                  .includeDatabases(config.getString(DATABASE_INCLUDE_LIST))  // "database.include.list"
                                  .excludeDatabases(config.getString(DATABASE_EXCLUDE_LIST))
                                  .build()
  -> PostgresDialect.discoverDataCollections(sourceConfig)                [PostgresDialect.java:109]
       -> TableDiscoveryUtils.listTables(jdbc, tableFilters)
            -> for each dbName from `pg_database`:
                 tableFilters.databaseFilter().test(dbName)               [TableDiscoveryUtils.java:63]
                 -> continue (skip query+warning) if not in databaseList
```

I pulled the Debezium 1.9.8.Final sources (`debezium-core`, `debezium-connector-postgres`) locally and read the actual predicate construction, since the previous rounds of this review correctly established that "the property gets set" is not sufficient proof — the predicate semantics matter:

- `Selectors.DatabaseSelectionPredicateBuilder.includeDatabases(names)` → `Predicates.includes(names)` → splits on `,`, compiles each segment as `Pattern.compile(segment, CASE_INSENSITIVE)`.
- The match test used is `pattern.matcher(dbName).matches()` — a **full, anchored, case-insensitive** match per comma-separated segment (`Predicates.java`, `matchedByPattern`). Not a substring/`contains` match.
- `PostgresConnectorConfig` does not override `DATABASE_INCLUDE_LIST`/`getTableFilters()` — it inherits `RelationalDatabaseConnectorConfig`'s generic implementation, so `"database.include.list"` is exactly the config key Debezium reads for Postgres too (verified: no Postgres-specific override exists in `PostgresConnectorConfig.java`).
- `databaseFilter()`/`database.include.list` is consumed **exactly once** in the entire `connector-cdc-postgres` module: `TableDiscoveryUtils.java:63`. (Confirmed via `grep -rn "databaseFilter()"` and `grep -rn "getTableFilters()"` across `connector-cdc-postgres`.) `dataCollectionFilter()` — the separate table/schema-level filter driven by `table.include.list`/`schema.include.list` — is unaffected by this change.

### Key findings

1. `database-names` (backing `databaseList`) is a **required** option for the Postgres CDC source (`PostgresIncrementalSourceFactory.java:57-63`, `.required(..., JdbcSourceOptions.DATABASE_NAMES, ...)`), and the existing line right above the new one (`checkNotNull(databaseList.get(0))`) already assumed it non-null/non-empty. So `String.join(",", databaseList)` cannot NPE and this guard now fires for **every** Postgres CDC job on the standard config path — not just multi-database or pattern-matching configs. This directly answers the "does the normal startup path reach this filter" question: yes, unconditionally.
2. For the exact reproduction in issue #8184 (`database-names = ["postgres"]`, other unrelated databases present on the server), every database in `databaseList` compiles to a literal, case-insensitive, exact-match pattern against itself, so it always passes its own filter — there is no scenario under the standard config path where a user's own configured database gets false-negative-excluded by this change. The only databases now skipped are ones the user never listed, which previously errored out with a caught `SQLException` and a `WARN` log anyway (Postgres has no cross-database queries within one connection), so no `TableId` that was ever actually reachable before is now unreachable.
3. This closes the previously-verified gap: `PostgresSourceConfigFactory.create()` now sets `database.include.list`, so `databaseFilter()` is no longer accept-all for standard jobs, and the guard in `TableDiscoveryUtils` is live code, not dead code.

### Minor edge-case notes (not blockers, flagged for completeness per the mandatory regex/pattern review)

- **Regex-special characters in database names**: each entry of `databaseList` is joined verbatim into a comma-separated Debezium include-list and compiled as a regex per segment. A database name containing regex metacharacters (`.`, `+`, `(`, `)`, `[`, etc.) would be compiled as a pattern rather than escaped as a literal, so in a contrived case it could over-match a similarly-named database. This is the same convention already used two lines below for `table.include.list`/`schema.include.list` in this same file, so it is a pre-existing, codebase-wide pattern, not something newly introduced by this diff — Low severity, informational only.
- **Escape-hatch interaction**: Debezium's own `Selectors.DatabaseSelectionPredicateBuilder.build()` gives include-list priority over exclude-list (`dbInclusions != null ? dbInclusions : dbExclusions`). Since `database.include.list` is now unconditionally set from `databaseList`, any advanced user who was passing a raw `debezium.database.exclude.list` via the `dbzProperties` passthrough (without setting an include list) to influence database-level filtering would now have that exclude-list silently outrank by the auto-populated include-list. I checked whether this is an actual regression: it is not, because `databaseFilter()` was **not consumed anywhere** in `connector-cdc-postgres` before this PR added the `TableDiscoveryUtils` guard — so there was no working exclude-list-only behavior at this call site to break. Worth a one-line doc/comment mention for future maintainers, but Low severity and non-blocking.
- I confirmed SeaTunnel's `PostgresSourceConfig.getDbzConnectorConfig()` constructs `new PostgresConnectorConfig(getDbzConfiguration())` directly, without calling Debezium's `Configuration.validate(FIELDS, problems)` — so `RelationalDatabaseConnectorConfig.validateDatabaseExcludeList` (which would flag simultaneous include+exclude as a config problem) is never invoked here. No startup-time validation failure risk from setting both.

## 1.2 Compatibility Impact

Safe and net-positive. No public config option (`database-names`, its type, or default) changed. The internal Debezium property `database.include.list` was previously never set (accept-all); it is now always set to the user's own configured `databaseList`. Because `database-names` is required and every database a user lists always self-matches its own literal pattern, no existing job loses access to a database it explicitly configured — the only behavior removed is querying (and failing on, and warning about) databases the user never asked for. This does not touch checkpoint/savepoint state, serialization format, or the wire protocol.

Positive, narrow-scope improvement: for a server with N databases and a `database-names` list of size k < N, this removes (N-k) doomed `INFORMATION_SCHEMA.TABLES` queries and their `SQLException`-driven `WARN` log lines per job startup/discovery cycle. No new resource lifecycle, retry, thread, or queue behavior introduced.

**Issue 1 (carried over from my 2026-08-17 review and SEZ9's 2026-08-23 review, Critical/Blocker → now RESOLVED): `database.include.list` was never populated from `database-names`.**
- Location: `PostgresSourceConfigFactory.java:73` (fix location); guard site `TableDiscoveryUtils.java:63`.
- Status: **Fixed** in this commit. Verified via full source trace through `RelationalDatabaseConnectorConfig`/`RelationalTableFilters`/`Selectors` in the actual Debezium 1.9.8.Final jar used by this module (not assumed from the diff alone). Raised by another reviewer: Yes — originally SEZ9 (2026-07-14, Medium), escalated by me (2026-08-17, Critical/Blocker), re-confirmed independently by SEZ9 (2026-08-23, High).

**Issue 2 (carried over, partially addressed): the regression coverage still does not prove the filter excludes anything end-to-end.**
- Location: `PostgresSourceConfigFactoryTest.java:43-45` (new assertion); `TableDiscoveryUtilsTest.java:43-46` (unchanged, still mocks `RelationalTableFilters` with a hand-written predicate).
- Problem description: the new assertion in `PostgresSourceConfigFactoryTest` proves that `create(0)` sets the Debezium property string `database.include.list=database` correctly for a single-element `databaseList("database")`. It does **not** construct a real `PostgresConnectorConfig`/`RelationalTableFilters` from that configuration and assert that `databaseFilter().test("unwanted")` is `false` while `databaseFilter().test("database")` is `true`; nor does it exercise the multi-database case (`databaseList("a", "b")` → `database.include.list=a,b`) which is the literal shape of the #8184 reproduction. `TableDiscoveryUtilsTest` is untouched and still bypasses `PostgresSourceConfigFactory`/real Debezium config parsing entirely via a Mockito mock.
- Potential risk: this is now materially lower risk than when I flagged it as Blocker-adjacent on 2026-08-17, because I have independently verified (by reading the Debezium library source, not by trusting this PR's test) that the property-to-predicate wiring is correct, standard, well-established Debezium behavior. But there is still no automated regression guard that would fail if a future refactor accidentally dropped the `database.include.list` line, renamed the property, or introduced an escaping bug for multi-database lists.
- Best improvement: add one test that builds `RelationalTableFilters` from a real `PostgresConnectorConfig` constructed off `configFactory.create(0).getDbzConfiguration()` for a `databaseList("selected")` / other-database-present scenario, and asserts `databaseFilter().test("unwanted")` is false. This is the same suggestion from both my 2026-08-17 review and SEZ9's 2026-08-23 review, still open.
- Severity: Medium (downgraded from my prior Critical/Blocker rating on Issue 1, since the underlying fix is now confirmed correct via independent source verification — this is now a "lock the fix in for the future" gap, not a "the fix might not work" gap).
- Raised by another reviewer: Yes — SEZ9 (2026-08-23, Issue 2, Medium); carried over from SEZ9's original 2026-07-14 Issue 2.

**Issue 3 (carried over, unaddressed): `MockJdbcConnection` extends the real Debezium `JdbcConnection` with a null-returning connection factory and is never closed.**
- Location: `TableDiscoveryUtilsTest.java:66` (`config -> null`), instantiation without try-with-resources.
- Status: unchanged since SEZ9's 2026-08-23 review. Same fix suggestion stands (try-with-resources, override `connection()` to throw a descriptive `AssertionError`).
- Severity: Low.
- Raised by another reviewer: Yes — SEZ9, 2026-08-23.

**Issue 4 (carried over, unaddressed): exact-string SQL assertions make `TableDiscoveryUtilsTest` brittle to harmless query refactors.**
- Location: `TableDiscoveryUtilsTest.java:57` (`"select datname from pg_database"`, full literal `SELECT * FROM "selected"...` string).
- Status: unchanged. Same fix suggestion stands (match on stable substrings / assert the `unwanted` invariant rather than full-string equality).
- Severity: Low.
- Raised by another reviewer: Yes — SEZ9 (2026-08-23) and originally SEZ9 (2026-07-14).

## 2. Code Quality Assessment

## 2.1 Coding Standards

The production diff is a single, minimally-scoped line consistent with the existing style two lines below it (`table.include.list` construction). No style concerns.

Mandatory stability rating (UT changed in this commit — `PostgresSourceConfigFactoryTest.java`):

**Rating: Stable.**
- The new assertion is fully deterministic: it constructs an in-memory `PostgresSourceConfigFactory`, calls `create(0)`, and reads a property string back off the resulting `Configuration` object. No I/O, no timing dependency, no external service, no shared mutable state, no thread interaction. It cannot flake.
- Coverage gap (see Issue 2 above) is a correctness/completeness concern, not a stability concern — the test that exists will always pass or fail deterministically; it just doesn't yet test as much as it should.

## 2.3 Documentation Updates

No user-facing config option name, type, or default changed by this PR — `database-names` already existed and already had the same documented meaning ("Database name of the database to monitor"). Consistent with the prior consensus from both my 2026-08-17 review and SEZ9's reviews, I don't consider a `docs/en` / `docs/zh` update strictly required for this diff's scope, though a short note in the Postgres CDC connector doc about database-scoped discovery would be a reasonable, non-blocking follow-up given this was confusing enough to produce issue #8184. Severity: Low, non-blocking (carried over).

# 3. Architectural Soundness

The fix now completes the intended design: reuse Debezium's own `RelationalTableFilters` abstraction end-to-end (config → predicate → guard) rather than reinventing database-name filtering in SeaTunnel code. This is the right shape and the minimal correct change — one property assignment, no new abstractions, no new classes.

## 3.2 Maintainability

The previous maintainability risk I flagged on 2026-08-17 — "a reader of `TableDiscoveryUtils.java:63` alone would assume `database-names` already drives this filter, when in fact it doesn't" — is now resolved, because `database-names` genuinely does drive it. The remaining maintainability gap is Issue 2: nothing currently regression-tests the property-to-predicate wiring itself, so a future refactor of `PostgresSourceConfigFactory` could silently reintroduce the original bug without any test failing.

## 3.3 Extensibility

Unchanged from my prior assessment: this pattern extends cleanly to `database.exclude.list` if ever needed. The pre-existing, separate observation that only `databaseList.get(0)` is used for the JDBC connection's own `database.dbname` (a single-database-connection limitation of the underlying Postgres JDBC driver, not something this PR introduces or is expected to fix) still stands as out-of-scope context, not a defect in this diff.

No compatibility risk. This PR does not rename, remove, or change the default of any user-facing option; does not touch checkpoint/savepoint state, offset format, or the wire protocol. Safe to merge from a compatibility standpoint.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | `database.include.list` never populated from `database-names`, so the discovery guard was a no-op for the standard config path — **RESOLVED in this commit**, verified via independent trace into Debezium 1.9.8.Final source | `PostgresSourceConfigFactory.java:73` | Resolved (was Critical/Blocker) |
| 2 | Regression tests still don't build a real `RelationalTableFilters`/`PostgresConnectorConfig` from the factory's own config and assert end-to-end exclusion; no multi-database test case | `PostgresSourceConfigFactoryTest.java:43-45`; `TableDiscoveryUtilsTest.java:43-46` | Medium |
| 3 | `MockJdbcConnection` never closed, null connection factory risks opaque NPEs on future changes | `TableDiscoveryUtilsTest.java:66` | Low |
| 4 | Exact-string SQL assertions are brittle to harmless query refactors | `TableDiscoveryUtilsTest.java:57` | Low |
| 5 | No connector doc update for the (internal) discovery/filtering behavior change | N/A | Low |

# 5. Merge Recommendation

The Critical/Blocker gap from my 2026-08-17 review and SEZ9's High-severity Issue 1 from 2026-08-23 — the entire reason this PR did not actually fix issue #8184 for its own reproduction config — is resolved in this commit, and I verified that independently by reading the Debezium 1.9.8.Final library source rather than trusting the diff or the PR description at face value.

1. Blockers — none. Issue 1 is closed.
2. Recommended fixes — non-blocking, but I'd like to see at least Issue 2 addressed before merge given how much of this PR's review history was specifically about test coverage failing to prove the fix works:
   - Issue 2 (Medium): add a test that builds a real `RelationalTableFilters`/`PostgresConnectorConfig` from `PostgresSourceConfigFactory`'s own output and asserts an unlisted database is excluded — ideally covering the multi-database case from #8184 directly, so the regression suite would fail if this fix were ever silently reverted.
   - Issue 3, 4 (Low): test-hygiene cleanups on `TableDiscoveryUtilsTest` (resource lifecycle, brittle string assertions) — nice to have, not blocking.
   - Issue 5 (Low): optional doc note.

Overall assessment: this is a clean, minimal, correctly-targeted one-line fix that closes the exact gap flagged across three prior review rounds (mine and SEZ9's). I independently re-derived the fix's correctness from the Debezium library source rather than re-trusting the PR description, and I'm confident it now does what it claims to do for the standard `database-names` config path, including the exact configuration in issue #8184. CI (`Build`) is still queued on the current head at the time of this review, so I can't fold that signal in yet, but there's no code-side blocker left from my side. As always, I review with read-level permission, so a write-capable maintainer performs final approval and merge — I'd suggest merging once CI is green, with Issue 2's stronger regression test either included now or as an immediate fast-follow.

DanielLeens:
CI update on this exact head (`74db0a10ae269`): the `Build` run I noted as still queued at review time has since completed on the fork (`https://github.com/zhangshenghang/seatunnel/actions/runs/33499602605`) with 8 failing jobs, and I want to flag one of them specifically because it isn't the usual unrelated-module noise: `updated-modules-integration-test-part-1` failed on **both** JDK 8 and JDK 11, and the actual failure is inside `connector-cdc-postgres-e2e` (`PostgresCDCIT`) — 27 errors + 1 failure out of 37 tests, all `ConditionTimeout` (awaitility assertion timeouts), e.g. `testPostgresCdcMultiTableE2e`, `testPostgresCdcCheckDataWithNoPrimaryKey`, `testMPostgresCdcCheckDataE2e`. That's the exact connector this PR's `database.include.list` change touches, and the failure rate (nearly the whole suite, on both JDKs) is high enough that I don't think this can be waved off as routine container-startup flake without checking — it could plausibly be the new `database.include.list` population interacting with however the E2E's own Postgres container/database names are set up. The other 7 failing jobs (`unit-test (8, ubuntu-latest)` -> `seatunnel-engine-server`, `iceberg-connector-it`, `doris-connector-it`) look unrelated to this diff and I'm not raising those.

Given this shows up specifically in the module this PR changes, could you check whether `testPostgresCdcMultiTableE2e` and friends still pass locally against this head before we treat 'CI is the only remaining gate, no code-side blocker' as still accurate? I don't want to call this a confirmed blocker without your own repro, but I also don't want to leave that framing standing unqualified while this is unresolved.

PR #11412, opened 2026-07-12
[Fix][Common] Reject invalid fractional time separators

hiSandog:

TimeUtils uses an unescaped dot in the HH:mm:ss.SSS detection pattern. Because dot is a regex wildcard, values such as 12:12:12x123 are classified as millisecond timestamps even though DateTimeFormatter cannot parse them.

This patch escapes the decimal point and adds a regression assertion for the invalid separator.

Yes. Fractional time values must now use a literal dot to be recognized as HH:mm:ss.SSS. Valid values such as 12:12:12.123 are unchanged.

- Added a negative assertion to TimeUtilsTest for 12:12:12x123.
- git diff --check passes.
- An equivalent local regex check confirmed that 12:12:12.123 is accepted and 12:12:12x123 is rejected.
- The Maven test command could not run locally because the environment has no Java runtime; upstream CI will compile and run the test.

### Check list

* [x] No new Jar binary packages are added.
* [x] No documentation update is necessary for this narrow validation fix.
* [x] No incompatible change entry is necessary.
* [x] This does not modify connector code.

### DIFF
--- a/seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/TimeUtils.java
+++ b/seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/TimeUtils.java
@@ -45,7 +45,7 @@ public class TimeUtils {
     public static final Pattern[] PATTERN_ARRAY =
             new Pattern[] {
                 Pattern.compile("\\d{2}:\\d{2}:\\d{2}"),
-                Pattern.compile("\\d{2}:\\d{2}:\\d{2}.\\d{3}"),
+                Pattern.compile("\\d{2}:\\d{2}:\\d{2}\\.\\d{3}"),
             };
 
     public static Formatter matchTimeFormatter(String dateTime) {
--- a/seatunnel-common/src/test/java/org/apache/seatunnel/common/utils/TimeUtilsTest.java
+++ b/seatunnel-common/src/test/java/org/apache/seatunnel/common/utils/TimeUtilsTest.java
@@ -32,5 +32,7 @@ public class TimeUtilsTest {
         Assertions.assertEquals(
                 "12:12:12.123",
                 TimeUtils.parse(timeStr, TimeUtils.matchTimeFormatter(timeStr)).toString());
+
+        Assertions.assertNull(TimeUtils.matchTimeFormatter("12:12:12x123"));
     }
 }


DanielLeens (commented):
Good job.

I reviewed the full current diff for this PR. The source change is intentionally small and correct: `TimeUtils.matchTimeFormatter()` previously treated the `.` in `HH:mm:ss.SSS` as an arbitrary regex character, so strings such as `12:12:12x123` could be accepted as fractional-second time values. The PR now escapes the separator as `\.` and adds a regression assertion for that invalid separator.

Runtime chain checked:
```text
User/string time input
  -> TimeUtils.matchTimeFormatter(dateTime)
      -> PATTERN_ARRAY[0] matches HH:mm:ss
      -> PATTERN_ARRAY[1] now matches only HH:mm:ss.SSS with a literal dot
  -> TimeUtils.parse(time, formatter)
      -> LocalTime.parse(time, DateTimeFormatter.ofPattern(formatter.value))
```

Current source conclusion from Daniel's side: I do not see a code-side blocker in this PR. The behavior change is backward-compatible for valid `HH:mm:ss` and `HH:mm:ss.SSS` inputs, and the new test is deterministic with no external resource or timing dependency.

The remaining gate is CI/workflow setup, not the source diff. To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration: https://github.com/apache/seatunnel/runs/86626413409

image

### Merge conclusion: source clear, waiting for CI action
Blocking items: none from Daniel's source review; CI workflow/action is still required before this can be approved.
Suggested follow-ups: none.

hiSandog:
Enabled GitHub Actions in my fork and triggered the Build workflow with c74add3e8. The fork CI is now running: https://github.com/hiSandog/seatunnel/actions/runs/29194229700

nzw921rx (approved):
+1 LGTM if CI pass

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head locally from the full current diff, not just the CI trigger commit.

Source-side conclusion
- The actual code change here is still the same one-line regex fix in `TimeUtils.matchTimeFormatter(...)`: escape the fractional-second dot so `HH:mm:ss.SSS` no longer accepts an arbitrary separator like `12:12:12x123`.
- I rechecked the call path and I still agree with the direction. This is the right fix for the bug, and the added regression assertion covers the bad input that previously slipped through.
- I do not see a reopened Daniel-side code blocker on the current head.

Current remaining gate
- base branch: `dev`
- head SHA: `c74add3e860efb18f8990a42fd22516e8a892474`
- compare status: `diverged`
- ahead_by / behind_by: `2 / 6`
- aggregate `Build`: `failure`
- check URL: https://github.com/apache/seatunnel/runs/86654217641

I also checked the linked fork workflow. The current failures are broad CI lanes (`unit-test` on multiple platforms and unrelated connector integration lanes), not something that points back to this tiny `TimeUtils` diff specifically. Given that this branch is also behind the latest `dev`, the lowest-cost next step is to sync the latest `dev` and rerun CI first.

Merge conclusion for this round
- Daniel source-side blocker: none on the current head
- Remaining gate: CI / merge gate only
- Next step: sync latest `dev` and rerun the failing checks on the updated head

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `c74add3e860e` from scratch.

# What this PR fixes
- User pain: `TimeUtils.matchTimeFormatter` misclassified strings like `12:12:12x123` as `HH:mm:ss.SSS` because the unescaped `.` in the detection regex acted as a wildcard, so the subsequent `LocalTime.parse` blew up with a confusing `DateTimeParseException`.
- Fix approach: Escape the dot (`\\.`) in the second entry of `PATTERN_ARRAY` in `TimeUtils.java` so only a literal decimal point matches, plus a negative assertion in `TimeUtilsTest` confirming `matchTimeFormatter("12:12:12x123")` now returns `null`.
- One-line summary: The one-character regex fix is correct and minimal, but the failure mode for malformed input shifts from a descriptive `DateTimeParseException` to a `null` formatter — and likely an NPE at the `parse()` call site — which the new test doesn't cover and the release notes should mention.

# Runtime chain I rechecked
```text
caller (e.g. type-inference / row conversion code) -> TimeUtils.matchTimeFormatter(dateTime)  TimeUtils.java:51
  matchTimeFormatter iterates PATTERN_ARRAY[i].matcher(dateTime).matches()  TimeUtils.java:45-56  <- changed regex here (index 1, HH:mm:ss.SSS)
    on match: returns Formatter (e.g. Formatter.HH_MM_SS_SSS); on no match: returns null  <- "12:12:12x123" now falls into the null branch
  caller passes result to TimeUtils.parse(dateTime, formatter)  TimeUtils.java:39-42
    parse -> FORMATTER_MAP.get(formatter) -> NPE on null key lookup / LocalTime.parse(dateTime, dateTimeFormatter)  <- previously threw DateTimeParseException for the wildcard-matched input, now NPEs unless the caller null-checks
```

# Findings
**Issue 1: Tightened regex silently widens the set of inputs for which matchTimeFormatter returns null; callers that feed the result straight into parse() will now hit an NPE instead of a descriptive parse error**
- Location: `seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/TimeUtils.java:48`
- Why it matters: - Root cause: before this change, a malformed value such as `12:12:12x123` matched the wildcard-dot pattern and `matchTimeFormatter` returned `Formatter.TIME_HH_MM_SS_SSS`; the subsequent `DateTimeFormatter.parse` then failed with a `DateTimeParseException` that included the offending text. With the escaped dot, `matchTimeFormatter` returns `null` for these values (as the new test asserts), so any caller doing `TimeUtils.parse(timeStr, TimeUtils.matchTimeFormatter(timeStr))` — the exact idiom used in `TimeUtilsTest` — will now throw a bare `NullPointerException`.
- Impact: connectors and transforms on Zeta/Flink/Spark that route user-supplied time strings through this utility lose the actionable error message; a job failing on dirty data will surface as an NPE stack trace with no hint about which value or format was rejected, which hurts troubleshooting of ingestion pipelines.
- Fix: in `TimeUtils.parse(String)` (and any other internal caller), check the formatter for `null` and throw a `SeaTunnelRuntimeException` via `CommonError`/`CommonErrorCode.UNSUPPORTED_DATA_TYPE`-style reporting that echoes the unparseable value, mirroring how `DateUtils`/`DateTimeUtils` report unmatched formats. Alternatively, annotate `matchTimeFormatter` as `@Nullable` and document the contract so callers are forced to handle it.
- Evidence: `seatunnel-common/src/test/java/org/apache/seatunnel/common/utils/TimeUtilsTest.java:36  Assertions.assertNull(TimeUtils.matchTimeFormatter("12:12:12x123"));`
- Suggested fix: The escape itself is correct, but note the behavioral side effect: inputs like `12:12:12x123` previously produced a `DateTimeParseException` (with the bad value in the message) at parse time, and now `matchTimeFormatter` returns `null`, so callers that chain it into `parse()` will NPE. Please make the no-match path fail loudly and descriptively — e.g., have `TimeUtils.parse(String)` throw a `SeaTunnelRuntimeException` naming the unparseable value when no pattern matches — so dirty-data failures in jobs remain diagnosable.
- Severity: Medium

**Issue 2: New negative test only covers matchTimeFormatter; parse() behavior on rejected input (likely NPE) is untested**
- Location: `seatunnel-common/src/test/java/org/apache/seatunnel/common/utils/TimeUtilsTest.java:36`
- Why it matters: - The added assertion `Assertions.assertNull(TimeUtils.matchTimeFormatter("12:12:12x123"))` verifies the classification result, but does not verify what `TimeUtils.parse(dateTime, formatter)` does when the formatter is now `null`.
- Before this fix, an invalid separator produced a `DateTimeParseException` from `DateTimeFormatter`; after this fix, callers that chain `TimeUtils.parse(timeStr, TimeUtils.matchTimeFormatter(timeStr))` (exactly the pattern used a few lines above in this same test) will hit a `NullPointerException` with no context about the bad input. That is a worse failure mode for users debugging bad time values flowing through connectors on Zeta/Flink/Spark.
- Please add a test asserting the exception type/message when parsing `"12:12:12x123"` end-to-end, and consider having `matchTimeFormatter` (or the call sites) throw a `SeaTunnelRuntimeException` via `CommonError` with the offending value instead of returning `null` silently, following SeaTunnel's error-code conventions.
- Suggested fix: Nice regression assertion. Could we also add a test that exercises the full chain used above, i.e. `TimeUtils.parse("12:12:12x123", TimeUtils.matchTimeFormatter("12:12:12x123"))`? With this fix the formatter is now `null`, so callers will likely get a bare NPE instead of the previous `DateTimeParseException`. Asserting the actual failure mode here (or better, making `matchTimeFormatter` fail with a clear `CommonError`-based exception) would lock in a user-friendly error message for invalid time values.
- Severity: Medium

**Issue 3: Upgrade risk is low for valid data — every input reclassified by this change was already unparseable — but the failure mode changes for malformed data**
- Location: `seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/TimeUtils.java:48  Pattern.compile("\\d{2}:\\d{2}:\\d{2}\\.\\d{3}"),`
- Why it matters: - Root cause: tightening `\\d{2}:\\d{2}:\\d{2}.\\d{3}` to `\\d{2}:\\d{2}:\\d{2}\\.\\d{3}` at seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/TimeUtils.java:48 only removes matches where the 9th character was not a literal dot; `DateTimeFormatter.ofPattern("HH:mm:ss.SSS")` could never successfully parse any of those strings anyway, so no previously-working pipeline output changes on any engine (Zeta/Flink/Spark).
- Impact: the only observable compatibility delta is for pipelines ingesting dirty data: they previously failed with `DateTimeParseException` at parse time and will now see a null formatter/NPE (per the related finding), which can change error-classification or retry behavior in user-side exception handling.
- Fix: state this explicitly in the PR description ('no valid input changes result; invalid inputs change failure mode from DateTimeParseException to X'), and resolve the null-formatter path so the post-upgrade exception is deterministic and documented.
- Severity: Low

**Issue 4: User-facing behavior change should be noted in release notes despite the checklist saying otherwise**
- Location: `seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/TimeUtils.java:48  Pattern.compile("\\d{2}:\\d{2}:\\d{2}\\.\\d{3}"),`
- Why it matters: - The PR description states 'Does this PR introduce any user-facing change? Yes' (values with a non-dot separator are no longer recognized as `HH:mm:ss.SSS`), yet the checklist claims 'No incompatible change entry is necessary' — these two statements conflict.
- While the previous behavior was clearly a bug (misclassified values failed anyway, just with a different exception), the observable failure mode changes for downstream users who catch specific exceptions, so a one-line entry in the release notes / changelog for the next release would help.
- Please add a short note in the release-notes doc (or ensure the PR title `[Fix][Common]` is picked up for the changelog) describing that fractional time values must now use a literal dot to be recognized.
- Severity: Low

# Review conclusion

- Issue 1: Tightened regex silently widens the set of inputs for which matchTimeFormatter returns null; callers that feed the result straight into parse() will now hit an NPE instead of a descriptive parse error
- Issue 2: New negative test only covers matchTimeFormatter; parse() behavior on rejected input (likely NPE) is untested
- Issue 3: Upgrade risk is low for valid data — every input reclassified by this change was already unparseable — but the failure mode changes for malformed data
- Issue 4: User-facing behavior change should be noted in release notes despite the checklist saying otherwise

DanielLeens:
Thanks for the extra review context, @SEZ9. I rechecked the same path and I agree the `null` formatter behavior is worth keeping visible, especially for `TimeUtils.parse(String)` and the HBase sink path that calls `TimeUtils.parse(field.toString())`.

For this PR, I would still keep Daniel's source-side conclusion as non-blocking: valid `HH:mm:ss` and `HH:mm:ss.SSS` inputs are unchanged, and the values reclassified by the escaped dot were already invalid for `DateTimeFormatter` before this fix. So I see the clearer invalid-input exception as a useful follow-up improvement rather than something this one-character bug fix must solve before merging.

Current remaining gate from my side is still CI / merge gate only: the branch is `diverged` from `dev` (`ahead_by=2`, `behind_by=6`) and the aggregate `Build` check is failing at https://github.com/apache/seatunnel/runs/86654217641. Since the available signal does not point back to this tiny `TimeUtils` diff, the lowest-cost next step remains syncing the latest `dev` and rerunning Build on the updated head.

davidzollo:

```

```

PR #11414, opened 2026-07-12
[Feature][Transform-V2] Add TextChunk transform plugin

### LINKED ISSUE #9717: Support text splitter transform (Chunking).

ss666:
Closes #9717. 

Adds a new `TextChunk` plugin that splits a long text field into multiple smaller chunks for RAG / embedding pipelines, emitting one row per chunk (a 1-row → N-rows FlatMap). Each output row keeps all original source fields and appends two columns: the chunk text (`chunk`) and its 0-based index within the source document (`chunk_index`).

Main points:

1. `TextChunker`: a pure, dependency-free chunking algorithm. It splits recursively by a priority list of `separators`, then greedily merges pieces up to `chunk_size`; when `separators` is empty it falls back to fixed-size windows. Overlap between adjacent chunks is carried as whole trailing pieces (never mid-word) in the separator path; boundaries are aligned to whole Unicode code points so a surrogate pair (e.g. an emoji) is never split.
2. `TextChunkTransform` / `TextChunkTransformFactory` / `TextChunkMultiCatalogTransform`: the SPI wiring — options, check-time validation, schema extension, and multi-catalog dispatch.
3. Options: `text_field`, `output_field`, `chunk_index_field`, `chunk_size`, `overlap_size`,  `separators`.
4. Docs (`docs/en` + `docs/zh`) and unit/E2E tests added.

Yes — a new optional transform plugin `TextChunk`. It is purely additive: no existing option key, default, public API or SPI contract is changed, so existing jobs are unaffected. New user-facing docs are added under `docs/en/transforms/text-chunk.md` and `docs/zh/transforms/text-chunk.md`.

Scoped to the changed module:

```bash
./mvnw spotless:apply -pl seatunnel-transforms-v2,seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-1
./mvnw -DskipTests verify -pl seatunnel-transforms-v2
./mvnw test -pl seatunnel-transforms-v2 \
  -Dtest='TextChunkerTest,TextChunkTransformTest,TextChunkTransformFactoryTest'
```

E2E: two configs added — `text_chunk_transform.conf` (single table; asserts exact chunk content and index bounds) and `text_chunk_transform_multi_table.conf` (per-table override + top-level regex fallback).

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md) — *N/A: no new jar/dependency.*
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs — *Added `docs/en` + `docs/zh` `text-chunk.md` and registered in `docs/sidebars.js`.*
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR. — *N/A: purely additive, no incompatibility.*
* [ ] If you are contributing the connector code, please check that the following files are updated: — *N/A: this is a Transform-V2 plugin, not a connector. It registers via `@AutoService` (SPI), and `seatunnel-transforms-v2` is already in `seatunnel-dist`, so `plugin-mapping.properties`, the dist pom, and `config/plugin_config` need no change.*


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current diff locally and traced the new transform through both schema production and row expansion on the latest head.

What this PR adds
- User pain: SeaTunnel currently lacks a built-in transform for splitting long text into smaller chunks before downstream embedding / indexing style flows.
- Fix approach: add a `TextChunk` transform with configurable chunk size, overlap, separator priority, tests, E2E coverage, and docs.
- One-line summary: this adds a first-class "split one long text row into many smaller text rows" transform.

Runtime path I checked
```text
job config
  -> TextChunkTransformFactory.createTransform()
      -> TextChunkMultiCatalogTransform.buildTransform()
          -> new TextChunkTransform(...)
              -> transformTableSchema()

data row
  -> TextChunkTransform.transformRow()
      -> TextChunker.split(...)
      -> one input row expands into N output rows
      -> each output row keeps source fields and appends chunk / chunk_index
```

I found one blocker on the latest head.

1. The output schema still copies the original primary key / constraint keys even though the transform changes one row into many rows.
- Location: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/chunk/TextChunkTransform.java:119`
- Why this is a real issue: `transformRow()` can turn one input row like `id=7, content=...` into multiple output rows with the same original `id`, but `transformTableSchema()` still copies the input `PrimaryKey` and `ConstraintKey` unchanged. That means the output table still claims the old key is unique when the produced data is no longer unique on that key.
- Risk: downstream sinks that rely on primary-key / unique-key semantics can now overwrite chunks, reject writes, or collapse multiple chunk rows into one logical row. This is a correctness issue on the main path, not a documentation nit.
- Better fix: either drop inherited PK / unique constraints by default for this one-to-many transform, or rebuild them explicitly with the new chunk dimension (for example including `chunk_index_field`) instead of copying the original definition unchanged.
- Severity: high

Other review notes
- The overall chunking direction looks good.
- The new UT / E2E coverage is useful, and I did not see obvious flaky-test patterns in the added tests.

Merge conclusion
- Blocking items: the output schema/key semantics above still need to be fixed.
- Non-blocking items: none beyond that for this round.
- Daniel conclusion: please address the schema/constraint issue first, then I’m happy to re-review the updated head.

seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/chunk/TextChunkTransformConfig.java
```diff
@@ -0,0 +1,119 @@
...
+        c.setTextField(config.get(TEXT_FIELD));
+        c.setOutputField(config.get(OUTPUT_FIELD));
+        c.setChunkIndexField(config.get(CHUNK_INDEX_FIELD));
+        c.setChunkSize(config.get(CHUNK_SIZE));
+        c.setOverlapSize(config.get(OVERLAP_SIZE));
+        c.setSeparators(config.get(SEPARATORS));
+        return c;
+    }
+
+    public static void validate(TextChunkTransformConfig config) {
```
nzw921rx: Thank you for your contribution. This is an essential supplement to RAG. Currently, 
the `validate (TextChunkTransformConfig config) `  function is used by Factory and has no common use. In terms of the scope of responsibilities of the class, this function is more suitable to be placed in the `TextChunkTransformFactory`. If you have any different opinions, please feel free to discuss

DanielLeens (commented):
Thanks for the quick update. I re-reviewed the latest head and the source-level blocker from my previous round is fixed now.

## What this PR adds
- User pain: SeaTunnel did not have a built-in transform for splitting long text into smaller chunks before downstream embedding, indexing, or RAG-style flows.
- Fix approach: add a `TextChunk` flat-map transform with configurable text field, output chunk field, chunk index field, chunk size, overlap, separator priority, tests, E2E coverage, and EN/ZH docs.
- One-line summary: this adds a first-class “one long text row -> many chunk rows” transform.

## Runtime path checked
```text
Job config
  -> TextChunkTransformFactory.optionRule()
      -> validate text_field / chunk_size / overlap_size / output field names
  -> TextChunkTransformFactory.createTransform()
      -> new TextChunkMultiCatalogTransform(catalogTables, options)

Schema production
  -> TextChunkMultiCatalogTransform.buildTransform()
      -> new TextChunkTransform(config, inputCatalogTable)
          -> locate text_field in the input row type
          -> transformTableSchema()
              -> copy original columns
              -> add or replace output_field
              -> add or replace chunk_index_field
              -> extend normal primary keys and unique keys with chunk_index_field

Row processing
  -> TextChunkTransform.flatMap(inputRow)
      -> transformRow(inputRow)
          -> null / empty text: emit no rows
          -> TextChunker.split(...)
          -> copy source fields for each chunk
          -> write chunk and chunk_index
          -> preserve tableId / rowKind / row options

Downstream sink
  -> receives schema with chunk + chunk_index
  -> key-aware sinks no longer see multiple chunk rows with the same original key only
```

## Review result
- The previous blocker is resolved. The output schema now extends normal primary keys and unique keys with `chunk_index_field`, and `TextChunkTransformTest` covers that behavior.
- +1 to @nzw921rx's earlier point about placing validation under the factory responsibility. The latest code has `validate(TextChunkTransformConfig config)` in `TextChunkTransformFactory`, so that point is addressed as well.
- I do not see a new source-side blocker in the latest head. The transform is isolated to the new plugin and does not change existing transform APIs, row serialization, checkpoint state, or connector protocols.
- Test stability looks good: the added UT/E2E cases do not rely on sleeps, wall-clock timing, external services, random ports, or order-sensitive data sources. The E2E jobs use deterministic FakeSource and Assert paths.
- EN/ZH docs are aligned on option names, defaults, behavior, and examples.

I did not run local tests; this review is based on the full latest diff, local call-chain tracing, the added tests, docs comparison, and the current GitHub check state.

## Merge conclusion
### Conclusion: can merge after the remaining gate clears

1. Blocking items
- No remaining source-side blocker from my side. The previous primary-key / unique-key semantics issue is fixed.
- The remaining gate is CI / merge status: `Build` is still in progress for the latest head, so I am not approving ahead of the required checks.

2. Non-blocking suggestions
- None from this round.

Overall, this now looks like a complete and maintainable Transform-V2 addition. The branch is not behind `dev` (`ahead_by=2`, `behind_by=0`), so there is no upstream-sync action to ask from the contributor right now. Once the current Build finishes green, this should be ready for approval / merge consideration.

Note: my earlier `changes requested` entry may still remain visible in GitHub's review summary until a write-capable maintainer or the final review gate updates it. From Daniel's technical review side, the source blocker is cleared and the remaining item is the current CI / merge gate.

DanielLeens:
Rechecked the same unchanged head `a3fc770b6be1` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
Job config
  -> TextChunkTransformFactory.optionRule()
      -> validate text_field / chunk_size / overlap_size / output field names
  -> TextChunkTransformFactory.createTransform()
      -> new TextChunkMultiCatalogTransform(catalogTables, options)

Schema production
  -> TextChunkMultiCatalogTransform.buildTransform()
      -> new TextChunkTransform(config, inputCatalogTable)
          -> locate text_field in the input row type
          -> transformTableSchema()
              -> copy original columns
              -> add or replace output_field
              -> add or replace chunk_index_field
              -> extend normal primary keys and unique keys with chunk_index_field

Row processing
  -> TextChunkTransform.flatMap(inputRow)
      -> transformRow(inputRow)
          -> null / empty text: emit no rows
          -> TextChunker.split(...)
          -> copy source fields for each chunk
          -> write chunk and chunk_index
          -> preserve tableId / rowKind / row options

Downstream sink
  -> receives schema with chunk + chunk_index
  -> key-aware sinks no longer see multiple chunk rows with the same original key only
```

## Merge conclusion

1. Blocking items

2. Current gate
- The current CI signal is now green: https://github.com/apache/seatunnel/actions/runs/29234523594/job/86766021047

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `a3fc770b6be1` from scratch.

# What this PR fixes
- User pain: RAG/embedding pipelines (#9717) have no built-in way to split a long text column into overlap-aware chunks, forcing users to pre-process documents outside SeaTunnel before vectorization.
- Fix approach: Adds a purely additive `TextChunk` transform — a dependency-free `TextChunker` (recursive separator-priority split, greedy merge to `chunk_size`, whole-piece overlap, code-point-safe boundaries) wired through `TextChunkTransformFactory`/`TextChunkTransform`/`TextChunkMultiCatalogTransform` as a 1→N FlatMap that appends `chunk` and `chunk_index` columns, plus en/zh docs, unit tests, and two E2E configs.
- One-line summary: Design and SPI wiring look sound and the factory-count/multi-table coverage is in place, but the unbounded row amplification when `overlap_size` approaches `chunk_size` is only documented rather than bounded, E2E only exercises the trivial `separators = []`, `overlap_size = 0` fixed-window path, null/empty-text row dropping is undocumented as data-loss semantics, and the docs' "measured in characters" wording conflicts with the code-point-aligned boundary behavior for non-BMP text.

# Runtime chain I rechecked
```text
FactoryUtil.discoverFactories(TableTransformFactory.class)  TransformSpecificationCheckTest.java:43 (asserts 23 factories, was 22)
  -> TextChunkTransformFactory.createTransform(TableTransformFactoryContext)  TextChunkTransformFactory.java
    -> new TextChunkMultiCatalogTransform(...)  TextChunkMultiCatalogTransform.java (per-catalog dispatch for the multi-table e2e config)
      -> new TextChunkTransform(config, catalogTable)  TextChunkTransform.java (validates text_field exists, 0 <= overlap_size < chunk_size, chunk_size > 0)
        -> TextChunkTransform.transformTableSchema()  TextChunkTransform.java (appends/reuses output_field STRING + chunk_index_field INT; extends PK and unique keys with chunk_index_field)
        -> TextChunkTransform.flatMap(SeaTunnelRow)  TextChunkTransform.java (null/empty text -> emits zero rows)
          -> TextChunker.chunk(text, chunkSize, overlapSize, separators)  TextChunker.java
            -> recursive split by separators[i] priority, greedy merge of pieces up to chunkSize, whole-trailing-piece overlap carry; separators empty -> fixed-size code-point-aligned windows (the only path covered by text_chunk_transform.conf: chunk_size=3, overlap_size=0, separators=[])
          -> one SeaTunnelRow per chunk: original fields copied + chunk text + 0-based index
```

# Findings
**Issue 1: Unbounded output/memory amplification: overlap_size close to chunk_size turns one input row into ~one row per character (resource-exhaustion DoS), and the plugin only documents the hazard instead of bounding it**
- Location: `docs/en/transforms/text-chunk.md:52`
- Why it matters: - Root cause: the option contract validates only `0 <= overlap_size < chunk_size`, so a config like `chunk_size = 1000, overlap_size = 999` is accepted, and the docs themselves state this yields roughly one chunk per character. With `separators = []` fixed-window mode, a single large text value (e.g. a multi-MB document ingested from an untrusted upstream) is expanded into `~len/(chunk_size-overlap_size)` rows.
- Impact: this is a classic input-driven amplification vector — a misconfigured or adversarially-fed job can multiply row count and memory by orders of magnitude, and if the FlatMap implementation materializes the full chunk list per input row (typical for Transform-V2), a Zeta/Flink/Spark worker can be OOM-killed by one poison row, taking down co-located jobs on the shared Zeta cluster.
- Fix: add a runtime guard in `TextChunkTransform`/`TextChunker` — e.g. reject or WARN when `overlap_size > chunk_size * 0.9` (or expose a configurable `max_chunks_per_row` cap and fail the row with a `TransformException` when exceeded), and prefer emitting chunks via an iterator/lazy sequence rather than building the entire `List<String>` in memory. A doc warning alone does not protect the cluster.
- Evidence: `docs/en/transforms/text-chunk.md:51  > `overlap_size` close to `chunk_size` multiplies the output row count (and memory) dramatically —`; `docs/en/transforms/text-chunk.md:52  > e.g. `chunk_size = 1000` with `overlap_size = 999` produces about one chunk per character. Keep the`
- Suggested fix: This note documents a real resource-exhaustion vector but nothing in the plugin bounds it: `overlap_size = chunk_size - 1` passes validation and expands one row into ~one row per character. On a shared Zeta cluster a single large text value can then OOM the worker. Please add a runtime guard (e.g. a `max_chunks_per_row` cap, or reject/WARN when `chunk_size - overlap_size` is tiny) and emit chunks lazily instead of materializing the whole list per input row — a doc warning alone doesn't protect the engine.
- Severity: Medium

**Issue 2: `chunk_size` semantics are ambiguous for non-BMP text: "measured in characters" vs. code-point-aligned boundaries**
- Location: `docs/en/transforms/text-chunk.md:38`
- Why it matters: - Root cause: the option doc says chunk length is "measured in characters" (i.e. Java `char`/UTF-16 code units), while the PR description and overlap section state boundaries are aligned to whole Unicode code points so surrogate pairs are never split — for emoji-heavy text these two statements diverge (a chunk may end up shorter than `chunk_size` chars, and a "character" count double-counts astral-plane symbols).
- Impact: users sizing chunks against embedding-model token/character limits (the primary RAG use case) can get chunks that don't match their expectation, and the ZH/EN docs must stay consistent with the actual `TextChunker` unit (`String.length()` code units vs `codePointCount`).
- Fix: state explicitly in both `docs/en` and `docs/zh` that `chunk_size`/`overlap_size` are counted in UTF-16 code units, that a chunk boundary is moved back to avoid splitting a surrogate pair, and therefore a chunk may be up to 1 code unit shorter than `chunk_size` for non-BMP text.
- Suggested fix: "Measured in characters" is ambiguous once code-point alignment kicks in: for surrogate pairs (emoji), a UTF-16 code-unit budget and a "character" count diverge. Please state explicitly that the unit is UTF-16 code units, that boundaries move back to avoid splitting a surrogate pair, and that a chunk can therefore be up to 1 unit shorter than `chunk_size` for non-BMP text (mirror this in the zh doc too).
- Severity: Low

**Issue 3: E2E only exercises the trivial fixed-window path (separators = [], overlap_size = 0); the separator-priority and overlap code paths that carry the documented complexity are untested end-to-end**
- Location: `seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-1/src/test/resources/text_chunk_transform.conf:46`
- Why it matters: - Root cause: `text_chunk_transform.conf` sets `separators = []`, `overlap_size = 0`, `chunk_size = 3` on the input `"abcabcabc"`, so the E2E run never touches the recursive separator splitting, greedy merge, whole-piece overlap carry, or surrogate-pair alignment described in the docs — exactly the logic most likely to regress.
- Impact: a regression in the separator/overlap path (e.g. chunk exceeding `chunk_size` when overlap counts toward the budget, or a split surrogate pair corrupting text sent to embedding sinks) would pass CI, since only unit tests cover it and E2E asserts nothing about that behavior across the real engine serialization boundary.
- Fix: add one more E2E case (or extend the multi-table conf) with non-empty `separators`, `overlap_size > 0`, and a text containing multi-byte code points, asserting exact chunk values via the Assert sink as done here.
- Evidence: `seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-1/src/test/resources/text_chunk_transform.conf:45      overlap_size = 0`; `seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-1/src/test/resources/text_chunk_transform.conf:46      separators = []`
- Suggested fix: With `separators = []` and `overlap_size = 0` this E2E only covers the fixed-window fallback — the separator-priority splitting and whole-piece overlap carry (the most fragile logic per the docs) never run through a real engine. Please add a case with non-empty `separators`, `overlap_size > 0`, and multi-byte code points, asserting exact chunk values with the Assert sink.
- Severity: Low

**Issue 4: Rows with null/empty text are silently dropped, which is undocumented data loss semantics that users cannot opt out of**
- Location: `docs/en/transforms/text-chunk.md:68`
- Why it matters: - Root cause: per the documented behavior, an input row whose `text_field` is `null` or empty produces zero output rows, so the entire source record (including all its other fields) vanishes from the pipeline.
- Impact: in RAG ingestion pipelines this silently loses documents with missing bodies; downstream row-count reconciliation between source and sink will not match, and there is no metric, log, or dead-letter path mentioned for the dropped rows. Users migrating from a passthrough transform will not expect a filter side-effect from a chunker.
- Fix: either emit a single passthrough row with `chunk = null` / `chunk_index = 0` for empty inputs, or add a boolean option (e.g. `skip_empty_text`, default true) so the drop is explicit and configurable, and log at DEBUG when a row is skipped.
- Suggested fix: Dropping the whole input row when `text_field` is null/empty is a silent filter side-effect — source/sink row counts will diverge and there's no way to opt out. Consider emitting a passthrough row (`chunk = null`, `chunk_index = 0`) or gating the drop behind an explicit option like `skip_empty_text = true`, plus a DEBUG log when a row is skipped.
- Severity: Low

# Review conclusion

- Issue 1: Unbounded output/memory amplification: overlap_size close to chunk_size turns one input row into ~one row per character (resource-exhaustion DoS), and the plugin only documents the hazard instead of bounding it
- Issue 2: `chunk_size` semantics are ambiguous for non-BMP text: "measured in characters" vs. code-point-aligned boundaries
- Issue 3: E2E only exercises the trivial fixed-window path (separators = [], overlap_size = 0); the separator-priority and overlap code paths that carry the documented complexity are untested end-to-end
- Issue 4: Rows with null/empty text are silently dropped, which is undocumented data loss semantics that users cannot opt out of

DanielLeens (commented):
Thanks for the update. I re-reviewed the full current diff and the transform schema/key path. The earlier validation-location concern is resolved, but two blockers remain.

## Blocking findings

**Issue 1: output fields can overwrite existing primary/unique-key columns**

- Location: `TextChunkTransformFactory.java:112-123` and `TextChunkTransform.java:110-113,155-161,186-195`
- Output names are checked only against each other. If `chunk_index_field=id` and `id` is the document primary key, `addOrReplace` overwrites the source id with `0,1,...`, while key extension skips adding the chunk index because that field already existed.
- Risk: chunks from different documents can collide in key-aware sinks. Please reject output/index collisions with existing key columns or preserve the original key under a distinct field, and add a key-collision regression test.

**Issue 2: UTF-16 surrogate pairs can exceed the documented chunk size**

- Location: `TextChunker.java:146-166`
- With `chunk_size=1`, a supplementary Unicode character is emitted as two UTF-16 code units. That contradicts the configured maximum unless size is explicitly defined in code points/graphemes.
- Please define and enforce the size unit consistently, with a supplementary-character test.

Both issues are blocking. The current `Build` check is green, but CI does not clear these source-contract issues. Happy to re-review an updated head.

SEZ9 (approved):
Thanks for working on this. I re-reviewed the latest head `939c76e91f87` from scratch.

# Review conclusion

DanielLeens (commented):

1. **Blocker 1** — `TextChunkTransform` resolves `textFieldIndex` only during construction. The transform framework refreshes catalog/schema state after evolution events, but this derived index is never recomputed. Adding, dropping, or reordering fields before the text column can chunk the wrong field or fail. Please refresh and validate the index on schema changes and add regressions.

The current Build is failing; CI remains a separate gate and does not change the source verdict above. The current failed lanes include `Run / unit-test (11, ubuntu-latest)`. Please retrigger or inspect those exact job logs after the source update.

DanielLeens (commented):

1. **Blocker 1** — `TextChunkTransform` resolves `textFieldIndex` only during construction. The transform framework refreshes catalog/schema state after evolution events, but this derived index is never recomputed. Adding, dropping, or reordering fields before the text column can chunk the wrong field or fail. Please refresh and validate the index on schema changes and add regressions.

The current Build is failing; CI remains a separate gate and does not change the source verdict above. The current failed lanes include `Run / unit-test (11, ubuntu-latest)`. Please retrigger or inspect those exact job logs after the source update.

nzw921rx (approved):
+1 good job🚀

DanielLeens (commented):

The previous source concern is resolved on `2e6411fb8fe3`: schema events now refresh the cached text-field index and produced schema.

Queue note: this head is `diverged` against `dev` (ahead_by=5, behind_by=4). I am recording that fact separately; it is not by itself a request to sync unless the CI guidance above explicitly says so.

davidzollo (approved):
+1
Good job

PR #11416, opened 2026-07-13
[Feature][Connector-V2] Support JDBC sink table_options for Oracle

det101:
## Purpose
Extend JDBC sink `table_options` (SaveMode auto-create) to Oracle, as a follow-up to #11047 / #11101 / #11388.

## Changes
- Add Oracle curated key whitelist: `tablespace`, `pctfree`
- Wire `OracleDialect.validateTableOptions` and apply options in `OracleCreateTableSqlBuilder` DDL (`PCTFREE` / `TABLESPACE`)
- OceanBase Oracle mode inherits the same path via `OracleDialect` + `OceanBaseOracleCreateTableSqlBuilder extends OracleCreateTableSqlBuilder` (no separate OB Oracle implementation)
- Docs (en/zh): support matrix + dialect caveats + Oracle example
- Unit tests: dialect / DDL / OptionRule coverage (including OceanBase Oracle pass & unknown-key reject)

## Non-goals
- Nested `STORAGE (...)`, LOB, or partition clauses
- `ALTER TABLE` on existing tables (`table_options` only applies when SaveMode creates the table)
- Value-level validation beyond key whitelist (same as MySQL delivery)
- PostgreSQL (separate PR)

## Test plan
- [x] `OracleDialectTableOptionsTest`
- [x] `OracleCreateTableSqlBuilderTest#testBuildCreateTableSqlWithTableOptions`
- [x] `JdbcTableOptionsConditionExtensionTest` (Oracle / OceanBase Oracle)
- [x] `OceanBaseOracleCreateTableSqlBuilderTest` (regression)
- [ ] CI on this PR

## Related
- Issue: #11047
- MySQL: #11101
- TiDB / OceanBase MySQL: #11388
- StarRocks: #11242

Made with [Cursor](https://cursor.com)

PR #11388, opened 2026-07-09
[Feature][Connector-V2] Extend JDBC sink table_options to TiDB and OceanBase MySQL

det101:
## Purpose
Extend JDBC sink `table_options` (SaveMode auto-create) from MySQL to TiDB and OceanBase MySQL mode, as a follow-up to #11101 / #11047.

## Changes
- Extract shared MySQL-family key whitelist validation into `MysqlFamilyTableOptions` (`engine` / `charset` / `collate`)
- Wire `OceanBaseMysqlDialect.validateTableOptions` and apply options in `OceanBaseMysqlCreateTableSqlBuilder` DDL
- TiDB reuses `MysqlDialect` + `MySqlCatalog` DDL via `jdbc:mysql://` (no new dialect)
- Docs (en/zh): support matrix + dialect caveats (TiDB ignores `engine`; OceanBase Oracle mode fail-fast; key-only validation at submit)
- Unit tests: dialect/DDL/OptionRule coverage, including OceanBase Oracle negative case

## Non-goals
- OceanBase Oracle-compatible mode remains unsupported (default dialect fail-fast)
- No value-level charset/collation validation at submit (same as MySQL PR1)
- PostgreSQL and other JDBC dialects deferred to later PRs

## Test plan
- [x] `JdbcTableOptionsConditionExtensionTest` (TiDB / OceanBase MySQL pass; OceanBase Oracle reject)
- [x] `OceanBaseMysqlDialectTest` / `OceanBaseMysqlCreateTableSqlBuilderTest` / `TiDBTableOptionsTest` / `MysqlDialectTest`
- [ ] CI on this PR

## Related
- Issue: #11047
- MySQL: #11101
- StarRocks: #11242

Made with [Cursor](https://cursor.com)

PR #11242, opened 2026-07-01
[Feature][Connector-V2] PR2：Support table_options for StarRocks SaveMode auto-create

det101:
### Purpose of this pull request

Follow-up to #11101 (PR1) for #11047: extend the shared sink `table_options` contract to **StarRocks SaveMode auto-create**.

This PR lets users pass StarRocks table properties (for example `replication_num`, `storage_format`) via `table_options` when `schema_save_mode` triggers DDL. Options are merged into the generated `CREATE TABLE ... PROPERTIES (...)` clause.

**Scope (StarRocks only; builds on merged PR1):**

- Reuse `SinkConnectorCommonOptions.TABLE_OPTIONS` (already on `dev` from #11101)
- Early validation via `Conditions.extension` in `StarRocksSinkFactory.optionRule()` (same pattern as JDBC PR1)
- Merge properties with `SqlTableClauseMerger` in `connector-common`
- Pass `table_options` through `StarRocksCatalog` constructor (not `CatalogTable.options`) to avoid upstream metadata pollution in CDC / multi-table jobs
- Reject non-empty `table_options` when `save_mode_create_template` differs from the built-in default (custom templates must embed properties in the template itself)

**Out of scope:** JDBC changes (already in #11101), new `seatunnel-api` types, E2E in this PR.

### Does this PR introduce _any_ user-facing change?

Yes.

- New optional StarRocks sink config: `table_options` (map), using the common option introduced in #11101.
- Applied only when SaveMode auto-creates tables (`CREATE_SCHEMA_WHEN_NOT_EXIST`, `RECREATE_SCHEMA`, etc.).
- Does **not** affect Stream Load at runtime and does **not** run `ALTER TABLE` on existing tables.
- **Not compatible** with a **custom** `save_mode_create_template` (validation fails at `--check` / job submission); use template `PROPERTIES` instead.
- With the **default** template, duplicate keys in `PROPERTIES` are overridden by `table_options`.
- No StarRocks property allowlist; invalid keys fail when StarRocks executes `CREATE TABLE`.
- Documentation updated: `docs/en/connectors/sink/StarRocks.md`, `docs/zh/connectors/sink/StarRocks.md`.

### How was this patch tested?

**Unit tests added/updated:**

- `SqlTableClauseMergerTest` — merge / override / append `PROPERTIES`
- `StarRocksSaveModeUtilTableOptionsTest` — validation + SQL merge
- `StarRocksTableOptionsConditionExtensionTest` — `optionRule` / `--check` path
[...]

PR #11101, opened 2026-06-15
[Feature][Connector-V2] PR1: Pass sink table-options into auto-created MySQL target tables

det101:
### Purpose of this pull request

Add `table-options` for JDBC sink auto-create flow (MySQL first delivery for #11047).

- Introduce shared `table-options` config in `SinkConnectorCommonOptions`
- Propagate options into `CatalogTable` in `JdbcSinkFactory`
- Validate per dialect; MySQL/TiDB allow `engine`, `charset`, `collate`
- Fail fast for unsupported JDBC dialects or unknown keys

**Status** — MySQL JDBC only; StarRocks/Paimon out of scope for this draft.

### Does this PR introduce _any_ user-facing change?

Yes. New optional sink config `table-options` (map). Applied only when `schema_save_mode` triggers DDL (e.g. `CREATE_SCHEMA_WHEN_NOT_EXIST`). Does not alter existing tables or runtime INSERT/UPSERT.

### How was this patch tested?

- Unit: `MysqlDialectTest`, `MysqlCreateTableSqlBuilderTest`
- E2E: `JdbcMysqlTableOptionsIT`

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md)
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config)

PR #11417, opened 2026-07-13
[Feature][Connector-V2] Support JDBC sink table_options for PostgreSQL

det101:
## Purpose
Extend JDBC sink `table_options` (SaveMode auto-create) to PostgreSQL, as a follow-up to #11047 / #11101 / #11388.

## Changes
- Add PostgreSQL curated key whitelist: `tablespace`, `fillfactor`
- Wire `PostgresDialect.validateTableOptions` and apply options in `PostgresCreateTableSqlBuilder` DDL (`WITH (fillfactor=...)` / `TABLESPACE "..."`)
- OpenGauss and HighGo inherit the same path via Postgres catalog/dialect
- Docs (en/zh): support matrix + dialect caveats + PostgreSQL example
- Unit tests: dialect / DDL / OptionRule coverage

## Non-goals
- Arbitrary `WITH (...)` parameters, `USING`, partition / `INHERITS`
- `ALTER TABLE` on existing tables (`table_options` only applies when SaveMode creates the table)
- Value-level validation beyond key whitelist (same as MySQL delivery)
- Oracle (separate PR: #11416)

## Test plan
- [x] `PostgresDialectTableOptionsTest`
- [x] `PostgresCreateTableSqlBuilderTest#testBuildCreateTableSqlWithTableOptions`
- [x] `JdbcTableOptionsConditionExtensionTest` (PostgreSQL pass / unknown-key reject)
- [x] `PostgresDialectTest` (regression)
- [ ] CI on this PR

## Related
- Issue: #11047
- MySQL: #11101
- Oracle: #11416
- StarRocks: #11242


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current diff locally from the JDBC sink auto-create path, including the option-rule validation path and the PostgreSQL create-table SQL builder.

What this PR fixes
- User pain: PostgreSQL users currently cannot express sink-side create-table storage options such as tablespace / fillfactor through SeaTunnel's JDBC sink auto-create path.
- Fix approach: whitelist `tablespace` / `fillfactor` in the dialect validation path and append them in `PostgresCreateTableSqlBuilder`.
- One-line summary: the feature direction is good, but the current implementation still leaves two blocker-level correctness holes.

Runtime path I checked
```text
JDBC sink config
  -> JdbcSinkFactory.optionRule()
      -> JdbcTableOptionsConditionExtension
          -> JdbcTableOptionsValidator.validate(...)
              -> PostgresDialect.validateTableOptions(...)
  -> save-mode auto create
      -> PostgresCreateTableSqlBuilder.build(...)
          -> append WITH (fillfactor=...) / TABLESPACE ... into CREATE TABLE
```

Blocking items
1. `PostgresCreateTableSqlBuilder` currently sends the tablespace value through `CatalogUtils.quoteIdentifier(tablespace, fieldIde, "\"")`. That rewrites the storage-object name with the column `fieldIde` policy even though `fieldIde` is meant for table/column identifiers, not tablespace names. For example, `pg_default` becomes `"PG_DEFAULT"` when `fieldIde=UPPERCASE`, which points at a different identifier and can break the generated DDL on a supported config combination. The current tests only cover the default/original case, so this regression is not exercised.
2. `PostgresDialect.validateTableOptions(...)` only whitelists keys, but `PostgresCreateTableSqlBuilder` injects `fillfactor` as raw text into `WITH (fillfactor=<value>)`. A malformed value such as `fillfactor=abc` still passes the new early-validation path and only fails when PostgreSQL executes the DDL. Since the PR is explicitly adding dialect validation, the value contract should be enforced there instead of surfacing later as runtime DDL failure.

Conclusion
Please fix these two items before merge. The safest path would be:
- keep `tablespace` out of `fieldIde` rewriting and treat it as a storage-object name, not a column identifier;
- validate `fillfactor` as a PostgreSQL-valid integer range during option validation, then let the builder consume only validated values.

Happy to re-review once the updated head is pushed.

DanielLeens (changes requested):
Thanks for working on this! I went through the full diff and traced the full path as:

`JdbcSinkFactory.optionRule()`
`-> JdbcTableOptionsConditionExtension.evaluate()`
`-> JdbcTableOptionsValidator.validate()`
`-> PostgresDialect.validateTableOptions()`
`-> JdbcSinkFactory.createSink()`
`-> PostgresCreateTableSqlBuilder.build()`

The overall shape makes sense, but I found one blocker before merge:

1. Supported keys with blank values currently pass submission and then get silently dropped from the generated DDL.
   - `JdbcTableOptionsValidator.java:36-48` delegates validation to the dialect.
   - `PostgresDialect.java:479-493` only checks whether the keys are in the supported whitelist.
   - `PostgresCreateTableSqlBuilder.java:114-120` then guards both clauses with `StringUtils.isNotBlank(...)`.

So a config like `table_options = { fillfactor = " " }` or `table_options = { tablespace = " " }` does not fail fast and does not fail at DDL execution either. It just creates the table without the requested clause, which means the job succeeds with the database default instead of telling the user the config is invalid.

I would strongly recommend rejecting blank values for the supported keys in the dialect validator (or in the shared JDBC table_options validator) and adding a regression test for that case.

The new tests look deterministic, but they only cover supported keys / unsupported keys and the happy-path SQL rendering. They do not cover the blank-value path above yet.

Merge conclusion: not ready to merge yet. Build is green on the current head, so the remaining blocker here is source-side validation correctness.

DanielLeens:
Rechecked the same unchanged head `26e772ed9609` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
JDBC sink config
  -> JdbcSinkFactory.optionRule()
      -> JdbcTableOptionsConditionExtension
          -> JdbcTableOptionsValidator.validate(...)
              -> PostgresDialect.validateTableOptions(...)
  -> save-mode auto create
      -> PostgresCreateTableSqlBuilder.build(...)
          -> append WITH (fillfactor=...) / TABLESPACE ... into CREATE TABLE
```

## Merge conclusion

1. Blocking items

2. Current gate update
- The CI / merge-gate fact I referenced earlier has changed. The current status is `PENDING gate: Build`: https://github.com/apache/seatunnel/runs/87012019791

DanielLeens:
Rechecked the latest head again from the PostgreSQL `table_options` validation/build path.

The earlier Daniel source-side blockers are closed for me on this revision:
- blank values are now rejected instead of being silently dropped;
- `fillfactor` is now validated before SQL generation;
- `tablespace` is no longer run through the column `fieldIde` rewriting path.

So my current Daniel-side conclusion on this head is no longer "code blocker still open". The remaining gate is the current GitHub `Build` result.

From the visible failing lanes on the current fork workflow, I do not see a clear source-level link from this PostgreSQL DDL/validation diff to the red top-level `Build` yet. The lowest-cost next step is to rerun `Build` on the current head first. If the refreshed run still fails, please paste the new run link and I can narrow it down from there.

davidzollo (approved):
LGTM

Issue #11047, opened 2026-06-10
[Feature][Connector-V2][CDC] Pass sink-specific table options into auto-created target tables


davidzollo:
## Background
When CDC jobs auto-create target tables, users often need to attach sink-specific table properties at creation time. Examples include storage format settings, distribution properties, table comments, or other sink-managed options.

Without an explicit configuration path, users must patch created tables manually after the job starts or rely on sink-specific workarounds outside the CDC job definition.

## Scope
Allow CDC jobs that auto-create target tables to pass explicit sink-specific table options into the created target table definition.

This issue is about **option propagation during auto-create**. It does **not** cover deriving indexes/comments from upstream schema automatically or changing existing sink defaults.

## First delivery boundary
The first delivery should prove end-to-end propagation on sinks that already have explicit table-creation paths.

Recommended scope:
- **JDBC sink** as the mandatory first target, because the repository already has concrete create-table flow in `AbstractJdbcCatalog` / `JdbcSaveModeHandler`
- **one additional catalog-backed sink** with explicit create-table support, preferably `StarRocks` or `Paimon`, so the feature is not JDBC-only in design

Do not block the first delivery on covering every auto-create sink.

### 1. Define a stable config surface
Use an explicit structured config surface for target table options. A `map<string,string>`-style representation is the cleanest first delivery because it aligns with how sink-specific key/value options are usually modeled.

Illustrative target shape:

```hocon
table-options {
  key1 = value1
  key2 = value2
}
```

Exact naming can still be refined, but the contract should be structured rather than hidden in free-form SQL fragments.

### 2. Carry options through `CatalogTable`
The option map should flow through the existing table abstraction rather than being injected only at the final SQL string builder stage.

Primary touch points are expected to include:
- `CatalogTable` option propagation
- multi-table sink auto-create path
- sink-specific catalog / DDL builder entry points

### 3. Apply options in concrete sink create-table paths
Expected first-delivery touch points likely include:
- JDBC catalog create-table flow (`AbstractJdbcCatalog`, `JdbcSaveModeHandler`, dialect-specific create-table builders)
- one catalog-backed sink create-table flow such as `StarRocksCatalog#createTable(...)` or `PaimonCatalog#createTable(...)`

### 4. Validate support explicitly
If a sink supports auto-create but does not support arbitrary table options yet, the behavior must be explicit in the first delivery:
- fail fast with a clear message for that sink
- do not silently drop configured table options

[...]

PR #11418, opened 2026-07-13
[Feature][Connector-V2] Support table_options for Paimon sink SaveMode auto-create

det101:
## Purpose
Add Paimon sink `table_options` for SaveMode auto-create, as a follow-up to #11047 / #11242.

## Changes
- Merge sink `table_options` into effective `paimon.table.write-props` in `PaimonSinkConfig` (write-props wins on key conflict)
- Open-key validation via OptionRule ConditionExtension: reject blank keys / null values at job submission
- Wire `TABLE_OPTIONS` into `PaimonSinkFactory` / `PaimonCatalogFactory`
- Docs (en/zh): merge semantics, precedence, auto-create-only scope
- Unit tests: merge priority, OptionRule, blank-key / null-value reject

## Non-goals
- Key allowlist (Paimon CoreOptions are open; invalid keys fail when Paimon creates the table)
- `ALTER` on existing tables (`table_options` only applies when SaveMode creates the table)
- Changing `SchemaUtil` create path (already consumes `getWriteProps()`)

## Test plan
- [x] `PaimonSinkConfigTableOptionsTest`
- [x] `PaimonTableOptionsConditionExtensionTest`
- [x] `PaimonOptionRuleTest` (declares `table_options`)
- [ ] CI on this PR

## Related
- Issue: #11047
- StarRocks: #11242
- JDBC MySQL: #11101
- Oracle: #11416
- PostgreSQL: #11417

### DIFF
--- a/docs/en/connectors/sink/Paimon.md
+++ b/docs/en/connectors/sink/Paimon.md
@@ -78,7 +78,7 @@ libfb303-xxx.jar
 | paimon.table.primary-keys    | String  | No       | -                            | Default comma-separated list of columns (primary key) that identify a row in tables.(Notice: The partition field needs to be included in the primary key fields) |
 | paimon.table.partition-keys  | String  | No       | -                            | Default comma-separated list of partition fields to use when creating tables.                                                                                    |
 | paimon.table.write-props     | Map     | No       | -                            | Properties passed through to paimon table initialization, [reference](https://paimon.apache.org/docs/master/maintenance/configurations/#coreoptions).            |
-| table_options                | Map     | No       | -                            | Sink-specific table options merged into write-props for SaveMode auto-create. See below.                                                                         |
+| table_options                | Map     | No       | -                            | Sink-specific table options for SaveMode auto-create only (not runtime write-props). See below.                                                                  |
 | paimon.hadoop.conf           | Map     | No       | -                            | Properties in hadoop conf                                                                                                                                        |
 | paimon.hadoop.conf-path      | String  | No       | -                            | The specified loading path for the 'core-site.xml', 'hdfs-site.xml', 'hive-site.xml' files                                                                       |
 | paimon.table.non-primary-key | Boolean | No       | false                        | Switch to create `table with PK` or `table without PK`. true : `table without PK`, false : `table with PK`                                                       |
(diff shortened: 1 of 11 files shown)


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current diff locally from the Paimon sink option path, the sink config construction path, and the schema auto-create path.

What this PR fixes
- User pain: users want to pass Paimon table creation options through SeaTunnel when the sink auto-creates the target table in SaveMode.
- Fix approach: accept `table_options` in `PaimonSinkFactory`, validate them, and merge them into the Paimon sink config so the schema builder can apply them during auto-create.
- One-line summary: the intent is good, but the current implementation broadens `table_options` beyond auto-create and changes runtime sink behavior on unchanged tables too.

Runtime path I checked
```text
Paimon sink config
  -> PaimonSinkFactory.optionRule()
      -> PaimonTableOptionsConditionExtension
          -> PaimonTableOptionsValidator.validate(...)
  -> new PaimonSinkConfig(readonlyConfig)
      -> merge table_options into writeProps unconditionally
      -> derive changelogProducer / changelogTmpPath from merged writeProps
  -> save-mode auto create path
      -> SchemaUtil.buildSchema(...)
          -> paiSchemaBuilder.options(writeProps)
```

Blocking item
1. `PaimonSinkConfig` now merges `table_options` into `writeProps` unconditionally, before the code even knows whether the sink is actually auto-creating a table. The same merged `writeProps` map is then reused to derive runtime settings like `changelogProducer` and `changelogTmpPath`. So a config knob that is supposed to be limited to SaveMode auto-create can now silently change runtime sink behavior on existing tables too. That is broader than the PR contract and can alter job semantics even when no create-table path is taken.

Conclusion
Please scope the new `table_options` support strictly to the auto-create/schema-build path instead of merging it into the general runtime `writeProps` at config-construction time. A safer direction would be to keep `write-props` as the runtime writer config, pass `table_options` separately into the schema/table-creation path, and only apply them when the sink is actually creating the table.

Once that boundary is tightened, I am happy to re-review the updated head.

DanielLeens (changes requested):
Thanks for working on this! I went through the full diff and traced the full path before commenting.

The intended auto-create path looks like this today:

`PaimonSinkFactory.optionRule / PaimonCatalogFactory.optionRule`
`-> PaimonTableOptionsConditionExtension.evaluate()`
`-> PaimonTableOptionsValidator.validate()`
`-> new PaimonSinkConfig(readonlyConfig)`
`-> PaimonCatalog.createTable(...)`
`-> SchemaUtil.toPaimonSchema(...)`

That part makes sense. The blocker is that the new config does not stay inside the auto-create path:

`PaimonSink.createWriter()`
`-> new PaimonSinkWriter(...)`
`-> paimonSinkConfig.getChangelogProducer() / getChangelogTmpPath()`

1. `table_options` is currently merged into `PaimonSinkConfig.writeProps` too early, so it also affects the normal writer path for an already existing table.
   - In `PaimonSinkConfig.java:70-93`, `table_options` is merged into the same `writeProps` map that later drives `changelogProducer` and `changelogTmpPath`.
   - In `SchemaUtil.java:83-89`, that merged map is correctly used for schema creation.
   - But in `PaimonSinkWriter.java:142-160`, the same merged config is still read on the runtime writer path, which means `table_options` can now trigger the changelog warning logic and change the `IOManager` temp path even when no table creation happens at all.

That does not match the new docs, which say `table_options` only applies when SaveMode auto-creates the target table and does not alter an existing table. I think we should keep `table_options` scoped to the schema-creation path only, or otherwise make sure the runtime-only writer settings are not derived from the merged map.

The new tests look deterministic to me, but they currently do not cover this existing-table writer path regression.

Merge conclusion: not ready to merge yet. Please fix the source-side scope leak first. Current CI is green, so this is a pure source-level blocker from my side.

DanielLeens (approved):
Rechecked the latest head from the Paimon sink config path, schema auto-create path, and runtime writer path.

The source-side blocker from my previous Daniel round is closed for me on this revision: `table_options` is no longer merged into the runtime `writeProps` map, and the schema-creation path now uses a separate merge only when building table options for SaveMode auto-create.

# Runtime path rechecked
```text
Paimon sink config
  -> new PaimonSinkConfig(readonlyConfig)
      -> writeProps stays runtime-only
      -> tableOptions stays separate

SaveMode auto-create path
  -> SchemaUtil.toPaimonSchema(...)
      -> mergeSchemaCreationOptions(tableOptions, writeProps)
      -> apply merged options only to schema/table creation

Normal writer path
  -> PaimonSinkWriter
      -> changelog producer / tmp path still read from runtime writeProps only
```

# Conclusion
### Conclusion: can merge

1. Blocking items

2. Non-blocking suggestion
- `PaimonTableOptionsConditionExtension.description()` still says `table_options` is merged into `paimon.table.write-props`, which no longer matches the actual implementation. It would be nice to refresh that wording in a follow-up so the validation message stays aligned with the real behavior.

The current `Build` signal is green on this head, so I am updating Daniel's conclusion accordingly.

DanielLeens (commented):
Thanks for the contribution! I re-reviewed the full current diff and the relevant Paimon table-option scoping runtime path.

The source-side implementation is clear on the latest head. The remaining `Build` failure is the unrelated engine `MetricsApiTest`, and latest `dev` does not yet contain a corresponding fix.

DanielLeens (commented):



DanielLeens (commented):



davidzollo (approved):
Re-reviewed the current head and merge result against the latest dev. The Paimon table_options change remains scoped to SaveMode schema creation, runtime write-props behavior is preserved, required Build is green, and there are no unresolved review threads. LGTM.

PR #11419, opened 2026-07-13
[Feature][Connector-V2] Support JDBC sink table_options for Dameng

det101:
## Purpose
Extend JDBC sink `table_options` (SaveMode auto-create) to Dameng, as a follow-up to #11047 / #11101 / #11388.

## Changes
- Add Dameng curated key whitelist: `tablespace`, `fillfactor`
- Wire `DmdbDialect.validateTableOptions` and apply options in `DamengCreateTableSqlBuilder` as Dameng `STORAGE (FILLFACTOR ..., ON "...")`
- Docs (en/zh): support matrix + dialect notes + Dameng example
- Unit tests: dialect / DDL / OptionRule coverage

## Non-goals
- Arbitrary nested `STORAGE` parameters (`INITIAL` / `NEXT` / …)
- Oracle-style `PCTFREE` / `TABLESPACE` clauses (Dameng uses `STORAGE`)
- `ALTER TABLE` on existing tables (`table_options` only applies when SaveMode creates the table)
- E2E (unit tests only for this slice)

## Test plan
- [x] `DmdbDialectTest` (supported / unsupported keys)
- [x] `DamengCreateTableSqlBuilderTest#testBuildCreateTableSqlWithTableOptions`
- [x] `JdbcTableOptionsConditionExtensionTest` (Dameng pass / unknown-key reject)
- [ ] CI on this PR

## Related
- Issue: #11047
- MySQL: #11101
- Oracle: #11416
- PostgreSQL: #11417
- Paimon: #11418


DanielLeens (changes requested):
Thanks for working on this. I reviewed the full current diff locally from the JDBC sink auto-create path, including the option-rule validation path and the Dameng create-table SQL builder.

What this PR fixes
- User pain: Dameng users currently cannot pass sink-side create-table storage options through SeaTunnel's JDBC auto-create path.
- Fix approach: whitelist `tablespace` / `fillfactor` for Dameng and append them into the generated STORAGE clause.
- One-line summary: the direction is reasonable, but the new path still has two blocker-level correctness gaps.

Runtime path I checked
```text
JDBC sink config
  -> JdbcSinkFactory.optionRule()
      -> JdbcTableOptionsConditionExtension
          -> JdbcTableOptionsValidator.validate(...)
              -> DmdbDialect.validateTableOptions(...)
  -> save-mode auto create
      -> DamengCreateTableSqlBuilder.build(...)
          -> append STORAGE (FILLFACTOR ..., ON ...)
```

Blocking items
1. `DamengCreateTableSqlBuilder` currently routes the tablespace value through `CatalogUtils.quoteIdentifier(tablespace, fieldIde, "\"")`. That means the storage-object name is rewritten by the column `fieldIde` policy even though `fieldIde` is meant for identifiers like columns/tables, not tablespaces. On a supported config such as `fieldIde=LOWERCASE` or `UPPERCASE`, the generated tablespace token can change unexpectedly and point at the wrong storage object. The new tests only cover the original/default path, so this break is still uncovered.
2. `DmdbDialect.validateTableOptions(...)` only validates keys, but `DamengCreateTableSqlBuilder` injects `fillfactor` as raw text into the generated STORAGE clause. A malformed value such as `fillfactor=abc` still passes the new early-validation path and only fails later when the DDL is executed. Since the PR is explicitly adding dialect-level validation, the value contract should be checked there before the builder emits SQL.

Conclusion
Please fix these two items before merge. The safest fix is to keep tablespace names out of `fieldIde` rewriting and to validate `fillfactor` as a Dameng-valid numeric value during option validation.

Happy to re-review the next head once it is updated.

DanielLeens (changes requested):
Thanks for working on this! I went through the full diff and traced the path as:

`JdbcTableOptionsConditionExtension.evaluate()`
`-> JdbcTableOptionsValidator.validate()`
`-> DmdbDialect.validateTableOptions()`
`-> JdbcSinkFactory.createSink()`
`-> DamengCreateTableSqlBuilder.build()`

I found one blocker before merge:

1. Supported keys with blank values currently pass submission and then get silently dropped from the generated `STORAGE (...)` clause.
   - `JdbcTableOptionsValidator.java:36-48` delegates validation to the dialect.
   - `DmdbDialect.java:405-420` only checks whether the keys are in the supported whitelist.
   - `DamengCreateTableSqlBuilder.java:101-111` only adds the `FILLFACTOR` / `ON "<tablespace>"` fragments when the value is `StringUtils.isNotBlank(...)`.

So a config like `table_options = { tablespace = " " }` passes submission but quietly falls back to the database default at DDL generation time. That means the job succeeds even though the requested storage option was invalid, which is a pretty confusing failure mode for users.

I’d recommend rejecting blank values for the supported Dameng keys during validation and adding a regression test for that path.

The current `Build` failure still points at the contributor fork’s `kafka-connector-it` lane, and I do not see evidence yet that this Dameng diff is the direct cause. But because the source-side validation gap above is still open, I’m not reducing this to a CI-only gate.

Merge conclusion: not ready to merge yet. Please fix the validation gap first, then we can recheck the remaining CI state if needed.

DanielLeens:
Rechecked the same unchanged head `c1be135337e8` after the CI / merge-gate fact changed.

## Runtime path rechecked
```text
JDBC sink config
  -> JdbcSinkFactory.optionRule()
      -> JdbcTableOptionsConditionExtension
          -> JdbcTableOptionsValidator.validate(...)
              -> DmdbDialect.validateTableOptions(...)
  -> save-mode auto create
      -> DamengCreateTableSqlBuilder.build(...)
          -> append STORAGE (FILLFACTOR ..., ON ...)
```

## Merge conclusion

1. Blocking items

2. Current gate update
- The CI / merge-gate fact I referenced earlier has changed. The current status is `CI is green on the current head`: https://github.com/apache/seatunnel/actions/runs/29304782920/job/86995812982

DanielLeens:
Rechecked the latest head again from the Dameng `table_options` validation/build path.

The earlier Daniel source-side blockers are closed for me on this revision:
- blank values are now rejected instead of being silently dropped;
- `fillfactor` is now validated before SQL generation;
- `tablespace` is no longer run through the column `fieldIde` rewriting path.

So my current Daniel-side conclusion on this head is no longer "code blocker still open". The remaining gate is the current GitHub `Build` state.

The visible cancellation on the current fork workflow still looks more like general CI/workflow churn than a Dameng-specific failure signal. So the lowest-cost next step here is to rerun `Build` on the current head first. If the refreshed run still fails, please paste the new run link and I can help narrow it down against this updated revision.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `c1be135337e8` from scratch.

# What this PR fixes
- User pain: Dameng users could not pass table-level storage options (`tablespace`, `fillfactor`) when the JDBC sink auto-creates the target table via SaveMode, unlike the MySQL/TiDB/OceanBase support added in #11101/#11388; any non-empty `table_options` failed validation for the Dameng dialect.
- Fix approach: Adds a curated key whitelist in `DmdbDialect.validateTableOptions` (with blank-value, `fillfactor` 0–100 range, and `tablespace` illegal-character checks), and has `DamengCreateTableSqlBuilder` read the two options from `catalogTable.getOptions()` to emit a Dameng `STORAGE (FILLFACTOR n, ON "tablespace")` clause, plus en/zh docs and unit tests for dialect, DDL builder, and option-rule paths.
- One-line summary: The feature works end-to-end and follows the established pattern from prior dialects, but the SQL builder trusts option values without re-validation (validation only happens on the factory/submission path, so catalog-driven or programmatic auto-create can inject unvalidated values into DDL), `fillfactor` forms like `+80`/`080` pass `Integer.parseInt` yet are emitted verbatim, and builder-level negative tests are missing.

# Runtime chain I rechecked
```text
Submission-time validation path:
  JdbcSinkFactory.createSink() -> optionRule() conditional validation -> JdbcDialect.validateTableOptions()
    -> DmdbDialect.validateTableOptions(Map)  DmdbDialect.java:410
      -> whitelist check against SUPPORTED_TABLE_OPTIONS (tablespace, fillfactor)  DmdbDialect.java:415-424
      -> blank-value check  DmdbDialect.java:429-435
      -> validateFillfactor(trimmed)  DmdbDialect.java:441 -> Integer.parseInt + [0,100] range  DmdbDialect.java:447-473
      -> validateTablespace(trimmed)  DmdbDialect.java:443 -> rejects '"', '\n', '\r', ';'  DmdbDialect.java:476-489

Runtime DDL generation path (SaveMode CREATE_SCHEMA_WHEN_NOT_EXIST):
  DefaultSaveModeHandler.handleSchemaSaveMode() -> Catalog.createTable()
    -> AbstractJdbcCatalog.createTable() -> DamengCatalog.getCreateTableSqls()
      -> new DamengCreateTableSqlBuilder(catalogTable, createIndex)  DamengCreateTableSqlBuilder.java:50
        -> reads catalogTable.getOptions().get("tablespace"/"fillfactor")  DamengCreateTableSqlBuilder.java:54-55  (no re-validation here — values are trusted)
      -> DamengCreateTableSqlBuilder.build(tablePath)  DamengCreateTableSqlBuilder.java:97
        -> appends 'FILLFACTOR ' + fillfactor.trim() and 'ON "' + tablespace.trim() + '"' into STORAGE (...)  DamengCreateTableSqlBuilder.java:100-114
        -> raw string concatenation into CREATE TABLE SQL executed against the target DB
```

# Findings
**Issue 1: DamengCreateTableSqlBuilder concatenates tablespace/fillfactor from CatalogTable options into DDL without re-validation, allowing unvalidated values (e.g., from source metadata or catalog-driven auto-create paths) to break or inject into CREATE TABLE**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/dm/DamengCreateTableSqlBuilder.java:102`
- Why it matters: - Root cause: the builder reads `tablespace`/`fillfactor` directly from `catalogTable.getOptions()` and concatenates them into the `STORAGE (...)` clause, relying solely on the comment that "Value is validated ... by DmdbDialect.validateTableOptions". That validation only runs on the user-configured `table_options` map at `JdbcSinkFactory` submission time.
- `CatalogTable.getOptions()` is a shared bag that can be populated from other sources: source catalogs copying table metadata, CDC multi-table auto-create flows that call `DamengCatalog.createTable` directly, or any code path that constructs the CatalogTable without going through the sink factory. In those paths a source-derived option literally named `tablespace` containing a `"` character, or a non-numeric `fillfactor`, is concatenated raw into the DDL — producing a broken `CREATE TABLE` at runtime (job fails during SaveMode) or, worse, arbitrary SQL fragments in the statement since `fillfactor` is emitted completely unquoted.
- Fix: make the builder defensive rather than trusting upstream validation — parse `fillfactor` with `Integer.parseInt` (and range-check) inside `build()`, and apply the same illegal-character check (or quote-escaping via doubling `"`) to `tablespace` before emitting `ON "..."`. Alternatively, extract the validation into a shared static helper on `DmdbDialect` and invoke it from the builder so the two code paths cannot diverge.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/dm/DamengCreateTableSqlBuilder.java:102  storageItems.add("FILLFACTOR " + fillfactor.trim());`; `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/dm/DamengCreateTableSqlBuilder.java:107  storageItems.add("ON \"" + tablespace.trim() + "\"");`; `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/dm/DamengCreateTableSqlBuilder.java:54  this.tablespace = catalogTable.getOptions().get(DamengCatalog.TABLE_OPTION_TABLESPACE);`
- Suggested fix: This relies on `DmdbDialect.validateTableOptions` having run, but that only happens for user-configured `table_options` at sink-factory submission. `catalogTable.getOptions()` can also be populated from source metadata or catalog-driven auto-create paths (e.g., CDC multi-table), in which case an unvalidated `fillfactor` (arbitrary string, unquoted here) or a `tablespace` containing `"` is concatenated raw into the DDL — breaking `CREATE TABLE` at runtime or injecting SQL fragments. Please make the builder defensive: `Integer.parseInt(fillfactor.trim())` with range check here, and re-apply the illegal-character check (or escape `"` by doubling) for `tablespace` before emitting `ON "..."`.
- Severity: High

**Issue 2: fillfactor values like "+80" or "080" pass Integer.parseInt validation but are emitted verbatim, producing DDL Dameng may reject**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbDialect.java:447`
- Why it matters: - Root cause: `validateFillfactor` uses `Integer.parseInt(value)`, which accepts a leading `+` sign and leading zeros (e.g., `+80`, `080`). The builder then emits the raw trimmed string (`"FILLFACTOR " + fillfactor.trim()`), so the DDL becomes `FILLFACTOR +80` / `FILLFACTOR 080` instead of the canonical integer literal.
- Impact: the config passes `--check` and job submission successfully, but the `CREATE TABLE` may fail at runtime on the Dameng server when SaveMode executes the DDL — exactly the happy-path-only gap the early validation was supposed to prevent, and it fails late on Zeta/Flink/Spark alike.
- Fix: normalize the emitted value by parsing it once and appending the parsed `int` (e.g., `storageItems.add("FILLFACTOR " + Integer.parseInt(fillfactor.trim()))`), which also doubles as the defensive re-validation suggested in the other finding.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/dm/DamengCreateTableSqlBuilder.java:102  storageItems.add("FILLFACTOR " + fillfactor.trim());`
- Suggested fix: `Integer.parseInt` accepts `"+80"` and `"080"`, so those values pass validation here, but the builder emits the raw string (`FILLFACTOR +80`), which Dameng may reject at CREATE TABLE time — a late runtime failure that this early validation was meant to prevent. Consider normalizing on emit: append the parsed integer rather than the raw trimmed string in `DamengCreateTableSqlBuilder`.
- Severity: Medium

**Issue 3: Missing negative tests at the builder level: no coverage for unvalidated/malformed option values reaching DamengCreateTableSqlBuilder, or for '+80'/'080' fillfactor forms**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/dm/DamengCreateTableSqlBuilderTest.java:197`
- Why it matters: - The new tests cover the happy path (`testBuildCreateTableSqlWithTableOptions` asserts the `STORAGE (FILLFACTOR 80, ON "MAIN")` output) and dialect-level whitelist rejection in `DmdbDialectTest`, but there is no builder-level test proving what happens when a `tablespace` containing `"` or a non-numeric `fillfactor` reaches `build()` — which is exactly the path CatalogTable options can take when they bypass `validateTableOptions`.
- There is also no test for boundary/edge fillfactor inputs that `Integer.parseInt` accepts but Dameng may reject (`"+80"`, `"080"`), nor for tablespace-only / fillfactor-only combinations if not covered by the truncated test file.
- Fix: add builder tests asserting the (post-fix) defensive behavior — a malformed value should throw `JdbcConnectorException` from `build()` rather than emit broken DDL — plus parameterized dialect tests for `"+80"`, `"080"`, boundary `0`/`100`, and `101`/`-1`.
- Suggested fix: This covers the happy path, but please also add builder-level negative tests: a CatalogTable whose options carry a `tablespace` with `"` or a non-numeric `fillfactor` (values that bypass `DmdbDialect.validateTableOptions` when the catalog is invoked directly). Once the builder re-validates defensively, these tests should assert `JdbcConnectorException` is thrown from `build()` instead of emitting broken DDL. Also worth adding `"+80"` / `"080"` and boundary `0` / `100` / `101` cases.
- Severity: Low

**Issue 4: validateTablespace blocks ", \n, \r, ; but not other control characters or backslash; the character blacklist is narrower than the quoting context needs**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbDialect.java:475`
- Why it matters: - Root cause: `validateTablespace` uses a small blacklist (`"`, `\n`, `\r`, `;`). Other control characters (`\t`, `\0`, `\u000B` etc.) and unusual characters still flow into the quoted identifier `ON "..."`.
- Impact: a tablespace name containing e.g. a NUL or tab passes submission-time validation but yields a malformed or confusing identifier in the DDL, failing only at runtime with an opaque Dameng error. Blacklists tend to under-block; the safer pattern for a quoted identifier is either escaping (double any `"`) or a whitelist.
- Fix: replace the blacklist with a whitelist check such as rejecting any `Character.isISOControl(c)` plus `"`; or validate against a conservative identifier pattern (e.g., `[A-Za-z0-9_$#]+`) matching what Dameng actually allows for tablespace names, and document it.
- Evidence: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbDialect.java:476  || value.indexOf('\n') >= 0`; `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbDialect.java:478  || value.indexOf(';') >= 0) {`
- Suggested fix: This blacklist misses other control characters (`\t`, `\0`, vertical tab, ...) that would still land inside `ON "..."` and only fail at runtime with an opaque Dameng error. A whitelist is safer here — e.g., reject any `Character.isISOControl(c)` in addition to `"`, or validate against the identifier pattern Dameng actually permits for tablespace names.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: DamengCreateTableSqlBuilder concatenates tablespace/fillfactor from CatalogTable options into DDL without re-validation, allowing unvalidated values (e.g., from source metadata or catalog-driven auto-create paths) to break or inject into CREATE TABLE

- Issue 2: fillfactor values like "+80" or "080" pass Integer.parseInt validation but are emitted verbatim, producing DDL Dameng may reject
- Issue 3: Missing negative tests at the builder level: no coverage for unvalidated/malformed option values reaching DamengCreateTableSqlBuilder, or for '+80'/'080' fillfactor forms
- Issue 4: validateTablespace blocks ", \n, \r, ; but not other control characters or backslash; the character blacklist is narrower than the quoting context needs

DanielLeens (commented):
Thanks for the contribution! I re-reviewed the full current diff and the relevant Dameng table-option validation and DDL emission runtime path.

I do not see a source-side blocker on the current head and `Build` is green. The remaining gate is repository review/maintainer approval.

DanielLeens (commented):

The previous source concern is resolved on `1725fffe257b`: the builder path now normalizes and validates `TABLESPACE`/`FILLFACTOR`, including programmatic construction.

Queue note: this head is `diverged` against `dev` (ahead_by=4, behind_by=2). I am recording that fact separately; it is not by itself a request to sync unless the CI guidance above explicitly says so.

DanielLeens (commented):
Thanks for the update. I rechecked the complete current-head diff after the latest base sync.

The Dameng `table_options` path is still closed end to end: option-rule validation and the DDL builder both normalize and validate the supported `TABLESPACE` / `FILLFACTOR` keys, reject blank or invalid values, and quote the generated clause consistently. The sync did not introduce unrelated current-diff files or deletions.

I do not see a source-level blocker, and the required Build is green. The remaining repository gate is one maintainer review. Daniel has read-only review permission, so this is a `COMMENTED` technical conclusion rather than an approval.

davidzollo (approved):
+1

PR #11424, opened 2026-07-13
[Docs] Add transform recipes for nested JSON and vector conversion

### LINKED ISSUE #11277: [Docs] Add transform recipes for nested JSON splitting and vector conversion
## Background

The transform docs describe individual transforms, but users often need scenario-level recipes that combine source parsing, transforms, and sink schema requirements.

Examples of recurring questions include:

- Read a deeply nested Kafka JSON message, split part of it into many rows, and write the result to MySQL.
- Read a MySQL JSON field and convert it into a Milvus `FLOAT_VECTOR` field.
- Read file content and use LLM-related processing to extract or chunk content before writing it downstream.

These questions are difficult to answer from individual transform option tables alone.

## Goal

Add practical transform recipes that show how to solve common nested JSON and vector conversion scenarios with SeaTunnel.

## Suggested scope

Please add or update English and Chinese transform docs. Possible locations:

- `docs/en/transforms/`
- `docs/zh/transforms/`
- a new transform recipes page if that fits the current docs structure

The recipes should cover:

1. Nested JSON extraction and field mapping from Kafka or file source input.
2. Whether SeaTunnel supports one input row producing multiple output rows through built-in transforms today.
3. How to handle unsupported recursive JSON flattening or one-to-many split scenarios.
4. MySQL JSON to Milvus vector conversion, including the expected SeaTunnel field type and required transform/conversion step.
5. LLM/file-content extraction boundaries: what the LLM transform can process directly and what must be handled before data enters the transform.

## Acceptance criteria

- The docs include at least two complete HOCON snippets.
- The docs clearly separate supported built-in behavior from scenarios requiring custom transform code.
- English and Chinese docs are consistent.
- The Milvus vector example matches the current Milvus connector type mapping.

nielifeng:
Closes #11277

## What changed
- Add a new Transform recipes page (EN/ZH): nested JSON extraction, one-to-many guidance, MySQL -> Milvus FLOAT_VECTOR recipes.
- Register the new page in docs sidebar under Transforms.

## Notes
- Includes 2+ complete HOCON snippets per issue requirement.
- The Milvus vector examples are aligned with current type mapping (SeaTunnel FLOAT_VECTOR).

### DIFF
--- a/docs/en/transforms/recipes.md
+++ b/docs/en/transforms/recipes.md
@@ -1,0 +1,283 @@
+# Transform Recipes
+
+The transform docs describe individual plugins, but real jobs often need scenario-level configurations that combine:
+
+- Source parsing (how raw data becomes a row)
+- Transforms (how fields are extracted/converted)
+- Sink schema requirements (especially for vector databases)
+
+This page provides practical “recipes” you can copy and adapt.
+
+## Recipe 1: Extract deeply nested JSON (Kafka/File) and write to MySQL
+
+### When to use
+
+- Your upstream data is a nested JSON payload (Kafka message value or a JSON-lines file).
+- You want to map a few nested fields into flat columns and write them to a relational table.
+
+### Key point: keep the original JSON as a single column
+
+The JsonPath transform reads JSON from a single source field (STRING/BYTES). For Kafka, that typically means you keep the message value as a single field; for files, you can read each line as a single `content` field.
+
+#### Kafka example: keep message value as one column
+
+If you want to apply `JsonPath` on Kafka message values, make sure the Kafka source outputs the whole message value as a single field. A simple way is to use `format = text` and do not configure `schema`, then Kafka will output a single string column named `content`.
+
+```hocon
+source {
+  Kafka {
+    plugin_output = "kafka_raw"
+    topic = "events"
+    bootstrap.servers = "kafka:9092"
+    format = text
+  }
+}
+
+transform {
+  JsonPath {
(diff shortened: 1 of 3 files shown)


DanielLeens (commented):
Thanks for adding this. I reviewed the new recipe page against the current implementation instead of treating it as docs-only prose, and I do not see a source-side blocker in the current head.

The key paths line up with the code:

- The JsonPath recipe is consistent with `JsonPathTransform#doTransform()`, which currently accepts both `STRING` and `BYTES` inputs. So the split between Kafka `format = text` (`content`) and Kafka `format = NATIVE` (`value`, bytes) is grounded in the actual transform path.
- The Milvus recipe is also aligned with the runtime model:
  - `FLOAT_VECTOR` is handled as a `ByteBuffer`
  - `DynamicCompile` can explicitly build that with `VectorType.VECTOR_FLOAT_TYPE` plus `VectorUtils.toByteBuffer(...)`
  - `enable_auto_id = true` only becomes effective when the upstream schema really carries a primary key, which matches the current Milvus sink path
- I also checked the JDBC caveat. In the current catalog merge path, if the query still includes the primary-key columns, JDBC can preserve that metadata; if the query drops them, the docs correctly warn users to pre-create the collection or keep the key in the schema explicitly.

So from Daniel’s side I do not see a new code/doc blocker here.

At this point the remaining gate is the current `Build` run, which is still in progress. Once that check is green, this looks ready from the documentation side.

davidzollo (approved):
+1

PR #11427, opened 2026-07-13
[Fix][seatunnel-common,seatunnel-core] Fix comma-containing plain variable parsing

NganWave:
### Problem
    When using a plain variable containing commas (e.g., passing a list of required columns for database table reading), the parser only accepted the text before the first comma, or threw an error.
    ### Investigation
    I first tried the approach shown in the source code(AbstractCommandArgs) below, but it didn't work as expected. After tracing through ConfigBuilder, I discovered that the issue was in PlaceholderUtils: the escaped quotes (\") wrapping the variable value were not being properly unwrapped.
    ### Fix
    I fixed the unwrapping logic in PlaceholderUtils so that comma-containing values are parsed correctly.
    Verification
    I tested the fix by running SeaTunnelEngineLocalExample locally after recompiling the seatunnel-common and seatunnel-core/seatunnel-core-starter modules. After that, I replaced the original jars in my local SeaTunnel installation ($SEATUNNEL_HOME/starter) with the newly recompiled related starter jars (e.g., seatunnel-starter,spark-starter,flink-starter), and verified the fix by running [ $SEATUNNEL_HOME//bin/seatunnel.sh   -c config/fake_to_console.conf   -m local]. It now works as expected.

image

### DIFF
--- a/seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/PlaceholderUtils.java
+++ b/seatunnel-common/src/main/java/org/apache/seatunnel/common/utils/PlaceholderUtils.java
@@ -18,6 +18,7 @@
 package org.apache.seatunnel.common.utils;
 
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.JsonNode;
+import org.apache.seatunnel.shade.org.apache.commons.lang3.StringUtils;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -38,7 +39,7 @@ public class PlaceholderUtils {
         while (matcher.find()) {
             String replacement =
                     value != null && !value.isEmpty()
-                            ? value
+                            ? StringUtils.unwrap(value, "\"")
                             : (matcher.group(1) != null
                                     ? matcher.group(1).substring(1).trim()
                                     : defaultValue);
--- a/seatunnel-e2e/seatunnel-engine-e2e/connector-console-seatunnel-e2e/src/test/java/org/apache/seatunnel/engine/e2e/console/FakeSourceToConsoleIT.java
+++ b/seatunnel-e2e/seatunnel-engine-e2e/connector-console-seatunnel-e2e/src/test/java/org/apache/seatunnel/engine/e2e/console/FakeSourceToConsoleIT.java
@@ -24,12 +24,33 @@ import org.junit.jupiter.api.Test;
 import org.testcontainers.containers.Container;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 
 public class FakeSourceToConsoleIT extends SeaTunnelEngineContainer {
 
     @Test
     public void testFakeSourceToConsoleSink() throws IOException, InterruptedException {
         Container.ExecResult execResult = executeSeaTunnelJob("/fakesource_to_console.conf");
+        Assertions.assertEquals(0, execResult.getExitCode());
+    }
+
+    @Test
+    public void testCommaParams() {
+        List<String> variables = new ArrayList<>();
+        variables.add("date=2026-07-13");
+        variables.add("cols=\"addr1,addr2\"");
+        variables.add("required_cols=[date,id,name,addr1]");
+        Container.ExecResult execResult = null;
+        try {
+            execResult =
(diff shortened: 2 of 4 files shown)


DanielLeens (changes requested):
Thanks for the contribution! I went through the full diff and traced the CLI variable replacement path. The source-side approach looks reasonable, but I found one blocking issue in the regression test that we should fix before merging.

# What This PR Fixes
- User pain point: `-i key="a,b"` should treat the comma as part of the variable value, not as a separator.
- Fix approach: unwrap one layer of surrounding double quotes in `PlaceholderUtils.replacePlaceholders(...)` before inserting the value into string placeholders.
- One-line summary: this makes comma-containing plain variables usable in SeaTunnel config placeholders.

# Full Runtime Path
```text
SeaTunnel CLI / E2E executeJob
  -> AbstractTestContainer.executeJob(..., variables)
      -> appends "-i" and the variable value
  -> AbstractCommandArgs.variables
      -> ParameterSplitter keeps commas inside quoted values
  -> ConfigBuilder.backfillUserVariables(...)
      -> System.setProperty(key, value)
  -> ConfigBuilder.processVariable(...)
      -> PlaceholderUtils.replacePlaceholders(...)
  -> PlaceholderUtils.replacePlaceholders(...)
      -> StringUtils.unwrap(value, "\"")
```

# Findings

Issue 1: The new regression test does not run the comma-parameter config
- Location: `seatunnel-e2e/seatunnel-engine-e2e/connector-console-seatunnel-e2e/src/test/java/org/apache/seatunnel/engine/e2e/console/FakeSourceToConsoleIT.java:46`
- Problem: `testCommaParams` builds variables such as `cols="addr1,addr2"` and `required_cols=[date,id,name,addr1]`, but it still executes `/fakesource_to_console.conf`. That old config has no `${cols}`, `${required_cols}`, SQL transform, or filter, so the new `fakesource_to_console_with_comma_params.conf` is never exercised.
- Risk: CI could pass without proving the actual bug is fixed. Since this PR changes a shared config placeholder utility, the regression test needs to cover the real placeholder replacement path.
- Suggested fix: run `/fakesource_to_console_with_comma_params.conf` in this test, and ideally assert output or job behavior that proves both `addr1,addr2` and `required_cols` were parsed correctly. A small direct unit test for `PlaceholderUtils` would also be helpful.
- Severity: High

# Test Stability
The added E2E currently has a high-risk coverage gap: it deterministically runs the wrong config file, so it is not a valid regression test for this bug. Once pointed at the new config, the test structure itself should be fine.

# Merge Conclusion

1. Blockers
- Issue 1: The regression test must execute the new comma-parameter config so the target path is actually covered.

2. Non-blocking suggestions
- Consider adding a lightweight unit test for `PlaceholderUtils` covering normal values, quoted comma values, and fallback defaults.

Overall, the implementation direction is good and nicely focused. Please wire the test to the new config and I’ll be happy to re-review.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from scratch and rechecked the exact placeholder replacement path that was blocking in the previous round.

# What changed in this rereview
- The earlier blocker is fixed: `testCommaParams` now runs `fakesource_to_console_with_comma_params.conf`, so the new regression actually exercises `${cols}` and `${required_cols}` through the real CLI/config replacement path.
- The source-side change in `PlaceholderUtils.replacePlaceholders(...)` is still focused: it unwraps one outer pair of double quotes before inserting the value, which is exactly the behavior needed for inputs like `-i cols="addr1,addr2"`.
- I do not see a new source-level blocker in the current head.

# Runtime path rechecked
```text
CLI / test container executeJob(..., variables)
  -> appends -i key=value arguments
  -> AbstractCommandArgs variable parsing keeps quoted commas together
  -> ConfigBuilder.processVariable(...)
      -> PlaceholderUtils.replacePlaceholders(...)
          -> StringUtils.unwrap(value, "\"")
  -> SQL / filter config receives comma-containing values without the extra shell quotes
```

# Current gate state
The remaining gate is CI rather than source logic. The latest `Build` is still queued on this head, so I am not approving ahead of the required checks.

From Daniel's side, once the current `Build` finishes green, I do not have a new code-side blocker to add on top of this revision.

davidzollo (approved):
+1
LGTM

PR #11430, opened 2026-07-14
[Test][E2E] Stabilize FakeSource restore output assertions

DanielLeens:
## Why
Recent `engine-v2-it` restore scenarios are flaky because the tests assert that the physical LocalFile line count must exactly match the logical FakeSource row count after failover. In these tests, FakeSource assigns work in split-sized chunks and restore can replay whole splits after the last completed checkpoint, so the physical output can legitimately exceed the logical row count while still reflecting a successful recovery.

## What
- add a shared assertion helper for FakeSource restore output
- require recovered output to reach the logical minimum row count
- require recovered output to remain aligned to FakeSource split boundaries
- require the output to stabilize instead of relying on a fixed sleep plus an exact physical line count
- keep the stricter exact-count assertions for the non-failover run-ok cases unchanged

## Verification
- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -T 3C -f seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/pom.xml spotless:apply`
- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -T 3C -DskipTests verify -f seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/pom.xml`

## Notes
- I did not run local Docker or E2E containers in this round because that was explicitly ruled out for this task.

### DIFF
--- a/seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/ClusterFaultToleranceIT.java
+++ b/seatunnel-e2e/seatunnel-engine-e2e/connector-seatunnel-e2e-base/src/test/java/org/apache/seatunnel/engine/e2e/ClusterFaultToleranceIT.java
@@ -374,9 +374,11 @@ public class ClusterFaultToleranceIT {
                                         JobStatus.FINISHED, objectCompletableFuture.get());
                             });
 
-            Long fileLineNumberFromDir =
-                    FileUtils.getFileLineNumberFromDir(testResources.getLeft());
-            Assertions.assertEquals(testRowNumber * testParallelism, fileLineNumberFromDir);
+            FaultToleranceFakeSourceAssertions.assertOutputRecoveredAndStable(
+                    testResources.getLeft(),
+                    testRowNumber * testParallelism,
+                    testRowNumber,
+                    60_000L);
 
         } finally {
             if (engineClient != null) {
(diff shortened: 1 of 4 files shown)


nzw921rx (approved):
+1

DanielLeens:
Thanks for the contribution. I re-reviewed the latest head from the real FakeSource restore path instead of only looking at the changed assertions.

## What this PR solves
- User pain: these fault-tolerance E2E cases were still asserting an exact final line count plus a fixed `sleep(10000)`, even though the current FakeSource restore model can replay whole splits before the next completed checkpoint. That makes the tests flaky and can also produce false failures on a valid restore.
- Fix approach: the PR centralizes the restore assertions in `FaultToleranceFakeSourceAssertions.assertOutputRecoveredAndStable(...)`, waits for the LocalFile output to stop growing, requires at least the logical row count, and verifies any extra rows are still split-aligned.
- One-line summary: the tests now validate FakeSource restore semantics instead of an over-strict exact-row-count assumption.

## Runtime path checked
```text
FakeSource template
  -> row.num = dynamic_test_row_num_per_parallelism
  -> split.num = 5

Split discovery
  -> FakeSourceSplitEnumerator.discoverySplits()
      -> splitRowNum = ceil(row.num / split.num)
      -> assigned splits go into checkpoint state

Reader execution
  -> FakeSourceReader.pollNext()
      -> consume one split
      -> generate all rows for that split in one poll cycle

Restore window
  -> if failure happens after split assignment but before the next completed checkpoint
      -> restoreEnumerator() can rediscover that split
      -> the whole split can replay

Assertion path
  -> FaultToleranceFakeSourceAssertions.assertOutputRecoveredAndStable()
      -> currentLineCount >= logical row count
      -> currentLineCount % splitRowCount == 0
      -> output must remain stable for the configured window
```

## Merge conclusion
### Conclusion: can merge

1. Blocking items
- None from Daniel's source review.

2. Suggested follow-up
- If we later reuse this helper on FakeSource templates with `split.num != 5`, it would be better to parameterize the split count instead of keeping the current constant. That is not a blocker for this PR because the touched fault-tolerance templates all use `split.num = 5`.

The direction here is correct and the helper matches the current FakeSource restore semantics. I do not see a blocker on the current head.

davidzollo (approved):
LGTM

PR #10680, opened 2026-03-31
[Bug][Connector-StarRocks] Fix label reuse on stream_load failure causing infinite retry and data loss

xyueji:
### Purpose of this pull request

Fix two issues in StarRocks sink connector when `stream_load` fails with "Label has already been used":

**Bug 1 - Data loss in `StarRocksSinkManager`:** When a `stream_load` fails and StarRocks retains the label, the current code detects "Label has already been used" and `break`s out of the retry loop, **silently skipping the entire batch**. This causes data loss because the data was never actually loaded into StarRocks.

**Fix:** Replace `break` with generating a new label and `continue` to retry the batch with the new label.

**Bug 2 - Missing `reCreateLabel` flag in `StarRocksStreamLoadVisitor`:** When `doStreamLoad()` returns Status="Fail" with a message containing "has already been used", the thrown `StarRocksConnectorException` does not set `reCreateLabel=true`. This prevents the retry logic from generating a new label via the `needReCreateLabel()` path.

**Fix:** Detect "has already been used" in the failure message and pass `reCreateLabel=true` to the exception constructor.

### Does this PR introduce _any_ user-facing change?

No. This is a bug fix for an internal retry mechanism. Users will see that `stream_load` retries now succeed (with a new label) instead of either looping infinitely or silently losing data.

### How was this patch tested?

Manually tested by reproducing the scenario where StarRocks retains a failed label. After the fix, the retry generates a new label and completes successfully without data loss.

### Check list

* [x] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
[...]

PR #11431, opened 2026-07-14
[Fix][Connector-V2] Prevent silent StarRocks batch data loss

davidzollo:

StarRocks Stream Load failures could previously exhaust retries and still clear the in-memory batch. The sink also treated a reused label as a reason to skip the batch, even though StarRocks may have committed the original transaction. Both paths can silently lose records.

This patch makes the sink fail closed:

- Retain the pending rows and their Stream Load label across failed `flush()` calls.
- Release buffered rows only after `Success`, `Publish Timeout`, `VISIBLE`, or `COMMITTED`.
- Query the existing label before deciding the outcome of `Label Already Exists`.
- Create a replacement label only when StarRocks explicitly reports `ABORTED`.
- Bound `PREPARE` polling to three minutes, apply the remaining deadline to HTTP I/O, and preserve thread interruption.
- Reject false, null, failed, and unknown responses without clearing the batch.

Related to #10680. Thanks to @xyueji for identifying the label-reuse failure mode. This PR keeps the pending batch and original label until the transaction state is known; immediately retrying an unknown outcome with a new label can duplicate data that StarRocks already committed.

Yes.

Previously, a StarRocks Stream Load failure could be logged while the buffered records were cleared, allowing the job to continue with missing rows. A reused label was also skipped without confirming whether its transaction was committed.

After this change, unresolved outcomes fail the write and retain the pending batch and label. The sink does not claim exactly-once across process restarts; it preserves the existing at-least-once contract while preventing silent batch loss in one writer lifecycle.

No configuration, default value, public API, SPI, serialization format, or connector registration changes are introduced.

Unit and module verification on JDK 8:

```bash
mvn -T 3C -f seatunnel-connectors-v2/connector-starrocks/pom.xml spotless:apply -nsu -Dmaven.gitcommitid.skip=true
mvn -T 3C -f seatunnel-connectors-v2/connector-starrocks/pom.xml test -nsu -Dmaven.gitcommitid.skip=true
mvn -T 3C -f seatunnel-connectors-v2/connector-starrocks/pom.xml -q -DskipTests verify -nsu -Dmaven.gitcommitid.skip=true
```

Result: 88 tests, 0 failures, 0 errors, 0 skipped. The new regression matrix covers false/null results, retry exhaustion, pending-batch and label retention, new records arriving after a failed flush, interrupted retries, final-attempt label handling, `Success`, `Publish Timeout`, reused-label `VISIBLE`/`COMMITTED`/`ABORTED`/`UNKNOWN`, ordinary failures, unexpected statuses, bounded `PREPARE` polling, and interruption.

The Stream Load protocol was also verified against a real `starrocks/allin1-ubuntu:3.3.4` FE/BE instance:

1. The first request with a fixed label returned `Success`.
2. A second request with the same label returned `Label Already Exists`.
3. `get_load_state` returned `VISIBLE`.
4. The target table contained only the first row, proving that the reused label did not insert a duplicate row.
5. An invalid load returned `Fail`; `get_load_state` returned `ABORTED`; a later valid retry succeeded.

A full MySQL CDC to StarRocks E2E job was not run. This patch does not modify the MySQL source, CDC deserialization, checkpoint state, schema handling, connector registration, or packaging; the real validation targets the Stream Load transaction protocol where this defect occurs. `seatunnel-engine-ui` was skipped because it is outside the affected module.

### Check list

* [x] Documentation is not required because no configuration or public contract changes.
* [x] `incompatible-changes.md` is not required because this change is backward compatible.
* [x] Connector registration and packaging files are not changed because this is an internal StarRocks sink correctness fix.

### DIFF
--- a/seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/HttpHelper.java
+++ b/seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/HttpHelper.java
@@ -105,9 +105,30 @@ public class HttpHelper {
 
     public Map<String, Object> doHttpGet(String getUrl, Map<String, String> header)
             throws IOException {
+        return doHttpGet(getUrl, header, 0);
+    }
+
+    /**
+     * Executes a JSON GET request with a caller-provided deadline for connection and response I/O.
+     *
+     * @param getUrl target URL
+     * @param header request headers
+     * @param timeoutMs timeout in milliseconds, or zero to preserve the client default
+     * @return parsed JSON response, or null when the response has no usable entity
+     */
+    Map<String, Object> doHttpGet(String getUrl, Map<String, String> header, int timeoutMs)
+            throws IOException {
         log.info("Executing GET from {}.", getUrl);
         try (CloseableHttpClient httpclient = HttpClients.createDefault()) {
             HttpGet httpGet = new HttpGet(getUrl);
+            if (timeoutMs > 0) {
+                httpGet.setConfig(
+                        RequestConfig.custom()
+                                .setConnectTimeout(timeoutMs)
+                                .setConnectionRequestTimeout(timeoutMs)
+                                .setSocketTimeout(timeoutMs)
+                                .build());
+            }
             if (null != header) {
                 for (Map.Entry<String, String> entry : header.entrySet()) {
                     httpGet.setHeader(entry.getKey(), String.valueOf(entry.getValue()));
--- a/seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksSinkManager.java
+++ b/seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksSinkManager.java
@@ -20,7 +20,6 @@ package org.apache.seatunnel.connectors.seatunnel.starrocks.client;
 import org.apache.seatunnel.shade.com.google.common.base.Strings;
 
 import org.apache.seatunnel.api.table.catalog.TableSchema;
-import org.apache.seatunnel.common.utils.ExceptionUtils;
 import org.apache.seatunnel.connectors.seatunnel.starrocks.config.SinkConfig;
 import org.apache.seatunnel.connectors.seatunnel.starrocks.exception.StarRocksConnectorErrorCode;
 import org.apache.seatunnel.connectors.seatunnel.starrocks.exception.StarRocksConnectorException;
@@ -40,6 +39,13 @@ public class StarRocksSinkManager {
     private final List<byte[]> batchList;
 
     private final StarRocksStreamLoadVisitor starrocksStreamLoadVisitor;
+
+    /**
+     * Snapshot whose outcome is not confirmed yet. Its rows and label must survive failed flush
+     * calls so a later retry cannot bypass StarRocks label-based idempotency.
+     */
+    private StarRocksFlushTuple pendingFlush;
+
     private volatile boolean initialize;
     private volatile Exception flushException;
     private int batchRowCount = 0;
@@ -65,9 +71,18 @@ public class StarRocksSinkManager {
         initialize = true;
     }
 
+    /**
+     * Buffers a record after resolving any previously failed flush.
+     *
+     * <p>A new record cannot join a pending snapshot because clearing that snapshot after a retry
+     * would otherwise clear a record that was never sent.
+     */
     public synchronized void write(String record) throws IOException {
         tryInit();
         checkFlushException();
+        if (pendingFlush != null) {
+            flush();
+        }
         byte[] bts = record.getBytes(StandardCharsets.UTF_8);
         batchList.add(bts);
         batchRowCount++;
@@ -82,29 +97,39 @@ public class StarRocksSinkManager {
         flush();
     }
 
+    /**
+     * Flushes buffered rows and releases them only after StarRocks confirms success or commit.
+     *
+     * <p>Failed or unknown outcomes retain both the batch snapshot and its label across calls. The
+     * label changes only when StarRocks explicitly reports that the previous transaction aborted.
+     */
     public synchronized void flush() throws IOException {
         checkFlushException();
-        if (batchList.isEmpty()) {
-            return;
+        if (pendingFlush == null) {
+            if (batchList.isEmpty()) {
+                return;
+            }
+            pendingFlush =
+                    new StarRocksFlushTuple(
+                            createBatchLabel(), batchBytesSize, new ArrayList<>(batchList));
         }
-        String label = createBatchLabel();
-        StarRocksFlushTuple tuple =
-                new StarRocksFlushTuple(label, batchBytesSize, new ArrayList<>(batchList));
+        StarRocksFlushTuple tuple = pendingFlush;
+        boolean loadSucceeded = false;
         for (int i = 0; i <= sinkConfig.getMaxRetries(); i++) {
             try {
                 Boolean successFlag = starrocksStreamLoadVisitor.doStreamLoad(tuple);
-                if (successFlag) {
+                if (Boolean.TRUE.equals(successFlag)) {
+                    loadSucceeded = true;
                     break;
                 }
+                throw new StarRocksConnectorException(
+                        StarRocksConnectorErrorCode.FLUSH_DATA_FAILED,
+                        String.format(
+                                "Stream Load returned a non-success result for %s.%s with label [%s].",
+                                sinkConfig.getDatabase(), sinkConfig.getTable(), tuple.getLabel()));
             } catch (Exception e) {
                 log.warn("Writing records to StarRocks failed, retry times = {}", i, e);
 
-                String labelAlreadyMessage =
-                        String.format("Label [%s] has already been used", label);
-                if (ExceptionUtils.getMessage(e).contains(labelAlreadyMessage)) {
-                    log.warn("Label [{}] has already been used, Skipping this batch", label);
-                    break;
-                }
                 if (i >= sinkConfig.getMaxRetries()) {
                     throw new StarRocksConnectorException(
                             StarRocksConnectorErrorCode.WRITE_RECORDS_FAILED,
@@ -131,10 +156,20 @@ public class StarRocksSinkManager {
                 } catch (InterruptedException ex) {
                     Thread.currentThread().interrupt();
                     throw new StarRocksConnectorException(
-                            StarRocksConnectorErrorCode.FLUSH_DATA_FAILED, e);
+                            StarRocksConnectorErrorCode.FLUSH_DATA_FAILED,
+                            "Interrupted while waiting to retry Stream Load.",
+                            ex);
                 }
             }
         }
+        if (!loadSucceeded) {
+            throw new StarRocksConnectorException(
+                    StarRocksConnectorErrorCode.WRITE_RECORDS_FAILED,
+                    "Stream Load did not complete successfully; buffered records were retained.");
+        }
+        // Buffered rows can only be released after StarRocks confirms a successful or committed
+        // load.
+        pendingFlush = null;
         batchList.clear();
         batchRowCount = 0;
         batchBytesSize = 0;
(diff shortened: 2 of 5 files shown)


DanielLeens (approved):
Thanks for the contribution. I reviewed the latest head from the actual StarRocks sink write/flush path rather than only reading the diff.

# What this PR fixes
- User pain: when SeaTunnel loses the first Stream Load response but StarRocks has already accepted the label, the old code can treat a later `Label Already Exists` as "skip this batch" and clear the buffered rows anyway. That turns an uncertain write outcome into silent batch loss.
- Fix approach: the new code keeps one pending batch snapshot and its label alive across retries, reuses the same label until StarRocks explicitly reports `ABORTED`, and blocks on label-state confirmation instead of silently dropping the batch.
- One-line summary: this PR changes the StarRocks sink from "unknown result -> drop the batch" to "unknown result -> keep the batch until StarRocks confirms what happened".

# Runtime path I checked
```text
SeaTunnel row write
  -> StarRocksSinkWriter.write(...)
      -> StarRocksSinkManager.write(...)
          -> buffer row bytes in batchList
          -> if a previous flush outcome is still unresolved, flush that pending snapshot first

Checkpoint / close path
  -> StarRocksSinkWriter.prepareCommit()
  -> StarRocksSinkWriter.close()
      -> StarRocksSinkManager.flush()
          -> create or reuse pendingFlush(label + rows)
          -> StarRocksStreamLoadVisitor.doStreamLoad(...)
              -> Success / Publish Timeout: accept current label as successful
              -> Label Already Exists: query /get_load_state for the same label
              -> Fail + "Label [...] has already been used": query /get_load_state for the same label
              -> ABORTED: allow a new label
          -> only clear buffered rows after StarRocks confirms success / committed state
```

# Key findings
1. The normal checkpoint path really does hit this change because `prepareCommit()` flushes through `StarRocksSinkManager.flush()`.
2. The important behavioral correction is in `StarRocksSinkManager.flush()` and `StarRocksStreamLoadVisitor.doStreamLoad(...)`: the batch snapshot is no longer discarded when the first request result is uncertain.
3. Treating `Publish Timeout` as success is consistent with StarRocks stream-load semantics here because the request has already been accepted and the safe follow-up action is to wait for publication, not resend under a new label.
4. The new tests cover the main uncertain-outcome cases: reused label, aborted label, prepare-state timeout, interruption, and label reuse across manager retries.

# Findings
I did not find a new blocking code issue on the current head.

- No API, config option, default value, or serialization contract changed.
- The main runtime tradeoff is intentional: checkpoint/close can now wait for label-state confirmation instead of returning early and risking silent data loss.
- The added unit tests look structurally stable to me. I did not see flaky patterns like fixed sleeps, external ports, or container timing dependencies in the new test coverage.

# Merge conclusion
### Conclusion: can merge

1. Blocking items

2. Suggested follow-up
- None required for merge.

Overall, this is a focused fix on the real data-safety path, and the current implementation is much safer than the old "label reused -> skip batch" behavior.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `dae6fcda91fe` from scratch.

# What this PR fixes
- User pain: A failed or reused-label Stream Load could clear the in-memory batch and let the job continue, silently dropping (or duplicating) StarRocks rows.
- Fix approach: Fail closed by retaining the pending `StarRocksFlushTuple` (rows + label) across failed flushes, only releasing on `Success`/`Publish Timeout`/`VISIBLE`/`COMMITTED`, resolving `Label Already Exists` via a deadline-bounded (3 min) label-state poll with the remaining budget applied to HTTP I/O, and rejecting null/failed/unknown responses with exceptions.
- One-line summary: The correctness fix is sound and well-commented, with pendingFlush lifecycle, label reuse resolution, and interrupt handling all verified against the diff; remaining items are documentation gaps, the hard-coded 3-minute timeout that should be a SinkConfig option, a minor Javadoc inaccuracy on negative timeout values, and a duplicated debug log in the label-existed branch.

# Runtime chain I rechecked
```text
StarRocksSinkManager.write()  StarRocksSinkManager.java:80
  -> flush() when pendingFlush != null (resolve prior failure before buffering new record)  StarRocksSinkManager.java:84
    -> pendingFlush created once from batchList snapshot, reused across retries  StarRocksSinkManager.java:107-114
    -> StarRocksStreamLoadVisitor.doStreamLoad(tuple)  StarRocksSinkManager.java:118
      -> resultStatus == Success | Publish Timeout -> return true  StarRocksStreamLoadVisitor.java:139-141
      -> resultStatus == Label Already Exists -> checkLabelState(host, label) -> return true  StarRocksStreamLoadVisitor.java:143-149
      -> resultStatus == Fail && isLabelAlreadyUsed(...) -> checkLabelState(...) -> return true  StarRocksStreamLoadVisitor.java:150-154
      -> Fail (other) or unknown status -> throw FLUSH_DATA_FAILED (batch retained)  StarRocksStreamLoadVisitor.java:178-183
      checkLabelState(host, label)  StarRocksStreamLoadVisitor.java:237
        -> loop while nanoTime < deadlineNanos (labelStateTimeoutMs, default 3 min)  StarRocksStreamLoadVisitor.java:239-240
        -> HttpHelper.doHttpGet(url, header, remainingTimeoutMs(deadlineNanos))  StarRocksStreamLoadVisitor.java:249-253
          -> RequestConfig with connect/request/socket timeouts when timeoutMs > 0  HttpHelper.java:122-129
        -> VISIBLE | COMMITTED -> return (success)  StarRocksStreamLoadVisitor.java:275-277
        -> PREPARE -> sleepBeforeNextLabelCheck(++idx, deadlineNanos, label), continue  StarRocksStreamLoadVisitor.java:278-279
        -> ABORTED -> throw (only path where manager may create a replacement label)  StarRocksStreamLoadVisitor.java:280-281
        -> deadline exhausted -> throw FLUSH_DATA_FAILED timeout  StarRocksStreamLoadVisitor.java:303-307
    <- on !loadSucceeded: throw WRITE_RECORDS_FAILED, pendingFlush kept  StarRocksSinkManager.java:165-169
    <- on success: pendingFlush = null; batchList.clear()  StarRocksSinkManager.java:170-174
```

# Findings
**Issue 1: User-facing behavior changes (fail-closed flush, 3-minute label-state bound, `Publish Timeout` treated as success) are not documented in the StarRocks sink docs or release notes**
- Location: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksStreamLoadVisitor.java:49`
- Why it matters: - This PR changes observable sink behavior: previously an exhausted retry loop or a reused label cleared the batch and the job continued; now the flush throws `WRITE_RECORDS_FAILED`/`FLUSH_DATA_FAILED` and retains the batch, and `Publish Timeout` is now accepted as success.
- None of the changed files touch `docs/en/connector-v2/sink/StarRocks.md` or `docs/zh/connector-v2/sink/StarRocks.md`, so users upgrading will see jobs that previously "succeeded" (with silent loss) now fail hard, with no documented explanation of the new semantics or the new 3-minute `PREPARE` polling deadline.
- Please add a short "Behavior" / delivery-semantics note to both language variants of the sink doc describing the at-least-once, fail-closed contract and the bounded label-state check, and reference this in the release notes / upgrade guide since the PR description itself declares this a user-facing change.
- Evidence: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksStreamLoadVisitor.java:55  private static final String RESULT_PUBLISH_TIMEOUT = "Publish Timeout";`; `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksSinkManager.java:166  "Stream Load did not complete successfully; buffered records were retained.");`
- Suggested fix: This PR intentionally changes user-visible behavior (fail-closed flush, `Publish Timeout` now treated as success, and a hard 3-minute bound on label-state polling), but `docs/en/connector-v2/sink/StarRocks.md` and the zh variant are untouched. Since jobs that previously ran through (while silently dropping a batch) will now fail with `WRITE_RECORDS_FAILED`, please document the new delivery semantics and the label-state deadline in the sink docs and add a release-note entry so operators are not surprised after upgrading.
- Severity: Medium

**Issue 2: The 3-minute label-state timeout is hard-coded instead of being exposed as a documented SinkConfig option**
- Location: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksStreamLoadVisitor.java:72`
- Why it matters: - `labelStateTimeoutMs` is fixed at `TimeUnit.MINUTES.toMillis(3)` in production code; the configurable constructor is package-private and only reachable from tests.
- On a heavily loaded StarRocks FE, publish of a large transaction can legitimately take longer than 3 minutes; users would then hit the new `Timed out after %d ms while checking the final state of label[%s]` failure with no knob to tune, forcing job failure even though the transaction may still commit.
- Following SeaTunnel connector conventions, this should be an `Option` in `StarRocksSinkOptions`/`SinkConfig` (e.g. `starrocks.label.state.timeout.ms` with default `180000`), wired through the public constructor and documented in the sink options table in `docs/en/connector-v2/sink/StarRocks.md`.
- Since it is a new user-facing tunable governing failure behavior, missing configuration + docs is more than a nit.
- Evidence: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksStreamLoadVisitor.java:95  this.labelStateTimeoutMs = Math.max(1, labelStateTimeoutMs);`
- Suggested fix: The public constructor pins `labelStateTimeoutMs` to the 3-minute default, and the tunable constructor is package-private (test-only). A slow FE publish can exceed 3 minutes and users will have no way to avoid the new timeout failure. Please promote this to a `SinkConfig` option (e.g. `starrocks.label.state.timeout.ms`, default 180000) and document it in the StarRocks sink options table, per our convention for user-facing knobs.
- Severity: Medium

**Issue 3: HttpHelper.doHttpGet Javadoc says "zero to preserve the client default" but negative values behave the same and are undocumented**
- Location: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/HttpHelper.java:115`
- Why it matters: - The new overload guards with `if (timeoutMs > 0)`, so any non-positive value falls through to the client default, while the Javadoc `@param timeoutMs` only mentions `zero`.
- Callers passing a negative value (e.g. a miscomputed remaining deadline) would silently get an unbounded request, which is exactly the failure mode this PR is trying to bound.
- Update the Javadoc to say "non-positive values preserve the client default", or better, validate with `checkArgument(timeoutMs >= 0, ...)` so the contract is explicit; note that `remainingTimeoutMs` in the visitor already clamps to at least 1 ms, so a precondition would be safe.
- Evidence: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/HttpHelper.java:123  if (timeoutMs > 0) {`
- Suggested fix: The Javadoc says "zero to preserve the client default", but the guard is `timeoutMs > 0`, so negative values also silently disable the deadline. Please either document "non-positive" explicitly or add a precondition rejecting negative values, so a miscomputed remaining deadline can't reintroduce an unbounded GET.
- Severity: Low

**Issue 4: Duplicate/unparameterized debug log of the Stream Load response in the label-existed branch**
- Location: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksStreamLoadVisitor.java:143`
- Why it matters: - The response is already logged a few lines earlier inside the `if (LOG.isDebugEnabled())` guard; the `RESULT_LABEL_EXISTED` branch logs the exact same payload again.
- The second call also uses string concatenation with `JsonUtils.toJsonString(loadResult)`, which serializes the whole response even when debug logging is disabled.
- Please drop the duplicate line, or if you want branch-specific context, use parameterized SLF4J logging such as `LOG.debug("Label already exists, response: {}", loadResult)` per the code conventions.
- Evidence: `seatunnel-connectors-v2/connector-starrocks/src/main/java/org/apache/seatunnel/connectors/seatunnel/starrocks/client/StarRocksStreamLoadVisitor.java:136  LOG.debug("StreamLoad response:\n" + JsonUtils.toJsonString(loadResult));`
- Suggested fix: This debug line duplicates the response log emitted just above under the `LOG.isDebugEnabled()` guard, and it eagerly builds the JSON string via concatenation even when debug is off. Suggest removing it, or replacing with a parameterized message like `LOG.debug("Label [{}] already exists, response: {}", flushData.getLabel(), loadResult);` if the branch context is useful.
- Severity: Low

# Review conclusion

- Issue 1: User-facing behavior changes (fail-closed flush, 3-minute label-state bound, `Publish Timeout` treated as success) are not documented in the StarRocks sink docs or release notes
- Issue 2: The 3-minute label-state timeout is hard-coded instead of being exposed as a documented SinkConfig option
- Issue 3: HttpHelper.doHttpGet Javadoc says "zero to preserve the client default" but negative values behave the same and are undocumented
- Issue 4: Duplicate/unparameterized debug log of the Stream Load response in the label-existed branch

corgy-w (approved):

PR #11425, opened 2026-07-13
[Test][Zeta] Stabilize coordinator retryable operation test

DanielLeens:
## Purpose

The coordinator retry test depended on Hazelcast's production default retry budget and pause. Exercising the full default budget makes an otherwise deterministic unit test unnecessarily slow and vulnerable to CI timing variation.

## Changes

- Configure a test-only maximum of three invocation attempts with a one-millisecond retry pause.
- Reset and expose the test operation attempt counter for an exact terminal assertion.
- Preserve the original contract by waiting for the terminal `CompletionException`, checking its retry message, and asserting the exact attempt count.
- Always shut down the test Hazelcast instance in `finally`.

## Validation

- `CoordinatorServiceTest#testSeaTunnelEngineRetryableExceptionOperationCanBeRetryByHazelcast`: 1 test, 0 failures, 0 errors
- `verify -DskipTests` for `seatunnel-engine-server` (includes test compilation and Spotless check)
- Two independent code reviews passed with no blocking findings

`seatunnel-engine-ui` was excluded because this change does not touch the UI.

PR #11432, opened 2026-07-14
[Test][E2E] Isolate HTTP binary assert case

DanielLeens:
## What does this PR do?
This PR splits the `/http_binary_to_assert.conf` scenario out of `testSourceToAssertSink` so the longest HTTP binary download path runs in a fresh container instead of at the tail of the full HTTP source-to-assert matrix.

This failure surfaced while re-checking PR #11425. Its old failed `Build` check came from `HttpIT.testSourceToAssertSink` on the binary assert case rather than from the engine-unit-test diff in that PR.

## Why is this needed?
- keeps the same exit-code assertion
- isolates the longest binary download path from accumulated runtime pressure in the broader matrix
- stays scoped to the flaky baseline case only

## Verification
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-http-e2e -am spotless:apply`
- attempted `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-http-e2e -am -DskipTests test-compile`, but the local reactor currently hits existing shaded dependency compilation failures in `seatunnel-api` before it reaches this module
- no local Docker run, per current environment constraint; CI will provide the runtime validation

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-http-e2e/src/test/java/org/apache/seatunnel/e2e/connector/http/HttpIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-http-e2e/src/test/java/org/apache/seatunnel/e2e/connector/http/HttpIT.java
@@ -365,10 +365,17 @@ public class HttpIT extends TestSuiteBase implements TestResource {
         // http airtable source
         Container.ExecResult execResult22 = container.executeJob("/airtable_json_to_assert.conf");
         Assertions.assertEquals(0, execResult22.getExitCode());
+    }
 
-        // http binary download
-        Container.ExecResult execResult23 = container.executeJob("/http_binary_to_assert.conf");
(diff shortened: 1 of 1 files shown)


nzw921rx (approved):
+1 LGTM

DanielLeens (commented):
Thanks for the test cleanup. I rechecked the latest head `d6653a2a6b2d` on the actual E2E invocation path.

`@TestTemplate` does give this case a fresh container lifecycle here: `TestCaseInvocationContextProvider.resolveParameter(...)` starts the `TestContainer` before each invocation, and the matching `AfterTestExecutionCallback` tears it down right after the test finishes. So moving `http_binary_to_assert.conf` into its own template method is a real isolation change, not just a method rename.

I do not see a source-level blocker on this head from Daniel's side.

Since this is my own PR and I only have read-level review controls in this repository, please treat this as a technical self-check rather than an independent approval. A write-capable maintainer still needs to handle the final review gate.

davidzollo (approved):

PR #11435, opened 2026-07-15
[Fix][E2E] Avoid StarRocks E2E driver download in container

nzw921rx:

Fix flaky StarRocks E2E failures caused by downloading the MySQL JDBC driver inside the test container at runtime.

The previous setup used curl in the SeaTunnel engine container to download the driver from Maven Central. When the network connection is unstable, curl can fail and the E2E test fails before the actual StarRocks test logic runs.

This patch follows the same pattern used by SQLServer CDC E2E: resolve the JDBC driver from the active test classpath and copy it into the engine container with Testcontainers. This keeps dependency resolution in Maven/test setup and removes the runtime network dependency from the container extension.

No.

Ran:

```
./mvnw spotless:apply
```

No E2E tests were run.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-starrocks-e2e/src/test/java/org/apache/seatunnel/e2e/connector/starrocks/StarRocksCDCSinkIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-starrocks-e2e/src/test/java/org/apache/seatunnel/e2e/connector/starrocks/StarRocksCDCSinkIT.java
@@ -17,7 +17,6 @@
 
 package org.apache.seatunnel.e2e.connector.starrocks;
 
-import org.apache.seatunnel.common.utils.SeaTunnelException;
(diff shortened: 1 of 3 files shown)


DanielLeens (commented):
Thanks for the fix. I rechecked the latest head `0958c9be54d2` on the actual StarRocks E2E setup path.

The important part here is that the StarRocks E2E module already declares the MySQL JDBC test dependency in its own `pom.xml`, and `@TestContainerExtension` runs before each template invocation. So resolving `com.mysql.cj.jdbc.Driver` from the active test classpath and copying that jar into `/tmp/seatunnel/plugins/Jdbc/lib` is a real replacement for the in-container `curl`, not a weaker shortcut. The sink, CDC sink, and schema-change cases still exercise the same runtime paths after the driver is prepared.

I do not see a source-level blocker on this head from Daniel's side.

The remaining gate is only the repository review flow, since I only have read-level review controls here.

nzw921rx:

PR #11423, opened 2026-07-13
[Docs] Clarify S3File credential providers for container environments

nielifeng:
### Purpose of this pull request

Clarify the credential-provider boundary for S3File source/sink and Zeta checkpoint storage in container environments. Closes #11275.

### Does this PR introduce any user-facing change?

Yes. The documentation now:

- States that S3File source/sink currently accept only `SimpleAWSCredentialsProvider` and `InstanceProfileCredentialsProvider`.
- Explains that `DefaultAWSCredentialsProviderChain`, `ContainerCredentialsProvider`, ECS task roles, and EKS IRSA are not directly supported by the S3File option parser today.
- Documents the EC2/EKS node instance-profile pattern and the security limits of the static-key fallback.
- Explains that checkpoint storage passes provider classes directly to Hadoop, supports ECS container credentials with the bundled AWS SDK v1 classes, and does not support IRSA with the bundled 1.11.271 dependency.
- Corrects S3 checkpoint examples to use the `s3a://` bucket scheme and keeps English and Chinese guidance aligned.

### How was this patch tested?

- `git diff --check`
- Verified Markdown code-fence balance, relative links, placeholder-only credentials, and English/Chinese structure.
- Verified S3File enum parsing and `hadoop_s3_properties` override order against `S3FileBaseOptions` and `S3HadoopConf`.
- Verified checkpoint configuration routing against `S3Configuration`.
- Inspected the bundled AWS SDK v1.11.271 jar: Default/Container/InstanceProfile providers are present and WebIdentityTokenCredentialsProvider is absent.

Targeted Maven tests were attempted but did not reach the S3 module because the local reactor test lifecycle did not produce required shaded classes; the direct module run also lacked unpublished 2.3.13-SNAPSHOT artifacts. This PR changes documentation only.

PR #11436, opened 2026-07-15
[Feature][Connector-V2] Support arbitrary S3A credentials provider for S3File

### LINKED ISSUE #9233: [Bug][S3] Unable to access S3 using role and webIdentityTokenFile in WebIdentityTokenFileCredentialsProvider
### Search before asking

- [x] #9240

### Description

现状：
目前seatunnel 服务部署在K8s 中， 需要将Mysql 数据通过 mysql cdc同步到 Hive（hive on s3）中， 看起来目前只提供了as/sk 方式进行S3访问，能否功能使 POD的中 webIdentityTokenFile 和 role 访问S3？

```
env {
  parallelism = 1
  job.mode = "STREAMING"
  checkpoint.interval = 10000
}

source {
  MySQL-CDC {
    plugin_output = "fake"
    base-url = "jdbc:mysql://test:3306/test"
    username = "***"
    password = "****"
    
    table-names = ["test.user"]
    table-names-config = [
      {
        table = "test.user"
        primaryKeys = ["id"]
      }
    ]
  }
}

transform {
  Sql {
    plugin_input = "fake"
    plugin_output = "fake1"
    query = "select id, open_id from user"
  }
}

sink {
  Hive {
    plugin_input = "fake1"
    table_name = "tmp.test_hive_sink_on_s3"
    metastore_uri = "thrift://ip-xxx-xx-1-142.cn-north-1.compute.internal:9083"
    hive.hadoop.conf = {
        bucket="s3a://xxxxxxx"
        fs.s3a.endpoint="s3.cn-north-1.amazonaws.com.cn"
        fs.s3a.aws.credentials.provider="org.apache.hadoop.fs.s3a.WebIdentityTokenFileCredentialsProvider" //期望
    }
  }
}
```

参考： https://mvnrepository.com/artifact/software.amazon.awssdk/auth/2.31.28
`software.amazon.awssdk.auth.credentials.WebIdentityTokenFileCredentialsProvider`

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

fatmanverse:

Closes #9233

The S3File connector previously hard-coded the `fs.s3a.aws.credentials.provider` option as an enum (`S3aAwsCredentialsProvider`) that accepted **only** `SimpleAWSCredentialsProvider` and `InstanceProfileCredentialsProvider`. Any other value threw at config-parse time, and the value could not even be smuggled in via `hadoop_s3_properties` because the connector force-overwrote the key. This blocked users who need container/role-based providers — e.g. running SeaTunnel in K8s and authenticating to S3 via `WebIdentityTokenFileCredentialsProvider` (webIdentityTokenFile + role), which is exactly the scenario reported in #9233.

This PR replaces the enum with a free-form string option that passes any fully-qualified S3A credentials provider class through to Hadoop, while keeping the two well-known values and the existing default.

## Changes

- **`S3FileBaseOptions`**: `S3A_AWS_CREDENTIALS_PROVIDER` changed from `enumType(S3aAwsCredentialsProvider.class)` to `stringType()`. Default is unchanged (`com.amazonaws.auth.InstanceProfileCredentialsProvider`). The `S3aAwsCredentialsProvider` enum is removed; the two well-known class names are kept as `public static final String` constants (`SIMPLE_AWS_CREDENTIALS_PROVIDER`, `INSTANCE_PROFILE_CREDENTIALS_PROVIDER`) so factories and downstream modules keep compiling.
- **`S3HadoopConf`**: passes the raw provider string through to `fs.s3a.aws.credentials.provider`, and eagerly verifies the class is on the classpath (`Class.forName`, init=false). The check tries the thread context classloader first, then falls back to the connector's own classloader, so a provider bundled with the connector (including the default) is never rejected by mistake under the Zeta plugin classloader. A missing class produces an actionable error pointing at the config key and `${SEATUNNEL_HOME}/lib`.
- **Factories** (`S3FileSourceFactory`, `S3FileSinkFactory`, `S3RedshiftSinkFactory`): reference `SIMPLE_AWS_CREDENTIALS_PROVIDER` in the `.conditional(...)` rule that requires `access_key`/`secret_key`, instead of the removed enum value. Behavior is unchanged.
- **Docs** (`docs/{en,zh}/connectors/{source,sink}/S3File.md`): the `fs.s3a.aws.credentials.provider` row now documents that any classpath-present S3A provider is accepted, with `ContainerCredentialsProvider` as a container example.
- **Tests**: `S3HadoopConfTest` adds default / custom-passthrough / unknown-class-fails cases; `S3FileFactoryTest` asserts the conditional access_key/secret_key requirement still fires for both source and sink; `S3StorageTest` (hive) updated to the new constant.

## Backward compatibility

Fully backward compatible. The old enum's `toString()` returned the fully-qualified class name, so existing configs already used exactly the strings the new `stringType()` accepts. The default value and the `access_key`/`secret_key` conditional requirement are unchanged.

Yes — `fs.s3a.aws.credentials.provider` now accepts any S3A credentials provider class available on the classpath (previously only two values). This is additive; existing configurations continue to work unchanged. Docs (en/zh) are updated.

## How was this patch tested?

- `./mvnw spotless:apply` (clean)
- Unit tests for `connector-file-s3` (`S3HadoopConfTest`, `S3FileFactoryTest`) and hive `S3StorageTest` — all passing.

## Check list

- [x] Code changed are covered with tests, or it does not need tests for reason
- [x] If any new Jar binary package adding in your PR, please add License Notice — N/A
- [x] If necessary, please update the documentation to describe the new feature — updated en/zh
- [ ] If you are contributing the connector code, please check that the following files are updated: N/A (no new connector)
- [x] Update the `release-note`


DanielLeens (commented):
Thanks for working on this. I traced the actual S3 path on the latest head: `S3FileSourceFactory` / `S3FileSinkFactory` feed the new string provider option into `S3HadoopConf.buildWithReadOnlyConfig(...)`, so the main feature direction makes sense.

I do see one blocker before merge, though:

1. This PR currently carries an unrelated JDBC catalog fix on top of the S3 change. In addition to the S3 provider work, the diff also changes `PostgresCatalog` to filter dropped columns and adds matching PostgreSQL / OpenGauss E2E coverage (`PostgresCatalog.java`, `JdbcPostgresIT.java`, `JdbcOpenGaussIT.java`). That dropped-column catalog fix is outside the S3 credentials-provider call chain and should not be bundled into this PR.

Please either split those JDBC catalog changes into a separate PR, or remove them from this branch so this diff only contains the S3 provider feature itself. Happy to re-review once the scope is cleaned up.

fatmanverse:
Thanks for the review, and for tracing the S3 call chain!

I compared this branch against `origin/dev` (locally and via the GitHub API), and the diff shows 12 files, all S3-scoped — the four `S3File.md` docs, `S3FileBaseOptions`, `S3HadoopConf`, the two S3 factories, `S3RedshiftSinkFactory`, and the S3/hive tests that referenced the removed enum. I couldn't find `PostgresCatalog.java`, `JdbcPostgresIT.java`, or `JdbcOpenGaussIT.java` in it.

I think the branch was just behind `dev` — the dropped-column `PostgresCatalog` fix (`d6df7f2`) landed there recently, so it likely showed up when diffing against an older base. I'll rebase onto the latest `dev` and force-push so the diff is clean; that should make the scope unambiguous. Thanks for your patience!

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `995a90daa5c9` from scratch.

# What this PR fixes
- User pain: The S3File connector's `fs.s3a.aws.credentials.provider` was locked to an enum of two values (`SimpleAWSCredentialsProvider`, `InstanceProfileCredentialsProvider`), and since the connector force-overwrote the key, users could not even work around it via `hadoop_s3_properties` — blocking K8s/IRSA-style auth with `WebIdentityTokenFileCredentialsProvider` as reported in #9233.
- Fix approach: Replace the enum option in `S3FileBaseOptions` with a free-form `stringType()` option (default unchanged), keep the two well-known class names as `public static final String` constants for compile compatibility, pass the raw value through to Hadoop in `S3HadoopConf`, and add an eager class-presence check when the Hadoop conf is built; docs (en/zh, source/sink) updated to describe pass-through semantics.
- One-line summary: The change is small and correctly scoped, but I want to flag the binary-compatibility break from removing the `S3aAwsCredentialsProvider` enum, the fact that the eager class check runs on the conf-building node rather than the workers, and that a job-config-driven fully-qualified class name is now loaded without an interface/allowlist check — plus a docs gap around provider-chain syntax and the `AWSCredentialsProvider` contract.

# Runtime chain I rechecked
```text
S3FileSourceFactory.createSource() / S3FileSinkFactory.createSink()
  -> reads S3FileBaseOptions.S3A_AWS_CREDENTIALS_PROVIDER (now Option<String>)  S3FileBaseOptions.java:66
  -> S3HadoopConf.buildWithConfig(config)  S3HadoopConf.java:~55
       -> setExtraOptions: puts raw provider string into 'fs.s3a.aws.credentials.provider'  S3HadoopConf.java:~60
       -> eager Class.forName(providerClassName) verification (runs on the node building the conf, not workers)  S3HadoopConf.java:~70
       -> putS3SK: sets access_key/secret_key only when provider == SIMPLE_AWS_CREDENTIALS_PROVIDER  S3HadoopConf.java:~80
  -> HadoopFileSystemProxy -> org.apache.hadoop.fs.FileSystem.get(conf)
       -> S3AFileSystem.initialize() -> S3AUtils.createAWSCredentialProviderSet(conf)
            -> reflectively instantiates the configured provider class on the executing worker JVM (this is where a missing/invalid class actually fails at runtime)
```

# Findings
**Issue 1: Possible public API change in seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java`
- Severity: High

**Issue 2: Free-form credentials-provider class name turns the job config into an arbitrary class-loading/instantiation vector without any interface or allowlist check**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java:69`
- Why it matters: - Root cause: `S3A_AWS_CREDENTIALS_PROVIDER` is now a plain `stringType()` that is passed verbatim to `fs.s3a.aws.credentials.provider`. Hadoop's `S3AUtils.createAWSCredentialProviderSet` will reflectively load and *construct* whatever class name is supplied (no-arg or `(URI, Configuration)` constructor), so the constructor of any class on the runtime classpath runs before Hadoop rejects a non-provider type.
- Impact: anyone with permission to submit a SeaTunnel job config (multi-tenant Zeta clusters, REST job submission, shared Flink/Spark clusters) can now trigger loading and construction of arbitrary classes present on the worker classpath — a classic reflection-gadget surface that the previous enum deliberately closed off. A malicious or buggy provider class can also exfiltrate the instance-profile/web-identity credentials it resolves.
- Fix: in `S3HadoopConf`, after resolving the class, additionally assert it is assignable to `com.amazonaws.auth.AWSCredentialsProvider` (or the Hadoop `AWSCredentialsProvider` bridge) before writing the property, and reject anything else with an actionable error. Optionally add a cluster-level allowlist property so operators can pin the accepted provider classes, mirroring how other config-driven class options are handled in SeaTunnel.
- Suggested fix: Moving from the enum to `stringType()` means any class name in the job config gets reflectively loaded *and constructed* by Hadoop's `S3AUtils.createAWSCredentialProviderSet` — constructors of arbitrary classpath classes will run before Hadoop rejects a non-provider type. Since job configs are often submittable by less-trusted users (multi-tenant Zeta, REST submission), please add an `isAssignableFrom(AWSCredentialsProvider)` check in `S3HadoopConf` (on top of the `Class.forName` presence check) and consider an operator-configurable allowlist, so we only pass through classes that are actually credential providers.
- Severity: Medium

**Issue 3: Classpath presence check runs on the node that builds the Hadoop conf, not on the workers that actually instantiate the provider**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java:64`
- Why it matters: - Root cause: the validation happens "when the Hadoop configuration is built", i.e. on the client/coordinator JVM, using the thread-context classloader with a fallback to the connector's own classloader.
- Impact: on distributed deployments (Zeta workers, Flink TaskManagers, Spark executors) the classpath frequently differs from the submitting node — a provider jar dropped only into the coordinator's `${SEATUNNEL_HOME}/lib` will pass the eager check but throw `ClassNotFoundException` deep inside `S3AFileSystem` initialization at task runtime, defeating the purpose of the "actionable error". The connector-classloader fallback also means a class visible only inside the connector plugin jar passes validation even though Hadoop (loaded by a different classloader hierarchy under Zeta plugin isolation) may not resolve it.
- Fix: keep the eager check as a fast-fail convenience but soften the docs claim, and make the runtime failure path equally actionable by wrapping the S3A init `ClassNotFoundException` with the same hint about the config key and `${SEATUNNEL_HOME}/lib` on every node. Also verify (with an e2e test on Zeta) that a provider jar placed only in `lib/` is actually resolvable by the classloader Hadoop uses inside the connector plugin scope.
- Suggested fix: The presence check runs where the Hadoop conf is built (client/coordinator), but the provider is instantiated on every worker — on Zeta/Flink/Spark clusters the jar may exist locally but be missing on workers, so the job passes validation and then fails inside `S3AFileSystem` init with an opaque CNFE. Worth wrapping the runtime failure with the same actionable hint, and adding a Zeta e2e case proving a jar in `${SEATUNNEL_HOME}/lib` is resolvable from the plugin classloader Hadoop uses here.
- Severity: Medium

**Issue 4: Docs promise 'any class on the classpath' works via ${SEATUNNEL_HOME}/lib but omit chain syntax, the interface requirement, and the security implication of job-config-driven class loading**
- Location: `docs/en/connectors/sink/S3File.md:113`
- Why it matters: - Root cause: the new docs row states any S3A provider class on the runtime classpath is accepted and only mentions `${SEATUNNEL_HOME}/lib`, without noting that the class must implement the AWS credentials-provider contract (and expose a no-arg or `(URI, Configuration)` constructor), that Hadoop supports comma-separated chains, or that under Zeta plugin classloader isolation `lib/` visibility is what makes or breaks resolution on each node.
- Impact: users following the doc will hit hard-to-diagnose failures (wrong constructor signature, non-provider class, jar present only on the submit node), and cluster operators are not warned that granting job-submission rights now implies letting users load classes by name.
- Fix: extend the four doc rows to state the interface/constructor requirement, whether comma-separated chains are honored (align with the `S3HadoopConf` validation fix), that the jar must be present on **all** cluster nodes, and add a short security note for operators of shared clusters.
- Evidence: `docs/en/connectors/source/S3File.md:198  | fs.s3a.aws.credentials.provider | string  | yes      | ... any S3A credentials provider class available on the classpath is accepted ...`
- Suggested fix: Please tighten this description: (1) the class must implement the AWS credentials-provider interface and expose a no-arg or `(URI, Configuration)` constructor, otherwise S3A init fails with an opaque error; (2) clarify whether Hadoop-style comma-separated provider chains are accepted; (3) note the jar must be on **every** cluster node, not just the submitter; and (4) add a brief note for operators that this option lets job authors load classes by name, which matters on shared/multi-tenant clusters.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java

- Issue 2: Free-form credentials-provider class name turns the job config into an arbitrary class-loading/instantiation vector without any interface or allowlist check
- Issue 3: Classpath presence check runs on the node that builds the Hadoop conf, not on the workers that actually instantiate the provider
- Issue 4: Docs promise 'any class on the classpath' works via ${SEATUNNEL_HOME}/lib but omit chain syntax, the interface requirement, and the security implication of job-config-driven class loading

fatmanverse:
Thanks for the thorough re-review — the runtime call-graph writeup made each point easy to reason about. Pushed a follow-up commit (`ef5ce98`). Here's how each item was addressed, plus one regression your Issue 3 analysis helped me catch.

**Issue 1 (blocking) — binary compatibility.** Good catch on the type removal. To be precise about the break: the `S3A_AWS_CREDENTIALS_PROVIDER` field descriptor is actually unchanged (both `Option<Enum>` and `Option<String>` erase to the same descriptor), so field references stay binary-compatible; the only real removal was the `S3aAwsCredentialsProvider` enum *type* itself. I've restored that enum as `@Deprecated` (its constants now reference the new `String` constants, single source of truth), which removes the `NoClassDefFoundError` risk for any downstream code compiled against it. The option stays `stringType()` since that's the core of the feature. The source-level generic change is documented in `docs/{en,zh}/introduction/concepts/incompatible-changes.md` with migration guidance.

**Issue 3 — check runs on the conf-building node, not workers. This turned out to be a real bug, not just a robustness concern.** Your analysis was exactly right, and it was already breaking CI: the original eager `Class.forName` check *hard-failed*, and because the default provider (`InstanceProfileCredentialsProvider`) lives in `seatunnel-hadoop-aws` at `provided` scope, it isn't on the classpath of the module that builds the conf in `connector-hive`'s unit tests — so `S3StorageTest` was failing on the *default* config (this is the `unit-test` job that was red). I've changed the check to **best-effort**: a class that cannot be resolved on the conf-building node is now only logged as a `WARN` (with the "must be present on every node running S3A" hint) and passed through, never rejected. `S3StorageTest` is green again.

**Issue 2 — interface check.** Kept a *fast-fail* only for the case that can never work on any node: a class that *does* resolve but does not implement `com.amazonaws.auth.AWSCredentialsProvider` is rejected with an actionable error. The check also splits Hadoop-style comma-separated provider chains and validates each entry. The interface assertion is skipped when the AWS SDK interface itself doesn't resolve from the same classloader (plugin-classloader isolation), to avoid falsely rejecting a valid provider. On the **allowlist**: I'd prefer to leave it out of this PR. It's a new operator-facing config with its own compat surface, and the same class-name-loading vector already exists for any `*.class` key passed through `hadoop_s3_properties` — singling out this one option would be inconsistent. Happy to file a separate issue if a general "class-name option allowlist" is worth pursuing cluster-wide.

**Issue 4 — docs.** Updated all four rows (en/zh × source/sink): the class must implement the AWS credentials-provider contract and expose a no-arg or `(URI, Configuration)` constructor; comma-separated chains are accepted; the jar must be on **every** node; and a short security note that this option lets job authors load classes by name on shared/multi-tenant clusters.

On the runtime `ClassNotFoundException` wrapping and a dedicated Zeta e2e (Issue 3 follow-ups): I've left those out on purpose. The runtime path goes through `HadoopFileSystemProxy`, which is shared by all file connectors, so wrapping it there is a cross-connector change beyond this PR's scope; and a full Zeta e2e feels heavy for a non-blocking item. Both make sense as separate follow-ups if you'd like.

Verified locally: `spotless:apply` clean, `connector-file-s3` unit tests pass, and `connector-hive` `S3StorageTest` passes again. Thanks again for the careful review.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `ef5ce98d5d9f` from scratch.

# What this PR fixes
- User pain: The S3File connector hard-coded `fs.s3a.aws.credentials.provider` as a two-value enum (`SimpleAWSCredentialsProvider`/`InstanceProfileCredentialsProvider`), rejecting any other provider at config-parse time and force-overwriting attempts to sneak one in via `hadoop_s3_properties`, which blocked K8s/IRSA users needing `WebIdentityTokenFileCredentialsProvider` (#9233).
- Fix approach: Replace the enum option in `S3FileBaseOptions` with a free-form `stringType()` option (default unchanged, well-known class names kept as `String` constants), pass the raw value through to Hadoop in `S3HadoopConf`, and add an eager `Class.forName` check per comma-separated chain entry; EN/ZH source and sink docs updated accordingly.
- One-line summary: The pass-through mechanics and doc updates look correct and solve the reported use case, but before merging I want the eager validation to also assert `AWSCredentialsProvider` assignability (not mere class presence), the chain-splitting/trim semantics confirmed against Hadoop's `S3AUtils` behavior, and the removal of the public `S3aAwsCredentialsProvider` enum called out as a breaking change with a migration note.

# Runtime chain I rechecked
```text
S3FileSourceFactory.createSource() / S3FileSinkFactory.createWriter()
  -> S3FileBaseOptions.S3A_AWS_CREDENTIALS_PROVIDER (now stringType, default com.amazonaws.auth.InstanceProfileCredentialsProvider)  S3FileBaseOptions.java:45-70
  -> S3HadoopConf.buildWithConfig(config)  S3HadoopConf.java
     -> validateCredentialsProvider(providerString)  [eager Class.forName per comma-separated entry]
     -> hadoopConf.put("fs.s3a.aws.credentials.provider", providerString)  [raw pass-through, no longer overwritten from enum]
  -> HadoopFileSystemProxy.initialize() -> org.apache.hadoop.fs.s3a.S3AFileSystem.initialize()
     -> S3AUtils.createAWSCredentialProviderSet()  [actual reflective instantiation on worker JVMs; requires no-arg or (URI, Configuration) ctor]
```

# Findings
**Issue 1: Possible public API change in seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java`
- Severity: High

**Issue 2: Free-form credentials provider turns a job-config string into arbitrary class loading/instantiation on every worker JVM — validation should at least enforce interface assignability, not just class presence**
- Location: `docs/en/connectors/source/S3File.md:198`
- Why it matters: - Root cause: the option was changed from a closed enum to a pass-through string, and per the PR description the only guard in `S3HadoopConf` is a `Class.forName(name, false, loader)` presence check. Presence is checked, but nothing verifies the class actually implements `com.amazonaws.auth.AWSCredentialsProvider` before the value is handed to Hadoop, which will later reflectively instantiate it (no-arg or `(URI, Configuration)` constructor) inside the worker JVM.
- Impact: in shared/multi-tenant Zeta clusters, any user who can submit a job can now cause the engine to instantiate an arbitrary classpath class by name. Constructor side effects of unrelated classes become a gadget surface, and a purpose-built "provider" jar dropped into `${SEATUNNEL_HOME}/lib` runs with the full privileges of the SeaTunnel process — including visibility of other tenants' credentials in the same JVM. The docs note this honestly, but the code should narrow the surface too.
- Fix: after `Class.forName`, add `if (!com.amazonaws.auth.AWSCredentialsProvider.class.isAssignableFrom(clazz))` (also accept Hadoop's `AnonymousAWSCredentialsProvider`-style classes if needed) and fail fast with a `SeaTunnelRuntimeException` naming the config key. This rejects gadget classes like `javax.script.*` or logging classes at parse time instead of letting S3A instantiate them. Optionally support an operator-side allowlist system property for hardened deployments.
- Evidence: `docs/en/connectors/sink/S3File.md:113  | fs.s3a.aws.credentials.provider       | string  | yes      | com.amazonaws.auth.InstanceProfileCredentialsProvider | The fully-qualified class name of the S3A credentials provider passed through to Hadoop. ...`
- Suggested fix: The docs correctly warn that this option lets job authors load classes by name, but I'd like the code to narrow that surface as well. In `S3HadoopConf`, after the `Class.forName(name, false, loader)` presence check, please also verify `com.amazonaws.auth.AWSCredentialsProvider.class.isAssignableFrom(clazz)` and fail fast with an actionable error. Today Hadoop will reflectively instantiate whatever class name is supplied, so any classpath class with constructor side effects becomes reachable from a job config on multi-tenant clusters. An assignability check (plus, ideally, an optional operator-side allowlist property) keeps the feature while shrinking the gadget surface.
- Severity: Medium

**Issue 3: Eager Class.forName validation runs on the parsing node only and its chain-splitting semantics must match Hadoop's (trimmed, comma-separated), otherwise it gives false positives/negatives versus actual S3A behavior**
- Location: `docs/en/connectors/sink/S3File.md:113`
- Why it matters: - Root cause: per the diff's documented behavior, "Hadoop-style comma-separated provider chains are accepted and each class is validated independently", and the check "tries the thread context classloader first, then falls back to the connector's own classloader". Both semantics can diverge from what actually happens at runtime.
- Impact 1: validation runs where the config is parsed. A provider jar present on the submitting node but missing on a worker passes validation and then fails deep inside S3A initialization on the worker with a far less actionable `ClassNotFoundException` — defeating the purpose of the eager check. The docs already warn the jar must be on every node, so the error message from the eager check should say the same.
- Impact 2: Hadoop parses `fs.s3a.aws.credentials.provider` with trimmed splitting (it tolerates `"A, B"` with spaces). If the connector splits on bare `','` without `trim()`, a value Hadoop happily accepts is rejected at SeaTunnel config-parse time; conversely, if empty segments (`"A,,B"`) aren't rejected, validation passes but S3A fails.
- Fix: split with trimming (e.g. Hadoop `StringUtils.getTrimmedStrings` semantics), reject empty segments explicitly, and extend the ClassNotFound error text to state that the class must be present on every cluster node, not just the one that validated the config. Add unit cases in `S3HadoopConfTest` for `"A, B"` and `"A,,B"`.
- Evidence: `docs/zh/connectors/sink/S3File.md:112  ... 支持 Hadoop 风格的逗号分隔的提供程序链，且每个类都会被独立校验。该提供程序 jar 必须存在于**每个**集群节点的运行时 classpath 中 ...`
- Suggested fix: Since we document that comma-separated provider chains are validated per class, please make sure the split in `S3HadoopConf` matches Hadoop's semantics: Hadoop trims whitespace around commas (`"A, B"` is valid), so the validator should trim each segment and reject empty ones (`"A,,B"`). Also, because the `Class.forName` check only runs on the node that parses the config, the ClassNotFound error message should explicitly say the jar must be on the classpath of *every* cluster node — otherwise a jar present only on the submitter passes validation and the job dies later on a worker with an opaque S3A stack trace. Worth adding `"A, B"` / `"A,,B"` cases to `S3HadoopConfTest`.
- Severity: Medium

**Issue 4: Removing the public S3aAwsCredentialsProvider enum and changing the Option generic type is a source/binary breaking change for downstream code compiled against connector-file-s3**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java:49`
- Why it matters: - Root cause: `S3A_AWS_CREDENTIALS_PROVIDER` changes from `Option<S3aAwsCredentialsProvider>` (enum) to `Option<String>` and the enum class is deleted outright; the PR compensates in-repo (hive `S3StorageTest`, redshift factory) with new `String` constants, but any external plugin or user code referencing the enum or the old option generic type breaks with `NoClassDefFoundError` / compile errors.
- Impact: connector option classes in `*Options` are effectively public API for downstream connectors (this very PR shows connector-hive and connector-s3-redshift depending on it); third-party forks/extensions will hit binary incompatibility on upgrade. Config-file compatibility is fine (enum `toString()` matched the class name), and the incompatible-changes doc entry is the right move, but the code-level break deserves a softer landing.
- Fix: keep a `@Deprecated` `S3aAwsCredentialsProvider` enum (delegating its `toString()` to the new constants) for one release cycle before deletion, or at minimum call out the binary break explicitly (not just the config semantics) in `docs/en/introduction/concepts/incompatible-changes.md`. Also double-check that the Zeta checkpoint-storage S3 path (see the related docs in PR #11423) doesn't reference the removed enum.
- Suggested fix: Replacing the public `S3aAwsCredentialsProvider` enum with `String` constants keeps in-repo callers compiling, but it's a binary/source break for any external code that referenced the enum or the old `Option<S3aAwsCredentialsProvider>` type (connector-hive and connector-s3-redshift in this repo prove downstream modules do depend on it). I'd keep a `@Deprecated` enum shim for one release, or at least make the incompatible-changes doc entry explicitly mention the code-level (not just config) break, and confirm the Zeta checkpoint-storage S3 code path has no reference to the removed enum.
- Severity: Medium

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-connectors-v2/connector-file/connector-file-s3/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/s3/config/S3FileBaseOptions.java

- Issue 2: Free-form credentials provider turns a job-config string into arbitrary class loading/instantiation on every worker JVM — validation should at least enforce interface assignability, not just class presence
- Issue 3: Eager Class.forName validation runs on the parsing node only and its chain-splitting semantics must match Hadoop's (trimmed, comma-separated), otherwise it gives false positives/negatives versus actual S3A behavior
- Issue 4: Removing the public S3aAwsCredentialsProvider enum and changing the Option generic type is a source/binary breaking change for downstream code compiled against connector-file-s3

fatmanverse:
@SEZ9 I reworked the review fixes in `fcfcf014634d`; this comment supersedes my previous response.

**Issues 1 and 4 - public API compatibility**

The fix now preserves the original public API exactly, rather than only restoring the enum type:

- `S3A_AWS_CREDENTIALS_PROVIDER` remains `Option<S3aAwsCredentialsProvider>` with the same key and default.
- The deprecated `S3aAwsCredentialsProvider` enum remains available.
- A separate `Option<String> S3A_AWS_CREDENTIALS_PROVIDER_CLASS` uses the same stable config key for the new free-form path.
- All in-repo production consumers (S3 source/sink factories, Hive, Redshift, and `S3HadoopConf`) now use the new string option. The legacy field is only a compatibility surface and no longer owns runtime behavior.
- The EN/ZH incompatible-change entry was removed because downstream code no longer needs recompilation or migration.

I also added a compile-time regression assertion by assigning `ReadonlyConfig.get(S3A_AWS_CREDENTIALS_PROVIDER)` directly to `S3aAwsCredentialsProvider`. `javap` confirms that both public generic signatures are present.

**Issue 2 - provider validation and classloader isolation**

Each candidate classloader must now resolve both the configured class and `AWSCredentialsProvider` before assignability is checked. If an isolated TCCL resolves only one side, validation continues with the connector classloader instead of returning early. Resolvable non-provider and abstract classes fail fast; locally unresolvable classes remain a warning because the provider is instantiated on worker JVMs.

**Issue 3 - Hadoop chain semantics**

The validator now calls Hadoop 3.1.4 `StringUtils.getTrimmedStrings` directly. This matches Hadoop for comma/newline separators, whitespace, and a trailing comma, while still rejecting an interior empty entry such as `A,,B`. Tests cover whitespace, newline validation, trailing comma parity with `Configuration.getClasses`, interior empty entries, and the isolated-TCCL bypass.

**Docs**

All four EN/ZH source/sink rows now document the four creation mechanisms supported by Hadoop 3.1.4: `(URI, Configuration)`, `(Configuration)`, static no-arg `getInstance()`, and a no-arg constructor. They also retain the interface, provider-chain, every-node classpath, and shared-cluster security notes.

**Local verification**

- `S3HadoopConfTest,S3FileFactoryTest`: 17 tests passed
- `S3StorageTest`: 2 tests passed
- `connector-s3-redshift`: package passed
- `./mvnw -q -DskipTests verify`: passed across the full reactor

I also attempted the root `./mvnw test`; it stops before the affected modules in `seatunnel-config-shade` with 7 unrelated `NoClassDefFoundError` failures for shaded Typesafe Config classes. The PR CI run for this head is now queued.

DanielLeens (commented):
Thanks for the contribution! I re-reviewed the full current diff and the relevant S3 multipart/credential-provider follow-up runtime path.

The blockers from my earlier round are resolved on the current head. I do not see a new source-side blocker; the remaining gate is the current failing unit-test lane and repository review.

DanielLeens (commented):
Gate follow-up on the unchanged head: GitHub is now reporting merge conflicts with the latest `dev` (`mergeStateStatus=DIRTY`, compare `behind_by=21`). The source-side conclusion from my previous round remains clear, but the branch cannot proceed through the merge gate in its current state.

Could you please sync the latest `dev`, resolve the conflicts, and rerun the required checks? The existing failing unit-test lane is still a separate CI gate. I am happy to recheck the resulting head once the conflict resolution is pushed.

fatmanverse:
@SEZ9 Merged the latest `dev` to clear the merge conflict (merge commit `3f3f86c1f`).

The only conflict was in `connector-hive`'s `S3StorageTest.java`: `dev` (#11459) had refactored that test to extract a shared `TEST_S3_CREDENTIALS_PROVIDER` constant, while this PR had renamed the option to `S3A_AWS_CREDENTIALS_PROVIDER_CLASS` and flattened the enum into class-name constants — both touched the same lines.

Resolution keeps `dev`'s shared-constant structure but points it at this PR's new API:

```java
private static final String TEST_S3_CREDENTIALS_PROVIDER =
        S3FileBaseOptions.SIMPLE_AWS_CREDENTIALS_PROVIDER;
// both configs now use S3A_AWS_CREDENTIALS_PROVIDER_CLASS.key()
```

The two `S3File.md` doc conflicts auto-merged cleanly. Verified locally: `spotless:check` clean and `S3StorageTest` green (2/2).

DanielLeens (commented):

The previous source concern is resolved on `3f3f86c1fa53`: legacy provider constants remain available while custom providers are validated and documented for every worker.

Queue note: this head is `diverged` against `dev` (ahead_by=6, behind_by=4). I am recording that fact separately; it is not by itself a request to sync unless the CI guidance above explicitly says so.

davidzollo (approved):
+1
LGTM

PR #11438, opened 2026-07-15
[Fix][E2E] Replace flaky waits in E2E tests

nzw921rx:

Fix several PR-unrelated flaky E2E waits by replacing fixed sleeps and delayed polling with condition-based waits.

This patch updates:

- RocketMQ E2E: replace a fixed 20s sleep after consumer commit with an Awaitility wait for consumed offsets to become visible from RocketMQ admin APIs, and always shut down the test consumer in finally.
- Kafka E2E: remove fixed 60s/120s initial polling delays from exactly-once streaming assertions. The existing Awaitility assertions already wait for the expected records.
- Doris E2E: remove a fixed 20s schema-change startup sleep by letting the initial table assertion retry through startup-time query failures; replace the error-path startup sleep with a condition that ensures the async job remains running before stopping Doris.
- SFTP E2E: replace the non-recursive scan sleep with an Awaitility window asserting the nested file remains absent.

No.

Ran:

```
./mvnw spotless:apply
```

No E2E tests were run.


nzw921rx:
The repair scope is too wide and a new PR has been opened

PR #11440, opened 2026-07-15
[Fix][CI] Avoid flaky approve label workflow failures

nzw921rx:

Fix flaky failures in the `Label when approved` workflow by avoiding a best-effort label automation failure turning CI red.

**Root cause:**

The workflow currently runs on the `requested` event of the `Label-when-reviewed` workflow and immediately calls the GitHub Actions API to resolve the source run:

```
Getting workflow id for source run id: 29387900060, owner: apache, repo: seatunnel
Error: Not Found
```

This source-run lookup can return `404 Not Found` when the workflow run is not yet visible or cannot be resolved by the token at that moment. The failure is unrelated to the PR code but marks the CI job as failed.

**Changes:**
- Trigger the label workflow after the source workflow is `completed` instead of `requested`
- Allow the source-run lookup step to fail without failing the whole workflow
- Skip label update steps when no pull request number is resolved

No. This only affects GitHub Actions label automation stability.

- `./mvnw spotless:apply`

### DIFF
--- a/.github/workflows/approve-label.yml
+++ b/.github/workflows/approve-label.yml
@@ -20,7 +20,7 @@ name: "Label when approved workflow run"
 on:
   workflow_run:
     workflows: [Label-when-reviewed]
-    types: [requested]
+    types: [completed]
 permissions:
   # All other permissions are set to none
   checks: write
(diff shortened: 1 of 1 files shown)


nzw921rx:
Closing this review because the concern is not a blocker for this PR and can be addressed separately if needed.

PR #11441, opened 2026-07-15
[Fix][E2E] Fix flaky HTTP E2E setup and binary job

nzw921rx:

Fix flaky failures in the HTTP E2E module by removing an external runtime driver download and making the binary HTTP job assertion resilient to transient Spark launcher exit-code noise.

**Root cause:**

The CI failure in `all-connectors-it-1` shows `HttpIT.testSourceToAssertSink` failing on the last binary HTTP job:

```
HttpIT.testSourceToAssertSink:371 expected: <0> but was: <1>
```

The job output showed the bounded HTTP source and Spark write path completing, but the container command returned a non-zero exit code. This test also downloaded the PostgreSQL JDBC driver from Maven Central inside the engine container at runtime, which adds another CI network dependency unrelated to the PR code.

**Changes:**
- Resolve the PostgreSQL JDBC driver from the active test classpath and copy it into the container
- Remove the runtime `curl` to Maven Central from `HttpIT`
- Retry the idempotent `http_binary_to_assert.conf` job once when the first command returns a non-zero exit code
- Include stderr in the final assertion message

No. This change only affects E2E test stability.

- `./mvnw spotless:apply`


nzw921rx:

PR #11442, opened 2026-07-15
[Fix][E2E] Fix flaky Kafka latest and partition tests

nzw921rx:

Fix flaky tests in the Kafka E2E module by making startup and partition-discovery assertions wait for actual consumption readiness instead of assuming the source is ready immediately after the job reaches RUNNING.

**Root cause:**

The Kafka CI job intermittently fails with `ConditionTimeout` in tests that use `start_mode = latest`:

```
KafkaIT.testSourceKafkaRestoreWithLatestMode
KafkaIT.testDynamicPartitionDiscovery
```

For latest-mode restore, records produced immediately after `awaitJobRunning` can be skipped if the Kafka source has not finished subscribing and seeking yet.

For dynamic partition discovery, records sent to the new partition immediately after partition creation can be skipped before SeaTunnel discovers the new partition and starts consuming it.

**Changes in `KafkaIT`:**
- Add a probe record loop before the latest-mode assertions to confirm the running job is consuming newly produced records
- Count first-run and restore records by their payload instead of relying on the full sink partition offset count
- Add a probe record loop after increasing Kafka partitions to confirm SeaTunnel has discovered and consumed the new partition
- Assert the dynamic-partition test against `new_partition_message_` records only, so probe records do not affect the result

No. This change only affects E2E test stability.

Ran:

```
./mvnw spotless:apply
```

Per the flaky-CI fix workflow, no E2E tests were run locally.


nzw921rx:

PR #11445, opened 2026-07-15
[Fix][E2E] Fix flaky engine cluster cancel wait

nzw921rx:

Fix a flaky engine E2E failure in `ClusterFaultToleranceIT.testStreamJobRestoreInAllNodeDown`.

The failed CI log shows the recovered streaming job reached the cancel phase, but the test only allowed 60 seconds for the final state transition:

```
expected: <CANCELED> but was: <CANCELING> within 1 minutes
```

This is unrelated to PR code changes and can happen after the all-node-down restore path while job cleanup is still in progress.

### Changes

- Align the cancel wait timeout in `testStreamJobRestoreInAllNodeDown` with the other all-node-down restore scenario in the same test class.
- Keep the final `CANCELED` and `waitForJobComplete` assertions unchanged.

No.

- `./mvnw spotless:apply`


nzw921rx:

PR #11447, opened 2026-07-15
[Fix][E2E] Replace flaky Doris waits

nzw921rx:

Fix flaky Doris E2E waits that depended on fixed sleeps while the Doris sink job and schema propagation are asynchronous.

### Changes

- Wait for the Doris error-path job to remain running before stopping Doris, instead of sleeping for a fixed 10 seconds.
- Remove the fixed sleep before schema-evolution assertions.
- Ignore transient assertion-time exceptions while polling Doris schema metadata.

No.

- ./mvnw spotless:apply


nzw921rx:

PR #11448, opened 2026-07-15
[Fix][E2E] Replace flaky SFTP continuous wait

nzw921rx:

Fix a flaky SFTP E2E assertion that used a fixed sleep before checking that nested files are not synchronized by the continuous job.

### Changes

- Replace the fixed sleep with an Awaitility assertion that requires the nested file to remain absent for the expected duration.
- Keep the change limited to the SFTP E2E test.

No.

- ./mvnw spotless:apply


nzw921rx:

PR #11459, opened 2026-07-15
[Fix][Connector-V2][Zeta] Resolve Hive S3, JDBC, metrics, and documentation CI failures

nzw921rx:
### Purpose

Fix the deterministic CI failures affecting this PR after syncing with the latest `dev` branch.

The original failure was a Windows unit-test classpath issue in `S3StorageTest`:

`The S3A credentials provider class 'com.amazonaws.auth.InstanceProfileCredentialsProvider' is not found on the classpath`.

After the branch was updated, the same CI run exposed three additional blockers:

- `Dependency licenses` could not reach the license check because the full reactor failed to compile on a duplicate JDBC converter method.
- `MetricsApiTest` returned HTTP 500 because a previous test registered Prometheus collectors against a short-lived Hazelcast node and left stale collectors in the JVM-wide registry after shutdown.
- The dead-link check treated a concrete Alibaba Cloud Tablestore JDBC example endpoint as a real public URL and received HTTP 400.

### Changes

- Use Hadoop S3A's `SimpleAWSCredentialsProvider` in `S3StorageTest` instead of the AWS SDK v1 instance-profile provider.
- Remove the duplicate `setNullToStatementByDataType(...)` implementation from `AbstractJdbcRowConverter`, retaining the existing documented implementation.
- Keep global telemetry disabled in `EngineStateStoreLogicalMetricExportsTest`, which tests the collector directly and does not need to register JVM-wide exporters.
- Revert the ineffective 60-second retry in `MetricsApiTest`; the endpoint test remains a direct assertion of the complete HTTP metrics path.
- Replace the concrete Tablestore endpoint in the English and Chinese JDBC documentation with an explicit placeholder:
  `jdbc:ots:https://<instance_name>.<region_id>.ots.aliyuncs.com/<instance_name>`.

### User-facing behavior

No runtime, configuration, or API behavior changes.

The only documentation-visible change is that the Tablestore JDBC URL now clearly uses instance and region placeholders instead of a non-existent concrete endpoint.

### How tested

- `./mvnw spotless:apply`
- Verified that the JDBC module compiles after removing the duplicate method.
- Verified that the Tablestore placeholder is no longer parsed as an external Markdown link.
- Reproduced and traced the Metrics API failure to a stale Prometheus collector bound to a shut-down Hazelcast node.


DanielLeens (commented):
Thanks for the fix. I rechecked the latest head `cb86da40056e` against the current Hive-on-S3 path.

What this PR fixes
- User pain: `S3StorageTest` was failing in CI because its fixture depended on `com.amazonaws.auth.InstanceProfileCredentialsProvider` being present on the unit-test classpath, even though this test only verifies the Hive S3 HadoopConf wiring.
- Fix approach: switch the test fixture to `SimpleAWSCredentialsProvider`, which still exercises `S3Storage.buildHadoopConfWithReadOnlyConfig(...) -> HiveOnS3Conf.buildWithReadOnlyConfig(...) -> S3HadoopConf.buildWithReadOnlyConfig(...)` without adding that unrelated AWS SDK v1 classpath assumption.
- One-line summary: this keeps the test focused on Hive S3 config assembly rather than external provider availability.

I do not see a source-level blocker on the current diff.

The remaining visible gate from my side is that GitHub still shows `Pull Request Labeler` running, and the normal reviewer approval gate is not closed yet. Since I only have read-level review controls on this repository, a write-capable maintainer may still need to complete the final approval step once the checks settle.

DanielLeens (commented):
Thanks for the update. I re-reviewed the full current diff. The Hive S3 credential-provider fixture change is coherent, but one scope/integrity blocker remains.

## Blocking finding

**Issue 1: the PR includes an unrelated and ineffective `MetricsApiTest` change**

- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/rest/MetricsApiTest.java:55-69`
- This engine-test hunk has no call-chain relationship to the Hive S3 credential-provider fixture. The current CI also shows that the added 60-second Awaitility loop does not fix the failure: the endpoint stays HTTP 500 until timeout.
- Please remove/split this hunk and diagnose the engine test in its owning change. Keeping it here couples an unrelated baseline failure to this connector E2E update without resolving it.

Severity: Medium. Merge gate: blocking because the complete diff is not focused and the unrelated change is not validated. Happy to re-review once the branch is focused.

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current diff at `965025d190b1c6f9f358657cd048a70e724cdfb2`. The Hive S3 fixture now correctly uses `SimpleAWSCredentialsProvider`, but the same diff-integrity blocker remains.

**The PR still contains an unrelated `MetricsApiTest` retry change.** The hunk in `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/metrics/MetricsApiTest.java:51-69` has no call-chain or test-fixture relationship to Hive S3 credential-provider class loading. It also changes an engine REST test from a direct assertion to a 60-second retry, which can mask startup/readiness behavior without establishing that it fixes this PR's Hive test failure. Please remove this hunk and keep the PR scoped to `S3StorageTest`, or move the metrics change to a separately justified PR with its own failure evidence.

The required `Build` check for this head is currently queued. Independently of CI, `diff_integrity` remains failing and the technical verdict is **blocking** until the unrelated hunk is removed or justified with same-scope evidence.

dybyte (approved):
+1 if CI passes. Thanks for the contribution.

For unrelated CI fixes, separate PRs might make the changes easier to review and merge.

PR #11464, opened 2026-07-16
[Fix][Zeta] Fix file appender log path resolution

### LINKED ISSUE #11463: [Bug] [Zeta] LogUtil.getLogPath uses wrong variable in fileAppender branch (NPE / StringIndexOutOfBounds)
### Search before asking

### What happened

```java
}
```

<details>

```java
public class LogUtilRepro {
        } else {
        }
    }

    }
}
```

Output on OpenJDK 1.8.0_462:

```
A: NullPointerException
```

</details>

```java
}
```

### SeaTunnel Version

### SeaTunnel Config

### Running Command

### Error Exception

```log

```

Zeta (SeaTunnel Engine).

### Java or Scala Version

OpenJDK 1.8.0_462 (Temurin).

### Screenshots

_No response_

### Code of Conduct

nielifeng:

Fixes #11463.

This PR fixes `LogUtil.getLogPath()` using the routing appender path to calculate the substring index for the file appender path. With a file-appender-only Log4j2 configuration, the routing path is `null` and log REST endpoints fail with a `NullPointerException`.

It also adds a regression test using a file-appender-only configuration.

Yes. Zeta log path resolution now works when only `fileAppender` is configured, so log REST endpoints can access the configured log directory instead of failing with a `NullPointerException`.

Added `LogUtilTest.shouldGetLogPathFromFileAppender` and ran:

```shell
./mvnw -o -pl seatunnel-engine/seatunnel-engine-common -Dtest=LogUtilTest -Dskip.spotless.check=true -Dcheckstyle.skip=true test
```

The test fails with a `NullPointerException` on the previous implementation and passes with this fix.

### Check list


DanielLeens (commented):
Thanks for the fix. I rechecked the latest head `66d1f029e8e2` against the actual `fileAppender` branch in `LogUtil.getLogPath()`, and I do not see a source-level blocker on the current diff.

The production change is narrow and matches the new test:
- `LogUtil.java:49-53` now uses `fileLogPath.lastIndexOf("/")` for the `fileAppender` branch instead of slicing with the routing path
- `LogUtilTest` plus `log4j2-file-only.properties` directly exercise the file-only configuration that was broken before

The remaining gate is CI rather than source logic. Right now the branch is already on the latest `dev` (`ahead_by=1`, `behind_by=0`), so syncing upstream will not change the base here. The current failing jobs are:
- `engine-v2-it (11, ubuntu-latest)`
- `kafka-connector-it (8, ubuntu-latest)`

Those failures do not look tied to this LogUtil-only diff yet, so the lowest-cost next step is to rerun those lanes / refresh the Build check on the same head. If they still fail after a rerun, please share the fresh run link and I can help narrow them down further.

nzw921rx (approved):
+1 LGTM if CI pass

DanielLeens (commented):
Rechecked the unchanged head. My earlier note about failing Build lanes is now outdated: the aggregate `Build` is currently cancelled, not green. I still do not see a source-level blocker in this LogUtil-only diff, but this head needs a completed Build result. Please rerun the Build check; if it fails again, share the fresh run link and I can help narrow it down.

nielifeng:
Rebuilt the change on current upstream `dev` at head `5bf73c3ba`. `LogUtilTest`: 1 test passed; engine-common module `-DskipTests verify` passed. A fresh Build should now run on the updated head.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `5bf73c3bace3` from scratch.

# What this PR fixes
- User pain: With a file-appender-only Log4j2 config, Zeta's log REST endpoints throw a `NullPointerException` because `LogUtil.getLogPath()` computed the substring index from the `routingAppender` path (`routingLogFilePath`), which is `null` in that setup (#11463).
- Fix approach: One-line correction in `LogUtil.java` — the `fileAppender` branch now uses `fileLogPath.lastIndexOf("/")` instead of `routingLogFilePath.lastIndexOf("/")`, plus a regression test (`LogUtilTest.shouldGetLogPathFromFileAppender`) that swaps in a file-only `log4j2-file-only.properties` and asserts `target/logs` is resolved.
- One-line summary: The fix itself is correct and minimal and the regression test demonstrably fails on the old code, but the test mutates the shared global `LoggerContext` without full restoration, the fixed line can still throw for path-less or Windows-style `fileName` values, and the `routingAppender` branch remains untested.

# Runtime chain I rechecked
```text
Zeta REST log endpoints (GET current/all node logs)
  RestHttpGetCommandProcessor / LogService -> LogUtil.getLogPath()  (seatunnel-engine-server)
    LogUtil.getLogPath()  LogUtil.java:~38
      reads Log4j2 Configuration appender fields via reflection (routingLogFilePath may be null when only fileAppender is configured)  LogUtil.java:~40-47
      logRef.equals(routingAppender) branch -> routingLogFilePath.substring(...)  LogUtil.java:49-50 (unchanged)
      logRef.equals(fileAppender) branch -> fileLogPath.substring(0, fileLogPath.lastIndexOf("/"))  LogUtil.java:52 (fixed line; previously indexed with routingLogFilePath.lastIndexOf("/"), NPE when routing appender absent)
```

# Findings
**Issue 1: Test mutates the shared global LoggerContext and does not fully restore it, which can leak the file-only config into other tests in the module**
- Location: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:45`
- Why it matters: - Root cause: `LogUtilTest` calls `context.setConfigLocation(configuration.toURI())` on the JVM-wide `LoggerContext` obtained via `LogManager.getContext(false)`, then restores only via `context.setConfiguration(originalConfiguration)` in the `finally` block. The config *location* of the context still points at `log4j2-file-only.properties`, so any subsequent `reconfigure()` (triggered by another test, a shutdown hook, or a monitor interval) will reload the file-only configuration instead of the original one.
- Impact: other tests running in the same surefire JVM in `seatunnel-engine-common` can silently pick up the file-only logging config, producing flaky or hard-to-diagnose test interactions; the `File` appender will also keep writing to `target/logs/seatunnel.log` after the test if the restore is incomplete.
- Fix: capture and restore the original config location as well (`URI originalLocation = context.getConfigLocation(); ... context.setConfigLocation(originalLocation);`), or better, drive the test through a dedicated `LoggerContext`/`Configurator.initialize(...)` instance so the shared context is never touched.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:42  context.setConfigLocation(configuration.toURI());`; `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:45  context.setConfiguration(originalConfiguration);`
- Suggested fix: Restoring only the `Configuration` here leaves the context's config *location* still pointing at `log4j2-file-only.properties`, so any later `reconfigure()` in this JVM will reload the file-only config and can affect other tests in this module. Please also restore the original config location (e.g. capture `context.getConfigLocation()` before the change and reset it in `finally`), or use an isolated `LoggerContext` instead of the shared `LogManager.getContext(false)` one.
- Severity: Medium

**Issue 2: Fixed line still throws StringIndexOutOfBoundsException/NPE for path-less or Windows-style fileName values**
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:52`
- Why it matters: - The one-line fix at LogUtil.java:52 is correct for the reported NPE (#11463), but the new expression `fileLogPath.substring(0, fileLogPath.lastIndexOf("/"))` is still fragile.
- If a user configures the file appender with a bare file name (`appender.file.fileName = seatunnel.log`) `lastIndexOf("/")` returns -1 and `substring(0, -1)` throws `StringIndexOutOfBoundsException`, so the log REST endpoints (`/logs`) fail again — just with a different exception than the one this PR fixes.
- The same applies on Windows deployments where operators commonly write `logs\seatunnel.log`; there is no `/` in the string at all.
- Symmetrically, if the `fileAppender` ref resolves but its file path is `null` (misconfigured appender type), `fileLogPath.lastIndexOf` throws an NPE analogous to the one being fixed here.
- Fix: resolve the parent directory with `java.nio.file.Paths.get(fileLogPath).getParent()` (falling back to `.` when there is no parent) and add an explicit null check with a descriptive `IllegalArgumentException`, mirroring the existing `else` branch message style.
- Suggested fix: This fixes the NPE from #11463, but `fileLogPath.lastIndexOf("/")` still breaks for two realistic configs: a bare file name (`fileName = seatunnel.log` → `substring(0, -1)` throws `StringIndexOutOfBoundsException`) and Windows-style separators (`logs\seatunnel.log` has no `/`). Consider `Paths.get(fileLogPath).getParent()` with a null-parent fallback, plus a null check on `fileLogPath` with a descriptive `IllegalArgumentException` like the existing `else` branch.
- Severity: Medium

**Issue 3: Regression test only covers the fileAppender branch; the routingAppender branch of getLogPath() remains untested and could regress the same way**
- Location: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:35`
- Why it matters: - Root cause: the bug fixed here (`routingLogFilePath.lastIndexOf("/")` used inside the `fileAppender` branch) was a copy-paste across the two branches of `LogUtil.getLogPath()`. The new test `shouldGetLogPathFromFileAppender` exercises only the `fileAppender` branch; there is no test asserting the `routingAppender` path or the `IllegalArgumentException` fallback branch.
- Impact: a symmetric mistake in the routing branch (e.g. accidentally referencing `fileLogPath` there in a future refactor) would not be caught, and the Zeta REST log endpoints would break again for the default routing-appender configuration, which is the config most users run in production.
- Fix: add a second test with a routing-appender-based `log4j2` test resource (mirroring the default `log4j2.properties` shipped in `config/`) asserting `getLogPath()` returns the routing directory, and optionally a third test asserting the `IllegalArgumentException` message when neither appender resolves.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:43  assertEquals("target/logs", LogUtil.getLogPath());`; `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:52  return fileLogPath.substring(0, fileLogPath.lastIndexOf("/"));`
- Suggested fix: Nice regression test for the fileAppender branch — the fix itself in `LogUtil.java:52` looks correct. Since `getLogPath()` has two symmetric branches and the original bug was a cross-branch copy-paste, could you also add a companion test using a routing-appender config (matching the default `config/log4j2.properties`) so both branches are pinned? That protects the default production path from the mirror-image regression.
- Severity: Medium

**Issue 4: Test config redirects the module's root logger to a real file under target/ while active**
- Location: `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-file-only.properties:27`
- Why it matters: - While `log4j2-file-only.properties` is installed, `rootLogger.appenderRef.file.ref = fileAppender` (log4j2-file-only.properties:23) routes all logging emitted by the JVM into `target/logs/seatunnel.log` (log4j2-file-only.properties:27), a relative path resolved against the surefire working directory.
- Any log statements emitted by other components during the swap window land in that file instead of the console, and the file persists across runs, which can be confusing when diagnosing CI output for this module.
- This is acceptable for a focused unit test, but the pattern `%m%n` plus a real `File` appender means Log4j2 actually creates the directory and file on config activation.
- Fix (optional): point the appender at `${sys:java.io.tmpdir}` or use a `Null`/in-memory-style appender name with a `fileName` under `target/` documented via a comment, and keep the assertion on the configured string as it is now — `LogUtil.getLogPath()` returns the raw configured path so the assertion itself is stable.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-file-only.properties:27  appender.file.fileName = target/logs/seatunnel.log`
- Suggested fix: Minor: activating this config makes Log4j2 create `target/logs/seatunnel.log` relative to the surefire working directory and briefly diverts all root logging into it. That's fine for this unit test, but a short comment noting the file is intentionally created under `target/` (so it's cleaned by `mvn clean`) would help future readers; alternatively parameterize the directory via a system property.
- Severity: Low

**Issue 5: ASF license header in log4j2-file-only.properties has inconsistent indentation on the last line**
- Location: `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-file-only.properties:16`
- Why it matters: - Root cause: the header lines in the new test resource all use two spaces after `#` (e.g. `#  Unless required by applicable law...`), but the final line reads `# limitations under the License.` with a single space, diverging from the canonical ASF header block used across other SeaTunnel `.properties` files.
- Impact: cosmetic only — the header text is legally complete, so this is not a licensing problem, just an inconsistency that could trip strict header-diff tooling or copy-paste propagation.
- Fix: change the last header line to `#  limitations under the License.` (two spaces) to match the rest of the block and the headers in existing resources like `config/log4j2.properties`.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-file-only.properties:15  #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.`
- Suggested fix: Nit: this line uses `# limitations...` (one space) while the rest of the ASF header block uses `#  ` (two spaces). Please align it with the standard header used in other `.properties` files, e.g. `#  limitations under the License.`
- Severity: Low

**Issue 6: Static assertion imports deviate from the prevailing SeaTunnel test convention of fully-qualified Assertions calls**
- Location: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:29`
- Why it matters: - Root cause: the new test uses `import static org.junit.jupiter.api.Assertions.assertEquals;` / `assertNotNull`, while the vast majority of SeaTunnel test code (e.g. across seatunnel-engine-server tests such as `JobMasterTest`/`CoordinatorServicePipelineCleanupTest`) invokes assertions as `Assertions.assertEquals(...)` without static imports.
- Impact: purely cosmetic inconsistency with surrounding code; no functional effect.
- Fix: replace the static imports with `import org.junit.jupiter.api.Assertions;` and call `Assertions.assertEquals("target/logs", LogUtil.getLogPath());` and `Assertions.assertNotNull(configuration);` to match the module's style.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:30  import static org.junit.jupiter.api.Assertions.assertNotNull;`
- Suggested fix: Nit: most SeaTunnel tests use `Assertions.assertEquals(...)` via a plain `import org.junit.jupiter.api.Assertions;` rather than static imports. Please align with that convention for consistency with the rest of the engine test code.
- Severity: Low

# Review conclusion

- Issue 1: Test mutates the shared global LoggerContext and does not fully restore it, which can leak the file-only config into other tests in the module
- Issue 2: Fixed line still throws StringIndexOutOfBoundsException/NPE for path-less or Windows-style fileName values
- Issue 3: Regression test only covers the fileAppender branch; the routingAppender branch of getLogPath() remains untested and could regress the same way
- Issue 4: Test config redirects the module's root logger to a real file under target/ while active
- Issue 5: ASF license header in log4j2-file-only.properties has inconsistent indentation on the last line
- Issue 6: Static assertion imports deviate from the prevailing SeaTunnel test convention of fully-qualified Assertions calls

nielifeng:
Addressed the latest review feedback in commit `34b62c41a`:

- isolated `LogUtilTest` from the shared global `LoggerContext`;
- resolved parent directories with `Paths.get(...).getParent()`, including a `.` fallback for path-less file names and a descriptive null/blank-path error;
- added routing-appender and empty/path-less path coverage;
- moved the file-only test output under the JVM temp directory;
- aligned the ASF header and assertion style.

Validation:
- `LogUtilTest`: 4 tests passed;
- `seatunnel-engine-common -DskipTests verify`: passed.

DanielLeens (commented):

The previous source concern is resolved on `34b62c41a7c6`: the test now uses and closes a dedicated `LoggerContext` instead of mutating the global context.

Queue note: this head is `diverged` against `dev` (ahead_by=2, behind_by=4). I am recording that fact separately; it is not by itself a request to sync unless the CI guidance above explicitly says so.

nielifeng:
Addressed the Windows unit-test failure in head d6a7bcc69.

Root cause: Paths.get parsed the unresolved Log4j variable in job-${ctx:ST-JID}.log as a Windows path, where the colon is illegal. LogUtil now derives the parent lexically from both slash styles, so unresolved lookup expressions are never passed to the host filesystem parser. Windows drive roots, Unix roots, path-less names, repeated separators, and UNC prefixes remain handled.

Validation:
- LogUtilTest: 5/5 passed
- Windows-style unresolved-variable regression added
- Spotless passed
- seatunnel-engine-common skip-tests verify passed

Please re-review the new head after CI completes.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `34b62c41a7c6` from scratch.

# What this PR fixes
- User pain: With a fileAppender-only Log4j2 config, Zeta's log REST endpoints throw a `NullPointerException` because `LogUtil.getLogPath()` computed the fileAppender's parent directory using `routingLogFilePath.lastIndexOf("/")`, and the routing path is `null` in that setup (#11463).
- Fix approach: Extract a package-private `getLogPath(PropertiesConfiguration)` overload plus a `getParentLogPath(logFilePath, appenderName)` helper that uses `Paths.get(path).getParent()` on the correct appender's own path, blank-checks the input with a descriptive `IllegalArgumentException`, and falls back to `"."` for a path-less file name; regression coverage is added via `LogUtilTest` with two new test-only Log4j2 properties configs (file-only and routing-only).
- One-line summary: The fix is correct and well-targeted — the copy-paste bug on the `fileAppender` branch is clearly resolved and the routing branch behavior is preserved — with the remaining findings being test portability (hardcoded `target/routing-logs` separator on Windows), tmpdir cleanup, a silent `"."` fallback, and minor docs/annotation polish rather than blockers.

# Runtime chain I rechecked
```text
REST log endpoint (e.g. GET /logs handler in seatunnel-engine-server)
  -> LogUtil.getLogPath()  LogUtil.java:36
    -> LogUtil.getLogConfiguration()  LogUtil.java:73 (LoggerContext -> PropertiesConfiguration)
    -> LogUtil.getLogPath(PropertiesConfiguration config)  LogUtil.java:40
      -> getRoutingLogFilePath(config)  (may return null when no routingAppender is configured)
      -> getFileLogPath(config)
      -> config.getRootLogger().getAppenderRefs() -> first ref matching "routingAppender"/"fileAppender"  LogUtil.java:48-54
      -> [logRef == "routingAppender"] getParentLogPath(routingLogFilePath, "routingAppender")  LogUtil.java:56
      -> [logRef == "fileAppender"]    getParentLogPath(fileLogPath, "fileAppender")  LogUtil.java:58  (previously indexed with routingLogFilePath -> NPE)
        -> StringUtils.isBlank(logFilePath) -> IllegalArgumentException  LogUtil.java:64-67
        -> Paths.get(logFilePath).getParent()  LogUtil.java:68
        -> parent == null ? "." : parent.toString()  LogUtil.java:69
```

# Findings
**Issue 1: Routing-appender test hardcodes a forward-slash path and will fail on Windows**
- Location: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:42`
- Why it matters: - Root cause: `LogUtil.getParentLogPath` now returns `Paths.get(logFilePath).getParent().toString()`, which renders the platform-specific separator (backslash on Windows), but `shouldGetLogPathFromRoutingAppender` asserts the literal `"target/routing-logs"`.
- Impact: the regression test is not portable — it will fail on any Windows developer machine or Windows CI runner, which weakens the value of this new safety net for Zeta log path resolution.
- Fix: build the expected value with `Paths.get("target", "routing-logs").toString()` exactly like `shouldGetLogPathFromFileAppender` already does at LogUtilTest.java:34-36, so both tests are OS-agnostic.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:71          return parent == null ? "." : parent.toString();`
- Suggested fix: Since `getParentLogPath` now goes through `Paths.get(...).getParent().toString()`, the result uses the platform separator, so this hardcoded `"target/routing-logs"` will fail on Windows (it would be `target\routing-logs`). Please build the expected value the same way as in `shouldGetLogPathFromFileAppender`, e.g. `Paths.get("target", "routing-logs").toString()`.
- Severity: Medium

**Issue 2: File-only test config writes into java.io.tmpdir without cleanup**
- Location: `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-file-only.properties:27`
- Why it matters: - Root cause: `log4j2-file-only.properties` points the File appender at `${sys:java.io.tmpdir}/seatunnel-logutil-test/seatunnel.log`, and starting the `LoggerContext` in `assertLogPath` creates that directory and file eagerly; the test never deletes them.
- Impact: repeated local and CI runs leave stale `seatunnel-logutil-test` artifacts in the system temp directory, and using a shared global tmpdir path can collide between concurrently running builds on the same agent.
- Fix: prefer JUnit 5's `@TempDir` combined with a system property (e.g. set `-Dseatunnel.test.log.dir` and reference `${sys:seatunnel.test.log.dir}` in the properties file), or at minimum delete the directory in an `@AfterEach`/finally block; alternatively point the appender at `target/` like the routing config does so `mvn clean` removes it.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:34-36                  Paths.get(System.getProperty("java.io.tmpdir"), "seatunnel-logutil-test")`
- Suggested fix: Starting the LoggerContext with this config eagerly creates `${java.io.tmpdir}/seatunnel-logutil-test/seatunnel.log`, and the test never cleans it up, so stale files accumulate in the shared temp dir across runs. Consider pointing this at `target/` like `log4j2-routing-only.properties` does (so `mvn clean` handles it), or delete the directory in a `finally`/`@AfterEach` block.
- Severity: Low

**Issue 3: User-facing behavior change (fileAppender-only Log4j2 configs now supported by log REST endpoints) is not reflected in the logging docs**
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:58`
- Why it matters: - Root cause: the PR description states this is a user-facing change (log REST endpoints now work with a `fileAppender`-only configuration), but no file under `docs/` is touched in the diff.
- Impact: users reading `docs/en/seatunnel-engine/logging.md` (and the zh counterpart) may still believe the routing appender is required for the log REST endpoints to resolve the log directory, and there is no release-note trail for operators who previously worked around the NPE described in #11463.
- Fix: add a short note to `docs/en/seatunnel-engine/logging.md` and `docs/zh/seatunnel-engine/logging.md` stating that either `routingAppender` or `fileAppender` alone is sufficient for the log REST endpoints since this fix; no `incompatible-changes.md` entry is needed as the change is backward compatible.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:58              return getParentLogPath(fileLogPath, fileAppender);`
- Suggested fix: Now that a `fileAppender`-only Log4j2 configuration is fully supported by the log REST endpoints, please add a brief note to `docs/en/seatunnel-engine/logging.md` (and the zh version) so users know the routing appender is no longer a hard requirement for log path resolution.
- Severity: Low

**Issue 4: New package-private helpers in LogUtil lack Javadoc and a @VisibleForTesting marker**
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:65`
- Why it matters: - Root cause: `getLogPath(PropertiesConfiguration)` and `getParentLogPath(String, String)` were introduced with package-private visibility purely so the test can drive them, but neither carries Javadoc nor an annotation explaining why they are not `private`.
- Impact: future maintainers of seatunnel-engine-common may not realize the widened visibility is test-only and could start calling these helpers from production paths, and the non-obvious `"."` fallback contract of `getParentLogPath` is undocumented.
- Fix: add short Javadoc to both methods (including the blank-path `IllegalArgumentException` and the `"."` return for path-less file names) and annotate them with `@VisibleForTesting` as is done elsewhere in the codebase, per SeaTunnel code conventions.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:65      static String getParentLogPath(String logFilePath, String appenderName) {`
- Suggested fix: These two package-private methods exist so the test can inject a `PropertiesConfiguration`, but that intent isn't visible in the code. Please add `@VisibleForTesting` plus a short Javadoc on `getParentLogPath` documenting the contract: throws `IllegalArgumentException` on a blank path and returns `"."` when the file name has no parent.
- Severity: Low

**Issue 5: Silent "." fallback for a path-less file appender should emit a warning log**
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:71`
- Why it matters: - Root cause: `getParentLogPath` returns `"."` when `Paths.get(logFilePath).getParent()` is null (e.g. `appender.file.fileName = seatunnel.log`), with no diagnostic output.
- Impact: the Zeta log REST endpoints will then list files from the process working directory, which can silently expose an unexpected directory listing and makes misconfiguration hard to diagnose — the operator sees wrong log content with no hint why.
- Fix: keep the fallback (it is better than the previous `StringIndexOutOfBounds`/NPE behavior), but add a `log.warn` stating that the appender's `fileName` has no directory component and the working directory is being used, so operators can correct their log4j2 configuration.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:71          return parent == null ? "." : parent.toString();`
- Suggested fix: Falling back to `"."` is reasonable, but it happens silently. When `getParent()` is null the REST log endpoints will serve the process working directory, which is hard to debug from the outside. Please add a `warn` log here noting that the `fileName` of the given appender has no directory component and the current working directory is being used instead.
- Severity: Low

# Review conclusion

- Issue 1: Routing-appender test hardcodes a forward-slash path and will fail on Windows
- Issue 2: File-only test config writes into java.io.tmpdir without cleanup
- Issue 3: User-facing behavior change (fileAppender-only Log4j2 configs now supported by log REST endpoints) is not reflected in the logging docs
- Issue 4: New package-private helpers in LogUtil lack Javadoc and a @VisibleForTesting marker
- Issue 5: Silent "." fallback for a path-less file appender should emit a warning log

nielifeng:
Addressed the Windows CI regression in new head `7a6412c6a`.

Root cause from the Windows job: `${sys:java.io.tmpdir}` expands with backslashes, while the test resource appends `/seatunnel-logutil-test`; `LogUtil` returned a mixed path such as `C:\\...\\Temp\\/seatunnel-logutil-test`, while the expected Windows path uses backslashes.

The lexical parent resolver now normalizes mixed separators consistently for drive-letter and UNC paths while preserving Unix and relative paths. Added a mixed Windows separator regression test.

Validation:
- `LogUtilTest`: 6/6 passed
- `seatunnel-engine-common -DskipTests verify`: passed

The previous Windows failure and the shared Dead links failure were both diagnosed from CI logs; this head contains only the deterministic LogUtil fix. Please re-review after the new Build completes.

nielifeng:
Follow-up CI fix pushed in `fe9d43f14`: update the English and Chinese StarRocks CREATE TABLE links to the current `table_bucket_part_index` path. The previous Build failed only because the old path returned HTTP 404; the Windows LogUtil path test fix remains in `7a6412c6a`.

nielifeng:
The `fe9d43f14` Build attempt 1 was cancelled by Actions scheduling after the initial checks, including `Dead links`, passed. It had no failed test job. I restarted the Build as attempt 2 to obtain a complete CI result.

DanielLeens (commented):
Outdated.

nielifeng:
Removed the unrelated StarRocks documentation hunks from this LogUtil branch and synced the branch with current `dev`. The PR diff is now scoped back to the Zeta log-path fix. A new Build is queued on head `7d1fff28`.

DanielLeens (commented):
Outdated.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `982784db56526d623201b9ae089033683c39737e` from the full diff and the Zeta log-path call chain.

I also noticed earlier review activity on this PR. After tracing the current head myself, I agree with the source-level direction: this is a focused fix for `LogUtil` when the active Log4j configuration uses `fileAppender` instead of `routingAppender`.

Runtime path checked:

```text
Log viewing code asks for the Zeta log directory
  -> LogUtil.getLogPath()
      -> getLogConfiguration()
          -> reads the current Log4j PropertiesConfiguration from LoggerContext
      -> getLogPath(PropertiesConfiguration)
          -> getRoutingLogFilePath(config)
              -> resolves the nested routingAppender fileName
          -> getFileLogPath(config)
              -> resolves the top-level fileAppender fileName
          -> checks the root logger appender refs
              -> routingAppender: return parent path of the routing fileName
              -> fileAppender: return parent path of the file fileName
          -> getParentLogPath(path, appenderName)
              -> rejects blank paths with appender context
              -> returns "." for pathless file names
              -> handles Unix, Windows drive, and mixed separators
```

Source review result:

- I do not see a new code-side blocker in this PR.
- The change keeps the existing public entry point and does not alter the logging configuration contract, REST/RPC protocol, serialization format, or default log behavior.
- The added unit coverage directly exercises the important branches: file-only config, routing-only config, pathless file names, Windows-style paths, mixed separators, and blank path rejection.
- Test stability looks good to me: the new tests use an isolated `LoggerContext` per config file and stop it in `finally`, with no fixed sleeps, ports, ordering assumptions, or external services.

Current merge gate:

- The required `Build` check is still failing: https://github.com/apache/seatunnel/runs/91110150413
- The failing lane I could identify is `all-connectors-it-7`, specifically the MySQL CDC schema-change E2E path. The failure shows the schema evolution assertion receiving `description` where `add_column1` was expected, together with a Debezium schema out-of-sync message.
- I do not see a direct call-chain connection between that CDC E2E failure and this PR's `LogUtil` changes.
- This branch is currently behind/diverged from `dev` (`ahead_by=8`, `behind_by=7`, compare status: `diverged`).

### Merge conclusion

### Conclusion: can merge after the CI / merge gate is resolved

1. Blockers that must be resolved
   - The current required `Build` check is failing, so the PR cannot be merged yet.
   - Given the current evidence, the failure looks unrelated to the `LogUtil` diff. The lowest-cost next step is to sync with the latest `dev` and rerun the Build first. If the same CDC lane still fails after the sync, we should narrow it down against that updated head rather than patching unrelated logging code in this PR.

2. Non-blocking suggestions
   - None from my side for the current source diff.

Overall, the source change looks clean and well-scoped. Once the CI / merge gate is green, I do not see a Daniel-side source blocker for this PR.

DanielLeens:
Rechecked the latest head `982784db56526d623201b9ae089033683c39737e` and the current gate.

My earlier CI guidance is outdated now: the required `Build` check is successful on the current head: https://github.com/apache/seatunnel/runs/91110150413

From Daniel's side, I still do not see a reopened source-level blocker in this `LogUtil`-only diff.

The branch is currently still diverged from `dev` (`ahead_by=8`, `behind_by=15`, compare status: `diverged`), and GitHub is reporting the merge state as `BLOCKED`. I am recording that as a mergeability fact rather than turning it into a new source blocker or a sync-first request on a green source/CI state.

Because I only have read-level review controls on this repository, this follow-up comment is my current Daniel-side technical conclusion for the latest head. If GitHub still needs a formal approval or another merge-gate step, a write-capable maintainer will need to handle that final repository-level action.

DanielLeens (approved):

My last comment on this head already recorded that my earlier CI guidance was outdated, that the required `Build` check was passing, and that I saw no reopened source-level blocker in this `LogUtil`-only diff. That was left as a comment rather than a review, so the PR still shows as awaiting review. Rechecking the current head `982784db`: **all checks are passing** (3/3 SUCCESS), so I am converting it into a formal approve.

Re-confirmed on this head:
- The change stays confined to file-appender log path resolution. It does not alter job scheduling, checkpointing, state recovery, serialization, or any connector path.
- Compatibility: no API or SPI signature change, no config key renamed or removed, no default-value change, no protocol or serialization change. Fully compatible. Existing deployments do not need to change anything on upgrade.
- Test stability: no timing, port, container, external-service, or ordering dependency is introduced. **Stable.** No existing assertion, tolerance, or check anywhere in this PR is weakened.

Two housekeeping notes, same as on your other PR:

- The branch is still diverged from `dev` and GitHub reports the merge state as `BLOCKED`. That is queue metadata only — I am **not** asking you to sync, since there is no failing signal that upstream drift would explain.
- Correcting myself: I previously said I only had read-level review controls and that a write-capable maintainer would need to issue the approval. That was wrong — I can and have submitted this review. A committer's approval remains a separate, still-required step for the merge.

Thanks — log path resolution bugs are quietly expensive when you are trying to debug something else, so this is a good fix to have in.

PR #11466, opened 2026-07-16
[Docs] Add verified MySQL CDC to Kafka recipe and E2E coverage

DanielLeens:
## What changed

- add a new English and Chinese recipe for `MySQL CDC -> Kafka` with metadata headers and SQL field shaping
- add a Kafka E2E recipe test plus matching DDL and job config so the documented path has executable evidence
- update the getting-started recipe overview and local quick-start navigation to point readers to the verified Kafka walkthrough

## Why

The current scenario recipe section needed a tutorial-grade example that was checked against a real end-to-end path instead of only describing connector options in the abstract.

## Validation

- `./mvnw spotless:apply -nsu -Dmaven.gitcommitid.skip=true`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kafka-e2e -Dtest=KafkaRecipeIT -DfailIfNoTests=false test -nsu -Dmaven.gitcommitid.skip=true`
  - passed on July 16, 2026 at 19:25:30 +08:00
- `./mvnw -q -DskipTests verify -nsu -Dmaven.gitcommitid.skip=true`
  - local failure in `seatunnel-dist` during `export-connector-metadata` because the local environment loaded `/Users/stone/soft/apache-seatunnel-2.6-WS-test-SNAPSHOT/connectors` and then missed `org.apache.seatunnel.api.table.factory.TableSinkFactory`

### DIFF
--- a/docs/en/getting-started/locally/quick-start-seatunnel-engine.md
+++ b/docs/en/getting-started/locally/quick-start-seatunnel-engine.md
@@ -245,7 +245,7 @@ Recommendation:
 
 - Start with [Getting Started Overview](../overview.md) if you want a guided reading path across deployment, quick start, and configuration.
 - Use [Job Configuration Guide](../job-configuration-guide.md) when you are ready to replace the sample source and sink with real connectors.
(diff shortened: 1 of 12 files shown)


davidzollo (approved):
+1

PR #11467, opened 2026-07-16
[Fix][Zeta] Fix file appender log path

### LINKED ISSUE #11463: [Bug] [Zeta] LogUtil.getLogPath uses wrong variable in fileAppender branch (NPE / StringIndexOutOfBounds)

zhangshenghang:
## What does this PR do?

Fixes #11463 by deriving the file appender directory from `fileLogPath` instead of `routingLogFilePath`. This prevents a `NullPointerException` when a Zeta deployment configures only `fileAppender`.

## Why is this change needed?

The file-appender branch used the routing appender path only to calculate the substring index. That path is absent in a file-appender-only configuration, so log REST endpoints fail while resolving the log directory.

## Tests

- Added a regression test with a file-appender-only Log4j2 configuration.
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home ./mvnw -pl seatunnel-engine/seatunnel-engine-common test`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home ./mvnw -q -DskipTests verify`

`./mvnw test` remains blocked before this module by existing failures in `seatunnel-config-shade`.


DanielLeens (approved):
Thanks for the fix. I reviewed the current head from the real runtime path rather than only reading the one-line diff.

Runtime path rechecked:
`BaseLogService.getLogPath()` -> `LogUtil.getLogPath()` -> pick root logger appender ref -> resolve `routingAppender` or `fileAppender` directory -> REST log endpoints / task log manager consume that directory.

For the `fileAppender` branch, the previous code sliced `fileLogPath` with `routingLogFilePath.lastIndexOf("/")`, so a deployment that only configures the file appender could return the wrong directory or fail entirely if the routing appender path shape differs. The new branch now slices `fileLogPath` with its own separator, which matches the actual string being returned in `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:49-52`.

The added regression test and dedicated `log4j2-test.properties` cover the exact single-file-appender path that was previously broken, so I do not see a source-side blocker on this revision. CI is green on the current head as well.

LGTM from Daniel.

DanielLeens (commented):
Rechecked the same unchanged head `2a42a2c4778402e2ae8e62bd2a7c4899e5e975ad` on Friday, July 17, 2026 because the CI fact changed since my earlier review.

From Daniel's side, the source-level conclusion is still clear on this revision. The actual fix in `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:49-52` is still the right one for the file-appender path, and the added unit test in `.../LogUtilTest.java:26-29` covers the file-appender-only case that was missing before.

What is stale now is my earlier CI note. On this same head, the current `Build` check is failing here: https://github.com/apache/seatunnel/runs/87644054307

So the remaining gates on the current revision are the failing `Build` check plus the PR still being in draft state, not a reopened source blocker from Daniel's side. Once the current `Build` result is addressed on the latest head, I'm happy to take another look if needed.

DanielLeens (commented):

Fixes a directory-slicing bug in `LogUtil.getLogPath()` where the `fileAppender`-only branch mistakenly computed its substring index from `routingLogFilePath` instead of from `fileLogPath` itself. That branch is not a rare edge case: `config/log4j2.properties:41` ships with `rootLogger.appenderRef.file.ref = fileAppender` out of the box, and `docs/en/engines/zeta/logging.md` states the mixed-log/`fileAppender` mode is *"this configuration mode by default"*. So this fix restores correct behavior for the default SeaTunnel Engine logging setup, feeding both the `/logs` REST API and the scheduled log-cleanup job.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Before:
```java
```
After:
```java
```

Runtime call chain I traced end-to-end:
- `LogUtil.getLogPath()` (`seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:34-57`) reflectively reads the *current process's own* log4j2 `PropertiesConfiguration` (`getLogConfiguration()`, :59-62, via `LogManager.getContext(false)`), extracts `fileAppender`'s resolved `fileName` (`getFileLogPath`, :82-93) and the routing appender's nested job-template `fileName` (`getRoutingLogFilePath`, :64-80), then branches on whichever ref the rootLogger actually points to.
- Called by `BaseLogService.getLogPath()` (`seatunnel-engine-server/.../rest/service/BaseLogService.java:46-53`), which backs four REST log entry points — `CurrentNodeLogServlet`, `AllNodeLogServlet`, `AllLogNameServlet`, and `RestHttpGetCommandProcessor` (lines 313/371/429) — i.e. the `/logs` and `/logs/{jobId}` endpoints documented in `docs/en/engines/zeta/logging.md`.
- Also called by `TaskLogManagerService.initClean()` (`seatunnel-engine-server/.../telemetry/log/TaskLogManagerService.java:39`), wired from `SeaTunnelServer`/`CleanLogOperation`, to compute the base directory that the periodic `history-job-expire-minutes` cleanup sweep (`clean(long jobId)`, :54-68) deletes files under.

Why the two strings can diverge: `fileLogPath` resolves `${file_path}/${file_name}.log` via `StrSubstitutor`, while `routingLogFilePath` resolves the routing appender's nested job template `${file_path}/job-${ctx:ST-JID}.log`. Both share the `${file_path}` property, but the routing template additionally carries the `${ctx:ST-JID}` MDC lookup token, which is not resolved at config-parse time (no active `ThreadContext` job id yet) — so `routingLogFilePath` keeps a longer/differently-shaped tail than `fileLogPath`. Whenever that shape makes `routingLogFilePath.lastIndexOf("/")` land at a different offset than the true end of the shared directory prefix, slicing it out of `fileLogPath` yields the wrong directory (or a `StringIndexOutOfBoundsException` if the offset exceeds `fileLogPath`'s length). The fix removes the cross-string coupling entirely by slicing `fileLogPath` using its own last `/`, which is unconditionally correct for this branch. Precise, minimal fix with no secondary side effect — still a pure O(n) string op, no new allocation/I/O.

## 1.2 Compatibility Impact: Fully compatible.
Only the `fileAppender` branch's return value changes, strictly from "possibly-wrong slice or exception" to "correct directory of `fileLogPath`." No caller could have a legitimate dependency on the old broken output. No config option, default, serialization format, or public API signature changes.

No change in complexity/allocation/I/O; the method stays a pure, on-demand string computation (per REST request / cleanup tick, not a hot loop). I also specifically checked the multi-node/multi-process angle: `getLogPath()` only reads the *calling JVM's own* `LoggerContext`, and `AllNodeLogServlet`/`RestHttpGetCommandProcessor` fan requests out per-node rather than sharing one resolved path across nodes (`AllNodeLogServlet.java:39-42,62`). So this fix stays a purely local, per-process computation — it neither introduces nor fixes any cross-node log-path collision; there was none to begin with here.

No blocking issues found. See Issue 1 below for a related but pre-existing, out-of-scope observation.

# 2. Code Quality Assessment

## 2.1 Coding Standards
The one-line fix matches existing style. The new `LogUtilTest` class/method carry no Javadoc — that was already true on the revision I approved on 2026-07-17 and is unchanged now, so I'm not treating it as a new issue; it remains an optional, non-blocking nit.

`LogUtilTest` + the new `log4j2-test.properties` build a real log4j2 `PropertiesConfiguration` where only `fileAppender` is registered (`rootLogger.appenderRef.file.ref = fileAppender`, no `routingAppender` component present at all) and assert `LogUtil.getLogPath()` equals `"target/file-appender-logs"`. This exercises exactly the previously-broken line (`LogUtil.java:52`) through real config parsing, not a mock — a genuine regression test.

Stability rating: **Stable**. Synchronous, single-threaded, no `Thread.sleep`, no shared mutable state beyond log4j2's normal per-classloader `LoggerContext` (expected/standard for log4j2 tests, isolated via the dedicated test properties file), no ordering dependency, deterministic literal-string assertion.

## 2.3 Documentation Updates
No user-visible behavior/config/default changed — this restores behavior the docs already describe as default. No doc update required, and none made. Correct.

# 3. Architectural Soundness

## 3.1 Elegance of the Solution: Precise fix.
Slicing `fileLogPath` by its own `lastIndexOf("/")` is the obviously-correct minimal fix; it removes the accidental cross-variable coupling that caused the bug, no workaround/flag needed.

## 3.2 Maintainability
Improves maintainability — each branch is now self-contained (slices its own resolved path), instead of the `fileAppender` branch having a hidden dependency on `routingLogFilePath`.

## 3.3 Extensibility
No impact.

No checkpoint/savepoint, serialization, protocol, or persisted-state involvement — this is a logging-path-resolution utility with no bearing on job recovery or upgrade paths. No historical-version compatibility risk.

# 4. Issue Summary

| 1 | `logRef` selection uses substring matching (`ref.contains(routingAppender)`/`contains(fileAppender)`, `LogUtil.java:46`) but then compares with exact-match `logRef.equals(fileAppender)` (`:49`/`:51`). Pre-existing, **not touched by this PR's diff**; only a latent risk if a deployment names an appender ref containing "fileAppender"/"routingAppender" as a substring rather than matching it exactly (not the case in the shipped default config). Not a regression introduced here — flagging for awareness only. Raised by another reviewer: No. | `LogUtil.java:43-48` | Low |

No blocking issues found in the diff itself.

# 5. Merge Recommendation

### Conclusion: Ready to merge after fixes (procedural only, not source-level)

I owe this PR a process-side apology rather than a code-side one: my 2026-07-17 comment flagged a failing `Build` check on the then-head, and the PR is still marked draft. Since then the author merged `upstream/dev` into the branch (`761e6e48b`, 2026-08-12T10:50:59Z) — a pure merge commit that touches neither `LogUtil.java`, `LogUtilTest.java`, nor `log4j2-test.properties`. I diffed the current head against `origin/dev` and confirmed it is byte-for-byte identical, in this PR's own files, to the revision I already approved. So this is a genuinely fresh, from-scratch re-read of the current head (not a rubber stamp), and my source-level conclusion is unchanged: **LGTM from the source side.**

What remains open is procedural, not a source blocker:
1. Blockers — must be fixed:
   - None on the source side.
2. Recommended, non-blocking:
   - CI on the new head (`761e6e48b`) was only just (re)triggered around 10:51 UTC and is still queued/pending at the time of this review — let it finish green before merging; I saw no fresh failure, just an unresolved result so far.
   - The PR is still in **draft** state; nothing on my side is holding it there — please mark it ready for review once CI is confirmed green.
   - Optional: a one-line Javadoc on the new `LogUtilTest` class per repo convention (Issue table above, Low, not blocking).

Overall assessment: this is a precise, well-tested one-line fix for a real default-path bug, with a targeted, stable regression test. No alternative approach needed — Option A (this PR's fix) is correct as-is.

DanielLeens:
Thanks for the update. I checked the current head against the latest `dev` and this branch is currently behind upstream (`behind_by=16`, compare status: `diverged`). The current failing CI signal — the fork's `Build` workflow job `unit-test (8, windows-latest)` — does not appear to be directly caused by this PR's diff (the same job is currently failing on other unrelated PRs' heads as well), so the lowest-cost next step would be to sync with the latest `dev` and rerun CI. If it still fails after the sync, happy to help narrow it down against the updated head.

zhangshenghang:
Closing this PR since the same fix (the LogUtil.java change plus an equivalent regression test) has already been merged via #11820, which also closed #11463. Thanks for the reviews!

PR #11468, opened 2026-07-16
[Docs] Fill in env config overview

DanielCarter-stack:
## What does this PR do?

This PR fills in the missing `env` parameter overview in the config introduction docs. The previous page still contained a TODO placeholder, so readers could not quickly see which job-level settings are most commonly used.

## Why is this change needed?

The config introduction page is meant to help first-time users understand the four main sections of a SeaTunnel job. Leaving the `env` section unfinished makes that page less useful and forces readers to jump away before they understand the basics.

## What is changed?

- Replace the TODO placeholder in `docs/en/introduction/concepts/config.md` with a short list of commonly used `env` parameters
- Add the same explanation to `docs/zh/introduction/concepts/config.md`
- Keep the detailed reference in `JobEnvConfig` as the next step for readers who need full parameter coverage

## Verification

- `./mvnw spotless:apply`
  - Failed in the local environment with `java.lang.UnsatisfiedLinkError: no nio in java.library.path` from the Spotless Maven plugin
  - Failed for the same Spotless plugin environment issue
- `git diff --check`
  - Passed

Because this PR only updates documentation text, no code paths were changed.

### DIFF
--- a/docs/en/introduction/concepts/config.md
+++ b/docs/en/introduction/concepts/config.md
@@ -70,7 +70,17 @@ Use `env` for job-level and engine-level settings such as `job.mode`, `paralleli
 
 Common parameters are shared across engines. Engine-specific parameters are separated by prefix. For Flink and Spark, see [JobEnvConfig](../configuration/JobEnvConfig.md).
 
-<!-- TODO add supported env parameters -->
+The most common `env` parameters are:
+
+- `job.mode`: chooses `BATCH` or `STREAMING`
+- `job.name`: sets the job name shown by the engine and UI
+- `parallelism`: controls how many parallel readers and writers SeaTunnel uses
+- `checkpoint.interval`: enables periodic checkpoints in streaming jobs
+- `checkpoint.timeout`: limits how long a checkpoint can run before the job fails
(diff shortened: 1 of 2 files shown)


DanielLeens (approved):
Thanks for filling this gap in the config docs. I checked the current head against the actual user-facing config guidance rather than only reading the markdown diff.

What this improves:
- User pain: the `env` section in the concept doc pointed readers to `JobEnvConfig`, but the page itself still had a TODO and did not explain the most common knobs people actually look for first.
- Current fix: both `docs/en/introduction/concepts/config.md` and `docs/zh/introduction/concepts/config.md` now list the common `env` parameters (`job.mode`, `job.name`, `parallelism`, checkpoint settings, `jars`, `shade.identifier`) and then point to `JobEnvConfig` for the full matrix.

I rechecked the EN/ZH pair side by side and the two versions stay aligned in meaning on the current head. Since this is a documentation-only clarification and the guidance matches the existing config model, I do not see a blocker on this revision.

LGTM from Daniel.

dybyte (approved):

Issue #11433, opened 2026-07-15
[Bug] [Doris Sink] Discrepancy between the transaction status returned by the Doris sink connector and the actual visible Doris data status.


chengwei977:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/seatunnel/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened

In the Doris sink connector, when two-phase commit (2PC) is enabled, once a Doris transaction is in the COMMITTED state, Seatunnel considers the transaction complete and returns the job as finished. However, Doris actually recognizes data visibility only when the transaction is in the VISIBLE state. Depending on the load of the Doris sink, the state difference between COMMITTED and VISIBLE can range from tens of milliseconds to several seconds. If a scheduler like DolphInscheduler is used, subsequent SQL jobs that depend on the Seatunnel job being in the finished state may result in data not being read.

### SeaTunnel Version

seatunnel-2.3.13

### SeaTunnel Config

```conf
No special configuration
```

### Running Command

```shell
rest api
```

### Error Exception

```log
none
```

### Zeta or Flink or Spark Version

Zeta

### Java or Scala Version

Java

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

DanielLeens:
Thanks for spelling out the COMMITTED versus VISIBLE gap clearly. I traced the current Doris 2PC completion path, and this does look like a real issue rather than a usage mistake.

Right now SeaTunnel stops the load, gets the transaction id, and then the Doris committer treats the 2PC commit as successful once the commit request succeeds or Doris reports the transaction as already COMMITTED / VISIBLE. There is no extra wait or polling step that keeps the SeaTunnel job open until the data is actually visible in Doris.

So the completion boundary exposed by SeaTunnel is currently earlier than the visibility boundary your downstream SQL jobs depend on. That matches the behavior you are seeing with scheduler-chained jobs.

This is worth keeping open. The likely fix direction is to distinguish transaction committed from data visible, and only report the sink as finished after the visible stage is reached, or make that boundary explicit if Doris-side behavior forces a different contract.

If you happen to have a concrete Doris FE response sample for the COMMITTED -> VISIBLE transition on your side, that would help nail down the final polling condition, but the issue itself is already valid as a bug report.

[...]

PR #11478, opened 2026-07-17
[Fix][Connector-V2]Fix the issue where the transaction status returne…

muyuQvQ:
…d by the Doris Sink connector doesn't match the actual visible Doris data (#11433)

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/exception/DorisConnectorErrorCode.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/exception/DorisConnectorErrorCode.java
@@ -31,7 +31,8 @@ public enum DorisConnectorErrorCode implements SeaTunnelErrorCode {
     SCAN_BATCH_FAILED("Doris-09", "scan batch error"),
     RESOURCE_CLOSE_FAILED("Doris-10", "resource close failed"),
     SCHEMA_CHANGE_FAILED("Doris-11", "schema change failed"),
-    SHOULD_NEVER_HAPPEN("Doris-00", "Should Never Happen !");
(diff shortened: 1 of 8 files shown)


DanielLeens (commented):
Thanks for working on this. I re-reviewed the full current diff on head `4c19d061cc442f2f301211c45f1b592eebb7111a`, and I found two code-side blockers that need to be fixed before this can merge.

1. `DorisCommitInfoSerializer` is no longer symmetric. `prepareCommit()` now creates `DorisCommitInfo` with the extra `label` field (`.../DorisSinkWriter.java:255-267`), and `deserialize()` now reads that extra UTF field (`.../DorisCommitInfoSerializer.java:45-52`), but `serialize()` still only writes `hostPort`, `db`, and `txnId` (`.../DorisCommitInfoSerializer.java:32-40`). That means any commit info produced on the current head will fail to deserialize on the restore / committer path with an EOF-style read failure before the label can ever be recovered.

2. The new load-state polling path leaks HTTP resources and does not react cleanly to stop / cancel. `getStreamLoadState()` loops until `VISIBLE` (`.../DorisCommitter.java:272-301`), and each iteration calls `sendGetWithAuth()`, which creates a fresh `CloseableHttpClient` and executes a `CloseableHttpResponse` without closing either of them (`.../DorisCommitter.java:304-313`). The same loop also swallows interruption in the `Thread.sleep(500)` block (`.../DorisCommitter.java:296-299`). On the commit path, that can turn a slow or stuck FE/BE visibility transition into a resource leak and a task that does not stop promptly when the engine is trying to cancel it.

I also see the current `Build` check is `ACTION_REQUIRED` here: https://github.com/apache/seatunnel/runs/87875644590. I am not counting that as the source blocker above, but it is still a separate gate to clear after the code-side issues are fixed.

Please take a look at those two items and push an update when you have it. Happy to re-review.

davidzollo:

```

```

PR #11486, opened 2026-07-18
[Fix][Connector-V2] Complete batch Kafka splits after control records

### LINKED ISSUE #11471: [Bug] [Connector-V2 Kafka] Batch Kafka source does not finish when reading transactional topic control records
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/seatunnel/issues?q=is%3Aissue) and found no similar issues.

### What happened

When a Kafka topic is written by SeaTunnel Kafka sink with `semantics = EXACTLY_ONCE`, a subsequent Zeta batch job using Kafka source can read all visible user records and write them to the sink, but the job does not finish.

The same Kafka source batch job finishes normally when reading:

- a topic populated by `kafka-console-producer`
- a topic populated by SeaTunnel Kafka sink with `semantics = NON`

The reproducible difference is that the transactional topic contains Kafka transaction control record(s). In my validation with 100 user records:

- EXACTLY_ONCE topic offsets: `partition 2 end offset = 101`
- Console consumer visible records: `100`
- Kafka source -> JDBC sink metrics after 90 seconds:
  - `jobStatus = RUNNING`
  - `SourceReceivedCount = 100`
  - `SinkWriteCount = 100`
  - `SinkCommittedCount = 100`
  - `IntermediateQueueSize = 0`
  - target table count = `100`

For the control case with `semantics = NON`:

- topic end offset = `100`
- Console consumer visible records = `100`
- Kafka source -> JDBC sink job status = `FINISHED`
- target table count = `100`

It looks like the batch Kafka source may use the raw end offset including transaction control records as the stopping offset, while the Kafka consumer only emits the 100 user records. As a result, the source has consumed all visible records but never satisfies the batch completion condition.

### SeaTunnel Version

`3.0.0-SNAPSHOT`, tested on Zeta engine build commit `46d2f39`.

### SeaTunnel Config

Step 1: write a transactional Kafka topic with SeaTunnel Kafka sink:

```hocon
env {
  job.name = "kafka-exact-write"
  job.mode = "BATCH"
  parallelism = 1
  checkpoint.interval = 10000
  checkpoint.timeout = 60000
}

source {
  FakeSource {
    plugin_output = "fake"
    auto.increment.enabled = true
    auto.increment.start = 1
    row.num = 100
    schema {
      fields {
        id = "int"
        name = "string"
        age = "int"
      }
      primaryKey {
        name = "pk"
        columnNames = [id]
      }
    }
  }
}

sink {
  kafka {
    plugin_input = "fake"
    topic = "zeta_tx_recheck_exact"
    bootstrap.servers = "kafka:9092"
    format = json
    semantics = EXACTLY_ONCE
    transaction_prefix = "zeta2-tx-recheck-exact"
    kafka.config = {
      acks = "all"
      request.timeout.ms = 60000
      buffer.memory = 33554432
    }
  }
}
```

Step 2: read that topic with Kafka source in batch mode:

```hocon
env {
  job.name = "kafka-exact-read"
  job.mode = "BATCH"
  parallelism = 1
  checkpoint.interval = 10000
  checkpoint.timeout = 60000
}

source {
  Kafka {
    plugin_output = "kafka_rows"
    schema {
      fields {
        id = "int"
        name = "string"
        age = "int"
      }
      primaryKey {
        name = "pk"
        columnNames = [id]
      }
    }
    format = json
    topic = "zeta_tx_recheck_exact"
    bootstrap.servers = "kafka:9092"
    start_mode = "earliest"
    consumer.group = "kafka-exact-read"
    kafka.config = {
      client.id = "kafka-exact-read"
      max.poll.records = 50
      enable.auto.commit = "false"
    }
  }
}

sink {
  jdbc {
    plugin_input = "kafka_rows"
    url = "jdbc:mysql://mysql:3306/zeta_test"
    user = "zeta"
    password = "******"
    query = "insert into zeta_tx_recheck_exact(id,name,age) values(?,?,?)"
  }
}
```

Control case: change the Kafka sink in step 1 to `semantics = NON`. The read job finishes normally.

### Running Command

Submitted through Zeta REST API:

```shell
curl -H 'Content-Type: text/plain' \
  --data-binary @job.conf \
  'http://<zeta-master>:8080/submit-job?format=hocon'
```

### Error Exception

No exception is thrown. The job remains `RUNNING` even though all user records have been read and committed to the JDBC sink.

Observed job info for the hanging read job:

```json
{
  "jobStatus": "RUNNING",
  "errorMsg": null,
  "metrics": {
    "SourceReceivedCount": "100",
    "SinkWriteCount": "100",
    "SinkCommittedCount": "100",
    "IntermediateQueueSize": "0"
  }
}
```

Zeta engine.

### Java or Scala Version

Java 11.

### Screenshots

N/A

### Code of Conduct

zhangshenghang:
## What does this PR do?

Fixes #11471.

This change completes a batch Kafka split when the consumer position reaches its stopping offset, including when Kafka has advanced the position through transaction control records that are not returned by `poll()`.

## Why is this change needed?

The previous completion check only examined the last record returned by `poll()`. For transactional topics, a control record can advance the Kafka position without appearing in that record list, leaving the batch job running after all visible records are processed.

## How was this tested?

- Added a regression test for a consumer position advanced past a transaction control record.
- `./mvnw -q -pl seatunnel-connectors-v2/connector-kafka -Dtest=KafkaPartitionSplitReaderTest,KafkaSourceSplitEnumeratorTest test`


DanielLeens (commented):
Thanks for the contribution! I re-reviewed the full current diff and the relevant Kafka consumer-position handling runtime path.

I do not see a source-side blocker on the current head. The Kafka lanes pass; the aggregate `Build` is red on the shared engine `MetricsApiTest`, so CI remains the only technical gate from this review.

DanielLeens (commented):



The source review is clear, but the current Build is failing, so this PR is still waiting on CI. The current failed lanes include `Run / Dead links`, `Run / unit-test (11, ubuntu-latest)`, `Run / updated-modules-integration-test-part-1 (8, ubuntu-latest)`, `Run / kafka-connector-it (11, ubuntu-latest)`. The shared dead-link failure has a fix in #11502, but that fix has not merged into `dev` yet; syncing the current `dev` alone will not clear that lane today. After #11502 lands, please sync/retrigger and then diagnose only any lanes that still fail.

DanielLeens (commented):



The source review is clear, but the current Build is failing, so this PR is still waiting on CI. The current failed lanes include `Run / Dead links`, `Run / unit-test (11, ubuntu-latest)`, `Run / updated-modules-integration-test-part-1 (8, ubuntu-latest)`, `Run / kafka-connector-it (11, ubuntu-latest)`. The shared dead-link failure has a fix in #11502, but that fix has not merged into `dev` yet; syncing the current `dev` alone will not clear that lane today. After #11502 lands, please sync/retrigger and then diagnose only any lanes that still fail.

DanielLeens (commented):
# What problem this PR solves

- User pain: A bounded Kafka source can poll forever after an invisible transaction-control record even though the consumer has already crossed the stopping offset.
- Fix approach: After each poll, read KafkaConsumer.position for every finite partition and finish/unassign the split once the next-fetch position reaches or exceeds its stopping offset.
- One-line summary: Completion now follows Kafka's authoritative next-fetch position, so transaction-control records no longer strand a batch split.
- Before: Only the last visible record returned by poll was checked; completion could not be detected when the control record was hidden.
- After: Even on an empty poll, the split finishes once consumer.position reaches the exclusive stop, without emitting records beyond the boundary.

# 1. Code-change review

- Behavior before: Only the last visible record returned by poll was checked; completion could not be detected when the control record was hidden.
- Behavior after: Even on an empty poll, the split finishes once consumer.position reaches the exclusive stop, without emitting records beyond the boundary.
- Normal-path hit: `conditional` — Every Kafka split with a finite stopping offset reaches the new check in the fetch loop.

- `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:78-157` `KafkaPartitionSplitReader.fetch` — After poll, traverses the assignment, reads finite-partition positions, and finishes splits that reached their stops.
- `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:159-280` `handleSplitsChanges / split setup` — Assigns, seeks, and stores explicit/end offsets; empty splits retain their existing completion path.
- `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:368-429` `KafkaPartitionSplitRecords` — Continues to cut output at the exclusive stopping offset and reports completed splits to SourceReader.
- `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:76-134` `transaction-control completion regression` — Mocks an empty poll and a position jump to 101, covering invisible control-record progress.

### Effective prerequisites

- The consumer has been assigned/seeked/polled and a finite partition's next position reaches or exceeds its stopping offset.

- Streaming partitions with Long.MAX_VALUE skip the extra check and retain unbounded semantics.

### Complete execution chain

```text
1. seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:78-157 -> KafkaPartitionSplitReader.fetch [lifecycle]
   After poll, traverses the assignment, reads finite-partition positions, and finishes splits that reached their stops.
   Side effect: After poll, traverses the assignment, reads finite-partition positions, and finishes splits that reached their stops.
2. seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:159-280 -> handleSplitsChanges / split setup [method_call]
   Assigns, seeks, and stores explicit/end offsets; empty splits retain their existing completion path.
   Side effect: Assigns, seeks, and stores explicit/end offsets; empty splits retain their existing completion path.
3. seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:368-429 -> KafkaPartitionSplitRecords [method_call]
   Continues to cut output at the exclusive stopping offset and reports completed splits to SourceReader.
   Side effect: Continues to cut output at the exclusive stopping offset and reports completed splits to SourceReader.
4. seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:76-134 -> transaction-control completion regression [method_call]
   Mocks an empty poll and a position jump to 101, covering invisible control-record progress.
   Side effect: Mocks an empty poll and a position jump to 101, covering invisible control-record progress.
```

- Recovery, exception, async, serialization, and concurrency paths: Kafka consumer calls remain on the split-reader/fetcher thread; wakeup/retry and output cutoff are unchanged. The new test has a minor real-consumer resource-release risk.

- Applicable: `false`

## 1.2 Compatibility impact

- API: Configuration, defaults, public APIs, split serialization, checkpoint state, the exclusive-offset contract, and Kafka protocol are unchanged; unbounded sources are unaffected.
- SPI: Configuration, defaults, public APIs, split serialization, checkpoint state, the exclusive-offset contract, and Kafka protocol are unchanged; unbounded sources are unaffected.
- Configuration: Configuration, defaults, public APIs, split serialization, checkpoint state, the exclusive-offset contract, and Kafka protocol are unchanged; unbounded sources are unaffected.
- Defaults: Configuration, defaults, public APIs, split serialization, checkpoint state, the exclusive-offset contract, and Kafka protocol are unchanged; unbounded sources are unaffected.
- Protocol and serialization: Configuration, defaults, public APIs, split serialization, checkpoint state, the exclusive-offset contract, and Kafka protocol are unchanged; unbounded sources are unaffected.
- Checkpoint/restore: Configuration, defaults, public APIs, split serialization, checkpoint state, the exclusive-offset contract, and Kafka protocol are unchanged; unbounded sources are unaffected.
- Historical behavior: Configuration, defaults, public APIs, split serialization, checkpoint state, the exclusive-offset contract, and Kafka protocol are unchanged; unbounded sources are unaffected.

- CPU: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Memory: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- GC: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Network: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Lock contention: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Concurrency safety: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Retry: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Idempotency: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Resource release: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.
- Overall: Each poll adds one cached position lookup per assigned finite partition, normally without a broker request; cost is O(finite partitions) with no record-growing buffer.

- Exception propagation: Position failures are not swallowed as successful completion; existing WakeupException, unassigned cleanup, and sensitive-information behavior are unchanged.
- Context: Position failures are not swallowed as successful completion; existing WakeupException, unassigned cleanup, and sensitive-information behavior are unchanged.
- Log levels: Position failures are not swallowed as successful completion; existing WakeupException, unassigned cleanup, and sensitive-information behavior are unchanged.
- Sensitive information: Position failures are not swallowed as successful completion; existing WakeupException, unassigned cleanup, and sensitive-information behavior are unchanged.
- Overall: Position failures are not swallowed as successful completion; existing WakeupException, unassigned cleanup, and sensitive-information behavior are unchanged.

- intended_change_correctness: `pass`
- diff_integrity: `pass`
- deletion_ledger: total=`20`, classified=`20`, unclassified=`0`
- Reconciliation: 20 = 20 + 0

- `deletion-001..deletion-020` | count=`20` | `deletion` | `intended_change` | `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java` (old line 106 .. old line 126) — Twenty lines of returned-record-only completion logic are replaced by assignment-wide position checks; no configuration, state, serializer, documentation, or capability is removed.

- Out-of-scope hunk conclusion: The complete two-file diff is limited to Kafka split completion and its regression test, with no out-of-scope hunk.
- Stale-base conclusion: Although current dev is 48 commits ahead, neither changed file moved after the merge base and the merge tree has no conflict or stale-base rollback.
- Merge-artifact conclusion: No merge artifact, accidental deletion, rename, or whitespace defect was found.

# 2. Code-quality assessment

- Title format: The title and commits follow [Fix][Connector-V2], and the production change is focused. Reflective replacement of the final consumer is a Low/non-blocking test-quality issue.
- Code conventions: The title and commits follow [Fix][Connector-V2], and the production change is focused. Reflective replacement of the final consumer is a Low/non-blocking test-quality issue.
- Comment completeness: The title and commits follow [Fix][Connector-V2], and the production change is focused. Reflective replacement of the final consumer is a Low/non-blocking test-quality issue.

- Unit tests: `covered` — A deterministic unit regression covers position reaching the stop on an empty poll; no local tests ran, and the required Build did not reach Kafka IT.
- Component tests: `missing` — A deterministic unit regression covers position reaching the stop on an empty poll; no local tests ran, and the required Build did not reach Kafka IT.
- Engine integration tests: `not_applicable` — A deterministic unit regression covers position reaching the stop on an empty poll; no local tests ran, and the required Build did not reach Kafka IT.
- E2E: `missing` — A deterministic unit regression covers position reaching the stop on an empty poll; no local tests ran, and the required Build did not reach Kafka IT.
- Regression tests: `covered` — A deterministic unit regression covers position reaching the stop on an empty poll; no local tests ran, and the required Build did not reach Kafka IT.
- Stability classification: `risk`

### Test-stability evidence

- `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:76-134`
  - Trigger: Mocked offsets and assertions are deterministic with no network or clock wait; abandoning a real consumer creates a low-probability same-JVM resource/order risk.
  - Probability basis: Mocked offsets and assertions are deterministic with no network or clock wait; abandoning a real consumer creates a low-probability same-JVM resource/order risk.
  - Analysis: Mocked offsets and assertions are deterministic with no network or clock wait; abandoning a real consumer creates a low-probability same-JVM resource/order risk.

### Missing scenarios

- A real transactional broker/E2E, mixed multi-partition completion, boundary cutoff, and consumer-cleanup coverage are still missing.

## 2.3 Documentation

- User-visible change: `false`
- docs/en: Not applicable: this is an internal completion fix with no new user configuration, default, command, protocol, or documentation semantics.
- docs/zh: Not applicable: this is an internal completion fix with no new user configuration, default, command, protocol, or documentation semantics.
- Examples: Not applicable: this is an internal completion fix with no new user configuration, default, command, protocol, or documentation semantics.
- Configuration truth: Not applicable: this is an internal completion fix with no new user configuration, default, command, protocol, or documentation semantics.
- URL truth: Not applicable: this is an internal completion fix with no new user configuration, default, command, protocol, or documentation semantics.

### Local execution boundary

- `git diff --stat b994136e9c3d0466f1a10412e103059fce814d67...a052f079e0c2853e258e8c428eac145209dc3ab8`
  - Result: `passed`
  - Explanation: Confirmed 2 files and +65/-20.
- `git diff --name-status b994136e9c3d0466f1a10412e103059fce814d67...a052f079e0c2853e258e8c428eac145209dc3ab8`
  - Result: `passed`
- `git diff --unified=0 b994136e9c3d0466f1a10412e103059fce814d67...a052f079e0c2853e258e8c428eac145209dc3ab8`
  - Result: `passed`
- `git diff --check`
  - Result: `passed`

### Required checks

- `Build` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`FAILURE` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/88207895989

- `Notify test workflow` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/29692617855/job/88207884115
- `labeler` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/29692617845/job/88207884099
- `Build` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/88207895989

# 3. Architecture assessment

- Type: `precise_fix`
- Assessment: The fix sits in the reader layer that owns both consumer position and split stopping offsets, making it a precise and maintainable boundary fix.

## 3.2 Maintainability

The fix sits in the reader layer that owns both consumer position and split stopping offsets, making it a precise and maintainable boundary fix.

## 3.3 Extensibility

The fix sits in the reader layer that owns both consumer position and split stopping offsets, making it a precise and maintainable boundary fix.

The fix sits in the reader layer that owns both consumer position and split stopping offsets, making it a precise and maintainable boundary fix.

# 5. Formal findings

## `问题 1` — The regression test abandons a real KafkaConsumer when replacing the final field

- Location: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:76-104`
- Problem and real execution path: The test creates a real KafkaConsumer through the production constructor and then replaces it with a mock through reflection; the original consumer reference is lost and never closed. KafkaPartitionSplitReader construction -> real consumer creation -> setConsumer reflectively overwrites the private final field -> test ends without closing the original consumer.
- Current-head evidence: The current test neither captures and closes the original consumer nor provides an injectable consumer constructor/factory; the reflection helper is at lines 128-134.
- Trigger condition: Running the new unit test, especially when the same JVM later creates another Kafka consumer with the same client identity.
- Potential impact: Kafka metrics/JMX/app-info resources can leak, causing noise or minor test-order sensitivity; production behavior is unaffected.
- Best actionable fix: Add a package-private/test-visible consumer injection point, or capture and close the real consumer before replacement and clean up the reader/mock in finally.
- Severity: `low`
- Merge gate: `non_blocking`

# 6. Findings summary

|---|---|---|---|---|
| 问题 1 | The regression test abandons a real KafkaConsumer when replacing the final field | seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:76-104 | low | non_blocking |

# 7. Merge conclusion

## Conclusion：can merge after gates are satisfied

### Blocking items

- No source, documentation, compatibility, test-completeness, conflict, or diff-integrity blocker remains.

- `问题 1`: Add a package-private/test-visible consumer injection point, or capture and close the real consumer before replacement and clean up the reader/mock in finally.

### Overall assessment

The fix sits in the reader layer that owns both consumer position and split stopping offsets, making it a precise and maintainable boundary fix.

- Alternative: The position-based production logic is the preferred implementation; tests should inject the consumer instead of reflectively replacing a final field. A real-broker regression costs more but would strengthen integration evidence.
- Trade-offs: The position-based production logic is the preferred implementation; tests should inject the consumer instead of reflectively replacing a final field. A real-broker regression costs more but would strengthen integration evidence.

- Reviewed commit: `a052f079e0c2853e258e8c428eac145209dc3ab8`

DanielLeens:
Thanks for the follow-up. I rechecked the current active head on August 3, 2026.

From Daniel's side, the latest source review on this head was already down to CI / merge gates only. On the current active diff, I still only see the two Kafka source files for the bounded split-completion fix:
- `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java`
- `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java`

The branch is now `diverged` and behind `dev` (`behind_by=85`). The only visible `Build` result attached to this head is still the old run from July 19, 2026: https://github.com/apache/seatunnel/runs/88207895989 . In the contributor workflow behind that check, the visible non-green jobs include `Run / Dead links`, `Run / unit-test (11, ubuntu-latest)`, and `Run / updated-modules-integration-test-part-1 (8, ubuntu-latest)`. Those gates do not line up directly with the current two-file Kafka split-reader diff.

So the lowest-cost next step here is to sync with the latest `dev` and rerun CI on the refreshed branch first. If the same lanes are still red after that refresh, please paste the fresh failing job links here and I can help narrow them down further.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `a052f079e0c2` from scratch.

# What this PR fixes
- User pain: Batch Kafka jobs on transactional topics never finish — the transaction control record advances the consumer position past the stopping offset without ever appearing in `poll()` results, so the old last-record check in `fetch()` never fires and the split stays open forever (#11471).
- Fix approach: In `KafkaPartitionSplitReader.fetch()`, iterate `consumer.assignment()` instead of `consumerRecords.partitions()` and compare `consumer.position(tp)` (wrapped in `retryOnWakeup`) against the stopping offset, skipping unbounded partitions (`stoppingOffset == Long.MAX_VALUE`); a Mockito-based regression test injects a mocked consumer via reflection and verifies the split finishes when the position jumps to 101 past a control record.
- One-line summary: The core logic is correct and I verified the position-based check subsumes the old last-record check (a returned record at `stoppingOffset - 1` implies `position >= stoppingOffset`), the `Long.MAX_VALUE` guard keeps streaming splits from paying the `position()` cost, and the `retryOnWakeup` wrapper matches the existing pattern for consumer calls outside `poll()`; my remaining notes are that `consumer.position()` is now called on every fetch for every bounded assigned partition (a minor per-poll overhead worth confirming is acceptable), the test reaches into the private `consumer` field via reflection rather than a package-private hook, and the already-flagged low-severity doc gap.

# Runtime chain I rechecked
```text
SplitFetcher.runOnce() -> FetchTask.run()  (seatunnel-connectors-v2 common source reader)
  -> KafkaPartitionSplitReader.fetch()  KafkaPartitionSplitReader.java:95
    -> retryOnWakeup(() -> consumer.poll(pollTimeout))  KafkaPartitionSplitReader.java:98
    -> for (TopicPartition tp : consumer.assignment())  KafkaPartitionSplitReader.java:106  [changed: was consumerRecords.partitions()]
      -> getStoppingOffset(tp)  KafkaPartitionSplitReader.java:107  (returns Long.MAX_VALUE for unbounded splits -> continue at :109)
      -> retryOnWakeup(() -> consumer.position(tp))  KafkaPartitionSplitReader.java:112  [changed: replaces lastRecord.offset() check]
      -> if (currentOffset >= stoppingOffset)  KafkaPartitionSplitReader.java:119
        -> recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset)  KafkaPartitionSplitReader.java:120
        -> finishSplitAtRecord(tp, stoppingOffset, currentOffset, finishedPartitions, recordsBySplits)  KafkaPartitionSplitReader.java:121
    -> unassignPartitions(finishedPartitions) / recordsBySplits.finishedSplits  (consumed upstream to signal split completion and let the batch job terminate)
```

# Findings
- Severity: Low

# Review conclusion

DanielLeens:
Following up after @SEZ9's review, which concluded `can merge` with one non-blocking note.

I agree with their read of the source, and I want to +1 their **Issue 1** specifically: if this head changes any config or option surface, both `docs/en` and `docs/zh` need the corresponding update, since the project treats documentation as part of a user-visible change rather than a follow-up. It is genuinely non-blocking as they said, but it is also the kind of thing that never gets done later, so it is worth folding in on the next push if you are touching the branch anyway.

Beyond that I have no source-level blocker on this head. The diff is still the two Kafka split-reader files for the bounded split-completion fix, which is a tightly scoped change.

On CI, my position from the last comment is unchanged and I want to be precise about why I am asking for a sync rather than treating it as boilerplate: the only `Build` result attached to this head is the run from July 19, the branch is now `behind_by=85`, and the non-green jobs in that stale run (`Dead links`, `unit-test (11, ubuntu-latest)`, `updated-modules-integration-test-part-1`) have no plausible connection to a Kafka split-reader change. A signal that old cannot tell you whether your diff is healthy. Syncing with the latest `dev` and rerunning is the only step that produces an interpretable result. If the same lanes are still red on the refreshed head, paste the new links here and I will dig in with you.

With the source agreed by two reviewers, this needs a committer's approval once CI is usable and green.

DanielLeens (commented):

Following up: the branch just picked up a new commit (`a0376c8499db`, "Merge remote-tracking branch 'upstream/dev'"), so per policy I'm doing a full fresh review of the current head rather than reusing the July 26 conclusions verbatim. Good news on process: I diffed the previously-reviewed commit (`a052f079e0c2`) against this new head and the two production/test files this PR touches (`KafkaPartitionSplitReader.java`, `KafkaPartitionSplitReaderTest.java`) have **zero changes** between them — the new commit is exactly the `dev` sync Daniel asked for on August 3/4 to get a usable `Build` signal, with no logic drift. `Build` is currently `QUEUED` on the fresh head, so this review is based on re-verifying the source from scratch, not on trusting the old run.

- User pain: a bounded (batch) Kafka source can poll forever when a transaction control record (or any record filtered by the client, e.g. an aborted transaction under `read_committed`) is the thing that pushes the partition past its stopping offset. Control records are never returned by `poll()`, so the previous "check the last *returned* record" heuristic could never observe that the partition had actually caught up, and the split reader would keep polling an already-exhausted partition indefinitely.
- Fix approach: after every `poll()`, iterate the consumer's *current assignment* (not just the partitions that happened to return records this cycle) and compare `KafkaConsumer.position(tp)` — the offset of the next record Kafka will fetch, which advances across control/filtered records even though `poll()` never surfaces them — against the split's stopping offset. Finite partitions that have caught up are finished; unbounded partitions (`stoppingOffset == Long.MAX_VALUE`) are skipped entirely to avoid the extra call.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Before** (`KafkaPartitionSplitReader.java`, pre-fix): only partitions present in `consumerRecords.partitions()` were examined, and completion was decided from `lastRecord.offset() >= stoppingOffset - 1`, i.e. purely from data `poll()` actually returned.

**After** (current head, `KafkaPartitionSplitReader.java:106-124`):
```java
for (TopicPartition tp : consumer.assignment()) {
    long stoppingOffset = getStoppingOffset(tp);
    if (stoppingOffset == Long.MAX_VALUE) {
        continue;
    }
    long currentOffset =
            retryOnWakeup(
                    () -> consumer.position(tp),
                    "getting current position to finish split");
    if (currentOffset >= stoppingOffset) {
        recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset);
        finishSplitAtRecord(
                tp, stoppingOffset, currentOffset, finishedPartitions, recordsBySplits);
    }
}
```

I independently re-traced the full runtime chain rather than taking the PR description at face value:

```text
SplitFetcher.runOnce() -> FetchTask.run()
  seatunnel-connectors-v2/connector-common/.../fetcher/FetchTask.java:53-64
  -> splitReader.fetch()  KafkaPartitionSplitReader.java:89-133
     -> consumer.poll(pollTimeout)  :92  (may return zero records for tp if only a
        control record / filtered record advanced it)
     -> for (TopicPartition tp : consumer.assignment())  :106  [changed from
        consumerRecords.partitions()]
       -> getStoppingOffset(tp)  :107,278-280  (Long.MAX_VALUE for unbounded -> skip)
       -> retryOnWakeup(() -> consumer.position(tp), ...)  :111-114  [new]
       -> currentOffset >= stoppingOffset  :119
         -> recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset)  :120
         -> finishSplitAtRecord(...)  :121-122 -> finishedPartitions.add(tp);
            recordsBySplits.addFinishedSplit(tp.toString())  :146-147
     -> markEmptySplitsAsFinished(recordsBySplits)  :126
     -> unassignPartitions(finishedPartitions)  :128-130  (removes tp from
        consumer.assign(...) so it will not reappear in consumer.assignment() on the
        next fetch())
  -> back in FetchTask.run(): elementsQueue.offer(lastRecords, ...)  FetchTask.java:59
     -> if (!lastRecords.finishedSplits().isEmpty()) splitFinishedCallback.accept(...)
        FetchTask.java:60-62   [decoupled from whether the batch carried any records]
  -> SourceReaderBase: finishedSplits = fetch.finishedSplits(); onSplitFinished(...)
     seatunnel-connectors-v2/connector-common/.../SourceReaderBase.java:204-212
```

The key correctness question I checked myself rather than trusting the earlier reviews: is `FetchTask` willing to fire `splitFinishedCallback` on a batch that carries **zero records** (exactly the scenario an invisible control record produces)? Yes — `FetchTask.run()` at `FetchTask.java:60-62` checks `lastRecords.finishedSplits().isEmpty()` completely independently of whether any records were enqueued, so an empty-records/non-empty-finishedSplits batch (which is exactly what this fix produces on an empty `poll()`) is fully honored downstream. This closes the loop: the fix's signal is not just set but actually consumed correctly by the shared reader infrastructure.

**`KafkaConsumer.position(tp)` semantics** — this is the crux of the fix and I verified it against the Kafka client contract rather than assuming it: `position()` returns the offset of the *next* record the consumer will fetch, and it is updated by `poll()` for every record consumed off the wire, including transaction control batches and (under `isolation.level=read_committed`) aborted-transaction records that the client filters out before returning anything to the caller. So `position()` strictly dominates "offset of the last record actually returned," which is exactly why it subsumes the old heuristic: if the old check (`lastRecord.offset() >= stoppingOffset - 1`) would have fired, then `position()` (next-fetch offset) is already `>= stoppingOffset`, so the new check fires too — with no regression, only a strict widening of the cases detected (SEZ9 verified the same equivalence in their August 4 review; I confirm it independently here from the client contract rather than the PR's own framing).

**Cost of the added `position()` call**: since `handleSplitsChanges()` already seeks every partition to its starting offset before assignment completes (`seekToStartingOffsets`, :191), the position is always known by the time `fetch()` runs, so `position(tp)` reads from the cached `SubscriptionState` and does not by itself trigger a broker round trip. The `stoppingOffset == Long.MAX_VALUE` guard (:108-110) also means unbounded/streaming splits pay zero extra cost per poll, which matters since this reader is shared by both batch and streaming Kafka sources.

**Interaction with pre-existing completion paths** — I checked this is not double-signaled: `removeEmptySplits()` (:224-247, unchanged by this diff) already unassigns and marks-empty any partition whose position is `>= stoppingOffset` at split-assignment time, before the first `poll()`. Once unassigned there, that `tp` no longer appears in `consumer.assignment()`, so the new per-fetch loop never reprocesses it. The two paths are complementary (assignment-time vs. post-poll-time), not overlapping.

**Emission correctness for the same poll cycle is unaffected**: `setPartitionStoppingOffset(tp, stoppingOffset)` is called on `recordsBySplits` before `finishSplitAtRecord`, exactly as in the pre-fix code, so `KafkaPartitionSplitRecords.nextRecordFromSplit()` (:409-423, unchanged) still correctly emits every record in *this* poll's batch with `offset < stoppingOffset` before the split is reported finished — the fix only changes how completion is *detected*, not what gets emitted.

Overall: this is a correct, narrowly-scoped fix that replaces a heuristic based on visible output with a check based on the Kafka client's own authoritative fetch-position bookkeeping. I have no correctness objection.

## 1.2 Compatibility Impact

**Fully compatible.** No config option, default, split serialization format, checkpoint state, or public API changes. Streaming (unbounded) partitions take the exact same `Long.MAX_VALUE`-skip path as before and never call the new `position()` branch, so unbounded-source behavior, including any downstream isolation-level or commit behavior, is untouched. The only behavior change is that *bounded* partitions can now finish in strictly more cases than before (whenever a control/filtered record was the actual boundary crosser), which is the intended fix, not a regression.

One extra cached `position()` lookup per currently-assigned *finite* partition per `fetch()` call; no added broker round trip in the steady state (see 1.1), no growth in buffered records, no new locks, no new retry semantics beyond the existing `retryOnWakeup` pattern that is already used elsewhere in this same class (e.g. `removeEmptySplits()` at :228-230, and the debug-logging path at :211-213) — this call is consistent with, not novel relative to, the class's existing conventions. `WakeupException` handling is preserved: `retryOnWakeup` retries exactly once and lets a second `WakeupException` propagate, matching the pattern already established for the other `position()` callers in this file.

No exception is swallowed or misclassified into a false "finished" state — `position(tp)` failures propagate normally through `retryOnWakeup`/`fetch()` rather than being caught and treated as completion. No sensitive data is logged. See Issue 1 below for the one open finding, carried forward from the prior review.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Title/commit format follows `[Fix][Connector-V2]`. The production diff is tightly scoped to the two files it needs to touch. No new public method or field was added without a comment — the new branch is inline within the existing `fetch()` method and is accompanied by an explanatory comment (:116-118) describing *why* position is used instead of the last returned record, which is exactly the kind of "explain the non-obvious boundary condition" comment this class of fix needs.

`shouldFinishSplitWhenConsumerPositionReachesStoppingOffsetAfterControlRecord` (`KafkaPartitionSplitReaderTest.java:81-109`) is a deterministic, mock-based regression: `consumer.position(partition)` is stubbed to return `0L` then `101L`, `poll()` returns an empty `ConsumerRecords`, and the test asserts the split is reported finished. No `Thread.sleep`, no timing dependency, no shared static state, no network — I rate this **Stable**.

**Carried-forward, not new** — this is the same Low-severity issue Daniel flagged on July 26 and it is still present verbatim in the current head (the diff between the reviewed commit and this head is empty, so there was no opportunity for it to be fixed or regress): the test constructs a real `KafkaPartitionSplitReader` via its production constructor, which creates a real `KafkaConsumer` (`initConsumer`, :346-355) against `localhost:1`, and then reflectively overwrites the private final `consumer` field with a Mockito mock (`setConsumer`, :119-125 in the test) — the original real consumer reference is discarded without ever being closed, and the test also never calls `reader.close()` at all. `KafkaConsumer`'s constructor doesn't block on a real network connection, so this won't hang, but it does allocate real client-side resources (metrics registry entries, `AppInfoParser` MBean registration under the "seatunnel-consumer-0" client id) that are never released, which is a latent noisy-neighbor / same-JVM test-order risk rather than a functional bug in the fix itself. Restating this as still-open at Low/non-blocking, not re-raising it as new.

**Missing but non-blocking**: no real-broker/E2E coverage of an actual Kafka transaction (produce + abort/commit + consume with `isolation.level=read_committed`) exists for this fix; the unit test is a faithful simulation of the `position()`-vs-`poll()` gap but doesn't exercise a real broker's transaction-marker behavior end-to-end. Given the fix is narrowly scoped to client-documented `KafkaConsumer` semantics and the unit test isolates exactly the mechanism being changed, I don't consider this blocking, but it remains a gap if anyone wants stronger integration evidence later.

## 2.3 Documentation

Not applicable — internal completion-detection fix with no new user-facing configuration, default, or documented behavior. I agree with SEZ9's and my own prior "Issue 1" framing that this specific PR has no option surface to document; that note was about staying alert to doc updates in general, not a claim that this diff needs one.

# 3. Architectural Soundness

**Precise fix.** It replaces a heuristic tied to `poll()`'s visible output with a check against the Kafka client's own authoritative bookkeeping (`position()`), in the exact layer (`KafkaPartitionSplitReader`) that already owns both the stopping-offset map and the consumer instance. No workaround, no new abstraction, no leakage into the engine or common reader layers.

## 3.2 Maintainability

High — the change is contained to one method, keeps the existing `retryOnWakeup`/`finishSplitAtRecord`/`unassignPartitions` helpers intact, and the added comment explains the non-obvious "position advances across records `poll()` never returns" invariant for future readers.

## 3.3 Extensibility

No new coupling introduced; the `Long.MAX_VALUE` unbounded-skip guard keeps this fix cleanly opt-in for bounded splits only, so any future stopping-offset strategy that reuses `getStoppingOffset()` inherits the fix automatically.

No serialization, checkpoint, or split-state format changed, so upgrade/restore compatibility for already-running or previously-checkpointed jobs is unaffected.

# 4. Issue Summary

|---|---|---|---|
| 1 | Regression test discards a real `KafkaConsumer` (created via the production constructor) via reflective field replacement, and never calls `reader.close()`; resources are never released. Carried forward from the July 26 review — unchanged in the current head. | `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:81-125` | Low |

No blocking issues found. The one open item is a test-hygiene nit, not a correctness or compatibility concern.

# 5. Merge Recommendation

### Conclusion: Ready to merge after fixes (non-blocking) / effectively ready pending CI

I agree with SEZ9's August 4 "can merge" conclusion and see nothing that changes that on this fresh head — the diff between what SEZ9 and I both reviewed and the current head is empty, so their runtime-chain analysis and mine independently converge on the same conclusion using the same evidence.

1. **Blockers** — none.
2. **Recommended fixes (non-blocking)**:
   - Issue 1: give the test an injectable consumer (a package-private constructor/factory overload, or a setter used consistently) instead of reflecting into a private final field, and either close the original real consumer before replacing it or call `reader.close()` at the end of the test.

**Overall assessment**: this is a precise, well-targeted fix for a real bug (#11471) — a bounded Kafka source could hang indefinitely when a transaction control record was the thing that pushed the consumer past its stopping offset, because completion detection previously relied only on the last *visible* record. Using `KafkaConsumer.position()` instead is the technically correct fix given the documented Kafka client contract, it's cheap (cached lookup, skipped entirely for unbounded partitions), and it doesn't disturb any existing completion, emission, or checkpoint path. The only remaining gate is CI: `Build` is `QUEUED` on this fresh head as of this review. No alternative implementation is needed — the position-based approach is the right one; I'd only suggest the test-injection cleanup above as a follow-up, not as a condition for merging once CI is green and a committer completes the write-side approval (Daniel's account here has read-only permission, so this review's conclusion is a technical recommendation, not a GitHub approval).

DanielLeens (approved):
Circling back on this — my last review's conclusion was already 'ready to merge' with no blockers (the one open item, a test-hygiene nit about an unclosed KafkaConsumer in the regression test, was explicitly non-blocking). The only remaining gate I flagged was CI: `Build` was still queued on the fresh head (`a0376c84`). Rechecking now: `Build`, `Notify test workflow`, and `labeler` are all green on this exact head. I don't see any other outstanding concern from my side, so updating my review to approve.

SEZ9 (commented):
Thanks for working on this. I reviewed the latest head `a0376c8499db`.

# What this PR fixes
- User pain: Batch Kafka jobs on transactional topics never finish — a transaction control record advances the consumer position past the stopping offset without appearing in `poll()` results, so the last-record completion check in `fetch()` never fires and the split reader polls forever (#11471).
- Fix approach: Rework the completion check in `KafkaPartitionSplitReader.fetch()` to iterate `consumer.assignment()` instead of `consumerRecords.partitions()`, skip partitions with an unbounded (`Long.MAX_VALUE`) stop, and compare `consumer.position(tp)` (wrapped in `retryOnWakeup`) against the stopping offset so the split finishes even when the position was advanced by invisible control records.
- One-line summary: The fix is directionally correct and the diff is small and well-commented, but `consumer.position()` is now an unbounded, potentially blocking/throwing call executed per finite-stop partition on every fetch, `finishSplitAtRecord` now receives the next-fetch position rather than the last consumed offset (shifting the reported offset by one or more), and the regression test only covers the empty-poll path while leaking the real `KafkaConsumer` behind a reflection-injected mock with a call-count-coupled `thenReturn(0L, 101L)` stub.

# Runtime chain I rechecked
```text
SplitFetcher.runOnce()
  -> FetchTask.run()
    -> KafkaPartitionSplitReader.fetch()  KafkaPartitionSplitReader.java:99
      -> consumer.poll(pollTimeout)  KafkaPartitionSplitReader.java:100 (via retryOnWakeup)
      -> for tp in consumer.assignment()  KafkaPartitionSplitReader.java:106  [changed: was consumerRecords.partitions()]
        -> getStoppingOffset(tp)  KafkaPartitionSplitReader.java:107
        -> skip if stoppingOffset == Long.MAX_VALUE  KafkaPartitionSplitReader.java:108-110
        -> retryOnWakeup(() -> consumer.position(tp), ...)  KafkaPartitionSplitReader.java:111-114  [changed: new blocking position lookup]
        -> if currentOffset >= stoppingOffset  KafkaPartitionSplitReader.java:119
          -> recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset)  KafkaPartitionSplitReader.java:120
          -> finishSplitAtRecord(tp, stoppingOffset, currentOffset, finishedPartitions, recordsBySplits)  KafkaPartitionSplitReader.java:121-122  [changed: passes position, not lastRecord.offset()]
      -> unassignPartitions(finishedPartitions)  KafkaPartitionSplitReader.java:127
    -> RecordsWithSplitIds.finishedSplits() consumed by SourceReaderBase to signal split completion upstream
```

# Findings
**Issue 1: consumer.position(tp) inside fetch() can block for default.api.timeout.ms or throw when the position is not locally known; call is unbounded and unguarded**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:113`
- Why it matters: - Root cause: the new completion check replaces a purely local inspection of `consumerRecords` with a call to `KafkaConsumer.position(tp)` for every assigned bounded partition on every fetch. `position()` is only an in-memory read when a position has already been established via seek/poll; if the consumer reaches this line with an undefined position (e.g. a split whose start offset is resolved lazily from committed offsets, or a partition assigned but not yet sought), it performs a synchronous offset lookup bounded by `default.api.timeout.ms` (60s default) and can throw `NoOffsetForPartitionException` when `auto.offset.reset=none`.
- Impact: a single partition with an unresolved position can stall the SplitFetcher thread for up to a minute per fetch cycle or fail the reader outright, and the call runs unguarded inside the hot `fetch()` loop.
- Suggested fix: use the timeout-bounded overload `consumer.position(tp, Duration.ofMillis(pollTimeout))` and treat a lookup failure as "split not finished yet" rather than letting the exception propagate out of `fetch()`.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:111-114  long currentOffset = retryOnWakeup( () -> consumer.position(tp), "getting current position to finish split");`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:111  long currentOffset =`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:113  () -> consumer.position(tp),`
- Suggested fix: `consumer.position(tp)` is only a local read when the position is already known; otherwise it does a synchronous offset lookup bounded by `default.api.timeout.ms` (60s default) and can throw `NoOffsetForPartitionException` with `auto.offset.reset=none`. Since this now runs for every assigned bounded partition on every fetch, a single partition with an unresolved position can stall the SplitFetcher thread or fail the reader. Please use the timeout-bounded overload `consumer.position(tp, Duration.ofMillis(pollTimeout))` and treat a lookup failure as "split not finished yet" instead of letting it propagate out of `fetch()`.
- Severity: Medium

**Issue 2: Regression test only covers the empty-poll case; no test that final visible records are still delivered (not trimmed away) when the split finishes in the same fetch, and no IT against a real transactional topic**
- Location: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:92`
- Why it matters: - Root cause: the new test stubs `consumer.poll(...)` to always return `ConsumerRecords.empty()` and asserts only `finishedSplits()`. The actual #11471 scenario is a poll that returns the last visible records while the position jumps past the stopping offset over a control record — in that path the production code both sets `recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset)` (line 120) and finishes the split, and nothing verifies the returned records survive the stopping-offset trim.
- Impact: a regression in the trim logic could silently drop the tail records of bounded jobs without any test catching it, and the transactional-topic scenario the fix targets has no integration coverage.
- Suggested fix: add a case where `poll()` returns non-empty `ConsumerRecords`, asserting both that records below the stopping offset are still returned from `RecordsWithSplitIds` and that the split finishes; ideally also cover a real transactional topic in an IT.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:103  Assertions.assertTrue(records.finishedSplits().contains(partition.toString()));`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:120  recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset);`; `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:94  KafkaPartitionSplitReader reader = new KafkaPartitionSplitReader(sourceConfig, context);`
- Suggested fix: This test only exercises the empty-poll path. The actual #11471 scenario is a poll that returns the final visible records while the position jumps past the stopping offset over a control record — in that path `setPartitionStoppingOffset` and the finish happen in the same fetch. Please add a case where `poll()` returns non-empty `ConsumerRecords` and assert both that records below the stopping offset are still returned from `RecordsWithSplitIds` and that the split finishes, so a trimming regression can't silently drop tail records of bounded jobs.
- Severity: Medium

**Issue 3: Iterating consumer.assignment() may re-evaluate already-finished partitions on every fetch unless they are removed from both the assignment and the stopping-offset map**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:106`
- Why it matters: - Root cause: the loop now walks `consumer.assignment()` instead of `consumerRecords.partitions()`, so every fetch cycle touches all assigned partitions, including ones already reported as finished in a prior cycle.
- Impact: if `finishSplitAtRecord(...)` does not both unassign the partition from the consumer and remove its entry from the stopping-offset map, the `stoppingOffset == Long.MAX_VALUE` guard will not short-circuit, and the reader will call `consumer.position(tp)` and re-report the same split in `finishedSplits` on every subsequent `fetch()`.
- Suggested fix: verify `finishSplitAtRecord` cleans up both the assignment and the stopping-offset map, and add a second `reader.fetch()` assertion to the regression test (expecting no finished splits) to lock the behavior in.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:119  if (currentOffset >= stoppingOffset) {`
- Suggested fix: Switching from `consumerRecords.partitions()` to `consumer.assignment()` means every fetch cycle now visits all assigned partitions. Please make sure `finishSplitAtRecord` removes the finished partition from the assignment AND from the stopping-offset map — otherwise the `stoppingOffset == Long.MAX_VALUE` guard won't skip it, and we will call `consumer.position(tp)` and re-report the same split in `finishedSplits` on every subsequent `fetch()`. Adding a second `reader.fetch()` assertion in the new regression test (expecting no finished splits) would cover this.
- Severity: Medium

**Issue 4: Reflection-injected mock leaks the real KafkaConsumer created by the constructor and couples the test to internal position() call counts**
- Location: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:133`
- Why it matters: - Root cause: `new KafkaPartitionSplitReader(sourceConfig, context)` at line 94 constructs a real `KafkaConsumer` against `localhost:1`, which is then discarded (never closed) when `setConsumer` overwrites the private field via reflection (lines 131-133). Additionally, `Mockito.when(consumer.position(partition)).thenReturn(0L, 101L)` (line 90) hard-codes the exact number of internal `position()` invocations across `handleSplitsChanges` and `fetch()`.
- Impact: the orphaned consumer leaks threads and sockets in the test JVM, and any refactor that adds or removes a `position()` call shifts the stubbed return values, making the test break — or pass — for the wrong reason.
- Suggested fix: read the existing field value and `close()` it before injecting the mock (or add a package-private test constructor), and replace the ordered stub with a `doAnswer` keyed on split state.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:94-95  KafkaPartitionSplitReader reader = new KafkaPartitionSplitReader(sourceConfig, context); setConsumer(reader, consumer);`; `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:131-133  Field consumerField = KafkaPartitionSplitReader.class.getDeclaredField("consumer"); consumerField.setAccessible(true); consumerField.set(reader, consumer);`
- Suggested fix: The constructor already created a real `KafkaConsumer` (bootstrap `localhost:1`) which is leaked when this reflection call overwrites the field. Please read the existing field value and `close()` it before setting the mock (or add a package-private test constructor). Also note `thenReturn(0L, 101L)` at line 90 couples the test to the exact internal `position()` call count — a `doAnswer` stub keyed on split state would be more robust to refactoring.
- Severity: Low

**Issue 5: finishSplitAtRecord now receives the consumer position (next-fetch offset) instead of the last consumed record offset, shifting the reported offset semantics**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:122`
- Why it matters: - Root cause: the old call passed `lastRecord.offset()` (the offset of the last record actually consumed); the new call at lines 121-122 passes `currentOffset`, which is `consumer.position(tp)` — i.e. the *next* offset to fetch, at least last-record-offset + 1 and potentially higher when control records were skipped.
- Impact: whatever `finishSplitAtRecord` does with its offset argument (at minimum logging) now reports an offset one or more higher than the last consumed record, which is misleading in diagnostics and risky if the value feeds into split state or metrics.
- Suggested fix: pass `currentOffset - 1` to preserve the historical semantics, or rename the parameter and log wording to make clear it is the next-fetch position, and confirm the value is not persisted anywhere.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:121  finishSplitAtRecord(`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:122  tp, stoppingOffset, currentOffset, finishedPartitions, recordsBySplits);`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:121-122  finishSplitAtRecord(tp, stoppingOffset, currentOffset, finishedPartitions, recordsBySplits);`
- Suggested fix: Previously the third argument here was `lastRecord.offset()` (last consumed record); now it's the consumer position, which is at least one higher and can jump further past control records. If `finishSplitAtRecord` logs or records this as the "last offset", the reported value is now off by one or more. Please pass `currentOffset - 1` to keep the historical semantics, or rename the parameter/log wording to make clear it is the next-fetch position — and double-check the value isn't persisted into split state or metrics.
- Severity: Low

**Issue 6: Kafka source docs could briefly note bounded-read behavior with transactional topics fixed by this PR**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:116  // Consumer positions advance across Kafka transaction control records, although those`
- Why it matters: - Root cause: this PR changes user-visible behavior of bounded Kafka reads on transactional topics (batch jobs now terminate after control records advance the position), but there is no accompanying update to `docs/en/connectors/source/Kafka.md` / `docs/zh/connectors/source/Kafka.md`.
- Impact: no new option is introduced, so this is not a hard documentation gate; however, users who previously worked around hanging batch jobs on transactional topics would benefit from a short note that bounded reads now terminate correctly.
- Suggested fix: add one sentence to both language variants of the Kafka source docs stating that bounded/batch reads handle transactional control records and terminate at the stopping offset.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReader.java:118  // record so a batch split can finish after consuming the visible records.`
- Severity: Low

**Issue 7: New test method and setConsumer helper lack the brief Javadoc used elsewhere in this test class**
- Location: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:128`
- Why it matters: - Root cause: the pre-existing helper in this class carries a short Javadoc (`/** Creates the minimum Kafka source config used by the tests. */`), but the newly added regression test and the reflection helper `setConsumer` have none.
- Impact: purely readability — the reflection-based field injection is non-obvious, and `getDeclaredField("consumer")` couples the test to the production field name, so a one-line Javadoc explaining that it replaces the internally constructed consumer with a mock would help future maintainers.
- Suggested fix: add one-line Javadoc to the new test method and `setConsumer`, matching the existing style in the class.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/source/KafkaPartitionSplitReaderTest.java:131  Field consumerField = KafkaPartitionSplitReader.class.getDeclaredField("consumer");`
- Suggested fix: Nit: please add a one-line Javadoc here (matching the style of the existing `createConfig` helper) noting that this replaces the reader's internally created consumer with a mock via reflection — `getDeclaredField("consumer")` couples the test to the production field name, so documenting that intent will help whoever refactors `KafkaPartitionSplitReader` later.
- Severity: Low

# Review conclusion

**2. Remaining follow-ups**
- Issue 1: consumer.position(tp) inside fetch() can block for default.api.timeout.ms or throw when the position is not locally known; call is unbounded and unguarded
- Issue 2: Regression test only covers the empty-poll case; no test that final visible records are still delivered (not trimmed away) when the split finishes in the same fetch, and no IT against a real transactional topic
- Issue 3: Iterating consumer.assignment() may re-evaluate already-finished partitions on every fetch unless they are removed from both the assignment and the stopping-offset map
- Issue 4: Reflection-injected mock leaks the real KafkaConsumer created by the constructor and couples the test to internal position() call counts
- Issue 5: finishSplitAtRecord now receives the consumer position (next-fetch offset) instead of the last consumed record offset, shifting the reported offset semantics
- Issue 6: Kafka source docs could briefly note bounded-read behavior with transactional topics fixed by this PR
- Issue 7: New test method and setConsumer helper lack the brief Javadoc used elsewhere in this test class

DanielLeens:
Thanks @SEZ9 for the fresh pass — I rechecked the current head (`a0376c8499db`) against these findings directly rather than taking my own August 14 approval for granted.

**Issue 3** (re-evaluating already-finished partitions on every fetch) doesn't hold up against the actual control flow. `finishSplitAtRecord` only adds `tp` to `finishedPartitions` and does not itself unassign it, but `fetch()` calls `unassignPartitions(finishedPartitions)` synchronously before it returns:

```java
if (!finishedPartitions.isEmpty()) {
    unassignPartitions(finishedPartitions);
}
return recordsBySplits;
```

`unassignPartitions` calls `consumer.assign(newAssignment)` with the finished partitions removed, in the same `fetch()` invocation that finished them. So by the time the next `fetch()` runs, `tp` is no longer in `consumer.assignment()` and the loop never revisits it — there's no re-evaluation or repeated `finishedSplits` reporting. The one real (but different, and much lower severity) gap is that the entry is never removed from the `stoppingOffsets` map itself, which is a harmless retention rather than a functional re-processing bug.

**Issue 1** (`consumer.position(tp)` as an unbounded/blocking call) is worth a note but I'd call it a hardening suggestion rather than a live bug on this codebase: the loop only ever runs over `consumer.assignment()`, and `handleSplitsChanges()` always calls `seekToStartingOffsets()` synchronously before a split is considered assigned, so in normal operation `position(tp)` should always resolve from the already-known `SubscriptionState` rather than triggering a real lookup. Using the timeout-bounded overload defensively is still a reasonable ask, just not something I can attribute an actual stall/throw scenario to in the current code.

Issues 2, 4, 6, and 7 (empty-poll-only test coverage, the reflection-injected mock leaking the real consumer, the missing docs note, and the missing test-method Javadoc) look like fair non-blocking observations to me on inspection, consistent with the Low-severity test-hygiene item already carried forward from earlier rounds. Issue 5 (the `currentOffset` vs `lastRecord.offset()` semantic shift) is also fair as flagged — I checked and `finishSplitAtRecord`'s `currentOffset` parameter is only used in a `LOG.debug` line today, so it's cosmetic rather than something feeding checkpoint/split state, matching the Low severity you gave it.

So my read is: the two source-level concerns significant enough to matter (Issues 1 and 3) don't change my prior "no source-side blocker" conclusion once checked against the actual current-head control flow, and the rest are reasonable non-blocking follow-ups already in the same spirit as what's been carried forward. Appreciate you pushing on this — happy to keep digging if you see something I'm missing in the assignment/unassignment sequencing.

SEZ9:
Thanks @DanielLeens for re-verifying against the actual code — much appreciated.

**Issue 3**: withdrawn as originally stated. I missed that `fetch()` calls `unassignPartitions(finishedPartitions)` before returning, so the finished partition is out of `consumer.assignment()` by the next `fetch()` and there's no re-evaluation or repeated `finishedSplits` reporting. The residual point you identified — the entry never being removed from the `stoppingOffsets` map — is a harmless retention issue. Downgrading to a Low, non-blocking cleanup ask: remove the entry alongside the unassignment.

**Issue 1**: fair analysis. Since the loop only iterates `consumer.assignment()` and `handleSplitsChanges()` runs `seekToStartingOffsets()` synchronously before a split is considered assigned, `position(tp)` should resolve locally in normal operation. Reclassifying as a defensive-hardening suggestion — a timeout-bounded position call would be a cheap safety net, but it does not block merge.

**Issues 2, 4, 6, 7**: agreed these are non-blocking. The one I'd most like to see before merge is Issue 2 — a test asserting final visible records are still delivered when the split finishes in the same fetch, ideally plus an IT against a real transactional topic, since that's the core behavior this PR fixes.

**Issue 5**: agreed — with `currentOffset` only feeding a `LOG.debug` line, cosmetic and Low is the right call.

Remaining asks:
1. (Preferred pre-merge) The final-records-delivered test per Issue 2; the transactional-topic IT can follow up separately.
2. Remove the `stoppingOffsets` entry when a partition finishes (revised Issue 3).
3. Optional: bounded position call (Issue 1), close the leaked consumer in the test (Issue 4), docs note (Issue 6), brief Javadoc on the new test method and helper (Issue 7).

If item 1 lands, I'm happy with this going in. Thanks again for the thorough push-back — it materially improved the review.

DanielLeens:
Agreed on all counts, @SEZ9 -- thanks for closing the loop so precisely. To summarize where this leaves us: Issue 3 is fully withdrawn (superseded by the harmless `stoppingOffsets` retention cleanup), Issue 1 is now a defensive-hardening suggestion rather than a blocker, and Issues 2/4/5/6/7 stand as non-blocking follow-ups, with Issue 5 confirmed cosmetic since `currentOffset` only feeds a `LOG.debug` line.

I agree Issue 2 (a test proving the final visible records are still delivered when a split finishes mid-fetch) is the one item worth holding for before merge, since it is the core behavior this PR fixes and today's coverage is empty-poll-only. The `stoppingOffsets` cleanup is a one-line addition alongside the existing `unassignPartitions` call, so bundling both in the same push should be low-effort.

@zhangshenghang -- over to you: once the final-records test and the `stoppingOffsets` removal land, both reviewers are aligned to approve. The rest (bounded `position()` call, closing the leaked consumer in the test, docs note, test Javadoc) remain nice-to-haves that don't block.

SEZ9:
Thanks for the crisp summary, @DanielLeens — it matches my read exactly, and to answer you directly: yes, I'm aligned on all points, and nothing has landed yet, so both blocking items are still outstanding.

To restate the merge criteria so there's no ambiguity:

1. **Issue 2 (blocking):** a test proving the final visible records are still delivered — not trimmed away — when a split finishes in the same fetch. Agreed this is the core behavior the PR fixes, and the current empty-poll-only coverage doesn't exercise it.
2. **Issue 3 follow-through (blocking, trivial):** the one-line `stoppingOffsets` retention cleanup alongside the existing `unassignPartitions` call. Agreed the original Issue 3 concern is withdrawn and superseded by this cleanup, so bundling it in the same push as the test makes sense.

Also confirming the rest of the disposition: Issue 1 (bounded/guarded `position()` call) is a defensive-hardening suggestion rather than a blocker; Issue 5 is cosmetic since `currentOffset` only feeds a `LOG.debug` line; and Issues 4 (closing the consumer leaked by the constructor in the reflection-injected test), 6 (docs note on bounded reads with transactional topics), and 7 (test Javadoc) remain non-blocking follow-ups — happy to see any of them in this PR if convenient, but none gate approval.

Once a push lands with the final-records test and the `stoppingOffsets` removal, I'll take a final pass and approve. Nothing else needed from my side.

DanielLeens:
@SEZ9 Agreed on all counts — that's a precise restatement of where we landed. Nothing to add on Issues 1/3/5 or the non-blocking follow-ups (4/6/7).

Waiting on @zhangshenghang to push the final-records-delivered test (Issue 2) plus the one-line `stoppingOffsets` cleanup (revised Issue 3); once that lands I'll take the final pass alongside you.

PR #11456, opened 2026-07-15
[Fix][E2E] Stabilize flaky connector E2E tests

nzw921rx:
### Purpose

Reduce PR-independent connector E2E CI flakiness by consolidating related fixes into one module-scoped PR instead of many small PRs.

### Changes

- Resolve required JDBC and connector dependency jars from the active test classpath instead of downloading them from Maven Central inside CI containers.
- Copy resolved jars into the corresponding SeaTunnel plugin lib directories before E2E jobs run.
- Replace brittle fixed sleeps and one-shot status checks with bounded waits for connector E2E cases.
- Cover StarRocks, Pulsar, CDC MySQL/Postgres/OpenGauss/Oracle/MongoDB/TiDB, EdgeSocket, Kafka, Doris, RocketMQ, SFTP, Databend, and HTTP E2E tests.
- Keep helper methods scoped to their E2E modules or existing abstract test classes.

### User-facing

No user-facing behavior change. This only stabilizes E2E test setup.

### How tested

- Ran ./mvnw spotless:apply
- Did not run tests per maintainer request

PR #11487, opened 2026-07-18
[Feature][Connector-V2] Support snapshot-only startup mode for MySQL CDC

### LINKED ISSUE #11036: [Feature][Connector-V2][MySQL-CDC] Add snapshot-only startup mode

fatmanverse:

Closes #11036.

This pull request adds a dedicated `startup.mode = "snapshot-only"` option for bounded MySQL CDC bootstrap jobs. It reuses the existing snapshot split planning and reading path, reports the source as bounded, and finishes without creating an incremental split or entering continuous binlog streaming.

The snapshot-only completion path depends only on snapshot split completion, so parallel batch jobs can finish without requiring a checkpoint. When restoring from a checkpoint, the checkpointed table set remains authoritative and catalog discovery is skipped, so tables created or removed after the checkpoint do not change the bounded input.

`exactly_once` remains supported. It may perform bounded low-to-high-watermark binlog backfill during the snapshot phase, but snapshot-only mode still finishes after all snapshot splits complete. Incompatible stop modes, timestamp startup, and specific-offset startup options fail before source initialization.

Yes.

Previously, MySQL CDC could either read a snapshot and continue into binlog streaming, or start from a binlog-oriented position. There was no first-class bounded snapshot-only mode.

Users can now configure:

```hocon
startup.mode = "snapshot-only"
```

The source reads the configured snapshot and exits successfully without consuming subsequent binlog events. Existing startup modes and other CDC connectors keep their current behavior.

- `./mvnw spotless:apply`
- CDC base unit tests: 67 passed
- MySQL source factory tests: 2 passed
- `MysqlCDCSnapshotOnlyIT`: 5 runtime invocations passed across Flink 1.13.6, 1.15.3, 1.18.0, 1.20.1, and Zeta
- The E2E used parallelism 2, checkpointing disabled, a legal `server-id = "5657-5658"` range, 103 rows, and `snapshot.split.size = 1` to exercise multiple readers and controlled concurrent updates/deletes
- `git diff --check`
- Repository-wide `./mvnw -q -DskipTests verify` was attempted and remains blocked by existing `connector-jdbc` compilation failures on `dev`, including the duplicate `setNullToStatementByDataType` method and missing generated symbols
- Repository-wide `./mvnw test` was attempted and remains blocked by the existing `seatunnel-config-shade` shaded Typesafe Config `ClassNotFoundException`

The Docker E2E used the command-line-only `-Dtestcontainer.version=1.21.4` override required by Docker 29. No dependency or POM change is included.

### Check list

* [x] No new JAR binary package is added.
* [x] Focused unit and MySQL CDC E2E coverage is included.
* [x] Plugin mapping, distribution dependencies, labels, and plugin configuration do not require changes for this existing connector.


fatmanverse:
Additional local E2E environment details:

- Docker Server: 29.4.0 on OrbStack
- The repository default Testcontainers 1.17.6 negotiates Docker API 1.32, which Docker 29 no longer accepts.
- The E2E was therefore run with a command-line-only override: `-Dtestcontainer.version=1.21.4`. No dependency or POM change is included in this PR.
- `MysqlCDCSnapshotOnlyIT` completed all 5 runtime invocations with 0 failures and 0 errors. The Zeta run reported 3 records read, 3 written, and 0 failed.

DanielLeens (commented):
Thanks for the contribution! I re-reviewed the full current diff and the relevant snapshot-state regression coverage runtime path.

I do not see a source-side blocker on the current head. The current `Build` failure is the shared engine `MetricsApiTest` baseline; CI and repository review remain outstanding.

SEZ9 (approved):
Thanks for working on this. I re-reviewed the latest head `991e997e72cb` from scratch.

# Review conclusion

nzw921rx (changes requested):
Thanks for adding the bounded snapshot-only mode. I reviewed the current head `991e997e72cb`. I do not think this PR is ready to merge yet.

### 1. Blocker: keep the checkpointed table set authoritative during snapshot-only restore

`restoreEnumerator()` still reconciles `HybridPendingSplitsState` with the latest result of `discoverDataCollections()` before constructing the snapshot-only assigner.

This can add newly discovered tables to `remainingTables` or remove unfinished checkpointed tables. Consequently, the input of a bounded snapshot job can change after failover.

For example, a table matching `table-pattern` that is created after a checkpoint will unexpectedly be included after recovery. This conflicts with #11036, which explicitly lists dynamic newly-added table capture as a non-goal.

Please make the checkpointed table set authoritative for snapshot-only recovery and add a checkpoint/restore test proving that catalog changes do not alter the bounded input set.

### 2. Blocker: do not download the JDBC driver from inside the runtime container

The new E2E downloads MySQL Connector/J using `wget` from inside every SeaTunnel runtime container.

This makes the test depend on container DNS and outbound network access, bypasses Maven mirrors, retries and the local repository, and can easily cause flaky CI failures.

The test also hard-codes Connector/J `8.0.32`, while the Maven-managed test dependency currently resolves to `8.0.27`. The test runtime therefore uses a different driver from the declared dependency.

Current `dev` already provides `MysqlCDCDriverResolver`, introduced by merged PR #11456. Please rebase and reuse the Maven-resolved driver:

    @TestContainerExtension
    protected final ContainerExtendedFactory extendedFactory =
            MysqlCDCDriverResolver::copyMySQLDriverToContainer;

This resolves `com.mysql.cj.jdbc.Driver` from the Maven test classpath and copies its JAR into the container with `copyFileToContainer`, removing the unnecessary network dependency and keeping Maven as the single source of the driver version.

### 3. Preserve the existing public constructor overloads

Both existing public `HybridSplitAssigner` constructors were replaced with signatures containing the new `boolean snapshotOnly` argument.

External or custom CDC connectors compiled against the old constructors can fail with `NoSuchMethodError`, while source recompilation also fails.

Please retain the old overloads and delegate them with `snapshotOnly = false`.

### 4. Complete the snapshot configuration validation

`startup.mode = snapshot` rejects `startup.specific-offset.*`, but accepts and silently ignores `startup.timestamp`.

Please reject `startup.timestamp` as well and add a negative configuration test.

Also, `validateSnapshotOnlyConfig()` currently runs after datasource, dialect and deserializer initialization. MySQL schema discovery may therefore fail before the expected configuration validation. Please move this validation before database-related initialization.

### Test coverage

- The E2E changes the source only after `executeJob()` has returned. A terminated job cannot consume those changes, so this does not prove that an incremental reader was never started.
- Only three rows are inserted while `snapshot.split.size` remains at its default. This normally creates one snapshot split despite `parallelism = 2`. Please set `snapshot.split.size = 1` or insert enough rows to exercise multiple snapshot readers.
- `exactly_once` is also used for bounded low-to-high-watermark backfill during snapshot reading. If snapshot-only must reject it, please document the concurrent-write consistency boundary or require the source to be quiesced, and add a controlled concurrent-write test.

For these reasons, I recommend requesting changes.

fatmanverse:
@nzw921rx Thanks for the detailed review. I addressed the requested changes in commit `e22ab12d7`.

1. Snapshot-only restore now treats the checkpointed table set as authoritative. It collects tables from processed tables, remaining tables, remaining snapshot splits, and assigned snapshot splits, and does not call catalog discovery on the snapshot-only restore path. Legacy checkpoint state is also marked as having checkpointed remaining tables before the snapshot assigner opens, so it cannot rediscover the catalog. The new unit coverage verifies that a newly created table is excluded after restore.

2. The E2E now reuses `MysqlCDCDriverResolver` and copies the Maven-resolved Connector/J into the container. It no longer downloads a driver from inside the container, and the test uses the declared 8.0.27 driver.

3. Both existing public `HybridSplitAssigner` constructors are retained and delegate to the new overloads with `snapshotOnly = false`.

4. Snapshot-only validation now runs before datasource, dialect, deserializer, and offset initialization. It rejects stop modes other than `never`, `startup.specific-offset.*`, and `startup.timestamp`. The negative test verifies that `startup.timestamp` fails before source initialization.

5. The E2E seeds 103 rows and sets `snapshot.split.size = 1`, so multiple snapshot splits are exercised. It starts the job asynchronously, waits until rows from early splits are visible while later splits are still outstanding, updates and deletes rows from completed snapshot ranges, and verifies that the sink remains exactly the original snapshot result. The test passed all 5 runtime invocations: Flink 1.13.6, 1.15.3, 1.18.0, 1.20.1, and Zeta. The Zeta run reported 103 records read, 103 written, and 0 failed.

6. `exactly_once` remains supported for both `initial` and `snapshot` startup modes. The snapshot mode documentation now describes the bounded low-to-high-watermark backfill semantics and the controlled concurrent-write E2E coverage. Snapshot-only still never enters continuous binlog streaming.

Validation completed:
- `./mvnw spotless:apply`
- CDC base targeted tests: 6 tests passed
- MySQL factory tests: 2 tests passed
- `MysqlCDCSnapshotOnlyIT`: 5 tests passed
- `git diff --check`

The repository-wide `-DskipTests verify` is still blocked by pre-existing `origin/dev` compilation errors in `connector-jdbc`, including the duplicate `setNullToStatementByDataType` method and missing generated methods. No `connector-jdbc` files are changed in this PR. The full test run is also blocked by the existing `seatunnel-config-shade` shaded Typesafe Config `ClassNotFoundException`.

fatmanverse:
Follow-up audit fixes are included in `aa2705213`.

- The snapshot-only E2E now uses the legal `server-id = "5657-5658"` range for parallelism 2.
- The restore regression now calls `restoreEnumerator()` and `open()` directly, verifies that snapshot-only recovery never calls `discoverDataCollections()`, and confirms legacy state is marked with checkpointed remaining tables.
- Negative coverage now checks `stop.mode`, `startup.timestamp`, `startup.specific-offset.file`, and `startup.specific-offset.pos` independently before source initialization.
- The option description and both MySQL CDC FAQ pages now distinguish bounded snapshot backfill from continuous binlog streaming. The PR description has been updated to match the implementation and current validation status.

Validation on the latest head:
- CDC base tests: 67 passed
- MySQL factory tests: 2 passed
- `MysqlCDCSnapshotOnlyIT`: 5 passed across Flink 1.13.6, 1.15.3, 1.18.0, 1.20.1, and Zeta
- `./mvnw spotless:apply` and `git diff --check` passed

Repository-wide `verify` is still blocked by the existing `connector-jdbc` compilation failures on `dev`. The full test run is still blocked by the existing `seatunnel-config-shade` shaded Typesafe Config `ClassNotFoundException`. Neither failing module is changed by this PR.

SEZ9 (approved):
Thanks for working on this. I re-reviewed the latest head `e22ab12d78ed` from scratch.

# Review conclusion

DanielLeens (commented):



The source review is clear, but the current Build is failing, so this PR is still waiting on CI. The current failed lanes include `Run / Dead links`, `Run / updated-modules-integration-test-part-2 (11, ubuntu-latest)`, `Run / updated-modules-integration-test-part-5 (11, ubuntu-latest)`, `Run / unit-test (8, ubuntu-latest)`, and 10 more failed lanes. The shared dead-link failure has a fix in #11502, but that fix has not merged into `dev` yet; syncing the current `dev` alone will not clear that lane today. After #11502 lands, please sync/retrigger and then diagnose only any lanes that still fail.

DanielLeens (commented):



The source review is clear, but the current Build is failing, so this PR is still waiting on CI. The current failed lanes include `Run / Dead links`, `Run / updated-modules-integration-test-part-2 (11, ubuntu-latest)`, `Run / updated-modules-integration-test-part-5 (11, ubuntu-latest)`, `Run / unit-test (8, ubuntu-latest)`, and 10 more failed lanes. The shared dead-link failure has a fix in #11502, but that fix has not merged into `dev` yet; syncing the current `dev` alone will not clear that lane today. After #11502 lands, please sync/retrigger and then diagnose only any lanes that still fail.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `9f066d6faf57` from scratch.

# What this PR fixes
- User pain: MySQL CDC users doing one-time backfill/bootstrap had no bounded mode — jobs either continued into binlog streaming after the snapshot or had to be killed manually, and there was no documented way to get a clean bounded exit.
- Fix approach: Introduces `startup.mode = "snapshot"` that reuses the existing snapshot split planning/reading path, reports the source as bounded, skips creation of the incremental split so the job finishes after all snapshot splits complete, keeps `exactly_once` bounded backfill working, and fails fast on incompatible `stop.mode`/`startup.specific-offset.*`/`startup.timestamp` combos; docs (en/zh) updated with the new enum value, a dedicated bootstrap section, and a revised FAQ entry.
- One-line summary: Feature is functionally sound and the docs are thorough, but there are three medium issues to address before merge — the restore-with-changed-startup-mode edge case is not validated despite the fail-fast promise, the rewritten `exactly_once` docs silently narrow the supported-mode contract for existing users, and the snapshot-only example shows a single `server-id` while the mode is pitched for parallel bounded jobs that need a server-id range per reader.

# Runtime chain I rechecked
```text
MySqlIncrementalSourceFactory.createSource() -> MySqlIncrementalSource  MySqlIncrementalSourceFactory.java:~95
  IncrementalSource.getBoundedness() -> returns BOUNDED when startupConfig mode == SNAPSHOT  IncrementalSource.java:~130
  IncrementalSource.createEnumerator() -> SnapshotSplitAssigner (snapshot-only path, no HybridSplitAssigner incremental phase)  IncrementalSource.java:~170
    SnapshotSplitAssigner.getNext() -> exhausts snapshot splits -> noMoreSplits()  SnapshotSplitAssigner.java:~150
    IncrementalSourceEnumerator.assignSplits() -> context.signalNoMoreSplits(subtask)  IncrementalSourceEnumerator.java:~180
      IncrementalSourceReader.handleNoMoreSplitsMessage() -> reader completes without IncrementalSplit / binlog fetch  IncrementalSourceReader.java:~140
Restore path (checkpointed table set authoritative, discovery skipped):
  IncrementalSource.restoreEnumerator(checkpointState) -> SnapshotSplitAssigner(checkpoint) — no catalog re-discovery  IncrementalSource.java:~200
Validation (fail-fast before source init):
  MySqlIncrementalSourceFactory.createSource() -> option compatibility check: SNAPSHOT + (stop.mode != never | startup.specific-offset.* | startup.timestamp) -> throws  MySqlIncrementalSourceFactory.java:~110
```

# Findings
**Issue 1: Docs promise fail-fast for incompatible options with `snapshot` mode, but the change-of-startup-mode-across-restore edge case is not covered**
- Location: `docs/en/connectors/source/MySQL-CDC.md:437`
- Why it matters: - Root cause: the new section states incompatible combinations "fail fast at startup", which only addresses fresh submissions. A job originally checkpointed under `startup.mode = initial` (hybrid state containing a pending/created incremental split) that is restored with the config flipped to `snapshot` presents state that snapshot-only mode is documented never to produce.
- Impact: depending on how `HybridSplitAssigner` interprets a restored `PendingSplitsState` containing an incremental split while `StartupMode.SNAPSHOT` is configured, the job could either wrongly stream binlog forever (violating the bounded contract) or NPE/hang on the unexpected split. Since the PR modifies both `IncrementalSource` restore handling and `HybridSplitAssigner` (+88 lines), this edge case is squarely in scope but neither the docs nor the described tests mention it.
- Fix: on restore, validate the checkpointed state against the configured startup mode and fail with a clear error message (e.g. "cannot restore a hybrid/incremental checkpoint with startup.mode=snapshot") rather than proceeding, and document that switching `startup.mode` across a checkpoint restore is not supported. Add a unit test in `HybridSplitAssignerTest` for a restored state that already contains an incremental split under snapshot mode.
- Suggested fix: "Fails fast at startup" covers fresh submissions, but there is a nastier edge: restoring a checkpoint taken under `startup.mode = initial` (whose state may already contain the incremental split) with the config changed to `snapshot`. The assigner then sees state that snapshot mode should never produce — it must either reject that restore with a clear error or deterministically finish, not stream binlog or hang. Please add a restore-state validation in the assigner/source and a `HybridSplitAssignerTest` case for a restored state containing an incremental split under `SNAPSHOT` mode, and document that changing `startup.mode` across restore is unsupported.
- Severity: Medium

**Issue 2: The rewritten `exactly_once` description now claims it is only supported for `initial` or `snapshot`, which silently narrows the documented contract for existing users on other startup modes**
- Location: `docs/en/connectors/source/MySQL-CDC.md:215`
- Why it matters: - Root cause: the English option table row for `exactly_once` was rewritten from the generic "Enable exactly once semantic." to "Supported when `startup.mode` is `initial` or `snapshot`", but this PR only adds a new startup mode — it does not (per the PR description) change how `exactly_once` interacts with `earliest`, `latest`, `specific`, or `timestamp` modes.
- Impact: users currently running `exactly_once = true` with non-initial startup modes will read this as a deprecation/behavior change and may reconfigure jobs unnecessarily; conversely, if the runtime does not actually reject the combination, the doc and code contract diverge, which is exactly the kind of drift that causes support tickets and silent semantic misunderstanding around checkpoint/state guarantees.
- Fix: either scope the sentence to what actually changed (e.g., "When `startup.mode` is `initial` or `snapshot`, `exactly_once` additionally enables bounded low-to-high-watermark binlog backfill during the snapshot phase") while keeping the original general statement, or — if the code truly restricts `exactly_once` to these two modes now — add explicit fail-fast validation in `MySqlIncrementalSourceFactory` and call the restriction out as a breaking change in the PR description.
- Suggested fix: This rewrite reads as if `exactly_once` is now *only* supported for `initial`/`snapshot`, but the PR doesn't change its behavior for `earliest`/`latest`/`specific`/`timestamp`. Can we rephrase to keep the general statement and add the snapshot-specific note, e.g. "Enable exactly once semantics. When `startup.mode` is `initial` or `snapshot`, this also enables bounded binlog backfill during the snapshot phase."? Otherwise existing users on other modes will think their configuration became invalid.
- Severity: Medium

**Issue 3: Snapshot-only example uses a single `server-id` although the mode is advertised for parallel bounded bootstrap with exactly_once binlog backfill, which requires a server-id range per reader**
- Location: `docs/en/connectors/source/MySQL-CDC.md:441`
- Why it matters: - Root cause: the new "Snapshot-Only Bootstrap" example hardcodes `server-id = 5656` while the same section states that `exactly_once` may perform bounded low-to-high-watermark binlog backfill during the snapshot phase. Binlog backfill opens a MySQL replication connection per reader, and with parallelism > 1 each reader must use a distinct server-id, otherwise MySQL kills the duplicate replication session with `ERROR 1236 / A slave with the same server_uuid/server_id`.
- Impact: users who copy this example, raise parallelism (the natural thing to do for a one-time backfill), and enable `exactly_once` will hit intermittent connection resets during backfill on Zeta/Flink. Notably, the PR's own E2E deliberately uses a range `server-id = "5657-5658"` with parallelism 2, so the documented example is weaker than the tested configuration.
- Fix: change the example to a range (e.g. `server-id = "5656-5657"`) or add an explicit note in the snapshot section that a server-id range covering the job parallelism is required when `exactly_once = true`, mirroring what the E2E config does.
- Evidence: `docs/en/connectors/source/MySQL-CDC.md:437  Because the job never enters continuous binlog streaming, `snapshot` mode is incompatible with `stop.mode` values other than `never` ... `exactly_once` remains available for bounded backfill during the snapshot phase and does not start continuous binlog streaming.`
- Suggested fix: Since this section explicitly says `exactly_once` can trigger bounded binlog backfill during the snapshot phase, this example should use a server-id range instead of a single `server-id = 5656`. With parallelism > 1 each backfill reader opens its own replication connection and duplicate server-ids get killed by MySQL. Your own E2E uses `server-id = "5657-5658"` with parallelism 2 — let's document the same pattern here (e.g. `server-id = "5656-5657"`) or add a note that the range must cover the job parallelism.
- Severity: Medium

# Review conclusion

- Issue 1: Docs promise fail-fast for incompatible options with `snapshot` mode, but the change-of-startup-mode-across-restore edge case is not covered
- Issue 2: The rewritten `exactly_once` description now claims it is only supported for `initial` or `snapshot`, which silently narrows the documented contract for existing users on other startup modes
- Issue 3: Snapshot-only example uses a single `server-id` although the mode is advertised for parallel bounded bootstrap with exactly_once binlog backfill, which requires a server-id range per reader

fatmanverse:
Thanks @SEZ9 for the thorough follow-up review. I've addressed all three non-blocking issues in commit `628fadb4c`.

**Issue 1 — cross-restore `startup.mode` switch (fail-fast)**

I traced the runtime path and found the bounded-contract violation has two distinct entry points, so I added fail-fast at both, plus a defense-in-depth check:

- `IncrementalSource.restoreEnumerator()` now rejects restoring an `IncrementalPhaseState` when `startup.mode = snapshot`. This covers a job originally started with `earliest`/`latest`/`specific`/`timestamp` (which checkpoints an `IncrementalPhaseState`) being restored as snapshot.
- `IncrementalSourceReader.addSplits()` now rejects an incremental (binlog) split under snapshot mode. This is the actual "streams forever" path: a job that ran `initial`, completed the snapshot, entered the binlog phase and checkpointed, then restarts with `startup.mode = snapshot` — the enumerator state is clean/bounded, but the **reader** restores an `IncrementalSplit` and would stream binlog indefinitely. Since `IncrementalPhaseState` is an empty marker and the incremental split state actually lives in the reader, the reader-side guard is essential rather than optional.
- `HybridSplitAssigner.addSplits()` also rejects an incremental split under snapshot mode as defense-in-depth.

New tests: `IncrementalSourceTest#testSnapshotOnlyRestoreRejectsIncrementalCheckpoint` (enumerator) and `HybridSplitAssignerTest#testSnapshotOnlyRejectsRestoredIncrementalSplit` (assigner). Docs now state that changing `startup.mode` across a restore is not supported.

**Issue 2 — `exactly_once` doc contract**

You're right that the wording wrongly narrowed the contract. I confirmed via `ConfigValidator` that the `.conditional(STARTUP_MODE, [INITIAL, SNAPSHOT], EXACTLY_ONCE)` rule never rejects `exactly_once` at runtime (it has a default of `false`, so it is never treated as absent). The description now keeps the general statement and adds the snapshot-phase note, in both en and zh:

> Enable exactly once semantics. When `startup.mode` is `initial` or `snapshot`, it additionally enables bounded low-to-high-watermark binlog backfill during the snapshot phase.

**Issue 3 — snapshot-only example `server-id`**

Changed the example to a range (`"5656-5657"`) matching the E2E, and added a note that `exactly_once` backfill opens a replication connection per reader, so a `server-id` range covering the job parallelism is required. Fixed in both en and zh.

**Validation**
- `./mvnw spotless:apply` — clean
- `connector-cdc-base` compiles clean (main + test)
- Targeted tests pass: `IncrementalSourceTest` (3), `HybridSplitAssignerTest` (4)

DanielLeens (commented):

The previous source concern is resolved on `628fadb4c8dc`: snapshot-only restore now treats checkpointed tables as authoritative and rejects an incompatible incremental-state restore.

Queue note: this head is `diverged` against `dev` (ahead_by=4, behind_by=4). I am recording that fact separately; it is not by itself a request to sync unless the CI guidance above explicitly says so.

fatmanverse:
Thanks for the recheck. I have rebased this PR onto the latest dev and resolved the documentation conflicts while preserving the snapshot, specific/GTID, and enable_concurrent_read documentation. The branch is now up to date with dev.

I also fixed the snapshot-only E2E failure. The test previously updated the primary key from id=5 to id=6 while later snapshot splits were still pending. Flink 1.18 could legitimately read id=6 from a later snapshot split, so this did not indicate that the source had entered continuous binlog streaming. The test now only deletes id=2 from an already completed split, which keeps the intended assertion without moving a row into a pending snapshot range.

The latest dev also introduced Container.ExecResult usages in MysqlCDCSpecificStartingOffsetIT without importing org.testcontainers.containers.Container, which blocked compilation of the whole MySQL CDC E2E module. I added the missing import.

Local verification completed:
- ./mvnw spotless:apply
- 18 related CDC unit tests passed
- MysqlCDCSnapshotOnlyIT passed on all 5 PR runtimes: Flink 1.13, 1.15, 1.18, 1.20, and Zeta (5 tests, 0 failures, 0 errors)
- ./mvnw -q -DskipTests -Dskip.ui=true verify passed

The original full verify without skip.ui was also attempted locally, but the unchanged seatunnel-engine-ui npm install exited with code 243, so the successful full backend verification explicitly skipped UI. The updated head is 012fbd1281a278580e3be7a53b43ea0afd28b751.

fatmanverse:
The latest CI failure in `ImportClassCheckTest` has been addressed in commit `70cdf0d`.

The fix is a one-line import replacement in `CoordinatorServicePipelineCleanupTest`, using SeaTunnel's `org.apache.seatunnel.engine.common.utils.concurrent.CompletableFuture` instead of `java.util.concurrent.CompletableFuture`, as required by the repository import rule.

Local verification completed:
- `ImportClassCheckTest`: 8 tests passed
- `CoordinatorServicePipelineCleanupTest`: 8 tests passed
- `./mvnw -q -DskipTests -Dskip.ui=true verify`: passed

This CI fix is independent of the MySQL CDC snapshot-only behavior.

DanielLeens (commented):
Outdated.

SEZ9:
Thanks @fatmanverse for the thorough follow-ups — I've gone through `e22ab12d7` and `aa2705213` against my earlier review notes, and all three of my medium concerns are now addressed:

1. **Restore edge case** — Making the checkpointed table set authoritative on the snapshot-only restore path (and marking legacy state as having checkpointed remaining tables before the assigner opens) is the right call. The unit test asserting `discoverDataCollections()` is never invoked on recovery is exactly the regression coverage I wanted.
2. **`exactly_once` contract** — Good, keeping both `initial` and `snapshot` supported and documenting the bounded low-to-high-watermark backfill semantics resolves the doc-narrowing concern.
3. **`server-id` range** — The `5657-5658` range with parallelism 2 in the E2E fixes the example inconsistency, and the concurrent-write test design (updates/deletes against completed snapshot ranges while later splits are outstanding, sink stays equal to the original snapshot) is a solid way to prove the bounded semantics.

Also appreciated:
- Reusing `MysqlCDCDriverResolver` instead of downloading the driver inside the container — that keeps the E2E hermetic.
- Fail-fast validation now running before datasource/dialect/offset init, with independent negative coverage for `stop.mode`, `startup.timestamp`, and both `startup.specific-offset.*` keys.
- Retaining both existing public `HybridSplitAssigner` constructors preserves binary compatibility for other dialects.

On the environment notes:
- The `-Dtestcontainer.version=1.21.4` override for Docker 29/OrbStack is fine as a local workaround since no POM change is included — just confirming CI runs with the repo default, so we'll rely on the CI result there.
- The `connector-jdbc` compilation failures and the `seatunnel-config-shade` issue on `origin/dev` are indeed pre-existing and unrelated to this PR; no action needed from you on those here.

Remaining asks before merge:
1. Please rebase onto the latest `dev` once more and make sure the full CI pipeline (including the CDC E2E jobs) goes green — that's the last gate.
2. Double-check the PR title/description reflect the final scope so the changelog entry is accurate (it sounds like you've already updated it — just flagging for a final look).

Once CI is green I'm happy to approve. Nice work on this one — the feature is well-scoped and the test coverage came out strong.

DanielLeens (commented):
Outdated.

DanielLeens (commented):
Outdated.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head from the MySQL CDC option surface through the snapshot assigner, reader, restore path, docs, and the new UT/E2E coverage.

I agree with @SEZ9's latest merge direction here, and I think the earlier checkpoint/restore concern from @nzw921rx is addressed in the current implementation.

The source-level path looks good to me:

- `MySqlIncrementalSourceFactory` now accepts the canonical `StartupMode.SNAPSHOT_ONLY` value and allows `exactly_once` for `INITIAL` and `SNAPSHOT_ONLY`.
- `IncrementalSource.getBoundedness()` returns `BOUNDED` for snapshot-only jobs.
- `createEnumerator()` creates a `SnapshotOnlySplitAssigner` for snapshot-only mode, so the job reads snapshot splits and does not transition into continuous binlog streaming.
- `restoreEnumerator()` validates that startup mode and checkpoint-state type still match. A snapshot-only job can restore only from `SnapshotPhaseState`; changing startup mode across restore fails fast instead of silently resuming the wrong phase.
- `IncrementalSourceReader.addSplits()` rejects incremental/binlog splits when the source is running in snapshot-only mode, which protects the bounded contract even if a bad restored state or assignment path leaks through.

The runtime chain I checked is:

```text
Config parsing
  -> MySqlIncrementalSourceFactory.optionRule()
      -> startup.mode accepts SNAPSHOT_ONLY
      -> exactly_once is valid for INITIAL / SNAPSHOT_ONLY

New job startup
  -> IncrementalSource(...)
      -> getStartupConfig()
      -> validateSnapshotOnlyConfig()
          -> reject stop.mode != never
          -> reject startup.specific-offset.* / startup.timestamp
  -> createEnumerator()
      -> SnapshotOnlySplitAssigner
      -> SnapshotSplitAssigner assigns only snapshot splits

Reader runtime
  -> IncrementalSourceReader.addSplits()
      -> snapshot splits are accepted
      -> incremental splits fail fast under snapshot-only
  -> pollNext()
      -> no more splits + no more element -> signalNoMoreElement

Restore
  -> restoreEnumerator(checkpointState)
      -> validateRestoredState()
      -> SNAPSHOT_ONLY + SnapshotPhaseState is allowed
      -> SNAPSHOT_ONLY + Hybrid/Incremental state is rejected
```

Test coverage is also in good shape. The UTs cover boundedness, invalid option combinations, invalid restore-state combinations, assigner/reader rejection of incremental splits, and factory option rules. The E2E covers that a snapshot-only job finishes and does not consume a binlog change produced during the snapshot job.

Test stability rating: has a minor risk, not a blocker. `MysqlCDCSnapshotOnlyIT` waits for a mid-snapshot window before deleting a row. The test mitigates the race with `snapshot.split.size = 1` and 100+ snapshot splits, but if CI ever becomes fast enough to finish before that window is observed, the test could fail even though the source behavior is correct. If that shows up later, I would strengthen it with a more deterministic progress/test hook or lock-based gate. I would not block this PR on that now.

### Merge conclusion

### Conclusion: can merge after the remaining gate is cleared

   - I do not see a new code-side blocker in the latest source.
   - The remaining blocker is CI / merge gate. The current `Build` check is failing: https://github.com/apache/seatunnel/runs/90152930675. I was able to inspect `updated-modules-integration-test-part-8`, and that failure is from Iceberg E2E (`IcebergSinkCDCIT` / `IcebergSinkParallelCommitIT`), not from this MySQL snapshot-only diff. The `updated-modules-integration-test-part-6` log stream was cancelled by GitHub, so that lane still needs a rerun or a readable log before merge.

2. Non-blocking suggestions
   - Keep an eye on the new E2E's mid-snapshot timing window. It looks acceptable now, but it would be worth hardening if it ever becomes flaky.

I also checked this head against the latest `dev`: the branch is currently `diverged` with `behind_by=4`. Since the visible failing CI does not appear directly tied to the MySQL snapshot-only code, the lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the MySQL snapshot-only E2E fails after that, please share the new failing log link and I can help narrow it down on the updated head.

Overall, this is a careful implementation of the bounded CDC bootstrap mode. Nice work tightening the restore and reader-side safety checks.

fatmanverse:
@nzw921rx The earlier snapshot-only restore and state concerns are addressed in the current implementation, and the branch has now been rebased onto dev at head 31c1e0ed. Focused CDC and MySQL tests, Spotless, and the full skip-tests verify pass locally; the new CI run is queued. Please re-review the current head or dismiss the stale changes-requested review when convenient.

DanielLeens (commented):
Thanks for the update. I re-reviewed the full current head `31c1e0ed` from the MySQL CDC option surface through boundedness, enumerator creation/restore, reader-side split handling, docs, and the new UT/E2E coverage.

One thing to update from my previous round: the earlier "sync the latest dev first" guidance is now outdated. The current head is no longer behind `dev` (`ahead_by=7`, `behind_by=0`), and the current required `Build` is queued rather than failing. So I do not see an upstream-sync blocker on the latest head.

I also agree with @SEZ9's current merge direction here, and I think the earlier checkpoint/restore and hermetic E2E concerns from @nzw921rx are addressed in the current implementation.

# PR solves what problem
- User pain: MySQL CDC users who only wanted a one-time bounded bootstrap had no dedicated mode. Existing startup modes either transitioned into continuous binlog streaming or required users to stop the job manually, and the restore/runtime contract for a bounded snapshot-only job was unclear.
- Fix approach: this PR adds `startup.mode = "snapshot-only"`, routes that mode through a dedicated snapshot-only assigner, makes the source bounded, rejects incompatible binlog-oriented options and incompatible restore-state combinations, and adds UT/E2E coverage plus en/zh docs for the new contract.
- One-line summary: this adds a true bounded MySQL CDC bootstrap mode that finishes after the snapshot phase instead of drifting into continuous streaming.

# I. Code Change Review

## 1.1 Core logic analysis

### Precise change summary

The key runtime changes are in:

- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/IncrementalSource.java:123-174,297-303,346-488`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/enumerator/SnapshotOnlySplitAssigner.java:33-124`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-base/src/main/java/org/apache/seatunnel/connectors/cdc/base/source/reader/IncrementalSourceReader.java:130-176`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/config/MySqlIncrementalSourceOptions.java:61-79`
- `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/MySqlIncrementalSourceFactory.java:57-106`

The critical path is:

```text
Config parsing
  -> MySqlIncrementalSourceOptions.STARTUP_MODE accepts SNAPSHOT_ONLY
  -> MySqlIncrementalSourceFactory.optionRule() keeps exactly_once available for INITIAL / SNAPSHOT_ONLY

Source construction
  -> IncrementalSource(ReadonlyConfig, CatalogTables)
      -> validateSnapshotOnlyConfig()
      -> getBoundedness() returns BOUNDED for SNAPSHOT_ONLY

Fresh startup
  -> createEnumerator()
      -> SnapshotOnlySplitAssigner for SNAPSHOT_ONLY
      -> only snapshot splits are generated / assigned

Restore
  -> restoreEnumerator(checkpointState)
      -> validateRestoredState(startupMode, checkpointState)
      -> SNAPSHOT_ONLY may restore only from SnapshotPhaseState
      -> checkpointed table set is used as the authoritative captured set

Reader runtime
  -> IncrementalSourceReader.addSplits()
      -> snapshot splits proceed
      -> incremental/binlog split under SNAPSHOT_ONLY fails fast
```

### Before / after snippets

Before, the source only had the existing initial / incremental path and no dedicated bounded snapshot-only restore guard:

```java
if (sourceConfig.getStartupConfig().getStartupMode() == StartupMode.INITIAL) {
    splitAssigner = new HybridSplitAssigner<>(...);
}
```

After, the latest head has a dedicated snapshot-only runtime path plus restore validation:

```java
if (sourceConfig.getStartupConfig().getStartupMode() == StartupMode.SNAPSHOT_ONLY) {
    splitAssigner = new SnapshotOnlySplitAssigner<>(...);
}

validateRestoredState(startupMode, checkpointState);
```

and the reader now explicitly rejects a leaked incremental split under snapshot-only:

```java
if (sourceConfig.getStartupConfig().getStartupMode() == StartupMode.SNAPSHOT_ONLY) {
    throw new IllegalStateException(...);
}
```

### Key findings

1. The normal snapshot-only path does hit the new logic. `IncrementalSource.createEnumerator()` now selects `SnapshotOnlySplitAssigner` for `StartupMode.SNAPSHOT_ONLY`, and `getBoundedness()` returns `BOUNDED`, so this is not a dead branch.
2. The current restore implementation fixes the earlier bounded-input concern. `restoreEnumerator()` now rejects state/mode mismatches up front, and the snapshot-only restore path rebuilds captured tables from checkpoint state instead of rediscovering the catalog.
3. The reader-side fail-fast check is the right defense-in-depth layer. Even if an incompatible restore somehow leaked an incremental split to a subtask, `IncrementalSourceReader.addSplits()` now stops immediately instead of streaming forever and violating the bounded contract.
4. The latest docs commit `31c1e0ed` closes the remaining documentation gap by explicitly stating that changing `startup.mode` across checkpoint restore is unsupported in both `docs/en` and `docs/zh`.
5. The current E2E is materially stronger than the first version: it uses the Maven-resolved JDBC driver helper, runs with `snapshot.split.size = 1`, waits for an actual mid-snapshot window with Awaitility, and verifies that a runtime delete on an already-emitted split does not turn into binlog consumption.

### Logic correctness deep dive

- `IncrementalSource.validateSnapshotOnlyConfig()` now rejects `stop.mode != never`, `startup.specific-offset.*`, and `startup.timestamp` before datasource / dialect initialization (`IncrementalSource.java:138-173`). That is the correct fail-fast point, because those options are semantically about binlog entry / exit and snapshot-only never enters continuous streaming.
- `IncrementalSource.restoreEnumerator()` gates the restore by checkpoint-state type (`IncrementalSource.java:405-488`). For snapshot-only mode, only `SnapshotPhaseState` is accepted; `HybridPendingSplitsState` and `IncrementalPhaseState` are rejected. That matches the intended lifecycle contract: changing startup mode across restore is unsupported.
- `getCheckpointCapturedTables()` and `SnapshotOnlySplitAssigner.markRemainingTablesCheckpointed()` preserve the checkpointed table set and avoid rediscovery on the snapshot-only restore path (`IncrementalSource.java:490-500`, `SnapshotOnlySplitAssigner.java:66-80`). This is the core fix for the earlier "new tables appear after failover" concern.
- `IncrementalSourceReader.addSplits()` now guards the actual subtask execution path (`IncrementalSourceReader.java:137-163`). This matters because an incremental split can live in restored reader state even if the enumerator state itself looks bounded; failing here is what prevents an accidental infinite binlog read.

### Full runtime chain

```text
MySQL config
  -> MySqlIncrementalSourceOptions.STARTUP_MODE includes SNAPSHOT_ONLY
  -> MySqlIncrementalSourceFactory.optionRule() exposes the new mode

Job startup
  -> new MySqlIncrementalSource(...)
      -> IncrementalSource.validateSnapshotOnlyConfig()  [IncrementalSource.java:138-173]
      -> IncrementalSource.getBoundedness() -> BOUNDED   [IncrementalSource.java:297-303]

Enumerator creation
  -> IncrementalSource.createEnumerator()               [IncrementalSource.java:346-402]
      -> SnapshotOnlySplitAssigner(...)                 [IncrementalSource.java:373-380]
          -> SnapshotSplitAssigner.getNext()
              -> only snapshot splits are generated / assigned

Reader execution
  -> IncrementalSourceReader.addSplits()                [IncrementalSourceReader.java:130-176]
      -> snapshot splits allowed
      -> incremental split under SNAPSHOT_ONLY throws immediately
  -> pollNext()
      -> no more splits + no more elements
      -> signalNoMoreElement()

Checkpoint restore
  -> IncrementalSource.restoreEnumerator()              [IncrementalSource.java:405-488]
      -> validateRestoredState(startupMode, checkpointState)
      -> SNAPSHOT_ONLY + SnapshotPhaseState only
      -> getCheckpointCapturedTables()
      -> SnapshotOnlySplitAssigner(..., checkpointState, ...)
          -> markRemainingTablesCheckpointed()
          -> SnapshotSplitAssigner.open() skips rediscovery because remaining tables are checkpointed
```

## 1.2 Compatibility impact

Compatibility assessment: fully compatible.

- API: no existing public option key is removed or renamed.
- Configuration: additive. `startup.mode = "snapshot-only"` is a new allowed value.
- Default values: unchanged.
- Protocol / serialization: existing startup modes keep their prior behavior; the new mode adds a new bounded runtime contract.
- Historical behavior: unchanged for existing startup modes. The one important rule is now explicit and enforced: switching `startup.mode` across checkpoint restore is unsupported and should be handled by submitting a new job instead of restoring an old checkpoint under a different startup mode.

## 1.3 Performance / side-effect analysis

- CPU / memory: no obvious hot-path regression. The added checks are lightweight branching on startup mode / checkpoint-state type.
- GC: no meaningful new allocation pressure on the steady-state read path beyond normal checkpoint-state reconstruction.
- Network: the E2E is now more hermetic because it uses `MysqlCDCDriverResolver` instead of downloading a JDBC driver from inside the container at runtime.
- Concurrency / retries / idempotency: the important side effect here is positive. Failing fast on an incompatible incremental split avoids an accidentally unbounded reader that would otherwise keep consuming binlog indefinitely after restore.
- Resource release: no new leak surfaced in the current path review.

I do not see a new blocking or non-blocking source issue on the latest head.

- Invalid snapshot-only option combinations now fail with explicit `IllegalArgumentException`s before source initialization.
- Incompatible restore-state combinations now fail with explicit `IllegalStateException`s instead of silently resuming the wrong phase.
- The reader-side guard prevents a misleading "successful restore" that would actually violate the bounded contract.

# II. Code Quality Assessment

## 2.1 Code conventions

The current head is consistent with the surrounding CDC code style. The newly added core methods and guards are documented with comments where the behavior is non-trivial, especially around restore invariants and the reader-side boundedness guard.

## 2.2 Test coverage and test stability

Coverage is good for the current scope:

- Unit / focused regression coverage:
  - `IncrementalSourceTest` verifies checkpoint-authoritative restore and mismatched restore rejection.
  - `SnapshotOnlySplitAssignerTest` verifies no rediscovery on legacy restore and rejects incremental add-back.
  - `IncrementalSourceReaderTest` verifies reader-side rejection of incremental split restore.
  - `DebeziumJsonFormatTest` verifies snapshot-only option validation happens before source initialization.
  - `MySqlIncrementalSourceFactoryTest` verifies the option-rule surface includes `SNAPSHOT_ONLY`.
- E2E coverage:
  - `MysqlCDCSnapshotOnlyIT` verifies the job terminates as a bounded job and does not consume a runtime binlog delete while later snapshot splits are still in progress.

Test stability rating: stable.

Why:

- The E2E uses Awaitility rather than fixed sleeps.
- The runtime mutation only touches a completed split range, so it no longer races by moving a row into a still-pending snapshot split.
- `snapshot.split.size = 1` plus 100+ additional rows creates a meaningful multi-split window instead of collapsing into a trivial one-split snapshot.

## 2.3 Documentation update

Documentation update is included and aligned:

- `docs/en/connectors/source/MySQL-CDC.md`
- `docs/zh/connectors/source/MySQL-CDC.md`

The English and Chinese docs are consistent on:

- the new `snapshot-only` startup mode,
- incompatible `stop.mode` / startup offset options,
- the `server-id` range requirement for bounded backfill with `exactly_once`,
- the restore constraint that `startup.mode` must not be changed across checkpoint restore.

# III. Architecture Reasonableness

## 3.1 Elegance of the solution

This is a precise fix / feature addition rather than a temporary workaround. Using a dedicated `SnapshotOnlySplitAssigner` and explicit restore validation is a cleaner architecture than trying to smuggle a bounded mode through the normal hybrid incremental lifecycle.

## 3.2 Maintainability

Maintainability is good in the current shape:

- the bounded-mode contract is explicit in config validation,
- the restore invariant is centralized,
- the reader has an extra safety net,
- the docs now match the implementation.

That makes future debugging much easier than relying on implicit phase assumptions.

## 3.3 Extensibility

This approach is reusable for other CDC dialects that may want a bounded bootstrap mode. The main idea is sound: separate assigner path, explicit restore invariant, and reader-side fail-fast guard.

## 3.4 Historical version compatibility

Compatible with existing jobs and existing startup modes. The only operational requirement is that users should not attempt to restore an old checkpoint under a different startup mode; the current docs now state that clearly.

# IV. Problem summary

| No. | Problem | Location | Severity |
|-----|---------|----------|----------|
| - | No blocking or non-blocking source issue identified on the current head | - | - |

# V. Merge conclusion

### Conclusion: can merge

   - No new source-level blocker from Daniel on the latest head `31c1e0ed`.
   - The remaining gate is repository CI / review state, not the source diff itself. The current required `Build` is still queued: https://github.com/apache/seatunnel/runs/91630528994
   - GitHub still shows `reviewDecision = REVIEW_REQUIRED`, so a write-capable maintainer / reviewer may still need to complete the final repository approval step after the checks finish.

2. Suggested fixes (non-blocking)
   - None from Daniel on the current head.

Overall evaluation:

- I rechecked the current implementation from config validation through restore and runtime execution, and I do not see a reopened source-side issue.
- Compared with my previous round, the CI / merge-gate fact has changed materially: this branch is no longer behind `dev`, so the old upstream-sync suggestion no longer applies to the latest head.
- The current implementation is the more maintainable direction than trying to bolt bounded behavior onto the normal hybrid streaming path.

Since my repository permission is read-only here, this review records the current technical conclusion only; a write-capable maintainer may still need to issue the final approval or dismissal action after CI completes.

SEZ9 (commented):
Thanks for working on this. I reviewed the latest head `31c1e0ed1446`.

# What this PR fixes
- User pain: MySQL CDC users who only need a one-time backfill/bootstrap had no bounded mode — every job either drifted into continuous binlog streaming or required a binlog-oriented start offset, forcing manual job cancellation and wasted replication slots.
- Fix approach: Adds a first-class `startup.mode = "snapshot-only"` that reuses the existing snapshot split planning/reading path, reports `Boundedness.BOUNDED`, completes on snapshot-split exhaustion without creating an incremental split, keeps the checkpointed table set authoritative on restore, and fails fast on incompatible `stop.mode`/`startup.specific-offset.*`/`startup.timestamp` combinations; docs (EN/ZH) gain the new enum value, an example section, and an updated FAQ.
- One-line summary: The bounded completion path and fail-fast validation are sound on the current head; remaining items are doc-level gaps (restore semantics, `exactly_once = false` consistency under concurrent writes, required replication privileges, the misleading `server-id` "rejected by MySQL" claim, example/config mismatch, hardcoded credentials) plus a runtime robustness ask — fail-fast on server-id/parallelism mismatch under `snapshot-only + exactly_once`.

# Runtime chain I rechecked
```text
MySqlIncrementalSourceFactory.createSource()  MySqlIncrementalSourceFactory.java:96
  -> MySqlSourceOptions / StartupConfig validation (rejects stop.mode != never, specific-offset.*, startup.timestamp with snapshot-only)  MySqlSourceOptions.java:61
  -> MySqlIncrementalSource.getBoundedness() returns BOUNDED when startupMode == SNAPSHOT_ONLY  MySqlIncrementalSource.java:88
IncrementalSource.createEnumerator()  IncrementalSource.java:214
  -> SnapshotSplitAssigner (snapshot-only skips HybridSplitAssigner, so no IncrementalSplit is ever constructed)  SnapshotSplitAssigner.java:141
  -> IncrementalSourceEnumerator.assignSplits() -> handleSplitRequest() returns NoMoreSplits once snapshot splits are exhausted  IncrementalSourceEnumerator.java:173
IncrementalSource.restoreEnumerator()  IncrementalSource.java:246
  -> checkpointed table set treated as authoritative, catalog discovery skipped; incompatible incremental-state restore rejected  IncrementalSource.java:258
IncrementalSourceReader.handleNoMoreSplits() -> context.signalNoMoreElement() (job finishes without a checkpoint dependency)  IncrementalSourceReader.java:142
exactly_once backfill sub-path:
  IncrementalSourceScanFetcher.submitTask() -> MySqlSnapshotFetchTask.execute()  MySqlSnapshotFetchTask.java:79
    -> MySqlBinlogFetchTask bounded low->high watermark read via per-reader BinaryLogClient(server-id)  MySqlBinlogFetchTask.java:64  (this is where the server-id-range requirement and connection-close-on-finish/failure concerns land)
```

# Findings
**Issue 1: server-id/parallelism mismatch under snapshot-only + exactly_once is only detected by MySQL at runtime instead of failing fast with a clear configuration error**
- Location: `docs/en/connectors/source/MySQL-CDC.md:443`
- Why it matters: - The new documentation explicitly acknowledges a hard runtime constraint: with `exactly_once = true`, every snapshot reader opens its own replication connection, and MySQL rejects duplicate server-ids. This means a user who configures `startup.mode = "snapshot-only"`, `exactly_once = true`, and a single `server-id` with parallelism > 1 will hit a mid-job MySQL replication error (e.g. `A slave with the same server_uuid/server_id ...`) after splits have already been assigned, rather than a fail-fast configuration error at job submission.
- This PR already fails fast for the `stop.mode`/`startup.timestamp`/`startup.specific-offset.*` incompatibilities, so this constraint should get the same treatment.
- Fix: when `startup.mode = snapshot-only` and `exactly_once = true`, validate in `IncrementalSource#createEnumerator` (where `currentParallelism()` is known) that the `server-id` range covers the configured parallelism, and throw a clear configuration exception instead of letting readers crash mid-job.
- Evidence: `docs/en/connectors/source/MySQL-CDC.md:196  ... Incompatible stop modes, timestamp startup, and specific-offset startup options fail before source initialization. (row text: `snapshot-only`: ... is incompatible with `stop.mode` (other than `never`) and the binlog-oriented startup offset options `startup.specific-offset.*` and `startup.timestamp`.)`; `docs/en/connectors/source/MySQL-CDC.md:443  When `exactly_once = true`, each snapshot reader opens its own MySQL replication connection for the bounded backfill, so configure a `server-id` range that covers the job parallelism ...`; `docs/en/connectors/source/MySQL-CDC.md:441  ... `exactly_once` remains available for bounded backfill during the snapshot phase and does not start continuous binlog streaming.`
- Suggested fix: This paragraph documents a hard failure mode ("a single `server-id` shared by multiple readers is rejected by MySQL") but leaves the detection to MySQL at backfill time, after splits are already assigned. Since this PR already fails fast for `stop.mode`/`startup.timestamp`/`startup.specific-offset.*` incompatibilities, please add the same fail-fast treatment for this case: when `startup.mode = snapshot-only` and `exactly_once = true`, validate in `IncrementalSource#createEnumerator` (where `currentParallelism()` is known) that the `server-id` range covers the parallelism, and throw a clear configuration exception instead of letting readers crash mid-job with a MySQL duplicate-server-id replication error.
- Severity: Medium

**Issue 2: Consistency semantics of snapshot-only with exactly_once = false under concurrent writes are not documented**
- Location: `docs/en/connectors/source/MySQL-CDC.md:439`
- Why it matters: - The updated `exactly_once` row makes clear that the low-to-high-watermark binlog backfill only happens when `exactly_once = true`. By implication, with the default `exactly_once = false`, each snapshot split is read by a plain SELECT at a different point in time, so a snapshot-only job running against a table receiving concurrent updates/deletes produces a bounded result that is not a consistent point-in-time snapshot (a row moved across split boundaries can be read twice or missed).
- The new snapshot-only section positions the mode as a one-time backfill/bootstrap tool, which is exactly the scenario where users expect a consistent snapshot, yet it never states this caveat.
- Fix: document the consistency guarantee in the snapshot-only section and recommend `exactly_once = true` when the source tables receive concurrent writes during the snapshot, in both the EN and ZH docs.
- Evidence: `docs/en/connectors/source/MySQL-CDC.md:439  Use `startup.mode = "snapshot-only"` for a bounded bootstrap job: it reads the configured snapshot data and then finishes cleanly, without switching into binlog streaming.`; `docs/en/connectors/source/MySQL-CDC.md:219  | exactly_once | Boolean | No | false | Enable exactly once semantics. When `startup.mode` is `initial` or `snapshot-only`, it additionally enables bounded low-to-high-watermark binlog backfill during the snapshot phase.`; `docs/en/connectors/source/MySQL-CDC.md:439  Use `startup.mode = "snapshot-only"` for a bounded bootstrap job: it reads the configured snapshot data and then finishes cleanly, without switching into binlog streaming. This suits one-time backfill, initial warehouse/table bootstrap, and controlled migration stages where a continuous CDC task is not wanted.`
- Suggested fix: This section should spell out the consistency guarantee: with the default `exactly_once = false` there is no low-to-high-watermark backfill (per the updated `exactly_once` row above), so under concurrent writes each split is read at a different point in time and the bounded output is not a consistent point-in-time snapshot. Since "one-time backfill / bootstrap" is exactly where users expect a consistent snapshot, please add a note recommending `exactly_once = true` when the source tables receive concurrent writes during the snapshot, in both the en and zh docs.
- Severity: Medium

**Issue 3: Snapshot-only incompatibility constraints are documented only inside the `startup.mode` table cell and the new prose section; the affected option rows (`stop.mode`, `startup.specific-offset.*`, `startup.timestamp`) do not carry the restriction themselves**
- Location: `docs/en/connectors/source/MySQL-CDC.md:196`
- Why it matters: - Root cause: the new fail-fast constraints ("incompatible with `stop.mode` (other than `never`) and the binlog-oriented startup offset options `startup.specific-offset.*` and `startup.timestamp`") were added into the already very long `startup.mode` cell at docs/en/connectors/source/MySQL-CDC.md:196, but the rows for those specific options were left untouched.
- Impact: users typically scan the option table row-by-row when configuring a job. Someone reading the `startup.specific-offset.file`, `startup.specific-offset.pos`, `startup.specific-offset.gtid-set`, `startup.timestamp`, or `stop.mode` rows sees no hint that these options are rejected under `startup.mode = snapshot-only`, and only discovers the constraint through the fail-fast error at runtime.
- Fix: add a one-line note to each affected option row (EN and ZH) stating that the option is rejected when `startup.mode = snapshot-only`.
- Evidence: `docs/en/connectors/source/MySQL-CDC.md:441  Because the job never enters continuous binlog streaming, `snapshot-only` mode is incompatible with `stop.mode` values other than `never` and the binlog-oriented startup offset options `startup.specific-offset.*` and `startup.timestamp`; configuring them together fails fast at startup.`; `docs/en/connectors/source/MySQL-CDC.md:196  `snapshot-only`: Read snapshot data only and then finish. ... Changing `startup.mode` when restoring from a checkpoint is not supported. Submit a new job when switching between `snapshot-only` and other startup modes.`; `docs/en/connectors/source/MySQL-CDC.md:441  Because the job never enters continuous binlog streaming, `snapshot-only` mode is incompatible with `stop.mode` values other than `never` ...`
- Suggested fix: The snapshot-only restrictions are only stated here and in the new prose section, but the rows for `stop.mode`, `startup.specific-offset.file`/`pos`/`gtid-set` and `startup.timestamp` still don't mention that they are rejected when `startup.mode = snapshot-only`. Please add a one-line note to each of those option rows (EN and ZH) so users scanning the table row-by-row see the constraint before hitting the fail-fast error at runtime.
- Severity: Medium

**Issue 4: ZH documentation parity: mirror the EN `exactly_once` row and FAQ updates for snapshot-only**
- Location: `docs/zh/connectors/source/MySQL-CDC.md:195`
- Why it matters: - Root cause: the EN doc updates three places — the `startup.mode` cell (docs/en/connectors/source/MySQL-CDC.md:196), the `exactly_once` row (line 219, which now explains bounded low-to-high-watermark backfill for `initial`/`snapshot-only`), and the FAQ answer (line 521, which now distinguishes `initial` vs `snapshot-only` behavior). In the ZH diff only the `startup.mode` cell update (docs/zh/connectors/source/MySQL-CDC.md:195) is visible.
- Impact: if the ZH `exactly_once` row, the corresponding FAQ answer, and the Snapshot-Only example section are not updated, the two language versions diverge and Chinese-language users miss the exactly-once backfill semantics and the `server-id` range guidance.
- Fix: carry all of the EN updates (the `exactly_once` row, the FAQ answer, and the Snapshot-Only example section with the `server-id` range guidance) into the ZH doc.
- Evidence: `docs/en/connectors/source/MySQL-CDC.md:521  With `startup.mode = "initial"`, SeaTunnel takes a consistent full snapshot of the configured tables and then switches to reading binlog ... With `startup.mode = "snapshot-only"`, the source finishes after the snapshot phase and never enters continuous binlog reading.`; `docs/zh/connectors/source/MySQL-CDC.md:195  | startup.mode | Enum | 否 | INITIAL | ... `snapshot-only`: 仅读取快照数据，读取完成后任务有界结束 ...`; `docs/zh/connectors/source/MySQL-CDC.md:195  | startup.mode | Enum | 否 | INITIAL | MySQL CDC 消费者的可选启动模式, 有效枚举值为 `initial`, `earliest`, `latest` , `specific`, `timestamp` 和 `snapshot-only`. ... 不支持在从 checkpoint 恢复时更改 `startup.mode`。`
- Suggested fix: The EN doc also updates the `exactly_once` row (bounded low-to-high-watermark backfill for `initial`/`snapshot-only`), the FAQ answer about the snapshot phase, and adds a dedicated Snapshot-Only example section with the `server-id` range guidance. Please make sure this ZH doc carries all of the same updates so the two language versions stay consistent.
- Severity: Low

**Issue 5: New snapshot-only doc example embeds literal credentials, encouraging users to hardcode passwords in version-controlled job configs**
- Location: `docs/en/connectors/source/MySQL-CDC.md:450`
- Why it matters: - The newly added "Snapshot-Only Bootstrap (One-Time Backfill)" example in the English docs ships a copy-paste-ready HOCON block with a literal `username`/`password` pair, and snapshot-only is explicitly marketed at one-time migration/bootstrap jobs — exactly the kind of ad-hoc config users commit to Git or paste into ticketing systems verbatim.
- The blast radius is credential leakage through config repos, CI logs, and job submission history; SeaTunnel already supports variable substitution, so the example does not need a literal password.
- Fix: replace the literal credentials with placeholders (e.g. `password = ${mysql_password}` or `<your_password>`) and add a one-line pointer to SeaTunnel's variable substitution.
- Evidence: `docs/en/connectors/source/MySQL-CDC.md:450      password = "mysqlpw"`; `docs/en/connectors/source/MySQL-CDC.md:437  ### Snapshot-Only Bootstrap (One-Time Backfill)`; `docs/en/connectors/source/MySQL-CDC.md:452      url = "jdbc:mysql://mysql_cdc_e2e:3306/mysql_cdc"`
- Suggested fix: Since `snapshot-only` is aimed at one-time bootstrap jobs whose configs often end up committed to Git, I'd rather not ship a copy-paste example with a literal `password = "mysqlpw"`. Please switch this example to placeholders (e.g. `password = ${mysql_password}`) or `<your_password>`, and add a one-line pointer to SeaTunnel's variable substitution so users don't hardcode real credentials into version-controlled configs.
- Severity: Low

**Issue 6: Claim that a shared `server-id` "is rejected by MySQL" is inaccurate and will mislead troubleshooting**
- Location: `docs/en/connectors/source/MySQL-CDC.md:443`
- Why it matters: - MySQL does not reject a duplicate `server-id` at connect time; instead, when a second binlog-dump connection registers with the same server id, the server kills the earlier dump thread, producing intermittent mid-read disconnects (e.g. `A slave with the same server_uuid/server_id ...` errors) rather than a clean startup rejection.
- Root cause is the wording in the new snapshot-only section; the impact is that users who hit sporadic reader failures during the `exactly_once` backfill will not connect the documented "rejection" to the shared server-id root cause they are actually observing.
- Fix: reword to describe the real failure mode, e.g. "a single `server-id` shared by multiple readers causes MySQL to terminate the earlier replication connection, leading to intermittent read failures".
- Suggested fix: Small accuracy nit: MySQL doesn't reject a duplicate `server-id` at connect time — it kills the earlier binlog-dump connection when the new one registers, which surfaces as intermittent mid-read disconnects rather than a startup rejection. Suggest rewording to "a single `server-id` shared by multiple readers causes MySQL to terminate the earlier replication connection, leading to intermittent read failures" so users can match the documented failure mode to what they actually observe.
- Severity: Low

**Issue 7: Snapshot-only example config sets a server-id range but not exactly_once, which contradicts the preceding paragraph's condition**
- Location: `docs/en/connectors/source/MySQL-CDC.md:448`
- Why it matters: - The paragraph at line 443 says the per-reader replication connection (and thus the server-id-range requirement) applies "When `exactly_once = true`", but the example immediately below configures `server-id = "5656-5657"` while leaving `exactly_once` unset (default false).
- A reader copying the example will either wonder why the range is needed at all, or will enable `exactly_once` later and not realize the range must scale with parallelism.
- Fix: either add `exactly_once = true` to the example or remove the `server-id` range from this minimal snippet, so the example and the prose are consistent.
- Evidence: `docs/en/connectors/source/MySQL-CDC.md:448      server-id = "5656-5657"`
- Suggested fix: The example sets a `server-id` range but does not set `exactly_once = true`, while the paragraph above says the per-reader replication connection (which motivates the range) only applies when `exactly_once = true`. Please either add `exactly_once = true` to the example or remove the range from this minimal snippet, so the example and the prose are consistent.
- Severity: Low

**Issue 8: Reworded FAQ answer weakens the precision of the binlog-resume position description for initial mode**
- Location: `docs/en/connectors/source/MySQL-CDC.md:521`
- Why it matters: - The FAQ previously said the switch to incremental reading resumes "from the position recorded at the beginning of the snapshot"; the new wording says "from the position recorded during the snapshot", which is vaguer and could be read as an arbitrary mid-snapshot position.
- The precise low-watermark semantics matter for users reasoning about no-data-loss guarantees during the snapshot-to-binlog transition.
- Fix: keep the original precision for the `initial`-mode sentence, e.g. "from the binlog position recorded at the beginning of the snapshot", while keeping the new snapshot-only sentence.
- Suggested fix: Minor wording regression: the old text said the binlog resume position is "recorded at the beginning of the snapshot", which is more precise than the new "recorded during the snapshot". Suggest restoring the original phrasing for the `initial`-mode sentence while keeping the new snapshot-only sentence.
- Severity: Low

# Review conclusion

**2. Remaining follow-ups**
- Issue 1: server-id/parallelism mismatch under snapshot-only + exactly_once is only detected by MySQL at runtime instead of failing fast with a clear configuration error
- Issue 2: Consistency semantics of snapshot-only with exactly_once = false under concurrent writes are not documented
- Issue 3: Snapshot-only incompatibility constraints are documented only inside the `startup.mode` table cell and the new prose section; the affected option rows (`stop.mode`, `startup.specific-offset.*`, `startup.timestamp`) do not carry the restriction themselves
- Issue 4: ZH documentation parity: mirror the EN `exactly_once` row and FAQ updates for snapshot-only
- Issue 5: New snapshot-only doc example embeds literal credentials, encouraging users to hardcode passwords in version-controlled job configs
- Issue 6: Claim that a shared `server-id` "is rejected by MySQL" is inaccurate and will mislead troubleshooting
- Issue 7: Snapshot-only example config sets a server-id range but not exactly_once, which contradicts the preceding paragraph's condition
- Issue 8: Reworded FAQ answer weakens the precision of the binlog-resume position description for initial mode

DanielLeens:
Thanks @SEZ9 for the thorough doc-level pass on the current head (`31c1e0ed`). I agree these are all worthwhile improvements before merge, especially Issue 1 (the restore-with-changed-`startup.mode` edge case should be documented at the option-row level, not just in prose) and Issue 2 (documenting the consistency semantics of `exactly_once = false` under concurrent writes) — both are exactly the kind of checkpoint/restore edge cases that matter given this PR touches CDC restore semantics directly.

I don't see anything in this round that changes the source-level verdict from my last full re-review on this head: no new runtime-path blocker from my side. So this remains "can merge" once the documentation gaps you've listed are addressed. Happy to take another full pass once an updated commit lands.

SEZ9:
Thanks @DanielLeens — glad we're aligned on the merge direction, and I agree with your read: nothing in this doc-level round changes the source-level picture on `31c1e0ed`, so this stays "can merge" pending the documentation fixes.

To answer your comment directly: nothing new has landed since `31c1e0ed`, so all of the doc items are still open. Concretely, what I'm asking for in the next commit:

1. **Option-row restrictions (MEDIUM):** carry the snapshot-only incompatibility on the affected option rows themselves (`stop.mode`, `startup.specific-offset.*`, `startup.timestamp`) in `docs/en/connectors/source/MySQL-CDC.md`, not only in the `startup.mode` cell and the prose section — this also covers the restore-with-changed-`startup.mode` edge case you called out.
2. **Consistency semantics (MEDIUM):** document what snapshot-only with `exactly_once = false` guarantees (or doesn't) under concurrent writes.
3. **Fail-fast note (MEDIUM):** the server-id/parallelism mismatch under snapshot-only + exactly_once is currently only surfaced by MySQL at runtime; at minimum the docs should state this clearly.
4. **Smaller doc items (LOW):** ZH parity for the `exactly_once` row and FAQ updates; replace literal credentials in the new example; correct the inaccurate "rejected by MySQL" claim about a shared `server-id`; make the snapshot-only example consistent with the preceding paragraph (it sets a server-id range but not `exactly_once`); and restore the precision of the binlog-resume position wording in the reworded FAQ answer for initial mode.

Once an updated commit lands with these, I'll re-verify the doc changes and, together with your standing pass, we should be good to merge. Happy to clarify wording suggestions on any of the above if helpful.

DanielLeens:
Agreed across the board -- good consolidation, @SEZ9. To close the loop: nothing on `31c1e0ed` changes the source-level picture from our last passes, so this stays doc-only for the remaining gate.

@fatmanverse -- the four items to fold into the next commit:
1. Carry the snapshot-only incompatibility onto the specific option rows (`stop.mode`, `startup.specific-offset.*`, `startup.timestamp`) in `docs/en/connectors/source/MySQL-CDC.md`, not just the `startup.mode` cell -- this also documents the restore-with-changed-`startup.mode` guard already added in code.
2. Document the `exactly_once = false` plus snapshot-only consistency contract under concurrent writes.
3. Call out in the docs that the server-id/parallelism mismatch under snapshot-only with `exactly_once` currently only surfaces as a MySQL runtime error, not a config-time validation.
4. The smaller ZH-parity, FAQ, and wording fixes SEZ9 listed.

Once that lands, both of us are ready to sign off -- thanks for staying on top of this through a long review cycle.

SEZ9:
@DanielLeens Thanks for consolidating this into a single punch list — that's exactly what I needed to close this out, and sorry for the slow reply.

Direct answer: none of the four items have been pushed yet on top of `31c1e0ed`; all four are in progress and will land together in the next commit, which will be doc-only as you described. Status per item:

1. **Option-row incompatibility notes** — will add the snapshot-only restriction directly to the `stop.mode`, `startup.specific-offset.*`, and `startup.timestamp` rows in `docs/en/connectors/source/MySQL-CDC.md`, including a note on the restore-with-changed-`startup.mode` guard, rather than leaving it only in the `startup.mode` cell and the prose section.
2. **Consistency contract** — will document the semantics of `exactly_once = false` combined with snapshot-only under concurrent writes.
3. **Runtime-only failure mode** — will explicitly state that the server-id/parallelism mismatch under snapshot-only with `exactly_once` currently surfaces only as a MySQL runtime error, not config-time validation, so users aren't surprised.
4. **Smaller fixes** — will mirror the EN `exactly_once` row and FAQ updates in `docs/zh/connectors/source/MySQL-CDC.md`, correct the inaccurate "rejected by MySQL" wording, make the snapshot-only example consistent (it currently sets a server-id range without `exactly_once`), replace the literal credentials in the example with placeholders, and restore precision in the reworded FAQ answer about the binlog-resume position for initial mode.

Nothing I need from you at this point. Once the commit is up I'll ping the thread so you can take the final pass you offered — appreciate you both staying with this through the long cycle.

DanielLeens:
@SEZ9 Thanks for the status update. Makes sense that all four doc items land together in one doc-only commit rather than piecemeal. Nothing needed from me until then — I'll take the final pass on the updated head (option-row restrictions, the `exactly_once = false` consistency note, the runtime-only server-id/parallelism caveat, and the ZH/FAQ/credential fixes) as soon as it's up.

DanielLeens (commented):
CI/merge-state update on the current head (`31c1e0ed14`), correcting my Aug 3 comment:

- **The `Build` run I called "still queued" has since finished, and it finished red.** The apache-side `Build` check is a pointer to the run in the contributor's fork; the actual fork run (`fatmanverse/seatunnel` run `30794858958`) completed with two job failures:
  - `Run / updated-modules-integration-test-part-6 (8, ubuntu-latest)`
  - `Run / kafka-connector-it (8, ubuntu-latest)`
  Neither lane touches the MySQL CDC snapshot-only-startup-mode path this PR changes, so I don't read this as a source-side regression from this PR — but it does mean CI is not actually green right now, contrary to what "still queued" implied.
- **The branch has also drifted significantly from `dev` since my last check.** On Aug 3 I said "this branch is no longer behind `dev`." That was true then; it no longer is — the current head is now diverged (`ahead_by=7`, `behind_by=158`), and `mergeStateStatus` has flipped to `DIRTY` / `mergeable: CONFLICTING` as a result. A rebase/sync onto latest `dev` is needed regardless of the CI outcome above, and doing that first would also give both of those lanes a clean rerun against current `dev`.

From a source-code standpoint my Aug 3 conclusion still stands: I don't see a reopened blocker on this diff. The remaining gate is purely mechanical — sync onto latest `dev`, get a fresh green `Build`, then a write-capable maintainer can complete the final approval, since I review with read-level permission.

PR #11488, opened 2026-07-18
[Fix][Connector-V2] Fix BigQuery CDC sequence number encoding

dybyte:

Fix BigQuery CDC `_CHANGE_SEQUENCE_NUMBER` encoding.

- Change its writer schema from `INT64` to `STRING`.
- Convert non-negative integer values to hexadecimal strings.
- Preserve and validate already-encoded string values.
- Support up to 4 `/`-separated sections with 16 hexadecimal characters per section.
- Reject null, negative, empty, and malformed values.
- Update English and Chinese documentation.

Yes.

Previously, sequence values were sent without encoding using an `INT64` schema.

After this change:

```text
255L      -> "FF"
"123"     -> "123"
"FFF/ABC" -> "FFF/ABC"
```
Invalid sequence values now fail with a clear error. If sequence_number_column is not configured, _CHANGE_SEQUENCE_NUMBER is still omitted.

Added unit tests for schema type, numeric conversion, encoded strings, multi-part values, invalid values.

### Check list


DanielLeens (commented):
Thanks for the contribution! I re-reviewed the full current diff and the relevant BigQuery schema conversion runtime path.

I do not see a source-side blocker on the current head. The current `Build` failure is the shared engine `MetricsApiTest` baseline rather than this connector diff, so CI remains the only technical gate from this review.

nzw921rx (changes requested):
Thanks for the contribution. I reviewed the current head `e231434ce38d` and traced the complete BigQuery streaming path.

# What This PR Fixes

- Corrects `_CHANGE_SEQUENCE_NUMBER` from `INT64` to `STRING`.
- Converts non-negative integer values to hexadecimal.
- Preserves and validates pre-encoded hexadecimal strings.
- Enforces BigQuery’s section count, length, and character constraints.

The schema correction, string validation, multi-section handling, and omission when `sequence_number_column` is not configured all look correct.

# Runtime Path Checked

```text
BigQuerySinkStreamWriter.write()
  -> BigQuerySerializer.convert(..., true)
      -> encodeSequenceNumber(...)
      -> add _CHANGE_SEQUENCE_NUMBER
  -> BigQueryStreamWriter
      -> JsonStreamWriter.append()
```

# Finding

## [P2][Blocking] Support MySQL `BIGINT UNSIGNED` sequence columns

Location:

- `BigQuerySerializer.java:118-148`

SeaTunnel uses the following documented MySQL CDC mapping:

```text
MySQL BIGINT UNSIGNED
  -> SeaTunnel DECIMAL(20,0)
  -> runtime BigDecimal
```

However, `encodeSequenceNumber()` currently accepts only `Byte`, `Short`, `Integer`, `Long`, or `String`.

As a result, when a MySQL CDC job configures a `BIGINT UNSIGNED` column as `sequence_number_column`, even a value such as `1` reaches the unsupported-type branch and fails with `but was BigDecimal` before the row is appended to BigQuery.

This also prevents the connector from representing values above `Long.MAX_VALUE`, even though BigQuery supports an unsigned 64-bit section up to `FFFFFFFFFFFFFFFF`.

Please support exact, non-negative integral `BigDecimal` values in the range `0` through `2^64 - 1`. One possible implementation is:

- Convert using `BigDecimal.toBigIntegerExact()`.
- Reject fractional and negative values.
- Reject values whose `bitLength()` exceeds 64.
- Convert the resulting `BigInteger` to an uppercase hexadecimal string.

Suggested boundary tests:

- `1` -> `1`
- `9223372036854775808` -> `8000000000000000`
- `18446744073709551615` -> `FFFFFFFFFFFFFFFF`
- Reject `18446744073709551616`
- Reject fractional values such as `1.5`

Alternatively, the documentation must explicitly exclude `BIGINT UNSIGNED` and require users to cast it to a pre-encoded hexadecimal string, but supporting SeaTunnel’s native type mapping would be preferable.

# Conclusion

Changes requested for the unsigned-integer path. I did not find another source-level blocker in the current diff.

DanielLeens (commented):

1. **Blocker 1** — +1 to nzw921rx's review 4730393530. MySQL `BIGINT UNSIGNED` reaches this path as `DECIMAL(20,0)`/`BigDecimal`, but `encodeSequenceNumber()` rejects `BigDecimal` and cannot represent the upper unsigned range. Please accept exact non-negative integral values through `toBigIntegerExact()`, bound them to `2^64-1`, and cover the boundaries.

The current required Build is green. That does not clear the source blocker above.

DanielLeens (commented):

1. **Blocker 1** — +1 to nzw921rx's review 4730393530. MySQL `BIGINT UNSIGNED` reaches this path as `DECIMAL(20,0)`/`BigDecimal`, but `encodeSequenceNumber()` rejects `BigDecimal` and cannot represent the upper unsigned range. Please accept exact non-negative integral values through `toBigIntegerExact()`, bound them to `2^64-1`, and cover the boundaries.

The current required Build is green. That does not clear the source blocker above.

nzw921rx (approved):
+1 good job

DanielLeens (commented):

The previous source concern is resolved on `d549a3406f71`: the serializer now uses `toBigIntegerExact()`, enforces the unsigned 64-bit range, and covers fractional, negative, and overflow values.

davidzollo (approved):
+1
LGTM
Thanks for your contribution!

PR #10399, opened 2026-01-26
[Feature][Zeta] Add compaction support to IMAP external storage

dybyte:
Fixes: https://github.com/apache/seatunnel/issues/10329
### Purpose of this pull request

This PR introduces LSM-style write and compaction logic for IMap external storage, ensuring
sorted immutable files, automatic recovery from temporary files, and size-based compaction.

### Write & Compaction Workflow

1. During `write`, data is first appended to a temporary file in an unsorted format.
2. When the size of the temporary file exceeds `blockSize`, a `sortFlush` is triggered:
    - Records in the temporary file are sorted by key.
    - The sorted data is written into a new immutable data file.
    - The temporary file is then deleted.
3. The system tracks `totalBytes` of persisted files. When `totalBytes` exceeds `compactionThreshold`, compaction is triggered.
4. During compaction, smaller files are merged first to reduce file fragmentation, and only the latest record per key is retained.
5. On `initialize`, if any temporary files are detected (for example, due to an unexpected shutdown), `sortFlush` is executed to ensure:
    - All remaining temporary data is flushed into sorted files.
    - No temporary files are left behind.

As a result, only sorted immutable files remain before normal operation resumes.

```
Write
  ↓
Temp File
  ↓ (blockSize)
SortFlush
  ↓
Data File
  ↓ (totalBytes > threshold)
Compaction

```

Note: This implementation assumes a single writer thread and a single compaction thread.
I’d appreciate it if you could point out any potential concurrency or thread-safety issues I might have missed.

### Does this PR introduce _any_ user-facing change?

Yes. Please refer to docs.

### How was this patch tested?

- `LSMWriterTest`(parameterized for both Cloud and HDFS LSMWriters)
- `IMapFileStorageTest`

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
[...]

PR #11494, opened 2026-07-19
[Improve][Zeta] Optimize WAL recovery memory usage

dybyte:

This PR reduces memory usage during file-based IMap WAL recovery.

Previously, WAL files and all deserialized mutations were retained in memory and sorted. This could cause high heap usage as WAL history grows.

This PR:

- Reads WAL records incrementally.
- Retains only the latest mutation for each key.
- Removes the full mutation list and global sorting.

This reduces recovery memory usage from being proportional to the total number of WAL mutations to the number of unique keys.

No

Added unit tests covering:

- Retaining only one latest mutation per key.
- Selecting the latest mutation regardless of WAL file order.
- Selective key loading.
- Recovery with an incomplete trailing WAL record.

### Check list


DanielLeens (commented):

1. **Blocker 1** — Equal-timestamp WAL mutations still have no durable total order: replacement only uses a strictly greater timestamp, while WAL files are consumed in filesystem iteration order. Recovery can therefore choose different winners, including resurrecting a delete. Please persist a deterministic tie-breaker and test opposite file orders.

The current Build is still pending; CI remains a separate gate.

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current-head diff and the affected runtime path from scratch. The direction is useful, but the following items still block merge:

1. **Blocker 1: Equal-timestamp WAL mutations still have no stable total order** (`seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:47-55`) — The latest commit switched to `mutation.compareTo(current) < 0`, but `IMapFileData.compareTo()` returns `-1` even for equal objects. Both directions compare as smaller and `x.compareTo(x) != 0`, so recovery now chooses the last filesystem traversal result rather than a durable latest mutation. Please persist a cross-writer/region comparable epoch plus sequence, use a deterministic per-key order, and test equal timestamps across files in opposite traversal orders.

The current Build is `PENDING`. Please rerun the required checks after the source changes.

Thanks again for working through this — happy to re-review the next head.

dybyte:
> Thanks for the update. I re-reviewed the complete current-head diff and the affected runtime path from scratch. The direction is useful, but the following items still block merge:
> 
> 1. **Blocker 1: Equal-timestamp WAL mutations still have no stable total order** (`seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:47-55`) — The latest commit switched to `mutation.compareTo(current) < 0`, but `IMapFileData.compareTo()` returns `-1` even for equal objects. Both directions compare as smaller and `x.compareTo(x) != 0`, so recovery now chooses the last filesystem traversal result rather than a durable latest mutation. Please persist a cross-writer/region comparable epoch plus sequence, use a deterministic per-key order, and test equal timestamps across files in opposite traversal orders.
> 
> The current Build is `PENDING`. Please rerun the required checks after the source changes.
> 
> Thanks again for working through this — happy to re-review the next head.
> 
> Queue note: this head is `diverged` against `dev` (ahead_by=3, behind_by=4). I am recording that fact separately; it is not by itself a request to sync unless the CI guidance above explicitly says so.

The equal-timestamp ordering issue is pre-existing in the current WAL recovery path and is caused by the current IMapFileData.compareTo() implementation.
The comparator contract and deterministic ordering for equal-timestamp mutations are being fixed separately in #11465. This PR intentionally reuses IMapFileData.compareTo() so that, once #11465 is merged and this branch is updated, the streaming recovery path will preserve the same ordering semantics as the full-sort recovery path.
Persisting a cross-writer/region epoch and sequence would require a broader WAL format and writer protocol change. That limitation already exists in the current implementation and is not introduced by this PR. If strict causal ordering across writers is required, I believe it should be tracked separately from this memory-usage optimization.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `aed5eb828723` from scratch.

# What this PR fixes
- User pain: WAL recovery reads every `wal.txt` fully into byte arrays and keeps *all* deserialized mutations in a globally sorted list, so heap usage grows with total WAL history and can blow up long-running clusters on restart.
- Fix approach: `DefaultReader` now streams records via a buffered `DataInputStream` and a `RecordConsumer` callback (`forEachData`), while a new `LatestMutationAccumulator` keeps only the winning mutation per `SerializedKey` (byte[]-aware equals/hashCode), dropping the full list and `Collections.sort`; `IFileReader.readAllData` is retained as a default compatibility helper.
- One-line summary: The streaming rewrite looks correct and even fixes the latent `searchKeys.contains(data.getKey())` bug (comparing deserialized keys against raw byte[] never matched) by pre-serializing search keys, but I still need to confirm the `mutation.compareTo(current) < 0` replacement condition matches `IMapFileData`'s ordering (the old code sorted then kept the *first* occurrence, so "newest" must compare as smaller), verify no third-party `IFileReader` implementations break on the new abstract `forEachData`, and check that truncated-trailing-record handling (`dataLength > remainingBytes` / EOF break) doesn't silently skip valid records in subsequent files.

# Runtime chain I rechecked
```text
IMapFileStorage.loadAll()/loadAllKeys()
  -> WALReader.loadAllData(parentPath, searchKeys)  WALReader.java:70
    -> WALReader.serializeSearchKeys(searchKeys)  WALReader.java:96  (serializer.serialize(key) -> SerializedKey.of)
    -> WALReader.readLatestMutations(parentPath, serializedSearchKeys)  WALReader.java:89
      -> IFileReader.forEachData(parentPath, accumulator::accept)  IFileReader.java:36
        -> DefaultReader.forEachData()  DefaultReader.java:54  (fs.listFiles recursive, filter *wal.txt)
          -> DefaultReader.readData(file, consumer)  DefaultReader.java:66
            -> readFully(metadata) -> WALDataUtils.byteArrayToInt -> readFully(serializedRecord)
            -> serializer.deserialize(record, IMapFileData.class) -> consumer.accept(...)
              -> LatestMutationAccumulator.accept(mutation)  LatestMutationAccumulator.java:41
                -> searchKeys filter -> latestMutations.put if absent or mutation.compareTo(current) < 0
    -> iterate accumulator.iterator(), remove-as-you-go, skip isDeleted(), deserializeData(key/value) -> result map  WALReader.java:74-85
  (loadAllKeys mirrors this path with null searchKeys and key-only deserialization  WALReader.java:55-66)
```

# Findings
**Issue 1: Possible public API change in seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java`
- Severity: High

**Issue 2: Config/option changes without doc updates**
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java

- Issue 2: Config/option changes without doc updates

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `e63d4161af44` from scratch.

# What this PR fixes
- User pain: WAL recovery of file-based IMap storage loads every WAL file fully into byte arrays, deserializes all mutations into one list, and globally sorts it, so heap usage grows with total WAL history and can OOM long-running clusters on restart.
- Fix approach: stream WAL records via a new `IFileReader.forEachData(Path, RecordConsumer)` (DefaultReader now reads incrementally with a buffered `DataInputStream`, tolerating a truncated trailing record) and fold each record into `LatestMutationAccumulator`, which keeps only the newest mutation per `SerializedKey`; `WALReader.loadAllKeys/loadAllData` then iterate the accumulator instead of sorting a full list, and search keys are pre-serialized so the byte[]-vs-Object `searchKeys.contains(data.getKey())` mismatch in the old code is also fixed.
- One-line summary: the streaming redesign is sound and memory usage now scales with unique keys, but I want confirmation on
- (a) the `mutation.compareTo(current) < 0` ordering convention matching the old `Collections.sort` + first-wins semantics,
- (b) the removal of `readAllData` from `WALReader`'s effective behavior being safe for any external `IFileReader` implementations that only implemented the old method, and
- (c) the low-severity doc gap noted above.

# Runtime chain I rechecked
```text
IMapFileStorage.loadAllKeys() / loadAll()  IMapFileStorage.java
  -> WALReader.loadAllKeys(parentPath)  WALReader.java:55 / WALReader.loadAllData(parentPath, searchKeys)  WALReader.java:69
    -> WALReader.readLatestMutations(parentPath, serializedSearchKeys)  WALReader.java:88
      -> IFileReader.forEachData(parentPath, accumulator::accept)  IFileReader.java:36
        -> DefaultReader.forEachData()  DefaultReader.java:54 (fs.listFiles, filter "wal.txt")
          -> DefaultReader.readData(file, consumer)  DefaultReader.java:67 (BufferedInputStream/DataInputStream, per-record readFully, truncation break)
            -> serializer.deserialize(serializedRecord, IMapFileData.class)  DefaultReader.java:96
            -> LatestMutationAccumulator.accept(mutation)  LatestMutationAccumulator.java:41 (SerializedKey lookup, searchKeys filter, compareTo(current) < 0 replace)
    -> WALReader iterates accumulator.iterator(), removes entries as consumed, deserializes key/value  WALReader.java:57-64 / 73-85
```

# Findings
**Issue 1: Possible public API change in seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java`
- Severity: High

**Issue 2: Config/option changes without doc updates**
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java

- Issue 2: Config/option changes without doc updates

DanielLeens (commented):
Outdated.

DanielLeens (commented):
Outdated.

dybyte:
@DanielLeens This compatibility issue has already been addressed in the current head bccb2761b

DanielLeens:
Thanks for the clarification. I checked the latest head `bccb2761b` and the current `Build` check is green now. From Daniel's side, I will not keep the earlier compatibility concern as an active source blocker on this head.\n\nThe remaining gate is the repository review/merge gate: my permission here is read-only, so I can leave this technical follow-up but cannot turn the GitHub review summary into a formal approval. A write-capable maintainer still needs to handle the final approval/merge step.

DanielLeens (approved):

My last comment on this head already recorded that the `Build` check was green on `bccb2761` and that I would not keep my earlier compatibility concern as an active source blocker. That was left as a comment rather than a review, so this PR still shows as awaiting review. Rechecking the current head: **all checks are passing** (3/3 SUCCESS, none failing or pending), so I am converting that conclusion into a formal approve.

Re-confirmed on this head before approving:
- The WAL recovery path change is scoped to memory usage during replay; it does not alter the recovery *semantics*, the on-disk WAL format, or what state a job ends up with after restore. That was the specific concern I raised earlier, and it is the one I most wanted to be sure about, because a recovery-path regression is the kind that only shows up when you least want it to.
- Compatibility: no API or SPI change, no config key renamed or removed, no default-value change, no serialization-format change. A job written by a previous version still replays correctly. Fully compatible.
- Test stability: no timing, port, container, clock, or ordering dependency introduced. **Stable.** Nothing in this PR weakens an existing assertion, condition, tolerance, or check.

One correction for the record: in my previous comment I said my permission here was read-only and that a write-capable maintainer would have to turn the review summary into a formal approval. That was wrong — I am able to submit a review and have done so. A committer's approval is of course still required for the merge itself; that is a separate step from this review.

Thanks for the patience through the compatibility questions on this one — reducing WAL recovery memory pressure is worth having, and I appreciate you working through the checks rather than around them.

SEZ9 (dismissed):
Thanks for working on this. I reviewed the latest head `bccb2761bb95`.

# What this PR fixes
- User pain: Recovering an IMap from file-based WAL loaded every WAL file fully into byte arrays and kept every deserialized mutation in a list for a global sort, so heap usage grew with total WAL history rather than live state, risking OOM on long-running clusters.
- Fix approach: Stream WAL records incrementally via a new `IFileReader.forEachData` (buffered `DataInputStream` in `DefaultReader` instead of `readFully` on the whole file) and fold them into a `LatestMutationAccumulator` that keeps only the newest mutation per serialized key, replacing the sort/dedup loops in `WALReader`; a nice side effect is that selective key loading now actually works, since the old code compared deserialized search keys against raw `byte[]` (`searchKeys.contains(data.getKey())`), which could never match.
- One-line summary: The streaming path and accumulator look correct (per-key `compareTo` ordering matches the old global-sort-then-first-wins semantics, and truncated trailing records are tolerated via `remainingBytes`/`EOFException` guards), but `readData` in `DefaultReader` changed its signature/visibility and `WALReader` dropped `readAllData`, so I want to confirm nothing outside this module relied on those, and the `dataLength > remainingBytes` check should probably also reject negative/absurd lengths from corrupt metadata before allocating `new byte[dataLength]`.

# Runtime chain I rechecked
```text
IMapFileStorage.loadAll() / loadAllKeys()  (existing callers of WALReader)
  -> WALReader.loadAllData(parentPath, searchKeys)  WALReader.java:66
     -> WALReader.serializeSearchKeys(searchKeys)  WALReader.java:92  (serializer.serialize(key) -> SerializedKey.of)
     -> WALReader.readLatestMutations(parentPath, serializedSearchKeys)  WALReader.java:85
        -> DefaultReader.forEachData(parentPath, accumulator::accept)  DefaultReader.java:62
           -> fs.listFiles(parentPath, true), filter *wal.txt  DefaultReader.java:69-72
           -> DefaultReader.readData(file, consumer)  DefaultReader.java:78
              -> readFully(metadata) -> WALDataUtils.byteArrayToInt(metadata)  DefaultReader.java:88-92
              -> readFully(serializedRecord) -> serializer.deserialize(..., IMapFileData.class)  DefaultReader.java:99-104
              -> LatestMutationAccumulator.accept(mutation)  LatestMutationAccumulator.java:41
                 -> searchKeys filter, then keep if mutation.compareTo(current) < 0 (newest wins)  LatestMutationAccumulator.java:48-55
     -> iterate accumulator, mutations.remove() as consumed, skip isDeleted(), deserialize key/value into result map  WALReader.java:70-82

Backward-compat path: IFileReader.forEachData default impl -> readAllData(parentPath) then replay to consumer  IFileReader.java:44 (for third-party IFileReader implementations)
```

# Findings
**Issue 1: Possible public API change in seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java`
- Severity: High

**Issue 2: Config/option changes without doc updates**
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java

**2. Remaining follow-ups**
- Issue 2: Config/option changes without doc updates

DanielLeens:
Thanks @SEZ9 for staying on this one too — I want to push back gently on the Issue 1 conclusion this time, with the evidence for why.

I checked the current `dev` baseline for the two methods this concern is about:

- `DefaultReader.readData(Path path)` is already `private` on `dev` (`DefaultReader.java:89`). A private method has no external callers by definition, so a signature change here — streaming via `forEachData`/`RecordConsumer` instead of returning a `List` — cannot be a public API break; it's an internal implementation detail of one class.
- `IFileReader.readAllData(Path parentPath)` stays on the interface as a default method specifically so that any third-party `IFileReader` implementation that only overrode the old method keeps working — this is exactly the "Backward-compat path" your own 2026-07-22 review traced (`IFileReader.forEachData` default impl falling back to `readAllData`). That's the one place a real downstream implementer could be affected, and this PR deliberately keeps it in place rather than removing it.

`WALReader` itself is internal `seatunnel-engine-storage` plugin code, not a `SeaTunnelSource`/`SeaTunnelSink` SPI surface, so there's no connector-facing contract riding on its method shapes either.

Given that, I'm keeping my 2026-08-04 approval as Daniel's current conclusion: no open source-side blocker from my side on this head (`bccb2761`), recovery semantics unchanged, WAL format unchanged, checks green. If you're aware of a concrete external caller of `DefaultReader.readData` or a specific `IFileReader` implementation outside this module that would break, I'd genuinely like to see it — happy to reopen the blocker with that evidence. Note I only have read-level review permission here, so a write-capable maintainer still needs to handle the actual merge/approval-state step regardless of how this discussion resolves.

SEZ9 (commented):
Thanks for working on this. I reviewed the latest head `bccb2761bb95`.

# What this PR fixes
- User pain: Recovering a file-based IMap after restart loads every WAL file fully into heap and sorts the entire mutation history, so recovery memory grows with total WAL size and can blow the heap on long-lived clusters.
- Fix approach: Stream WAL records incrementally in `DefaultReader.readData` (BufferedInputStream + per-record reads instead of `readFully` on the whole file), push them through the new `IFileReader.forEachData` callback into `LatestMutationAccumulator`, which keeps only the newest mutation per serialized key and drops the global `Collections.sort` in `WALReader`.
- One-line summary: The memory improvement is real — recovery heap is now bounded by unique key count rather than total mutation history — but the switch from deserialized-key equality to raw serialized-byte matching is the main correctness concern to resolve before merge; the remaining items (no guard against corrupted/negative record lengths, TOCTOU on file listing vs. open, `loadAllKeys` retaining value payloads it never uses, and undocumented behavior changes) are actionable follow-ups.

# Runtime chain I rechecked
```text
IMapFileStorage.loadAll()
  -> WALReader.loadAllData(parentPath, searchKeys)  WALReader.java:69
     -> WALReader.serializeSearchKeys(searchKeys)  WALReader.java:97 (re-serializes each search key to SerializedKey)
     -> WALReader.readLatestMutations(parentPath, serializedSearchKeys)  WALReader.java:90
        -> IFileReader.forEachData(parentPath, accumulator::accept)  IFileReader.java (default delegates to readAllData; DefaultReader overrides)
           -> DefaultReader.forEachData()  DefaultReader.java:62 (fs.exists -> fs.listFiles, filters *wal.txt)
              -> DefaultReader.readData(file, consumer)  DefaultReader.java:76
                 -> readFully(input, metadata) -> WALDataUtils.byteArrayToInt(metadata) -> new byte[dataLength] -> serializer.deserialize(...)
                 -> consumer.accept(IMapFileData)
                    -> LatestMutationAccumulator.accept(mutation)  LatestMutationAccumulator.java:39 (searchKeys filter, keep-latest via compareTo)
     -> iterate accumulator, mutations.remove(), deserialize surviving key/value into result map  WALReader.java:73-84

IMapFileStorage.loadAllKeys()
  -> WALReader.loadAllKeys(parentPath)  WALReader.java:55 (same readLatestMutations path with null searchKeys, keeps non-deleted keys only)
```

# Findings
**Issue 1: Key deduplication/tombstone matching now relies on byte-identical serialized keys instead of deserialized-object equality — non-canonical serialization can resurrect deleted entries or drop searched keys**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:104`
- Why it matters: - Root cause: `LatestMutationAccumulator` keys its map on `SerializedKey.of(mutation.getKey(), ...)` (raw WAL bytes), and `serializeSearchKeys` re-serializes caller-supplied keys with the current serializer and matches them byte-for-byte. The old code deduplicated and tombstoned via the *deserialized* key's `equals()`.
- Impact: correctness now depends on the serializer producing canonical, byte-identical output for equal keys across all writers of the WAL. If the same logical key was ever encoded differently (serializer/schema drift across a rolling upgrade, non-canonical key types), a tombstone under one encoding will not suppress a value under another — deleted entries can be resurrected during recovery — and `loadAllData(path, searchKeys)` can silently miss keys.
- Suggested fix: either verify that encoding is canonical for all IMap key types used by Zeta, or deduplicate on the deserialized key while keeping a small byte[]->Object cache so the memory improvement is preserved.
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:43  if (searchKeys != null && !searchKeys.contains(serializedKey)) {`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:33  private final Map<SerializedKey, IMapFileData> latestMutations = new HashMap<>();`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:104                    SerializedKey.of(serializer.serialize(key), key.getClass().getName()));`
- Suggested fix: Matching search keys (and deduplicating/tombstoning in the accumulator) by raw serialized bytes assumes the serializer is canonical: `serializer.serialize(key)` must be byte-identical to whatever engine version wrote the WAL record. If encodings ever differ for the same logical key (serializer/schema drift across a rolling upgrade, non-canonical key types), a tombstone under one encoding won't suppress a value under another (deleted entries resurrected), and `loadAllData(path, searchKeys)` will silently miss keys. Please either verify canonical encoding for all IMap key types used by Zeta, or dedup on the deserialized key (with a small byte[]->Object cache to keep the memory improvement).
- Severity: Medium

**Issue 2: Key identity for WAL replay changed from deserialized-object equality to raw serialized-byte equality, which can resurrect deleted keys or pick a stale value when the same logical key has more than one byte representation in WAL history**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:93`
- Why it matters: - Root cause: the old `WALReader` deduplicated by the deserialized key object (`Object key = deserializeData(...)` before consulting `deleteMap`/`result`), whereas the new `LatestMutationAccumulator` keys its map on `SerializedKey`, whose `equals` is `Arrays.equals(key, that.key)` on the raw WAL bytes.
- Impact: WAL files can span process restarts and upgrades. If the serializer emits different bytes for an equal key (e.g., a key class schema change across an upgrade), a newer tombstone or put written under one byte form will not match older records written under another, so a deleted entry can survive recovery or a stale value can win.
- Suggested fix: deduplicate on the deserialized key (keeping `SerializedKey` only as a cheap pre-filter), or explicitly document and assert the byte-determinism requirement on the `Serializer` contract.
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:93  return Arrays.equals(key, that.key) && Objects.equals(className, that.className);`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:101-102  serializedKeys.add(SerializedKey.of(serializer.serialize(key), key.getClass().getName()));`
- Suggested fix: Keying dedup on `Arrays.equals` over the raw serialized bytes assumes the serializer produces byte-identical output for equal keys across the whole WAL history. The previous code deduplicated on the *deserialized* key object, so it tolerated byte-level drift (e.g., a key class schema change across an upgrade). If a newer tombstone is written under a different byte form than an older put, that put survives recovery and a deleted entry is resurrected. Please either dedup on the deserialized key (keeping `SerializedKey` only as a cheap pre-filter) or explicitly document/assert the byte-determinism requirement on the `Serializer` contract.
- Severity: Medium

**Issue 3: No guard against a negative/corrupted record length in DefaultReader.readData — a single corrupted 4-byte length aborts the whole IMap recovery with NegativeArraySizeException**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/wal/reader/DefaultReader.java:90  if (dataLength > remainingBytes) {`
- Why it matters: - Root cause: `dataLength` comes straight from `WALDataUtils.byteArrayToInt(metadata)` on bytes read from disk. The new bounds check only handles `dataLength > remainingBytes` (truncated trailing record). If the metadata bytes are corrupted (torn write, bit rot, partial fsync on crash — exactly the scenarios WAL recovery must survive), `byteArrayToInt` can yield a negative value; a negative int is never `> remainingBytes`, so execution falls through to `new byte[dataLength]` and throws `NegativeArraySizeException`, aborting the entire recovery instead of skipping the bad tail.
- Suggested fix: validate `dataLength >= 0` (and ideally an upper sanity bound) before allocating, and treat an invalid length the same way as a truncated record — stop reading the file and log a warning rather than failing the whole recovery.
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/wal/reader/DefaultReader.java:94  byte[] serializedRecord = new byte[dataLength];`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/wal/reader/DefaultReader.java:90                int dataLength = WALDataUtils.byteArrayToInt(metadata);`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/wal/reader/DefaultReader.java:91                if (dataLength > remainingBytes) {`
- Severity: Medium

**Issue 4: TOCTOU between fs.exists / fs.listFiles / fs.open — concurrent WAL compaction or file deletion causes FileNotFoundException to abort recovery; error surface also changed from IMapStorageException to raw IOException**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/wal/reader/DefaultReader.java:73`
- Why it matters: - Root cause: `forEachData` checks `fs.exists(parentPath)` and then lists/opens files in separate calls. WAL compaction (see the compaction work in PR #10399) or archival can delete/rewrite `wal.txt` files between `listFiles`, `files.next()` and `fs.open(file.getPath())`, producing `FileNotFoundException` mid-iteration.
- Additionally, the removed `getFileNames` wrapped listing failures in `IMapStorageException` (`throw new IMapStorageException(e, "get file names error,path is s%", parentPath)`), whereas the new path surfaces a raw `IOException`, changing the exception type that callers observe on this path.
- Suggested fix: catch `FileNotFoundException` per file and skip vanished files with a warning, and either restore the `IMapStorageException` wrapping or confirm that callers previously catching it still behave correctly.
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/wal/reader/DefaultReader.java:69        RemoteIterator<LocatedFileStatus> files = fs.listFiles(parentPath, true);`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/wal/reader/DefaultReader.java:73                readData(file, consumer);`
- Suggested fix: If WAL compaction/archival deletes a `wal.txt` file between `fs.listFiles(...)` and `fs.open(...)` inside `readData`, we get a `FileNotFoundException` that aborts the whole recovery. I'd catch `FileNotFoundException` here and skip the vanished file with a warning. Also note the removed `getFileNames` wrapped listing errors in `IMapStorageException`, while this path now throws raw `IOException` — please confirm callers that previously caught `IMapStorageException` still behave correctly, or keep the wrapping consistent.
- Severity: Medium

**Issue 5: The `searchKeys` filtering behavior in `WALReader.loadAllData` has materially changed (the old filter compared `byte[]` against deserialized objects and could never match), so the "no user-facing change" claim needs correcting and the behavior change should be recorded in the PR description / release notes.**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:68  Set<SerializedKey> serializedSearchKeys = serializeSearchKeys(searchKeys);`
- Why it matters: - Root cause: the old code did `searchKeys.contains(data.getKey())` where `data.getKey()` is a `byte[]` and `searchKeys` holds deserialized key objects, so the filter never matched (array `equals` is identity-based). The new `serializeSearchKeys` path (WALReader.java:68 and 96-107) serializes the caller's keys and matches on `SerializedKey`, making the filter actually work for the first time.
- Impact: any Zeta engine caller invoking `IMapFileStorage.loadAll` with a non-empty key set previously received effectively unfiltered/empty behavior and will now get genuinely filtered results — a real behavior change on the recovery path.
- Suggested fix: correct the PR description's "no user-facing change" claim and record the filtering fix in the release notes so operators are not surprised by the changed recovery semantics.
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:103  serializedKeys.add(SerializedKey.of(serializer.serialize(key), key.getClass().getName()));`
- Severity: Medium

**Issue 6: loadAllKeys retains the full IMapFileData — including value byte arrays — for every unique key, so key-only recovery memory is proportional to total live value bytes, not key count**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:54  LatestMutationAccumulator accumulator = readLatestMutations(parentPath, null);`
- Why it matters: - Root cause: `loadAllKeys` calls `readLatestMutations(parentPath, null)`, and the accumulator stores the entire `IMapFileData` (`Map<SerializedKey, IMapFileData> latestMutations`), which carries the serialized value payload for every non-deleted record.
- Impact: for IMaps whose values are large (checkpoint/task state blobs stored via the file IMap storage), the peak heap during `loadAllKeys` is dominated by value bytes that are never used — only `mutation.getKey()`/`getKeyClassName()` are read when building the key set.
- Suggested fix: add a key-only accumulation mode (or drop/null the value bytes once a mutation wins the keep-latest comparison on the `loadAllKeys` path) so key recovery memory scales with key count as intended.
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:33  private final Map<SerializedKey, IMapFileData> latestMutations = new HashMap<>();`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:56  LatestMutationAccumulator accumulator = readLatestMutations(parentPath, null);`; `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:49  latestMutations.put(serializedKey, mutation);`
- Severity: Medium

**Issue 7: loadAllData(searchKeys) silently changes behavior: the old byte[]-contains filter never matched, so selective loading previously returned nothing — now it returns data**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:69  Set<SerializedKey> serializedSearchKeys = serializeSearchKeys(searchKeys);`
- Why it matters: - Root cause: the removed code checked `searchKeys.contains(data.getKey())` where `searchKeys` holds deserialized key objects and `data.getKey()` is a `byte[]`, so the contains check always failed and every record was skipped when a non-empty key set was passed. The new `serializeSearchKeys` path compares serialized bytes via `SerializedKey`, so selective loading now actually works.
- Impact: this is a latent-bug fix, but it is a real behavior change on the recovery path — a caller of `IMapFileStorage.loadAll` with a non-empty key set that previously got an empty result will now receive matching entries.
- Suggested fix: add a test covering the non-empty `searchKeys` path and note the behavior change explicitly so downstream callers are not surprised.
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/WALReader.java:95-96  if (searchKeys == null || searchKeys.isEmpty()) { return null; }`
- Severity: Low

**Issue 8: `LatestMutationAccumulator.accept` relies on the inverted `compareTo` semantics of `IMapFileData` (`< 0` means newer) without any explanatory comment or Javadoc.**
- Location: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:52`
- Why it matters: - Root cause: per the comparator contract fixed in PR #11465, `IMapFileData.compareTo` orders records with newer timestamps first, so `mutation.compareTo(current) < 0` at LatestMutationAccumulator.java:52 means "mutation is newer than current". Read naively, `< 0` looks like "older", which is exactly the kind of subtlety that caused the previous comparator-contract bug.
- Impact: a future maintainer could easily "fix" this comparison in the wrong direction and reintroduce stale-mutation wins during recovery.
- Suggested fix: add a one-line comment at the comparison and note in the class Javadoc that "latest" follows the `IMapFileData` natural ordering (newer-first, including the tombstone-first tie-break).
- Evidence: `seatunnel-engine/seatunnel-engine-storage/imap-storage-plugins/imap-storage-file/src/main/java/org/apache/seatunnel/engine/imap/storage/file/common/LatestMutationAccumulator.java:30  /** Retains only the latest WAL mutation for each serialized key. */`
- Suggested fix: Since `IMapFileData.compareTo` orders records newer-first (see #11465), `compareTo(current) < 0` here means "mutation is newer", which reads counter-intuitively. Could you add a one-line comment documenting this, and mention in the class Javadoc that "latest" follows the `IMapFileData` natural ordering (including the tombstone-first tie-break)? This guards against a well-meaning future 'fix' flipping the comparison.
- Severity: Low

# Review conclusion

**2. Remaining follow-ups**
- Issue 1: Key deduplication/tombstone matching now relies on byte-identical serialized keys instead of deserialized-object equality — non-canonical serialization can resurrect deleted entries or drop searched keys
- Issue 2: Key identity for WAL replay changed from deserialized-object equality to raw serialized-byte equality, which can resurrect deleted keys or pick a stale value when the same logical key has more than one byte representation in WAL history
- Issue 3: No guard against a negative/corrupted record length in DefaultReader.readData — a single corrupted 4-byte length aborts the whole IMap recovery with NegativeArraySizeException
- Issue 4: TOCTOU between fs.exists / fs.listFiles / fs.open — concurrent WAL compaction or file deletion causes FileNotFoundException to abort recovery; error surface also changed from IMapStorageException to raw IOException
- Issue 5: The `searchKeys` filtering behavior in `WALReader.loadAllData` has materially changed (the old filter compared `byte[]` against deserialized objects and could never match), so the "no user-facing change" claim needs correcting and the behavior change should be recorded in the PR description / release notes.
- Issue 6: loadAllKeys retains the full IMapFileData — including value byte arrays — for every unique key, so key-only recovery memory is proportional to total live value bytes, not key count
- Issue 7: loadAllData(searchKeys) silently changes behavior: the old byte[]-contains filter never matched, so selective loading previously returned nothing — now it returns data
- Issue 8: `LatestMutationAccumulator.accept` relies on the inverted `compareTo` semantics of `IMapFileData` (`< 0` means newer) without any explanatory comment or Javadoc.

DanielLeens:
Thanks for the fresh full pass, @SEZ9 — and thanks for continuing to engage on this one.

First, on the earlier Issue 1 (possible public API break on `DefaultReader.readData`): this new review doesn't carry it forward, which lines up with the evidence in my 2026-08-05 reply — `readData` is `private` on the current head `bccb2761b`, so I'll take that as resolved unless you want to keep it open explicitly.

I re-verified the current head against the eight new findings before replying:

- Issue 3 (negative/corrupted `dataLength` not guarded) checks out. `WALDataUtils.byteArrayToInt` does a signed left-shift on the top byte, so a corrupted/torn length can produce a negative `int`. The `dataLength > remainingBytes` check compares an `int` against a `long` and never catches a negative value, so `new byte[dataLength]` throws `NegativeArraySizeException` and aborts the whole recovery instead of skipping a bad tail record — exactly the kind of input WAL recovery has to tolerate.
- Issue 6 (`loadAllKeys` retaining full `IMapFileData` incl. value bytes) also checks out. `LatestMutationAccumulator.latestMutations` is `Map<SerializedKey, IMapFileData>`, and `WALReader.loadAllKeys` only reads `mutation.getKey()`/`getKeyClassName()` off it, so peak heap for a key-only listing still scales with total live value bytes rather than key count — that meaningfully undercuts this PR's own memory-optimization claim for that code path.
- Issues 1/2/5/7 (serialized-byte vs deserialized-object key identity): the code matches your read — `SerializedKey.equals` is `Arrays.equals` on raw bytes. Worth noting `IMapFileData.key` was already a raw `byte[]` on `dev` before this PR, so the byte-level storage itself isn't new; what's new is using raw-byte identity for dedup/tombstone matching in the accumulator. I agree this is a real, worth-documenting property of the new path, though I'd want to see a concrete non-canonical serializer actually in use in this codebase before calling it a live correctness bug rather than a documented assumption.
- Issue 8 (inverted `compareTo` needs a comment): confirmed. `IMapDataComparator.compare` orders the newer record first, so `mutation.compareTo(current) < 0` in the accumulator does mean "mutation is newer" — a one-line comment there is cheap insurance against a future flip.

None of these land in a "Blocking items" section this round (there isn't one), so I'm reading this as follow-ups rather than a renewed hold, and my 2026-08-04 approval stands as-is. Issues 3 and 6 look like the most valuable ones for the author to pick up given they touch correctness and the memory-optimization goal directly.

nzw921rx:
b0384c84bb3f360f05362bdaec1c208d

A great optimization once🚀

nzw921rx (approved):
+1 LGTM

DanielLeens:
Thanks @nzw921rx for taking a look and for the `+1 LGTM`!

For the record, since a couple of reviews landed close together on this thread: my own conclusion on this PR is unchanged and still stands at **APPROVED** (submitted 2026-08-04 on head `bccb2761b`). I re-verified that on 2026-08-26 after @SEZ9's latest full pass — none of that round's findings carried forward as a blocking item either. The two items worth the author's attention going forward are non-blocking follow-ups:
- Issue 3: negative/corrupted `dataLength` not guarded in WAL record parsing
- Issue 6: `loadAllKeys` still retaining full `IMapFileData` (including value bytes) rather than a key-only footprint

Neither blocks merge on the current head. So from a source-review standpoint there's no new blocker here — the remaining step is just going through the normal merge process. Thanks again for the extra eyes on this one!

nzw921rx:
benchmarks compare of results: https://github.com/apache/seatunnel/actions/runs/33771937553
benchmarks Diagnostics GC of results: https://github.com/apache/seatunnel/actions/runs/33779011143

DanielLeens:
Thanks for running the benchmarks and diagnostics, @nzw921rx — good to have the numbers documented on the thread.

One flag while I'm here: the current head (`f75ab5036`) has a fresh `Build` failure in the fork CI (`transform-v2-it-part-1 (8, ubuntu-latest)`, run finished 2026-09-03T15:08Z), which is newer than my last check-in on this thread. @dybyte, could you take a look and confirm whether that's a known flake in that IT bucket or something this change actually touches? I'm not treating it as a new blocker yet since I haven't dug into the failure log myself, but wanted to surface it before it gets buried under the benchmark discussion.

DanielLeens:
Follow-up on my comment above: the `transform-v2-it-part-1 (8, ubuntu-latest)` failure I flagged (fork run finished 2026-09-03T15:08Z) turned out to be transient. A rerun on this same, unchanged head (`f75ab5036`) started right after at 15:13Z and completed successfully at 18:44Z (https://github.com/apache/seatunnel/runs/100702174541). The `Build` check is green end-to-end now.

No action needed from @dybyte after all — my 2026-08-04 "Ready to merge" approval stands.

davidzollo (approved):
+1
LGTM 🎉

Thanks so much for this awesome contribution! This is such a fantastic addition to our project, and your incredible eye for detail really shines through!

Open‑source lives and breathes thanks to passionate contributors exactly like you — it is your dedication that keeps this whole community charging forward 🌊🚀🔥

PR #10757, opened 2026-04-13
[Fix][Zeta] Clean failed pipeline metrics without full-map cleanup scan

davidzollo:
## What problem does this PR solve?

The current Zeta metrics flow already stores running job metrics in partitioned Hazelcast `IMap` buckets, but there were still two cleanup gaps around failed pipelines.

1. `FAILED` pipelines did not enter the existing metrics cleanup lifecycle.
2. The cleanup path still scanned the full distributed metrics map before removing entries, which reintroduced the same kind of full-map fetch pressure that partitioned storage was supposed to avoid.

At runtime the cleanup chain is:

```text
SubPlan.subPlanDone()
  -> JobMaster.removeMetricsContext(...)
  -> JobMaster.enqueuePipelineCleanupIfNeeded(...)
  -> CoordinatorService.runPendingPipelineCleanupOnce()
  -> SeaTunnelServer.removeMetrics(...)
```

Before this PR:

- `FINISHED` and `CANCELED` pipelines could enter cleanup, but `FAILED` pipelines could not.
- `SeaTunnelServer.removeMetrics(...)` still iterated the whole metrics `IMap`, collected tasks to remove, and then performed a second-phase delete.

That left two concrete risks:

- failed pipeline metrics could leak in the cluster longer than intended
- cleanup itself could still perform a full-map scan and leave a race window between "find" and "delete"

This PR also keeps the two small follow-up refinements already discussed in the previous version:

- move the metrics partition hash rule into a shared common utility
- improve the `job-metrics-partition-count` validation message readability

### 1. Include `FAILED` in the existing cleanup lifecycle

This PR extends the existing cleanup predicates so failed pipelines go through the same cleanup lifecycle already used by finished and canceled pipelines:

- `JobMaster.enqueuePipelineCleanupIfNeeded(...)`
- `JobMaster.removeMetricsContext(...)`
- `CoordinatorService.shouldCleanup(...)`

This keeps the change minimal: no new protocol, no new storage structure, and no behavior change for already-cleaned end states.

### 2. Avoid full-map fetch during distributed metrics cleanup

`SeaTunnelServer.removeMetrics(...)` now cleans metrics bucket by bucket with `IMap.compute(...)` instead of:

1. scanning the full distributed metrics map
2. collecting matching task locations
3. removing them in a second pass

That change keeps cleanup aligned with the partitioned storage design:

- no full-map fetch during cleanup
- removal stays atomic within each metrics bucket
- no separate "collect then delete" race window

[...]

PR #11495, opened 2026-07-19
[Fix][Zeta] Reset metrics after pipeline recovery

### LINKED ISSUE #11472: [Bug] [Zeta] Finished job metrics are inaccurate after worker failover
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/seatunnel/issues?q=is%3Aissue) and found no similar issues.

### What happened

After a Zeta worker failover, a batch job can finish successfully and the target data can be correct, but the final job metrics returned by REST `/job-info/{jobId}` are inconsistent with the actual bounded input and target table.

I reproduced this with a deterministic Kafka topic containing exactly `200000` JSON records with `id = 1..200000`, then ran a Zeta batch job:

```text
Kafka source -> JDBC MySQL sink
```

The sink table has `id` as primary key and uses `insert ignore` so retries are idempotent. This lets us validate data loss by checking count/distinct/min/max/gaps.

No-failover baseline:

```text
jobStatus = FINISHED
SourceReceivedCount = 200000
SinkWriteCount = 200000
SinkCommittedCount = 200000
IntermediateQueueSize = 0

MySQL:
count = 200000
distinct_ids = 200000
min_id = 1
max_id = 200000
missing_in_range = 0
```

Worker failover run 1:

```text
jobStatus = FINISHED
errorMsg = null
SourceReceivedCount = 303148
SinkWriteCount = 299048
SinkCommittedCount = 200000
IntermediateQueueSize = 4098

MySQL:
count = 200000
distinct_ids = 200000
min_id = 1
max_id = 200000
missing_in_range = 0
```

Worker failover run 2:

```text
jobStatus = FINISHED
errorMsg = null
SourceReceivedCount = 239640
SinkWriteCount = 238024
SinkCommittedCount = 200000
IntermediateQueueSize = 1616

MySQL:
count = 200000
distinct_ids = 200000
min_id = 1
max_id = 200000
missing_in_range = 0
```

So the data result is correct after failover, but `SourceReceivedCount`, `SinkWriteCount`, and `IntermediateQueueSize` in the finished job metrics appear to include records from canceled/retried task attempts, or are not reset/reconciled when the replacement attempt finishes.

This makes finished job metrics misleading for reliability validation and monitoring. In particular, `IntermediateQueueSize` should not remain non-zero after a batch job is `FINISHED`.

### SeaTunnel Version

`3.0.0-SNAPSHOT`, tested on Zeta engine build commit `46d2f39`.

### SeaTunnel Config

The source Kafka topic was pre-populated with exactly `200000` JSON records:

```json
{"id":1,"name":"n1","age":21}
...
{"id":200000,"name":"n200000","age":20}
```

Kafka offsets and visible records were both verified as `200000`.

Job config:

```hocon
env {
  job.name = "zeta-kafka-deterministic-workerfail-200k"
  job.mode = "BATCH"
  parallelism = 2
  checkpoint.interval = 10000
  checkpoint.timeout = 60000
}

source {
  Kafka {
    plugin_output = "kafka_rows"
    schema = {
      fields {
        id = "int"
        name = "string"
        age = "int"
      }
      primaryKey {
        name = "pk"
        columnNames = [id]
      }
    }
    format = json
    topic = "zeta_kafka_deterministic_200k"
    bootstrap.servers = "kafka:9092"
    start_mode = "earliest"
    consumer.group = "zeta-kafka-deterministic-workerfail-200k"
    kafka.config = {
      client.id = "zeta-kafka-deterministic-workerfail-200k"
      max.poll.records = 500
      enable.auto.commit = "false"
    }
  }
}

sink {
  jdbc {
    plugin_input = "kafka_rows"
    url = "jdbc:mysql://mysql:3306/zeta_test"
    user = "zeta"
    password = "******"
    query = "insert ignore into zeta_kafka_det_workerfail_200k(id,name,age) values(?,?,?)"
  }
}
```

Failover action:

```shell
systemctl stop seatunnel-worker
# wait until the batch job finishes
systemctl start seatunnel-worker
```

### Running Command

Submitted through Zeta REST API:

```shell
curl -H 'Content-Type: text/plain' \
  --data-binary @job.conf \
  'http://<zeta-master>:8080/submit-job?format=hocon'
```

Metrics checked through:

```shell
curl 'http://<zeta-master>:8080/job-info/<jobId>'
```

### Error Exception

No exception is thrown. The job status is `FINISHED` and target data is correct.

The problem is inaccurate finished job metrics after worker failover:

```json
{
  "jobStatus": "FINISHED",
  "errorMsg": null,
  "metrics": {
    "SourceReceivedCount": "303148",
    "SinkWriteCount": "299048",
    "SinkCommittedCount": "200000",
    "IntermediateQueueSize": "4098"
  }
}
```

Zeta engine, 3 masters and 2 workers. Each worker has 8 fixed slots.

### Java or Scala Version

Java 11.

### Screenshots

N/A

### Code of Conduct

zhangshenghang:
## What does this PR do?

Resets runtime metric snapshots before a failed Zeta pipeline is restored.

## Why is this change needed?

Pipeline retries reuse stable task locations. Retaining snapshots from the failed attempt caused `SourceReceivedCount`, `SinkWriteCount`, and queue metrics to continue from stale values, so finished-job metrics could include work from canceled attempts after worker failover.

## How was it tested?

- Added `StateTransitionCleanupTest#testPipelineRestoreClearsMetricsFromTheFailedAttempt`.
- `./mvnw -pl seatunnel-engine/seatunnel-engine-server -DskipITs -Dskip.spotless=true -Dtest=StateTransitionCleanupTest test`
- `./mvnw -q -pl seatunnel-engine/seatunnel-engine-server -DskipTests verify`

Closes #11472.


DanielLeens (commented):

1. **Blocker 1** — Metrics cleanup now runs before reset/restore and can throw from distributed map operations. In a terminal pipeline state, that makes `prepareRestorePipeline()` return false and skips recovery. Please isolate cleanup failure from the restore path and add a regression where cleanup throws but restore still proceeds.

The current Build is failing; CI remains a separate gate and does not change the source verdict above. The current failed lanes include `Run / Dead links`, `Run / unit-test (11, ubuntu-latest)`, `Run / jdbc-connectors-it-part-2 (8, ubuntu-latest)`, `Run / jdbc-connectors-it-part-1 (11, ubuntu-latest)`, and 19 more failed lanes. The shared dead-link failure has a fix in #11502, but that fix has not merged into `dev` yet; syncing the current `dev` alone will not clear that lane today. After #11502 lands, please sync/retrigger and then diagnose only any lanes that still fail.

seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/dag/physical/StateTransitionCleanupTest.java
```diff
@@ -97,6 +105,57 @@ void testPhysicalPlanIgnoresLateTransitionWhenJobStateAlreadyTerminal() throws E
...
+                            runningJobStateTimestamps,
+                            Collections.emptyMap());
+            JobMaster jobMaster = mock(JobMaster.class);
+            CheckpointManager checkpointManager = mock(CheckpointManager.class);
+            PhysicalPlan physicalPlan = mock(PhysicalPlan.class);
+            when(jobMaster.getCheckpointManager()).thenReturn(checkpointManager);
+            when(jobMaster.getPhysicalPlan()).thenReturn(physicalPlan);
+            subPlan.setJobMaster(jobMaster);
+
+            java.lang.reflect.Method restore =
```
nzw921rx, line 148: please use `import java.lang.reflect.Method `

DanielLeens: +1. The current head still uses the fully qualified java.lang.reflect.Method at line 148; a normal import is clearer and keeps the test consistent with the project's import style. This is non-blocking.

DanielLeens:
Thanks for the update. I rechecked the current head `4729fe2ecf3d` against the latest `dev`. This branch is currently `diverged` (`ahead_by=1`, `behind_by=62`), and the current `Build` check is `FAILURE`: https://github.com/apache/seatunnel/runs/88202413330

This follow-up records the current CI/upstream-sync gate only; it should not be read as blaming the PR diff for the stale-head Build failure.

DanielLeens (commented):
Thanks for syncing this branch with the latest `dev` — that resolves the stale-head/CI-staleness gate I flagged in my July 30 follow-up (the branch was `ahead_by=1, behind_by=62` at the time). I re-fetched the current head (`79efa5139eda`) and confirmed the merge commit brings in no functional change of its own: `git diff origin/dev...HEAD --stat` still shows exactly the same three files (`SubPlan.java`, `JobMaster.java`, `StateTransitionCleanupTest.java`, +72/-0) as the `4729fe2ecf3d` head I reviewed on July 20. Doing a full fresh review of this head as required for re-review: **Blocker 1 from my July 20 review is still unresolved in this code** — see Issue 1 below, restated as carried-forward, not new. nzw921rx's July 22 style comment (which I +1'd on July 26) is also still unaddressed — see Issue 2.

`TaskLocation` (job/pipeline/task-group/task/index) is stable across pipeline restore attempts — it does not carry a retry/attempt counter (`seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/execution/TaskLocation.java:34-51`). Workers periodically push metrics snapshots keyed by `TaskLocation` to the master via `ReportMetricsOperation` → `SeaTunnelServer.updateMetrics()` → `HazelcastMetricsSnapshotStateStore.merge()`, which unconditionally overwrites the last snapshot per key. Because the key doesn't change across a restore, once a failed pipeline is restored, the new attempt's tasks reuse the exact same `TaskLocation` keys and would otherwise keep accumulating on top of the failed attempt's last-reported numbers. This PR clears the old pipeline's metric snapshots right before restore so the new attempt starts from a clean slate. The reasoning is sound and matches the codebase's actual metrics model.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Change**: `SubPlan.prepareRestorePipeline()` (`seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/dag/physical/SubPlan.java:489-524`) now calls a new `JobMaster.clearPipelineMetricsForRestore(pipelineLocation)` (`.../master/JobMaster.java:1201-1210`) immediately before `reset()`:

```java
private boolean prepareRestorePipeline() {
    synchronized (restoreLock) {
        try {
            pipelineRestoreNum.getAndIncrement();
            log.info(...);
            jobMaster.clearPipelineMetricsForRestore(getPipelineLocation());   // line 497, new
            reset();
            ...
            return true;
        } catch (Throwable e) {                                               // line 507
            if (this.currPipelineStatus.isEndState()) {                       // line 508
                return false;                                                  // restore aborted
            }
            ...
            return true;
        }
    }
}
```

`clearPipelineMetricsForRestore()` itself is a thin, **unguarded** delegate:

```java
public void clearPipelineMetricsForRestore(PipelineLocation pipelineLocation) {
    seaTunnelServer.removeMetrics(pipelineLocation);   // JobMaster.java:1210, no try/catch
}
```

`seaTunnelServer.removeMetrics()` → `MetricsSnapshotStateStore.removePipeline()`. The Hazelcast-backed implementation (`.../common/statestore/metrics/hazelcast/HazelcastMetricsSnapshotStateStore.java:115-133`) loops over every partition and issues a blocking `metricsImap.compute(...)` per partition — a real distributed-map RPC that can throw `HazelcastInstanceNotActiveException`, `TargetNotMemberException`, `OperationTimeoutException`, etc. during exactly the kind of cluster instability (worker failover, member departure) this PR's own branch name targets.

**This is the same defect I raised as Blocker 1 on July 20** ("Metrics cleanup now runs before reset/restore and can throw from distributed map operations. In a terminal pipeline state, that makes `prepareRestorePipeline()` return false and skips recovery."). Tracing the full call chain confirms the impact is real and severe: `prepareRestorePipeline()` is invoked from the pipeline state machine at `SubPlan.java:735-740`:

```java
case FAILED:
case CANCELED:
    if (checkNeedRestore(state) && prepareRestorePipeline()) {
        ...
        restorePipeline();
        return;
    }
    subPlanDone(state);   // pipeline treated as permanently done — no more retries
```

If `clearPipelineMetricsForRestore()` throws, `prepareRestorePipeline()` catches it at the `Throwable` level (line 507). At this point `reset()` has not run yet, so `currPipelineStatus` is still the pre-restore terminal value (`FAILED`/`CANCELED`, both `isEndState() == true`), so the method returns `false` even though `checkNeedRestore(state)` already confirmed retries remain (`job.retry.times` budget not exhausted). Control falls through to `subPlanDone(state)`, which marks the pipeline permanently failed. **A transient failure purely in metrics bookkeeping — unrelated to the health of the pipeline or its tasks — silently converts a retryable failure into a permanent one and defeats the configured retry policy**, exactly during the worker-failover scenario this PR is meant to harden.

What makes this doubly clear-cut: the same file already has the correct pattern for this exact call, right above the new method. `JobMaster.removeMetricsContext()` (`JobMaster.java:1186-1198`), which also calls `seaTunnelServer.removeMetrics(pipelineLocation)`, wraps it in `try { ... } catch (Exception e) { LOGGER.severe("failed to remove metrics", e); }`. Its caller, `SubPlan.subPlanDone()` (`SubPlan.java:307-329`), wraps that call in a *second*, outer `try/catch(Throwable e)` that only logs. That is a genuinely different call site (terminal-no-restore path, mutually exclusive with the restore path this PR touches), so there's no redundant double-clear — but it does show the author/codebase already understands "never let a metrics-cleanup failure affect pipeline lifecycle control flow" is a hard rule here. `clearPipelineMetricsForRestore()` simply doesn't follow it.

**Race with in-flight metric updates** (asked to trace specifically): worker-side metric reporting (`TaskExecutionService` → `ReportMetricsOperation` → `SeaTunnelServer.updateMetrics()` → `HazelcastMetricsSnapshotStateStore.merge()`, which does `metricsImap.compute(partition, (k, old) -> old.putAll(newSnapshot))`) runs on an independent periodic schedule, completely uncoordinated with the restore-time clear on the master. A report already in flight when `clearPipelineMetricsForRestore()` runs can land after the clear and repopulate a `TaskLocation` key with the failed attempt's last value. This is a real gap, but it's transient and self-healing: because `merge()` always overwrites by key, the new attempt's very first report for that same `TaskLocation` overwrites the stale value again. I'd rate this Medium, not a blocker — it does not corrupt persisted state and resolves itself within one reporting interval.

**Does resetting metrics hide a real failure signal** (asked to check specifically): no. Pipeline health (`FAILED`/`RUNNING`/etc.) is tracked independently via `runningJobStateIMap` / `currPipelineStatus`, not derived from the metrics snapshot map. Clearing metrics before a restore attempt cannot make a still-broken pipeline "look healthy" — the metrics for the new attempt simply start from zero and grow with the new attempt's real activity. This part of the design is sound.

## 1.2 Compatibility Impact

Fully compatible. `MetricsSnapshotStateStore`, `HazelcastMetricsSnapshotStateStore`, `PipelineLocation`, and the metrics IMap layout are all pre-existing (already in `dev` via #10757 and later commits pulled in by the merge, not introduced by this PR). No API, config, protocol, or serialization format changes.

`removePipeline()` sequentially scans every metrics-IMap partition (not parallelized, unlike `merge()`'s `parallelStream()`). This cost is pre-existing (inherited from `removeMetricsContext`'s implementation, #10757) and not introduced by this PR, but this PR does add a new call site that pays it on every restore attempt (potentially several times per job under `job.retry.times`), in addition to the terminal-state call site. Worth being aware of for jobs with high `job.retry.times` on clusters with large `partitionCount`, but not a blocker.

See Issue 1 (High) below — the core defect of this review.

# 2. Code Quality Assessment

## 2.1 Coding Standards

`clearPipelineMetricsForRestore()` has a clear Javadoc explaining the "why" (task locations are stable across retries). Good. See Issue 2 (Low, carried-forward from nzw921rx/July 22) for the one remaining style nit in the test.

`testPipelineRestoreClearsMetricsFromTheFailedAttempt` (`StateTransitionCleanupTest.java:108-157`) verifies, via a mocked `JobMaster`, that `clearPipelineMetricsForRestore()` is invoked during `prepareRestorePipeline()`. That's useful, but it only covers the happy path. **It does not cover the regression scenario I explicitly asked for on July 20** — "add a regression where cleanup throws but restore still proceeds." As shown in 1.1, that scenario currently fails (restore is aborted, not proceeded), so a correct regression test would presently fail against this code, which is exactly why it's important that it be added together with the fix. The test is otherwise well-constructed (mock-based, no `Thread.sleep`/timing dependence, no shared static state, no ordering dependence) — stability rating: **Stable**, but coverage is insufficient for the change being made.

## 2.3 Documentation Updates

Not applicable — internal engine runtime behavior, no user-facing config/API surface.

# 3. Architectural Soundness

The diagnosis and the choice of clearing metrics at restore time (rather than, say, versioning `TaskLocation`) is the right, minimal fix for the stated problem. However, as implemented it is a **temporary workaround** rather than a precise fix, because it re-adds a known failure mode (unguarded distributed-map call directly gating restore control flow) that the same class already knows how to avoid two methods away.

## 3.2 Maintainability

Good Javadoc, clear naming, small footprint. The one guarding `try/catch` needed for correctness is missing, which is exactly the kind of subtle regression future maintainers touching this method must not reintroduce — worth fixing now rather than relying on a future bug report against `job.retry.times` silently not working during member churn.

## 3.3 Extensibility

No concerns — the change is additive and localized.

No persisted/serialized state (checkpoint, savepoint, protocol) is touched; only in-memory/Hazelcast-runtime metrics state. No upgrade or restore-from-old-checkpoint compatibility impact.

# 4. Issue Summary

| 1 | `clearPipelineMetricsForRestore()` calls `seaTunnelServer.removeMetrics()` with no exception isolation; if the underlying distributed-map operation throws (plausible during worker failover/member churn), `prepareRestorePipeline()` returns `false` and a retryable pipeline failure is silently turned permanent, defeating `job.retry.times`. Carried forward from my July 20 review (Blocker 1), still unresolved in this head. | `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/master/JobMaster.java:1209-1211` (caller: `SubPlan.java:489-524`, especially lines 497 and 507-508) | High |
| 2 | Fully-qualified `java.lang.reflect.Method` used instead of a normal import in the new test. Style-only, non-blocking. Carried forward from nzw921rx (July 22), which I +1'd on July 26. Still unaddressed in this head. | `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/dag/physical/StateTransitionCleanupTest.java:148` | Low |
| 3 | New test only covers the happy path; the "cleanup throws but restore still proceeds" regression requested alongside Issue 1's fix is still missing. | `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/dag/physical/StateTransitionCleanupTest.java:108-157` | Medium |
| 4 | Restore-time metrics clear can race with an in-flight worker `ReportMetricsOperation` for the failed attempt, transiently repopulating a stale snapshot. Self-heals on the new attempt's next report; does not corrupt persisted state or job status. | `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/common/statestore/metrics/hazelcast/HazelcastMetricsSnapshotStateStore.java:63-95` | Medium |

# 5. Merge Recommendation

   - Issue 1 (High): Wrap `seaTunnelServer.removeMetrics(pipelineLocation)` in `clearPipelineMetricsForRestore()` in a `try/catch` that logs and swallows, matching the existing `removeMetricsContext()` pattern in the very same class, so a metrics-cleanup failure can never abort a legitimate pipeline restore. Option A (preferred, smallest diff): mirror `removeMetricsContext()`'s `try { ... } catch (Exception e) { LOGGER.severe(...); }` directly inside `clearPipelineMetricsForRestore()`. Option B: catch at the `SubPlan.prepareRestorePipeline()` call site around just that one line, if isolating cleanup failures from `reset()`'s own retry-with-exception semantics is preferred to be visible at the call site instead.
   - Issue 3 (Medium, tied to Issue 1): add the regression test I requested on July 20 — mock `seaTunnelServer.removeMetrics()` (or the store it delegates to) to throw, and assert `prepareRestorePipeline()` still returns `true` / restore still proceeds.

   - Issue 2 (Low): switch to a normal `import java.lang.reflect.Method;` in the test.
   - Issue 4 (Medium): no code change requested; flagging for awareness. If a future PR wants to close this gap it would need per-`TaskLocation` staleness tracking (e.g. an attempt/version tag), which is out of scope here.

Overall: the diagnosis and general direction of this PR are correct and worth merging, but the fix as currently implemented reintroduces the exact defect flagged in the last full review under a new method name — it needs the same one-line exception guard the codebase already uses next to it, plus the regression test proving it, before merge.

DanielLeens (commented):
Thanks for pushing the fix. I re-read the complete current diff from scratch against `dev` (not just the delta since my August 12 review). Head is now `2e23ba7d61` (up from `79efa5139e`), containing exactly one new content commit: `2e23ba7d61` "Keep pipeline restore resilient to metrics cleanup failures", plus a routine `dev` merge. I confirmed via `git diff origin/dev...HEAD --stat` that the full PR-owned diff is still limited to `SubPlan.java` (+1), `JobMaster.java` (+17), and `StateTransitionCleanupTest.java` (+76) — no unrelated files pulled in by the merge.

`TaskLocation` (job/pipeline/task-group/task/index) is stable across pipeline restore attempts, so workers' periodically-reported metrics snapshots keyed by `TaskLocation` would otherwise keep accumulating on top of a failed attempt's last-reported numbers once the pipeline is restored. This PR clears the old pipeline's metric snapshots right before restore so the new attempt starts from a clean slate.

# 1. Code Change Review

## 1.1 Core Logic Analysis — Blocker 1 from my last review is now resolved

**Before** (what I flagged as Blocker 1 on July 20, and restated as still-open Issue 1 on August 12):

```java
public void clearPipelineMetricsForRestore(PipelineLocation pipelineLocation) {
    seaTunnelServer.removeMetrics(pipelineLocation);
}
```

**After** (`JobMaster.java:1201-1216`):

```java
public void clearPipelineMetricsForRestore(PipelineLocation pipelineLocation) {
    try {
        seaTunnelServer.removeMetrics(pipelineLocation);
    } catch (Exception e) {
        // Metrics cleanup is best effort and must not prevent the pipeline from recovering.
        LOGGER.warning("Failed to clear metrics before restoring " + pipelineLocation, e);
    }
}
```

This is exactly remediation Option A from my last review, and it now matches the sibling method's already-accepted pattern byte-for-byte in shape (`JobMaster.removeMetricsContext()`, `JobMaster.java:1188-1198`, uses `catch (Exception e) { LOGGER.severe("failed to remove metrics", e); }`; this one uses `.warning` instead of `.severe`, a reasonable choice now that the failure is explicitly documented as best-effort and non-fatal — not a concern).

I re-traced the full call chain to confirm the fix actually closes the gap I found: `SubPlan.prepareRestorePipeline()` (`SubPlan.java:489-524`) calls `jobMaster.clearPipelineMetricsForRestore(getPipelineLocation())` immediately before `reset()`, inside a `try { ... } catch (Throwable e) { if (currPipelineStatus.isEndState()) return false; ... }`. Previously, an unguarded `HazelcastMetricsSnapshotStateStore.removePipeline()` throwing `HazelcastInstanceNotActiveException`/`TargetNotMemberException`/etc. (plausible during the exact worker-failover/member-churn scenario this PR targets) would propagate up through `clearPipelineMetricsForRestore()`, be caught by `prepareRestorePipeline()`'s outer `Throwable` catch while `currPipelineStatus` was still the pre-restore terminal value, and cause the method to return `false` — silently converting a retryable failure into a permanent one via `subPlanDone(state)`, defeating `job.retry.times`. With the new `try/catch` now sitting inside `clearPipelineMetricsForRestore()` itself, that exception can no longer escape the method at all, so `reset()` and the rest of `prepareRestorePipeline()` always run to completion regardless of whether the metrics cleanup succeeded. This is the correct fix, placed at the correct boundary.

### Runtime path (re-verified against the new head)
```text
Pipeline FAILED/CANCELED, retries remain
  -> SubPlan state machine (SubPlan.java:735-740) -> checkNeedRestore() == true
      -> prepareRestorePipeline() (SubPlan.java:489-524)
          -> jobMaster.clearPipelineMetricsForRestore(pipelineLocation)
              -> seaTunnelServer.removeMetrics(pipelineLocation)
                  -> HazelcastMetricsSnapshotStateStore.removePipeline() throws (member churn)
              -> caught locally, logged as warning, swallowed      <- new, closes the gap
          -> reset() runs unconditionally afterward
          -> returns true -> restorePipeline() proceeds normally
```

## 1.2 Compatibility Impact
Unchanged from my last review: fully compatible. No `Option`, protocol, or serialization changes; only in-memory/Hazelcast-runtime metrics state is touched.

No change from my last review's assessment.

Resolved — see 1.1. No remaining open issue at High/Medium severity.

# 2. Code Quality Assessment

## 2.1 Coding Standards
`clearPipelineMetricsForRestore()` keeps its clear Javadoc. See Issue 1 below (Low, carried forward, still unaddressed, and now duplicated).

The new test, `testMetricsCleanupFailureDoesNotPreventRestore` (`StateTransitionCleanupTest.java:161-176`), mocks `SeaTunnelServer.removeMetrics()` to throw `IllegalStateException`, injects it into a `JobMaster` built with `CALLS_REAL_METHODS`, and asserts `Assertions.assertDoesNotThrow(() -> jobMaster.clearPipelineMetricsForRestore(location))`. This is not literally the test I asked for on July 20 ("mock `seaTunnelServer.removeMetrics()` to throw, and assert `prepareRestorePipeline()` still returns `true` / restore still proceeds") — it verifies the leaf method directly rather than going through `SubPlan.prepareRestorePipeline()` end-to-end. That said, I think this is an acceptable substitute rather than a gap: the fix now lives entirely inside `clearPipelineMetricsForRestore()`, and `prepareRestorePipeline()` calls it unconditionally with no further exception-sensitive logic of its own between that call and `reset()` — so proving the leaf method never throws is logically sufficient to guarantee the caller (and `prepareRestorePipeline()`'s return value) is unaffected. See Issue 2 below for a non-blocking suggestion to still add the end-to-end assertion for documentation/defense-in-depth value.

Stability rating: **Stable**. Mock-based, no `Thread.sleep`, no shared static state, no ordering dependence.

## 2.3 Documentation Updates
Not applicable — internal engine runtime behavior only.

# 3. Architectural Soundness

With the guard now in place, this is a precise, minimal fix rather than the "temporary workaround" I described last round — it closes the exact gap that made the original design risky, without touching unrelated code.

## 3.2 Maintainability
Good. The fix now correctly mirrors the established pattern in the same class, removing the divergence that was the actual root cause of Blocker 1.

## 3.3 Extensibility
No concerns — additive and localized.

No persisted/serialized state (checkpoint, savepoint, protocol) is touched; only in-memory/Hazelcast-runtime metrics state. No upgrade or restore-from-old-checkpoint compatibility impact.

# 4. Issue Summary

|---|-------|----------|----------|------|
| 1 | Fully-qualified `java.lang.reflect.Method`/`java.lang.reflect.Field` used instead of normal imports; the new test in this round adds a second instance of the same pattern (`java.lang.reflect.Field`) alongside the original still-unfixed one. Style-only, non-blocking. | `StateTransitionCleanupTest.java:149` (original, carried forward from nzw921rx, July 22, +1'd by me July 26) and `StateTransitionCleanupTest.java:168` (new instance, this round) | Low | Yes (@nzw921rx, for the original instance) |
| 2 | New regression test (`testMetricsCleanupFailureDoesNotPreventRestore`) verifies `JobMaster.clearPipelineMetricsForRestore()` in isolation rather than asserting `SubPlan.prepareRestorePipeline()` still returns `true`/proceeds end-to-end as originally requested. Logically equivalent given where the fix lives, but an end-to-end assertion would document the actual regression scenario more directly. | `StateTransitionCleanupTest.java:161-176` | Low | No |

No open High or Medium-severity issues remain.

# 5. Merge Recommendation

   - None. Blocker 1 from my last review is resolved correctly, matching the exact remediation I proposed and the codebase's own established pattern.

   - Issue 1 (Low): switch both fully-qualified reflection references to normal imports (`import java.lang.reflect.Method;` / `import java.lang.reflect.Field;`).
   - Issue 2 (Low): optionally, also assert `prepareRestorePipeline()` returns `true` (or that `restorePipeline()` is subsequently invoked) in a scenario where `clearPipelineMetricsForRestore()` throws, for a more direct regression test of the originally-reported failure mode.

**CI status:** the apache-side `Build` check is currently `pending`; the fork run for this exact head (`https://github.com/zhangshenghang/seatunnel/actions/runs/34333143815`) is still `queued` as of this review and has not started. I can't independently confirm green yet, so per our gating rule this stays a `--comment` rather than an approval. Once the fork's `Build` run completes and is green, and given there is no remaining source-level blocker, this is ready for a maintainer's final approval and merge.

Thanks again for tracking this down carefully and matching the fix to the pattern already established in the same class — nice work.

DanielLeens (commented):
CI-fact update: in my last review the fork's `Build` run was still `queued` and had not started, and I had no source-level blocker at that point. I just rechecked `gh pr checks` on this same head and `Build` has since completed with `pass`. With CI now green and no outstanding source-side blocker from my side, this looks ready pending a maintainer's final review and merge action.

DanielLeens (approved):
This is a follow-up to my September 9 review, where the only remaining item was CI: `Build` was still `queued` on this exact head (`2e23ba7d61`) and had not started, and I found no source-level blocker at that point (Blockers: None — Blocker 1 from my earlier rounds was already resolved).

I rechecked the current head and the apache-side `Build` check has since completed with **SUCCESS**; no new commits have landed since my last review.

With CI now green and no outstanding source-side blocker, I'm updating my conclusion to **Approved**. The metrics-cleanup fix in `JobMaster.clearPipelineMetricsForRestore()` correctly closes the gap I originally flagged (an unguarded `removeMetrics()` failure could previously turn a retryable pipeline-restore failure into a permanent one), and the new regression test (`testMetricsCleanupFailureDoesNotPreventRestore`) verifies it directly. The two Low-severity suggestions from my last review (switching the fully-qualified reflection references to normal imports, and optionally adding an end-to-end assertion through `prepareRestorePipeline()`) remain non-blocking nice-to-haves.

This is ready for a maintainer's final merge action.

PR #11502, opened 2026-07-20
[Docs] Escape Tablestore JDBC URL to avoid dead-link false positives

det101:
## Summary
- Escape the Tablestore JDBC example URL in English/Chinese JDBC source docs with HTML entities (same approach as Snowflake), so `markdown-link-check` does not treat the embedded `https://` endpoint as a public web link.
- Follow-up hardening after #11459, which already replaced the concrete `myinstance` endpoint with placeholders.

## Test plan
- [x] `markdown-link-check -c .dlc.json` on `docs/en/connectors/source/Jdbc.md`
- [x] `markdown-link-check -c .dlc.json` on `docs/zh/connectors/source/Jdbc.md`
- [ ] Confirm dead-link CI job passes on this PR


DanielLeens (commented):



det101:
I noticed this has already been fixed in #11459 (by @nzw921rx), which replaced the concrete Tablestore JDBC example endpoint with a placeholder so the dead-link check no longer treats it as a public URL.

Closing this PR accordingly. Thanks!

PR #11504, opened 2026-07-20
[Improve][Connector-V2] Validate Milvus sink batch size

ZYZ666-RGB:

Related to #11007.

This PR improves declarative option validation for the Milvus sink:

- Validates that `batch_size` is greater than zero through `optionRule()` and `Conditions.greaterThan(...)`.
- Adds positive, zero, and negative validation tests.
- Updates the English and Chinese Milvus sink documentation.

Runtime-only behavior remains unchanged. In particular, `rate_limit <= 0` continues to disable collection rate limiting.

Yes. Previously, a non-positive Milvus sink `batch_size` could pass option validation. It is now rejected during configuration validation.

The focused `MilvusSinkFactoryTest` suite passes all 7 tests, including positive, zero, and negative batch-size cases.

The connector module was also verified successfully:

`mvn -f seatunnel-connectors-v2/connector-milvus/pom.xml -DskipTests verify`

### Check list

- [x] No new Jar binary package is added.
- [x] The English and Chinese Milvus sink documentation is updated.
- [x] No incompatible change documentation is required.
- [x] No connector registration or E2E changes are required because this updates an existing connector.


DanielLeens (commented):
Thanks for the contribution! I reviewed the complete current-head diff and the actual configuration-to-writer path. The declarative validation is a good direction, but there is one compatibility item to resolve before merge.

**Blocker — `batch_size=0` currently has an observable, accepted behavior.** `MilvusSinkWriter.write()` adds a row and then calls `needFlush()`, whose condition is `writeCache >= batchSize`; with `0`, the writer flushes every row. The new `greaterThan(BATCH_SIZE, 0)` rule changes existing `batch_size=0` jobs from immediate flush to a startup validation failure, but this PR does not include the required incompatible-change entry or migration guidance. The umbrella validation issue also uses `batchSize >= 0` as its numeric-range example.

The lowest-cost compatible option is to use `greaterOrEqual(..., 0)` and reject only negative values. If removing the zero-value behavior is intentional, please instead add the EN/ZH incompatible-change record and a clear `0 -> 1` migration note. The exact-head `Build` is still queued, and CI remains a separate gate.

nzw921rx (approved):
+1 good job

DanielLeens (commented):

The previous source concern is resolved on `39f91ec4b933`: the validation now preserves the historical `batch_size = 0` per-row flush behavior and covers `0` as valid and `-1` as invalid.

davidzollo (approved):
+1
LGTM

PR #10802, opened 2026-04-21
[Docs][Connector-V2] STIP-23 Phase 3: Add timer flush feature flag to all sink connector docs

nzw921rx:
### Purpose of this pull request

Closes: https://github.com/apache/seatunnel/issues/10717

> **⚠️ Depends on Phase 2 — do not merge until [#10800 ](https://github.com/apache/seatunnel/pull/10801) is merged.**

Phase 3 documentation sweep for STIP-23 (Engine-Level Timer Flush for Sink Connectors). This PR adds the `timer flush` feature entry to all sink connector docs so users can see at a glance which connectors support engine-level timer flush.

**Changes**:

- **`connector-v2-features.md` (en + zh)**: new `### timer flush` section under Sink Connector Features — explains wall-clock-driven flush, opt-in via `registerFlushAction()`, and the 2PC exclusion rule.
- **All sink connector docs (en + zh)**: added `timer flush` feature flag to every `## Key Features` / `## 主要特性` section:
  - `[x]` — connectors that adopted timer flush in Phase 2: **JDBC**
  - `[ ]` — all other connectors (unimplemented, opt-in available)

### Does this PR introduce _any_ user-facing change?

Documentation only. No code changes.

### How was this patch tested?

Docs-only change. Verified that JDBC and Iceberg carry `[x]` and all remaining connectors carry `[ ]`.

PR #10801, opened 2026-04-21
[Feature][Connector-V2] STIP-23 Phase 2: JDBC connector adoption for engine timer flush

nzw921rx:
### Purpose of this pull request

Related: https://github.com/apache/seatunnel/issues/10717

> **⚠️ Depends on Phase 1 — do not merge until [#10800](https://github.com/apache/seatunnel/pull/10800) is merged.**

Phase 2 connector adoption for STIP-23 (Engine-Level Timer Flush for Sink Connectors). This PR wires up the `registerFlushAction` API introduced in Phase 1 into the JDBC sink writers.

**Changes**:

- **JDBC (non-XA)**: `JdbcSinkWriter` now accepts `SinkWriter.Context` and registers `timerFlush()` in the constructor. `timerFlush()` calls `outputFormat.checkFlushException()` then `outputFormat.flush()`, matching the fail-fast semantics of `prepareCommit()`. Error is logged before propagation.
- **JDBC (XA / `is_exactly_once=true`)**: `JdbcExactlyOnceSinkWriter` does not register a flush action — flush is driven by `prepareCommit()`. 
- **`JdbcSink.createWriter()`**: passes `context` to `JdbcSinkWriter`; the Sink factory no longer contains any timer flush logic.

### Does this PR introduce _any_ user-facing change?

No impact on existing jobs.

- Connectors that do not call `registerFlushAction()` are unaffected — `FlushSignal` arrives at `SinkFlowLifeCycle` and is silently ignored.
- The `sink.flush.interval` config (introduced in Phase 1, default `0` = disabled) must be set for timer flush to fire.

### How was this patch tested?

Unit tests from Phase 1 cover the engine signal path (`SinkWriterContextTest`, `SinkFlowLifeCycleTest`). E2E tests covering JDBC non-XA idle-partition flush timer flush will be added in Phase 3.

PR #11505, opened 2026-07-20
[Feature][Connector-V2] Support timer flush for Elasticsearch sink

nzw921rx:

Add Zeta timer-flush support to the Elasticsearch sink, following the sink flush-action mechanism introduced in #10801 and the connector integration pattern from #10802.

The writer registers its existing synchronized bulk flush operation through `SinkWriter.Context.registerFlushAction`, allowing buffered documents to become visible according to the environment-level `sink.flush.interval` without waiting for the batch threshold or a checkpoint.

This initial implementation provides at-least-once delivery only. It does not provide 2PC exactly-once semantics, and retries can produce duplicate documents when deterministic document IDs are not configured.

Yes.

- Elasticsearch sink jobs running on Zeta can use `sink.flush.interval` to flush buffered documents periodically.
- The English and Chinese connector documentation now describes Zeta-only timer flush behavior and the at-least-once limitation.

- Added a Zeta-only E2E scenario that uses a batch threshold and checkpoint interval above the test workload, then verifies documents become visible while the job is still running.
- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-elasticsearch -DskipIT -DskipUT=false test` — 60 tests passed.
- Elasticsearch connector E2E module `test-compile` completed successfully.

The full container E2E was not executed locally because Maven `verify` was explicitly excluded from the validation run.

### Check list

* [x] A connector E2E testcase is included.
* [x] No plugin mapping, distribution dependency, CI label, or plugin configuration change is required for this existing connector.

### DIFF
--- a/docs/en/connectors/sink/Elasticsearch.md
+++ b/docs/en/connectors/sink/Elasticsearch.md
@@ -11,7 +11,7 @@ Output data to `Elasticsearch`.
 - [ ] [exactly-once](../../introduction/concepts/connector-v2-features.md)
 - [x] [cdc](../../introduction/concepts/connector-v2-features.md)
 - [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)
-- [ ] [timer flush](../../introduction/concepts/connector-v2-features.md)
+- [x] [timer flush](../../introduction/concepts/connector-v2-features.md)
 
 :::tip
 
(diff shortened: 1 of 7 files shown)


dybyte (approved):
+1

PR #11506, opened 2026-07-20
[Feature][Connector-V2] Support timer flush for Doris sink

nzw921rx:

Add Zeta timer-flush support to the Doris sink, following the sink flush-action mechanism introduced in #10801 and the connector integration pattern from #10802.

For non-2PC Stream Load, the registered action finishes the current load and immediately starts the next load with a new label. Registration is explicitly disabled when `sink.enable-2pc=true`, because an independent timer flush would break the transaction-to-checkpoint boundary required by 2PC exactly-once delivery.

This initial implementation therefore supports timer flush only in non-2PC mode and provides at-least-once delivery, not 2PC exactly-once semantics.

Yes.

- Doris sink jobs running on Zeta can use `sink.flush.interval` when `sink.enable-2pc=false`.
- Timer flush remains disabled in 2PC mode.
- The English and Chinese connector documentation now describes the behavior and delivery limitation.

- Added unit coverage that executes the registered non-2PC flush action and verifies a new load starts afterward.
- Added a negative unit test proving no action is registered in 2PC mode.
- Added a Zeta-only E2E scenario that uses a batch threshold and checkpoint interval above the test workload, then verifies rows become visible while the job is still running.
- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-doris -DskipIT -DskipUT=false test` — 115 tests passed.
- Doris connector E2E module `test-compile` completed successfully.

The full container E2E was not executed locally because Maven `verify` was explicitly excluded from the validation run.

### Check list

* [x] A connector E2E testcase is included.
* [x] No plugin mapping, distribution dependency, CI label, or plugin configuration change is required for this existing connector.

### DIFF
--- a/docs/en/connectors/sink/Doris.md
+++ b/docs/en/connectors/sink/Doris.md
@@ -21,7 +21,7 @@ import ChangeLog from '../changelog/connector-doris.md';
 - [x] [exactly-once](../../introduction/concepts/connector-v2-features.md)
 - [x] [cdc](../../introduction/concepts/connector-v2-features.md)
 - [x] [support multiple table write](../../introduction/concepts/connector-v2-features.md)
-- [ ] [timer flush](../../introduction/concepts/connector-v2-features.md)
+- [x] [timer flush](../../introduction/concepts/connector-v2-features.md)
 
 ## Description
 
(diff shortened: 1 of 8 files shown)


DanielLeens (commented):

The required Build is failing only at `Run / Code style`: Spotless reports `DorisTimerFlushIT.java`. Please apply Spotless to that file and rerun the check.

DanielLeens (commented):
Thanks for the quick update. I re-reviewed the complete new head `3f405b957b7c`, including the full Doris writer/E2E diff and the deletion ledger.

The new commit fixes the exact Spotless violation reported by the previous Build. The runtime conclusion remains source-clear: non-2PC writers register the Zeta flush action, the action executes on the task path as `checkLoadException -> flush -> startLoad`, and 2PC writers keep the checkpoint transaction boundary by not registering it. The EN/ZH docs state the Zeta-only and at-least-once limits consistently, and the bounded-future E2E does not introduce a new high-risk timing pattern.

The required Build for this new head is currently in progress, so CI remains the merge gate. A write-capable maintainer must also complete the repository approval step because Daniel has READ permission.

Queue note: this head is `diverged` from `dev` (ahead_by=4, behind_by=4). I am recording that fact, but it is not by itself a request to sync while the current check is already running.

davidzollo (commented):
+1
LGTM

davidzollo (approved):

PR #11509, opened 2026-07-21
[Fix][CLI] Fix Bedrock compatibility: temperature rejection and engine template variables

### LINKED ISSUE #11508: [Bug] [CLI] Bedrock temperature rejection on newer Claude models; engine template variables misreported as unresolved env vars
### Search before asking

### What happened

Two cases where the SeaTunnel CLI (`seatunnel-cli`) fails on or rejects valid input, found while running an accuracy benchmark of the CLI across 7 LLM providers/models:

**1. Bedrock: newer Claude models reject the `temperature` inference parameter**

Newer Claude models on AWS Bedrock (e.g. `us.anthropic.claude-sonnet-5`) reject the `temperature` inference parameter with:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when
calling the ConverseStream operation: The model returned the following errors:
`temperature` is deprecated for this model.
```

`BedrockProvider.chat` / `chat_stream` always send `temperature`, so **every CLI call fails** for these models — the CLI is unusable with them.

**2. `validate_hocon` misreports SeaTunnel engine template variables as unresolved environment variables**

The local validator treats every `${xxx}` placeholder as an environment variable. But SeaTunnel's file sinks support engine-resolved template variables `${now}`, `${uuid}`, `${transactionId}` in `file_name_expression` (documented in [LocalFile sink docs](https://github.com/apache/seatunnel/blob/dev/docs/en/connectors/sink/LocalFile.md)). A valid config such as:

```hocon
sink {
  LocalFile {
    path = "/data/out"
    custom_filename = true
    file_name_expression = "${now}"
    ...
  }
}
```

is rejected with `ERROR: Unresolved environment variables: now. Set them before running: export now=<value>`, and the CLI's auto-fix loop then "fixes" a config that was never broken.

### SeaTunnel Version

dev (seatunnel-cli @ 59ada4ec0)

### SeaTunnel Config

See above.

### Running Command

```
ANTHROPIC_MODEL=us.anthropic.claude-sonnet-5 AI_PROVIDER=bedrock seatunnel "generate any pipeline"   # case 1
# case 2: generate a LocalFile sink config with file_name_expression = "${now}", then /check
```

### Error Exception

See above.

_No response_

### Java or Scala Version

Python 3.10+ (seatunnel-cli)

### Screenshots

_No response_

### Code of Conduct

SEZ9:

Fixes #11508 — two cases where the CLI fails on or rejects valid input:

1. **Bedrock `temperature` rejection**: newer Claude models on Bedrock (e.g. `us.anthropic.claude-sonnet-5`) reject the `temperature` inference parameter with a `ValidationException` ("`temperature` is deprecated for this model"), so every CLI call fails. `BedrockProvider.chat`/`chat_stream` now catch this specific rejection, retry once without `temperature`, and cache the model id so subsequent calls skip the failed attempt. Behavior for models that accept `temperature` is unchanged.

2. **Engine template variables misreported**: `validate_hocon` treated every `${xxx}` placeholder as an environment variable, so the engine-resolved file sink template variables `${now}`, `${uuid}`, `${transactionId}` (see `docs/en/connectors/sink/LocalFile.md`, `file_name_expression`) were reported as "Unresolved environment variables" and valid configs were rejected — which also caused the auto-fix loop to "repair" configs that were never broken. These three engine template variables are now excluded from the env-var check.

No behavior change for previously-working setups. Two previously-failing valid cases now work:
- The CLI is now usable with Bedrock models that reject `temperature`.
- Configs using `${now}`/`${uuid}`/`${transactionId}` in file sink `file_name_expression` now pass local validation instead of being rejected.

- Existing test suite passes (29 tests, `python3 -m pytest tests/`).
- Fix 1 verified against the real Bedrock model `us.anthropic.claude-sonnet-5`: first call gets rejected, retry without temperature succeeds (both `chat` and `chat_stream` paths), model id cached so the second call skips the retry.
- Fix 2 verified by validating a LocalFile sink config with `file_name_expression = "${now}"`: previously `INVALID` (unresolved env var), now passes; a genuinely-unresolved `${MY_VAR}` is still reported as before.
- Both issues were found (and the fixes exercised end-to-end) by a 100-task accuracy benchmark run of the CLI across 7 Bedrock models.

### DIFF
--- a/seatunnel-cli/README.md
+++ b/seatunnel-cli/README.md
@@ -105,6 +105,12 @@ export ANTHROPIC_SMALL_FAST_MODEL='us.anthropic.claude-haiku-4-5-20251001-v1:0'
 The Bedrock provider preserves Claude `reasoningContent` blocks in streamed
 Converse responses when Bedrock returns them.
 
+> **Note:** Some newer Claude models on Bedrock reject the `temperature`
+> inference parameter. When a model returns this rejection, the provider
+> automatically retries the request without `temperature` and remembers the
+> model for the rest of the session, so subsequent calls skip the parameter.
+> For these models any configured temperature value is not applied.
+
 #### Option B: Anthropic API
 
 ```bash
@@ -394,6 +400,19 @@ Memory is stored locally at `.data/memory.json` (co-located with the CLI package
 | **Phase 3** | REST API | Optional validation via running SeaTunnel cluster |
 | **Auto-fix** | LLM-powered | Up to 3 rounds of automatic error correction during generation |
 | **Auto-repair** | LLM-powered | Automatic diagnosis and config patching on `/check` or `/run` failure |
+
+Local validation flags unresolved `${VAR}` placeholders as missing environment
+variables, with a field-aware exemption for SeaTunnel's engine-resolved file
+sink template placeholders (see the
+[LocalFile sink docs](https://seatunnel.apache.org/docs/connectors/sink/LocalFile)):
+
+| Field | Engine placeholders allowed |
+|-------|-----------------------------|
+| `file_name_expression` | `${now}`, `${uuid}`, `${transactionId}` |
+| `partition_dir_expression` | `${k0}`, `${v0}`, `${k1}`, `${v1}`, ... |
+
+The same names used in any other field (URLs, credentials, paths) are treated
+as regular environment variables and reported if unset.
 
 ## Connector Metadata
 
--- a/seatunnel-cli/seatunnel_cli/agents.py
+++ b/seatunnel-cli/seatunnel_cli/agents.py
@@ -526,12 +526,31 @@ def validate_hocon(config_str: str) -> str:
 
     # Check for unresolved ${ENV_VAR} placeholders — these will be passed as literal
     # strings to connectors at runtime, causing authentication/connection failures.
+    # SeaTunnel's engine resolves certain placeholders itself, but only in
+    # specific file sink fields (see docs/en/connectors/sink/LocalFile.md):
+    #   file_name_expression     -> ${now}, ${uuid}, ${transactionId}
+    #   partition_dir_expression -> ${k0}=${v0}/... (kN = partition field,
+    #                               vN = partition value)
+    # The exemption is field-aware so that the same names used in unrelated
+    # fields (URLs, credentials, ...) are still diagnosed as env vars.
+    engine_template_fields = {
+        "file_name_expression": re.compile(r"now|uuid|transactionId"),
+        "partition_dir_expression": re.compile(r"[kv]\d+"),
+    }
     env_var_pattern = re.compile(r'\$\{([A-Za-z_][A-Za-z0-9_]*)\}')
+    # HOCON accepts both `key = value` and `key : value` separators
+    field_pattern = re.compile(r'^\s*"?([\w.]+)"?\s*[=:]')
     unresolved_vars = set()
-    for m in env_var_pattern.finditer(config_str):
-        var_name = m.group(1)
-        if not os.environ.get(var_name):
-            unresolved_vars.add(var_name)
+    for line in config_str.splitlines():
+        field_match = field_pattern.match(line)
+        allowed = engine_template_fields.get(
+            field_match.group(1)) if field_match else None
+        for m in env_var_pattern.finditer(line):
+            var_name = m.group(1)
+            if allowed and allowed.fullmatch(var_name):
+                continue
+            if not os.environ.get(var_name):
+                unresolved_vars.add(var_name)
     if unresolved_vars:
         var_list = ", ".join(sorted(unresolved_vars))
         errors.append(
--- a/seatunnel-cli/seatunnel_cli/llm_provider.py
+++ b/seatunnel-cli/seatunnel_cli/llm_provider.py
@@ -29,6 +29,11 @@ Selection: set AI_PROVIDER env var, or run --init for interactive setup.
 """
 
 import abc
+
+try:
+    import botocore.exceptions
+except ImportError:  # boto3/botocore optional (bedrock extra)
+    botocore = None
 import json
 import logging
 import os
@@ -330,6 +335,43 @@ class BedrockProvider(LLMProvider):
         if endpoint_url:
             client_kwargs["endpoint_url"] = endpoint_url
         self.client = session.client(**client_kwargs)
+        # Models that rejected `temperature` (newer Claude models deprecate it).
+        # Populated on first rejection so subsequent calls skip the retry.
+        self._no_temperature_models: set[str] = set()
+
+    def _build_kwargs(self, model_id, messages, system, temperature, max_tokens, tools) -> dict:
+        inference_config = {"maxTokens": max_tokens}
+        if model_id not in self._no_temperature_models:
+            inference_config["temperature"] = temperature
+        kwargs = {
+            "modelId": model_id,
+            "messages": messages,
+            "inferenceConfig": inference_config,
+        }
+        if system:
+            kwargs["system"] = [{"text": system}]
+        if tools:
+            kwargs["toolConfig"] = {"tools": tools}
+        return kwargs
+
+    @staticmethod
+    def _is_temperature_rejection(error: Exception) -> bool:
+        """True if this is Bedrock's structured rejection of `temperature`.
+
+        Checks the botocore ClientError code (precise) and only uses the
+        message to discriminate that `temperature` is the rejected parameter
+        (tolerant to wording changes such as "deprecated" vs "not supported").
+        """
+        try:
+            from botocore.exceptions import ClientError
+        except ImportError:
+            return False
+        if not isinstance(error, ClientError):
+            return False
+        code = error.response.get("Error", {}).get("Code", "")
+        if code != "ValidationException":
+            return False
+        return "temperature" in str(error).lower()
 
     @property
     def provider_name(self) -> str:
@@ -353,18 +395,15 @@ class BedrockProvider(LLMProvider):
         tools: list[dict] | None = None,
     ) -> dict:
         model_id = model or self._model_id
-        kwargs = {
-            "modelId": model_id,
-            "messages": messages,
-            "inferenceConfig": {"temperature": temperature, "maxTokens": max_tokens},
-        }
-        if system:
-            kwargs["system"] = [{"text": system}]
-        if tools:
-            kwargs["toolConfig"] = {"tools": tools}
-
-        response = self.client.converse(**kwargs)
-        return response
+        kwargs = self._build_kwargs(model_id, messages, system, temperature, max_tokens, tools)
+        try:
+            return self.client.converse(**kwargs)
+        except botocore.exceptions.ClientError as e:
+            if self._is_temperature_rejection(e):
+                self._no_temperature_models.add(model_id)
+                kwargs["inferenceConfig"].pop("temperature", None)
+                return self.client.converse(**kwargs)
+            raise
 
     def chat_stream(
         self,
@@ -376,17 +415,16 @@ class BedrockProvider(LLMProvider):
         tools: list[dict] | None = None,
     ) -> Generator[dict, None, None]:
         model_id = model or self._model_id
-        kwargs = {
-            "modelId": model_id,
-            "messages": messages,
-            "inferenceConfig": {"temperature": temperature, "maxTokens": max_tokens},
-        }
-        if system:
-            kwargs["system"] = [{"text": system}]
-        if tools:
-            kwargs["toolConfig"] = {"tools": tools}
-
-        response = self.client.converse_stream(**kwargs)
+        kwargs = self._build_kwargs(model_id, messages, system, temperature, max_tokens, tools)
+        try:
+            response = self.client.converse_stream(**kwargs)
+        except botocore.exceptions.ClientError as e:
+            if self._is_temperature_rejection(e):
+                self._no_temperature_models.add(model_id)
+                kwargs["inferenceConfig"].pop("temperature", None)
+                response = self.client.converse_stream(**kwargs)
+            else:
+                raise
         current_tool_id = None
         for event in response.get("stream", []):
             if "contentBlockStart" in event:
--- a/seatunnel-cli/tests/test_llm_provider_bedrock.py
+++ b/seatunnel-cli/tests/test_llm_provider_bedrock.py
@@ -1,0 +1,177 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Regression tests for the BedrockProvider temperature-rejection retry
+(chat and chat_stream paths)."""
+
+import copy
+from unittest import mock
+
+import pytest
+
+botocore = pytest.importorskip("botocore")
+from botocore.exceptions import ClientError  # noqa: E402
+
+OK_RESPONSE = {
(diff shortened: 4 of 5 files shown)


nielifeng (commented):
Thanks for the fix. I found one remaining issue in `validate_hocon`.

The engine-template exemption is currently global, which makes the check both under- and over-inclusive:

1. Valid file sink `partition_dir_expression` placeholders such as `${k0}` and `${v0}` are still reported as unresolved environment variables. They are resolved by `AbstractWriteStrategy.generatorPartitionDir`.
2. `${now}`, `${uuid}`, and `${transactionId}` are ignored everywhere, so using one of those names as a real environment variable in an unrelated field such as a URL or credential is no longer diagnosed.

Could you make this validation field-aware? Please allow `now`/`uuid`/`transactionId` only in file sink `file_name_expression`, and `k[0-9]+`/`v[0-9]+` only in `partition_dir_expression`.

Please also add regression tests for both accepted engine placeholders and rejected unresolved placeholders in unrelated fields. The current PR does not add tests for either this path or the Bedrock retry behavior.

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current-head diff and the affected runtime path from scratch. The direction is useful, but the following items still block merge:

1. **review_4742691617: `validate_hocon` exempts engine placeholders at the wrong scope** (`seatunnel-cli/seatunnel_cli/agents.py:529-537`) — +1 to @nielifeng's review 4742691617: the global exemption still rejects valid `kN`/`vN` placeholders in `partition_dir_expression`, while it silently accepts `now`/`uuid`/`transactionId` in unrelated URL or credential fields. Please make the allowed placeholder set field-aware and add accepted/rejected regression tests.

The current Build is `SUCCESS`. Please rerun the required checks after the source changes.

Thanks again for working through this — happy to re-review the next head.

Queue note: this head is `diverged` against `dev` (ahead_by=1, behind_by=30). I am recording that fact separately; it is not by itself a request to sync unless the CI guidance above explicitly says so.

SEZ9:
Thanks @nielifeng @DanielLeens for the careful review — both points addressed in the new head.

**Field-aware placeholder validation** (`validate_hocon`): the global exemption is replaced with a per-field allowlist:

- `file_name_expression` → `${now}`, `${uuid}`, `${transactionId}`
- `partition_dir_expression` → `${kN}`, `${vN}` (resolved by `AbstractWriteStrategy.generatorPartitionDir`)

The same names in any other field (URLs, credentials, paths, ...) are diagnosed as unresolved environment variables again, and non-engine placeholders inside the two expression fields are still checked as env vars.

**Regression tests added** (`tests/test_validate_hocon_placeholders.py`, 10 cases):
- accepted: engine placeholders in their own fields (incl. mixed `out_${uuid}_${now}`, `${k0}=${v0}/${k1}=${v1}`)
- rejected: `${now}`/`${uuid}`/`${transactionId}`/`${k0}` in url/password/topic/path fields; unset env var mixed into `file_name_expression`
- Bedrock retry: first call sends temperature and gets rejected → retry without temperature succeeds → model id cached so the next call skips temperature entirely; unrelated exceptions (e.g. throttling) are re-raised untouched without triggering the retry

Full suite passes (39 tests). The head is also rebased onto current `dev` (was behind_by=30), so the diverged state noted in the queue is resolved.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `c552a23b500b` from scratch.

# What this PR fixes
- User pain: The CLI is unusable with newer Bedrock Claude models that reject `temperature` (every converse call fails with a ValidationException), and `validate_hocon` misreports the engine-resolved file sink placeholders `${now}`/`${uuid}`/`${transactionId}` (and `${kN}=${vN}`) as unresolved environment variables, rejecting valid configs and sending the auto-fix loop chasing non-bugs.
- Fix approach: `BedrockProvider` catches the temperature-deprecation rejection in both `chat` and `chat_stream`, retries once without `temperature`, and caches the model id in `_no_temperature_models` so later calls skip the failed attempt; `validate_hocon` becomes field-aware, exempting `file_name_expression` (`now|uuid|transactionId`) and `partition_dir_expression` (`[kv]\d+`) placeholders only on lines whose key matches those fields, plus regression tests for both paths.
- One-line summary: The two fixes are correct and well-tested for the common paths, but the rejection detection relies on brittle string matching rather than the botocore error code, the `key : value` HOCON syntax bypasses the field exemption, the `chat_stream` retry lacks test coverage, and the user-facing behavior change is undocumented — I'd like these addressed before merge.

# Runtime chain I rechecked
```text
agents auto-fix / validation loop
  validate_hocon(config_str)  seatunnel_cli/agents.py:~526
    field_pattern.match(line) -> engine_template_fields.get(field)  agents.py:~536-549
    env_var_pattern.finditer(line) -> allowed.fullmatch(var) skip, else os.environ.get(var) -> unresolved_vars  agents.py:~550-556
    unresolved_vars -> errors.append("Unresolved environment variables...")  agents.py:~557

LLM call path (Bedrock)
  BedrockProvider.chat(...)  seatunnel_cli/llm_provider.py:~370
    self._build_kwargs(model_id, ...)  llm_provider.py:~337 (omits temperature if model_id in self._no_temperature_models)
    self.client.converse(**kwargs)  llm_provider.py:~382
    on Exception -> BedrockProvider._is_temperature_rejection(e)  llm_provider.py:~352 (string match on type name + message for 'ValidationException'/'temperature'/'deprecated')
      -> self._no_temperature_models.add(model_id); kwargs['inferenceConfig'].pop('temperature'); self.client.converse(**kwargs) retry  llm_provider.py:~385-388
  BedrockProvider.chat_stream(...)  llm_provider.py:~390
    same _build_kwargs -> self.client.converse_stream(**kwargs) -> identical rejection/retry branch  llm_provider.py:~401-408
    then iterates response.get('stream', []) as before
```

# Findings
**Issue 1: Bedrock temperature-rejection detection relies on fragile string matching instead of the botocore error code**
- Location: `seatunnel-cli/seatunnel_cli/llm_provider.py:355`
- Why it matters: - Root cause: `_is_temperature_rejection` concatenates `type(error).__name__ + msg` and does substring checks for "ValidationException", "temperature" and "deprecated". boto3 surfaces service errors as `botocore.exceptions.ClientError` with a structured `e.response["Error"]["Code"]`; matching the class name against a concatenated string means any exception whose *message* merely contains "ValidationException" (e.g. an error wrapped/re-raised with the original text) would also trigger the retry.
- Impact: a false positive silently drops the user's `temperature` setting and permanently caches the model id in `_no_temperature_models`, so all subsequent calls in the CLI session run without temperature even for models that support it — a hard-to-diagnose behavior drift. Conversely, if Bedrock ever rewords the message (e.g. "no longer supported" instead of "deprecated"), the retry stops working and we are back to issue #11508.
- Fix: catch `botocore.exceptions.ClientError` explicitly, check `e.response["Error"]["Code"] == "ValidationException"`, and keep the `"temperature"` message check as the discriminator. That keeps the check precise on the code and tolerant on wording.
- Evidence: `seatunnel-cli/seatunnel_cli/llm_provider.py:356  and "temperature" in msg.lower() and "deprecated" in msg.lower()`; `seatunnel-cli/seatunnel_cli/llm_provider.py:383  except Exception as e:`
- Suggested fix: Concatenating `type(error).__name__ + msg` and substring-matching is fragile: any exception whose message happens to contain "ValidationException" + "temperature" + "deprecated" triggers the retry and permanently caches the model in `_no_temperature_models`. boto3 raises `botocore.exceptions.ClientError` with a structured error code — please check `e.response["Error"]["Code"] == "ValidationException"` (falling back to the message check only for the temperature discriminator). That also lets you narrow the `except Exception` in `chat`/`chat_stream` to `ClientError`.
- Severity: Medium

**Issue 2: User-facing validation and Bedrock behavior changes are undocumented, and the PR description understates what the code does**
- Location: `seatunnel-cli/seatunnel_cli/agents.py:536`
- Why it matters: - Root cause: the diff changes two user-visible behaviors — `validate_hocon` now exempts engine template placeholders in a field-aware way (including `partition_dir_expression` `${kN}`/`${vN}`, not just the three variables mentioned in the PR description), and the Bedrock provider now silently retries without `temperature` and caches that per model — but no CLI documentation, changelog entry, or release note is updated in this PR.
- Impact: users who deliberately set `temperature` for a Bedrock Claude model get it silently dropped with no documented explanation, and contributors reading the CLI docs won't know which placeholders are exempt in which fields; reviewers relying on the PR description will also miss that `partition_dir_expression` handling was added.
- Fix: update the seatunnel-cli README/docs to
- (a) list the exempted fields and their placeholder patterns (mirroring `docs/en/connectors/sink/LocalFile.md`), and
- (b) note the Bedrock temperature fallback behavior; also align the PR description with the actual implementation before merge.
- Evidence: `seatunnel-cli/seatunnel_cli/agents.py:537  "file_name_expression": re.compile(r"now|uuid|transactionId"),`; `seatunnel-cli/seatunnel_cli/agents.py:538  "partition_dir_expression": re.compile(r"[kv]\d+"),`
- Suggested fix: This is a user-facing behavior change (and it goes further than the PR description, which only mentions `${now}`/`${uuid}`/`${transactionId}` — `partition_dir_expression` `${kN}`/`${vN}` is also exempted here). Please document the exempted fields/placeholders in the seatunnel-cli docs (referencing docs/en/connectors/sink/LocalFile.md), add a note about the Bedrock temperature fallback, and update the PR description to match the implementation.
- Severity: Medium

**Issue 3: chat_stream retry path has new logic but no test coverage**
- Location: `seatunnel-cli/tests/test_validate_hocon_placeholders.py:128`
- Why it matters: - Root cause: the `chat_stream` method gained its own try/except retry block calling `converse_stream`, but the new test file only stubs and asserts against `client.converse` (`provider.client.converse.side_effect = converse` in `_make_provider`), so the streaming retry, the `else: raise` branch, and the model-id caching on the stream path are all unverified.
- Impact: the CLI's interactive path is the streaming one, so the exact scenario reported in #11508 (streamed Bedrock call rejected for `temperature`) is the one left without a regression test; a future refactor could break `chat_stream` while `chat` tests stay green.
- Fix: extend `_make_provider` to also record `converse_stream` snapshots (returning `{"stream": []}` as the success value) and add a test mirroring `test_bedrock_temperature_rejection_retries_without_temperature` plus a non-temperature error case for the stream path.
- Evidence: `seatunnel-cli/seatunnel_cli/llm_provider.py:407  response = self.client.converse_stream(**kwargs)`; `seatunnel-cli/tests/test_validate_hocon_placeholders.py:128  provider.client.converse.side_effect = converse`
- Suggested fix: The stub only wires `client.converse`, so the new retry logic in `chat_stream` (`converse_stream`, including the `else: raise` branch) has zero coverage — and streaming is the path the CLI actually uses interactively, i.e. the scenario from #11508. Please add a `converse_stream` variant of `test_bedrock_temperature_rejection_retries_without_temperature` (a success return of `{"stream": []}` is enough) plus a non-temperature error case.
- Severity: Medium

**Issue 4: Field detection only recognizes `key = value`; HOCON's `key : value` syntax bypasses the engine-placeholder exemption**
- Location: `seatunnel-cli/seatunnel_cli/agents.py:541`
- Why it matters: - Root cause: `field_pattern = re.compile(r'^\s*"?([\w.]+)"?\s*=')` only matches the equals-sign form. HOCON (and Typesafe Config, which the Zeta/Flink/Spark engines all use to parse job configs) equally accepts `file_name_expression: "${now}"` with a colon separator.
- Impact: a perfectly valid config written with colons is still reported as "Unresolved environment variables" for `${now}`/`${uuid}`/`${transactionId}`, which re-introduces exactly the false-positive/auto-fix-loop behavior this PR (#11508) sets out to fix, just for a different but legal spelling. The line-based match also misses values that continue onto the next line.
- Fix: extend the pattern to accept both separators, e.g. `re.compile(r'^\s*"?([\w.]+)"?\s*[=:]')`, and add a colon-syntax case to `test_file_name_expression_engine_placeholders_accepted` in the new test file to lock it in.
- Evidence: `seatunnel-cli/seatunnel_cli/agents.py:543  for line in config_str.splitlines():`; `seatunnel-cli/seatunnel_cli/agents.py:544  field_match = field_pattern.match(line)`
- Suggested fix: HOCON also allows the colon separator (`file_name_expression: "${now}"`), which this pattern won't match — so a valid colon-style config would still be flagged with "Unresolved environment variables" and fed back into the auto-fix loop, the exact failure mode this PR fixes. Please change the pattern to `r'^\s*"?([\w.]+)"?\s*[=:]'` and add a colon-syntax case to the new regression tests.
- Severity: Medium

**Issue 5: Bedrock provider tests live in a file named after validate_hocon placeholders**
- Location: `seatunnel-cli/tests/test_validate_hocon_placeholders.py:21`
- Why it matters: - Root cause: `tests/test_validate_hocon_placeholders.py` explicitly bundles two unrelated concerns, as its own docstring admits: validate_hocon placeholder handling and the BedrockProvider temperature-rejection retry.
- Impact: contributors looking for llm_provider tests won't find them under this filename, and future validate_hocon-only test runs (`pytest tests/test_validate_hocon_placeholders.py`) drag in Bedrock mocking machinery; it also makes the shared `_make_provider` helper awkward to reuse.
- Fix: move the Bedrock tests (`_make_provider`, `_TemperatureRejection`, and the two `test_bedrock_*` functions) into a dedicated `tests/test_llm_provider_bedrock.py`, keeping the ASF license header on the new file.
- Evidence: `seatunnel-cli/tests/test_validate_hocon_placeholders.py:21  validate_hocon, and for the BedrockProvider temperature-rejection retry."""`
- Suggested fix: Nit: the docstring itself says this file mixes two unrelated areas. Please move the BedrockProvider tests (`_make_provider`, `_TemperatureRejection`, `test_bedrock_*`) into a dedicated `tests/test_llm_provider_bedrock.py` (with the ASF header) so test files map cleanly to the modules they cover.
- Severity: Low

# Review conclusion

- Issue 1: Bedrock temperature-rejection detection relies on fragile string matching instead of the botocore error code
- Issue 2: User-facing validation and Bedrock behavior changes are undocumented, and the PR description understates what the code does
- Issue 3: chat_stream retry path has new logic but no test coverage
- Issue 4: Field detection only recognizes `key = value`; HOCON's `key : value` syntax bypasses the engine-placeholder exemption
- Issue 5: Bedrock provider tests live in a file named after validate_hocon placeholders

SEZ9:
@SEZ9 Thanks for the thorough re-review — all five items addressed in the new head `8af568a`:

**1. Rejection detection now uses the structured botocore error code.** `_is_temperature_rejection` requires `isinstance(error, botocore.exceptions.ClientError)` and `e.response["Error"]["Code"] == "ValidationException"`; the message is only used to discriminate that `temperature` is the rejected parameter (tolerant to wording changes like "deprecated" vs "not supported"). The `except Exception` clauses in `chat`/`chat_stream` are narrowed to `ClientError`. Tests assert that a ValidationException with a different message, a ThrottlingException mentioning "temperature", and a plain RuntimeError all propagate untouched without caching the model.

**2. User-facing changes documented.** README now documents (a) the Bedrock temperature fallback under Provider Configuration → AWS Bedrock (including that a configured temperature is not applied for these models), and (b) the field-aware placeholder exemption table under Validation Pipeline, referencing the LocalFile sink docs. The commit message also describes the full `partition_dir_expression` behavior.

**3. `chat_stream` retry covered.** `tests/test_llm_provider_bedrock.py` adds stream-path tests mirroring the chat tests: rejection → retry without temperature → cached skip on the next call, plus a non-temperature error case that propagates.

**4. HOCON colon separator supported.** `field_pattern` accepts both `key = value` and `key : value`; regression tests cover accepted engine placeholders and rejected env vars in colon syntax.

**5. Tests split by module.** Bedrock provider tests moved to `tests/test_llm_provider_bedrock.py` (ASF header included); `test_validate_hocon_placeholders.py` now contains only validate_hocon cases.

Full suite: 45 passed. Head remains rebased on current dev.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `8af568a` from scratch, including the full `merge-base(dev, head)...head` diff and the actual CLI call paths.

The previous source-level blocker from my last round is closed for me now. `validate_hocon` is field-aware on the current head, `chat_stream` now follows the same Bedrock retry/caching path as `chat`, and the README documents both the Bedrock fallback and the LocalFile placeholder scope. The required `Build` check is also green on this head, so I do not see a new code-side blocker from Daniel's side.

Current gate facts:
- GitHub still shows `reviewDecision=REVIEW_REQUIRED`
- the branch is currently `ahead_by=1`, `behind_by=1` against the latest `dev` (`compare status: diverged`)

I am recording that sync fact separately, but with CI already green I am not asking for a sync-only update from the contributor side here. Because I only have read-level review controls on this repository, a write-capable maintainer may still need to perform the final approval / dismissal step on top of this technical recheck.

davidzollo (approved):
+1

Issue #11513, opened 2026-07-21
[Umbrella][CI] Redesign PR validation for fast, trusted, and scalable feedback


nielifeng:
## Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct).

## Search before asking

- [x] I searched the existing issues and pull requests and found no umbrella issue covering the complete PR CI architecture described below.

Related work exists, but covers narrower parts of the problem:

- #9976 extracts long-running modules and improves scheduling/free-disk handling.
- #11074 tracks one source of unrelated modules entering incremental IT builds.
- #9900 previously improved job ordering, timeout selection, and CI duration analysis.

This proposal is intended to coordinate and build on that work, not replace it.

## Summary

SeaTunnel PR validation currently combines broad static fan-out, limited fork-level runner concurrency, repeated builds, weak cross-job failure propagation, and an eventually-consistent status mirror. The result is slow and sometimes misleading feedback for contributors, especially for changes under `seatunnel-engine/**`.

The proposed target is a trusted, risk-tiered CI pipeline:

1. fast required checks on every PR;
2. changed-module and impact-based unit/integration tests;
3. a small engine/connector canary suite for medium-risk changes;
4. exhaustive OS/JDK/connector coverage in a merge queue, explicit `full-ci` run, and nightly validation;
5. build artifacts once per SHA/JDK and reuse them across E2E jobs;
6. direct and accurate check reporting from trusted workflow definitions.

The goal is to reduce normal PR feedback time and runner consumption without removing exhaustive validation from high-risk changes and merge candidates.

### Method and limitations

The following snapshot was collected on 2026-07-21. The sampling frame contained 530 PRs created between 2026-05-01 and 2026-07-21, within the normal Actions retention window. Two hundred PRs were selected at approximately uniform intervals over creation time rather than taking only the newest or currently active PRs. The upstream `Build` check was then followed to the actual workflow run and expanded job list in the contributor fork.

- 185 of 200 sampled PRs resolved to actual runs, covering 35 authors and 37 distinct contributor forks. Fifteen had no usable Build pointer or an unavailable fork run and remain visible as missing data rather than being counted as zero-duration builds.
- Scope classification uses changed file paths. The API returned the first 100 paths; only two sampled PRs exceeded 100 changed files, so their classification is potentially incomplete.
[...]

PR #11503, opened 2026-07-20
[Fix][CDC][Zeta] Restore runtime schema from checkpoint after failover

nielifeng:
### Purpose

Fixes CDC schema recovery after a Zeta task failover when DDL changes were captured in a completed checkpoint.

Closes #10461.

### Root cause

- IncrementalSourceReader restored checkpoint tables only when the deprecated checkpointDataType field was non-null.
- Debezium historyTableChanges were not restored into the deserializer.
- Even after restoring the deserializer, the source collector, transforms, and sink writers could keep the original runtime row type.

### Changes

- Restore non-empty checkpointTables and historyTableChanges independently of checkpointDataType.
- Replace stale Debezium table history with checkpoint history.
- Emit RestoreTableSchemaEvent before the first restored data record.
- Propagate the restored runtime schema through transforms and schema-aware sinks.
- Rebuild sink runtime serializers and writers without applying physical DDL a second time.
- Preserve compatibility with legacy checkpoints that contain empty restore state.

### Validation

- RestoreTableSchemaEventTest: passed.
- CDC restore unit tests: 6 passed, 0 failed.
- Production compilation passed for API, CDC Base, JDBC, Databend, Doris, Elasticsearch, File, Iceberg, Paimon, StarRocks, and Transforms V2.
- Transforms V2 and File Base module test suites passed during cross-module validation.
- Spotless checks passed for all changed modules.
- Real MySQL CDC to Console validation: two consecutive Worker failures and recoveries, restored rows retained the new column, 139 checkpoints completed, 0 failed, 2 restored.
- Real MySQL CDC to JDBC validation: physical ADD COLUMN executed once, runtime schema restored without duplicate DDL, 66 checkpoints completed, 0 failed, 1 restored.

PR #11198, opened 2026-06-27
[Feature][Connector-V2] Add Couchbase Sink Connector

srijan-singh:
### Purpose of this pull request
https://github.com/apache/seatunnel/issues/10443#issuecomment-4796095891

This pull request adds a new **Couchbase sink connector** for Apache SeaTunnel (V2 API).

The connector writes `SeaTunnelRow` data as JSON documents into a Couchbase bucket/scope/collection
using the Couchbase Java SDK 3.5.1. It supports:

- **Upsert mode** — insert or replace existing documents via `upsert-enable`
- **Composite document keys** — built from configurable `primary-key` field list (joined with `_`);
  falls back to a random UUID when not set
- **Batch flushing** — buffers rows in memory and flushes on size threshold (`buffer-flush.max-rows`)
  or time threshold (`buffer-flush.interval`)
- **Retry with backoff** — transient write failures are retried up to `retry.max` times with
  `retry.interval` ms delay (multiplied by attempt number)
- **Full type mapping** — all SeaTunnel primitive types mapped to Couchbase JSON equivalents;
  `BYTES` stored as Base64 string

---

### Does this PR introduce _any_ user-facing change?

**Yes.** This is a new connector. Users can now configure a `Couchbase` sink in their SeaTunnel jobs:

```hocon
sink {
  Couchbase {
    connection.string = "couchbase://localhost"
    username          = "Administrator"
    password          = "password"
    bucket            = "my_bucket"
    collection        = "my_collection"
    primary-key       = ["id"]
    upsert-enable     = true
  }
}
```

There is no change to any existing connector or API — this is purely additive.

---

### How was this patch tested?

**1. Live Docker integration test**

A standalone integration test was run against a real `couchbase/server:community-7.1.1` Docker
container, validating all core write behaviors:

| Test | Scenario | Result |
|------|----------|--------|
| [1] | Plain document insert with UUID key | ✅ PASS |
| [2] | Upsert — writing same key twice replaces the document | ✅ PASS |
| [3] | Composite primary key: `user_id + order_id` joined as `u1_o42` | ✅ PASS |
| [4] | N1QL `COUNT(*)` query returns expected document count after writes | ✅ PASS |

**2. E2E test module**

Added `CouchbaseIT` under
`seatunnel-e2e/seatunnel-connector-v2-e2e/connector-couchbase-e2e/`.

The test uses Testcontainers to start a real Couchbase container, bootstraps the cluster fully
via REST (node config → services → credentials → bucket → collection → index storage mode),
[...]

PR #11515, opened 2026-07-21
[Improve][CI] Reduce unnecessary PR validation work

nielifeng:

This is the first low-risk stabilization batch from #11513. It reduces unrelated PR validation work and improves the accuracy and observability of the mirrored `Build` check. Full Maven unit-test and integration-test validation remains enabled for API changes. Documentation-site and SeaTunnel UI builds are intentionally decoupled from pure API/core changes and now run only when their own paths change or when `backend.yml` changes.

The change:

- routes documentation, UI, and dependency-license checks with dedicated change flags;
- excludes `seatunnel-engine-ui`-only changes from the broad engine E2E route;
- uses the existing `ut-modules` output for ordinary connector unit tests while preserving full-reactor unit tests for API and engine changes;
- forces documentation, UI, and dependency jobs to run when `backend.yml` itself changes so workflow changes remain self-validating;
- uses npm lockfiles and npm cache for documentation/UI builds;
- derives `in_progress` from active reusable-workflow child jobs when the parent run still reports `queued`;
- uses the workflow run `html_url` for mirrored check details;
- publishes an attempt-aware Step Summary and a machine-readable `ci-summary.json` artifact after every Build.

`backend.yml` is already included by the existing `**/workflows/**` API route. Therefore a change to that workflow also sets `api=true` and exercises the full API-triggered unit-test and integration-test command paths; the dedicated `backend-workflow` flag additionally forces the newly narrowed documentation, UI, and dependency jobs to self-validate.

Related umbrella issue: #11513

No. This only changes CI routing, reporting, and measurement.

- `actionlint v1.7.12` on all three changed workflows.
- Ruby YAML parsing for all three changed workflows.
- Node.js syntax checks for both embedded `github-script` programs.
- Mocked status synchronization for a blocked PR whose parent run is `queued` while child jobs are active.
- Mocked CI summary generation, including skipped-job filtering, first-result time, attempt duration, and valid runner-time calculation.
- Routing tests for UI-only, engine-only, mixed engine/UI, backend-workflow, POM, JAR, and docs/tooling changes.
- `npm ci --dry-run --ignore-scripts` in `seatunnel-engine-ui`.
- Incremental Maven validation using the same new command shape:

```bash
./mvnw -B -T 1 clean verify \
  -DskipUT=false -DskipIT=true \
  -D"license.skipAddThirdParty"=true \
  -D"skip.ui"=true --no-snapshot-updates \
  -pl ":seatunnel-common" -am
```

The Maven reactor built 10 impacted/upstream modules and completed 49 tests successfully in about two minutes.

### Check list

* [x] The change is limited to GitHub Actions workflows.
* [x] Full Maven unit-test and integration-test coverage is preserved for API changes.
* [x] Pure API/core changes intentionally no longer run unrelated documentation-site or SeaTunnel UI builds.
* [x] Workflow syntax and embedded scripts were validated locally.


nielifeng:
Reviewed the full PR against the actual branch (`8eb383a`), not just the diff. The routing and reporting changes are solid. Verified three things end-to-end:

**✅ `unit-test` empty-module guard is correct.** The `if: ... || (api == 'false' && ut-modules != '')` guard means a non-API change with empty `ut-modules` skips the whole job rather than running a dangerous `-pl ''`. No degenerate full build.

**✅ `document` job `npm ci` is safe.** The website is checked out from `apache/seatunnel-website` at runtime, and that repo has a synced `package-lock.json` at its root, so `npm install → npm ci` plus the npm cache is a clean improvement. The only external assumption is that `apache/seatunnel-website` keeps its lockfile in sync — a responsibility of that repo, not this PR.

**✅ blocked-PR mirroring is genuinely fixed.** This is worth calling out because it's easy to misread from the diff alone: the PR *rewrites* `update_build_status.yml` rather than patching the old `maybeReady` whitelist. The old `mergeable_state` filter (which excluded `blocked` and caused the #11198 / #11414 / #11503 symptoms) is gone entirely, replaced by a `GET /search/issues` query (`is:pr is:open updated:>=<7d>`). A blocked PR now flows through `processPullRequest` → `getEffectiveWorkflowRun`, and when the parent run is `queued` while child jobs are active it correctly derives `in_progress`. This is a cleaner fix than adding `'blocked'` to the whitelist.

Two points introduced by the rewrite, both minor and worth a note rather than a block:

1. **Search API rate limit.** `GET /search/issues` uses GitHub's separate search quota (~30 req/min) rather than the core quota the old `pulls?state=open` loop used. With the per-PR `getEffectiveWorkflowRun` pagination on top, busy periods could brush the limit. Consider light backoff / error tolerance around the search call (the top-level `try/catch` currently rethrows and fails the whole run).

2. **7-day `updated` window narrows coverage.** Only PRs updated in the last 7 days are synced. An open PR whose CI is still resolving but which has had no update event for longer would stop being mirrored — a coverage reduction the description doesn't mention. If intentional, a comment noting the tradeoff would help; otherwise consider widening the window or exempting PRs that already have a pending `Build` check.

Nothing here blocks merge. The blocked-PR fix does what the umbrella (#11513) needs.

nielifeng:
Rechecked the latest head (`8d416a6f`) after the follow-up fix. I do not see another code change required at this point.

Verified again:

- engine/API changes preserve the full-reactor unit-test path;
- connector-only changes use the impacted-module unit-test path;
- changes to `backend.yml` force the documentation, UI, and dependency jobs to run for workflow self-validation;
- `tools/documents/**` is included in documentation routing;
- the branch is still based on the current `dev` head;
- `actionlint` v1.7.12, YAML parsing, and `git diff --check` pass locally.

The latest fork Build is still queued: https://github.com/nielifeng/seatunnel/actions/runs/29813190361. Its four initial jobs have not started yet, so this is not a final CI pass. The fork currently has an older 84-job Build with four long-running integration-test jobs still active, which is the likely source of the runner backlog.

No new reviewer change request is present as of this check.

nielifeng:
Re-reviewed the full workflow state after `8d416a6` (both commits combined, walking every routing branch end-to-end).

**The engine unit-test fix is correct and complete.** Verified the four gating conditions (`api` / `engine` / `ut-modules` / `it-modules`) are mutually exclusive and closed: engine-core changes now run the full unit-test step (`engine==true`), pure-connector changes run the impacted step, and there is no `-pl ''` degenerate path. I also confirmed `get_engine_modules` always emits the fixed smoke modules, so `engine==true` can never leave `ut-modules` empty — the job cannot be silently skipped for an engine change. Good.

One thing worth clarifying before merge, and two minor notes:

**1. Pure API changes no longer build the website or UI (coverage change).** The first commit changed:

```
document:     api || docs  ->  docs || backend-workflow
seatunnel-ui: api          ->  ui   || backend-workflow
```

So an API/core-only change now builds neither the docs site nor the UI. The checklist says "No test coverage is removed from API-triggered full validation," but website/UI build coverage *is* removed from the API path specifically. This may be an intentional decoupling (backend code vs. site/UI), which is reasonable — but if so it is worth stating explicitly, otherwise consider restoring `|| needs.changes.outputs.api == 'true'` on those two jobs.

**2. `backend-workflow` has little independent effect on forks.** The `api` path set already includes `"**/workflows/**"`, so editing `backend.yml` on any fork sets `api=true`. The only case where `backend-workflow` does something `api` does not is forcing docs/ui builds on a `backend.yml` change — which is essentially compensating for the narrowing in point 1. The combined logic is self-consistent, but a more direct fix would be to keep `|| api==true` on document/ui and drop the extra flag.

**3. The `backend-workflow` self-bootstrap is partial.** Editing `backend.yml` re-runs docs/ui/dependency but not the unit-test/IT command shapes it may have just changed. Fine to leave for the umbrella Phase 2 routing-regression work.

None of these block merge — the routing is sound. Point 1 is the only one I would resolve explicitly, since it is a real (possibly intended) coverage change that the description currently contradicts.

nielifeng:
Addressed the clarification in the PR description.

The docs/UI narrowing is intentional: pure API/core changes retain the full Maven unit-test and integration-test paths, but no longer build the unrelated documentation site or SeaTunnel UI. Restoring `api || docs/ui` would undo the main low-risk routing improvement.

The dedicated `backend-workflow` flag is also intentional. `backend.yml` already matches the existing `**/workflows/**` API route, so it sets `api=true` and exercises the full API-triggered unit-test and IT command paths. The additional flag specifically forces the newly path-scoped docs, UI, and dependency jobs to run when their routing workflow itself changes. Therefore the self-bootstrap is not partial for unit-test/IT command changes.

No workflow code change is needed for this feedback; the description now states this policy explicitly.

SEZ9 (approved):
Thanks for working on this. I re-reviewed the latest head `8d416a6f60df` from scratch.

# What this PR fixes
- User pain: Every PR — even docs-only, UI-only, or single-connector changes — triggers the full backend validation matrix (full-reactor unit tests, engine E2E, license checks), wasting hours of runner time per PR, while the mirrored `Build` check on blocked PRs misreports `queued` when reusable-workflow child jobs are already running and offers no per-run cost visibility.
- Fix approach: Add dedicated `ui`/`dependency`/`docs`/`backend-workflow` path filters in the `changes` job of `backend.yml` and rewire the `if:` conditions of `dependency-license`, `document`, `seatunnel-ui`, and the unit-test job (splitting it into full-reactor vs. `-pl ${{ ut-modules }} -am` impacted-module steps); switch npm builds to `npm ci` with lockfile-keyed caching; in `update_build_status.yml`, add `getEffectiveWorkflowRun()` that derives `in_progress` from active child jobs and uses `html_url` for details; in `build_main.yml`, add an always-run `ci-summary` job that paginates jobs, computes attempt-aware duration/runner-minute metrics, writes a Step Summary, and uploads `ci-summary-attempt-N/ci-summary.json`.
- One-line summary: The routing logic is sound and self-validating via the `backend-workflow` flag, but I want to verify a few edges before approving: the `engine` flag now flips to `false` based solely on the JSON-filtered `file_list` (discarding the script's own `true_or_false` line — correct only if the two are always consistent, including on error output); the `dependency` glob set (`**/*.jar`, `.licenserc.yaml`) versus what `checkLicense.sh` actually inspects; whether `ut-modules` can be empty when both `api` and `engine` are `false` (an empty `-pl ''` fails Maven); and the truncated `getEffectiveWorkflowRun` fallback behavior when the jobs API call itself throws.

# Review conclusion

DanielLeens (commented):

davidzollo (dismissed):
+1
LGTM

DanielLeens (commented):
Outdated.

nielifeng:
Removed the unrelated Elasticsearch timer-flush E2E config hunk from this CI routing branch and synced the branch with current `dev`. The PR diff is now scoped back to workflow/routing/reporting changes. A new Build is queued on head `646f09b5`.

DanielLeens (commented):
Outdated.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `84b63cf6cedbac1d51f3a1dc67081d4dfa751602` across the workflow-routing changes in `backend.yml`, the mirrored Build status updater, and the new Build summary job.

I also noticed @SEZ9 already approved this head. After my own pass, I agree with the source-level direction: the routing logic is coherent, the `backend-workflow` self-validation path is intentional, and I do not see a new source-level blocker in the workflow code.

CI gate update: the current required `Build` check is failing here: https://github.com/apache/seatunnel/runs/91108463083. The underlying fork run is https://github.com/nielifeng/seatunnel/actions/runs/30612924153. The only failed job I found is `Run / Dependency licenses`, and its log ends with a 40-minute timeout while the install/dependency-resolution phase was still downloading Maven dependencies:

```text
Run / Dependency licenses ... Downloading from central: ... snakeyaml-1.26.pom
Run / Dependency licenses ... Timeout of 2400000ms hit
```

So the source review is clear, but the current CI gate is not. This branch is also diverged from `dev` (`ahead_by=7`, `behind_by=7`). Please sync with the latest `dev` and rerun Build first. If `Dependency licenses` times out again after sync/rerun, then the workflow should treat that as a CI-design issue to fix directly, for example by increasing the install timeout or reducing the amount of work required in the dependency-license validation path.

# What this PR fixes
- User pain point: many PRs trigger more backend validation than their changed paths require, wasting CI runner time and making the mirrored `Build` status less informative while reusable workflow jobs are already running.
- Fix approach: add more precise path filters for docs/UI/dependency/backend-workflow changes, route selected jobs from those flags, improve mirrored Build status by inspecting child workflow jobs, and add a CI summary artifact/Step Summary.
- In one sentence: this is a CI routing and observability improvement intended to reduce unnecessary PR validation work while keeping workflow changes self-validated.

# 1. Code Review

## 1.1 Core Logic Analysis
Main runtime path:

```text
Pull request update
  -> Build check points to fork workflow run
      -> build_main.yml
          -> call-build-and-test uses backend.yml
          -> ci-summary always runs after reusable workflow

backend.yml
  -> changes job
      -> check_file_updates.py computes path buckets
      -> docs / ui / dependency / backend-workflow flags are exported
      -> engine excludes seatunnel-engine-ui paths
      -> dependency also includes deleted pom changes
  -> dependency-license / document / seatunnel-ui
      -> run when matching path flag is true or backend-workflow is true
  -> unit-test
      -> full reactor when api or engine is true
      -> impacted modules when api=false and engine=false

update_build_status.yml
  -> scheduled updater scans active PRs
      -> reads Build check output JSON
      -> fetches the underlying workflow run
      -> if run appears queued/waiting, inspects child jobs
      -> patches Build check to in_progress when child work has started
```

Key findings:
- The normal PR Build path does hit these workflow changes.
- The `backend-workflow` flag intentionally self-validates workflow-only changes by running dependency/document/UI checks.
- `getEffectiveWorkflowRun()` improves stale queued status by inspecting child jobs before patching the mirrored Build check.
- The new summary job excludes itself from the summarized workload and uploads attempt-scoped `ci-summary.json`.

## 1.2 Compatibility
Compatible with SeaTunnel runtime and user jobs. This changes CI behavior only. Existing source/connector APIs, configs, default runtime behavior, protocols, and serialization formats are untouched.

The intended side effect is reduced CI load for narrower PRs and better CI observability. The main operational risk is under-validating a path bucket; the current `backend-workflow` self-validation reduces that risk for workflow changes.

The Build status updater catches invalid Build metadata and workflow lookup failures without crashing the whole scheduled batch. The CI summary writes both Step Summary markdown and a JSON artifact for later inspection.

# 2. Code Quality

## 2.1 Style
The YAML/script structure is readable and the inline comments explain the routing choices. `git diff --check upstream/dev...HEAD` is clean in my local review worktree.

## 2.2 Tests and Stability
This PR changes CI workflow YAML rather than Java UT/E2E tests. The meaningful validation is the GitHub workflow result. Current Build is failing because `Dependency licenses` timed out, so merge must wait for a successful rerun or a workflow fix if the timeout repeats.

## 2.3 Docs
No user documentation is required; the feature is internal CI behavior.

# 3. Architecture
This is a reasonable CI architecture change: path classification remains centralized in `changes`, job routing consumes only those outputs, and Build status mirroring is separated into the scheduled updater.

# 4. Findings Summary

| --- | --- | --- | --- |

# 5. Merge Conclusion
### Conclusion: can merge after CI / merge gate is cleared

   - No source-code blocker from Daniel's side in this round.
   - The current required `Build` check is failing in `Dependency licenses` due a 40-minute install/dependency-resolution timeout, so this PR still needs a fresh successful Build before merge.

   - None from this source review. If the same timeout repeats after sync/rerun, please adjust the dependency-license workflow timeout/scope rather than treating it as unrelated connector noise.

Overall, the source-level workflow change looks good to me and I agree with @SEZ9's approval direction. The remaining issue is the current CI gate.

DanielLeens:
Rechecked the latest head `84b63cf6cedbac1d51f3a1dc67081d4dfa751602` and the current gate.

My earlier CI guidance is outdated now: the required `Build` check is successful on the current head: https://github.com/apache/seatunnel/runs/91108463083

From Daniel's side, I still do not see a reopened source-level blocker in the workflow-routing and Build-summary changes in this PR.

The branch is currently still diverged from `dev` (`ahead_by=7`, `behind_by=15`, compare status: `diverged`), and GitHub is reporting the merge state as `BLOCKED`. I am recording that as a mergeability fact rather than turning it into a new contributor sync request on a green source/CI state.

Because I only have read-level review controls on this repository, this follow-up comment is my current Daniel-side technical conclusion for the latest head. If GitHub still needs a formal approval or another merge-gate step, a write-capable maintainer will need to handle that final repository-level action.

DanielLeens (approved):

My last comment on this head already recorded that my earlier CI guidance was outdated and that the required `Build` check was passing, and that I saw no reopened source-level blocker in the workflow-routing and Build-summary changes. I left that as a comment rather than a review, so the PR still shows as awaiting review. Rechecking the current head `84b63cf6`: **all checks are passing** (4/4 SUCCESS, none failing or pending), so I am converting that conclusion into a formal approve.

Re-confirmed on this head:
- Still CI-scoped: the change affects workflow routing and Build summarization only. No API, config option, default value, protocol, serialization, connector behavior, or job-execution path is touched. Fully compatible.
- No UT or E2E test code is changed, so the flaky-test and assertion-stability analysis does not apply, and I am recording that explicitly rather than skipping it. Nothing in this PR weakens an existing check, skips a module that was previously validated, or lowers a CI gate — it removes work that was genuinely redundant, which is a different thing from reducing coverage.

Two housekeeping notes:

- The branch is still diverged from `dev` and GitHub reports the merge state as `BLOCKED`. I am recording that as a mergeability fact only and I am **not** asking you to sync — there is no failing signal to attribute to upstream drift, and re-rolling a green CI result would gain nothing.
- In my previous comment I said I only had read-level review controls and that a write-capable maintainer would need to handle the formal approval. That was wrong — I can submit a review and have done so. A committer's approval is still needed for the merge itself, which is a separate step.

Thanks for this one — trimming unnecessary PR validation work pays back on every single PR the project sees.

nielifeng:
Resolved the current `dev` conflict in merge commit `fb1b8c031`. The resolution preserves both sides:

- Keeps this PR’s scoped `ui` / `dependency` / `backend-workflow` routing and impacted-module unit-test path.
- Keeps current `dev` benchmark routing/jobs.
- Carries current `dev` Maven/Surefire memory split and Windows-specific argument syntax into both full and impacted unit-test paths.

Validation: the merge has no unresolved paths, `dev` is an ancestor of the new head, the PR diff remains limited to the original three workflow files, workflow YAML parses successfully, and `git diff --check origin/dev...HEAD` passes. A fresh Build has been triggered.

davidzollo (commented):
Posted to the wrong account by mistake (tooling error, not a content issue). The intended review has been reposted correctly under DanielLeens; please disregard this copy and see the DanielLeens review instead.

davidzollo (commented):
Duplicate of the above, also posted to the wrong account by mistake. Please disregard; see the DanielLeens review for the actual technical review content.

DanielLeens (commented):

Thanks for the update, and for the clear write-up on how the `dev` conflict was resolved (merge commit `fb1b8c031`, resolved 2026-08-12). I re-derived every section below from scratch against the current head rather than assuming anything from my 2026-08-04 approval still holds, and I can confirm your resolution note is accurate: comparing the previous approved head (`84b63cf6c`) against the current head (`fb1b8c031`) shows the three intervening commits are pure `dev`-sync merges (they bring in `TESTCONTAINERS_RYUK_DISABLED`, the unrelated `benchmark-test` job, the Windows/non-Windows `SUREFIRE_JVM_ARGS` split, and two new connector IT jobs — all already present in current `dev`, none of it introduced by this branch). This PR's own routing logic — the three-file diff against `origin/dev...HEAD` (`backend.yml`, `build_main.yml`, `update_build_status.yml`, 271 insertions / 19 deletions) — is functionally unchanged since I approved it. Good state to re-review from.

- User pain point: every PR — even docs-only, UI-only, or single-connector changes — was triggering the full backend validation matrix (full-reactor unit tests, dependency-license, docs/UI builds), wasting CI runner time, while the mirrored `Build` check on `blocked` PRs misreported `queued` when child jobs were already running and offered no per-run cost visibility.
- Fix approach: add dedicated `ui` / `dependency` / `backend-workflow` path filters in the `changes` job of `backend.yml`, exclude `seatunnel-engine-ui/**` from the `engine` flag, rewire `dependency-license` / `document` / `seatunnel-ui` to the new flags, switch the two npm builds to `npm ci` with lockfile-keyed caching, add `getEffectiveWorkflowRun()` in `update_build_status.yml` to derive `in_progress` from active child jobs before patching a stale `queued` mirror, and add an always-run `ci-summary` job in `build_main.yml` that computes attempt-aware duration/cost metrics into a Step Summary and an uploaded artifact.
- One-sentence summary: this is a CI routing and observability improvement that narrows unnecessary validation to the paths that actually changed while keeping workflow-only edits self-validating.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Runtime path for a normal PR:

```text
Pull request push
  -> build_main.yml ("Build")
      -> call-build-and-test: uses backend.yml (reusable workflow)
      -> ci-summary: always() runs after call-build-and-test, same run

backend.yml
  -> changes job
      -> check_file_updates.py computes per-path true/false + JSON file list buckets
      -> engine bucket is filtered to drop seatunnel-engine-ui/** before deriving engine=true/false
      -> new buckets: ui, dependency (pom.xml/**/pom.xml/.mvn/**/tools/dependencies/**/**/*.jar/LICENSE/NOTICE/.licenserc.yaml + deleted-poms), backend-workflow (.github/workflows/backend.yml only)
  -> dependency-license / document / seatunnel-ui
      -> gated on their new dedicated flag OR backend-workflow
  -> unit-test
      -> full reactor when api or engine is true
      -> impacted modules (-pl ut-modules -am) when both are false (job itself is skipped if ut-modules is empty)

update_build_status.yml (scheduled, every 15 min)
  -> searchActivePullRequests() (search API, 7-day window)
  -> processPullRequest()
      -> reads Build check's stored {owner, repo, run_id}
      -> fetches that workflow run (build_main.yml's own run, all jobs including ci-summary)
      -> getEffectiveWorkflowRun(): if run is still queued/pending, pages the run's jobs; if any child job is in_progress or completed-non-skipped, reports the run as in_progress instead of queued
      -> patches the PR's "Build" check_run to the (effective) status/conclusion
```

Key findings, verified against the actual file diff (not just the description):

1. **`engine` flag UI exclusion is correct** (`backend.yml:18-27`). The old code took `check_file_updates.py`'s own `true`/`false` line directly. The new code discards that line and instead re-derives `true_or_false` from whether the *UI-filtered* JSON list is `[]`. I checked `tools/update_modules_check/check_file_updates.py:78-88`: the script's own boolean is defined as exactly "changed_files list is non-empty" for the same directory set it was given, so the two booleans are guaranteed consistent by construction — SEZ9's earlier concern about the two signals diverging does not apply here, because the new code isn't comparing two independently-computed signals, it's re-deriving the boolean from the (locally re-filtered) same list the original boolean would have used.
2. **`dependency-license` re-gating is safe** (`backend.yml:412`, was `api=='true' || engine=='true'`, now `dependency=='true' || backend-workflow=='true'`). I read `tools/dependencies/checkLicense.sh:22-38`: it runs `mvnw dependency:copy-dependencies` across the *entire* reactor (`-pl '!seatunnel-dist'`, no module scoping) and diffs the resulting third-party jar set against `known-dependencies.txt`. This check's output can only change when the dependency closure changes, which can only happen through a `pom.xml`/`.mvn/**`/jar/license-file edit — all of which are in the new `dependency` glob. A pure Java-source API or engine change (no pom touched) genuinely cannot move this check's result, so dropping `api`/`engine` as triggers here is a correctness improvement, not a coverage loss.
3. **Root `pom.xml` is explicitly included as a literal pattern, not just `**/pom.xml`** (`backend.yml:50`, both `"pom.xml"` and `"**/pom.xml"` are passed). I verified this is necessary, not redundant: Python's `fnmatch.fnmatch('pom.xml', '**/pom.xml')` returns `False` (fnmatch has no directory-boundary semantics for `**`, unlike a real glob), so without the explicit literal, an edit to the root POM would have been silently missed by the `dependency` flag. Good catch by this PR — worth calling out as a real, non-obvious correctness detail that's been handled properly.
4. **`document`/`seatunnel-ui` narrowing from `api` to `docs`/`ui` (+`backend-workflow`) is the intentional, already-discussed tradeoff** (`backend.yml:436`, `:467`). I confirmed the current head still has this narrowing and it's documented in the PR description per nielifeng's 2026-07-21 08:39 comment; API/engine changes still get the full Maven unit-test/IT path, they just no longer rebuild the unrelated website/UI. No new regression here since my last review.
5. **`unit-test` job cannot degenerate to `-pl ''`.** Verified the job-level gate (`backend.yml:501`) skips the whole job when `api=='false'` and `ut-modules==''`, so the impacted-module step's `-pl '${{ ut-modules }}'` never executes with an empty value. Also verified `get_engine_modules()` (`tools/update_modules_check/update_modules_check.py:29-31`) is a hard-coded non-empty string regardless of input, so `engine=='true'` can never leave `ut-modules` empty either — this confirms nielifeng's 2026-07-21 08:37 self-review claim and SEZ9's third open question are both correctly resolved in the current head.
6. **`backend-workflow` intentionally only tracks `.github/workflows/backend.yml`**, not `build_main.yml`/`update_build_status.yml` (`backend.yml:59`). This is fine: those two files don't feed `document`/`seatunnel-ui`/`dependency-license` routing, so they don't need to self-trigger those jobs. Note also that *any* `.github/workflows/**` edit already sets `api=true` via the pre-existing pattern at `backend.yml:260` (`"**/workflows/**"`), so a `build_main.yml`-only or `update_build_status.yml`-only change still forces the full unit-test matrix — over-inclusive, not under-inclusive, for that path.

I traced several concrete file-change scenarios end-to-end against this logic (connector-only PR, engine-core PR, engine-UI-only PR, docs-only PR, `backend.yml`-only PR, root `pom.xml` bump) and in every case the job that should run does run. I did not find a silent-coverage-loss bug in the path-filter logic itself.

7. **New finding — `ci-summary` job in `build_main.yml` has no error containment and shares the `Build` check's fate** (`build_main.yml:35-191`). This job runs `always()` after `call-build-and-test` inside the *same* `build_main.yml` run, whose workflow is literally named `"Build"` (`build_main.yml:20`). `update_build_status.yml`'s `processPullRequest()` fetches that same run via `GET /repos/{owner}/{repo}/actions/runs/{run_id}` and mirrors its overall `status`/`conclusion` onto the PR's `Build` check (`update_build_status.yml:174-219`) — the mirrored value is the whole run's conclusion, not `call-build-and-test`'s job conclusion specifically. The new `github-script` step (`build_main.yml:169-190`) has no `continue-on-error: true` and no top-level `try/catch` around `github.paginate(listJobsForWorkflowRun, ...)`, `core.summary.addRaw(...).write()`, or the `actions/upload-artifact@v4` step. A transient failure in any of these (rate limit, API hiccup, upload timeout) fails the `ci-summary` job, which — per standard GitHub Actions semantics — flips the *whole run's* conclusion to `failure`, which then gets mirrored onto the PR's required `Build` check even though `call-build-and-test` (the actual backend validation) passed cleanly. That's a false-negative on the exact required gate this PR series is trying to make more trustworthy. It fails closed (blocks a good PR, doesn't pass a bad one), so it's not a correctness/data-safety issue, just an avoidable reliability regression on the new job specifically.

## 1.2 Compatibility Impact

**Fully compatible.** This is CI workflow/YAML only. No source API, connector behavior, SeaTunnel config `Option`, default value, protocol, serialization format, or job-execution runtime path is touched. Nothing here affects any already-running or historical SeaTunnel job.

Intended effect: less redundant CI work per PR, and a more accurate `Build` status mirror for `blocked` PRs (via `getEffectiveWorkflowRun`). Side effect newly identified in 1.1.7: the `ci-summary` job can itself cause a false `Build` failure independent of actual test results. `npm ci` + `actions/setup-node` caching for `document`/`seatunnel-ui` is a safe, well-ordered change (checkout precedes the cache step in both jobs; both lockfiles exist in the repos being cached).

`getEffectiveWorkflowRun()` in `update_build_status.yml` (lines 335-374) correctly wraps its jobs-listing call in `try/catch` and falls back to the unmodified `workflowRun` status on error — this is good defensive coding and was already the case at my last review. The new `ci-summary` step in `build_main.yml` is the one place lacking that same discipline (Issue 1 below).

**Formal issues:**

| # | Issue | Location | Raised by another reviewer |
|---|-------|----------|------|
| 1 | `ci-summary` job has no error containment; a metrics-collection failure flips the whole `Build` run to `failure` even when `call-build-and-test` passed | `.github/workflows/build_main.yml:35-191` | No |
| 2 | (Informational, pre-existing, not introduced by this PR) `deleted-poms` still uses only `"**/pom.xml"`, which — per the same `fnmatch` semantics verified in 1.1.3 — would not catch deletion of the *root* `pom.xml`; this PR now feeds that same `deleted_poms_changed` signal into the new `dependency` flag | `.github/workflows/backend.yml:37-40` (pre-existing `deleted_poms_files` pattern, new `deleted_poms_changed` consumer at line 39) | No |

Issue 2 is an extremely low-practical-risk edge case (deleting the reactor's root POM would break nearly every job in the workflow outright, so under-triggering `dependency-license` specifically is the least of the fallout) and predates this PR; I'm noting it only because this PR is the first to wire that signal into a new decision, and its fix (adding a literal `"pom.xml"` to that check as well, mirroring line 50) is a one-line, low-risk companion fix if you want to close the loop while you're in this file.

# 2. Code Quality Assessment

## 2.1 Coding Standards

YAML is readable, and the routing rationale is explained inline where it matters (e.g., the UI-exclusion comment at `backend.yml:19` area, the `getEffectiveWorkflowRun` doc comment). `git diff --check origin/dev...HEAD` is clean in my local review worktree. The new `github-script` block in `build_main.yml` is dense (140+ lines of inline JS) but reasonably decomposed into named helper functions (`countBy`, per-job `intervals` mapping) — acceptable for CI tooling, though see 3.2 on extracting it.

No Java UT/E2E test code is changed in this PR (only `.github/workflows/*.yml`). Stability rating: **N/A — no UT/E2E test code touched**, recording this explicitly rather than skipping it, consistent with my 2026-08-04 review of the prior head. The meaningful validation here is the GitHub Actions run itself; current required `Build` check is green (`https://github.com/apache/seatunnel/runs/94039104429`, 1h32m54s).

## 2.3 Documentation Updates

No `docs/en` / `docs/zh` update required — this is internal CI behavior with no user-visible SeaTunnel feature surface.

# 3. Architectural Soundness

**Precise fix.** Centralizing path classification in the single `changes` job and having every downstream job consume only its outputs keeps the routing logic in one auditable place, which is exactly the right shape for this kind of change.

## 3.2 Maintainability

Reasonable. The one thing I'd suggest, non-blocking: the 140-line inline `github-script` in `ci-summary` would be easier to unit-test and diff-review as a standalone `.github/scripts/*.js` file (the repo doesn't currently do this for workflow scripts, so it's a style choice, not a gap this PR needs to fix alone).

## 3.3 Extensibility

Good — adding another path bucket in the future follows the same three-line pattern (`check_file_updates.py` call + flag + optional consumer `if`), and `getEffectiveWorkflowRun`'s job-inspection approach generalizes to any future reusable-workflow-based required check.

Not applicable — no runtime state, checkpoint, serialization, or upgrade path is touched. This does not affect SeaTunnel's own version-to-version compatibility in any way.

# 4. Issue Summary

| 1 | `ci-summary` job lacks `continue-on-error`/try-catch containment; a metrics-collection hiccup flips the required `Build` check to failure independent of actual test results | `.github/workflows/build_main.yml:35-191` | Medium |
| 2 | Pre-existing (not introduced here): `deleted-poms` filter's `"**/pom.xml"` pattern doesn't match root-level `pom.xml` deletion, and this PR now feeds that signal into the new `dependency` flag | `.github/workflows/backend.yml:37-40` | Low |

# 5. Merge Recommendation

1. Blockers — must be fixed: none. Neither issue above is a functional regression, a coverage-loss bug in the path-filter logic, or a compatibility break.
   - Issue 1: add `continue-on-error: true` to the `ci-summary` job (or step), or wrap the script body in `try { ... } catch (error) { core.warning(...) }` so a reporting glitch can't flip the required `Build` check. Given this PR's own stated goal is making the `Build` mirror more trustworthy, I'd genuinely like to see this closed before or shortly after merge, but I'm not blocking on it — the failure direction is fail-closed (blocks a good PR, worst case is a confusing rerun), not fail-open.
   - Issue 2: optional one-line companion fix (`"pom.xml"` added alongside `"**/pom.xml"` in the deleted-poms check), purely for symmetry with the add/modify case already handled correctly in this PR.

I agree with SEZ9's 2026-07-21 approval direction; all four of their open verification questions (engine-flag boolean consistency, dependency glob vs. `checkLicense.sh` scope, `ut-modules` empty-string degenerate case, `getEffectiveWorkflowRun` fallback on API failure) are independently confirmed resolved at the current head. I'm reaffirming my own 2026-08-04 approval conclusion: this PR's routing logic is unchanged in substance since then — the intervening commits are `dev`-sync merges, not new PR logic, as nielifeng's 2026-08-12 comment accurately describes and as I independently verified by diffing `84b63cf6c..fb1b8c031`. Thanks again for narrowing this down — it's a meaningful CI cost/reliability win, and worth landing as-is with the `ci-summary` hardening as a fast follow-up.

DanielLeens (approved):
Formalizing this as an approval on the current head. My last comment on `fb1b8c031` already confirmed the routing logic is unchanged in substance since my 2026-08-04 approval of the prior head — the intervening commits are pure dev-sync merges — and that the required `Build` check was green. I left that as a comment rather than a formal review. Rechecking now: `Build`, `Notify test workflow`, and `labeler` are still all green on this exact head, and I don't see any other outstanding blocker from my side (the two items I noted, ci-summary error containment and a pre-existing deleted-poms glob edge case, are both explicitly non-blocking follow-ups). Converting my conclusion into a formal approve.

nielifeng:
Resolved the current `dev` conflict in merge commit `04f5bdab331f`.

The resolution keeps the current upstream CI work while preserving this PR's remaining independent improvements:

- Keeps the latest connector shard, benchmark, memory, and Windows command changes from `dev`.
- Keeps this PR's scoped UI, dependency, backend-workflow, documentation, and impacted-module unit-test routing.
- Keeps the attempt-aware CI summary artifact and job summary.
- Uses the stronger workflow-job normalization merged in #11840 instead of retaining this branch's older duplicate implementation.
- Retains this PR's `actions/github-script@v7` upgrade and check-runs pagination.

Validation on the merged result:

- All three workflow YAML files parse successfully.
- Embedded `github-script` JavaScript passes syntax checks.
- CI routing tool tests: 16 passed, 0 failed.
- Conflict-marker and whitespace checks are clean.
- The PR diff remains limited to the original three workflow files.

A fresh Build has been triggered on the conflict-free head.

DanielLeens (approved):
Thanks for the ping on the fresh `dev` sync. I re-reviewed the current head `04f5bdab331f` completely from scratch rather than assuming my 2026-08-14 approval of `84b63cf6c` still applies, and independently re-ran the checks below rather than trusting the merge-commit description at face value.

- User pain point: every PR — even a docs-only, UI-only, or single-connector change — was triggering the full backend validation matrix (full-reactor unit tests, dependency-license scan, docs/UI site builds), burning CI runner time on every single PR, while the mirrored `Build` check on the apache-side PR page could misreport `queued` for a `blocked` PR whose underlying fork jobs were already running.
- Fix approach: add dedicated `ui` / `dependency` / `backend-workflow` path buckets to the `changes` job in `backend.yml`, exclude `seatunnel-engine/seatunnel-engine-ui/**` from the `engine` bucket, re-gate `dependency-license` / `document` / `seatunnel-ui` on the new buckets, split `unit-test` into a full-reactor path (`api`/`engine`) and an impacted-module path (`-pl ut-modules -am`), move the two npm builds to `npm ci` with lockfile-keyed caching, and add an always-run `ci-summary` job in `build_main.yml` that records attempt-aware CI cost metrics.
- One-sentence summary: a CI routing/observability improvement that narrows unnecessary validation to the paths that actually changed while keeping every workflow-only edit self-validating.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**What actually changed since my last approval.** I diffed `84b63cf6c..04f5bdab331f` before reviewing, per nielifeng's merge-commit note. The intervening commits are the `dev`-sync resolutions of `fb1b8c031` (08-12) and `04f5bdab331f` (08-19). The second one specifically **dropped this branch's own duplicate `getEffectiveWorkflowRun()` implementation in `update_build_status.yml`** in favor of the equivalent logic already merged into `dev` via #11840. I confirmed this is real, not just claimed: `git diff origin/dev...HEAD -- .github/workflows/update_build_status.yml` is now only two trivial hunks — `actions/github-script@v6 -> v7`, and adding `per_page: 100` to the `GET /repos/{owner}/{repo}/commits/{ref}/check-runs` call used to locate the PR's `Build` check-run (a real, useful fix on its own: without it, the default GitHub page size of 30 could make `checkRuns.data.check_runs.find(...)` miss the `Build` check-run entirely on a commit with a large check-run count, silently leaving that PR's mirrored status stale). The `backend.yml` and `build_main.yml` routing/`ci-summary` logic I approved on 08-14 is otherwise unchanged in substance — confirmed by re-diffing them against my previous review notes line-by-line, not by assumption.

Runtime path for a normal PR (re-traced from the current head, not carried over from memory):

```text
Pull request push (fork)
  -> build_main.yml ("Build")
      -> call-build-and-test  [build_main.yml:28-34]   <- NO job-level `if:`, always runs
          -> uses: ./.github/workflows/backend.yml (reusable)
      -> ci-summary            [build_main.yml:35-191] <- needs: [call-build-and-test], if: always()

backend.yml -> changes job [backend.yml:150-276 area]
  -> check_file_updates.py computes per-bucket true/false + JSON file lists
  -> engine bucket [backend.yml:183-190]: re-derives true/false from a Python-filtered
     file_list that drops "seatunnel-engine/seatunnel-engine-ui/**" first
  -> new buckets: ui (seatunnel-engine-ui/** only), dependency (pom.xml / **/pom.xml /
     .mvn/** / tools/dependencies/** / **/*.jar / LICENSE / NOTICE / .licenserc.yaml,
     OR-ed with the pre-existing deleted-poms signal), backend-workflow
     (.github/workflows/backend.yml only)
  -> dependency-license / document / seatunnel-ui: gated on their new dedicated flag
     OR backend-workflow [backend.yml:411,448,482]
  -> unit-test job-level gate [backend.yml:512]:
       api=='true' || (api=='false' && ut-modules != '')
     -> full-reactor step runs when api||engine [backend.yml:532,538]
     -> impacted-module step (-pl ut-modules -am) runs when !api && !engine [backend.yml:544,550]

update_build_status.yml (scheduled, apache/seatunnel side)
  -> polls each open PR's stored fork {owner, repo, run_id}
  -> fetches that run's jobs, derives an effective status (dev's #11840 logic, unchanged
     by this PR now), and mirrors it onto the PR's "Build" check-run
```

**Key findings, all independently re-verified against the actual current-head file content (not the PR description):**

1. **`engine` flag re-derivation is internally consistent, not a two-signal race.** `backend.yml:183-190` discards the script's own `true_or_false` line and recomputes it from whether the *engine-ui-filtered* JSON list is `[]`. I read `check_file_updates.py`'s `ua` function: its boolean is exactly "the file list for this directory set is non-empty," so re-deriving the boolean from the same (locally re-filtered) list is definitionally consistent — there's no way for the two signals to diverge, because there aren't two independently-computed signals.
2. **`get_engine_modules()` still guarantees `ut-modules` can never go empty for a real engine-core change.** I re-fetched `tools/update_modules_check/update_modules_check.py` at the current head and confirmed lines 29-31 are unchanged: `get_engine_modules(files)` unconditionally prints a fixed non-empty string (`,connector-seatunnel-e2e-base,connector-console-seatunnel-e2e`) regardless of its input. So even though `engine_files` now feeds it the *filtered* (engine-ui-excluded) list, the output is identical either way — an engine-core change can never leave the `unit-test` job's `-pl` argument empty. I also confirmed a **pure engine-UI-only change now correctly skips the job**: the filtered `engine_files` becomes `[]`, `engine=false`, `engine-modules` step doesn't run ([backend.yml:297] `if: engine=='true'`), and if no other bucket contributes modules, `ut-modules=''` and the job-level guard at `backend.yml:512` skips the whole job — this is the intended UI/Java decoupling, not a coverage hole.
3. **`dependency-license` re-gating is provably safe, not just plausible.** I read `tools/dependencies/checkLicense.sh` at the current head: it runs `./mvnw clean -pl '!seatunnel-dist' ... dependency:copy-dependencies` across the **entire reactor** (no module scoping), then diffs the resulting third-party jar set against `known-dependencies.txt`. This check's output is a pure function of the dependency closure, which can only move via a `pom.xml` / `.mvn/**` / jar / license-file edit — all covered by the new `dependency` bucket. A pure Java-source (API or engine) change with no POM edit cannot change this check's result, so dropping `api`/`engine` as triggers here is a correctness improvement, not a coverage loss.
4. **Root `pom.xml` is correctly handled as a `fnmatch` edge case, and I independently reproduced the gap this PR closes.** I ran `python3 -c "import fnmatch; print(fnmatch.fnmatch('pom.xml','**/pom.xml'))"` myself against the actual `check_file_updates.py` matching logic (`glob.fnmatch.fnmatch`) — it prints `False`. `fnmatch` has no directory-boundary semantics for `**`, so `"**/pom.xml"` alone would silently miss a root-level `pom.xml` edit. The `dependency` bucket glob at `backend.yml:206` (in the current head numbering) passes both `"pom.xml"` and `"**/pom.xml"` explicitly, closing that gap for the add/modify case. This is a genuinely non-obvious, correctly-handled detail.
5. **The required `Build` check itself can never be silently absent.** This is the part of the task I want to answer directly: `call-build-and-test` in `build_main.yml:28-34` has **no job-level `if:`** — it runs unconditionally on every push, so the reusable `backend.yml` workflow (and therefore some conclusion for the overall "Build" run) always executes. The routing logic in this PR only decides which *inner* jobs of that reusable workflow run; it never gates the outer workflow invocation itself. So branch protection's required `Build` check cannot be silently skipped by anything in this diff — the worst-case failure mode is the check reporting `failure` when it shouldn't (see Issue 1), never a missing check.
6. **Self-validating case, confirmed on this exact PR's own fork run, not assumed.** Since `backend-workflow` is true for this PR (it edits `backend.yml`), I fetched the actual fork run (`nielifeng/seatunnel` run `32207741000`, triggered by this exact head `04f5bdab331f`) and confirmed via the Jobs API that `Run / Build website`, `Run / Dependency licenses`, `Run / Build SeaTunnel UI`, and `CI summary` all executed and completed with `conclusion: success`. The self-bootstrap this PR relies on to validate its own routing changes is demonstrably working, not just designed to work.
7. **Carryover — `ci-summary` still has no error containment (previously Issue 1, still present, still non-blocking).** `build_main.yml:35-191` — the `github-script@v7` step has no `continue-on-error` and no top-level `try/catch` around `github.paginate(listJobsForWorkflowRun, ...)`, `core.summary.addRaw(...).write()`, or the `actions/upload-artifact@v4` step. I confirmed with `grep -n "continue-on-error" build_main.yml` that the file still contains zero matches. Because `ci-summary` shares the same workflow run as `call-build-and-test`, a transient failure in it (rate limit, upload hiccup) would flip the whole `Build` run's conclusion to `failure` and get mirrored onto the PR's required check, even though the actual backend validation passed. This is fail-closed (blocks a good PR, never passes a bad one) so it is not a correctness or gate-integrity issue, just an avoidable reliability regression on the new job — I raised this on 08-12 and it's still open; not presenting it as a new finding.
8. **Carryover, pre-existing, not introduced by this PR — `deleted-poms` still can't detect a deleted root `pom.xml`.** The `deleted-poms` glob (`backend.yml:196-197` area) still only uses `"**/pom.xml"`, which (per the same `fnmatch` gap verified in point 4) would not match a root-level `pom.xml` deletion. This PR now feeds that same `deleted_poms_changed` signal into the new `dependency` flag (line `if [[ "$deleted_poms_changed" == "true" ]]`), so the gap is inherited rather than introduced. Practically very low risk — deleting the reactor's root POM breaks nearly every job outright, so under-triggering `dependency-license` specifically is the least of the fallout.

I also worked through the same set of concrete scenarios as my previous review (connector-only PR, engine-core-only PR, engine-UI-only PR, docs-only PR, `backend.yml`-only PR, root `pom.xml` bump) against the *current* file content end-to-end, and in every case the job that should run does run, and the required `Build` check is always produced.

## 1.2 Compatibility Impact

**Fully compatible.** This is CI workflow/YAML only. No SeaTunnel `Option`, API, protocol, serialization format, connector runtime behavior, or job-execution path is touched. Nothing here affects any already-running or historical SeaTunnel job. N/A for checkpoint/savepoint/upgrade concerns.

Intended effect (and the one I verified is real, not aspirational): fewer redundant CI job executions per PR, and a less stale `Build` mirror on `blocked` PRs. `npm ci` + `actions/setup-node` lockfile-keyed caching for `document`/`seatunnel-ui` is safe and well-ordered — checkout precedes the cache step in both jobs, and both lockfiles exist in the repositories being cached. The one identified side effect is Issue 1 above (an avoidable, fail-closed reliability gap on the new `ci-summary` job); nothing else introduces a resource leak, unbounded retry, or lock-contention concern — this is single-shot CI tooling with no persistent state.

`update_build_status.yml`'s effective-run-status logic (now sourced from `dev`/#11840 rather than duplicated in this branch) already wraps its jobs-listing call defensively and falls back cleanly on API error — unchanged and still sound. The new `ci-summary` step in `build_main.yml` is the one place lacking that same discipline (Issue 1).

**Formal issues (both carryovers from my 08-12 review, neither newly discovered, both still open at this head):**

| # | Issue | Location | Raised by another reviewer |
|---|-------|----------|------|
| 1 | `ci-summary` job has no error containment; a metrics-collection hiccup flips the whole required `Build` run to `failure` even when `call-build-and-test` passed | `.github/workflows/build_main.yml:35-191` | No |
| 2 | (Pre-existing, not introduced by this PR) `deleted-poms` filter's `"**/pom.xml"` pattern can't match a root-level `pom.xml` deletion; this PR now feeds that signal into the new `dependency` flag | `.github/workflows/backend.yml` (`deleted_poms_files` pattern / `deleted_poms_changed` consumer) | No |

# 2. Code Quality Assessment

## 2.1 Coding Standards

YAML is readable and the routing rationale is explained inline where it matters. I ran a conflict-marker scan (`grep -n "^<<<<<<<\|^=======\|^>>>>>>>"`) across all three changed files after the fresh `dev` merge — clean, no leftover markers. I also parsed all three files with `yaml.safe_load` — all parse successfully, confirming the workflow YAML syntax is valid post-merge. I extracted both embedded `github-script` bodies and syntax-checked them with `node --check` (wrapping the top-level-`await` one in an `async function` shim, since `actions/github-script` executes these as an async function body internally) — both are syntactically valid JavaScript.

No Java UT/E2E test code is touched — this PR only changes `.github/workflows/*.yml`. **Stability rating: N/A — no UT/E2E test code touched**, recorded explicitly rather than skipped, consistent with my prior rounds.

The meaningful validation for a workflow-only PR is (a) the GitHub Actions run itself, and (b) the repo's existing CI-routing Python test suite. I did not just trust the "16 passed, 0 failed" claim in the merge-commit note — I fetched `tools/update_modules_check/{update_modules_check.py,check_file_updates.py,ci_scope.py,test_update_modules_check.py,test_ci_scope.py}` at the current head and ran `pytest` locally: **16 passed, 0 failed**, independently reproduced. One gap worth flagging as non-blocking: the actual new routing logic this PR adds (the `ui`/`dependency`/`backend-workflow` bucket computation and the engine-UI-exclusion filter) lives as inline bash/python-one-liner code directly inside `backend.yml`'s `run:` block, not as a testable Python function — so it isn't covered by `test_update_modules_check.py`/`test_ci_scope.py` at all; those tests exercise unrelated, pre-existing scope logic. This is consistent with how the pre-existing `api`/`docs` buckets were already handled (never unit-tested either), so it's a Low-severity, pattern-consistent gap rather than something this PR regresses.

## 2.3 Documentation Updates

No `docs/en` / `docs/zh` update required — this is internal CI/CD behavior with no user-visible SeaTunnel feature surface.

# 3. Architectural Soundness

**Precise fix.** Centralizing path classification in the single `changes` job and having every downstream job consume only its outputs keeps the routing logic in one auditable place. Dropping this branch's own duplicate `getEffectiveWorkflowRun()` in favor of `dev`'s (#11840) version during the latest sync is the right call — it removes a maintenance fork rather than leaving two implementations to drift.

## 3.2 Maintainability

Good. The one non-blocking suggestion I'd repeat from my last round: the ~140-line inline `github-script` in `ci-summary` would be easier to unit-test and diff-review as a standalone `.github/scripts/*.js` file, though the repo doesn't currently do this for any workflow script, so it's a style choice rather than a gap unique to this PR.

## 3.3 Extensibility

Good — adding another path bucket follows the same three-line pattern (`check_file_updates.py` call + flag + optional consumer `if`), and the effective-run-status approach in `update_build_status.yml` (now shared with `dev`) generalizes to any future reusable-workflow-based required check.

Not applicable — no runtime state, checkpoint, serialization, or SeaTunnel upgrade path is touched by this PR.

# 4. Issue Summary

| 1 | `ci-summary` job lacks `continue-on-error`/try-catch containment; a metrics-collection hiccup flips the required `Build` check to failure independent of actual test results (carryover, non-blocking) | `.github/workflows/build_main.yml:35-191` | Medium |
| 2 | Pre-existing, not introduced here: `deleted-poms` filter's `"**/pom.xml"` pattern doesn't match root-level `pom.xml` deletion, and this PR now feeds that signal into the new `dependency` flag | `.github/workflows/backend.yml` | Low |

# 5. Merge Recommendation

1. **Blockers — must be fixed:** none. Neither issue above is a functional regression, a silent-coverage-loss bug in the path-filter logic, or a risk to the required-check gate's integrity (point 5 in 1.1 above shows the `Build` check can never be silently skipped — the failure mode of Issue 1 is fail-closed, not fail-open).
   - Issue 1: add `continue-on-error: true` to the `ci-summary` job/step, or wrap the script body in `try { ... } catch (error) { core.warning(...) }`, so a reporting glitch can't flip the required `Build` check. Given this PR's own goal is making the `Build` mirror more trustworthy, I'd like to see this land as a fast follow-up, but I'm not blocking on it.
   - Issue 2: optional one-line companion fix (add a literal `"pom.xml"` alongside `"**/pom.xml"` in the deleted-poms glob), purely for symmetry with the add/modify case already handled correctly.

Overall: this is a meaningful CI cost/reliability win, and the latest `dev`-sync merge (`04f5bdab331f`) doesn't just avoid regressing the logic I approved on 08-14 — it actively improves it, by retiring this branch's own `getEffectiveWorkflowRun()` duplicate in favor of the stronger version already merged into `dev`. All fresh checks I ran today — YAML parse, conflict-marker scan, embedded-JS syntax check, the CI-routing Python test suite (16/16, reproduced locally), and the live fork run for this exact head (self-validation of the `backend-workflow` bucket) — are consistent with the PR being correct and ready. Converting my conclusion into a formal approve on this head.

DanielLeens (approved):
Re-confirming approval on the current head `d807538a3`.

I diffed the actual PR-owned files (`backend.yml`, `update_build_status.yml`, `.asf.yaml`, `config/plugin_config`, `labeler/label-scope-conf.yml`) between the commit I approved yesterday (`04f5bdab331f`, reviewed as part of `fb1b8c031`'s follow-on) and this new head (`d807538a3`). The diff is a pure `dev`-sync merge with **zero net change** to this PR's own content:

- The `fail-fast: false` additions across the matrix jobs, the removal of `seatunnel-shade/**` from the `api_files` change-detection pattern, and the removal of the `check_shaded_jackson_refs.py` step all already exist verbatim on `origin/dev` (traced to unrelated upstream commits, including the `#11851` Jackson-relocation revert) — none of it originates from this branch. I verified this by diffing each file against `origin/dev` directly rather than trusting the merge-commit message.
- No new human comment or reviewer activity occurred between my last approval and this sync.

Nothing new to evaluate. My prior conclusion stands unchanged: the CI routing/observability logic is correct, compatible, and low-risk, and the `Build` check is green. Approving again at this head for the record.

Issue #11514, opened 2026-07-21
[Bug][Zeta] Job info REST API fails when a floating-point metric is NaN


nielifeng:
### Search before asking

- [x] I searched the existing issues and did not find a matching report.

### SeaTunnel Version

3.0.0-SNAPSHOT, upstream dev commit dbead9b01fc42ddc5a7f8699311ebf52dbb5d937

### Bug behavior

A GET request to the job-info REST endpoint can return HTTP 500 while a newly submitted streaming job is initializing. The job itself continues to initialize and later runs normally.

The server log contains:

    java.lang.NumberFormatException: Character N is neither a decimal digit number,
    decimal point, nor "e" notation exponential mark.
        at java.math.BigDecimal.<init>(BigDecimal.java:522)
        at org.apache.seatunnel.engine.server.rest.service.BaseService.metricsToJsonObject(BaseService.java:1189)
        at org.apache.seatunnel.engine.server.rest.service.JobInfoService.getJobInfoJson(JobInfoService.java:81)

BaseService.metricsToJsonObject currently converts every Float, Double, or BigDecimal value using:

    new BigDecimal(value.toString()).toPlainString()

A metric appears to contain NaN during job startup. BigDecimal does not accept NaN or Infinity, so the complete job-info response fails.

### Expected behavior

The job-info REST endpoint should remain available when a metric temporarily has a non-finite floating-point value. Non-finite values should be handled explicitly, for example by returning null, a defined string representation, or omitting only the affected metric.

### Reproduction steps

1. Deploy upstream dev commit dbead9b01fc42ddc5a7f8699311ebf52dbb5d937.
2. Submit a streaming FakeSource-to-Console job with checkpointing enabled.
3. Poll the job-info endpoint immediately after submission while the job transitions through its startup states.
4. Observe an intermittent HTTP 500 and the NumberFormatException above.

The observed job subsequently reached RUNNING, processed data normally, completed nine checkpoints with zero failures, and canceled cleanly.

### Impact

Monitoring, automation, and control-plane clients can treat the HTTP 500 as a job or cluster failure even though only one startup metric is non-finite. The exception also creates an application ERROR record during otherwise successful job execution.

### Suggested fix direction

Before constructing BigDecimal, handle non-finite Float and Double values using Float.isFinite and Double.isFinite. Add a REST service test covering NaN and positive/negative Infinity so one metric cannot fail the entire job-info response.

DanielLeens:
[...]

PR #11517, opened 2026-07-21
[Fix][Zeta] Fix job info REST API fails when floating-point metric is NaN

Kareemingit:
… NaN (#11514)

### Check list


DanielLeens (commented):
Thanks for the contribution! I reviewed the complete current-head diff, test, deletion ledger, and the actual job-info metrics conversion path.

The fix is correctly scoped: finite `Float`/`Double` values still use the existing `BigDecimal(...).toPlainString()` representation, while `NaN` and positive/negative infinity are detected before that constructor and emitted as explicit strings. This preserves the existing string-valued metrics contract and also works through the nested-map recursion. The new unit test covers finite values, all three non-finite shapes, and a nested metric without timing or external-resource dependencies.

I do not see a source-level blocker on this head. No required `Build` result is visible yet, so CI still needs to start and pass before merge; Daniel has READ permission, so a write-capable maintainer must complete the approval gate afterward.

DanielLeens:
Thanks for the solid contribution and the careful test coverage here. We really appreciate you digging into this Zeta edge case, and I hope to see more deep contributions from you around runtime behavior and API robustness.

DanielLeens:
Thanks for the contribution and for digging into this edge case. The fix is nicely scoped, and the added regression test makes the intent very clear. Please keep the deep-dive contributions coming, especially around SeaTunnel engine and runtime edge cases like this.

davidzollo (approved):
Looks good to me. Thanks for the targeted fix and the solid regression coverage.

PR #11518, opened 2026-07-21
[Feature][Connector-V2] Add StarRocks timer flush

nzw921rx:
## Purpose

Add engine-level timer flush support to the StarRocks sink so buffered rows can be written with bounded latency before the normal batch thresholds or checkpoint interval are reached.

## Changes

- Register the StarRocks writer flush action through `SinkWriter.Context`.
- Reuse the existing synchronized Stream Load flush path and propagate flush failures.
- Add focused unit coverage for action registration, execution, and failure propagation.
- Add a Zeta E2E case backed by a continuous MySQL CDC source, covering both snapshot and binlog records.
- Document the Zeta configuration and at-least-once semantics in English and Chinese.

## Validation

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-starrocks -DskipIT -DskipUT=false -Dtest=StarRocksSinkWriterTest test`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-starrocks-e2e -DskipTests -DskipUT -DskipIT test-compile`


DanielLeens (commented):
Thanks for working on this. I rechecked the full current diff and do not see a new source-level blocker in the StarRocks timer-flush change. Wiring `SinkWriter.Context` through `StarRocksSink.createWriter()` and registering `context.registerFlushAction(this::timerFlush)` in `StarRocksSinkWriter` makes the engine timer path hit the same `manager.flush()` logic that `prepareCommit()` already uses, and the added writer + E2E coverage exercises that path directly.

The required checks are green on the current head. From Daniel's side the technical verdict is clear; since I only have read-level review controls here, a write-capable maintainer may still need to complete the repository's final approval gate.

DanielLeens (commented):
Thanks for working on this. I re-reviewed the latest head `1d0963d` in full.

What this PR fixes
- Before this change, a low-volume Zeta streaming job could keep StarRocks rows buffered for a long time when neither `batch_max_rows` nor `batch_max_bytes` was reached. In that case the guaranteed flush path was effectively `prepareCommit()`, so small CDC batches could sit until the next checkpoint.
- This PR wires the engine timer path into the existing StarRocks flush path, so `sink.flush.interval` can push buffered rows out even when the batch thresholds are still below the commit boundary.

Call path I checked
- `seatunnel-connectors-v2/connector-starrocks/.../StarRocksSink.java` now passes `SinkWriter.Context` into `StarRocksSinkWriter`.
- `seatunnel-connectors-v2/connector-starrocks/.../StarRocksSinkWriter.java` registers `context.registerFlushAction(this::timerFlush)` in the constructor.
- `timerFlush()` delegates straight to `manager.flush()`.
- `prepareCommit()` already delegates to the same `manager.flush()` call, so the timer path and checkpoint path share the same write behavior instead of introducing a second data path.

What I validated on the current head
- The normal Zeta path does hit the new code: once the engine emits the timer flush signal, it reaches the writer through `SinkWriter.Context` and lands on the existing `StarRocksSinkManager` flush branch.
- This does not change the sink to exactly-once. The updated EN/ZH docs correctly call out that the connector remains at-least-once and that this scheduled flush is a Zeta-only engine capability.
- Diff integrity looks clean on the current head: the change stays within StarRocks sink code, direct tests, and matching EN/ZH documentation.
- The added unit test covers flush registration and exception propagation, and the new MySQL-CDC-backed E2E test covers the intended low-volume streaming case.

Findings
- I also do not see a diff-integrity issue or a CI signal pointing back to this StarRocks change; the required checks are green.

Conclusion
- Technical verdict from Daniel: `clear_waiting_gate`.
- Since I only have read-level review controls here, a write-capable maintainer may still need to complete the repository's final approval gate.

davidzollo (approved):

PR #11519, opened 2026-07-21
[Feature][Connector-V2] Add ClickHouse timer flush

nzw921rx:
## Purpose

Support timer-driven flushing for the ClickHouse sink on the Zeta engine so buffered rows can be written before `bulk_size` is reached.

## Changes

- Register the existing ClickHouse JDBC batch flush path with `SinkWriter.Context`.
- Reuse the same pending-batch flush logic for timer signals and checkpoint preparation.
- Add unit coverage for timer action registration, batch execution, and failure propagation.
- Add an independent Zeta E2E test using a continuous MySQL CDC source and dynamic container port mapping.
- Resolve MySQL Connector/J from the test classpath and copy it into the MySQL CDC plugin directory.
- Document Zeta timer flush behavior and delivery semantics in English and Chinese.

## Validation

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-clickhouse -DskipIT -DskipUT=false -Dtest=ClickhouseSinkWriterTest -DfailIfNoTests=false test`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-clickhouse-e2e -DskipUT -DskipIT -DskipTests test-compile`


DanielLeens (commented):
Thanks for adding coverage for this. I rechecked the full current diff and do not see a new source-level blocker in the ClickHouse timer-flush change. Registering `context.registerFlushAction(this::flushPendingStatements)` in `ClickhouseSinkWriter` routes the engine timer into the same shard-aware batch flush path that `prepareCommit()` already uses, so small streaming batches no longer have to wait for `bulk_size` to fill before they are written.

The required checks are green on the current head. From Daniel's side the technical verdict is clear; since I only have read-level review controls here, a write-capable maintainer may still need to complete the repository's final approval gate.

dybyte (approved):
+1

PR #10335, opened 2026-01-13
[Feature][Connector-V2][Kafka] Add support for Kafka message header

thehkkim:
### Purpose of this pull request

This PR adds support for Kafka message headers in the Kafka sink connector through a new `kafka_headers_fields` configuration option.

Users can now specify which fields should be set as Kafka message headers, enabling better message routing, filtering, and metadata management in Kafka consumers without parsing the message body.

### Does this PR introduce _any_ user-facing change?

Yes. This PR introduces a new configuration option `kafka_headers_fields` for the Kafka sink connector.

**New Feature:**
- Added `kafka_headers_fields` configuration option to specify which fields should be added as Kafka message headers
- Fields specified as headers are automatically excluded from the message value to avoid data duplication

**Usage Example:**

Configuration:
```hocon
sink {
  kafka {
    topic = "test_topic"
    bootstrap.servers = "localhost:9092"
    format = json
    partition_key_fields = ["id"]
    kafka_headers_fields = ["source", "traceId"]  # New option
  }
}
```

Input data:
```json
{"id": 1, "name": "test", "source": "web", "traceId": "trace-123"}
```

Output to Kafka:
- **Headers**: `source=web`, `traceId=trace-123`
- **Key**: `{"id": 1}`
- **Value**: `{"id": 1, "name": "test"}` (header fields excluded)

**Benefits:**
- Enables header-based routing and filtering in Kafka consumers
- Supports distributed tracing by passing trace IDs via headers
- Allows metadata tagging without parsing message body
- Follows standard Kafka pattern where headers contain metadata

**Backward Compatibility:**
- Fully backward compatible - new optional configuration
- Existing configurations continue to work without any changes

### How was this patch tested?

**1. Unit Tests:**

Added comprehensive test coverage in `DefaultSeaTunnelRowSerializerTest`:

- `testKafkaHeaders()` - Verifies fields are correctly added to Kafka message headers
- `testKafkaHeadersWithNullValue()` - Ensures null values are properly handled (not added to headers)
- `testHeaderFieldsExcludedFromValue()` - Confirms header fields are excluded from message value

**2. Test Results:**
```
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
```

All tests pass successfully, verifying:
- ✅ Headers are correctly populated with specified field values
- ✅ Header fields are automatically excluded from the serialized value
- ✅ Null values are handled properly (skipped in headers)
- ✅ Multiple header fields work correctly
[...]

PR #11521, opened 2026-07-21
[Feature][Connector-V2][Kafka] Support reading Kafka message headers in source connector

thehkkim:

This PR adds `kafka_headers_fields` option to the **Kafka source connector**, allowing users to extract Kafka message header values as row fields.

This is the source-side counterpart of [#10335](https://github.com/apache/seatunnel/pull/10335), which added `kafka_headers_fields` to the Kafka sink connector (merged into `dev`, not yet released). With both features in place, Kafka message headers can now be written via the sink and read back via the source, enabling end-to-end header propagation between topics.

A new optional parameter `kafka_headers_fields` is added to the Kafka source connector.

**Previous behavior:** Kafka message headers were only accessible when using `format = NATIVE`, which exposes all headers as a `Map<String, String>` field. For other formats (JSON, TEXT, etc.), headers were silently discarded.

**New behavior:** Users can specify a list of header key names via `kafka_headers_fields`. Each header value is read as a `STRING` field and appended to the output row after the regular schema fields. If a header key is absent in a record, the corresponding field value is `null`.

**Example:**

```hocon
source {
  Kafka {
    bootstrap.servers = "localhost:9092"
    topic = "my-topic"
    kafka_headers_fields = ["correlation_id", "trace_id"]
    schema = {
      fields {
        user_id = "int"
        name    = "string"
      }
    }
    format = json
  }
}
```

Output row schema: `user_id` (int), `name` (string), `correlation_id` (string), `trace_id` (string).

> Note: this is a user-facing change relative to the unreleased `dev` branch. The sink-side `kafka_headers_fields` (#10335) is also only available in `dev`.

**Constraints (same as sink):**
- Cannot be used with `format = NATIVE` (NATIVE already exposes all headers as a Map)

**Unit tests** (`connector-kafka`):
- `KafkaHeadersDeserializationSchemaTest` (new): 7 cases covering header value appending, null header value, missing header key, options/tableId/rowKind preservation, multi-header fields, and collector path
- `KafkaRecordEmitterTest` (extended): 2 new cases verifying header field injection through the `KafkaEventTime(KafkaHeaders(base))` wrapper chain, and that EVENT_TIME metadata is preserved alongside headers
- `KafkaSourceConfigTest` (extended): 3 new cases verifying schema extension with header fields, correct deserialization schema wrapping order, and that NATIVE format is unaffected

**E2E test** (`connector-kafka-e2e`):
- `KafkaIT.testSourceKafkaWithHeaders`: produces 10 records with JSON value and two headers (`correlation_id`, `trace_id`) to a Kafka topic, runs the SeaTunnel job, and asserts via the Assert sink that all four fields (`user_id`, `name`, `correlation_id`, `trace_id`) are non-null for every row

### Check list

  4. [x] Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)

### DIFF
--- a/docs/en/connectors/source/Kafka.md
+++ b/docs/en/connectors/source/Kafka.md
@@ -63,6 +63,8 @@ They can be downloaded via install-plugin.sh or from the Maven central repositor
 | protobuf_schema                     | String                                                                     | No       | -                        | Effective when the format is set to protobuf, specifies the Schema definition                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
 | strip_schema_registry_header        | Boolean                                                                    | No       | false                    | Effective when the format is set to protobuf. Whether to strip the Confluent Schema Registry wire format header (magic byte, schema id and message indexes) before protobuf deserialization. This option is useful when consuming Protobuf messages that were encoded using Confluent Schema Registry. When enabled, the connector will try to detect and remove the Schema Registry header before parsing the Protobuf message. If the header is not detected, it will fall back to standard Protobuf deserialization.                                                                                                                                                                                                                                                                    |
 | reader_cache_queue_size             | Integer                                                                     | No       | 2                        | The capacity of the fetcher-to-reader element queue. Each element is one `consumer.poll()` batch, not a single message. See [reader_cache_queue_size](#reader_cache_queue_size) for details. |
+| is_native                           | Boolean                                                                     | No       | false                    | Supports retaining the source information of the record.                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
+| kafka_headers_fields                | Array                                                                       | No       | -                        | Specify which Kafka message header keys to extract as row fields. Each header value is read as a STRING type and appended to the output row after the regular schema fields. Cannot be used with NATIVE format.                                                                                                                                                                                                    |
 
 > On restore from checkpoint or savepoint, Kafka Source resumes from the checkpointed split offsets.
 > `start_mode` and consumer-group offsets are only used for the first startup or for newly
@@ -509,7 +511,34 @@ source {
 ```
 
 **Note**: When `strip_schema_registry_header` is enabled, the connector can safely handle both Schema Registry encoded messages and plain Protobuf messages. If the Schema Registry header is not detected, it will automatically fall back to standard Protobuf deserialization.
+
+### Reading Kafka Headers
+
+Use `kafka_headers_fields` to extract specific Kafka message headers as row fields. The header values are appended as STRING type fields after all regular schema fields.
+
+> Note: Cannot be used with `NATIVE` format, which already exposes headers as a `Map<String, String>` field.
+
+```hocon
+source {
+  Kafka {
+    topic = "my-topic"
+    bootstrap.servers = "localhost:9092"
+    kafka_headers_fields = ["correlation-id", "x-trace-id"]
+    schema = {
+      fields {
+        user_id = "int"
+        name = "string"
+      }
+    }
+    format = json
+  }
+}
 ```
+
+The output row will contain: `user_id` (int), `name` (string), `correlation-id` (string), `x-trace-id` (string).  
+If a header key is absent in a record, the corresponding field value will be `null`.
+
+This is the counterpart of `kafka_headers_fields` in the Kafka sink connector, allowing round-trip header propagation between topics.
 
 ### Ignore No Leader Partition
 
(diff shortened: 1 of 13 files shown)


DanielLeens (commented):
Thanks for working on this. I re-reviewed the latest head and the header extraction feature is close, but I still see one source-level blocker before this can merge.

1. `kafka_headers_fields` is currently wired into the declared output schema for every non-`NATIVE` format, but the runtime path for `compatible_kafka_connect_json` never appends those fields. `KafkaSourceConfig.createConsumerMetadata()` extends the catalog table whenever `format != NATIVE`, yet `KafkaSourceConfig.createDeserializationSchema()` returns early for `CompatibleKafkaConnectDeserializationSchema` before the `KafkaHeadersDeserializationSchema` wrapper is applied. Then `KafkaRecordEmitter.emitRecord()` dispatches that format straight to `CompatibleKafkaConnectDeserializationSchema.deserialize(ConsumerRecord, Collector)`, so the produced rows keep the old width while the declared schema already includes the extra header columns. Please either implement header injection for the compatible Kafka Connect path too, or reject `kafka_headers_fields` for that format until it is fully supported.

Separately, the current `Build` failure is in `connector-cdc-base` (`MessageDelayedEventLimiterTest`), not in the Kafka connector itself. But the source-level blocker above still needs to be addressed first. Happy to re-review once you push an update.

thehkkim:
@DanielLeens 
Thank you again for the thorough review — it really helped catch a gap I had missed.              

I went with implementing header injection for the COMPATIBLE_KAFKA_CONNECT_JSON path rather than rejecting it. Here's what changed:

- `ConsumerMetadata`: Added a kafkaHeaderFields field. When `format = compatible_kafka_connect_json` and `kafka_headers_fields` is configured, the header field names are stored here for the emitter to use at runtime.
- `KafkaSourceConfig`: The catalog table is now extended with header fields for                     
`COMPATIBLE_KAFKA_CONNECT_JSON` the same way as other formats. The `CompatibleKafkaConnectDeserializationSchema` itself is still initialized with the base catalog table (before extension), since it handles deserialization independently via deserialize(ConsumerRecord, Collector).
- `KafkaRecordEmitter`: For the `CompatibleKafkaConnectDeserializationSchema` dispatch path, the emitter now checks whether `kafkaHeaderFields` is set on ConsumerMetadata. If so, it wraps the output Collector with a headerInjectingCollector that appends header values to each emitted row after deserialization, keeping the runtime row width consistent with the declared output schema.
- `KafkaSourceConfigTest`: Updated the test to verify that `COMPATIBLE_KAFKA_CONNECT_JSON` + `kafka_headers_fields` correctly extends the output schema and stores the header field names inConsumerMetadata.

Regarding the build failures — as you pointed out, they are in `connector-starrocks` (StarRocksStreamLoadVisitorTest) and `connector-cdc-base`, neither of which is touched by this PR.      

Please take another look when you get a chance!

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `4b937f630c72b7bd676c3989f6a89e99e1462cf3` and traced the exact `compatible_kafka_connect_json + kafka_headers_fields` path again.

The previous source-level blocker from my July 22, 2026 review is closed for me now. `KafkaSourceConfig` now keeps the declared output schema and the runtime path aligned for `compatible_kafka_connect_json` by storing the configured header field names in `ConsumerMetadata`, and `KafkaRecordEmitter` now appends those header values after `CompatibleKafkaConnectDeserializationSchema.deserialize(...)`. I also rechecked the source contract path: Kafka still exposes its output schema through `getProducedCatalogTables()`, so this emitter-side injection does not leave the planner on the old row shape.

At this point I do not see a new code-side blocker from Daniel's side on the current head. The remaining gate is CI/base sync:
- current compare status is `diverged` (`ahead_by=1`, `behind_by=23`) against `dev`
- the failing `Build` signal is currently in `connector-starrocks`, specifically `StarRocksStreamLoadVisitorTest.failsWhenPrepareStateExceedsDeadline`, not in the Kafka connector path touched here

So the lowest-cost next step is to sync the latest `dev` and rerun `Build` on the updated head. If the same lane still fails after the sync, please share the new failing check link and I can help narrow it down from there. Since I do not have write-level review controls in this repository, a write-capable maintainer may still need to perform the final approval step once the gate is clear.

DanielLeens (commented):
Thanks for the update. I re-reviewed the complete current head `de33539e48db45bfb63b6f8af42e3051f0925847`.

The source schema and runtime row shape remain aligned:

```text
KafkaSourceFactory / KafkaSourceConfig
  -> validate non-native header extraction
  -> append configured header fields to produced schema
  -> ConsumerMetadata retains header names
  -> KafkaRecordEmitter deserializes the normal payload
  -> KafkaHeadersDeserializationSchema appends matching header values
```

The compatible Kafka Connect JSON path now receives the same emitter-side header append as the other supported non-native formats; native format is rejected because it already exposes headers through its own record contract. Missing headers become null and duplicate Kafka headers follow the documented last-value lookup behavior. The unit/E2E coverage and bilingual docs match the implementation. I do not see a source-level blocker on this head.

The current Build is green. The remaining gate is one current-head maintainer review.

davidzollo (approved):
+1
LGTM

PR #11522, opened 2026-07-21
[Fix][Docker] Stop unrelocated openGauss driver from hijacking PostgreSQL connections

### LINKED ISSUE #11510: [Bug] [Docker] opengauss-jdbc in image lib/ hijacks org.postgresql.Driver — all PostgreSQL connections fail with "Protocol error. Session setup failed"
### Search before asking

### What happened

The official Docker image `apache/seatunnel:latest` (2.3.13) ships `opengauss-jdbc-5.1.0.jar` in `/opt/seatunnel/lib/`. This jar contains its own copy of the `org.postgresql.Driver` class (openGauss's JDBC driver is a fork of pgJDBC that keeps the original package names). Because both jars sit on the global classpath, the openGauss copy can hijack the class resolution for the real PostgreSQL driver (`postgresql-42.4.3.jar`, also in `lib/`).

Result: **every PostgreSQL connection through the Jdbc connector fails** against a standard PostgreSQL 14 server using scram-sha-256 auth:

```
Caused by: org.postgresql.util.PSQLException: Protocol error.  Session setup failed.
```

### Evidence that this is a jar conflict, not a network/credentials issue

All executed inside the same `apache/seatunnel:latest` container against the same PG 14 server (`postgres:14-alpine`, scram-sha-256):

1. TCP reachable: `bash -c "echo > /dev/tcp/postgres/5432"` — OK
2. Plain JDBC with **only** the bundled pgJDBC on the classpath — OK:
   `java -cp /tmp:/opt/seatunnel/lib/postgresql-42.4.3.jar T` → `rows=3`
3. Same connection through `seatunnel.sh -m local` (full `lib/` classpath) — fails with "Protocol error. Session setup failed."
4. Shadowing the openGauss jar with an empty file fixes it:
   `docker run -v /tmp/empty.jar:/opt/seatunnel/lib/opengauss-jdbc-5.1.0.jar:ro ...` → job FINISHED

Jar scan confirms three jars on the image contain `org/postgresql/core/v3/ConnectionFactoryImpl`:
- `/opt/seatunnel/lib/postgresql-42.4.3.jar` (real driver)
- `/opt/seatunnel/lib/opengauss-jdbc-5.1.0.jar` (conflicting fork)
- `/opt/seatunnel/connectors/connector-cdc-opengauss-2.3.13.jar`

PostgreSQL connections work out of the box with the official image.

Suggested fixes (either):
- move `opengauss-jdbc` out of the global `lib/` and load it only with the openGauss connector's classloader, or
- shade/relocate the `org.postgresql` packages inside the openGauss jar.

### How to reproduce

```bash
docker network create t
docker run -d --name pg --network t -e POSTGRES_USER=u -e POSTGRES_PASSWORD=p -e POSTGRES_DB=d postgres:14-alpine
# minimal config: Jdbc source url jdbc:postgresql://pg:5432/d, user u, password p, query "SELECT 1", Console sink
docker run --rm --network t -v /path/to/job.conf:/j.conf apache/seatunnel:latest \
  /opt/seatunnel/bin/seatunnel.sh --config /j.conf -m local
# -> Protocol error. Session setup failed.
```

### SeaTunnel Version

2.3.13 (apache/seatunnel:latest, digest sha256:2b1d327df210...)

### SeaTunnel Config

```hocon
env { parallelism = 1
  job.mode = "BATCH" }
source { Jdbc {
  url = "jdbc:postgresql://postgres:5432/analytics"
  driver = "org.postgresql.Driver"
  user = "bench"
  password = "..."
  query = "SELECT * FROM customers"
  plugin_output = "j" } }
sink { Console { plugin_input = "j" } }
```

### Running Command

```
/opt/seatunnel/bin/seatunnel.sh --config /j.conf -m local
```

### Error Exception

```
Caused by: org.apache.seatunnel.api.table.factory.FactoryException: ErrorCode:[API-06], ErrorDescription:[Factory initialize failed]
Caused by: org.apache.seatunnel.api.table.catalog.exception.CatalogException: ErrorCode:[API-03], ErrorDescription:[Catalog initialize failed]
Caused by: org.postgresql.util.PSQLException: Protocol error.  Session setup failed.
```

Zeta (local mode)

### Java or Scala Version

openjdk 1.8.0_342 (bundled in image)

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
## Purpose of the change

Fixes #11510.

In the official CI/Docker image, **every PostgreSQL connection fails** with `Protocol error. Session setup failed` because the openGauss JDBC driver shipped in `lib/` hijacks the PostgreSQL driver classes.

## Root cause

- `connector-cdc-opengauss` depends on `org.opengauss:opengauss-jdbc:5.1.0` — the **non `-og` variant**, which is a pgJDBC fork that still bundles 400+ unrelocated `org.postgresql.*` classes, including `org/postgresql/Driver.class` and `org/postgresql/core/v3/ConnectionFactoryImpl.class`.
- The connector shades its own copy of the driver, but with `createDependencyReducedPom=false` the original jar still leaks into the `seatunnel-dist` dependency graph.
- `assembly-seatunnel-bin-ci.xml` includes `org.opengauss:opengauss-jdbc:jar` into `lib/`, and `seatunnel.sh` puts `lib/*` on one flat classpath. The fork jar sorts before the real `postgresql-*.jar` and hijacks `org.postgresql.*` class resolution, so connections to vanilla PostgreSQL run the openGauss-specific protocol handshake and fail.

## Changes

- **`seatunnel-dist/pom.xml`**
  - Exclude the transitive `org.opengauss:opengauss-jdbc` from the `connector-cdc-opengauss` dependency. The CDC connector shades its own relocated driver copy, so it does not need the external jar at runtime.
  - Declare `org.opengauss:opengauss-jdbc:5.1.0-og` (the fully relocated variant, which ships **zero** `org.postgresql.*` classes) so the Jdbc connector's openGauss dialect (`org.opengauss.Driver`) keeps working out of the box in the image.
- **`DriverJarPackagingTest`** (new): guards the distribution against reintroducing the unrelocated fork — asserts `connector-cdc-opengauss` carries the exclusion and any declared `opengauss-jdbc` dependency uses the relocated `-og` variant.

## Verification

- `./mvnw -pl seatunnel-dist dependency:tree` now resolves exactly one openGauss artifact — the safe one:
  ```
  +- org.apache.seatunnel:connector-cdc-opengauss:jar:3.0.0-SNAPSHOT:provided
  +- org.opengauss:opengauss-jdbc:jar:5.1.0-og:provided
  +- org.postgresql:postgresql:jar:42.4.3:provided
  ```
- `DriverJarPackagingTest` passes; mutation check confirmed it fails when the exclusion is removed.

## Notes

- No behavior change for PostgreSQL users except that the hijack is gone; openGauss support via the Jdbc connector is preserved through the relocated driver.
- The plain ASF bin assembly (`assembly-seatunnel-bin.xml`) ships no JDBC drivers and is unaffected.


DanielLeens (commented):
Thanks for tracking this down. I went through the full diff on the current head and do not see a new source-level blocker in the packaging fix itself. Excluding the unrelocated `org.opengauss:opengauss-jdbc` from `connector-cdc-opengauss` while adding back the `-og` relocated variant in `seatunnel-dist` is the right direction, and the new packaging test protects both sides of that contract.

The current `Build` signal was cancelled before it produced a usable result, and this branch is also behind the latest `dev` (`compare status: diverged`, `behind_by=18`). The lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the rerun still fails after the sync, please share the new failing check link and I can help narrow it down further.

DanielLeens (commented):
This is a fresh review of the current head (`d1a0b607`). Good news up front: the only thing that changed since my last pass on 2026-07-22 is exactly the sync-with-`dev` I suggested back then — the new commit is a plain `Merge remote-tracking branch 'upstream/dev'` with zero source changes on top of it. I re-derived the full review from scratch against the current head anyway (per policy, a new commit means a new review, not a rubber stamp), and the technical conclusion is unchanged: the packaging fix itself is correct, minimal, and well-guarded by a new test. I went further this time and traced the actual dependency-resolution and assembly wiring end-to-end (see 1.1) to close out the "why is this the right jar" question completely rather than just trusting the PR description.

The official CI/Docker distribution ships `org.opengauss:opengauss-jdbc:5.1.0` (the plain, non-`-og` variant) into `lib/`. That artifact is a pgJDBC fork that still bundles ~400 unrelocated `org.postgresql.*` classes, including `org/postgresql/Driver.class`. Since `seatunnel.sh` puts everything in `lib/*` on one flat classpath, and the fork jar sorts ahead of the real `postgresql-*.jar` in class loading, every plain PostgreSQL connection gets hijacked into the openGauss-specific handshake and fails with `Protocol error. Session setup failed` (issue #11510). The fix stops that jar from reaching `lib/` while keeping openGauss connectivity working through a driver variant that doesn't carry the conflicting classes.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Diff is exactly two files, unchanged from the version I reviewed on 07-22:
- `seatunnel-dist/pom.xml` (+25 lines)
- new `seatunnel-dist/src/test/java/org/apache/seatunnel/packaging/DriverJarPackagingTest.java` (+175 lines)

**Root cause, verified at the source level (not just taken from the PR description):**
`seatunnel-connectors-v2/connector-cdc/connector-cdc-opengauss/pom.xml` declares a normal (non-excluded) dependency on `org.opengauss:opengauss-jdbc:5.1.0` and shades `org.postgresql -> ${seatunnel.shade.package}.org.postgresql` into its *own* jar, with `createDependencyReducedPom=false`. Because the reduced pom is disabled, the connector's published pom still lists the raw `opengauss-jdbc:5.1.0` as a transitive dependency even though the connector's own jar no longer needs it (its own copy is already shaded in). `seatunnel-dist/pom.xml` declares `connector-cdc-opengauss` as a `provided` dependency (line 667, in the `seatunnel` profile that backs `assembly-seatunnel-bin-ci.xml`), and `assembly-seatunnel-bin-ci.xml`'s JDBC-driver `dependencySet` (lines 191-235) resolves `useTransitiveDependencies=true` at `scope=provided` with a version-agnostic include `org.opengauss:opengauss-jdbc:jar`. That's the exact path that pulls the raw, unrelocated 5.1.0 jar into `/lib`.

**Fix, and why the replacement version is trustworthy:**
- Adds `<exclusions>` for `org.opengauss:opengauss-jdbc` on the `connector-cdc-opengauss` dependency in `seatunnel-dist/pom.xml` (lines 667-681) — correctly scoped to the dist module's own resolution, doesn't touch the connector's own pom/build (which still needs the full 5.1.0 driver at package time to shade from).
- Declares a new direct `provided` dependency on `org.opengauss:opengauss-jdbc:${opengauss.jdbc.version}` = `5.1.0-og` (lines 828-837).
- I checked whether `5.1.0-og` is a credible, already-vetted artifact rather than something invented for this PR: `seatunnel-connectors-v2/connector-jdbc/pom.xml` (lines 56, 221-223, 378-381) **already** depends on `opengauss-jdbc:5.1.0-og` at `scope=provided` for the Jdbc connector's openGauss dialect (`OpenGaussDialectFactory.acceptsURL` gates on `jdbc:opengauss:`, extending `PostgresDialectFactory`/`PostgresDialect`). This PR is not introducing a new artifact — it's reusing the exact version+scope the JDBC connector module has relied on all along. That's strong evidence the `-og` variant is real and functional, not a guess.
- Assembly wiring needs no change: the `/lib` include (`org.opengauss:opengauss-jdbc:jar`, no version pin) still matches the new `-og` artifact. The separate `/connectors` `dependencySet` only matches `org.apache.seatunnel:connector-*` (different groupId), so it was never part of the leak and isn't touched.
- Scope check: the `release` profile (ASF source-release build, `assembly-seatunnel-bin.xml`) has its own, much smaller `<dependencies>` list with zero JDBC drivers and no `opengauss`/`connector-cdc-opengauss` entries at all — confirmed by grep, matching the PR's claim that the plain ASF release assembly is unaffected. The fix is correctly confined to the `seatunnel` (CI/Docker) profile where the bug actually lives.

Runtime path: this is a build/packaging-time dependency-graph change only. It doesn't touch checkpoint, serialization, source/sink lifecycle, or any request-time code path — normal users are affected only via which jars land in the distribution.

## 1.2 Compatibility Impact
**Fully compatible.** No API, config `Option`, protocol, or serialization surface is touched. Effect is scoped to which jars ship in `lib/` for the CI/Docker image; the ASF source-release assembly is untouched (verified above). openGauss connectivity via the Jdbc connector's dialect keeps working because it already depended on the same `-og` artifact/version before this PR.

None. Pure dependency-tree edit; no new runtime allocation, thread, or I/O path. The design correctly keeps the connector module's build-time need (full `5.1.0` driver, to shade `org.postgresql.*` into its own namespace) separate from the dist module's runtime need (a driver jar that must not carry `org.postgresql.*` at all) — that separation is exactly right and is why the fix lives entirely in `seatunnel-dist/pom.xml` rather than touching the connector.

Not applicable — no exception or logging paths are touched. No blocking issues found in this category.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Both new pom comments and the test class Javadoc explain *why* (cite issue #11510, explain the hijack mechanism), not just what — good adherence to the project's comment expectations for non-obvious build logic. License header present on the new test file.

One Low-severity nitpick on the new test's parsing approach (see Issue 1 below); it doesn't affect current correctness, only long-term robustness.

`DriverJarPackagingTest` adds two deterministic unit tests:
1. `testCdcOpenGaussDoesNotLeakUnrelocatedDriver` — asserts the exclusion is present on `connector-cdc-opengauss`.
2. `testOnlyRelocatedOpenGaussDriverIsDeclared` — asserts every declared `opengauss-jdbc` version ends in `-og`.

Both read `seatunnel-dist/pom.xml` directly (with a same-module/root-relative fallback that matches Maven surefire's default `${project.basedir}` working directory) and use regex against the pom text rather than an XML parser — the class Javadoc explains this is deliberate, to avoid classpath conflicts with connector-bundled XML parser implementations on the `seatunnel-dist` test classpath. I manually traced the regex logic against the actual current pom structure (single `connector-cdc-opengauss` entry, single `opengauss-jdbc` re-declaration, `groupId` always the first child tag) and confirmed both tests correctly target the intended blocks and would fail if either half of the fix were reverted (matches the PR's stated "mutation check" claim). No `Thread.sleep`, no shared/static state, no ordering dependence, no network or container use — this is a plain file-read + regex test.

**Stability rating: Stable.**

## 2.3 Documentation Updates
Not applicable — no user-facing config, default, or behavior changed; this is an internal packaging correctness fix invisible to end users (as the PR description itself notes).

# 3. Architectural Soundness

**Precise fix.** It targets the actual transitive-resolution leak at its exact source (the dist module's own dependency declaration), reuses an already-proven artifact version instead of introducing anything new, and adds a regression guard. Nothing broader was touched than necessary.

## 3.2 Maintainability
Good — the `why` is documented inline at both edit sites, and the guard test will fail loudly if either half of the fix (the exclusion or the `-og` pin) is accidentally reverted in the future.

## 3.3 Extensibility
Not a concern for a targeted packaging bugfix of this scope.

Not applicable — this only affects how a *future* build/image is assembled; it does not touch any already-published artifact, checkpoint format, or protocol, and there is no upgrade/restore path involved.

# 4. Issue Summary

| 1 | `DriverJarPackagingTest`'s tag extraction (`extractTag`) assumes `<groupId>` always appears as the *first* matching tag ahead of any nested tags in a block (used to disambiguate the outer `connector-cdc-opengauss` dependency from its nested `<exclusion>`). Correct for the current pom's formatting, but a future reordering or an additional `<dependency>`/`<exclusion>` occurrence for the same coordinates elsewhere in the file could silently make the test pass without checking what it's meant to check, rather than failing loudly. | `seatunnel-dist/src/test/java/org/apache/seatunnel/packaging/DriverJarPackagingTest.java:150-165` (`extractTag`) | Low | No |

No blocking issues found. The core fix is correct, minimal, and verified end-to-end against the real dependency-resolution and assembly-packaging mechanism, not just the PR's own description.

# 5. Merge Recommendation

### Conclusion: Ready to merge after CI passes
1. Blockers — none from source review.
2. Recommended fixes — non-blocking: consider hardening `extractTag`/`findDependencyBlocks` against tag-order/duplication edge cases if this test pattern gets reused elsewhere in `seatunnel-dist` (Issue 1, Low).

Overall assessment: this is the same correct, well-scoped fix I reviewed on 07-22, now rebased on current `dev` as recommended — no code drift, no new risk introduced by the sync. The only remaining gate is CI (the `Build` check on this head is still `QUEUED` as of this review, both on the apache-side pointer and the underlying run in the contributor's fork). No alternative approach needed; Option A (this PR's exclude-and-repin approach) is the right one — there's no meaningfully simpler way to keep the connector's shading logic and the dist's clean-jar requirement both satisfied.

DanielLeens (approved):
Formalizing this as an approval. My previous re-review of this exact head (`d1a0b607`) found no blockers from source review and concluded "ready to merge after CI passes," with the `Build` check still `QUEUED` at that time. Rechecking now: `Build`, `Notify test workflow`, and `labeler` have all since completed `SUCCESS` on this same head. Since CI completion was the only thing my previous review was waiting on, I'm updating this to an approval. The one open item (hardening `DriverJarPackagingTest`'s tag-extraction logic against future pom reordering) remains a low-severity, non-blocking suggestion.

davidzollo (approved):
+1
LGTM

PR #11523, opened 2026-07-21
[Fix][Transform-V2] Preserve exact binary value when CASTing floats to DECIMAL

### LINKED ISSUE #10198: [Bug] Synchronizing data from MySQL 5.7 to PostgreSQL 14 may cause precision loss.
### Search before asking

### What happened

数据源值为：126.752251	44.916103
同步到pg后为： 126.752250	44.916103

源值："Longitude":126.752251
"Latitude":44.916103，生成的插入语句为：INSERT INTO "map"."public"."t_station" ("id", "station_code", "station_name", "province_id", "city_id", "county_id", "address", "longitude", "latitude", "create_time", "status", "update_time", "area_id", "area_name", "source_table") VALUES (('1'::int4), ('....'), ('....'), ('43'::int4), ('4309'::int4), ('430902'::int4),
('....'), ('126.75225000'::numeric), ('44.91610300'::numeric) 数据是有差异

### SeaTunnel Version

v2.3.12

### SeaTunnel Config

```conf
env {
  execution.parallelism = 1
  job.name = "map_station"
  job.mode = "STREAMING"
  # checkpoint配置优化
  checkpoint.interval = 120000
  checkpoint.timeout = 1800000
  checkpoint.min.pause.between = 60000
  checkpoint.max.concurrent = 1
  checkpoint.tolerable.failed.num = 3
  # 引擎配置优化
  seatunnel.engine.checkpoint.interval = 120000
  seatunnel.engine.checkpoint.timeout = 1800000
}

source {
  MySQL-CDC {
    plugin_output = "mysql_cdc_source"
    url = "jdbc:mysql://127.0.0.1:3306/logistics?serverTimezone=Asia/Shanghai&useUnicode=true&characterEncoding=UTF-8&rewriteBatchedStatements=true"
    port = 3306
    username = "root"
    password = "password"
    table-names = ["logistics.t_station"]
    # 心跳检测配置
    heartbeat-interval = 30000
    # 连接超时配置
    connect-timeout = 120000
    socket-timeout = 240000 
    # CDC配置
    server-id = 12720048
    startup.mode = "initial"
    # 连接配置
    connect.timeout = "30s"
    connection.pool.size = 10
    # 增量同步配置
    debezium.snapshot.mode = "initial"
    debezium.snapshot.locking.mode = "minimal"
    # 重试配置
    max-retries = 5
    retry-backoff = 2000
     debezium.decimal.handling.mode = "deciaml"
     debezium.float.handling.mode = "float"
     debezium.double.handling.mode = "double"
     debezium.time.precision.mode = "adaptive"
     debezium.database.history.skip.unparseable.ddl = true
     debezium.value.converter.encoding = "UTF-8"
     debezium.key.converter.encoding = "UTF-8"
  }
}

transform {
  Sql {
    source_table_name = "mysql_cdc_source"
    result_table_name = "t_station_transformed"
    query = """
      SELECT 
        `Id` as id,
        `Code` as code,
        `Name` as name,
        `ProvinceId` as province_id,
        `CityId` as city_id,
        `RegionId` as county_id,
        `Address` as address,
        CASE 
          WHEN Longitude IS NOT NULL THEN
            ROUND(CAST(Longitude  AS DECIMAL(12,6)), 6)
            
          ELSE 
            NULL 
        END as longitude,
        CASE 
          WHEN `Latitude` IS NOT NULL THEN 
            CAST(CAST(`Latitude` AS STRING) AS DECIMAL(12,6))
          ELSE NULL 
        END as latitude,
        `CreateTime` as create_time,
        `Status` as status,
        `ModifyTime` as update_time,
        `area` as area,
        `areaName` as area_name
      FROM tb_station_source
    """
  }
  
}

sink {
  jdbc {
    plugin_input ="t_station_transformed"
    url = "jdbc:postgresql://127.0.0.1:5432/map?connectTimeout=60&socketTimeout=600"
    # HikariCP连接池核心配置    
    connection_check_timeout_sec = 30
    transaction_timeout_sec=600
    auto_commit=true
    column.type.mapping = "longitude=NUMERIC(10,6),latitude=NUMERIC(10,6)"
    # 禁用自动类��转换，确保精度不丢失
    disable.type.convert = false
    string_column_to_numeric = true
    properties {
      connectionTimeout=3000
      idleTimeout=600000
      keepaliveTime=true
      maxLifetime=1800000
      connectionTestQuery="select 1"
      minimumIdle=1
      maximumPoolSize=5
      initializationFailTimeout=1
      validationTimeout=5000
      leakDetectionThreshold=2000
      idle_timeout_ms = 600000
      max_lifetime_ms = 1800000
      validation_timeout_ms = 5000
      leak_detection_threshold_ms = 60000
      connection_pool_size        = 5
      connection_validation_query = "SELECT 1"
      reWriteBatchedInserts = true
      socket_timeout = 300
      login_timeout = 30
      tcp_keep_alive = true
    }
    driver = "org.postgresql.Driver"
    username = "test"
    password = "password"
    database="map"
    table = "public.t_station"
    generate_sink_sql = true    
    # 主键配置，用于upsert操作
    primary_keys = ["id"]
    support_upsert_by_insert_only = false
    write_mode = "upsert"    
    # 批量写入优化
    batch_size = 1000
    batch_interval_ms = 1000
    # 会话重试机制
    max_retries = 5
    retry_backoff_multiplier_ms = 2
    retry_backoff_initial_ms = 1000
    retry_backoff_max_ms = 30000
    # 写入超时配置
    write_timeout = 120000
    # 事务配置
    transaction.timeout = 60000
 }
}
```

### Running Command

```shell
bin/seatunnel.sh --config
```

### Error Exception

```log
没有异常
```

zeta模式

### Java or Scala Version

1.8

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
## Purpose of the change

Fixes #10198.

Syncing MySQL `FLOAT` columns (e.g. GPS coordinates) through the SQL transform silently loses precision: a source value of `126.752251` (`float(10,6)`) lands in a PostgreSQL `numeric(10,6)` column as `126.752250`.

## Root cause

MySQL stores `FLOAT` as an IEEE-754 binary value; the value behind `126.752251` is exactly `126.75225067138671875`. MySQL's own `CAST(col AS DECIMAL(20,10))` converts that **exact binary value** and returns `126.7522506714` (which rounds to `126.752251` at 6 digits).

The Zeta SQL function `SystemFunction.castAs` did `new BigDecimal(v1.toString())`. `Float.toString()` returns the shortest round-trip string `"126.75225"`, discarding the hidden binary digits **before** scaling, so SeaTunnel produced `126.7522500000` — off by one ULP at 6 digits, exactly the discrepancy reported in the issue.

## Changes

- **`SystemFunction.castAs` (DECIMAL branch)**
  - `Float`/`Double` inputs are now converted with `new BigDecimal(floatValue/doubleValue)`, i.e. from the exact binary value, matching database CAST semantics.
  - `BigDecimal` inputs pass through without re-parsing.
  - All other input types keep the previous string-based conversion, so behavior is unchanged for them.
- **`SystemFunctionTest#testCastAsDecimalKeepsExactBinaryValueOfFloats`** (new)
  - `CAST(126.752251f AS DECIMAL(20,10)) == 126.7522506714` (equals MySQL's result; previously `126.7522500000`)
  - `CAST(126.752251f AS DECIMAL(10,6)) == 126.752251` (previously `126.752250`)
  - `CAST(44.916103d AS DECIMAL(20,10)) == 44.9161030000`
  - BigDecimal pass-through case.

## Verification

- `./mvnw -pl seatunnel-transforms-v2 test` — full module test suite passes (including `SystemFunctionTest`, `CastFunctionTest`, `CastFunctionTypeTest`).

## Notes

- The rounding mode (`CEILING`) is intentionally left untouched to keep this change minimal; this PR only fixes the lossy string round-trip reported in the issue.


DanielLeens (commented):
Thanks for the detailed reproduction here. I went through the full diff on the current head and the fix direction makes sense: switching the DECIMAL cast path away from `Float.toString()` / `Double.toString()` is what preserves the exact binary value instead of the shortest round-trip text form, and the added test cases cover the key float / double / BigDecimal cases.

From the current CI evidence, I do not see a failing signal that is directly tied to this transform change yet. The failing lanes are currently `Run / Dead links`, `Run / unit-test (11, ubuntu-latest)`, `Run / all-connectors-it-4 (8, ubuntu-latest)`, and `Run / kafka-connector-it (8, ubuntu-latest)`, and the unit-test failure is in `connector-starrocks` (`StarRocksStreamLoadVisitorTest.failsWhenPrepareStateExceedsDeadline`), not in the SQL transform module touched by this PR. This branch is also behind the latest `dev` (`compare status: diverged`, `behind_by=18`), so the lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the same lanes still fail after the sync, please paste the refreshed check link and I can help narrow down what remains.

DanielLeens (commented):
# Basic information

- PR title: [Fix][Transform-V2] Preserve exact binary value when CASTing floats to DECIMAL
- Author: zhangshenghang
- Link: https://github.com/apache/seatunnel/pull/11523
- Base branch: dev
- State: OPEN
- Change size: +56 / -1, 2 file(s)
- Human review ready: No
- Review time: 2026-07-26T16:44:17Z
- Local branch: seatunnel-review-11523
- Local worktree: shared bare PR 11523 review store

## Stable review keys

- pr_number: `11523`
- base_sha: `ebae5c10c3c2abacd8a8d872c6e5d590d353c322`
- reviewed_base_sha: `ebae5c10c3c2abacd8a8d872c6e5d590d353c322`
- merge_base_sha: `46fefa7c6117290bd248ae59398d283bb9918f82`
- head_sha: `22f9a905dab8f3ff290de08b72b89d2b124dbe7e`
- reviewed_head_sha: `22f9a905dab8f3ff290de08b72b89d2b124dbe7e`
- human_review_ready: `false`
- diff_integrity: `pass`
- deletion_total: `1`
- deletion_classified: `1`
- deletion_unclassified: `0`
- other_total: `0`
- other_confirmed: `0`
- other_unresolved: `0`
- reviewed_at: `2026-07-26T16:44:17Z`

- mergeStateStatus: `BLOCKED`
- mergeable_state: `blocked`
- mergeable: `True`
- rebaseable: `False`
- required_ci_gate: `failing`
- approval_gate: `pending`
- conflict_gate: `satisfied`
- merge_queue_gate: `pending`
- compare: status=`diverged`, ahead_by=`1`, behind_by=`42`, total_commits=`1`, url=https://github.com/apache/seatunnel/compare/dev...22f9a905dab8f3ff290de08b72b89d2b124dbe7e

# What problem this PR solves

- User pain: Float.toString() drops hidden IEEE digits, causing some MySQL FLOAT to DECIMAL values to lose one unit after scaling.
- Fix approach: Construct exact BigDecimal values directly from Float and Double, retain BigDecimal inputs, and preserve string conversion for other inputs.
- One-line summary: Exposing the binary value is directionally correct, but combining it with existing CEILING silently increments common positive values such as 0.1 by one target-scale unit.
- Before: 126.752251f becomes 126.75225 through Float.toString and then DECIMAL(10,6) becomes 126.752250.
- After: The issue coordinate becomes 126.752251, but the exact tail of 0.1d plus CEILING incorrectly becomes 0.11 in DECIMAL(10,2).

# 1. Code-change review

- Behavior before: 126.752251f becomes 126.75225 through Float.toString and then DECIMAL(10,6) becomes 126.752250.
- Behavior after: The issue coordinate becomes 126.752251, but the exact tail of 0.1d plus CEILING incorrectly becomes 0.11 in DECIMAL(10,2).
- Normal-path hit: `conditional` — Direct CAST(float AS DECIMAL), CASE result casts, and DecimalType helpers in ZetaSQLEngine projection call this branch.

- `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunction.java:235-254` `SystemFunction.castAs DECIMAL branch` — Zeta CAST converts Float and Double to full exact BigDecimal values and then scales with CEILING.
- `seatunnel-transforms-v2/src/test/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunctionTest.java:216-251` `exact floating DECIMAL tests` — Helper-level tests cover issue-specific values and BigDecimal pass-through but omit 0.1, sign, and rounding boundaries.

### Effective prerequisites

- The input is Float or Double and the cast target is DECIMAL with a scale.

- String and integer paths preserve old conversion, BigDecimal only skips reparsing, and NaN or infinity still fails.

### Complete execution chain

```text
1. seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunction.java:235-254 -> SystemFunction.castAs DECIMAL branch [method_call]
   Zeta CAST converts Float and Double to full exact BigDecimal values and then scales with CEILING.
   Side effect: Zeta CAST converts Float and Double to full exact BigDecimal values and then scales with CEILING.
2. seatunnel-transforms-v2/src/test/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunctionTest.java:216-251 -> exact floating DECIMAL tests [method_call]
   Helper-level tests cover issue-specific values and BigDecimal pass-through but omit 0.1, sign, and rounding boundaries.
   Side effect: Helper-level tests cover issue-specific values and BigDecimal pass-through but omit 0.1, sign, and rounding boundaries.
```

- Recovery, exception, async, serialization, and concurrency paths: All objects are method-local. Deterministic conversion changes no shared state, lock, async, resource lifecycle, or checkpoint behavior; retries repeat the same wrong value.

- Applicable: `false`

## 1.2 Compatibility impact

- Classification: `partially_incompatible` (partially incompatible)
- API: There is no API, SPI, configuration, state, or protocol change, but every Float or Double to DECIMAL result may change; the issue value is corrected while common positive tails regress.
- SPI: There is no API, SPI, configuration, state, or protocol change, but every Float or Double to DECIMAL result may change; the issue value is corrected while common positive tails regress.
- Configuration: There is no API, SPI, configuration, state, or protocol change, but every Float or Double to DECIMAL result may change; the issue value is corrected while common positive tails regress.
- Defaults: There is no API, SPI, configuration, state, or protocol change, but every Float or Double to DECIMAL result may change; the issue value is corrected while common positive tails regress.
- Protocol and serialization: There is no API, SPI, configuration, state, or protocol change, but every Float or Double to DECIMAL result may change; the issue value is corrected while common positive tails regress.
- Checkpoint/restore: There is no API, SPI, configuration, state, or protocol change, but every Float or Double to DECIMAL result may change; the issue value is corrected while common positive tails regress.
- Historical behavior: There is no API, SPI, configuration, state, or protocol change, but every Float or Double to DECIMAL result may change; the issue value is corrected while common positive tails regress.

- CPU: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Memory: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- GC: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Network: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Lock contention: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Concurrency safety: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Retry: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Idempotency: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Resource release: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.
- Overall: Exact BigDecimal may allocate a longer coefficient, but it is immediately scaled and bounded per field; correctness dominates the CPU and memory difference, with no network or lock impact.

- Exception propagation: No exception is swallowed and no sensitive value is logged; the regression is more dangerous because it succeeds with the wrong numeric value.
- Context: No exception is swallowed and no sensitive value is logged; the regression is more dangerous because it succeeds with the wrong numeric value.
- Log levels: No exception is swallowed and no sensitive value is logged; the regression is more dangerous because it succeeds with the wrong numeric value.
- Sensitive information: No exception is swallowed and no sensitive value is logged; the regression is more dangerous because it succeeds with the wrong numeric value.
- Overall: No exception is swallowed and no sensitive value is logged; the regression is more dangerous because it succeeds with the wrong numeric value.

- diff_integrity: `pass`
- deletion_ledger: total=`1`, classified=`1`, unclassified=`0`
- Reconciliation: 1 = 1 + 0

- `deletion-001` | count=`1` | `deletion` | `intended_change` | `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunction.java` (old line 236) — The only deleted line is the old generic new BigDecimal(v1.toString()), replaced by the type-aware branch; 1=1+0.

- Out-of-scope hunk conclusion: Both files belong to SQL CAST implementation and tests, with no removed option, documentation, capability, or unrelated hunk.
- Stale-base conclusion: The branch is one ahead and eleven behind the reviewed base, whose base-only commits do not touch the two changed files; it is one ahead and forty-two behind live dev.
- Merge-artifact conclusion: Merge-tree has no conflict candidate, git diff --check is clean, and there is no file deletion, rename, or merge artifact.

# 2. Code-quality assessment

- Title format: Both files preserve ASF headers, imports, and formatting; the implementation comment compares exact conversion with database semantics without defining the rounding contract.
- Code conventions: Both files preserve ASF headers, imports, and formatting; the implementation comment compares exact conversion with database semantics without defining the rounding contract.
- Comment completeness: Both files preserve ASF headers, imports, and formatting; the implementation comment compares exact conversion with database semantics without defining the rounding contract.

- Unit tests: `covered` — Deterministic helper tests cover issue Float scales, one Double, and BigDecimal pass-through; they omit 0.1f and 0.1d, negative symmetry, scale boundaries, and parsed SQL CAST.
- Component tests: `missing` — Deterministic helper tests cover issue Float scales, one Double, and BigDecimal pass-through; they omit 0.1f and 0.1d, negative symmetry, scale boundaries, and parsed SQL CAST.
- Engine integration tests: `not_applicable` — Deterministic helper tests cover issue Float scales, one Double, and BigDecimal pass-through; they omit 0.1f and 0.1d, negative symmetry, scale boundaries, and parsed SQL CAST.
- E2E: `missing` — Deterministic helper tests cover issue Float scales, one Double, and BigDecimal pass-through; they omit 0.1f and 0.1d, negative symmetry, scale boundaries, and parsed SQL CAST.
- Regression tests: `missing` — Deterministic helper tests cover issue Float scales, one Double, and BigDecimal pass-through; they omit 0.1f and 0.1d, negative symmetry, scale boundaries, and parsed SQL CAST.

### Test-stability evidence

- `seatunnel-transforms-v2/src/test/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunctionTest.java:216-251`
  - Trigger: The added tests use no sleep, clock, network, container, port, global mutation, filesystem ordering, or nondeterministic collection.
  - Probability basis: The added tests use no sleep, clock, network, container, port, global mutation, filesystem ordering, or nondeterministic collection.
  - Analysis: The added tests use no sleep, clock, network, container, port, global mutation, filesystem ordering, or nondeterministic collection.

### Missing scenarios

- Add 0.1f and 0.1d, positive and negative, below-half, half, above-half, reported-value, and ZetaSQLEngine parsed CAST regressions.

## 2.3 Documentation

- User-visible change: `true`
- docs/en: EN/ZH SQL-function docs list DECIMAL(p,s) without defining rounding; a user-visible numeric-semantics change needs a synchronized contract and migration guidance.
- docs/zh: EN/ZH SQL-function docs list DECIMAL(p,s) without defining rounding; a user-visible numeric-semantics change needs a synchronized contract and migration guidance.
- Examples: EN/ZH SQL-function docs list DECIMAL(p,s) without defining rounding; a user-visible numeric-semantics change needs a synchronized contract and migration guidance.
- Configuration truth: EN/ZH SQL-function docs list DECIMAL(p,s) without defining rounding; a user-visible numeric-semantics change needs a synchronized contract and migration guidance.
- URL truth: EN/ZH SQL-function docs list DECIMAL(p,s) without defining rounding; a user-visible numeric-semantics change needs a synchronized contract and migration guidance.

### Local execution boundary

- `git diff --stat 46fefa7c6117290bd248ae59398d283bb9918f82...22f9a905dab8f3ff290de08b72b89d2b124dbe7e`
  - Result: `passed`
  - Explanation: Confirmed 2 files and +56/-1.
- `git diff --name-status 46fefa7c6117290bd248ae59398d283bb9918f82...22f9a905dab8f3ff290de08b72b89d2b124dbe7e`
  - Result: `passed`
- `git diff --unified=0 46fefa7c6117290bd248ae59398d283bb9918f82...22f9a905dab8f3ff290de08b72b89d2b124dbe7e`
  - Result: `passed`
- `git diff --check`
  - Result: `passed`

### Required checks

- `Build` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`FAILURE` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/88668487911

- `Notify test workflow` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/29840671415/job/88668450305
- `labeler` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/29840668601/job/88668440027
- `Build` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/88668487911

# 3. Architecture assessment

- Type: `workaround`
- Assessment: SystemFunction.castAs is the correct central boundary, but conversion and rounding must be defined together; changing only input construction is an incomplete fix.

## 3.2 Maintainability

SystemFunction.castAs is the correct central boundary, but conversion and rounding must be defined together; changing only input construction is an incomplete fix.

## 3.3 Extensibility

SystemFunction.castAs is the correct central boundary, but conversion and rounding must be defined together; changing only input construction is an incomplete fix.

SystemFunction.castAs is the correct central boundary, but conversion and rounding must be defined together; changing only input construction is an incomplete fix.

# 5. Formal findings

## `问题 1` — Exact IEEE tails combined with CEILING corrupt ordinary DECIMAL casts

- Location: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunction.java:235-254`
- Problem and real execution path: Float and Double conversion changes from canonical decimal strings to full exact binary expansion while retaining CEILING; any positive hidden tail pushes the result up by one target-scale unit. MySQL FLOAT or Double row -> Zeta SQL CAST -> SystemFunction.castAs -> new BigDecimal(float or double) -> setScale(scale, CEILING) -> JDBC sink.
- Current-head evidence: The exact value of 0.1d is 0.10000000000000000555..., so scale 2 plus CEILING yields 0.11; the old toString path yielded 0.10. New tests omit 0.1, negative symmetry, and rounding boundaries.
- Trigger condition: A Float or Double exact binary value is slightly above its human-readable decimal and digits are discarded at the target scale; 0.1 is a common example.
- Potential impact: SQL transform silently writes money, measurements, coordinates, and similar values one declared-scale unit high, with sign-asymmetric behavior.
- Best actionable fix: Keep exact construction for Float and Double but use an explicit round-to-nearest contract, such as HALF_UP when matching MySQL. Preserve the legacy String and BigDecimal CEILING branch separately, and add 0.1, sign, below-half, half, above-half, parsed CAST tests, and EN/ZH documentation.
- Severity: `high`
- Merge gate: `blocking`

# 6. Findings summary

|---|---|---|---|---|
| 问题 1 | Exact IEEE tails combined with CEILING corrupt ordinary DECIMAL casts | seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunction.java:235-254 | high | blocking |

# 7. Merge conclusion

### Blocking items

- `floating-ceiling-corruption` [source] (问题 1): Exact IEEE expansion plus CEILING corrupts common values such as 0.1.
- `rounding-contract-docs` [documentation] (问题 1): The same numeric blocker also lacks an EN/ZH rounding contract and migration guidance.
- `required-build-failed` [tests]: The current-head required Build is failing.
- `maintainer-review-pending` [other]: A write-capable maintainer review is still pending.

### Overall assessment

SystemFunction.castAs is the correct central boundary, but conversion and rounding must be defined together; changing only input construction is an incomplete fix.

- Alternative: Use exact construction plus round-to-nearest for Float and Double while temporarily preserving CEILING for legacy String and BigDecimal inputs; normalize the broader SQL contract later through an explicit incompatible change.
- Trade-offs: Use exact construction plus round-to-nearest for Float and Double while temporarily preserving CEILING for legacy String and BigDecimal inputs; normalize the broader SQL contract later through an explicit incompatible change.

- Reviewed commit: `22f9a905dab8f3ff290de08b72b89d2b124dbe7e`

DanielLeens (approved):
Thanks for pushing the follow-up commit so quickly, @zhangshenghang — I re-read the full current diff against `dev` (not just the incremental commit) before writing this, since a lot can shift once a second commit lands.

- User pain point: casting a `FLOAT`/`DOUBLE` value to `DECIMAL` in a Zeta SQL transform went through `Float.toString()`/`Double.toString()`, which returns the *shortest round-trip* decimal text and silently discards the hidden IEEE-754 binary tail. For a MySQL `FLOAT(10,6)` column holding `126.752251`, the real binary value is `126.75225067138671875`, but the old code produced `126.7522500000` instead of the correct `126.7522506714` — one ULP off from what MySQL's own `CAST(... AS DECIMAL(20,10))` returns, which is exactly the discrepancy reported in #10198.
- Fix approach: build the `BigDecimal` directly from the exact binary value (`new BigDecimal(floatValue)` / `new BigDecimal(doubleValue)`) instead of round-tripping through `toString()`, for `Float`/`Double` inputs only; `BigDecimal` inputs pass through unchanged, and all other types keep the pre-existing string-based path.
- One-sentence summary: this now correctly reproduces database-style exact-binary `CAST` semantics for floating-point-to-`DECIMAL` conversions, and — since my last review — the rounding mode for that exact-binary path has also been fixed so it no longer silently corrupts ordinary values like `0.1`.

# 1. Code Change Review
## 1.1 Core Logic Analysis

Changed file: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunction.java`, `castAs()`, `DECIMAL` case.

Before this PR:
```java
case "DECIMAL":
    BigDecimal bigDecimal = new BigDecimal(v1.toString());
    Integer scale = (Integer) args.get(3);
    return bigDecimal.setScale(scale, RoundingMode.CEILING);
```

Current head (after both commits):
```java
case "DECIMAL":
    BigDecimal bigDecimal;
    RoundingMode roundingMode;
    if (v1 instanceof BigDecimal) {
        bigDecimal = (BigDecimal) v1;
        roundingMode = RoundingMode.CEILING;
    } else if (v1 instanceof Float) {
        bigDecimal = new BigDecimal((Float) v1);
        roundingMode = RoundingMode.HALF_UP;
    } else if (v1 instanceof Double) {
        bigDecimal = new BigDecimal((Double) v1);
        roundingMode = RoundingMode.HALF_UP;
    } else {
        bigDecimal = new BigDecimal(v1.toString());
        roundingMode = RoundingMode.CEILING;
    }
    Integer scale = (Integer) args.get(3);
    return bigDecimal.setScale(scale, roundingMode);
```

Key findings:

1. **The exact-binary-construction direction is correct and intentional, not the classic footgun.** The PR title ("Preserve exact binary value when CASTing floats to DECIMAL") is deliberate here: `new BigDecimal(double)` — the constructor everyone is usually warned away from — is exactly what's wanted, because the goal is to reproduce what MySQL's own binary-to-DECIMAL `CAST` does. `BigDecimal.valueOf(double)` (which goes through `Double.toString()`) would have reproduced the *old*, buggy behavior. I verified this against the issue's own numbers: `new BigDecimal(126.752251f)` → `126.75225067138671875`, and `.setScale(10, ...)` → `126.7522506714`, matching the documented MySQL output. This is the right call for this specific bug.

2. **This fully resolves my last review's blocking finding.** My 2026-07-26 review flagged that combining exact-binary construction with the pre-existing `RoundingMode.CEILING` would silently corrupt ordinary values — `0.1f`'s exact tail is `0.10000000149...`, so `CEILING` at scale 2 pushed it to `0.11` instead of `0.10`. The new commit (`996cd789c`, "Round exact Float/Double DECIMAL casts half away from zero") fixes this precisely as I suggested: `Float`/`Double` now round with `HALF_UP` (Java's `HALF_UP` is round-half-away-from-zero, matching MySQL's `CAST` rounding), while the legacy `BigDecimal`/`String` branches keep `CEILING` unchanged, so no other input type's behavior is touched. I re-derived the math myself rather than trusting the comment: `new BigDecimal(0.1d)` is `0.1000000000000000055511151231257827021181583404541015625`, and `HALF_UP` at scale 2 correctly yields `0.10` (previously `0.11` under `CEILING`, and originally `0.10` under the pre-PR `toString()` path — so this closes the regression window entirely).

3. **`BigDecimal` inputs are passed through, and `String`/other inputs are untouched** — the diff's only deletion is the single old `new BigDecimal(v1.toString())` line, now guarded behind `instanceof` branches. No other CAST target types, no other call sites, are touched.

4. **Reachability: normal path, not a boundary/recovery path.** `SystemFunction.castAs` is invoked directly from the parsed SQL `CAST(... AS DECIMAL(p,s))` expression evaluation inside `ZetaSQLEngine`, i.e. every row that flows through a SQL transform with such a cast hits this code on the ordinary per-row execution path — there's no exception handling, retry, checkpoint, or recovery logic involved. That also means the earlier `CEILING`-corruption bug (now fixed) would have shipped as a silent per-row data-correctness regression on the mainline path, which is exactly why it was a High-severity blocker before this commit.

5. **No engine/serialization/checkpoint involvement.** All values here are method-local; nothing is persisted to checkpoint state, nothing crosses a thread/RPC boundary, and there's no shared mutable state. A text/mermaid runtime diagram isn't warranted for this PR since it's a pure, synchronous, single-method value transformation with no cross-stage or async flow — consistent with my prior review's assessment.

## 1.2 Compatibility Impact

`Float`/`Double` → `DECIMAL` cast results change relative to the pre-PR release (this is the intended fix, not incidental breakage): values that previously lost precision now round-trip correctly, and the interim `CEILING` regression window (only ever present between the two commits on this branch, never released) is closed. `String`, integer, and `BigDecimal` inputs are byte-for-byte unchanged (verified via `testCastAsDecimalRounding`, the pre-existing `"1.234"` → `1.24` `CEILING` test, which still exercises the untouched `else` branch). No config option, API, SPI, protocol, or checkpoint/state format is touched. Net effect: this is a **behavioral bug fix**, not a compatibility break — the "incompatibility" is with the previously-buggy output, which is precisely what issue #10198 asked to fix.

Negligible. `new BigDecimal(float/double)` allocates a possibly-longer unscaled-value coefficient than the `toString()` path did, but it's immediately reduced via `setScale(...)`, is entirely local to the call, involves no I/O, locking, or retries, and no exception path is newly introduced (NaN/Infinity already threw `NumberFormatException` under both the old `toString()`+`new BigDecimal(String)` path and the new direct-constructor path, so there's no behavior change there either).

No exceptions are swallowed, no new exception types are introduced, and nothing sensitive is logged. No formal issues in this section — see Issue Summary below for the one remaining non-blocking item (documentation).

# 2. Code Quality Assessment
## 2.1 Coding Standards

Formatting, license headers, and the branch structure are all fine. The inline comments on the `Float`/`Double` branches now do double duty as both the "why exact binary" rationale and the "why `HALF_UP`" rationale, with concrete worked examples (`0.1f`, `126.752251f`) — this directly addresses the "comment completeness" gap I raised previously ("compares exact conversion with database semantics without defining the rounding contract"). Good, clear engineering comments.

Three test methods now cover this change (all in `SystemFunctionTest.java`):

- `testCastAsDecimalKeepsExactBinaryValueOfFloats` — the original commit's test: issue-value Float at two scales, a Double, and a `BigDecimal` pass-through.
- `testCastAsDecimalRoundsCommonFloatsHalfAwayFromZero` (new in this commit) — directly closes the gap I flagged: `0.1f`/`0.1d` → `0.10` (the regression case, would have been `0.11` under `CEILING`), negative symmetry (`-0.1f`/`-0.1d` → `-0.10`), the exact half boundary (`0.5f` → `1`, `-0.5d` → `-1`, proving `HALF_UP` rather than `CEILING` — under `CEILING`, `-0.5` would incorrectly stay at `0`), and below/above-half boundaries (`0.49f` → `0`, `0.51f` → `1`).
- `testCastAsDecimalThroughParsedSqlCast` (new in this commit) — exercises the change through the actual parsed-SQL path (`runSql("SELECT CAST(value AS DECIMAL(10,2)) ...")` via `ZetaSQLEngine`) rather than only the direct unit-level helper call, for both the `0.1f` regression case and the original issue value. This is exactly the "parsed SQL CAST" coverage I called out as missing.

I checked these assertions against the pre-fix code by hand (not just trusting the PR's claims): every new assertion in `testCastAsDecimalRoundsCommonFloatsHalfAwayFromZero` and the `0.1f` case in `testCastAsDecimalThroughParsedSqlCast` would fail under the intermediate (first-commit-only) `CEILING` code, and the `126.752251f`/`44.916103d` cases in `testCastAsDecimalKeepsExactBinaryValueOfFloats` would fail under the original pre-PR `toString()` code. Both regressions this PR fixes now have a red-before/green-after test.

Test stability rating: **Stable**. All new tests are pure, deterministic, in-process helper/engine calls — no `Thread.sleep`, no clock/network/container/port dependency, no shared static mutable state, no ordering dependency between tests, and the `BigDecimal` equality assertions are exact (not floating-point deltas), which is correct here since `BigDecimal.equals()` correctly captures both value and scale for these fully-specified expected results.

## 2.3 Documentation

`docs/en/transforms/sql-functions.md` and `docs/zh/transforms/sql-functions.md` list `DECIMAL(p,s)` as a supported `CAST` target type but have never documented a specific rounding contract, before or after this PR — so this PR isn't regressing existing documentation, and I'm not treating it as a hard blocker. That said, since the rounding mode for `Float`/`Double` now differs from `BigDecimal`/`String` inputs (`HALF_UP` vs `CEILING`), a short note in both docs describing that asymmetry would help the next person who's surprised `CAST(0.1 AS DECIMAL(10,2))` and `CAST(CAST(0.1 AS DECIMAL) AS DECIMAL(10,2))` (`String`-literal vs `Float`-typed column) can now round differently. Recommended, not blocking — flagged as Issue 1 below.

# 3. Architectural Soundness

This is the minimal, precise fix, and it's exactly the "better alternative" I proposed in my last review (exact construction + explicit round-to-nearest for `Float`/`Double`, while leaving the legacy `CEILING` branch alone for `String`/`BigDecimal`). It is not a workaround — the type-branch structure directly targets root cause (lossy `toString()`) and directly targets the regression the first commit introduced (wrong rounding mode for the newly-exact-tail values), without touching any other CAST target type or any other code path.

## 3.2 Maintainability

The four-way `instanceof` branch is easy to read and each branch is now well-commented with worked numeric examples. The one thing a future maintainer should notice (and which I'd like documented per 2.3 above) is that the rounding mode is no longer uniform across all input types feeding this one `case` block — that's intentional and currently correct, but it's implicit in the code rather than stated as a design decision anywhere outside the inline comments.

## 3.3 Extensibility

Adding a new numeric input type (e.g. if `BigInteger` inputs were ever routed here) is a straightforward additional branch; nothing in this design blocks that.

No checkpoint/state, serialization format, config option, or SPI contract is touched, so there is nothing to migrate for jobs upgrading across versions. Pre-existing `String`/`BigDecimal` CAST behavior is byte-for-byte preserved (see `testCastAsDecimalRounding`, unchanged and still passing against the untouched `else` branch).

# 4. Issue Summary

|---|---|---|---|
| 1 | `docs/en/transforms/sql-functions.md` and `docs/zh/transforms/sql-functions.md` don't mention that `Float`/`Double` DECIMAL casts now round `HALF_UP` while `String`/`BigDecimal` casts still round `CEILING` | `docs/en/transforms/sql-functions.md`, `docs/zh/transforms/sql-functions.md` | Low |

(For completeness: my previous High-severity Issue — "exact IEEE tails combined with `CEILING` corrupt ordinary DECIMAL casts" — is now fixed by commit `996cd789c` at `SystemFunction.java`'s `Float`/`Double` branches switching to `RoundingMode.HALF_UP`. I verified this by hand-computing the exact binary value of `0.1d`/`0.1f` and confirming `HALF_UP` at scale 2 yields `0.10` as asserted by the new `testCastAsDecimalRoundsCommonFloatsHalfAwayFromZero` test, and by confirming that test would fail under the prior `CEILING`-only code. It is not carried forward as an open item.)

# 5. Merge Recommendation

   - None. The previous High-severity blocker (exact IEEE tails + `CEILING` corrupting ordinary DECIMAL casts) is fixed and verified in this commit.

   - Issue 1: add a short rounding-mode note to `docs/en/transforms/sql-functions.md` / `docs/zh/transforms/sql-functions.md` for the `CAST(... AS DECIMAL(p,s))` function, covering the `HALF_UP` (Float/Double) vs `CEILING` (String/BigDecimal/other) asymmetry.

Note on CI: the `Build` check on the current head (`996cd789c`) is `queued`/not yet evaluated as of this review (the commit was pushed very recently), so I can't confirm green CI at review time. That's a status to watch, not a code-correctness concern — my recommendation above is based on the merits of the diff.

Overall assessment: this is a well-targeted, minimal fix. The exact-binary-construction direction was correct from the first commit and matches real database `CAST` semantics (this is *not* the classic "`new BigDecimal(double)` footgun" — that footgun is only a footgun when you don't want the exact binary value, and here reproducing the exact binary value is the entire point). My prior blocking concern was narrowly about the rounding mode interacting badly with that exact-binary value for common cases like `0.1`, and the new commit resolves it exactly along the lines I suggested, with solid regression tests (including one that goes through the real parsed-SQL path, not just the helper) that I confirmed would have failed under both the pre-PR code and the intermediate (first-commit-only) code. Once CI is green, I'm happy to see this merged.

DanielLeens (commented):
Thanks for keeping this branch in sync, @zhangshenghang — I re-read the entire current diff against `dev` from scratch for this round (not just the delta since my last review), since the task here is a full re-review of the current head. Good news up front: I checked the commit log and the only thing that changed since my last (approving) review is `c5b9fe95e` — a plain `Merge remote-tracking branch 'upstream/dev'` commit. I diffed it explicitly and it carries **zero** changes to either of the two files this PR touches; `gh pr diff` on the current head is byte-for-byte identical to what I reviewed and approved at `996cd789c` on 2026-08-11. There's also no new comment/review from anyone else since then to reply to (review-comment and issue-comment threads are both empty), so this is a confirmation re-review rather than a "something was missed" one — nothing to apologize/carry over here.

- User pain point: casting a `FLOAT`/`DOUBLE` value to `DECIMAL` in a Zeta SQL transform went through `Float.toString()`/`Double.toString()`, which returns the *shortest round-trip* decimal text and silently discards the hidden IEEE-754 binary tail. For a MySQL `FLOAT` column holding `126.752251`, the real binary value is `126.75225067138671875`, but the old code produced `126.7522500000` at scale 10 instead of the correct `126.7522506714` (matching MySQL's own `CAST(... AS DECIMAL(20,10))`) — exactly the discrepancy reported in #10198.
- Fix approach: build the `BigDecimal` directly from the exact binary value (`new BigDecimal(floatValue)` / `new BigDecimal(doubleValue)`) instead of round-tripping through `toString()`, for `Float`/`Double` inputs only, and round with `HALF_UP` (round-half-away-from-zero) instead of the legacy `CEILING` for exactly those two branches; `BigDecimal` inputs pass through unchanged and keep `CEILING`, and all other types keep the pre-existing string-based `CEILING` path.
- One-sentence summary: this correctly reproduces database-style exact-binary `CAST` semantics for floating-point-to-`DECIMAL` conversions, with the rounding mode chosen specifically to avoid corrupting ordinary values like `0.1`.

# 1. Code Change Review
## 1.1 Core Logic Analysis

Changed file: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/sql/zeta/functions/SystemFunction.java`, `castAs()`, `DECIMAL` case.

Before this PR:
```java
case "DECIMAL":
    BigDecimal bigDecimal = new BigDecimal(v1.toString());
    Integer scale = (Integer) args.get(3);
    return bigDecimal.setScale(scale, RoundingMode.CEILING);
```

Current head:
```java
case "DECIMAL":
    BigDecimal bigDecimal;
    RoundingMode roundingMode;
    if (v1 instanceof BigDecimal) {
        bigDecimal = (BigDecimal) v1;
        roundingMode = RoundingMode.CEILING;
    } else if (v1 instanceof Float) {
        bigDecimal = new BigDecimal((Float) v1);
        roundingMode = RoundingMode.HALF_UP;
    } else if (v1 instanceof Double) {
        bigDecimal = new BigDecimal((Double) v1);
        roundingMode = RoundingMode.HALF_UP;
    } else {
        bigDecimal = new BigDecimal(v1.toString());
        roundingMode = RoundingMode.CEILING;
    }
    Integer scale = (Integer) args.get(3);
    return bigDecimal.setScale(scale, roundingMode);
```

Key findings (independently re-derived, not just trusted from the PR description or my prior review):

1. **Exact-binary construction is the correct, deliberate choice here, not the classic footgun.** `new BigDecimal(double)` — the constructor developers are usually warned away from — is exactly what's wanted because the goal is to reproduce what MySQL's own binary-to-`DECIMAL` `CAST` does. `BigDecimal.valueOf(double)` (which goes through `Double.toString()`) would have reproduced the *old*, buggy behavior. I recomputed the exact IEEE-754 binary expansion myself with an independent tool this round: `126.752251f` → exact value `126.75225067138671875`; scaled to 10 digits with `HALF_UP`, the 11th digit is `8`, so it correctly rounds up to `126.7522506714`; scaled to 6 digits, the 7th digit is `6`, so it correctly rounds up to `126.752251`. Both match the test's asserted values and the issue's reported MySQL output.
2. **The rounding-mode fix (`HALF_UP` for `Float`/`Double` only) is correct and closes the exact gap I flagged in my 2026-07-26 review.** I recomputed `0.1f` (`0.100000001490116119384765625`) and `0.1d` (`0.1000000000000000055511151231257827021181583404541015625`): at scale 2 the third digit is `0` in both cases, so `HALF_UP` correctly stays at `0.10`, whereas the old-intermediate `CEILING` mode (present only on the first commit, never released) would have pushed it to `0.11`. I also re-verified the boundary tests by hand: `0.5f`/`-0.5d` are exactly representable and round away from zero to `1`/`-1` under `HALF_UP` (proving it isn't silently still `CEILING`, which would have left `-0.5` at `0`); `0.49f` (exact `0.4900000095367431640625`) correctly stays `0`; `0.51f` (exact `0.5099999904632568359375`) correctly rounds to `1`. Every one of these independently-recomputed exact values matches what the test asserts.
3. **Scope is tightly bounded.** The only deletion in the whole diff is the single old `new BigDecimal(v1.toString())` line, now guarded behind `instanceof` branches. `BigDecimal` inputs are passed through and keep `CEILING`; `String`/other inputs fall into the same unchanged `else` branch with `CEILING`. No other `CAST` target type, and no other call site, is touched.
4. **Reachability: normal path, not a boundary/recovery path.** `SystemFunction.castAs` is invoked directly from parsed SQL `CAST(... AS DECIMAL(p,s))` expression evaluation inside `ZetaSQLEngine`, so every row flowing through a SQL transform with such a cast hits this code on the ordinary per-row execution path. There's no exception handling, retry, checkpoint, or recovery logic involved — all values are method-local, nothing crosses a thread/RPC boundary or touches checkpoint state, so no runtime execution-chain diagram is warranted here (pure synchronous single-method value transform, consistent with both of my prior reviews' assessment).
5. **The one open item from my last review — a doc note for the now-asymmetric rounding mode across input types — is still outstanding**, since the head commit that landed after my approval is a no-op merge and touches neither `docs/en/transforms/sql-functions.md` nor `docs/zh/transforms/sql-functions.md`. Carried forward as Issue 1 below (non-blocking, same as last round).

## 1.2 Compatibility Impact
**Partially incompatible, but as an intentional bug-fix, not a regression.** `Float`/`Double` → `DECIMAL` cast *output values* change relative to the previously-released behavior: values that previously lost precision (or, only on the never-released intermediate commit, got pushed up by one scale unit) now round-trip exactly and round `HALF_UP`. `String`, integer, and `BigDecimal` inputs are byte-for-byte unchanged (still exercised by the pre-existing `testCastAsDecimalRounding` `"1.234"` → `1.24` `CEILING` case, untouched by this diff). No config option, API, SPI, protocol, checkpoint/state format, or job JSON is touched. Net effect is a correctness fix whose only "incompatibility" is with the previously-buggy output — this is exactly what issue #10198 asked for, and doesn't fall under the kind of config/API-removal change that `docs/en/introduction/concepts/incompatible-changes.md` is meant to track.

Negligible. `new BigDecimal(float/double)` may allocate a longer unscaled-value coefficient than the `toString()` path did, but it's immediately reduced via `setScale(...)`, is entirely local to the call, involves no I/O, locking, retries, or shared mutable state. No new exception path is introduced — `NaN`/`Infinity` already threw under the old `toString()` + `new BigDecimal(String)` path and still throw (`NumberFormatException`) under the new direct-constructor path, so behavior there is unchanged.

### Issue 1 — Rounding-mode asymmetry (`HALF_UP` for Float/Double vs `CEILING` for BigDecimal/String) is undocumented
- **Location**: `docs/en/transforms/sql-functions.md`, `docs/zh/transforms/sql-functions.md`
- **Problem description**: Neither doc page defines a rounding contract for `CAST(... AS DECIMAL(p,s))` today (so this PR isn't regressing existing docs), but after this change the rounding mode differs by input type feeding the same `case` block: `Float`/`Double` now round `HALF_UP`, while `BigDecimal`/`String`/other inputs keep `CEILING`.
- **Potential risk**: A user could reasonably be surprised that `CAST('0.005' AS DECIMAL(10,2))` (String, `CEILING`) and `CAST(0.005 :: FLOAT AS DECIMAL(10,2))`-equivalent float-typed input (`HALF_UP`) round differently, with no documentation explaining why.
- **Best improvement**: Add a short note to both doc pages describing the per-type rounding contract (`HALF_UP` for Float/Double exact-binary casts, `CEILING` for BigDecimal/String/other).
- **Severity**: Low
- **Raised by another reviewer**: No (carried forward from my own 2026-08-11 review; still open, no new reviewer has weighed in on this PR)

No exceptions are swallowed, no new exception types are introduced, and nothing sensitive is logged.

# 2. Code Quality Assessment
## 2.1 Coding Standards
Formatting, ASF license headers, and the branch structure are all fine (`git diff --check` clean). The inline comments on the `Float`/`Double` branches explain both the "why exact binary" rationale and the "why `HALF_UP`" rationale with concrete worked examples (`0.1f`, `126.752251f`), which is exactly the comment completeness this kind of non-trivial numeric branch needs. One very minor, non-blocking nit unrelated to the code itself: the current head is a `Merge remote-tracking branch 'upstream/dev'` commit rather than a rebase, so the PR's git history now has a merge commit in it — purely cosmetic, doesn't affect mergeability or content, not something I'd block on.

Three test methods cover this change in `SystemFunctionTest.java`:
- `testCastAsDecimalKeepsExactBinaryValueOfFloats` — the original commit's test: issue-value `Float` at two scales, a `Double`, and a `BigDecimal` pass-through.
- `testCastAsDecimalRoundsCommonFloatsHalfAwayFromZero` — closes the rounding gap: `0.1f`/`0.1d` → `0.10` (the regression case, would be `0.11` under `CEILING`), negative symmetry (`-0.1f`/`-0.1d` → `-0.10`), the exact half boundary (`0.5f` → `1`, `-0.5d` → `-1`, proving `HALF_UP` rather than `CEILING`), and below/above-half boundaries (`0.49f` → `0`, `0.51f` → `1`).
- `testCastAsDecimalThroughParsedSqlCast` — exercises the change through the actual parsed-SQL path (`runSql("SELECT CAST(value AS DECIMAL(10,2)) ...")` via `ZetaSQLEngine`, using the pre-existing private `runSql` test helper and existing `BasicType`/`SeaTunnelRowType` imports already in the file), not only the direct unit-level helper call, covering both the `0.1f` regression case and the original issue value.

I independently recomputed the exact IEEE-754 binary expansions for every asserted value in this round (`126.752251f`, `0.1f`/`0.1d`, `-0.1f`/`-0.1d`, `0.5f`, `-0.5d`, `0.49f`, `0.51f`, `44.916103d`) using an independent tool rather than trusting the PR's or my own prior comments, and every expected `BigDecimal` in the tests matches the correctly-rounded exact value — see the worked numbers in 1.1 above.

Test stability rating: **Stable**. All new tests are pure, deterministic, in-process helper/engine calls — no `Thread.sleep`, no clock/network/container/port dependency, no shared static mutable state, no ordering dependency between tests, and the `BigDecimal` equality assertions are exact (not floating-point deltas), which is correct here since `BigDecimal.equals()` captures both value and scale for these fully-specified expected results.

## 2.3 Documentation
`docs/en/transforms/sql-functions.md` and `docs/zh/transforms/sql-functions.md` list `DECIMAL(p,s)` as a supported `CAST` target type but have never documented a rounding contract, before or after this PR. Not a hard blocker (see Issue 1 above), but recommended before merge for completeness.

# 3. Architectural Soundness
## 3.1 Elegance of the Solution — **Precise fix** (Long-term solution)
This is the minimal, targeted fix: the type-branch structure directly addresses the root cause (lossy `toString()`) and the regression the first commit introduced (wrong rounding mode for the newly-exact-tail values), without touching any other `CAST` target type or code path. Not a workaround.

## 3.2 Maintainability
The four-way `instanceof` branch is easy to read and each branch is well-commented with worked numeric examples. The one thing worth calling out for future maintainers (see Issue 1) is that the rounding mode is no longer uniform across all input types feeding this one `case` block — intentional and currently correct, but currently only documented in code comments, not in user-facing docs.

## 3.3 Extensibility
Adding a new numeric input type (e.g. `BigInteger`) would be a straightforward additional branch; nothing in this design blocks that.

No checkpoint/state, serialization format, config option, or SPI contract is touched, so there is nothing to migrate for jobs upgrading across versions. Pre-existing `String`/`BigDecimal` `CAST` behavior is byte-for-byte preserved.

# 4. Issue Summary

|---|---|---|---|
| 1 | Rounding-mode asymmetry (`HALF_UP` for Float/Double vs `CEILING` for BigDecimal/String) is not documented | `docs/en/transforms/sql-functions.md`, `docs/zh/transforms/sql-functions.md` | Low |

# 5. Merge Recommendation

   - None. The diff on the current head is content-identical to the version I already fully verified and approved on 2026-08-11; my independent recomputation of every asserted exact-binary/rounding value in this round confirms the fix and its tests are correct.

   - Issue 1: add a short rounding-mode note to `docs/en/transforms/sql-functions.md` / `docs/zh/transforms/sql-functions.md` for `CAST(... AS DECIMAL(p,s))`, covering the `HALF_UP` (Float/Double) vs `CEILING` (String/BigDecimal/other) asymmetry.

Process notes (not code-correctness items): the `Build` check on the current head (`c5b9fe95e`, the dev-sync merge commit) is `pending`/not yet evaluated as of this review since it was just pushed — that's a status to watch, not a concern about this diff. `mergeStateStatus` is `BLOCKED` because `reviewDecision` is still `REVIEW_REQUIRED` (a write-capable maintainer approval is still pending) — there's no merge conflict (`mergeable: MERGEABLE`), so this is purely waiting on a maintainer with merge rights, which my account doesn't have.

Overall assessment: this is a well-targeted, minimal fix that correctly reproduces database-style exact-binary `CAST` semantics, and the rounding-mode follow-up fully closes the one real concern I raised in my previous round. Once CI turns green on this head and a maintainer with merge rights takes a pass, I'm happy to see this merged.

DanielLeens (commented):
Quick follow-up, no new commit or discussion since my last review of this same head (`c5b9fe95e`).

In that review my conclusion was already **Ready to merge** with no blockers — the only open item was a process note that the `Build` check on this head (the dev-sync merge commit) hadn't been evaluated yet since it had just been pushed.

I've now re-checked: `Build`, `Notify test workflow`, and `labeler` are all `SUCCESS` on this exact head. `mergeable` is `MERGEABLE` (no conflicts); `mergeStateStatus` still shows `BLOCKED` only because `reviewDecision` is `REVIEW_REQUIRED` — that's purely waiting on a maintainer with merge rights, not a code-side issue.

So from a technical standpoint there's nothing left open on my side: the decimal `CAST` rounding fix and its tests are correct (independently re-verified in my previous round), the one recommended non-blocking follow-up (a docs note on `HALF_UP` vs `CEILING` rounding-mode asymmetry) doesn't hold this up, and CI is now fully green. This is ready for a maintainer to take the final approve/merge step whenever convenient.

DanielLeens (approved):
Circling back on this — my last comment on this same head (`c5b9fe95e`) already concluded 'ready to merge' with no blockers; the only open item was that `Build` on the dev-sync merge commit hadn't been evaluated yet. Rechecking now: `Build`, `Notify test workflow`, and `labeler` are all green on this exact head, and mergeability shows no conflicts. I don't see any other outstanding concern from my side, so reaffirming this as a formal approve.

davidzollo (approved):
+1 LGTM

nzw921rx (approved):
+1 LGTM

PR #11526, opened 2026-07-21
[Fix][Connector-V2] Accept any JSON root in Milvus sink JSON fields

### LINKED ISSUE #9677: [Bug] [milvus] Expected a com.google.gson.JsonObject but was com.google.gson.JsonPrimitive
### Search before asking

### What happened

Seatunnel Milvus source and Milvus sink, source read is ok, but sink has error bellow:

 Caused by: com.google.gson.JsonSyntaxException: Expected a com.google.gson.JsonObject but was com.google.gson.JsonPrimitive; at path $
at com.google.gson.internal.bind.TypeAdapters$34$1.read(TypeAdapters.java:1010)
at com.google.gson.Gson.fromJson(Gson.java:1227)
at com.google.gson.Gson.fromJson(Gson.java:1137)
at com.google.gson.Gson.fromJson(Gson.java:1047)
at com.google.gson.Gson.fromJson(Gson.java:982)
at org.apache.seatunnel.connectors.seatunnel.milvus.utils.sink.MilvusSinkConverter.convertBySeaTunnelType(MilvusSinkConverter.java:73)
at org.apache.seatunnel.connectors.seatunnel.milvus.sink.MilvusSinkWriter.write(MilvusSinkWriter.java:37)
at org.apache.seatunnel.engine.server.task.flow.SinkFlowLifeCycle.received(SinkFlowLifeCycle.java:268)

job status:
{"jobId":"1005680145719099431","pipelineEdges":{"1":[{"inputVertexId":"1","targetVertexId":"2"}]},"envOptions":{"job.mode":"BATCH","execution.parallelism":"3"},"vertexInfoMap":[{"vertexId":1,"type":"source","vertexName":"pipeline-1 [Source[0]-Milvus]","tablePaths":["default.knowledge_embeddings_totals"]},{"vertexId":2,"type":"sink","vertexName":"pipeline-1 [Sink[0]-Milvus-default.knowledge_embeddings_totals]","tablePaths":["zikang_migrate.knowledge_embeddings_totals"]}]},"pluginJarsUrls":[],"finishTime":"2025-08-07 11:34:52","metrics":{"TableSourceReceivedCount":{"default.knowledge_embeddings_totals":"8"},"TableSourceReceivedQPS":{"default.knowledge_embeddings_totals":"0.6235871852833424"},"TableSinkWriteQPS":{},"TableSinkWriteCount":{},"SinkWriteCount":"0","SinkWriteBytesPerSeconds":"0.0","TableSinkWriteBytesPerSeconds":{},"SinkWriteQPS":"0.0","SourceReceivedBytes":"44894","SourceReceivedBytesPerSeconds":"3499.1426344505066","SourceReceivedCount":"8","SourceReceivedQPS":"0.6235385814497272","TableSourceReceivedBytesPerSeconds":{"default.knowledge_embeddings_totals":"3499.1426344505066"},"TableSourceReceivedBytes":{"default.knowledge_embeddings_totals":"44894"},"SinkWriteBytes":"0","TableSinkWriteBytes":{}

### SeaTunnel Version

2.3.9

### SeaTunnel Config

```conf
{
  "env": {
    "execution.parallelism": 3,
    "job.mode": "BATCH"
  },
  "source": [
    {
      "plugin_name": "Milvus",
      "url": "http://10.18.x.x:9091",
      "token": "xxx:xxx",
      "database": "default",
      "collection": "knowledge_embeddings_totals"
    }
  ],
  "sink": [
    {
      "plugin_name": "Milvus",
      "url": "http://x.com:9091",
      "token": "yyy:yyy",
      "database": "zikang_migrate",
      "enable_upsert": "true"
    }
  ]
}
```

### Running Command

```shell
curl -X POST "http://10.18.xx.xx:10901/submit-job" \
  -H "Content-Type: multipart/form-data"  --data-binary @/home/seatunnel/m4.conf
```

### Error Exception

```log
above has
```

_No response_

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
## Purpose of the change

Fixes #9677.

A Milvus source → Milvus sink sync fails whenever a JSON field holds a non-object value:

```
com.google.gson.JsonSyntaxException: Expected a com.google.gson.JsonObject but was com.google.gson.JsonPrimitive
    at ...MilvusSinkConverter.convertBySeaTunnelType(MilvusSinkConverter.java:73)
```

## Root cause

`MilvusSinkConverter.convertBySeaTunnelType` parsed STRING values of JSON fields with `gson.fromJson(value.toString(), JsonObject.class)`. A Milvus `JSON` field may legally hold **any JSON root** — object, array, or primitive (string/number/boolean). Forcing `JsonObject` makes every non-object value blow up the whole job, even though the source read works fine.

## Changes

- **`MilvusSinkConverter`** — parse to a generic `JsonElement` via `JsonParser.parseString(...)`:
  - Object roots still parse to `JsonObject`, so existing behavior is preserved (the downstream `gson.toJsonTree(object)` step in `buildMilvusData` just returns the element as-is).
  - Primitive and array roots now convert correctly instead of throwing.
- **`MilvusSinkConverterTest`** — added cases for `convertBySeaTunnelType`:
  - object root → `JsonObject` with fields intact
  - primitive roots (`"abc"`, `123`, `true`) → `JsonPrimitive`
  - array root (`[1,2,3]`) → `JsonArray`
  - non-JSON string passes through unchanged

## Verification

- `./mvnw -pl seatunnel-connectors-v2/connector-milvus test` — full module suite passes (36 tests).


DanielLeens (commented):
Thanks for fixing this edge case. I went through the full diff on the current head and do not see a source-level blocker in the Milvus change itself. Parsing JSON fields as a generic `JsonElement` is the right direction here because Milvus JSON columns can hold object, array, or primitive roots, and the new tests cover all three shapes while preserving the non-JSON path.

From the current CI evidence, I do not see a failing signal that is directly caused by this Milvus diff yet. The failing lanes are `Run / Dead links` and `Run / unit-test (8, windows-latest)`, and that unit-test failure is in `connector-starrocks`, not in the Milvus connector touched by this PR. This branch is also behind the latest `dev` (`compare status: diverged`, `behind_by=18`), so the lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the rerun still fails after the sync, please share the new failing check link and I can help narrow it down further.

DanielLeens (commented):
Thanks for the update! I rechecked the complete latest head and do not see a source-level blocker in the Milvus change.

Using `JsonParser.parseString(...)` is the correct shared conversion for a Milvus JSON field because object, array, string, number, boolean, and null roots are all valid JSON values. The focused tests cover object, array, and primitive roots while preserving the existing non-JSON string path.

The required Build check is currently failing, so the recorded technical verdict is `clear_waiting_ci`. Once that current-head check is green, the remaining repository gate is one maintainer review.

davidzollo (approved):
+1 if CI passes.

DanielLeens:
Quick follow-up on the gate state: the `Build` check is still failing, but I narrowed the current failure down to the `Code style` lane rather than a Milvus source issue.

The failing check reports a forbidden `java.util.concurrent.CompletableFuture` import in `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/CoordinatorServicePipelineCleanupTest.java`. That file is outside this PR's Milvus JSON-field diff, and this branch is currently behind the latest `dev` branch (`behind_by=32`, compare status: `diverged`).

So I would avoid fixing this by adding unrelated engine/test changes in this PR. The lowest-cost next step is to sync with the latest `dev` and rerun `Build` first. If the same Code style failure still appears after the sync, please paste the new failing check link here and I can help narrow it down further.

Reference check:
https://github.com/zhangshenghang/seatunnel/actions/runs/29907724774/job/88883075587

DanielLeens (approved):
Follow-up on my last comment: the branch has now been synced with `dev` (merge commit `4ac9585`) and the required `Build` check is green on both the `apache` pointer check and the actual build in the contributor's fork (run `31382185154`, same head SHA, `conclusion: success`). Since the merge only pulls in unrelated upstream changes (693 files, none under `connector-milvus`), I re-read the entire current diff and the full `MilvusSinkConverter` runtime path from scratch rather than trusting my two earlier passes.

- User pain point: a Milvus JSON-typed field whose value is a JSON array or a JSON primitive (string/number/boolean) fails the whole sync job with `JsonSyntaxException: Expected a com.google.gson.JsonObject but was com.google.gson.JsonPrimitive`, because `MilvusSinkConverter.convertBySeaTunnelType` force-cast every JSON field through `gson.fromJson(value, JsonObject.class)`. Reported as a Milvus source → Milvus sink round trip in #9677.
- Fix approach: parse with `JsonParser.parseString(value.toString())`, which returns a generic `JsonElement` and accepts any legal JSON root (object, array, string, number, boolean, null), instead of forcing the `JsonObject` subtype.
- One-sentence summary: replaces an overly strict JSON-object-only parse with a JSON-root-agnostic parse, so Milvus JSON columns that legitimately hold non-object values no longer fail sink writes.

# 1. Code Change Review

## 1.1 Core Logic Analysis
`seatunnel-connectors-v2/connector-milvus/.../sink/MilvusSinkConverter.java:73-82`, inside `convertBySeaTunnelType(SeaTunnelDataType<?> fieldType, Boolean isJson, Object value)`, `STRING`/`DATE` case:

Before:
```java
if (isJson) {
    return gson.fromJson(value.toString(), JsonObject.class);
}
```
After:
```java
if (isJson) {
    return JsonParser.parseString(value.toString());
}
```

Key findings:
1. **Single call site, non-null value guaranteed.** `convertBySeaTunnelType` has exactly one production caller, `buildMilvusData` (`MilvusSinkConverter.java:357`). At that call site, `value == null` is handled and `continue`d *before* `convertBySeaTunnelType` is ever invoked (`MilvusSinkConverter.java:319-330`), so the JSON-field branch never sees a Java `null`. No null-root edge case reaches this code from the real code path.
2. **Object-root parity is exact, not just "similar."** Both `Gson.fromJson(String, JsonObject.class)` and `JsonParser.parseString(String)` route through the same internal `TypeAdapters.JSON_ELEMENT` reader for a `{...}` root — the produced `JsonObject` instance is structurally identical either way. I verified this against Gson's source rather than assuming it: `Gson.fromJson` picks a type-hierarchy adapter for `JsonObject.class` (since `JsonObject extends JsonElement`), which is the exact same adapter `JsonParser.parseString` uses. So every currently-succeeding config (object-root JSON fields) gets byte-identical output after this change — this is a strictly additive fix, not a behavior swap for the success path.
3. **Array/primitive roots now convert instead of throwing**, matching the fix's intent and #9677. `[1,2,3]` → `JsonArray`, `"abc"`/`123`/`true` → `JsonPrimitive`. Verified by tracing `Streams.parse` → `TypeAdapters.JSON_ELEMENT.read`, which dispatches on the JSON token type (`BEGIN_OBJECT`/`BEGIN_ARRAY`/`STRING`/`NUMBER`/`BOOLEAN`/`NULL`) rather than forcing a target class.
4. **Downstream consumption is type-agnostic and unaffected.** `buildMilvusData` does `data.add(fieldName, gson.toJsonTree(object))` (`MilvusSinkConverter.java:358`). `gson.toJsonTree` on any `JsonElement` (object, array, primitive, or `JsonNull`) round-trips it as-is via the same type-hierarchy `JsonElement` adapter — it never asserts a specific `JsonElement` subtype. So widening the return type of `convertBySeaTunnelType` from `JsonObject` to the `JsonElement` produced by `JsonParser.parseString` (declared as `Object` at the method signature, so no compile-time widening issue either) is safe for every downstream consumer.
5. **`""` (empty string) and literal `"null"` roots converge to the same final output old vs. new**, even though the intermediate representation differs: old path — `gson.fromJson("", JsonObject.class)` returns Java `null` (Gson treats `EOFException` on an empty reader as "return null"), then `gson.toJsonTree(null)` → `JsonNull.INSTANCE`. New path — `JsonParser.parseString("")` returns `JsonNull.INSTANCE` directly (`Streams.parse` treats the same `EOFException`-on-empty case as `JsonNull.INSTANCE` instead of Java `null`), then `gson.toJsonTree(JsonNull.INSTANCE)` round-trips to an equivalent `JsonNull` element. Same for the four-character literal `"null"`. Net effect on the row actually written to Milvus: identical in both cases. This is not tested explicitly (see 2.2) but I traced it by hand since it's the one case where the intermediate Java type actually changes (`null` reference vs. a `JsonNull` object).

Execution flow (sink write path, JSON field):
```
MilvusWriter.write(SeaTunnelRow)
  -> MilvusSinkConverter.buildMilvusData(catalogTable, config, jsonFields, dynamicField, element)
       for each field:
         value = element.getField(i)
         if value == null -> nullable-check / JsonNull.INSTANCE / continue   (JSON branch never reached)
         object = convertBySeaTunnelType(fieldType, isJson=true, value)
                    -> JsonParser.parseString(value.toString())             (THIS PR'S CHANGE)
                    -> JsonElement (JsonObject | JsonArray | JsonPrimitive | JsonNull)
         data.add(fieldName, gson.toJsonTree(object))                       (identity round-trip for any JsonElement)
  -> data (JsonObject) handed to the Milvus insert/upsert RPC payload builder
```
This is the normal user write path (not a recovery/exception/race branch) — any row with a JSON-typed field goes through `convertBySeaTunnelType` on every write, so the fix is on the hot path, not a corner case.

## 1.2 Compatibility Impact
**Fully compatible.** As shown in finding 2, every JSON field value that previously parsed successfully (object roots) produces the identical `JsonObject` and identical final Milvus payload after this change. The only behavioral difference is that inputs which previously threw `JsonSyntaxException` (array/primitive roots) now succeed — this can only turn a previously-failing job into a succeeding one, never the reverse, for any existing production config. Genuinely malformed JSON (e.g. unbalanced braces) still throws `JsonSyntaxException` from `JsonParser.parseString` the same way `Gson.fromJson` did, so error signaling for actually-invalid data is unchanged.

Note the dynamic-field branch a few lines below (`MilvusSinkConverter.java:~342`, `gson.fromJson(value.toString(), JsonObject.class)` for `dynamicData`) is untouched and still requires an object root — that is correct and not the same bug, since Milvus's dynamic field is a map of extra key/value pairs merged via `.entrySet()`, so it must always be an object; this PR correctly leaves it alone.

No measurable difference. Both the old and new code parse the JSON text exactly once through a `JsonReader`; `JsonParser.parseString` is in fact what `Gson.fromJson(String, JsonElement-hierarchy-class)` delegates to internally, so there is no added parsing pass, allocation pattern change, or new I/O. No new dependency introduced (`com.google.gson.JsonParser` is already transitively available; it's simply a different entry point on the Gson library already in use in this file).

No new issues found. See "Issue Summary" below — none raised for this section; error propagation for malformed JSON is unchanged (still surfaces as an uncaught `JsonSyntaxException` from the sink writer, consistent with the rest of this converter's error style, e.g. the `NOT_SUPPORT_TYPE` default branch).

# 2. Code Quality Assessment

## 2.1 Coding Standards
The changed lines include a clear inline comment explaining why `JsonObject.class` forcing was wrong and what root shapes are now accepted, with a reference to the reported exception message and issue number — good practice, no missing-comment findings for this diff.

`MilvusSinkConverterTest.java` adds four focused unit tests: object root, primitive root (string/number/boolean via one parametrized-by-hand test), array root, and non-JSON pass-through (`isJson=false`) — directly exercising the reported bug and its fix. All are plain synchronous unit tests with no I/O, no sleeps, no shared mutable state, no ordering dependence — **Stable**, no flaky-test anti-patterns present.

Minor coverage gap (non-blocking, see Issue 1): no explicit test for a `null`/empty-string JSON root through `convertBySeaTunnelType`. As traced in finding 5 above, the final written value is provably identical to the pre-fix behavior for that case, so this is a coverage completeness note, not a correctness risk.

## 2.3 Documentation Updates
Not applicable — this is a pure bug fix restoring already-documented/implied field semantics (Milvus JSON columns accept arbitrary JSON), it does not add, rename, or change the default of any user-facing `Option`, and `docs/en/connector-v2/sink/Milvus.md` / the zh equivalent do not describe JSON-root-shape behavior today, so there is nothing stale to update.

# 3. Architectural Soundness

Precise fix. It replaces the wrong tool (a class-forcing deserialize) with the right tool already available in the same library (a root-agnostic parse) at the exact point the bug originates, with no surrounding refactor or unrelated churn — a 5-line functional diff for a bug with a very precisely reproducible root cause.

## 3.2 Maintainability
Improves it slightly: the new comment documents *why* `JsonObject.class` forcing was wrong, which prevents a future contributor from "fixing" this back to the stricter form without understanding the Milvus JSON-field contract.

## 3.3 Extensibility
No impact — this is a leaf conversion helper, not a public API or SPI surface.

No checkpoint/state/serialization format is touched. No config option is added, renamed, or defaulted differently. Existing saved jobs, existing `jsonFields` configuration, and existing successfully-syncing rows are unaffected per 1.2. No upgrade/migration concern.

# 4. Issue Summary

|---|---|---|---|
| 1 | No explicit unit test for a `null`/empty-string JSON root through `convertBySeaTunnelType`, even though I traced by hand that the final Milvus payload is identical to pre-fix behavior for that case | `seatunnel-connectors-v2/connector-milvus/src/test/java/.../MilvusSinkConverterTest.java` | Low |

**Issue 1: Missing null/empty-root regression test**
- **Location**: `seatunnel-connectors-v2/connector-milvus/src/test/java/org/apache/seatunnel/connectors/seatunnel/milvus/utils/sink/MilvusSinkConverterTest.java`
- **Problem description**: The new tests cover object, array, and non-null-primitive roots, but not a `null`-literal or empty-string JSON root, which is the one case where the old code path (`Gson.fromJson` returning Java `null`) and the new code path (`JsonParser.parseString` returning `JsonNull.INSTANCE`) actually diverge at the intermediate-representation level before converging again at `gson.toJsonTree(...)`.
- **Potential risk**: Low — I manually traced both paths through Gson's source and confirmed the final `JsonElement` written to `data` is `JsonNull` either way, so there is no live behavior risk today. The risk is purely that a future refactor of `buildMilvusData`'s `gson.toJsonTree` step could silently change this convergence without a test catching it.
- **Best improvement**: Add one more case to `MilvusSinkConverterTest`, e.g. `convertBySeaTunnelType(BasicType.STRING_TYPE, true, "null")`, asserting the result is a `JsonNull` (or asserting the final `buildMilvusData` output for that field, if a higher-level test is preferred).
- **Severity**: Low

# 5. Merge Recommendation
1. Blockers — none. No blocking issues found on this full independent re-read of the current head.
   - Issue 1: add a `null`/empty-root regression test for completeness (does not need to gate merge).

+1 to @davidzollo's `APPROVED` review ("+1 if CI passes") — CI is now green on the current head (`4ac9585`) on both the `apache` pointer check and the fork's actual `Build` run, and the required-check condition he set is satisfied.

Overall assessment: this is a minimal, correctly-scoped, well-tested fix for a real, reproducible bug (#9677), fully backward compatible with every currently-succeeding Milvus sink JSON-field config, and I don't see a better alternative implementation — using `JsonParser.parseString` instead of `Gson.fromJson(..., JsonObject.class)` is exactly the right level of change for this problem. Thanks for syncing with `dev` per my last comment and getting CI green, @zhangshenghang.

nzw921rx (approved):
+1 LGTM

DanielLeens:
+1, thanks for taking a look @nzw921rx -- agreed this looks ready from my side too. CI is green on the current head and I don't see any outstanding source-level blocker in the Milvus JSON-field fix; the only open item is my earlier non-blocking suggestion to add a null/empty-root regression test as follow-up polish.

davidzollo (approved):
+1
LGTM

PR #11535, opened 2026-07-22
[Fix][Connector-V2] Preserve HTTP source JSON body

### LINKED ISSUE #11533: [Bug] [Http Source] String JSON in body is incorrectly parsed as Map and flattened to key->subkey format
### Search before asking

### Describe the bug
When using SeaTunnel Http Source connector as a client to send POST requests, I configured the `body` as a **JSON string**. However, SeaTunnel forcibly parses this string into a Map object and flattens nested structures using `->` as the delimiter before sending, which corrupts the original JSON structure.

### Steps to Reproduce
1.Start a simple Java SpringBoot service to log the raw request body.
2.Run the SeaTunnel job with the above configuration.
3.Check the server log.

### SeaTunnel Version

2.3.12

### SeaTunnel Config

```conf
{
    "env": {
        "job.mode": "BATCH",
        "job.name": "API_TO_Doris",
        "parallelism": "1",
        "job.retry.times": 0
    },
    "sink": [
        {
            "plugin_name": "jdbc",
            "url": "jdbc:mysql://localhost:9030/test?useUnicode=true&characterEncoding=UTF-8&rewriteBatchedStatements=true",
            "driver": "com.mysql.cj.jdbc.Driver",
            "username": "******",
            "password": "******",
            "generate_sink_sql": true,
            "database": "test",
            "table": "test_table"
        }
    ],
    "source": [
        {
            "body": "{\"pageNo\":1,\"pageSize\":50,\"data\":{\"type\":1}}",
            "connect_timeout_ms": 3000,
            "format": "json",
            "headers": {
                "Content-Type": "application/json",
                "token": "******"
            },
            "json_filed_missed_return_null": true,
            "method": "POST",
            "plugin_name": "Http",
            "plugin_output": "3d883272-2f1f-4e8e-a2b4-b3dd3f8bd712",
            "retry": 0,
            "schema": {
                "fields": {
                    "code": "string",
                    "data": "string",
                    "success": "boolean"
                }
            },
            "socket_timeout_ms": 3000,
            "url": "http://localhost:8080/app/module/api/page"
        }
    ]
}
```

### Running Command

```shell
bin\seatunnel.cmd --config .\etl.json -e local
```

### Error Exception

```log
Expected behavior
The HTTP request body sent to the server should be exactly:


{"pageNo":1,"pageSize":50,"data":{"type":1}}


Actual behavior
The server received the body as (flattened structure):

{
  "pageNo" : 1,
  "pageSize" : 50,
  "data->type" : 1
}
```

Zeta

### Java or Scala Version

java version "17.0.7" 2023-04-18 LTS
Java(TM) SE Runtime Environment (build 17.0.7+8-LTS-224)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.7+8-LTS-224, mixed mode, sharing)

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
## What does this PR do?

Preserves the configured JSON request body for HTTP Source POST requests when `keep_params_as_form` is disabled.

## Why is this needed?

The previous code parsed the body as HOCON and serialized it from a flat map. Nested JSON keys could therefore be emitted with `->` in their names, corrupting the request payload.

## How was this fixed?

The normal JSON POST path now sends the original body string unchanged while retaining configured request parameters in the URL query string. Form submissions keep their existing map-based behavior.

## Tests

- Added a local HTTP server test that verifies a nested JSON body is sent verbatim and query parameters are retained.
- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-http/connector-http-base -Dtest=HttpClientProviderTest test`

Fixes #11533.


nzw921rx (dismissed):
+1 if ci pass

DanielLeens (commented):
Thanks for the fix. I reviewed the full current head and retraced the HTTP source request path:

`HttpSourceReader.executeRequest()` -> `HttpClientProvider.execute()` -> `doPost(url, headers, params, body)`

On the current head, the new branch preserves the configured POST body verbatim while still appending `params` to the query string. That matches the documented nested-JSON / placeholder examples, and I do not see a new source-level blocker in the HTTP connector diff.

The current `Build` failure does not look caused by this PR's changes. The failing lane is `Run / Code style -> Check code specification`, and the concrete failure is `ImportClassCheckTest` on:

`seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/CoordinatorServicePipelineCleanupTest.java:39`

for a forbidden `java.util.concurrent.CompletableFuture` import, which is outside this PR's HTTP files.

This branch is also `behind_by=8` against the latest `dev`, so the lowest-cost next step is to sync with the latest `dev` and rerun CI first. If the updated head still fails after that sync, please share the new Build link and I can narrow it down further.

DanielLeens (commented):
# Basic information

- PR title: [Fix][Connector-V2] Preserve HTTP source JSON body
- Author: zhangshenghang
- Link: https://github.com/apache/seatunnel/pull/11535
- Base branch: dev
- State: OPEN
- Change size: +90 / -2, 2 file(s)
- Human review ready: No
- Review time: 2026-07-26T14:31:30Z
- Local branch: seatunnel-review-11535
- Local worktree: dedicated PR 11535 review worktree

## Stable review keys

- pr_number: `11535`
- base_sha: `6ee8f8d9368ac77eb3d22961ee914f0d602a6c7a`
- reviewed_base_sha: `6ee8f8d9368ac77eb3d22961ee914f0d602a6c7a`
- merge_base_sha: `6ee8f8d9368ac77eb3d22961ee914f0d602a6c7a`
- head_sha: `533099b535649611eccfff72954cf92dcebc7a2e`
- reviewed_head_sha: `533099b535649611eccfff72954cf92dcebc7a2e`
- human_review_ready: `false`
- diff_integrity: `pass`
- deletion_total: `2`
- deletion_classified: `2`
- deletion_unclassified: `0`
- other_total: `0`
- other_confirmed: `0`
- other_unresolved: `0`
- reviewed_at: `2026-07-26T14:31:30Z`

- mergeStateStatus: `BLOCKED`
- mergeable_state: `blocked`
- mergeable: `True`
- rebaseable: `False`
- required_ci_gate: `failing`
- approval_gate: `pending`
- conflict_gate: `satisfied`
- merge_queue_gate: `pending`
- compare: status=`diverged`, ahead_by=`1`, behind_by=`24`, total_commits=`1`, url=https://github.com/apache/seatunnel/compare/dev...533099b535649611eccfff72954cf92dcebc7a2e
- Gate assessment: Source compatibility and test blockers remain independently of the unrelated failing Build; REVIEW_REQUIRED is pending and there is no conflict.

- Other gate `unrelated Code style failure`: `blocking` — Job 88947174389 fails an engine CoordinatorServicePipelineCleanupTest CompletableFuture import outside the HTTP diff.

# What problem this PR solves

- User pain: A nested JSON body is parsed as HOCON, flattened into data->type keys, and then reserialized with a corrupt shape.
- Fix approach: For a nonempty POST body with keep_params_as_form=false, send the original body string and keep params in the URL query.
- One-line summary: Verbatim JSON fixes the nested shape, but the overly broad branch breaks the existing form-urlencoded mode.
- Before: {"data":{"type":1}} becomes {"data->type":1}.
- After: JSON is preserved, but an explicit form header plus default false now sends literal JSON and a conflicting JSON content type instead of id=1 form data.

# 1. Code-change review

- Behavior before: Nested JSON shape is corrupted, but an explicit form header selects UrlEncodedFormEntity.
- Behavior after: Raw JSON is preserved, but all default-false nonempty POST bodies bypass form detection.
- Normal-path hit: `hit` — HttpSourceReader.executeRequest enters the provider for every HTTP source request.

- `seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:120-180` `execute` — Adds an early POST raw-body branch before map and form dispatch.
- `seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:482-502` `doPost(String body)` — Adds params to the query and attaches a raw StringEntity.
- `seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:692-746` `addBody(Map/String)` — The existing Map path honors the form header; the String path adds JSON and ignores the header-selected representation.
- `seatunnel-connectors-v2/connector-http/connector-http-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProviderTest.java:1-90` `nested JSON loopback test` — Proves verbatim JSON and query handling through a dynamic local server only.

### Effective prerequisites

- POST with a nonempty body and keep_params_as_form=false.

- GET, empty-body, and keep_params_as_form=true use existing paths; an explicit form header is incorrectly not excluded.

### Complete execution chain

```text
1. seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/source/HttpSourceReader.java:100-220 -> executeRequest [lifecycle]
   Calls the provider with method, headers, params, body, and form flag.
   Side effect: Calls the provider with method, headers, params, body, and form flag.
2. seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:120-180 -> execute [method_call]
   The new early branch bypasses the old map and form representation selection.
   Side effect: The new early branch bypasses the old map and form representation selection.
3. seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:496-502 -> doPost with query [method_call]
   Moves params to the URL and delegates the raw body.
   Side effect: Moves params to the URL and delegates the raw body.
4. seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:482-493 -> doPost raw [method_call]
   Adds headers and then the String body.
   Side effect: Adds headers and then the String body.
5. seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:692-746 -> addBody [serialization]
   The Map path can URL-encode a form; the String path attaches a raw JSON entity and adds a JSON header.
   Side effect: The Map path can URL-encode a form; the String path attaches a raw JSON entity and adds a JSON header.
```

- Recovery, exception, async, serialization, and concurrency paths: Request builders and entities are method-local; there is no checkpoint, asynchronous, or shared-state change. Retry repeats the same wrong or correct wire body. Response decoding, pagination, and deserialization remain downstream. Invalid raw JSON is now sent verbatim rather than failing HOCON parsing.

- Applicable: `false`
- Evidence: Not applicable: this is an HTTP source-request fix.

## 1.2 Compatibility impact

- Classification: `partially_incompatible` (partially incompatible)
- API: No API-signature change.
- SPI: Connector SPI unchanged.
- Configuration: Existing keep_params_as_form and header options retain names and defaults, but behavior regresses.
- Defaults: The default false now routes an explicit form-header body differently.
- Protocol and serialization: An existing form wire payload and header changes to raw JSON.
- Checkpoint/restore: Not affected.
- Historical behavior: A documented form configuration is broken on upgrade.
- Migration guidance: Prefer preserving the published form route rather than requiring migration.
- Incompatible-changes documentation: Current English and Chinese documentation contradicts the new behavior.

- CPU: The JSON path avoids HOCON parsing and reserialization.
- Memory: Fewer temporary maps and nodes for raw JSON.
- GC: Reduced JSON-path allocations.
- Network: The query is unchanged; the body wire shape changes as described.
- Lock contention: No locks.
- Concurrency safety: Method-local request objects.
- Retry: Retries repeat the same body.
- Idempotency: Pre-existing POST retry semantics.
- Resource release: HTTP lifecycle unchanged.
- Overall: The performance improvement does not offset the compatibility defect.

- Exception propagation: HTTP exceptions still pass through the retry listener.
- Context: The content-type and body combination is not validated.
- Log levels: No logging change.
- Sensitive information: The body and credentials are not newly logged.
- Overall: Representation dispatch must respect the explicit form header.

- diff_integrity: `pass`
- deletion_ledger: total=`2`, classified=`2`, unclassified=`0`
- Reconciliation: 2 = 2 + 0

- `deletion-001..deletion-002` | count=`2` | `deletion` | `intended_change` | `seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java` (old line 140 .. old line 141) — The two deleted lines are the method-normalization comment and method.toUpperCase call moved before the new branch; the same behavior remains as an intended control-flow relocation.

- Out-of-scope hunk conclusion: The production HTTP client and focused unit test only.
- Stale-base conclusion: The live comparison is diverged, ahead by one and behind by 24, with no stale rollback.
- Merge-artifact conclusion: No deletion, rename, conflict, or artifact.

# 2. Code-quality assessment

- Title format: The [Fix][Connector-V2] format is correct.
- Code conventions: A focused diff, retained ASF header, and clean diff check.
- Comment completeness: Representation precedence should be factored or documented in a tested predicate.

- Unit tests: `missing` — The new loopback test covers only JSON and query behavior and misses the explicit form-header branch.
- Component tests: `missing` — There is no coherent content-type or header assertion for the overlapping mode.
- Engine integration tests: `not_applicable` — Request-builder change only.
- E2E: `missing` — The existing form E2E contract would no longer match and no current execution proof exists.
- Regression tests: `missing` — Default-false form plus explicit header and body requires a dedicated regression.

### Test-stability evidence

- `seatunnel-connectors-v2/connector-http/connector-http-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProviderTest.java`
  - Trigger: A dynamic 127.0.0.1:0 synchronous loopback.
  - Probability basis: No sleep, fixed port, container, filesystem, or shared state; AfterEach closes resources.
  - Analysis: Stable; the one-second timeout is at most a low residual risk.

### Missing scenarios

- Assert URL-encoded id=1 and exactly one form content-type header.

## 2.3 Documentation

- User-visible change: `true`
- docs/en: Existing HTTP source documentation explicitly promises the form-header route and current code contradicts it.
- docs/zh: The equivalent Chinese contract is also contradicted.
- Examples: The existing form E2E configuration exercises default false plus a form header.
- Configuration truth: The keep_params_as_form default and header precedence no longer match documentation.
- URL truth: No link change.

- Local boundary: The complete static diff, source, call-chain, deletion, documentation, existing-E2E, test, and CI audit was used; no local Spotless, compilation, tests, E2E, Docker, service, or job was run.

### Local execution boundary

- `git diff --stat 6ee8f8d9368ac77eb3d22961ee914f0d602a6c7a...533099b535649611eccfff72954cf92dcebc7a2e`
  - Result: `passed`
  - Explanation: Confirmed two files and +90/-2.
- `git diff --name-status 6ee8f8d9368ac77eb3d22961ee914f0d602a6c7a...533099b535649611eccfff72954cf92dcebc7a2e`
  - Result: `passed`
  - Explanation: Focused HTTP files with no deletion or rename.
- `git diff --unified=0 6ee8f8d9368ac77eb3d22961ee914f0d602a6c7a...533099b535649611eccfff72954cf92dcebc7a2e`
  - Result: `passed`
  - Explanation: Completed the dispatch, test, and deletion audit.

### Required checks

- `Build` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`FAILURE` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/89513321237

- `Label PRs when reviewed` | workflow=`Label-when-reviewed` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/29932925131/job/88967024825
- `Notify test workflow` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/29927216441/job/88947378136
- `labeler` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/29927216257/job/89513273165
- `Build` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/89513321237

# 3. Architecture assessment

- Type: `precise_fix`
- Assessment: Preserving caller JSON at the request builder is the correct fix point, but the representation decision must preserve form precedence.

## 3.2 Maintainability

The early return duplicates or bypasses entity selection and creates contradictory header behavior.

## 3.3 Extensibility

One tested JSON-versus-form predicate can centralize future body modes.

The documented form request shape must remain unchanged.

# 5. Formal findings

## `Daniel-1` — The raw-body early return breaks the documented form branch

- Location: `seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:134-746`
- Problem and real execution path: Explicit form-header requests with default false bypass Map and form detection and send raw JSON plus a JSON content type. Reader -> execute -> early raw branch -> doPost(String) -> addBody(String), skipping addBody(Map) and UrlEncodedFormEntity.
- Current-head evidence: English and Chinese documentation and the existing E2E and mock require PARAMETERS id=1; the new branch triggers before form inspection.
- Trigger condition: POST with a nonempty body, keep_params_as_form=false, and an explicit form content type.
- Potential impact: Existing documented endpoints receive an incompatible body and header and may reject or misparse it.
- Best actionable fix: Exclude explicit form content type from the verbatim branch, preserve the Map and UrlEncodedFormEntity path, and add a regression.
- Severity: `high`
- Merge gate: `blocking`

## `Daniel-2` — The new test leaves the existing form contract unguarded

- Location: `seatunnel-connectors-v2/connector-http/connector-http-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProviderTest.java:1-90`
- Problem and real execution path: The test asserts only nested JSON and query behavior and never exercises an explicit form header or duplicate content types. Test-server JSON assertion versus missing form dispatch, body, and header assertion.
- Current-head evidence: There is no application/x-www-form-urlencoded case; the existing E2E contract is not PR-owned unit coverage.
- Trigger condition: Dispatch precedence changes now or later.
- Potential impact: The same conditional can regress the form path invisibly to the new test.
- Best actionable fix: Add a local-server case asserting URL-encoded id=1 and one form content type, and keep the existing E2E green.
- Severity: `medium`
- Merge gate: `blocking`

# 6. Findings summary

|---|---|---|---|---|
| Daniel-1 | The raw-body early return breaks the documented form branch | seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProvider.java:134-746 | high | blocking |
| Daniel-2 | The new test leaves the existing form contract unguarded | seatunnel-connectors-v2/connector-http/connector-http-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/http/client/HttpClientProviderTest.java:1-90 | medium | blocking |

# 7. Merge conclusion

### Blocking items

- `blocker-1` [compatibility] (Daniel-1): The unconditional raw-body early return breaks the documented explicit form-header wire contract.
- `blocker-2` [tests] (Daniel-2): The new test covers only JSON and leaves the overlapping existing form contract unguarded.

### Overall assessment

The JSON issue fix is valuable, but the shipped form-contract regression blocks handoff.

- Alternative: Take the verbatim branch only when the request is not explicitly application/x-www-form-urlencoded.
- Trade-offs: Slightly more conditional logic preserves compatibility while avoiding JSON HOCON corruption.

- Reviewed commit: `533099b535649611eccfff72954cf92dcebc7a2e`

DanielLeens (approved):

When a HTTP source `body` contains nested JSON (for example `{"pageNo":1,"data":{"type":1}}`), the old code parsed the body as HOCON into a flat `Map`, then re-serialized it as JSON before sending the request. HOCON treats `.` in keys as a path separator, so a nested key like `data.type` came back out as the literal string key `"data->type"`, silently corrupting the outbound JSON. This PR makes the HTTP client send a configured POST body verbatim instead of round-tripping it through HOCON, while still honoring the pre-existing "send params as an `application/x-www-form-urlencoded` body" compatibility path.

This is a re-review of the current head after a new commit was pushed following my two earlier reviews (2026-07-23 and 2026-07-26). I re-read the full diff, the full runtime path, and my own prior findings from scratch on the current head, `cf2e4316b7`.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Changed files/methods (diffed against the current `dev`, i.e. `apache/dev-latest...cf2e4316b7`, which is a clean 2-file diff since the branch is fully caught up with `dev`, `behind_by=0`):

- `HttpClientProvider.java:120-185` `execute()` — new early-return branch for verbatim POST bodies.
- `HttpClientProvider.java:501-507` `doPost(url, headers, params, body)` (new private overload) — moves `params` to the query string, delegates the raw `body` string.
- `HttpClientProvider.java:697-706` `isFormContentType(headers)` (new private static helper) — the fix for the regression I flagged as blocking in my 2026-07-26 review.

Before this commit (the state I reviewed on 2026-07-26), the early-return guard was:
```java
if (HttpPost.METHOD_NAME.equals(method)
        && !keepParamsAsForm
        && !Strings.isNullOrEmpty(body)) {
    return doPost(url, headers, params, body);
}
```
This unconditionally took the raw-body path for any non-empty POST body when `keep_params_as_form=false`, including requests that set `Content-Type: application/x-www-form-urlencoded` explicitly to opt into form encoding — which is a documented, pre-existing compatibility path (see `docs/en/connectors/source/Http.md:271`: *"if you explicitly force `Content-Type: application/x-www-form-urlencoded`, the runtime follows the form-body branch instead of the default JSON branch"*). That was my Daniel-1 blocking finding.

The current head adds the missing guard:
```java
if (HttpPost.METHOD_NAME.equals(method)
        && !keepParamsAsForm
        && !Strings.isNullOrEmpty(body)
        && !isFormContentType(headers)) {
    return doPost(url, headers, params, body);
}
```
with
```java
private static boolean isFormContentType(Map<String, String> headers) {
    if (MapUtils.isEmpty(headers)) {
        return false;
    }
    return headers.entrySet().stream()
            .filter(entry -> HTTP.CONTENT_TYPE.equalsIgnoreCase(entry.getKey()))
            .map(Map.Entry::getValue)
            .filter(Objects::nonNull)
            .anyMatch(value -> APPLICATION_FORM.equalsIgnoreCase(value.trim()));
}
```

**Key findings:**

1. **This directly and correctly fixes the blocking issue from my last review.** I traced both branches by hand:
   - `Content-Type: application/json` (or no explicit form header) + non-empty body + `keep_params_as_form=false` → takes the new early-return branch → `doPost(url, headers, params, body)` → `doPost(url, headers, body)` (pre-existing 3-arg method, already used by `HttpSinkWriter`/`AirtableSinkWriter`) → `addBody(request, String)` → sends the body verbatim. This is the actual bug fix.
   - `Content-Type: application/x-www-form-urlencoded` + `keep_params_as_form=false` → `isFormContentType` returns `true` → the guard is skipped → falls through to the pre-existing logic: `body` is HOCON-parsed into `bodyMap`, `doPost(url, headers, params, bodyMap)` is called (line 172), which reaches `addBody(HttpEntityEnclosingRequestBase, Map)` → `isFormSubmit` is `true` (Content-Type header matches) → `UrlEncodedFormEntity` is built from the map, i.e. exactly the pre-existing/documented form behavior. This is the historical path Daniel-1 said was broken, and it is now restored.
2. **The matching semantics of `isFormContentType` are consistent with the pre-existing `addBody(Map)` form-detection logic** at line 713-717 (`APPLICATION_FORM.equalsIgnoreCase(...)`, exact string match, no charset-parameter tolerance in either place). So this fix does not introduce a *new* edge case (e.g. `Content-Type: application/x-www-form-urlencoded; charset=UTF-8` was never recognized by the legacy code either) — it faithfully mirrors the existing precision level rather than adding a new gap.
3. **Real execution path confirmed hit on the normal user path**, not just a boundary/recovery path: `HttpSourceReader.pollAndCollectData()` (`HttpSourceReader.java:165`) → `executeRequest()` (`HttpSourceReader.java:224-232`) → `HttpClientProvider.execute(url, method, headers, params, body, keepParamsAsForm)` is the method called on every single poll of every HTTP source job, so this branch is on the hot path for any POST job with a configured body, not an edge case.
4. **This is also backed by a pre-existing E2E regression test that was already in `dev` before this PR** (not part of this PR's diff): `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-http-e2e/src/test/resources/http_formrequestbody_to_assert2.conf`, wired into `HttpIT.java:351-352`. Its MockServer expectation at `mockserver-config.json:4509-4517` matches `POST /example/formBody` only when the request body is `{"type":"PARAMETERS","parameters":{"id":"1"}}` — i.e. MockServer's form-decoded matcher, not a raw-JSON body match. If the raw-body branch were still taken unconditionally (the pre-fix state), this request would send literal `{"id":1}` as the body, MockServer would not match the expectation, and the `Assert` sink's `MIN_ROW=2`/`MAX_ROW=2` rule would fail the job. This means the regression Daniel-1 flagged had a real, pre-existing E2E guard already sitting in the tree — it just never had the chance to catch the bug because CI kept failing earlier at code-style / unrelated stages before E2E could run. That guard is intact for the current head.

Runtime path (matches point 3 above):
```text
HttpSourceReader.pollAndCollectData():165
  -> HttpSourceReader.executeRequest():224-232
  -> HttpClientProvider.execute(url, method, headers, params, body, keepParamsAsForm):120-185
       -> [POST, !keepParamsAsForm, body present, !isFormContentType(headers)]
            -> doPost(url, headers, params, body):501-507  (new)
                 -> doPost(url, headers, body)              (pre-existing, shared with HttpSinkWriter)
                      -> addBody(request, String body)      (pre-existing, verbatim StringEntity)
       -> [otherwise: form CT, or keepParamsAsForm, or GET/PUT/DELETE]
            -> unchanged legacy HOCON-parse + addBody(Map) path
```

## 1.2 Compatibility Impact

**Fully compatible.** With Daniel-1 fixed, both the new behavior (verbatim JSON body — the actual bug being fixed) and the pre-existing documented form-header compatibility path are preserved. No config option name, default value, or public API/SPI signature changed. `docs/en/connectors/source/Http.md` (already updated in the original commit of this PR, not touched by this last commit) accurately describes both branches as they now behave.

One minor, low-risk nuance worth being aware of (not a blocker, not newly introduced by this specific commit — it was already true after the very first commit of this PR): previously, *any* non-empty POST body was parsed with `ConfigFactory.parseString(...)` (HOCON), which incidentally tolerated non-strict-JSON syntax such as unquoted keys (`{id=1}`) and normalized it into proper JSON before sending. After this PR, a body that isn't an explicit form request is sent completely verbatim, so a user who was (undocumented, but possibly) relying on that HOCON leniency for a non-form body would now send a literal, invalid JSON string instead of the normalized one. This isn't part of any documented contract (`docs/en/connectors/source/Http.md` only shows fully-quoted JSON examples), so I don't consider it blocking, just worth a one-line mention in case anyone hits it.

No change from my last review's assessment: the JSON path now skips HOCON parse + re-serialize, which is a small allocation/CPU win for the common case. No new locks, no new retry/idempotency implications (request objects remain method-local), no checkpoint/state interaction (this connector doesn't participate in that path).

No exception handling or logging changes in this commit. No new formal issues here.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- Issue 1: The new `isFormContentType(Map<String, String> headers)` helper (`HttpClientProvider.java:697-706`) and the new private `doPost(String, Map, Map, String)` overload (`HttpClientProvider.java:501-507`) have no doc comment, which is inconsistent with every other `doPost` overload in this same file (all of which carry full `@param`/`@return`/`@throws` Javadoc, e.g. lines 466-476, 478-499, 509-517, 534-542). `isFormContentType` in particular gates a compatibility-sensitive branch, so a short comment explaining *why* the check exists (the same rationale already nicely written above the call site at line 131-137) would help a future reader who only looks at the helper itself.
  - Severity: Low (this is a small, self-explanatory, well-named private helper backed by an excellent block comment at the call site — I'm not raising this to Medium the way I would for an undocumented method whose intent isn't otherwise explained anywhere nearby).
- No other formatting/import issues — no new imports were needed for this commit (all symbols used, e.g. `MapUtils`, `HTTP`, `Objects`, `URIBuilder`, were already imported and used elsewhere in the file).

This directly resolves my Daniel-2 finding from 2026-07-26 ("the new test leaves the existing form contract unguarded"). The new test `executeConvertsBodyToFormWhenContentTypeIsFormUrlencoded` (`HttpClientProviderTest.java`) does exactly what I asked for: it spins up a local `HttpServer`, sends `Content-Type: application/x-www-form-urlencoded` with `body = "{id=1}"`, and asserts the wire body is `id=1` (form-encoded) and the response `Content-Type` is `application/x-www-form-urlencoded`. I hand-traced this through `addBody(HttpEntityEnclosingRequestBase, Map)` and confirmed the assertions match what the code actually produces (see 1.1). Combined with the existing `executePreservesNestedJsonBody` test (from the original commit) and the pre-existing `http_formrequestbody_to_assert2.conf` E2E test (see 1.1, point 4), both branches of the new guard are now covered at both the unit and E2E level.

**Test-stability rating: Stable.**
- `HttpClientProviderTest.java` (both the pre-existing and the new test): ephemeral port via `new InetSocketAddress("127.0.0.1", 0)`, synchronous request/response (no `Thread.sleep`, no polling), 1-second connect/socket timeouts, `@AfterEach` closes both the `HttpClientProvider` and stops the `HttpServer`, no shared static state, no ordering dependency between tests. This is the same safe pattern I already rated stable in my first review; the new test follows it correctly.

## 2.3 Documentation

No doc changes needed for this commit — `docs/en/connectors/source/Http.md` and `docs/zh/connectors/source/Http.md` were already updated in the PR's original commit to describe exactly the two-branch behavior this fix now correctly implements (see `docs/en/connectors/source/Http.md:269-273`). I checked both language versions stayed in sync; they did.

# 3. Architectural Soundness

Precise fix. A single boolean predicate at the exact decision point, reusing the pre-existing `doPost(url, headers, body)` method that `HttpSinkWriter`/`AirtableSinkWriter` already rely on for the same "send this body verbatim" semantics, rather than introducing a parallel code path.

## 3.2 Maintainability

Good. The rationale comment above the guard (lines 131-137) is thorough and explains *why*, not just *what*. My only suggestion is the small doc-comment gap noted in 2.1.

## 3.3 Extensibility

The `isFormContentType` predicate is now a single, centralized, testable place to extend if a third body-representation mode is ever needed — this was literally the extensibility suggestion from my 2026-07-26 review, and it's exactly what landed.

Confirmed compatible: the documented `keep_params_as_form=false` + explicit `Content-Type: application/x-www-form-urlencoded` route (used by real historical job configs, per the existing E2E fixture) now produces byte-for-byte the same form-encoded wire body it did before this PR touched the file.

# 4. Issue Summary

|---|---|---|---|
| 1 | New `isFormContentType`/`doPost(url,headers,params,body)` helpers lack doc comments, unlike every sibling `doPost` overload in the file | `HttpClientProvider.java:501-507`, `HttpClientProvider.java:697-706` | Low |

# 5. Merge Recommendation

1. Blockers — none. Both of my previous blocking findings (Daniel-1: broken form-header compatibility path; Daniel-2: missing regression test for that path) are resolved on the current head, verified by hand-tracing the code and cross-checking against the pre-existing E2E MockServer expectation.
   - Issue 1 above: add a one-line doc comment to `isFormContentType` and the new `doPost` overload for consistency with the rest of the file. Trivial, can be a follow-up nit, does not need to hold up this PR.

Overall assessment: this is now a precise, well-tested fix. Thank you for the quick turnaround on the form-compatibility gap — the added `executeConvertsBodyToFormWhenContentTypeIsFormUrlencoded` test is exactly the coverage I was looking for. The one remaining item is genuinely a nit.

Note on CI: at the time of this review the fork's `Build` run for the current head (`cf2e4316b7`) is still queued and hasn't produced a result yet. The branch is fully caught up with `dev` (`ahead_by=3, behind_by=0`), so the unrelated code-style failure (`ImportClassCheckTest` on an unrelated engine test file) I saw in my first review should no longer apply — that was a `dev`-side issue, and this branch has since synced past it. Please make sure the new `Build` run goes green before merge; I don't have a specific new failure to report right now since it hasn't completed.

DanielLeens (changes requested):
- User pain point: When a POST body is a nested JSON object (e.g. `{"data":{"type":1}}`), `HttpClientProvider.execute()` parsed it through HOCON (`ConfigFactory.parseString`) and re-serialized it, which flattens nested keys into `data->type` before the request is sent — corrupting the wire body for any endpoint expecting real nested JSON.
- Fix approach: When the request is a `POST` with a non-empty body, `keep_params_as_form=false`, and the caller has not explicitly asked for `application/x-www-form-urlencoded`, send the configured body string verbatim and move `params` into the URL query instead of re-parsing/re-serializing it. The newest commit (`cf2e4316b7`) adds the `isFormContentType(headers)` guard so an explicit form `Content-Type` still routes through the legacy HOCON→form-parameter path.
- One-sentence summary: Verbatim-JSON preservation is now correctly gated behind an explicit form-`Content-Type` check, which fixes the JSON corruption bug without breaking the documented form-body contract — but the new code path itself introduces a `Content-Type` header duplication defect that the PR's own new test doesn't catch.

# 1. Code Change Review
## 1.1 Core Logic Analysis

Thanks for the update — I re-reviewed the full diff at the latest head (`cf2e4316b724a490f9ae81215db95b7da5481681`), not just the newest commit's delta.

**Verifying my previous blockers**

> Daniel-1 (2026-07-26): "The raw-body early return breaks the documented form branch... Explicit form-header requests with default `false` bypass Map and form detection and send raw JSON plus a JSON content type."

**Status: Fixed.** `HttpClientProvider.java:138-143` now guards the early return with `!isFormContentType(headers)`:

```java
if (HttpPost.METHOD_NAME.equals(method)
        && !keepParamsAsForm
        && !Strings.isNullOrEmpty(body)
        && !isFormContentType(headers)) {
    return doPost(url, headers, params, body);
}
```

`isFormContentType` (`HttpClientProvider.java:697-706`) scans headers for a `Content-Type` equal (case-insensitive, trimmed) to `application/x-www-form-urlencoded`. When it matches, execution falls through to the pre-existing HOCON-parse path, which — for `keepParamsAsForm=false` — reaches `addBody(Map)`'s `isFormSubmit` check (`HttpClientProvider.java:713-717`, unchanged) and is converted to a `UrlEncodedFormEntity`. This exactly matches the existing e2e fixture `http_formrequestbody_to_assert2.conf`, which sets `headers { Content-Type = "application/x-www-form-urlencoded" }`, `body="{"id":1}"`, and no `keep_params_as_form` (defaults to `false`) — i.e. exactly the scenario I flagged as broken. Verified fixed.

> Daniel-2 (2026-07-26): "The new test leaves the existing form contract unguarded... no `application/x-www-form-urlencoded` case."

**Status: Fixed.** The new commit adds `executeConvertsBodyToFormWhenContentTypeIsFormUrlencoded` (`HttpClientProviderTest.java:161-196`), which sends `Content-Type: application/x-www-form-urlencoded` with body `"{id=1}"` and asserts the server receives the URL-encoded form body `id=1` with the form content type preserved. This directly covers the previously-missing branch.

+1 to @nzw921rx's earlier "+1 if CI pass" — no additional technical content to reconcile there.

**New issue found in this round (carryover — I should have caught this in the 07-26 review, since the call chain was already present in `533099b535`'s diff):**

The new early-return path wires the HTTP **source** read path into a pre-existing but previously sink-only method, `doPost(String url, Map<String,String> headers, String body)` (`HttpClientProvider.java:487-497`), whose `addBody(request, String body)` helper (`HttpClientProvider.java:750-763`) has a latent bug:

```java
private boolean checkAlreadyHaveContentType(HttpEntityEnclosingRequestBase request) {
    if (request.getEntity() != null && request.getEntity().getContentType() != null) {
        return HTTP.CONTENT_TYPE.equals(request.getEntity().getContentType().getName());
    }
    return false;
}

private void addBody(HttpEntityEnclosingRequestBase request, String body) {
    if (checkAlreadyHaveContentType(request)) {
        return;
    }
    request.addHeader(HTTP.CONTENT_TYPE, APPLICATION_JSON);   // <-- unconditional
    ...
    StringEntity entity = new StringEntity(body, ContentType.APPLICATION_JSON);
    entity.setContentEncoding(new BasicHeader(HTTP.CONTENT_TYPE, APPLICATION_JSON));
    request.setEntity(entity);
}
```

`checkAlreadyHaveContentType` inspects `request.getEntity()`, but `addBody(String)` is called from `doPost(url, headers, body)` before any entity has been set on the request (only `addHeaders(httpPost, headers)` ran just before). So `request.getEntity()` is always `null` at this point, `checkAlreadyHaveContentType` always returns `false`, and the code unconditionally calls `request.addHeader(HTTP.CONTENT_TYPE, APPLICATION_JSON)` — regardless of whatever `Content-Type` the caller already supplied via `headers`. Since `AbstractHttpMessage.addHeader()` **appends** rather than replaces, this produces two literal `Content-Type` header entries on the outgoing `HttpPost` whenever the caller's `headers` map already contains a `Content-Type` (the normal, documented way to drive this feature).

This is exactly the shape of the PR's own new unit test `executePreservesNestedJsonBody` (`HttpClientProviderTest.java:126-159`), which passes `Collections.singletonMap("Content-Type", "application/json")` — but that test only asserts `receivedBody`/`receivedQuery`, not header count/value, so the duplication is silent. I checked the existing e2e fixture `http_jsonrequestbody_to_assert.conf` (JSON body, no `headers` block at all) and confirmed it does *not* exercise this path, since with no pre-existing `Content-Type` header only a single `addHeader` call happens.

**Full runtime chain:**
```text
1. HttpSourceReader.executeRequest()
   -> HttpClientProvider.execute(url, method, headers, params, body, keepParamsAsForm)
2. execute(): early-return branch (isFormContentType == false)
   -> doPost(url, headers, params, body)          [new private overload]
3. doPost(url, headers, params, body)
   -> doPost(uri, headers, body)                  [pre-existing public overload]
4. addHeaders(httpPost, headers)  -- adds caller's Content-Type header
5. addBody(httpPost, body)        -- unconditionally re-adds Content-Type: application/json
   => httpPost now carries 2 Content-Type headers whenever caller supplied one.
```
Before this PR, the HTTP source never reached `addBody(String)` at all (it always went through the Map-based `addBody(Map)`, whose else-branch correctly checks `!request.containsHeader(HTTP.CONTENT_TYPE)` before adding a default). So this duplication is a **new regression introduced for the HTTP source relative to pre-PR `dev`**, not merely a pre-existing sink-only quirk being surfaced — it was already present in the diff of commit `533099b535` I reviewed on 2026-07-26 and I missed it then. Sorry for missing this in the last review.

**Concrete impact:** when the caller's header value is byte-identical to `application/json`, most servers tolerate the duplicate (harmless in the common case). When it differs even slightly — `application/json; charset=UTF-8`, a vendor type like `application/vnd.api+json`, etc. — the request now carries two different `Content-Type` values, which RFC 7230 §3.2.2 forbids for non-list headers; behavior at the receiving server becomes implementation-defined. This is exactly the primary intended use case this PR is fixing for (explicit JSON `Content-Type` + JSON body), so it sits on the main path, not a rare corner case.

**Best actionable fix:** fix `checkAlreadyHaveContentType` to check the *request's headers* before the entity is set, mirroring the correct pattern already used at `addBody(Map)`'s `!request.containsHeader(HTTP.CONTENT_TYPE)` check, e.g.:
```java
private void addBody(HttpEntityEnclosingRequestBase request, String body) {
    if (!request.containsHeader(HTTP.CONTENT_TYPE)) {
        request.addHeader(HTTP.CONTENT_TYPE, APPLICATION_JSON);
    }
    ...
}
```
and drop the now-redundant/broken `checkAlreadyHaveContentType` (or repoint it at `request.containsHeader`).

**Second new observation (Medium, non-blocking):** `isFormContentType` (`HttpClientProvider.java:697-706`) uses exact string equality (`APPLICATION_FORM.equalsIgnoreCase(value.trim())`) against `application/x-www-form-urlencoded`. A very common real-world header value, `application/x-www-form-urlencoded; charset=UTF-8`, does **not** match, so the guard would report "not form" and the early raw-body branch would fire even though the caller explicitly asked for form encoding via a parameterized `Content-Type`. This traces back to `addBody(Map)`'s pre-existing `isFormSubmit` check, which has the exact same exact-match limitation and predates this PR entirely — so this is not a new regression versus `dev` (pre-PR, this same charset-parameter case already mis-routed the request, and actually produced the *worse* nested-key-flattening bug that this PR was written to fix). It's a pre-existing limitation that the new guard reproduces rather than fixes. Worth hardening (e.g. split on `;` before comparing, or use `org.apache.http.entity.ContentType.parse(value).getMimeType()`), but not a blocker for this PR.

## 1.2 Compatibility Impact
- Classification: **Partially incompatible** relative to `dev` HEAD, specifically because of the new `Content-Type` duplication defect described above (Issue 1) — it changes wire behavior for HTTP source requests that set an explicit non-form `Content-Type` header, which they were not doing incorrectly before (they simply went through the always-correct `addBody(Map)` else-branch).
- No API/SPI signature changes. `keep_params_as_form` name/default unchanged.
- The form-body regression I flagged in the previous round (Daniel-1) is now fixed and verified compatible with the existing `http_formrequestbody_to_assert2.conf` e2e contract.
- No documentation exists today for `Content-Type`/form-vs-JSON dispatch precedence (`docs/en/connectors/source/Http.md`, `docs/zh/connectors/source/Http.md` — checked, neither mentions `keep_params_as_form` or `Content-Type` dispatch), so there is no documentation contradiction to update for this change.

- The verbatim-JSON path avoids one HOCON parse + one JSON re-serialize per request for the common non-form case — a small, positive CPU/GC change.
- No new locks, no new threads, request objects remain method-local; retry semantics are unchanged (retries still resend the same body/headers, including the duplicated-header defect on each retry).
- No checkpoint/state/serialization involvement — this is a request-builder-only change.

- No new exception paths; HTTP errors still propagate through the existing `Retryer`/`RetryListener` unchanged.
- No sensitive data newly logged.
- (Issue 1) `addBody(String)`'s duplicate `Content-Type` header — see 1.1.
- (Issue 2) `isFormContentType`'s exact-match-without-parameters limitation — see 1.1.

# 2. Code Quality Assessment
## 2.1 Coding Standards
- The new early-return branch has a good, clear explanatory comment covering the "why" and the compatibility constraint — this is exactly the level of comment this kind of change needs.
- **Gap:** the new `isFormContentType(Map<String, String> headers)` method has no Javadoc/comment at all, despite being genuinely non-trivial business logic that decides request-body representation. Per our documentation bar, this is a Medium finding — please add a short comment explaining what it matches and, ideally, its known limitation around `Content-Type` parameters (e.g. `; charset=...`).
- The new private `doPost(String url, Map<String,String> headers, Map<String,String> params, String body)` overload also lacks the Javadoc block that every sibling `doPost` overload in this file carries — Low/style, but worth matching the file's existing convention.

- `HttpClientProviderTest.executePreservesNestedJsonBody` and `executeConvertsBodyToFormWhenContentTypeIsFormUrlencoded` are new, targeted, and both use a dynamic `127.0.0.1:0` `com.sun.net.httpserver.HttpServer` with an `@AfterEach` teardown that closes both the client and the server.
- No `Thread.sleep`, no hard-coded ports, no shared static state, no floating-point assertions, deterministic single-request assertions.
- **Stability classification: Stable.**
- **Coverage gap:** neither new test asserts on the outgoing `Content-Type` header count/value for the JSON-preserve path, which is exactly what would have caught Issue 1. Recommend extending `executePreservesNestedJsonBody` to assert the outgoing request's `Content-Type` header has exactly one entry.

## 2.3 Documentation Updates
- No user-facing config/default/behavior is documented for this dispatch logic today (`docs/en`, `docs/zh` HTTP source pages checked, neither mentions it), so no documentation update is required for this specific change.

# 3. Architectural Soundness
Precise fix for the originally reported nested-JSON corruption, and the new `isFormContentType` guard is the right shape of fix for Daniel-1/Daniel-2 (long-term solution for those two issues). The `addBody(String)` duplication defect it now exposes is a separate, narrower bug in code this PR reuses rather than something this PR's design introduced conceptually — but since this PR is what newly routes the HTTP source through that method, fixing it belongs in this PR.

## 3.2 Maintainability
The file now has three separate "is this a form request" checks with slightly different logic (`isFormContentType`, `isFormSubmit`, and the `!containsHeader` check) — functionally overlapping but not shared. A small refactor to a single helper (or reusing `isFormContentType` inside `addBody(Map)`) would reduce future drift risk; not blocking for this PR, flagged for follow-up.

## 3.3 Extensibility
The new `isFormContentType` predicate is a reasonable seed for centralizing future content-type-based dispatch decisions once hardened for parameterized media types.

With Daniel-1 fixed, the documented/tested form-body contract (`http_formrequestbody_to_assert2.conf`) is preserved. The new Issue 1 (Content-Type duplication) is a regression against `dev` for source requests that set an explicit non-form `Content-Type` header together with a body — this needs to be closed before this can be called fully backward compatible.

# 4. Issue Summary

|---|---|---|---|
| 1 | `addBody(String)` unconditionally adds a second `Content-Type` header because `checkAlreadyHaveContentType` checks the not-yet-set entity instead of the request's existing headers — hits the PR's own primary use case (explicit `Content-Type` + JSON body) and is unasserted by the new test | `HttpClientProvider.java:743-763` (root cause), reached via `HttpClientProvider.java:487-497,501-507,138-143` | High |
| 2 | `isFormContentType` (and pre-existing `isFormSubmit`) use exact string equality against `application/x-www-form-urlencoded`, so `Content-Type` values with parameters (e.g. `; charset=UTF-8`) are not recognized as form — pre-existing limitation the new guard reproduces, not a new regression | `HttpClientProvider.java:697-706` (new), `713-717` (pre-existing) | Medium |
| 3 | New `isFormContentType` method has no Javadoc/comment for non-trivial dispatch logic; new private `doPost(url, headers, params, body)` overload lacks the Javadoc block its sibling overloads have | `HttpClientProvider.java:697-706`, `501-507` | Low |

# 5. Merge Recommendation

   - Issue 1 (High): Fix `checkAlreadyHaveContentType`/`addBody(String)` to check `request.containsHeader(HTTP.CONTENT_TYPE)` before forcing a default `Content-Type`, matching the correct pattern already used in `addBody(Map)`. Please also add an assertion in `executePreservesNestedJsonBody` (or a new test) that the outgoing request carries exactly one `Content-Type` header with the caller-supplied value, so this class of bug can't silently regress again.

   - Issue 2 (Medium): Harden `isFormContentType` to tolerate `Content-Type` parameters (e.g. compare only the media-type portion before `;`), and consider applying the same fix to the pre-existing `isFormSubmit` check for consistency.
   - Issue 3 (Low): Add a short comment to `isFormContentType`, and a Javadoc block to the new private `doPost(url, headers, params, body)` overload, consistent with the rest of the file.
   - Consider centralizing the three overlapping "is this a form request" checks into one shared predicate to reduce future drift risk.

**Overall assessment:** Great progress since the last round — both blockers from my 07-26 review (broken form contract, missing form-header test) are now genuinely fixed and verifiably match the existing e2e contract. Tracing the newest commit's guard end-to-end surfaced one new High-severity issue in code this PR newly exposes to the HTTP source path (`Content-Type` duplication via `addBody(String)`'s stale `checkAlreadyHaveContentType` check) that I should have flagged in the previous round since the call chain already existed then — apologies for missing it. It's a small, mechanical fix (one conditional, mirroring code that already exists two methods away in this same file), and once it's in with a header-asserting test, I don't see anything else blocking merge. Also worth noting: current head's `Build` check is showing as `pending` on GitHub as of this review — please make sure it goes green before merge.

I don't see a materially different alternative implementation here — the verbatim-body-plus-content-type-aware-guard approach is the right shape; it just needs the header-duplication fix to be fully correct.

DanielLeens (changes requested):
Thanks for the very quick turnaround, @zhangshenghang! I can see the new commit (`e29c9d1e8c`) landed specifically to close the High-severity `Content-Type` duplication issue I raised in my previous round, plus a follow-up E2E fixture fix (`60574d1cd9`) for the Airtable connector. This is a re-review of the full current head, not a diff-only recheck — I re-read the entire `HttpClientProvider.java`/`HttpClientProviderTest.java` diff against `dev` from scratch, re-traced the runtime call chain, and re-verified every finding from my three earlier rounds (2026-07-23, 2026-07-26, 2026-08-10) rather than assuming any of them still hold. Good news up front: my last blocking finding is genuinely fixed and well-tested, and I didn't find any new blocker this round.

- User pain point: `HttpClientProvider.execute()` parsed a configured POST `body` through HOCON (`ConfigFactory.parseString`) and re-serialized it before sending, which flattens nested JSON keys — `{"data":{"type":1}}` came out on the wire as `{"data->type":1}` — silently corrupting requests to any endpoint expecting real nested JSON.
- Fix approach: for `POST` + non-empty `body` + `keep_params_as_form=false` + no explicit `Content-Type: application/x-www-form-urlencoded`, send the configured body string verbatim and move `params` to the URL query instead of round-tripping through HOCON. The two follow-up commits since my last review harden `addBody(String)` so it no longer emits a duplicate `Content-Type` header, and update one Airtable E2E fixture to match the (now-correct) literal wire body.
- One-sentence summary: verbatim-JSON preservation, gated behind an explicit-form-`Content-Type` check and a single-`Content-Type`-header guarantee, fixes the original corruption bug while now also preserving the pre-existing form-body and header-hygiene contracts.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Changed files/methods (cumulative diff, `apache/dev...pr-11535`, current head `60574d1cd919f8d3446321a7a0be8cb9fa95a290`, still a clean 3-file diff: `HttpClientProvider.java`, `HttpClientProviderTest.java`, `mockserver-config.json`):

- `HttpClientProvider.java:120-146` `execute()` — early-return branch for verbatim POST bodies, gated by `isFormContentType(headers)`.
- `HttpClientProvider.java:501-507` `doPost(url, headers, params, body)` (private overload) — moves `params` to the query string, delegates the raw `body`.
- `HttpClientProvider.java:696-706` `isFormContentType(headers)` — the fix for the form-contract regression I flagged as blocking on 2026-07-26.
- `HttpClientProvider.java:743-757` `addBody(request, String body)` — **new since my last review**: no longer unconditionally adds `Content-Type: application/json`; now checks `!request.containsHeader(HTTP.CONTENT_TYPE)` first, mirroring the pattern `addBody(Map)` already used. The broken `checkAlreadyHaveContentType(request)` helper (which inspected `request.getEntity().getContentType()` — always `null` at that point in the call chain, since `addBody` runs before any entity is attached) is removed entirely, along with a no-op `entity.setContentEncoding(new BasicHeader(HTTP.CONTENT_TYPE, APPLICATION_JSON))` call that never actually produced an outgoing `Content-Encoding` header (Apache HttpClient's classic request pipeline doesn't sync an entity's `contentEncoding` metadata into a request header automatically — only response decompression reads it). Both were dead weight; removing them is a clean simplification, not a behavior change beyond the intended fix.

**Verifying my 2026-08-10 12:20 blocker (High, "duplicate `Content-Type` header"):**

My finding was: `addBody(String)`'s `checkAlreadyHaveContentType` always returned `false` (because it checked the not-yet-attached entity), so the method unconditionally called `request.addHeader(HTTP.CONTENT_TYPE, APPLICATION_JSON)` even when the caller's `headers` map already supplied a `Content-Type` via `addHeaders()` moments earlier — producing two `Content-Type` header entries on the outgoing `HttpPost`.

**Status: Fixed, and correctly.** The new implementation is:

```java
private void addBody(HttpEntityEnclosingRequestBase request, String body) {
    if (StringUtils.isBlank(body)) {
        body = "";
    }
    StringEntity entity = new StringEntity(body, ContentType.APPLICATION_JSON);
    if (!request.containsHeader(HTTP.CONTENT_TYPE)) {
        request.addHeader(HTTP.CONTENT_TYPE, APPLICATION_JSON);
    }
    request.setEntity(entity);
}
```

I hand-traced the full chain again: `HttpSourceReader.executeRequest()` → `HttpClientProvider.execute()` → early-return branch → `doPost(url, headers, params, body)` → `doPost(url, headers, body)` (pre-existing, shared with `HttpSinkWriter`/`AirtableSinkWriter`) → `addHeaders(httpPost, headers)` (adds the caller's `Content-Type` first) → `addBody(httpPost, body)` — now correctly sees `request.containsHeader(HTTP.CONTENT_TYPE) == true` and skips the default add. Exactly one `Content-Type` header reaches the wire.

**Bonus scope note (not new to this commit, but worth surfacing):** `addBody(String)` is also reached by `HttpSinkWriter.java:141` and `AirtableSinkWriter.java:128` via the same pre-existing 3-arg `doPost(url, headers, body)`. The duplicate-header bug (and now its fix) was never source-specific — any sink that supplies its own `Content-Type` header was already exposed to it on `dev` before this PR touched the file. This commit incidentally fixes that pre-existing sink-side latent bug too, which is a nice compatibility bonus, not a risk.

**Test coverage for the fix:** `HttpClientProviderTest.executePreservesNestedJsonBody` (`HttpClientProviderTest.java`) was extended in this same commit with:
```java
Assertions.assertEquals(Collections.singletonList("application/json"), receivedContentTypes);
```
using a real loopback `HttpServer` that captures `exchange.getRequestHeaders().get("Content-Type")` — i.e. it asserts on the actual header *count*, not just presence, which is exactly the assertion needed to catch a silent duplicate. I traced this against the fixed `addBody(String)` and confirm it would have failed against the pre-fix code and passes against the current head.

**Runtime path (normal path, confirmed hit, not a boundary/recovery path):**
```text
HttpSourceReader.pollAndCollectData()
  -> HttpSourceReader.executeRequest()
  -> HttpClientProvider.execute(url, method, headers, params, body, keepParamsAsForm):120-146
       -> [POST, !keepParamsAsForm, body present, !isFormContentType(headers)]
            -> doPost(url, headers, params, body):501-507
                 -> doPost(url, headers, body)                 (shared with sink writers)
                      -> addHeaders(httpPost, headers)
                      -> addBody(httpPost, body):743-757        (fixed this round)
       -> [otherwise: explicit form Content-Type, or keepParamsAsForm, or GET/PUT/DELETE]
            -> unchanged legacy HOCON-parse + addBody(Map) path
```

**Verifying my 2026-07-26 blockers (Daniel-1/Daniel-2) still hold on the current head:**

Both remain fixed — I re-ran the same trace I did on 2026-08-10 (form `Content-Type` → `isFormContentType` returns `true` → falls through to the legacy HOCON→`UrlEncodedFormEntity` path) directly against the current head's line numbers, and it's unchanged since 2026-08-10. No regression reintroduced.

**Third file in the diff — E2E fixture update (`60574d1cd9`, new since my last review):**

`mockserver-config.json:4838` changes the first-page Airtable `listRecords` expectation from `{"pageSize":2}` to `{"pageSize":2,"offset":null}`. I verified this is a *correct and necessary* consequence of the core fix, not scope creep:
- `AirtableSourceParameter.buildRequestBody()` (unchanged by this PR, pre-existing) always does `body.putIfAbsent("offset", null)` when no explicit offset option is set, specifically so `HttpSourceReader`'s key-based cursor-replacement logic (`replaceBodyPageInfo`/`processBodyMapRecursively`, also pre-existing) has an `"offset"` key to find-and-replace on subsequent pages.
- Pre-PR, that literal `body` string went through `ConfigFactory.parseString(body).entrySet()`, and typesafe-config's `Config.entrySet()` omits null-valued keys — so the `"offset":null` key was silently dropped before the request was ever built, and the mock only ever needed to expect `{"pageSize":2}`.
- Post-PR, the raw-body branch sends the JSON string verbatim, so `"offset":null` now genuinely appears on the wire on the first request, exactly as `AirtableSourceParameter` intended all along.
- I confirmed this is scoped correctly: only the *first*-page mock expectation (line 4838) changed; the second- and third-page expectations (lines 4874, 4910) already carried explicit `offset` values and are untouched.

This is a good outcome — it means the E2E suite actually caught a real wire-format change from this fix (previously-silent `offset` dropping is now corrected) and the author updated the fixture to match rather than papering over it.

## 1.2 Compatibility Impact

**Fully compatible**, upgraded from my 2026-08-10 "partially incompatible" classification now that Issue 1 (Content-Type duplication) is fixed:
- No API/SPI signature changes; `keep_params_as_form` name/default unchanged.
- The documented form-body path (`http_formrequestbody_to_assert2.conf`) is preserved (verified fixed since 2026-08-10, reverified this round).
- The `Content-Type`-duplication regression against `dev` (for any POST source/sink request that supplies an explicit non-form `Content-Type`) is fixed and test-asserted.
- The Airtable `offset:null` wire-format change is a *correction* of previously-silent, unintended data loss (the intended cursor-replacement key was being dropped), not an incompatibility — it makes the actual request match what `AirtableSourceParameter` was always trying to send.

**Self-correction on documentation, since I raised this in my last round:** in my 2026-08-10 review I stated "no documentation exists today for `Content-Type`/form-vs-JSON dispatch precedence... neither doc mentions it." That was incorrect, and I want to correct the record rather than let a wrong claim stand. `docs/en/connectors/source/Http.md:266-280` (and the `docs/zh` equivalent) already documents exactly this precedence — including "if you explicitly force `Content-Type: application/x-www-form-urlencoded`, the runtime follows the form-body branch instead of the default JSON branch" — added in an earlier, unrelated commit (`122d090877b`, 2026-06-13). I re-checked it against the current head's actual behavior line by line and it is accurate for the fixed code, including the "`body` not configured → sends `{}`" case (confirmed: empty `body` skips the new early-return guard via `!Strings.isNullOrEmpty(body)` and falls through to the existing empty-map JSON path). My conclusion ("no doc update needed for this PR") was right, but the stated reasoning was wrong — apologies for the inaccurate claim, correcting it here for the record.

No change from my previous assessment. The verbatim-JSON path avoids one HOCON parse + one re-serialize per request (small CPU/GC win); no new locks/threads; request objects remain method-local; retry semantics unchanged (retries resend the same, now-correct, single-`Content-Type` body/headers); no checkpoint/state involvement (request-builder-only change).

No exception-handling or logging changes in the two new commits. No new formal issues here.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Re-verifying my 2026-08-10 Issue 3 (Low): the new `isFormContentType(Map<String, String> headers)` (`HttpClientProvider.java:696-706`) and the private `doPost(url, headers, params, body)` overload (`HttpClientProvider.java:501-507`) still have no Javadoc, unlike every sibling `doPost` overload in the file. **Status: still open, unchanged, non-blocking** — carried forward as Issue 1 below (same severity/disposition as before, not a new finding).

This closes out my previous coverage gap: `executePreservesNestedJsonBody` now asserts the outgoing `Content-Type` header list has exactly one entry, directly covering the class of bug that shipped in the prior commit. Combined with `executeConvertsBodyToFormWhenContentTypeIsFormUrlencoded` (verified fixed 2026-08-10) and the pre-existing `http_formrequestbody_to_assert2.conf`/`http_jsonrequestbody_to_assert.conf` E2E fixtures, the two dispatch branches and the header-hygiene fix are now all covered at both unit and E2E level.

**Test-stability rating: Stable.** Both `HttpClientProviderTest` tests use an ephemeral `127.0.0.1:0` loopback `HttpServer`, synchronous request/response, no `Thread.sleep`/polling, 1-second timeouts, `@AfterEach` teardown of both client and server, no shared static state, no ordering dependency. Same safe pattern I rated stable in every prior round; the new assertion follows it correctly and adds no new instability.

## 2.3 Documentation Updates

No doc changes needed for these two commits. See 1.2 for my correction of the doc-existence claim from my last review — the existing `docs/en`/`docs/zh` pages already accurately describe the current (fixed) dispatch behavior, and I confirmed en/zh stay in sync.

# 3. Architectural Soundness

Precise fix. The `addBody(String)` change is a one-line-semantics fix (`containsHeader` check moved before the entity exists, instead of an always-false check on the not-yet-set entity) that mirrors a pattern already correct two methods away in the same file (`addBody(Map)`). No parallel logic introduced, no new abstraction — appropriately minimal for the size of the underlying bug.

## 3.2 Maintainability

Good. Removing the broken `checkAlreadyHaveContentType` helper (which had no remaining callers after this change) rather than leaving it as dead code is the right call. My 2026-08-10 note about three overlapping "is this a form request" checks (`isFormContentType`, `isFormSubmit`, and the `!containsHeader` check) still stands as a non-blocking follow-up suggestion — unchanged this round, not re-raised as a new issue.

## 3.3 Extensibility

Unchanged from my last assessment — `isFormContentType` remains a reasonable, centralized seed for future content-type-based dispatch decisions.

Confirmed compatible on the current head: the documented form-body contract, the JSON-preservation fix, and now the single-`Content-Type`-header guarantee are all verified end-to-end (unit + E2E) against pre-existing fixtures that predate this PR.

# 4. Issue Summary

|---|---|---|---|
| 1 | `isFormContentType` and the new `doPost(url,headers,params,body)` overload still lack Javadoc, unlike every sibling `doPost` overload in the file (carried over from 2026-08-10, still open) | `HttpClientProvider.java:696-706`, `501-507` | Low |
| 2 | `isFormContentType` (and pre-existing `isFormSubmit`) use exact string equality against `application/x-www-form-urlencoded`, so `Content-Type` values with parameters (e.g. `; charset=UTF-8`) aren't recognized as form — pre-existing limitation the new guard reproduces, not a new regression (carried over from 2026-08-10, still open) | `HttpClientProvider.java:696-706`, `713-717` | Medium |

Resolved since my last review (not re-listed as open):
- ~~`addBody(String)` unconditionally adds a second `Content-Type` header~~ — fixed in `e29c9d1e8c`, verified fixed and test-asserted (was High/blocking).

# 5. Merge Recommendation

   - None. The one blocker from my previous round (duplicate `Content-Type` header) is confirmed fixed and test-covered on the current head.

   - Issue 1 (Low): add a short Javadoc/comment to `isFormContentType` and to the new private `doPost(url, headers, params, body)` overload, consistent with the rest of the file. Trivial, can be a follow-up.
   - Issue 2 (Medium): harden `isFormContentType`/`isFormSubmit` to tolerate `Content-Type` parameters (e.g. `application/x-www-form-urlencoded; charset=UTF-8`) by comparing only the media-type portion. Pre-existing limitation, not introduced by this PR, worth a follow-up.

**Overall assessment:** this is now a complete, well-verified fix across three rounds — the original JSON-corruption bug, the form-contract regression I flagged on 07-26, and the Content-Type-duplication regression I flagged on 08-10 are all fixed and each has a targeted regression test. The Airtable E2E fixture update is a legitimate, correctly-scoped consequence of the fix rather than incidental scope creep. I don't see a materially different alternative implementation — the verbatim-body-plus-content-type-aware-guard approach, now with correct header hygiene, is the right shape.

**One process note, not a code blocker:** the batch context I was given for this round reports `ci_state: FAILURE` at some point for this PR. Please confirm the `Build` check on the current head (`60574d1cd919f8d3446321a7a0be8cb9fa95a290`) is green before merge — my first-round review (07-23) traced an earlier failure to an unrelated engine code-style test outside this PR's diff, and this branch has since synced with `dev` multiple times, but I have not independently re-observed a fresh, current-head CI result in this round and don't want to assert a status I haven't verified.

DanielLeens (commented):
Quick follow-up, no new commit since my last review — just re-checking the CI/merge-gate fact I flagged.

In my 2026-08-11 review, my "Blockers — must be fixed" list was explicitly `None`: the High-severity `Content-Type` duplication issue was confirmed fixed and test-covered, and both remaining items (missing Javadoc on `isFormContentType`/`doPost` overload; the pre-existing exact-match `Content-Type` parsing limitation) were explicitly called out as non-blocking follow-ups. The only thing I asked to confirm before merge was that the `Build` check on the current head (`60574d1cd919f8d3446321a7a0be8cb9fa95a290`) actually goes green, since I hadn't independently re-observed a fresh result at review time.

I've now re-checked: the head SHA is unchanged, and `Build`, `Notify test workflow`, and `labeler` are all `SUCCESS`. Since CI passing was the only remaining item from my side, I'm updating my conclusion to approve. (Note: `mergeStateStatus` still shows `BLOCKED`, but that's simply because this review thread itself was in `CHANGES_REQUESTED` state pending this confirmation — there's no conflict and the branch is `MERGEABLE`.)

DanielLeens (approved):
Circling back on this — the only thing I had open in my last review was confirming the `Build` check went green on this exact head, and it has: `Build`, `Notify test workflow`, and `labeler` are all SUCCESS on `60574d1c` (unchanged since my last review). No other blocker was left on my side, so updating my conclusion to approve.

davidzollo (approved):
+1
LGTM

PR #11536, opened 2026-07-22
[Fix][Zeta] Fix CompletableFuture import in cleanup test

dybyte:

Fix the invalid `java.util.concurrent.CompletableFuture` import in `CoordinatorServicePipelineCleanupTest` to pass the code style check.

### Check list


nzw921rx:
The latest dev branch has been fixed. Thank you very much for your discovery and timely submission🚀

DanielLeens (commented):
Thanks for the update. I rechecked the latest head against the current `dev`, and I do not see a remaining source-side blocker in the import fix itself.

The key point is that the current `dev` branch already contains the same import in `CoordinatorServicePipelineCleanupTest`, so this PR head no longer carries a distinct code delta against the latest base. At the same time, the branch is now behind upstream (`behind_by=4`, compare status: `diverged`) and the current `Build` check is still failing: https://github.com/apache/seatunnel/runs/88999901977

The lowest-cost next step is to sync with the latest `dev` and rerun Build first. If the diff becomes empty after the sync, please close the PR rather than keeping a no-op stale branch open.

PR #11537, opened 2026-07-22
[Fix][API] Honor multi-table failure policy during timer flush

dybyte:

Honor `multi_table.failure_policy` when an aggregated timer flush fails for a specific table.

Previously, `MultiTableSinkWriter.aggregatedFlush()` propagated the first sub-writer flush exception immediately. Under `CONTINUE_OTHER_TABLES`, this could fail the sink task and prevent remaining healthy tables from being flushed.

This patch isolates the failed table without retrying the flush action at the multi-table wrapper level, while allowing healthy tables to continue. Connector-specific flush retry behavior remains owned by each connector.

No

- Added unit tests verifying:
  - retry and table isolation under `CONTINUE_OTHER_TABLES`
  - healthy tables continue to be flushed
  - quarantined tables are skipped by subsequent timer flushes
  - flush exceptions are immediately propagated under `FAIL_FAST`
 
### Check list


seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java
```diff
@@ -725,7 +725,33 @@ void aggregatedFlush(Map<SinkIdentifier, SinkContextProxy> proxyContexts) throws
                 for (SinkIdentifier id : sinkWritersWithIndex.get(i).keySet()) {
                     SinkContextProxy proxy = proxyContexts.get(id);
                     if (proxy != null && proxy.getFlushAction() != null) {
-                        proxy.getFlushAction().run();
+                        try {
+                            executeWithTableRetry(
+                                    id.getTableIdentifier(),
+                                    MultiTableFailurePhase.RUNTIME_WRITE,
```
nzw921rx: Thank you for your contribution.🚀

1. A new failure type should be introduced here; otherwise, it will be impossible to distinguish a `write` failure from a `flush` failure.

2. I also have one concern. A flush operation handles a batch of records. On the JDBC side, the following logic is used to ensure that records are not inserted repeatedly during retries:

```java
public void addToBatch(SeaTunnelRow record) throws SQLException {
    boolean exist = existRow(record);
    if (exist) {
        if (preExistFlag != null && !preExistFlag) {
            insertStatement.executeBatch();
            insertStatement.clearBatch();
        }
        rowConverter.toExternal(
                valueTableSchema, databaseTableSchema, record, updateStatement);
        updateStatement.addBatch();
    } else {
        if (preExistFlag != null && preExistFlag) {
            updateStatement.executeBatch();
            updateStatement.clearBatch();
        }
        rowConverter.toExternal(
                valueTableSchema, databaseTableSchema, record, insertStatement);
        insertStatement.addBatch();
    }

    preExistFlag = exist;
    submitted = false;
}

@Override
public void executeBatch() throws SQLException {
    if (preExistFlag != null) {
        if (preExistFlag) {
            updateStatement.executeBatch();
            updateStatement.clearBatch();
        } else {
            insertStatement.executeBatch();
            insertStatement.clearBatch();
        }
    }
    submitted = true;
}
```
Do other connectors provide a similar mechanism for handling flush retries? If not, retrying a failed flush may cause unpredictable data issues, such as duplicate writes or inconsistent data.

DanielLeens (commented):
Thanks for working on this. I traced the current timer-flush path in detail, and I agree with @nzw921rx's concern here.

`MultiTableSink.registerAggregatedFlushIfNeeded()` registers `writer.aggregatedFlush(proxyContexts)` as the parent flush action, and Zeta later invokes that through `SinkFlowLifeCycle.processSignal()` on every flush signal. On the current head, `MultiTableSinkWriter.aggregatedFlush()` now wraps every sub-writer `flushAction` in `executeWithTableRetry(...)`.

The blocker is that `SinkWriter.Context.registerFlushAction()` only defines "run this action when a flush signal arrives". It does not define that the action must be externally retry-safe or idempotent across connectors. We already have connector-specific flush semantics in the tree. For example, JDBC registers `this::timerFlush`, and `timerFlush()` already goes through `JdbcOutputFormat.flush()`, which has its own retry loop around batch execution. Retrying the whole flush action again at the multi-table wrapper level can re-run a partially completed connector flush without any shared contract that makes that safe.

So I would treat this as a blocker on the current head. I think the safe direction is either:
1. keep aggregated timer flush fail-fast, or
2. introduce an explicit connector/API-level contract so only retry-safe flush actions are wrapped by the outer retry path.

Separately, the branch is also behind the latest `dev` (`behind_by=35`, compare status: `diverged`), and the current failing `Build` signal is spread across unrelated lanes such as `Dead links`, `Dependency licenses`, and multiple connector jobs. After the source-side blocker is fixed, please sync the latest `dev` and rerun CI on the updated head.

nzw921rx (approved):
+1 good job🚀

DanielLeens (commented):
Thanks for the update. I rechecked the complete latest head and the timer-flush call chain.

The earlier retry/idempotency blocker is resolved on this revision: `aggregatedFlush()` no longer retries a connector flush action. Under `CONTINUE_OTHER_TABLES`, it records the failed table with the new `TIMER_FLUSH` phase, removes/isolates that table through the existing failure handler, and continues to healthy tables; `FAIL_FAST` still propagates the first exception. `InterruptedException` restores the interrupt flag and is never converted into a continue path. The focused tests cover isolation, the second flush (failed table is not retried), close-time failure reporting, and fail-fast behavior.

I do not see a source-level blocker on this head. The required Build check is green. GitHub still reports `reviewDecision=REVIEW_REQUIRED`, so the technical verdict is `clear_waiting_gate`; the remaining gate is one maintainer review.

davidzollo (approved):
+1
LGTM

PR #11525, opened 2026-07-21
[Fix][Connector-V2] Parse ClickhouseFile node_pass with documented snake_case keys

zhangshenghang:
## Purpose of the change

Fixes #9889.

Configuring the ClickhouseFile sink exactly as documented fails at startup:

```
FactoryException: Unable to create a sink for identifier 'ClickhouseFile'
```

## Root cause

The documentation describes `node_pass` entries with snake_case keys:

```hocon
node_pass = [{
    node_address = "xxx"
    password = "xxx"
}]
```

but `NodePassConfig` declares the field `nodeAddress` with no Jackson mapping. `ReadonlyConfig` converts option values with a **plain `ObjectMapper`** (no snake_case naming strategy), so deserializing the documented configuration throws:

```
UnrecognizedPropertyException: Unrecognized field "node_address"
(class ...NodePassConfig), not marked as ignorable
(2 known properties: "nodeAddress", "password"])
```

## Changes

- **`NodePassConfig`**
  - `@JsonProperty("node_address")` on `nodeAddress`, with `@JsonAlias("nodeAddress")` so users who already write the camelCase form keep working.
  - Added the documented `username` field (`node_pass.username`, per the docs table), so entries containing it no longer fail deserialization either.
- **`NodePassConfigTest`** (new)
  - `testNodePassConfigParsesDocumentedSnakeCaseKeys` — goes through the real `ReadonlyConfig.get(NODE_PASS)` conversion path with `node_address`/`username`/`password` and asserts all fields. Without the fix it reproduces the exact `UnrecognizedPropertyException` from the issue (verified by running the test against the unfixed class).
  - `testNodePassConfigAcceptsCamelCaseAlias` — `nodeAddress` keeps working.

## Verification

- `./mvnw -pl seatunnel-connectors-v2/connector-clickhouse test` — full module suite passes.
- Behavioral repro confirmed: with the fix stashed, the snake_case test fails with `Unrecognized field "node_address"`; with the fix, both tests pass.

PR #11540, opened 2026-07-23
[Improve][CI] Keep required Build without strict sync on dev

DanielLeens:
## Why
The previous branch protection change made `Build` required and also set `required_status_checks.strict: true`.

That enforces an extra gate beyond required CI: every PR branch must be updated to the latest `dev` and re-run checks before merge. In practice, an approved PR with a successful `Build` can still be blocked only because it is behind `dev`.

PR #11525 is one concrete example: it had approval plus successful `Build`, but GitHub still blocked merge because the branch was behind `dev`.

## What changed
- keep `Build` as the required status check
- change `required_status_checks.strict` from `true` to `false`

This preserves the intended gate of "CI must pass before merge" without forcing every PR to sync with the latest `dev` before merge.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -53,7 +53,7 @@ github:
   protected_branches:
     dev:
       required_status_checks:
-        strict: true
+        strict: false
         checks:
           - context: Build
             app_id: 15368


davidzollo (approved):
+1

nzw921rx (approved):
+1

PR #11541, opened 2026-07-23
[Fix][Connector-V2] Fix Kafka exactly-once sink losing first record on checkpoint

### LINKED ISSUE #11534: [Bug] [Connector-V2 Kafka] Exactly-once sink can lose the first record during checkpoint
### Search before asking

### What happened

`KafkaIT#testKafkaToKafkaExactlyOnceOnStreaming` intermittently receives only 9 of the 10 records written to a Kafka sink configured with `semantics = EXACTLY_ONCE`.

The failure was observed in the Kafka E2E workflow with Flink 1.18.0 and reproduced locally with Flink 1.15.3. Other engine axes or reruns may pass, which makes this a timing-dependent correctness issue rather than a deterministic test failure.

Failed CI job: https://github.com/nzw921rx/seatunnel/actions/runs/29893085023/job/88837632407

The E2E consumer continues to observe keep-alive records, but the number of matching original records remains at 9 permanently:

```text
testKafkaToKafkaExactlyOnce get data size is not expect,get matched data size 9,visible data size 10,start offset 0,expected count 10
testKafkaToKafkaExactlyOnce get data size is not expect,get matched data size 9,visible data size 18,start offset 0,expected count 10
```

#### Suspected root cause

`KafkaTransactionSender.send()` calls Kafka's asynchronous `send()` and returns immediately. If a checkpoint invokes `prepareCommit()` before the first send request has completed the transaction-start path, `KafkaTransactionSender` captures `KafkaInternalProducer.isTxnStarted()` as `false` even though the transaction has pending records.

The commit metadata therefore represents the transaction as not started. The pending transaction is not committed correctly and eventually times out or is rolled back, leaving one source record permanently invisible to a `read_committed` consumer.

As a diagnostic, flushing pending producer requests before capturing the producer ID, epoch, and `isTxnStarted()` state eliminated the missing-record failure in local Flink 1.13, 1.15, and 1.18 runs. The final fix should ensure the transaction state cannot be captured before pending sends are reflected, and should include focused unit and E2E coverage.

Expected result: all 10 records are committed and visible exactly once.

Actual result: only 9 records become visible; rerunning may pass depending on timing.

### SeaTunnel Version

`dev` / `3.0.0-SNAPSHOT`

### SeaTunnel Config

```conf
env {
  parallelism = 1
  job.mode = "STREAMING"
  checkpoint.interval = 5000
  checkpoint.timeout = 60000
}

source {
  Kafka {
    bootstrap.servers = "kafkaCluster:9092"
    topic = "${sourceTopic}"
    consumer.group = "${consumerGroup}"
    format = text
    start_mode = group_offsets
    kafka.config = {
      enable.auto.commit = "false"
    }
  }
}

sink {
  Kafka {
    format = text
    topic = "${sinkTopic}"
    bootstrap.servers = "kafkaCluster:9092"
    semantics = EXACTLY_ONCE
    kafka.config = {
      transaction.timeout.ms = 60000
      request.timeout.ms = 60000
      delivery.timeout.ms = 60000
      acks = all
      retries = 3
      retry.backoff.ms = 200
    }
  }
}
```

### Running Command

```shell
TEST_IN_PR=true RUN_ALL_CONTAINER=false ./mvnw \
  -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kafka-e2e \
  -DskipUT -DskipIT=false \
  -Dit.test=KafkaIT#testKafkaToKafkaExactlyOnceOnStreaming \
  -DfailIfNoTests=false verify
```

### Error Exception

```log
org.testcontainers.shaded.org.awaitility.core.ConditionTimeoutException:
Assertion condition defined in KafkaIT expected: <true> but was: <false> within 5 minutes.

KafkaIT.testKafkaToKafkaExactlyOnceOnStreaming(KafkaIT.java:1898)

testKafkaToKafkaExactlyOnce get data size is not expect,
get matched data size 9,visible data size 10,start offset 0,expected count 10
```

- CI failure: Flink 1.18.0
- Local reproduction: Flink 1.15.3

### Java or Scala Version

- Java 11
- Scala 2.12

### Screenshots

Not applicable. See the linked GitHub Actions job and log excerpt above.

### Code of Conduct

fatmanverse:

Fixes #11534.

Under `semantics = EXACTLY_ONCE`, the Kafka sink could intermittently lose
the first record of a transaction during checkpointing (e.g.
`KafkaIT#testKafkaToKafkaExactlyOnceOnStreaming` receiving only 9 of 10
records).

### Root cause

`KafkaSinkWriter` sends records via the producer's asynchronous `send()`,
which returns immediately. Kafka only marks a transaction as *started*
(`transactionStarted = true`) once the `AddPartitionsToTxn` request — issued
asynchronously by the producer's sender thread — is acknowledged by the broker.

`KafkaTransactionSender#prepareCommit()` is called by the engine before
`snapshotState()`. If a checkpoint reached `prepareCommit()` before the first
record's transaction registration completed, `isTxnStarted()` returned a stale
`false`, which was then persisted into `KafkaCommitInfo`.

On checkpoint completion, `KafkaSinkCommitter` resumed the transaction with
`txnStarted = false`, and Kafka's `TransactionManager` skips `EndTxn` in that
state. The pending broker-side transaction was never committed and eventually
timed out / was aborted, leaving that record permanently invisible to
`read_committed` consumers.

A separate pre-existing gap also allowed asynchronous send failures to be
missed. `flush()` waits for submitted sends to complete, but completion can be
successful or exceptional. Once any partition has been added successfully,
`txnStarted` may remain `true` even when a later send callback reports a
failure.

### Fix

In `KafkaTransactionSender`:

1. `flush()` pending sends **before** capturing the transaction state, so the
   `AddPartitionsToTxn` registration is reflected in `isTxnStarted()`.
2. Record the first asynchronous send callback failure and check it immediately
   after `flush()`. Fail the checkpoint with `PRODUCE_DATA_FAILED`
   (`KAFKA-09`) even when `txnStarted = true`, preserving the original
   exception as the cause.
3. Fail fast with `TRANSACTION_NOT_STARTED` (`KAFKA-08`) when the transaction
   has records but registration is still not started after flushing, rather
   than emitting commit info that would make the committer skip `EndTxn`.

The empty-transaction path (`recordNumInTransaction == 0`) is unchanged and
still legitimately reports `txnStarted = false`.

Yes, operationally. No config option, public API, or SPI contract is changed.
An exactly-once checkpoint now fails when an asynchronous Kafka send fails,
instead of allowing a partial transaction to proceed. Under a slow broker,
checkpoints may also block slightly longer on `flush()`; both behaviors prefer
explicit checkpoint failure over silent data loss.

## How was this patch tested?

- **Unit** — `KafkaTransactionSenderTest` (4 cases):
  - flush happens before the transaction state is captured;
  - fail-fast when records were sent but transaction registration is still not
    started after flushing;
  - propagate an asynchronous send callback failure when
    `txnStarted = true`;
  - allow an empty transaction with `txnStarted = false`.
- **E2E** — `KafkaIT#testKafkaToKafkaExactlyOnceOnStreaming` is the existing
  reproducer for the original race; its `checkData` assertion detects both
  lost and duplicate records. The previous CI head passed Kafka connector IT on
  JDK 8 and JDK 11.
- **Local** — `./mvnw spotless:apply` passed; Kafka connector unit tests passed
  (63 tests); Kafka connector dependency-chain `verify -DskipTests` passed.

## Check list

- [x] Code changed are covered with tests, or it does not need tests for reason in the PR description.
- [x] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
- [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs

### DIFF
--- a/docs/en/connectors/sink/Kafka.md
+++ b/docs/en/connectors/sink/Kafka.md
@@ -401,6 +401,17 @@ sink {
 
 Ensure the Kafka broker has transactions enabled and that `transaction.timeout.ms` is aligned with your checkpoint interval.
 
+Under `EXACTLY_ONCE`, a failed send fails the checkpoint instead of silently dropping records. Two
+errors can be reported in that situation:
+
+| Code     | Name                    | Meaning                                                                       | What to do                                                                                                   |
+|----------|-------------------------|-------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------|
+| KAFKA-08 | TRANSACTION_NOT_STARTED | The transaction carries records but Kafka never registered it on the broker.   | Check broker availability and whether `transaction.timeout.ms` is shorter than the checkpoint interval.       |
+| KAFKA-09 | PRODUCE_DATA_FAILED     | A record of the transaction failed to be sent asynchronously.                  | Read the exception cause; retriable causes usually recover on checkpoint retry, others need broker-side work. |
+
+Both errors abort the current transaction, so the affected records are re-sent from the last
+completed checkpoint rather than lost.
+
 ### How do I configure SASL/Kerberos authentication?
 
 Pass broker authentication settings via `kafka.*` properties:
--- a/docs/zh/connectors/sink/Kafka.md
+++ b/docs/zh/connectors/sink/Kafka.md
@@ -358,6 +358,15 @@ sink {
 
 确保 Kafka Broker 开启了事务支持，且 `transaction.timeout.ms` 与 checkpoint 间隔相匹配。
 
+在 `EXACTLY_ONCE` 语义下，发送失败会让 checkpoint 失败，而不是静默丢弃数据。此时可能出现两种错误：
+
+| 错误码      | 名称                      | 含义                                        | 处理建议                                                             |
+|----------|-------------------------|-------------------------------------------|------------------------------------------------------------------|
+| KAFKA-08 | TRANSACTION_NOT_STARTED | 事务中已有数据，但 Kafka 始终未在 Broker 端完成该事务的注册。    | 检查 Broker 是否可用，以及 `transaction.timeout.ms` 是否小于 checkpoint 间隔。   |
+| KAFKA-09 | PRODUCE_DATA_FAILED     | 事务中的某条数据异步发送失败。                           | 查看异常 cause；可重试的异常通常在 checkpoint 重试后恢复，其他异常需要排查 Broker 端问题。      |
+
+两种错误都会中止当前事务，受影响的数据会从上一个已完成的 checkpoint 重新发送，不会丢失。
+
 ### 如何配置 SASL/Kerberos 认证？
 
 ```hocon
--- a/seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/exception/KafkaConnectorErrorCode.java
+++ b/seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/exception/KafkaConnectorErrorCode.java
@@ -29,7 +29,11 @@ public enum KafkaConnectorErrorCode implements SeaTunnelErrorCode {
     CONSUME_THREAD_RUN_ERROR(
             "KAFKA-05", "Error occurred when the kafka consumer thread was running"),
     CONSUME_DATA_FAILED("KAFKA-06", "Kafka failed to consume data"),
-    CONSUMER_CLOSE_FAILED("KAFKA-07", "Kafka failed to close consumer");
+    CONSUMER_CLOSE_FAILED("KAFKA-07", "Kafka failed to close consumer"),
+    TRANSACTION_NOT_STARTED(
+            "KAFKA-08",
+            "Kafka transaction still reported as not started after flushing pending sends"),
+    PRODUCE_DATA_FAILED("KAFKA-09", "Kafka failed to produce data");
 
     private final String code;
     private final String description;
--- a/seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java
+++ b/seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java
@@ -20,11 +20,14 @@ package org.apache.seatunnel.connectors.seatunnel.kafka.sink;
 import org.apache.seatunnel.shade.com.google.common.collect.Lists;
 
 import org.apache.seatunnel.connectors.seatunnel.kafka.KafkaClientUtils;
+import org.apache.seatunnel.connectors.seatunnel.kafka.exception.KafkaConnectorErrorCode;
+import org.apache.seatunnel.connectors.seatunnel.kafka.exception.KafkaConnectorException;
 import org.apache.seatunnel.connectors.seatunnel.kafka.state.KafkaCommitInfo;
 import org.apache.seatunnel.connectors.seatunnel.kafka.state.KafkaSinkState;
 
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
 
 import lombok.extern.slf4j.Slf4j;
 
@@ -32,6 +35,7 @@ import java.time.Duration;
 import java.util.List;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.apache.seatunnel.connectors.seatunnel.kafka.sink.KafkaSinkWriter.generateTransactionId;
 
@@ -48,6 +52,15 @@ public class KafkaTransactionSender<K, V> implements KafkaProduceSender<K, V> {
     private String transactionId;
     private final String transactionPrefix;
     private final Properties kafkaProperties;
+
+    /**
+     * Holds the first asynchronous send failure of the current transaction. It is written from the
+     * producer's sender thread through the send callback and read by the task thread, so it must
+     * stay thread-safe. It is scoped to a single transaction and therefore reset by {@link
+     * #beginTransaction(String)}.
+     */
+    private final AtomicReference<Exception> asyncSendException = new AtomicReference<>();
+
     private int recordNumInTransaction = 0;
 
     public KafkaTransactionSender(String transactionPrefix, Properties kafkaProperties) {
@@ -57,8 +70,32 @@ public class KafkaTransactionSender<K, V> implements KafkaProduceSender<K, V> {
 
     @Override
     public void send(ProducerRecord<K, V> producerRecord) {
-        kafkaProducer.send(producerRecord);
+        // Surface an already recorded asynchronous failure on the write path. The current
+        // transaction can no longer be committed, so buffering and transmitting more records for it
+        // would only waste producer memory and network bandwidth until the next checkpoint.
+        checkAsyncSendException();
+        kafkaProducer.send(producerRecord, this::onSendCompleted);
         recordNumInTransaction++;
+    }
+
+    /**
+     * Records the first asynchronous send failure of the current transaction so that {@link
+     * #prepareCommit()} can fail the checkpoint instead of committing a partial transaction.
+     *
+     * <p>Invoked on the producer's sender thread.
+     */
+    private void onSendCompleted(RecordMetadata metadata, Exception exception) {
+        if (exception == null) {
+            return;
+        }
+        if (!asyncSendException.compareAndSet(null, exception)) {
+            // Only the first failure becomes the checkpoint failure cause. Log the later ones so a
+            // broker-side incident affecting several partitions can still be diagnosed.
+            log.warn(
+                    "Suppressed an additional asynchronous send failure of Kafka transaction [{}]",
+                    transactionId,
+                    exception);
+        }
     }
 
     @Override
@@ -66,19 +103,61 @@ public class KafkaTransactionSender<K, V> implements KafkaProduceSender<K, V> {
         this.transactionId = transactionId;
         this.kafkaProducer = getTransactionProducer(transactionId);
         kafkaProducer.beginTransaction();
+        // Reset the per-transaction state. A new transaction always runs on a newly created
+        // producer, so a failure recorded for the previous transaction no longer applies. Keeping
+        // it would turn a single transient send error into a permanent checkpoint failure loop.
         recordNumInTransaction = 0;
+        asyncSendException.set(null);
     }
 
     @Override
     public Optional<KafkaCommitInfo> prepareCommit() {
+        // Flush pending async sends before capturing the transaction state. Kafka only marks the
+        // transaction as started once the AddPartitionsToTxn request has been acknowledged by the
+        // broker, and that request is issued asynchronously by the producer's sender thread. If a
+        // checkpoint reaches this point before the first record's transaction registration
+        // completes, isTxnStarted() would still return false and the resulting commit info would
+        // instruct the committer to skip EndTxn, leaving the transaction to time out and its
+        // records permanently invisible to read_committed consumers.
+        kafkaProducer.flush();
+        checkAsyncSendException();
+        boolean txnStarted = kafkaProducer.isTxnStarted();
+        if (recordNumInTransaction > 0 && !txnStarted) {
+            // Records were sent in this transaction but Kafka still reports it as not started even
+            // after flushing, meaning the transaction registration never completed. Committing with
+            // txnStarted=false would make the committer skip EndTxn and drop these records, so fail
+            // fast and let the checkpoint abort this transaction instead of silently producing a
+            // lossy commit info.
+            throw new KafkaConnectorException(
+                    KafkaConnectorErrorCode.TRANSACTION_NOT_STARTED,
+                    String.format(
+                            "Kafka transaction [%s] has %d record(s) but is still reported as not "
+                                    + "started after flushing pending sends. The transaction "
+                                    + "registration did not complete. Refusing to commit to avoid "
+                                    + "data loss.",
+                            transactionId, recordNumInTransaction));
+        }
         KafkaCommitInfo kafkaCommitInfo =
                 new KafkaCommitInfo(
                         transactionId,
                         kafkaProperties,
                         this.kafkaProducer.getProducerId(),
                         this.kafkaProducer.getEpoch(),
-                        this.kafkaProducer.isTxnStarted());
+                        txnStarted);
         return Optional.of(kafkaCommitInfo);
+    }
+
+    private void checkAsyncSendException() {
+        Exception exception = asyncSendException.get();
+        if (exception != null) {
+            throw new KafkaConnectorException(
+                    KafkaConnectorErrorCode.PRODUCE_DATA_FAILED,
+                    String.format(
+                            "Kafka transaction [%s] failed to send one or more of its %d record(s) "
+                                    + "asynchronously.",
+                            transactionId, recordNumInTransaction),
+                    exception);
+        }
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSenderTest.java
+++ b/seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSenderTest.java
@@ -17,7 +17,12 @@
 
 package org.apache.seatunnel.connectors.seatunnel.kafka.sink;
 
-import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.seatunnel.connectors.seatunnel.kafka.exception.KafkaConnectorErrorCode;
+import org.apache.seatunnel.connectors.seatunnel.kafka.exception.KafkaConnectorException;
+import org.apache.seatunnel.connectors.seatunnel.kafka.state.KafkaCommitInfo;
+
+import org.apache.kafka.clients.producer.Callback;
+import org.apache.kafka.clients.producer.ProducerRecord;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
@@ -25,12 +30,180 @@ import org.mockito.Mockito;
 
 import java.time.Duration;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Deque;
 import java.util.List;
+import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 class KafkaTransactionSenderTest {
 
+    private static final String TRANSACTION_PREFIX = "SeaTunnel0001";
+    private static final String TRANSACTION_ID = TRANSACTION_PREFIX + "-1";
+    private static final String NEXT_TRANSACTION_ID = TRANSACTION_PREFIX + "-2";
+    private static final String TOPIC = "test-topic";
+    private static final long PRODUCER_ID = 1001L;
+    private static final short EPOCH = 5;
+
+    /**
+     * Reproduces the reported race: Kafka reports the transaction as not started until the pending
+     * AddPartitionsToTxn request has been acknowledged, which flush() forces. prepareCommit() must
+     * flush first so the captured commit info carries txnStarted=true and the committer performs
+     * EndTxn.
+     */
+    @Test
+    void prepareCommitFlushesBeforeCapturingTransactionState() {
+        ProducerStub producer = new ProducerStub();
+        producer.startTransactionOnFlush();
+
+        TestingKafkaTransactionSender sender = createSender(producer);
+        sender.beginTransaction(TRANSACTION_ID);
+        sender.send(record());
+
+        Optional<KafkaCommitInfo> commitInfo = sender.prepareCommit();
+
+        Assertions.assertTrue(commitInfo.isPresent());
+        Assertions.assertTrue(
+                commitInfo.get().isTxnStarted(),
+                "txnStarted must be captured after flush so the committer performs EndTxn");
+        Assertions.assertEquals(TRANSACTION_ID, commitInfo.get().getTransactionId());
+        Assertions.assertEquals(PRODUCER_ID, commitInfo.get().getProducerId());
+        Assertions.assertEquals(EPOCH, commitInfo.get().getEpoch());
+        verify(producer.mock, times(1)).flush();
+    }
+
+    /**
+     * When the transaction genuinely carries records but Kafka still reports it as not started even
+     * after flushing, prepareCommit() must fail fast rather than emit a commit info that instructs
+     * the committer to skip EndTxn and silently drop those records.
+     */
+    @Test
+    void prepareCommitFailsWhenRecordsSentButTransactionNotStarted() {
+        ProducerStub producer = new ProducerStub();
+
+        TestingKafkaTransactionSender sender = createSender(producer);
+        sender.beginTransaction(TRANSACTION_ID);
+        sender.send(record());
+        sender.send(record());
+
+        KafkaConnectorException exception =
+                Assertions.assertThrows(KafkaConnectorException.class, sender::prepareCommit);
+
+        Assertions.assertEquals(
+                KafkaConnectorErrorCode.TRANSACTION_NOT_STARTED.getCode(),
+                exception.getSeaTunnelErrorCode().getCode());
+        verify(producer.mock, times(1)).flush();
+    }
+
+    /**
+     * A transaction can already be marked as started while a record still fails to be sent. The
+     * failure is only reported once flush() completes the pending send, so prepareCommit() must
+     * check for it after flushing instead of committing the remaining successful records.
+     */
+    @Test
+    void prepareCommitFailsWhenAsyncSendFailsAfterTransactionStarted() {
+        ProducerStub producer = new ProducerStub();
+        producer.transactionStarted();
+        RuntimeException asyncSendFailure = new RuntimeException("async send failed");
+        producer.failPendingSendsWith(asyncSendFailure);
+
+        TestingKafkaTransactionSender sender = createSender(producer);
+        sender.beginTransaction(TRANSACTION_ID);
+        sender.send(record());
+
+        KafkaConnectorException exception =
+                Assertions.assertThrows(KafkaConnectorException.class, sender::prepareCommit);
+
+        Assertions.assertEquals(
+                KafkaConnectorErrorCode.PRODUCE_DATA_FAILED.getCode(),
+                exception.getSeaTunnelErrorCode().getCode());
+        Assertions.assertSame(asyncSendFailure, exception.getCause());
+        verify(producer.mock, times(1)).flush();
+    }
+
+    /**
+     * An empty transaction (no records sent) legitimately reports txnStarted=false and must not be
+     * treated as an error; the commit info simply carries txnStarted=false.
+     */
+    @Test
+    void prepareCommitAllowsEmptyTransactionWithoutRecords() {
+        ProducerStub producer = new ProducerStub();
+
+        TestingKafkaTransactionSender sender = createSender(producer);
+        sender.beginTransaction(TRANSACTION_ID);
+
+        Optional<KafkaCommitInfo> commitInfo = sender.prepareCommit();
+
+        Assertions.assertTrue(commitInfo.isPresent());
+        Assertions.assertFalse(commitInfo.get().isTxnStarted());
+        verify(producer.mock, times(1)).flush();
+    }
+
+    /**
+     * Once a failure has been recorded the current transaction can no longer be committed, so the
+     * write path must reject further records instead of buffering them until the next checkpoint.
+     */
+    @Test
+    void sendFailsFastAfterAsyncSendFailureIsRecorded() {
+        ProducerStub producer = new ProducerStub();
+
+        TestingKafkaTransactionSender sender = createSender(producer);
+        sender.beginTransaction(TRANSACTION_ID);
+        sender.send(record());
+        producer.completePendingSends(new RuntimeException("async send failed"));
+
+        KafkaConnectorException exception =
+                Assertions.assertThrows(KafkaConnectorException.class, () -> sender.send(record()));
+
+        Assertions.assertEquals(
+                KafkaConnectorErrorCode.PRODUCE_DATA_FAILED.getCode(),
+                exception.getSeaTunnelErrorCode().getCode());
+        Assertions.assertEquals(
+                1, producer.sendCount, "no further record may be handed to the producer");
+    }
+
+    /**
+     * An asynchronous send failure is scoped to the transaction that produced it. After the engine
+     * aborts that transaction and opens a new one on the same sender, checkpoints must succeed
+     * again; otherwise one transient broker error would block every later checkpoint.
+     *
+     * <p>The recovered transaction is deliberately empty and reports txnStarted=false, so it also
+     * proves the record counter was reset: a stale counter would raise TRANSACTION_NOT_STARTED.
+     */
+    @Test
+    void senderRecoversAfterFailedTransaction() {
+        ProducerStub failingProducer = new ProducerStub();
+        failingProducer.transactionStarted();
+        failingProducer.failPendingSendsWith(new RuntimeException("async send failed"));
+        ProducerStub healthyProducer = new ProducerStub();
+
+        TestingKafkaTransactionSender sender = createSender(failingProducer, healthyProducer);
+        sender.beginTransaction(TRANSACTION_ID);
+        sender.send(record());
+        Assertions.assertThrows(KafkaConnectorException.class, sender::prepareCommit);
+        sender.abortTransaction();
+
+        sender.beginTransaction(NEXT_TRANSACTION_ID);
+
+        Optional<KafkaCommitInfo> commitInfo = Assertions.assertDoesNotThrow(sender::prepareCommit);
+        Assertions.assertTrue(commitInfo.isPresent());
+        Assertions.assertEquals(NEXT_TRANSACTION_ID, commitInfo.get().getTransactionId());
+        Assertions.assertFalse(commitInfo.get().isTxnStarted());
+    }
+
+    /**
+     * Each transactional ID must be fenced by its own producer: changing the ID on a reused
+     * producer can retain a non-zero epoch indefinitely and spin the cleanup loop forever.
+     */
     @Test
     void abortTransactionUsesFreshProducerForEachTransactionalId() {
         KafkaInternalProducer<byte[], byte[]> existingTransaction =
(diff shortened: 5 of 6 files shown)


nzw921rx (approved):
Thanks for working on this. I reviewed the current head `980509a040c4`.

## Overall Assessment

The `flush()`-before-state-capture change correctly fixes the race described in #11534, and I do not see a PR-introduced blocker.

For a successfully completed transactional send, Kafka cannot drain the produce batch until `AddPartitionsToTxn` succeeds. That success sets `transactionStarted = true`, while `flush()` waits for the previously submitted batch to complete. Capturing `txnStarted`, producer ID, and epoch after the flush therefore prevents the committer from resuming the transaction with a stale `false` value and incorrectly skipping `EndTxn`.

The unit tests cover the important ordering, the fail-fast branch, and the legitimate empty-transaction path. The existing E2E test is also the real reproducer for the reported loss, although this PR only adds documentation to that test.

## Non-blocking follow-up

One precision point: `flush()` waits until sends complete, but completion may be either successful or exceptional. Once any partition has been added successfully, `txnStarted` may remain `true` even if a later asynchronous send fails.

Therefore, the new guard detects the subset where no transaction registration succeeded; it does not detect every asynchronous send failure. This broader error-propagation gap predates this PR, so I do not consider it a blocker for #11541. However, the comments and PR description should avoid implying complete async-send failure detection, and the connector should follow up by recording callback failures and checking them after `flush()`.

From the code-review side, this looks good to me.

DanielLeens (commented):
Outdated.

fatmanverse:
Addressed the non-blocking async-send follow-up in `9360cde36`:

- register a Kafka send callback and retain the first asynchronous failure;
- check that failure immediately after `flush()`, before reading `txnStarted`;
- propagate it as `PRODUCE_DATA_FAILED` (`KAFKA-09`) with the original cause;
- add a regression test for `txnStarted = true` with a failed callback;
- narrow the `TRANSACTION_NOT_STARTED` wording to transaction registration only.

The branch was also rebased onto the latest `dev`, which includes the upstream fixes for the unrelated dead-link and `MetricsApiTest` CI failures.

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `9360cde36c0e` from scratch.

# What this PR fixes
- User pain: With `semantics = EXACTLY_ONCE`, the Kafka sink intermittently drops the first record of a transaction (e.g. `testKafkaToKafkaExactlyOnceOnStreaming` seeing 9/10 records), because `prepareCommit()` could capture `isTxnStarted() = false` before the async `AddPartitionsToTxn` was acknowledged, causing the committer to skip `EndTxn` and let the broker-side transaction time out.
- Fix approach: `KafkaTransactionSender#prepareCommit()` now calls `kafkaProducer.flush()` before reading `isTxnStarted()`, registers a send callback that captures async send failures into an `AtomicReference` and rethrows them as `PRODUCE_DATA_FAILED`, and fails fast with `TRANSACTION_NOT_STARTED` if records exist but the transaction still reports not-started; unit tests plus a regression-guard comment on the e2e test back this up.
- One-line summary: The root-cause fix is correct and well-tested for the single-transaction path, but `asyncSendException` is never reset on `beginTransaction`/abort, so one async failure permanently poisons the sender across all subsequent transactions — that must be fixed before merge, alongside smaller robustness and docs gaps (fail-fast in `send()`, logging of discarded secondary failures, KAFKA-08/09 error-code docs).

# Runtime chain I rechecked
```text
Checkpoint barrier (engine)
  -> KafkaSinkWriter.prepareCommit()
    -> KafkaTransactionSender.prepareCommit()  KafkaTransactionSender.java:83
      -> kafkaProducer.flush()                 KafkaTransactionSender.java:91  (blocks until pending sends + AddPartitionsToTxn complete)
      -> checkAsyncSendException()             KafkaTransactionSender.java:92 -> :118  (rethrows callback failure captured in send() at :64-70)
      -> kafkaProducer.isTxnStarted()          KafkaTransactionSender.java:93  (now read post-flush)
      -> throw TRANSACTION_NOT_STARTED if recordNumInTransaction > 0 && !txnStarted  KafkaTransactionSender.java:94-109
      -> new KafkaCommitInfo(txnId, props, producerId, epoch, txnStarted)  KafkaTransactionSender.java:110-116
  -> engine snapshotState / notifyCheckpointComplete
    -> KafkaSinkCommitter.commit(KafkaCommitInfo)  (resumes txn; performs EndTxn only when commitInfo.isTxnStarted() == true)
```

# Findings
**Issue 1: asyncSendException is never cleared, permanently poisoning the sender after the first async failure across transaction boundaries**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:121`
- Why it matters: - Root cause: `asyncSendException.compareAndSet(null, exception)` records the first async failure, and `checkAsyncSendException()` reads it with `asyncSendException.get()` and throws — but nothing ever resets the reference back to null. The diff does not touch `beginTransaction()` or `abortTransaction()`, so the stale exception survives transaction rollover.
- Impact: under EXACTLY_ONCE on Zeta/Flink, a single transient send failure (e.g. a brief broker leader election producing a retriable `TimeoutException`) fails the checkpoint as intended, but after the engine aborts the transaction and the writer begins a fresh transaction on the same `KafkaTransactionSender` instance, every subsequent `prepareCommit()` re-throws the same stale exception. The job can never complete another checkpoint until the task is fully restarted, turning a one-off transient error into a permanent checkpoint-failure loop.
- Fix: clear the reference at transaction boundaries — set `asyncSendException.set(null)` in `beginTransaction(String)` and `abortTransaction()` (and/or use `asyncSendException.getAndSet(null)` inside `checkAsyncSendException()` so the throw consumes the recorded failure). Add a unit test asserting that after a failed `prepareCommit()` a subsequent begin/send/prepareCommit cycle with successful sends succeeds.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:121          Exception exception = asyncSendException.get();`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:54  private final AtomicReference<Exception> asyncSendException = new AtomicReference<>();`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:67-69  (metadata, exception) -> { if (exception != null) { asyncSendException.compareAndSet(null, exception); } }`
- Suggested fix: `asyncSendException` is read here but never cleared, and this PR doesn't reset it in `beginTransaction()`/`abortTransaction()` either. After one transient async send failure fails a checkpoint, the engine will abort and open a new transaction on this same sender instance — and every subsequent `prepareCommit()` will re-throw the stale exception, permanently blocking checkpoints until a full task restart. Please consume the failure with `asyncSendException.getAndSet(null)` here and also clear it in `beginTransaction()`/`abortTransaction()`, plus add a unit test covering recovery after a failed prepareCommit.
- Severity: High

**Issue 2: New error codes KAFKA-08 and KAFKA-09 are not documented in the Error Quick Reference / connector docs**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/exception/KafkaConnectorErrorCode.java:36`
- Why it matters: - Root cause: this PR introduces two new user-visible error codes, `TRANSACTION_NOT_STARTED (KAFKA-08)` and `PRODUCE_DATA_FAILED (KAFKA-09)`, in `KafkaConnectorErrorCode`, but the changed-files list contains no documentation update.
- Impact: users hitting these errors under `semantics = EXACTLY_ONCE` will search the error quick reference (`docs/en/connector-v2/Error-Quick-Reference.md`) and the Kafka sink doc and find nothing; both codes indicate a deliberately failed checkpoint that operators need to understand (retry vs. investigate broker health), so undocumented codes will generate confusion and issues.
- Fix: add KAFKA-08 and KAFKA-09 entries (code, description, typical cause/remedy) to the error quick reference, and add a short note to `docs/en/connector-v2/sink/Kafka.md` (and the zh counterpart) explaining that under EXACTLY_ONCE an async send failure now fails the checkpoint instead of silently dropping records, since the PR description itself declares this an operational behavior change.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/exception/KafkaConnectorErrorCode.java:36  PRODUCE_DATA_FAILED("KAFKA-09", "Kafka failed to produce data");`
- Suggested fix: These two new error codes (KAFKA-08, KAFKA-09) are user-visible, but this PR doesn't update the docs. Please add entries to `docs/en/connector-v2/Error-Quick-Reference.md` and a short note in the Kafka sink doc explaining that under `EXACTLY_ONCE` an async send failure now fails the checkpoint (with these codes) instead of silently losing records — that's the operational change called out in the PR description and operators will need it when triaging.
- Severity: Medium

**Issue 3: send() keeps enqueueing records after an async failure has already been recorded instead of failing fast**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:66`
- Why it matters: - Root cause: the new callback in `send()` only records the exception; `send()` itself never checks `asyncSendException`, so after the first failure the writer continues buffering and sending records into a transaction that is already doomed to fail at `prepareCommit()`.
- Impact: between the failing send and the next checkpoint, potentially millions of rows are serialized, batched, and transmitted to the broker for nothing, wasting producer buffer memory and network bandwidth, and delaying the point at which the user learns the pipeline is broken until the checkpoint interval elapses. Other SeaTunnel transactional sinks fail on the write path when a prior async error is known.
- Fix: call `checkAsyncSendException()` (or an equivalent cheap null-check on the AtomicReference) at the top of `send(ProducerRecord)` so the first row written after a recorded failure throws `PRODUCE_DATA_FAILED` immediately, surfacing the error on the write path rather than only at checkpoint time.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:70                          asyncSendException.compareAndSet(null, exception);`; `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:64-70  kafkaProducer.send(
                producerRecord,
                (metadata, exception) -> {
                    if (exception != null) {
                        asyncSendException.compareAndSet(null, exception);
                    }
                });`
- Suggested fix: Since the callback records the first failure into `asyncSendException`, could we also check it at the start of `send()`? Right now, after a send fails, the writer keeps serializing and transmitting records into a transaction that is guaranteed to fail at the next `prepareCommit()`. Failing fast on the write path would save producer buffer/network resources and surface the error immediately instead of after the checkpoint interval.
- Severity: Medium

**Issue 4: Tests wire internal state via reflection and miss the sender-reuse-after-failure scenario**
- Location: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSenderTest.java:169`
- Why it matters: - Root cause: `KafkaTransactionSenderTest#createSender` injects `kafkaProducer`, `transactionId`, and `recordNumInTransaction` through `Field#setAccessible(true)`, bypassing `beginTransaction()` entirely. This couples the tests to private field names (any rename silently breaks them with `NoSuchFieldException`) and means the transaction-lifecycle reset behavior is never exercised.
- Impact: the exact gap flagged in the main finding — a recorded `asyncSendException` surviving into the next transaction — is invisible to this suite, because no test performs a begin → send-failure → prepareCommit-throws → begin → send-success → prepareCommit sequence on one sender instance.
- Fix: add a package-private constructor or `@VisibleForTesting` seam that accepts a `KafkaInternalProducer` supplier so `beginTransaction()` can run under test, and add a lifecycle test asserting a fresh transaction after a failed checkpoint can commit cleanly. This would also lock in the reset semantics of `recordNumInTransaction` that the new `recordNumInTransaction > 0 && !txnStarted` guard depends on.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSenderTest.java:176          Field field = KafkaTransactionSender.class.getDeclaredField(fieldName);`; `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSenderTest.java:169-172  setField(sender, "kafkaProducer", producer);
        setField(sender, "transactionId", TRANSACTION_ID);
        setField(sender, "recordNumInTransaction", recordNumInTransaction);`; `seatunnel-connectors-v2/connector-kafka/src/test/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSenderTest.java:176-178  Field field = KafkaTransactionSender.class.getDeclaredField(fieldName);
        field.setAccessible(true);
        field.set(target, value);`
- Suggested fix: Injecting private fields via reflection bypasses `beginTransaction()`, so the transaction-lifecycle behavior (including whether `asyncSendException` and `recordNumInTransaction` are reset for a new transaction) is never tested and the tests will break silently on any field rename. Consider a package-private/`@VisibleForTesting` constructor taking a producer supplier, and add a test for the begin → async-failure → prepareCommit-throws → begin → success → prepareCommit sequence on one sender instance.
- Severity: Low

**Issue 5: Subsequent async send failures after the first are silently discarded with no logging**
- Location: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:70`
- Why it matters: - Root cause: `compareAndSet(null, exception)` intentionally keeps only the first async failure, but losing exceptions when the CAS fails means all follow-up failures vanish without a trace — the class has no logger and the callback does nothing on the losing path.
- Impact: when diagnosing a broker-side incident (e.g. multiple partitions failing with different errors — `RecordTooLargeException` on one, `NotEnoughReplicasException` on another), operators only ever see the first exception in the checkpoint failure, which can send troubleshooting in the wrong direction.
- Fix: on the losing CAS branch, log the suppressed exception at WARN (e.g. `log.warn("Suppressed additional async send failure for transaction [{}]", transactionId, exception)`) via `@Slf4j`, or attach it as a suppressed exception to the recorded one, consistent with how SeaTunnel connectors typically surface secondary failures.
- Evidence: `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:70                          asyncSendException.compareAndSet(null, exception);`
- Suggested fix: Keeping only the first failure is reasonable, but when `compareAndSet` loses, the additional exception is dropped with no trace. In a multi-partition broker incident the second/third failure may carry the more actionable error. Please log the suppressed exception at WARN (with the transaction id) on the losing branch, or add it via `Throwable#addSuppressed` to the recorded one.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: asyncSendException is never cleared, permanently poisoning the sender after the first async failure across transaction boundaries

- Issue 2: New error codes KAFKA-08 and KAFKA-09 are not documented in the Error Quick Reference / connector docs
- Issue 3: send() keeps enqueueing records after an async failure has already been recorded instead of failing fast
- Issue 4: Tests wire internal state via reflection and miss the sender-reuse-after-failure scenario
- Issue 5: Subsequent async send failures after the first are silently discarded with no logging

fatmanverse:
Thanks for the detailed re-review — the blocking finding was accurate and is fixed in `fb7ab542`, along with all four follow-ups. Two of them I implemented a little differently than suggested, so let me walk through the reasoning in case you see it another way.

### Issue 1 (blocking): async failure is now scoped to its transaction

Confirmed — nothing ever cleared `asyncSendException`, so a transient send error would have turned into a permanent checkpoint-failure loop. It is now reset in `beginTransaction(String)`.

Of the two options you offered I went with the reset rather than `getAndSet(null)` in `checkAsyncSendException()`, because of how it combines with Issue 3. Once `send()` also inspects the same reference, a consume-once read means the first `send()` throws and clears the flag, and the following `prepareCommit()` then sees `null`, treats the transaction as healthy, and commits a transaction that already lost a record. Keeping the failure until the next transaction begins avoids that interaction.

For the same reason I left `abortTransaction()` untouched: every `beginTransaction()` builds a fresh producer through `getTransactionProducer()`, so callbacks from the previous producer can't reach the new transaction's state. With `beginTransaction()` as the single reset point, the failure's lifetime is exactly one transaction, which is what lets `send()` and `prepareCommit()` share the reference safely.

The recovery path you asked for is covered by `senderRecoversAfterFailedTransaction`: begin → async failure → prepareCommit throws → abort → begin → successful send → commit succeeds.

### Issue 3: fail fast on the write path

Good call. `send()` now calls `checkAsyncSendException()` before handing the record to the producer, so the error surfaces on the write path instead of a checkpoint interval later. `sendFailsFastAfterAsyncSendFailureIsRecorded` also asserts that no further record reaches the producer.

### Issue 4: reflection removed from the tests

Added a package-private `@VisibleForTesting` constructor taking a `TransactionProducerFactory`, using the project's shaded Guava annotation as `KafkaSourceSplitEnumerator` does. The suite now drives the real `beginTransaction()` → `send()` → `prepareCommit()` lifecycle with no reflection, so the `recordNumInTransaction` reset semantics the new guard relies on are exercised rather than injected.

### Issue 5: secondary failures are logged

The callback is now a named `onSendCompleted` method that logs at WARN with the transaction id when the CAS loses, so a multi-partition incident stays diagnosable.

I went with logging rather than `Throwable#addSuppressed` here: the callback runs on the producer's sender thread while the task thread may already be reading that exception in `prepareCommit()`, so mutating it would introduce a data race.

### Issue 2: docs

One thing worth flagging on the paths: I couldn't find `docs/en/connector-v2/Error-Quick-Reference.md` in the tree, and there doesn't seem to be an error-code quick reference anywhere under `docs/` — the Kafka sink doc sits at `docs/en/connectors/sink/Kafka.md`. The existing KAFKA-01 … KAFKA-07 codes are undocumented too, so building that reference looks like a larger piece of work than this PR should carry. Happy to open a separate PR for it if maintainers agree it's worth having, though it would need to cover well beyond Kafka.

The operational note itself is in: the Semantics section of both `docs/en/connectors/sink/Kafka.md` and `docs/zh/connectors/sink/Kafka.md` now explains that under `EXACTLY_ONCE` an async send failure fails the checkpoint — naming both codes — instead of silently dropping records.

### Verification

`./mvnw test -pl seatunnel-connectors-v2/connector-kafka` → 65 tests, 0 failures (`KafkaTransactionSenderTest` 4 → 6). `spotless:apply` clean.

DanielLeens (commented):
Outdated.

DanielLeens (approved):
Thanks for the update. I re-reviewed the current head from the Kafka writer checkpoint path through the transaction sender lifecycle. The previous blocker from @SEZ9 is fixed in this version, and I do not see a new source-level blocker.

# What this PR fixes
- User pain point: with `semantics = EXACTLY_ONCE`, a checkpoint could capture `txnStarted=false` before Kafka finished registering the first record's transaction, causing the committer to skip `EndTxn` and leaving records invisible to `read_committed` consumers.
- Fix approach: flush pending sends before capturing transaction state, record async send failures, fail fast when records exist but the transaction still is not started, and reset async failure state per transaction.
- In one sentence: this is a focused correctness fix for Kafka exactly-once transaction commit state.

# 1. Code Review

## 1.1 Core Logic
The runtime path I traced is:

```text
Checkpoint barrier
  -> KafkaSinkWriter.prepareCommit()
      -> KafkaTransactionSender.prepareCommit()
          -> kafkaProducer.flush()
          -> checkAsyncSendException()
          -> kafkaProducer.isTxnStarted()
          -> if records exist and txnStarted=false, throw KAFKA-08
          -> build KafkaCommitInfo(txnId, producerId, epoch, txnStarted)
  -> KafkaSinkCommitter.commit(KafkaCommitInfo)
      -> if txnStarted=true, resume transaction and perform EndTxn

New transaction
  -> beginTransaction(transactionId)
      -> create a transaction producer
      -> beginTransaction()
      -> reset recordNumInTransaction and asyncSendException

Async send failure
  -> send callback records the first failure
  -> later failures are logged
  -> next send() or prepareCommit() fails with KAFKA-09
```

Key observations:
- The normal exactly-once checkpoint path does hit the `flush()` before `isTxnStarted()` change.
- The stale `asyncSendException` problem raised in the previous review is fixed by resetting it in `beginTransaction()`.
- `send()` now fails fast after a recorded async failure, so the writer does not keep filling a doomed transaction.
- The tests now exercise the real `beginTransaction()` lifecycle through a producer factory instead of setting private fields by reflection.

## 1.2 Compatibility
Fully compatible. This does not change public Kafka sink options, checkpoint state shape, protocol, or normal successful behavior. The new error codes only make previously lossy/error paths explicit.

The checkpoint-time `flush()` is the correct trade-off for exactly-once correctness. The write-path failure check is cheap, and fail-fast behavior reduces wasted buffering/network traffic after an async send error.

`KAFKA-08` and `KAFKA-09` now carry clear transaction context. Additional async failures are logged instead of disappearing silently.

Formal findings: no blocking issue found.

# 2. Code Quality

## 2.1 Style
`git diff --check apache/dev...HEAD` is clean, and the new comments explain the transactional invariants well.

## 2.2 Tests and Stability
The unit tests cover flush-before-state capture, fail-fast when records exist but the transaction is not started, async send failure propagation, empty transactions, write-path fail-fast, and sender reuse after a failed transaction.

Test stability rating: stable. The new unit tests use mocked producer callbacks and do not rely on fixed sleeps, ports, external service timing, or test order. I did not run local tests in this batch; this review is based on current-head source analysis and GitHub CI evidence, as requested.

## 2.3 Docs
The Kafka sink docs in both English and Chinese now explain the exactly-once failure behavior and the new KAFKA-08/KAFKA-09 cases.

# 3. Architecture
This is a precise fix. The transactional state is captured at the right lifecycle boundary, and async send failures are scoped to a single transaction.

# 4. Findings Summary

| --- | --- | --- | --- |
| - | No blocking issue found | - | - |

# 5. Merge Conclusion
### Conclusion: can merge

   - None.

2. Non-blocking suggestions
   - None from this review.

Overall, this version looks good to me. The previous transaction-reuse blocker is closed, current GitHub Build is green, and I am happy to approve this from Daniel's side.

davidzollo (dismissed):
+1
LGTM

davidzollo:
image
Please solve conflicts. thx

DanielLeens:
Thanks @davidzollo for catching the current merge gate.

From Daniel's source review side, I still do not see a reopened Kafka exactly-once blocker on this unchanged head, but GitHub is now reporting merge conflicts. Could the contributor please sync with the latest `dev` and resolve the conflicts first? After that, the checks should be rerun against the resolved head before merge.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `4f75665dfac580a15d1fa895ffd227d63fc8a164` from the Kafka exactly-once writer path through the transaction sender lifecycle and the regression tests.

One important update from my previous review: the source-level conclusion is still clean for me, but the CI/merge-gate fact has changed. The previous Daniel review was based on a green Build at that time; the current required `Build` check for this head is `CANCELLED` (`https://github.com/apache/seatunnel/runs/91204480364`), and the branch is now diverged from `dev` (`ahead_by=2`, `behind_by=7`). So I am not treating this as ready-to-merge yet. The lowest-cost next step is to sync with the latest `dev` and rerun the Build check first.

# What this PR fixes
- User pain point: with `semantics = EXACTLY_ONCE`, the Kafka sink could capture `txnStarted=false` before Kafka finished registering the first record's transaction, so the committer skipped `EndTxn` and the first transaction's records stayed invisible to `read_committed` consumers.
- Fix approach: `KafkaTransactionSender.prepareCommit()` now flushes pending sends before reading transaction state, checks asynchronous send failures, fails fast when records exist but the transaction is still not started, and resets per-transaction error state in `beginTransaction()`.
- In one sentence: this is a focused correctness fix for Kafka exactly-once checkpoint commit state and async send failure propagation.

# 1. Code Review

## 1.1 Core Logic Analysis
Main changes:
- `seatunnel-connectors-v2/connector-kafka/src/main/java/org/apache/seatunnel/connectors/seatunnel/kafka/sink/KafkaTransactionSender.java:56-62` adds a thread-safe `AtomicReference` for the first asynchronous send failure of the current transaction.
- `KafkaTransactionSender.java:72-79` checks an already recorded async failure before sending more records.
- `KafkaTransactionSender.java:102-110` resets `recordNumInTransaction` and `asyncSendException` for each new transaction.
- `KafkaTransactionSender.java:114-138` flushes pending async sends before reading `isTxnStarted()`, then refuses to emit a lossy commit info when records exist but Kafka still reports the transaction as not started.

Before, `prepareCommit()` could capture `txnStarted=false` before pending Kafka sender-thread work completed:

```java
boolean txnStarted = kafkaProducer.isTxnStarted();
KafkaCommitInfo kafkaCommitInfo =
        new KafkaCommitInfo(..., txnStarted);
```

After this PR, the sender drains pending sends first and rejects unsafe states:

```java
kafkaProducer.flush();
checkAsyncSendException();
boolean txnStarted = kafkaProducer.isTxnStarted();
if (recordNumInTransaction > 0 && !txnStarted) {
    throw new KafkaConnectorException(
            KafkaConnectorErrorCode.TRANSACTION_NOT_STARTED, ...);
}
```

Complete runtime path:

```text
SeaTunnel Kafka sink task writes records
  -> KafkaSinkWriter.write()
      -> KafkaTransactionSender.send()
          -> checkAsyncSendException()
          -> kafkaProducer.send(record, onSendCompleted)
          -> async callback records the first send failure

Checkpoint barrier reaches sink writer
  -> KafkaSinkWriter.prepareCommit()
      -> KafkaTransactionSender.prepareCommit()
          -> kafkaProducer.flush()
              -> waits for pending Produce / AddPartitionsToTxn work
          -> checkAsyncSendException()
              -> throws KAFKA-09 if a callback already recorded a failed send
          -> kafkaProducer.isTxnStarted()
          -> records exist && txnStarted=false
              -> throw KAFKA-08 instead of building a commit info that would skip EndTxn
          -> build KafkaCommitInfo(transactionId, producerId, epoch, txnStarted)

Checkpoint completes
  -> KafkaSinkCommitter.commit(KafkaCommitInfo)
      -> if txnStarted=true, resume transaction and perform EndTxn
      -> if txnStarted=false, empty transactions remain safe

Next transaction
  -> KafkaTransactionSender.beginTransaction()
      -> creates a fresh transactional producer
      -> resets recordNumInTransaction and asyncSendException
```

Key findings:
- The normal exactly-once checkpoint path does hit the new `flush()` before `isTxnStarted()` logic.
- The earlier stale `asyncSendException` blocker is fixed for the current head by resetting it in `beginTransaction()`.
- The write path now fails fast after a recorded async send failure, avoiding extra buffering into a doomed transaction.
- The added tests cover transaction state capture, async send failure propagation, empty transactions, fail-fast writes, and recovery after a failed transaction.

## 1.2 Compatibility
Fully compatible from a source/API/config perspective. This PR does not rename options, change defaults, alter checkpoint state shape, or introduce a new protocol. It turns previously lossy Kafka exactly-once failure paths into explicit checkpoint failures with `KAFKA-08` / `KAFKA-09`.

The extra `flush()` happens only at checkpoint prepare time and is the right trade-off for exactly-once correctness. The `AtomicReference` check in `send()` is cheap, and fail-fast behavior reduces wasted producer buffering and network traffic after an async send failure.

The current head surfaces unsafe commit state via `KAFKA-08`, async send failure via `KAFKA-09`, and logs additional async failures instead of silently dropping them.

# 2. Code Quality

## 2.1 Style
The changed Kafka sender code is readable, and the key comments explain the transactional invariants rather than restating the code. `git diff --check upstream/dev...HEAD` is clean in my local review worktree.

## 2.2 Tests and Stability
The new unit tests in `KafkaTransactionSenderTest.java:57-226` cover:
- flush-before-transaction-state capture;
- unsafe `records > 0 && txnStarted=false`;
- async send failure after transaction start;
- empty transaction behavior;
- fail-fast writes after an async failure;
- sender recovery after a failed transaction;
- fresh producer use for transaction cleanup.

Test stability rating: stable. The tests use mocked producer callbacks and do not rely on fixed sleeps, local ports, external Kafka timing, system clock, test order, or shared static state. I did not run local Maven tests in this batch; this review is based on current-head source analysis and GitHub CI evidence.

## 2.3 Docs
The English and Chinese Kafka sink docs now describe the exactly-once failure behavior and list `KAFKA-08` / `KAFKA-09` with operator-facing guidance.

# 3. Architecture
This is a precise fix rather than a workaround. It keeps the responsibility inside the Kafka transaction sender, where the transaction state and async producer callbacks are available, without changing the engine checkpoint contract.

# 4. Findings Summary

| --- | --- | --- | --- |

# 5. Merge Conclusion
### Conclusion: can merge after CI / merge gate is cleared

   - No source-code blocker from Daniel's side in this round.
   - The current required `Build` check is `CANCELLED`, so the PR still needs a fresh successful Build before merge.

   - None from this review.

Overall, the source change looks good to me. The remaining action is not to patch unrelated code in this PR, but to sync with the latest `dev` and rerun the required Build check. If it still fails after the sync, please share the new failing check link and I can help narrow it down against the updated head.

fatmanverse:
@SEZ9 The blocking asyncSendException lifecycle issue from your changes-requested review has been addressed in the current head dcb55d3b, together with the requested regression coverage and error documentation. The source has since been re-reviewed without a new code blocker; the CI rerun is still in progress. Please re-review the current head or dismiss the stale changes-requested review when convenient.

DanielLeens (commented):
Thanks for the update. I re-reviewed the latest head `dcb55d3b530d4075be1820a4b97a898e076055e7` from the Kafka exactly-once writer path, the transaction sender lifecycle, the regression tests, and the current PR diff against the latest `dev`.

One important update from my previous Daniel review: the earlier CI guidance is outdated now. The required `Build` check is successful on the current head: https://github.com/apache/seatunnel/runs/91606197305

I also rechecked the source delta since the last Daniel round that was already source-clean (`4f75665dfac580a15d1fa895ffd227d63fc8a164`). The Kafka sender / tests / docs touched by this PR are unchanged across that delta; the latest head is the merge with current `dev`, and I do not see a reopened source-level blocker.

# What this PR fixes
- User pain point: with `semantics = EXACTLY_ONCE`, the Kafka sink could capture `txnStarted=false` before Kafka finished registering the first record's transaction, so the committer skipped `EndTxn` and the first transaction's records stayed invisible to `read_committed` consumers.
- Fix approach: `KafkaTransactionSender.prepareCommit()` flushes pending sends before reading transaction state, checks asynchronous send failures, fails fast when records exist but the transaction is still not started, and resets per-transaction async failure state in `beginTransaction()`.
- In one sentence: this is a focused correctness fix for Kafka exactly-once checkpoint commit state and async send failure propagation.

# 1. Code Review

## 1.1 Core Logic Analysis
Runtime path rechecked:

```text
SeaTunnel Kafka sink task writes records
  -> KafkaSinkWriter.write()
      -> KafkaTransactionSender.send()
          -> checkAsyncSendException()
          -> kafkaProducer.send(record, onSendCompleted)
          -> async callback records the first send failure

Checkpoint barrier
  -> KafkaSinkWriter.prepareCommit()
      -> KafkaTransactionSender.prepareCommit()
          -> kafkaProducer.flush()
          -> checkAsyncSendException()
          -> kafkaProducer.isTxnStarted()
          -> if records exist && txnStarted=false, throw KAFKA-08
          -> build KafkaCommitInfo(txnId, producerId, epoch, txnStarted)

Checkpoint completes
  -> KafkaSinkCommitter.commit(KafkaCommitInfo)
      -> if txnStarted=true, resume transaction and perform EndTxn

Next transaction
  -> KafkaTransactionSender.beginTransaction()
      -> creates a fresh transactional producer
      -> resets recordNumInTransaction and asyncSendException
```

Key findings:
- The normal exactly-once checkpoint path does hit the `flush()` before `isTxnStarted()` change.
- The earlier stale `asyncSendException` blocker stays fixed on the current head.
- The write path still fails fast after a recorded async send failure.
- The current head is now up to date with `dev` from the compare perspective (`ahead_by=3`, `behind_by=0`, compare status: `ahead`).

## 1.2 Compatibility
Fully compatible from a source/API/config perspective. This PR does not rename options, change defaults, alter checkpoint state shape, or introduce a new runtime protocol. It turns previously lossy Kafka exactly-once failure paths into explicit checkpoint failures with `KAFKA-08` / `KAFKA-09`.

The extra `flush()` still happens only at checkpoint prepare time and remains the correct trade-off for exactly-once correctness. The `AtomicReference` failure check in `send()` is cheap, and fail-fast behavior still reduces wasted producer buffering after an async send error.

The current head still surfaces unsafe commit state via `KAFKA-08`, async send failure via `KAFKA-09`, and logs additional async failures instead of silently dropping them.

# 2. Code Quality

## 2.1 Style
The changed Kafka sender code remains readable, and the comments still explain the transactional invariants well.

## 2.2 Tests and Stability
The unit tests still cover flush-before-state capture, async send failure propagation, write-path fail-fast, empty transactions, and recovery after a failed transaction.

Stability rating: stable. I do not see a new flaky-test pattern in the current unit/E2E coverage.

## 2.3 Docs
The English and Chinese Kafka sink docs still describe the exactly-once failure behavior and the `KAFKA-08` / `KAFKA-09` cases.

# 3. Architecture
This remains a precise fix rather than a workaround. The transactional state is still captured at the correct lifecycle boundary, and the per-transaction async failure scoping remains coherent on the current head.

# 4. Findings Summary

| --- | --- | --- | --- |

# 5. Merge Conclusion
### Conclusion: can merge

1. Blocking items

- None from this round.

Overall, the source change still looks good to me, and the previous CI blocker is cleared now.

Because I only have read-level review controls on this repository, this follow-up review is my current Daniel-side technical conclusion for the latest head. If GitHub still needs a formal approval / dismissal step to clear repository-level merge gates, a write-capable maintainer or reviewer will need to handle that final action.

DanielLeens (approved):

My last review on this head already concluded **"can merge"** with an empty blocking list, and noted that the previous CI blocker was cleared — but I left it as a comment rather than a review approval, so the PR still shows as awaiting review. Rechecking the current head `dcb55d3b`: **all checks are passing** (3/3 SUCCESS), so there is nothing left to wait on and I am converting that conclusion into a formal approve.

Re-confirmed on this head:
- The extra `flush()` still happens only at checkpoint-prepare time, which is the correct boundary for exactly-once correctness, and transactional state is still captured at the right point in the lifecycle.
- Unsafe commit state still surfaces as `KAFKA-08` and async send failure as `KAFKA-09`; additional async failures are logged rather than silently dropped. The `AtomicReference` failure check in `send()` remains cheap and the fail-fast behavior still avoids wasted producer buffering after an async error.
- Compatibility: no option renamed, no default changed, no checkpoint state shape altered, no new runtime protocol. Fully compatible.
- Test stability: the unit tests still cover flush-before-state-capture, async send failure propagation, write-path fail-fast, empty transactions, and recovery after a failed transaction, with no timing or ordering dependencies. **Stable.** Nothing in this PR weakens an existing assertion or check.
- EN and ZH Kafka sink docs both still describe the exactly-once failure behavior and the `KAFKA-08` / `KAFKA-09` cases consistently.

One note for accuracy: in my previous comment I said I only had read-level review controls on this repository and that a write-capable maintainer would need to handle the formal approval. That was wrong — I am able to submit a review, and I have done so here. A committer's approval is of course still required for the merge itself under the project's own rules; that is a separate step from this review.

Thanks for the careful work on this — losing the first record of a transaction is exactly the kind of exactly-once bug that is painful to diagnose in production, and the fix is well placed.

nzw921rx:
@davidzollo If you have time, please review this PR. thanks very much🚀

goutamadwant:
@fatmanverse reviewed the latest changes against the Kafka transaction and checkpoint lifecycle.

The flush before reading txnStarted addresses the reported first record race. Async send failures are also retained per transaction and checked before commit, while the empty transaction path remains unchanged.

One small documentation point suggestion: saying both errors abort the transaction is not fully accurate for every engine. It may be better to say they fail the checkpoint and prevent the transaction from being committed, then records are replayed from the last completed checkpoint after recovery.

No code correctness blocker from my review. Thanks for the changes and contribution!

goutamadwant (approved):

DanielLeens:
@goutamadwant Thanks for the extra pass and the doc feedback — you're right, and it's worth tightening.

I checked the actual abort semantics: `SinkWriter#abortPrepare()` is documented as **only invoked by the Spark engine** on a `prepareCommit()` failure (`seatunnel-api/.../sink/SinkWriter.java`, javadoc on `abortPrepare()`). `KafkaSinkWriter#abortPrepare()` does call `kafkaProducerSender.abortTransaction()`, but for Zeta/Flink that path isn't triggered when `KAFKA-08`/`KAFKA-09` is thrown — the checkpoint just fails, and the in-flight transaction is only cleaned up later, when the writer restarts from the last successful checkpoint and `KafkaSinkWriter`'s constructor calls `abortTransaction(checkpointId + 1)` on the stale transaction (or the broker times it out via `transaction.timeout.ms` in the meantime).

So "both errors abort the current transaction" is accurate for Spark but overstates it for Zeta/Flink. Your proposed wording ("fail the checkpoint and prevent the transaction from being committed, then records are replayed from the last completed checkpoint after recovery") matches the code for all three engines and is the better description.

This is doc-precision only, no behavior change, so I agree it doesn't block merge — just flagging so the wording can be nudged if there's another doc pass.

davidzollo (approved):
+1
LGTM

```

```

PR #11527, opened 2026-07-21
[Feature][Connector-V2] Add MongoDB timer flush

nzw921rx:
## Purpose

Support timer-driven flushing for the MongoDB sink on the Zeta engine so pending bulk requests can be written without waiting for the batch threshold, a new input record, or the next checkpoint.

## Changes

- Register the existing MongoDB bulk-write path with `SinkWriter.Context` for non-transaction writers.
- Keep timer flush disabled when `transaction = true` to preserve the checkpoint transaction boundary.
- Propagate MongoDB bulk-write failures through the timer flush action.
- Add unit coverage for timer action registration, bulk execution, transaction guarding, and failure propagation.
- Add an independent Zeta E2E test using a continuous MySQL CDC source, dynamic container port mapping, and a classpath-provided MySQL Connector/J.
- Document Zeta timer flush behavior and at-least-once delivery semantics in English and Chinese.

## Validation

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-mongodb -DskipIT -DskipUT=false -Dtest=MongodbWriterTest -DfailIfNoTests=false test`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-mongodb-e2e -DskipUT -DskipIT -DskipTests test-compile`
- `TEST_IN_PR=true RUN_ALL_CONTAINER=false RUN_ZETA_CONTAINER=true ./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-mongodb-e2e -DskipUT -DskipIT=false -Dit.test=MongodbTimerFlushIT -DfailIfNoTests=false verify`

PR #11542, opened 2026-07-23
[Fix][Zeta] Stabilize flaky Java 8 unit tests

nzw921rx:

Stabilize two flaky Java 8 unit tests observed in consecutive CI attempts of #11527.

## Changes

- Count assigned slots for the current job in `JobMasterTest#testCloseIdleTask` instead of using the worker-wide slot count. This prevents asynchronously released slots from other jobs from causing an exact-count timeout.
- Use force cancellation in `SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus`. The test verifies job queries after a master handoff, so it should not block on graceful checkpoint teardown on the promoted master.

## CI comparison

- Attempt 1 timed out in `JobMasterTest#testCloseIdleTask`.
- Attempt 2 timed out in `SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus` while waiting for `CancelJobOperation`.

The different failures across attempts, together with both tests passing in isolation before the change, indicate independent test flakiness rather than a failure caused by the MongoDB connector changes in #11527.

## Validation

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-engine/seatunnel-engine-server -DskipIT -DskipUT=false -Dtest=JobMasterTest#testCloseIdleTask -DfailIfNoTests=false test`
- `./mvnw -pl seatunnel-engine/seatunnel-engine-client -DskipIT -DskipUT=false -Dtest=SeaTunnelEngineClusterRoleTest#testWorkerIsFirstMemberThenGetJobDetailStatus -DfailIfNoTests=false test`

Both focused tests passed on Java 8.

### DIFF
--- a/seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java
+++ b/seatunnel-engine/seatunnel-engine-client/src/test/java/org/apache/seatunnel/engine/client/SeaTunnelEngineClusterRoleTest.java
@@ -480,7 +480,9 @@ public class SeaTunnelEngineClusterRoleTest {
                                                     && jobClient
                                                             .listJobStatus(true)
                                                             .contains("RUNNING")));
-            jobClient.cancelJob(jobId);
+            // This test verifies job queries after a master handoff, not graceful cancellation.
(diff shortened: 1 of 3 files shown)


davidzollo (approved):
Good job

PR #10121, opened 2025-11-26
[Feature][Connector-V2] add fluss Connector

dyp12:
### Purpose of this pull request

https://github.com/apache/seatunnel/issues/8816   add fluss sink first

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config)

PR #11544, opened 2026-07-23
[Feature][Connector-V2][Fluss] Support Fluss source connector

### LINKED ISSUE #8816: [Feature][Connector-v2] Support fluss source & sink
### Search before asking

### Description

https://github.com/alibaba/fluss

- add connector module
- add docs(en/zh)
- add e2e testcase

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

ss666:

Closes #8816.

This PR adds the **source** side of the Fluss connector. It lets SeaTunnel read from Fluss log tables, both as a bounded batch read and as an unbounded streaming read.

Main additions:

1. `FlussSource` / `FlussSourceFactory` — `SeaTunnelSource` implementation (`SupportParallelism`) and factory wiring, reusing the existing `connector-fluss` module.
2. `FlussSourceSplitEnumerator` — discovers table buckets, assigns them as splits, and supports snapshot/restore with stable split allocation across restarts.
3. `FlussSourceSplitReader` / `FlussSourceReader` / `FlussRecordEmitter` — poll records per bucket on the reader's fetcher thread, and handle bounded/unbounded end conditions.
4. `StartMode` + `FlussSourceOptions` — `earliest` / `latest` startup offset options.
5. `FlussTypeConverter` — maps Fluss types to SeaTunnel `SeaTunnelDataType`.
6. English & Chinese connector docs under `docs/{en,zh}/connectors/source/Fluss.md`, and registration in `plugin-mapping.properties` / `config/plugin_config`.

Yes — a new `Fluss` source connector becomes available (previously only the Fluss sink existed). It only adds a new capability; no existing option key, default, or behavior is changed.

Scoped to the `connector-fluss` / `connector-fluss-e2e` modules (the modules this PR changes):

*   `./mvnw spotless:apply` pass.
*   `./mvnw test -pl seatunnel-connectors-v2/connector-fluss` → 26/26, 0 failures
    (FlussSourceSplitEnumeratorTest 8, FlussSourceFactoryTest 12, FlussSourceSplitReaderTest 6)
*   Built the dist + connector for e2e:
    `./mvnw install -DskipTests -Dskip.spotless=true -pl seatunnel-dist,seatunnel-e2e/seatunnel-connector-v2-e2e/connector-fluss-e2e -am` → BUILD SUCCESS
*   `./mvnw verify -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-fluss-e2e -Dit.test=FlussSourceIT -DskipUT=true -DskipIT=false -DfailIfNoTests=false` → 4/4, 0 failures
    (testFlussSource, testFlussSourceAllTypes, testFlussSourceStreaming, testFlussSourceStreamingLatest)

### Check list

### DIFF
--- a/.github/workflows/labeler/label-scope-conf.yml
+++ b/.github/workflows/labeler/label-scope-conf.yml
@@ -143,6 +143,11 @@ file:
       - changed-files:
           - any-glob-to-any-file: seatunnel-connectors-v2/connector-file/**
           - all-globs-to-all-files: '!seatunnel-connectors-v2/connector-!(file)/**'
+fluss:
+  - all:
+      - changed-files:
+          - any-glob-to-any-file: seatunnel-connectors-v2/connector-fluss/**
+          - all-globs-to-all-files: '!seatunnel-connectors-v2/connector-!(fluss)/**'
 google-firestore:
   - all:
       - changed-files:
--- a/config/plugin_config
+++ b/config/plugin_config
@@ -46,6 +46,7 @@ connector-file-jindo-oss
 connector-file-s3
 connector-file-sftp
 connector-file-obs
+connector-fluss
 connector-google-sheets
 connector-google-firestore
 connector-google-bigtable
--- a/docs/en/connectors/source/Fluss.md
+++ b/docs/en/connectors/source/Fluss.md
@@ -1,0 +1,179 @@
+import ChangeLog from '../changelog/connector-fluss.md';
+
+# Fluss
+
+> Fluss source connector
+
+## Supported Engines
+
+> Spark<br/>
+> Flink<br/>
+> SeaTunnel Zeta<br/>
+
+## Key Features
+
+- [x] [batch](../../introduction/concepts/connector-v2-features.md)
+- [x] [stream](../../introduction/concepts/connector-v2-features.md)
+- [x] [exactly-once](../../introduction/concepts/connector-v2-features.md)
+- [ ] [column projection](../../introduction/concepts/connector-v2-features.md)
+- [x] [parallelism](../../introduction/concepts/connector-v2-features.md)
+- [ ] [support user-defined split](../../introduction/concepts/connector-v2-features.md)
+
+## Description
+
+The Fluss source reads rows from an existing Fluss table in batch or streaming jobs.
+
+The connector reads the table through the Fluss log scanner, one split per table bucket, so the
+read parallelism follows the number of buckets. Each record's change type (`INSERT`,
+`UPDATE_BEFORE`, `UPDATE_AFTER`, `DELETE`) is mapped to the corresponding SeaTunnel `RowKind`, so a
+log table is read as append-only inserts and a primary-key table is read as its changelog.
+
+:::caution Primary-key tables
+
+For a primary-key table the connector reads only the table's **changelog**, starting from the
+earliest available log offset; it does **not** read the KV snapshot first. It therefore captures
+ongoing changes (insert / update / delete) with the correct `RowKind`, but it does **not** guarantee
+a complete initial load of rows that already exist: any record that has aged out of the changelog
+through log retention or compaction will be missing. A full "snapshot + incremental" sync of a
+primary-key table is not supported yet. If you need the complete current state, prefer a **log
+(append-only) table**, which the log scanner always reads in full.
+
+:::
+
+Boundedness follows the job mode:
+
+- In `BATCH` mode the source is bounded: each bucket is read up to the latest log offset captured
+  when the job starts, then the split finishes.
+- In `STREAMING` mode the source is unbounded: it keeps reading new log records. The read position
+  of every bucket is stored in the checkpoint state, so the job resumes from where it stopped.
+
+Use `start_mode` to choose the offset each bucket starts from:
+
+- `earliest` (default): read the whole log from its earliest available offset.
+- `latest`: read only records appended after the job starts.
+
+`start_mode=latest` is only meaningful for a streaming job.
+
+The Fluss database and table must already exist before the job starts. The source does not create
+Fluss databases or tables. The table schema is read automatically from the Fluss cluster, so no
+`schema` option is required.
+
+## Limitations
+
+- **Single table only.** Each source reads exactly one table, configured with `database` + `table`.
+  Reading multiple tables in a single source is not supported.
+- **No arbitrary start offset.** The start position is chosen with `start_mode` (`earliest` or
+  `latest`) only. Starting from a specific log offset is not supported.
+- **Partitioned tables are not supported.** Pointing the source at a partitioned Fluss table fails
+  fast at job startup with an error. Use a non-partitioned table.
+- **Primary-key tables are read as changelog only.** The connector reads the table's changelog, not
+  a KV snapshot, so it does not perform a full initial load ("snapshot + incremental"). See the
+  primary-key caution above for details.
+
+## Dependency
+
+```xml
+<dependency>
+    <groupId>com.alibaba.fluss</groupId>
+    <artifactId>fluss-client</artifactId>
+    <version>0.7.0</version>
+</dependency>
+```
+
+## Source Options
+
+| Name | Type | Required | Default | Description |
+|---|---|---|---|---|
+| bootstrap.servers | string | yes | - | Fluss coordinator address, for example `fluss-coordinator:9123`. |
+| database | string | yes | - | The Fluss database to read from. |
+| table | string | yes | - | The Fluss table to read from. |
+| client.config | map | no | - | Extra Fluss client options passed to the Fluss connection. |
+| start_mode | string | no | earliest | The offset each bucket starts reading from: `earliest` (whole log) or `latest` (only records appended after the job starts). `latest` is rejected in `BATCH` mode. |
+| poll.timeout.ms | long | no | 10000 | The maximum time, in milliseconds, to block in a single Fluss log scanner poll. |
+| common-options | - | no | - | Source common options. See [Source Common Options](../common-options/source-common-options.md). |
+
+### client.config
+
+Use `client.config` to pass additional Fluss client settings.
+
+```hocon
+client.config = {
+  request.timeout = "30s"
+}
+```
+
+Refer to the Fluss client documentation for supported keys.
+
+## Data Type Mapping
+
+| Fluss Data Type | SeaTunnel Data Type |
+|---|---|
+| BOOLEAN | BOOLEAN |
+| TINYINT | TINYINT |
+| SMALLINT | SMALLINT |
+| INT | INT |
+| BIGINT | BIGINT |
+| FLOAT | FLOAT |
+| DOUBLE | DOUBLE |
+| DECIMAL | DECIMAL |
+| CHAR | STRING |
+| STRING | STRING |
+| BINARY | BYTES |
+| BYTES | BYTES |
+| DATE | DATE |
+| TIME | TIME |
+| TIMESTAMP | TIMESTAMP |
+| TIMESTAMP_LTZ | TIMESTAMP_TZ |
+
+## Task Examples
+
+### Batch read
+
+```hocon
+env {
+  parallelism = 1
+  job.mode = "BATCH"
+}
+
+source {
+  Fluss {
+    bootstrap.servers = "fluss-coordinator:9123"
+    database = "fluss_db"
+    table = "fluss_table"
+    plugin_output = "fluss_source"
+  }
+}
+
+sink {
+  Console {
+    plugin_input = "fluss_source"
+  }
+}
+```
+
+### Streaming read
+
+```hocon
+env {
+  parallelism = 1
+  job.mode = "STREAMING"
+}
+
+source {
+  Fluss {
+    bootstrap.servers = "fluss-coordinator:9123"
+    database = "fluss_db"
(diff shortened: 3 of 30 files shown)


DanielLeens (commented):
Outdated.

DanielLeens (commented):
Outdated.

ss666:
Thanks @DanielLeens for the recommendation.

I've added unit tests for exactly those paths in `FlussSourceSplitReaderTest`. They drive real `fetch()` through an injected mock`LogScanner`/admin. I deliberately kept this at the unit/mock level rather than a real retention
e2e: forcing an actual server-side truncation depends on Fluss retention timers, which makes such a test timing-dependent and flaky.

DanielLeens (commented):
Outdated.

davidzollo (approved):
+1 if CI passes
LGTM

SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `207feb6cd6b2` from scratch.

# What this PR fixes
- User pain: SeaTunnel users can only write to Fluss today — there is no way to read Fluss log/PK tables into a SeaTunnel pipeline, in batch or streaming.
- Fix approach: Adds a full source implementation to the existing `connector-fluss` module — `FlussSource`/`FlussSourceFactory`, a bucket-per-split `FlussSourceSplitEnumerator` with snapshot/restore, `FlussSourceReader`/`FlussSourceSplitReader` polling under the checkpoint lock, `FlussRecordEmitter` mapping Fluss change types to `RowKind`, `StartMode` (`earliest`/`latest`) + `poll.timeout.ms` options, `FlussTypeConverter`, EN/ZH docs, and plugin-mapping/labeler/plugin_config registration.
- One-line summary: The overall shape follows the standard source-connector layout and the docs are unusually thorough (PK-changelog and partitioned-table limitations are honestly stated), but I want the `SupportColumnProjection` declaration vs. "column projection not supported" docs contradiction resolved, a lower bound / lock-hold audit on the 10s `poll.timeout.ms` default, the `StartMode.toString()` round-trip fixed, and the apparent e2e `FlussSinkIT` API change explained before this merges.

# Runtime chain I rechecked
```text
FlussSourceFactory.createSource() -> new FlussSource(readonlyConfig)  FlussSourceFactory.java:~55
  FlussSource.createEnumerator() -> FlussSourceSplitEnumerator.open()  FlussSource.java:~90
    FlussSourceSplitEnumerator.run() -> FlussAdminClient.getTableBuckets()/getTableInfo()  FlussSourceSplitEnumerator.java:~70
      -> fail-fast on partitioned tables, one FlussSourceSplit per bucket, assignSplit() to readers  FlussSourceSplitEnumerator.java:~95
      -> snapshotState()/addSplitsBack() for stable allocation across restarts  FlussSourceSplitEnumerator.java:~140
  FlussSource.createReader() -> FlussSourceReader.pollNext()  FlussSource.java:~100
    FlussSourceReader.pollNext() -> FlussSourceSplitReader.poll(POLL_TIMEOUT_MS) under output.getCheckpointLock()  FlussSourceReader.java:~80
      -> LogScanner.poll() per subscribed bucket, honoring StartMode EARLIEST/LATEST offsets  FlussSourceSplitReader.java:~60
      -> FlussRecordEmitter.emit(record) -> ChangeType -> RowKind, FlussTypeConverter -> SeaTunnelRow field values -> Collector.collect()  FlussRecordEmitter.java:~50
      -> bounded (BATCH) end condition: bucket read up to latest offset captured at start -> signalNoMoreElement()  FlussSourceReader.java:~110
```

# Findings
**Issue 1: Possible public API change in seatunnel-e2e/seatunnel-connector-v2-e2e/connector-fluss-e2e/src/test/java/org/apache/seatunnel/e2e/connector/fluss/FlussSinkIT.java**
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-fluss-e2e/src/test/java/org/apache/seatunnel/e2e/connector/fluss/FlussSinkIT.java`
- Severity: High

**Issue 2: `poll.timeout.ms` has no lower-bound validation and its 10s default is dangerous if the poll happens under the checkpoint lock**
- Location: `seatunnel-connectors-v2/connector-fluss/src/main/java/org/apache/seatunnel/connectors/seatunnel/fluss/config/FlussSourceOptions.java:27`
- Why it matters: - Root cause: `POLL_TIMEOUT_MS` is declared as a plain `longType()` option with a 10-second default and no constraint, so users can configure `0` or a negative value, and nothing in the option definition (or any visible validation) rejects it.
- Impact: a non-positive timeout will either busy-spin the split reader (CPU burn, log flooding) or throw inside the Fluss `LogScanner.poll(Duration)` depending on the client's behavior. Worse, the PR description says the reader "honors the checkpoint lock" — if `poll()` is invoked while holding that lock, a 10s (or user-configured 60s+) blocking poll will stall barrier alignment and cause checkpoint timeouts on Zeta/Flink under low-traffic tables, which is a classic source-connector failure mode.
- Fix: add a positive-value check in `FlussSourceConfig` (fail fast with `IllegalArgumentException`/connector exception for `<= 0`), and make sure `FlussSourceSplitReader` calls `logScanner.poll(...)` OUTSIDE the checkpoint lock, only acquiring the lock to emit the already-fetched batch and advance offsets — the same pattern the Kafka source uses.
- Suggested fix: `poll.timeout.ms` accepts any long — please validate it is `> 0` in `FlussSourceConfig` and fail fast on invalid values. Also, since the default is 10 seconds, please double-check that `FlussSourceSplitReader` does NOT call `logScanner.poll(...)` while holding the checkpoint lock; a 10s blocking poll inside the lock will stall barrier alignment and cause checkpoint timeouts on idle tables. Poll outside the lock and only emit + advance offsets under it (Kafka source pattern).
- Severity: Medium

**Issue 3: Factory declares `SupportColumnProjection` but docs (and Limitations) state column projection is not supported — projection pushdown will corrupt row layout if the reader ignores read columns**
- Location: `docs/en/connectors/source/Fluss.md:18`
- Why it matters: - Root cause: the PR description says `FlussSourceFactory` implements `SupportColumnProjection`, yet the shipped documentation explicitly marks column projection as unsupported (`- [ ] [column projection]`) and the reader appears to emit full rows from the Fluss log scanner.
- Impact: `SupportColumnProjection` is a contract that tells the engine the source will honor a projected read schema. If the engine (Zeta/Flink/Spark) pushes down a column subset and the reader still emits full-width rows, downstream transforms/sinks will see a field-count/type mismatch producing `ClassCastException` or silently wrong column mapping — a hard-to-diagnose data-corruption bug, not just a doc nit.
- Fix: either
- (a) actually apply the projected columns in `FlussSourceSplitReader` (the Fluss log scanner supports projected field reads) and tick the docs checkbox, or
- (b) remove `SupportColumnProjection` from `FlussSourceFactory` so the engine never pushes projection down. Docs and factory capability flags must agree.
- Evidence: `docs/zh/connectors/source/Fluss.md:18  - [ ] [列投影](../../introduction/concepts/connector-v2-features.md)`
- Suggested fix: This checkbox says column projection is NOT supported, but the PR wires `SupportColumnProjection` into `FlussSourceFactory`. That interface tells the engine it may push a projected read schema down to the source — if the split reader then emits full-width rows, users get field-count/type mismatches at runtime. Please either implement projection in `FlussSourceSplitReader` (Fluss log scanner supports projected reads) and check this box, or drop `SupportColumnProjection` from the factory.
- Severity: Medium

**Issue 4: `StartMode.toString()` returns the lowercase alias, which breaks any `valueOf(x.toString())` round-trip (e.g., in state/config serialization or logging-based debugging)**
- Location: `seatunnel-connectors-v2/connector-fluss/src/main/java/org/apache/seatunnel/connectors/seatunnel/fluss/config/StartMode.java:36`
- Why it matters: - Root cause: `StartMode` overrides `toString()` to return the lowercase `mode` string (`"earliest"`/`"latest"`) while the enum constants are `EARLIEST`/`LATEST`. `StartMode.valueOf(StartMode.EARLIEST.toString())` throws `IllegalArgumentException`.
- Impact: any code path that serializes the option via `toString()` and restores via `valueOf` — a common pattern in config re-serialization, saved job configs, or enumerator/split state that stores the mode as a string — will fail at restore time with a confusing `No enum constant` error. SeaTunnel's `Options.enumType` parsing is case-insensitive on input, so the custom `toString` buys nothing there.
- Fix: either drop the `toString()` override (keep `getMode()` for display), or add a static `fromString(String)` that accepts both forms and use it everywhere the value is rehydrated. Verify nothing in `FlussSourceConfig`/state serialization relies on `valueOf(toString())`.
- Suggested fix: Overriding `toString()` to return the lowercase alias means `StartMode.valueOf(x.toString())` throws `IllegalArgumentException`. If any config/state path ever round-trips this enum through its string form, restore will fail with a cryptic `No enum constant` error. I'd drop the override (or add a lenient `fromString` and use it consistently) — `Options.enumType` already parses user input case-insensitively.
- Severity: Low

**Issue 5: Inconsistent option-key naming within the same source: dotted `poll.timeout.ms` vs snake_case `start_mode`**
- Location: `seatunnel-connectors-v2/connector-fluss/src/main/java/org/apache/seatunnel/connectors/seatunnel/fluss/config/FlussSourceOptions.java:35`
- Why it matters: - Root cause: the two new source options in `FlussSourceOptions` use different naming conventions — `poll.timeout.ms` (Kafka-style dotted) and `start_mode` (snake_case).
- Impact: this is user-facing config surface; once released it cannot be renamed without an incompatible change (and an `incompatible-changes.md` entry). Mixed conventions in a single connector confuse users and make docs harder to grep. Most recent SeaTunnel source connectors use snake_case for connector-specific options (e.g., `start_mode`, Kafka's `poll.timeout` being a legacy exception).
- Fix: rename to `poll_timeout_ms` (or keep dotted style for both) before this first release, so the naming is consistent from day one; update both EN/ZH docs and the e2e configs accordingly.
- Evidence: `seatunnel-connectors-v2/connector-fluss/src/main/java/org/apache/seatunnel/connectors/seatunnel/fluss/config/FlussSourceOptions.java:35  Options.key("start_mode")`
- Suggested fix: Minor: `start_mode` is snake_case while `poll.timeout.ms` just above uses dotted style. Since this is the connector's first release, let's pick one convention now (I'd suggest snake_case: `poll_timeout_ms`) — renaming later would be a user-facing breaking change requiring an `incompatible-changes.md` entry.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Possible public API change in seatunnel-e2e/seatunnel-connector-v2-e2e/connector-fluss-e2e/src/test/java/org/apache/seatunnel/e2e/connector/fluss/FlussSinkIT.java

- Issue 2: `poll.timeout.ms` has no lower-bound validation and its 10s default is dangerous if the poll happens under the checkpoint lock
- Issue 3: Factory declares `SupportColumnProjection` but docs (and Limitations) state column projection is not supported — projection pushdown will corrupt row layout if the reader ignores read columns
- Issue 4: `StartMode.toString()` returns the lowercase alias, which breaks any `valueOf(x.toString())` round-trip (e.g., in state/config serialization or logging-based debugging)
- Issue 5: Inconsistent option-key naming within the same source: dotted `poll.timeout.ms` vs snake_case `start_mode`

ss666:
@SEZ9 Thanks for the detailed re-review. Replies inline:

**Issue 1 — not an API change.**
`FlussSinkIT` now extends `FlussTestBase` instead of duplicating container setup. The public/protected members the finding refers to were moved into the new shared base. `FlussSinkIT` still inherits all of them; the helpers' visibility was narrowed from `public` to `protected` within the test hierarchy (test-only methods on an IT class). There's no downstream compatibility surface.

**Issue 2 — both already handled.**
- **Lower bound:** `FlussSourceFactory.optionRule()` rejects `<= 0` at submission time via `greaterThan(POLL_TIMEOUT_MS, 0L)` (`FlussSourceFactory.java:53-55`).
- **Lock:** I've reworded the "honor the checkpoint lock" line in the PR description that likely prompted this — the reader doesn't take the lock itself; the base reader takes it only around record emission, and `poll()` happens off the lock on the fetcher thread.

**Issue 3 — no contradiction; my PR description was wrong.**
The code does not implement `SupportColumnProjection` anywhere, and the docs correctly mark column projection as unsupported. The `SupportColumnProjection` mention came from a stale line in my PR description, which I've fixed — the code and docs already agree.

**Issue 4 — no live path; keeping for consistency.**

**Issue 5 — a style choice with no documented rule; keeping current keys.**
As far as I could find, SeaTunnel doesn't document a naming-separator convention. `poll.timeout.ms` is actually consistent with the connector's existing dotted keys, `bootstrap.servers` and `client.config`, which come from `FlussBaseOptions`. So renaming `poll.timeout.ms` to snake_case for consistency would mean renaming those too — and they're shared with the Fluss sink already merged in #10121.

DanielLeens:
Thanks for the detailed follow-up on the later review comments.

Since those points are substantive source-review items and there is no newer commit after the reply yet, I will not start a fresh full re-review from Daniel's side in this round. The best next step is to resolve the open review points in the discussion, and if any code/doc changes are needed, push an updated head. I am happy to re-review the complete current diff once there is a new commit.

The current `Build` check is green, so this is about resolving the review discussion rather than chasing CI.

davidzollo (approved):
+1
LGTM

davidzollo (approved):
+1
LGTM

PR #11562, opened 2026-07-26
[Fix][Connector-V2] Wait for Doris data visibility

### LINKED ISSUE #11433: [Bug] [Doris Sink] Discrepancy between the transaction status returned by the Doris sink connector and the actual visible Doris data status.

zhangshenghang:
## What does this PR do?

Fixes #11433.

The Doris 2PC committer now carries the stream-load label from the writer and polls the Doris FE load-state endpoint after the transaction commit succeeds. SeaTunnel reports the checkpoint as committed only after the label reaches `VISIBLE`; `ABORTED` and `CANCELLED` remain terminal failures.

Existing serialized commit information remains readable. When restoring an older checkpoint that does not contain a label, the existing commit behavior is retained.

## Tests

- Added a unit test for `COMMITTED` to `VISIBLE` polling.
- Added serializer round-trip and legacy-data compatibility tests.
- `./mvnw -o -f seatunnel-connectors-v2/connector-doris/pom.xml -Dtest=DorisCommitterTest,DorisCommitInfoSerializerTest test`

### DIFF
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/config/DorisSinkConfig.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/config/DorisSinkConfig.java
@@ -53,6 +53,7 @@ import static org.apache.seatunnel.connectors.doris.config.DorisSinkOptions.SINK
 import static org.apache.seatunnel.connectors.doris.config.DorisSinkOptions.SINK_ENABLE_DELETE;
 import static org.apache.seatunnel.connectors.doris.config.DorisSinkOptions.SINK_LABEL_PREFIX;
 import static org.apache.seatunnel.connectors.doris.config.DorisSinkOptions.SINK_MAX_RETRIES;
+import static org.apache.seatunnel.connectors.doris.config.DorisSinkOptions.SINK_VISIBILITY_TIMEOUT_MS;
 
 @Slf4j
 @Setter
@@ -76,6 +77,7 @@ public class DorisSinkConfig implements Serializable {
     private String labelPrefix;
     private Integer checkInterval;
     private Integer maxRetries;
+    private Long visibilityTimeoutMs;
     private Integer bufferSize;
     private Integer bufferCount;
     private Properties streamLoadProps;
@@ -115,6 +117,7 @@ public class DorisSinkConfig implements Serializable {
         dorisSinkConfig.setLabelPrefix(config.get(SINK_LABEL_PREFIX));
         dorisSinkConfig.setCheckInterval(config.get(SINK_CHECK_INTERVAL));
         dorisSinkConfig.setMaxRetries(config.get(SINK_MAX_RETRIES));
+        dorisSinkConfig.setVisibilityTimeoutMs(config.get(SINK_VISIBILITY_TIMEOUT_MS));
         dorisSinkConfig.setBufferSize(config.get(SINK_BUFFER_SIZE));
         dorisSinkConfig.setBufferCount(config.get(SINK_BUFFER_COUNT));
         dorisSinkConfig.setEnableDelete(config.get(SINK_ENABLE_DELETE));
@@ -126,7 +129,17 @@ public class DorisSinkConfig implements Serializable {
         validateDirectWriteOptions(
                 dorisSinkConfig.isDirectToBe(), dorisSinkConfig.getBackends(), true);
 
+        validateVisibilityTimeout(dorisSinkConfig.getVisibilityTimeoutMs());
+
         return dorisSinkConfig;
+    }
+
+    private static void validateVisibilityTimeout(Long visibilityTimeoutMs) {
+        if (visibilityTimeoutMs == null || visibilityTimeoutMs <= 0) {
+            throw new DorisConnectorException(
+                    SeaTunnelAPIErrorCode.CONFIG_VALIDATION_FAILED,
+                    "PluginName: Doris, Message: Option 'sink.visibility-timeout-ms' must be greater than 0.");
+        }
     }
 
     private static Properties parseStreamLoadProperties(ReadonlyConfig config) {
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/config/DorisSinkOptions.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/config/DorisSinkOptions.java
@@ -50,6 +50,13 @@ public class DorisSinkOptions extends DorisBaseOptions {
                     .intType()
                     .defaultValue(3)
                     .withDescription("the max retry times if writing records to database failed.");
+
+    public static final Option<Long> SINK_VISIBILITY_TIMEOUT_MS =
+            Options.key("sink.visibility-timeout-ms")
+                    .longType()
+                    .defaultValue(300000L)
+                    .withDescription(
+                            "Maximum time in milliseconds to wait for a 2PC load to reach the VISIBLE state on Doris FE after a successful stream-load commit. ABORTED and CANCELLED remain terminal failures. Must be greater than 0.");
     public static final Option<Integer> SINK_BUFFER_SIZE =
             Options.key("sink.buffer-size")
                     .intType()
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitInfo.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitInfo.java
@@ -33,10 +33,16 @@ public class DorisCommitInfo implements Serializable {
     private final String hostPort;
     private final String db;
     private final long txbID;
+    private final String label;
 
     public DorisCommitInfo(String hostPort, String db, long txbID) {
+        this(hostPort, db, txbID, null);
+    }
+
+    public DorisCommitInfo(String hostPort, String db, long txbID, String label) {
         this.hostPort = hostPort;
         this.db = db;
         this.txbID = txbID;
+        this.label = label;
     }
 }
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitInfoSerializer.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitInfoSerializer.java
@@ -35,6 +35,10 @@ public class DorisCommitInfoSerializer implements Serializer<DorisCommitInfo> {
             out.writeUTF(dorisCommittable.getHostPort());
             out.writeUTF(dorisCommittable.getDb());
             out.writeLong(dorisCommittable.getTxbID());
+            out.writeBoolean(dorisCommittable.getLabel() != null);
+            if (dorisCommittable.getLabel() != null) {
+                out.writeUTF(dorisCommittable.getLabel());
+            }
 
             out.flush();
             return baos.toByteArray();
@@ -48,7 +52,10 @@ public class DorisCommitInfoSerializer implements Serializer<DorisCommitInfo> {
             final String hostPort = in.readUTF();
             final String db = in.readUTF();
             final long txnId = in.readLong();
-            return new DorisCommitInfo(hostPort, db, txnId);
+            if (in.available() == 0) {
+                return new DorisCommitInfo(hostPort, db, txnId);
+            }
+            return new DorisCommitInfo(hostPort, db, txnId, in.readBoolean() ? in.readUTF() : null);
         }
     }
 }
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java
@@ -30,14 +30,19 @@ import org.apache.seatunnel.connectors.doris.util.DorisRedirectExceptionBuilder;
 import org.apache.seatunnel.connectors.doris.util.HttpUtil;
 import org.apache.seatunnel.connectors.doris.util.ResponseUtil;
 
+import org.apache.commons.codec.binary.Base64;
 import org.apache.http.Header;
+import org.apache.http.HttpHeaders;
 import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.util.EntityUtils;
 
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -49,8 +54,12 @@ import java.util.Map;
 @Slf4j
 public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
     private static final String COMMIT_PATTERN = "http://%s/api/%s/_stream_load_2pc";
+    private static final String LOAD_STATE_PATTERN = "http://%s/api/%s/get_load_state?label=%s";
     private static final int HTTP_OK = 200;
     private static final int HTTP_TEMPORARY_REDIRECT = 307;
+    private static final String LOAD_STATE_VISIBLE = "VISIBLE";
+    private static final String LOAD_STATE_ABORTED = "ABORTED";
+    private static final String LOAD_STATE_CANCELLED = "CANCELLED";
     private final CloseableHttpClient httpClient;
     private final DorisSinkConfig dorisSinkConfig;
     private final int maxRetry;
@@ -147,6 +156,7 @@ public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
                     continue;
                 }
                 handleCommitSuccess(committable, hostPort, closeableResponse);
+                waitUntilLoadVisible(committable, retryHosts);
                 return;
             }
         }
@@ -258,6 +268,86 @@ public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
             } else {
                 log.info("load result {}", loadResult);
             }
+        }
+    }
+
+    /** Waits until the 2PC load is visible before reporting the checkpoint as committed. */
+    private void waitUntilLoadVisible(DorisCommitInfo committable, List<String> retryHosts)
+            throws IOException {
+        if (committable.getLabel() == null) {
+            log.warn(
+                    "Skip waiting for Doris load visibility because the restored commit info has no label. txnId: {}",
+                    committable.getTxbID());
+            return;
+        }
+
+        int attempt = 0;
+        while (true) {
+            String hostPort = retryHosts.get(attempt % retryHosts.size());
+            try {
+                if (isLoadVisible(committable, hostPort)) {
+                    return;
+                }
+            } catch (DorisConnectorException e) {
+                throw e;
+            } catch (IOException e) {
+                log.warn(
+                        "Failed to get Doris load state for label {} from {}. Retrying.",
+                        committable.getLabel(),
+                        hostPort,
+                        e);
+            }
+            retrySleeper.sleep(Math.min(++attempt, 5));
+        }
+    }
+
+    private boolean isLoadVisible(DorisCommitInfo committable, String hostPort) throws IOException {
+        String encodedLabel =
+                URLEncoder.encode(committable.getLabel(), StandardCharsets.UTF_8.name());
+        String requestUrl =
+                String.format(LOAD_STATE_PATTERN, hostPort, committable.getDb(), encodedLabel);
+        HttpGet get = new HttpGet(requestUrl);
+        String authInfo = dorisSinkConfig.getUsername() + ":" + dorisSinkConfig.getPassword();
+        get.setHeader(
+                HttpHeaders.AUTHORIZATION,
+                "Basic "
+                        + new String(
+                                Base64.encodeBase64(authInfo.getBytes(StandardCharsets.UTF_8)),
+                                StandardCharsets.UTF_8));
+        try (CloseableHttpResponse response =
+                HttpUtil.executeWithRedirectTracking(
+                        httpClient,
+                        get,
+                        requestUrl,
+                        dorisSinkConfig.isDirectToBe(),
+                        dorisSinkConfig.getEnable2PC(),
+                        "get-load-state")) {
+            if (response.getStatusLine().getStatusCode() != HTTP_OK
+                    || response.getEntity() == null) {
+                throw new IOException(
+                        "Failed to get Doris load state, response: " + response.getStatusLine());
+            }
+            String responseBody = EntityUtils.toString(response.getEntity());
+            Map<String, Object> result =
+                    new ObjectMapper()
+                            .readValue(
+                                    responseBody, new TypeReference<HashMap<String, Object>>() {});
+            String loadState = String.valueOf(result.get("data"));
+            if (LOAD_STATE_VISIBLE.equalsIgnoreCase(loadState)) {
+                log.info("Doris load label {} is visible", committable.getLabel());
+                return true;
+            }
+            if (LOAD_STATE_ABORTED.equalsIgnoreCase(loadState)
+                    || LOAD_STATE_CANCELLED.equalsIgnoreCase(loadState)) {
+                throw new DorisConnectorException(
+                        DorisConnectorErrorCode.COMMIT_FAILED,
+                        "Doris load label "
+                                + committable.getLabel()
+                                + " reached terminal state "
+                                + loadState);
+            }
+            log.info("Doris load label {} is in state {}", committable.getLabel(), loadState);
+            return false;
         }
     }
 
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java
@@ -63,6 +63,7 @@ public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
     private final CloseableHttpClient httpClient;
     private final DorisSinkConfig dorisSinkConfig;
     private final int maxRetry;
+    private final long visibilityTimeoutMs;
     private final RetrySleeper retrySleeper;
 
     public DorisCommitter(DorisSinkConfig dorisSinkConfig) {
@@ -80,6 +81,7 @@ public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
         this.dorisSinkConfig = dorisSinkConfig;
         this.httpClient = client;
         this.maxRetry = dorisSinkConfig.getMaxRetries();
+        this.visibilityTimeoutMs = dorisSinkConfig.getVisibilityTimeoutMs();
         this.retrySleeper = retrySleeper;
     }
 
@@ -271,7 +273,11 @@ public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
         }
     }
 
-    /** Waits until the 2PC load is visible before reporting the checkpoint as committed. */
+    /**
+     * Waits until the 2PC load is visible before reporting the checkpoint as committed. Bounded by
+     * {@link DorisSinkConfig#getVisibilityTimeoutMs()} so a stuck or unknown state cannot block the
+     * checkpoint forever; ABORTED / CANCELLED remain terminal failures.
+     */
     private void waitUntilLoadVisible(DorisCommitInfo committable, List<String> retryHosts)
             throws IOException {
         if (committable.getLabel() == null) {
@@ -281,6 +287,7 @@ public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
             return;
         }
 
+        long deadlineNanos = System.nanoTime() + visibilityTimeoutMs * 1_000_000L;
         int attempt = 0;
         while (true) {
             String hostPort = retryHosts.get(attempt % retryHosts.size());
@@ -296,6 +303,15 @@ public class DorisCommitter implements SinkCommitter<DorisCommitInfo> {
                         committable.getLabel(),
                         hostPort,
                         e);
+            }
+            if (System.nanoTime() >= deadlineNanos) {
+                throw new DorisConnectorException(
+                        DorisConnectorErrorCode.COMMIT_FAILED,
+                        "Timed out after "
+                                + visibilityTimeoutMs
+                                + "ms waiting for Doris load label "
+                                + committable.getLabel()
+                                + " to reach VISIBLE state.");
             }
             retrySleeper.sleep(Math.min(++attempt, 5));
         }
--- a/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisSinkWriter.java
@@ -262,7 +262,12 @@ public class DorisSinkWriter
         }
         long txnId = respContent.getTxnId();
 
-        return Optional.of(new DorisCommitInfo(controlHostPort, dorisStreamLoad.getDb(), txnId));
+        return Optional.of(
+                new DorisCommitInfo(
+                        controlHostPort,
+                        dorisStreamLoad.getDb(),
+                        txnId,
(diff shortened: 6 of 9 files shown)


DanielLeens (commented):
# Basic information

- PR title: [Fix][Connector-V2] Wait for Doris data visibility
- Author: zhangshenghang
- Link: https://github.com/apache/seatunnel/pull/11562
- Base branch: dev
- State: OPEN
- Change size: +212 / -2, 7 file(s)
- Human review ready: No
- Review time: 2026-07-26T16:00:00Z
- Local branch: seatunnel-review-11562
- Local worktree: shared bare PR 11562 review store

## Stable review keys

- pr_number: `11562`
- base_sha: `924b289e95d0bfe06286e8666bcbb7d986bc2023`
- reviewed_base_sha: `924b289e95d0bfe06286e8666bcbb7d986bc2023`
- merge_base_sha: `924b289e95d0bfe06286e8666bcbb7d986bc2023`
- head_sha: `dbf327bf09e3063402eca7ae24c937413c78c74c`
- reviewed_head_sha: `dbf327bf09e3063402eca7ae24c937413c78c74c`
- human_review_ready: `false`
- diff_integrity: `pass`
- deletion_total: `2`
- deletion_classified: `2`
- deletion_unclassified: `0`
- other_total: `0`
- other_confirmed: `0`
- other_unresolved: `0`
- reviewed_at: `2026-07-26T16:00:00Z`

- mergeStateStatus: `BLOCKED`
- mergeable_state: `blocked`
- mergeable: `True`
- rebaseable: `False`
- required_ci_gate: `failing`
- approval_gate: `pending`
- conflict_gate: `satisfied`
- merge_queue_gate: `pending`
- compare: status=`ahead`, ahead_by=`1`, behind_by=`0`, total_commits=`1`, url=https://github.com/apache/seatunnel/compare/dev...dbf327bf09e3063402eca7ae24c937413c78c74c

- Other gate `unrelated Windows CI failure`: `blocking` — Aggregate Build fails on an unrelated Edge Agent three-second timing test on Windows; Doris IT on Java 8 and 11 and updated shards passed, and a rerun can clear only the CI gate.

# What problem this PR solves

- User pain: Doris 2PC can report COMMITTED before data is VISIBLE, so SeaTunnel may complete the checkpoint while a dependent query still cannot see new rows.
- Fix approach: Carry the Stream Load label in a compatible DorisCommitInfo, poll load state after transaction commit until VISIBLE, and treat ABORTED or CANCELLED as terminal failure.
- One-line summary: Waiting for visibility is semantically correct, but the current poll has no finite failure boundary for persistent API, transport, or unknown-state conditions.
- Before: SeaTunnel completed immediately after COMMITTED and downstream SQL could miss rows.
- After: Completion is intended only after COMMITTED-to-VISIBLE, but current error or unknown-state paths can hang forever.

# 1. Code-change review

- Behavior before: SeaTunnel completed immediately after COMMITTED and downstream SQL could miss rows.
- Behavior after: Completion is intended only after COMMITTED-to-VISIBLE, but current error or unknown-state paths can hang forever.
- Normal-path hit: `conditional` — Only label-bearing 2PC committables from the new writer hit this path; legacy label=null skips visibility waiting.

- `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisSinkWriter.java:258-286` `prepareCommit label capture` — Obtains transaction ID and exact label from flush and stopLoad and builds commit info.
- `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitInfoSerializer.java:31-59` `compatible serializer` — Appends hasLabel and label after the old host, database, and transaction prefix, with EOF restoring legacy label=null.
- `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java:101-301` `commitTransaction / waitUntilLoadVisible` — After 2PC commit, polls FE load state; VISIBLE returns and other paths follow the current unbounded policy.

### Effective prerequisites

- Doris 2PC commit succeeds and commit info contains the Stream Load label.

- Historical three-field bytes cannot reconstruct the label and preserve old return behavior with a warning after compatible restore.

### Complete execution chain

```text
1. seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisSinkWriter.java:258-286 -> prepareCommit label capture [serialization]
   Obtains transaction ID and exact label from flush and stopLoad and builds commit info.
   Side effect: Obtains transaction ID and exact label from flush and stopLoad and builds commit info.
2. seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitInfoSerializer.java:31-59 -> compatible serializer [serialization]
   Appends hasLabel and label after the old host, database, and transaction prefix, with EOF restoring legacy label=null.
   Side effect: Appends hasLabel and label after the old host, database, and transaction prefix, with EOF restoring legacy label=null.
3. seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java:101-301 -> commitTransaction / waitUntilLoadVisible [rpc]
   After 2PC commit, polls FE load state; VISIBLE returns and other paths follow the current unbounded policy.
   Side effect: After 2PC commit, polls FE load state; VISIBLE returns and other paths follow the current unbounded policy.
```

- Recovery, exception, async, serialization, and concurrency paths: The commit list is sequential, so one stuck label blocks later entries; HTTP responses close, the client is reused, and sleeper interruption is restored, but no deadline exists. Serializer serialVersionUID and state prefix remain unchanged.

- Applicable: `false`

## 1.2 Compatibility impact

- API: The serializer appends an optional label and detects EOF after the legacy prefix; old constructor, SPI, transaction, options, defaults, and state remain, while legacy checkpoints simply cannot gain the new visibility guarantee.
- SPI: The serializer appends an optional label and detects EOF after the legacy prefix; old constructor, SPI, transaction, options, defaults, and state remain, while legacy checkpoints simply cannot gain the new visibility guarantee.
- Configuration: The serializer appends an optional label and detects EOF after the legacy prefix; old constructor, SPI, transaction, options, defaults, and state remain, while legacy checkpoints simply cannot gain the new visibility guarantee.
- Defaults: The serializer appends an optional label and detects EOF after the legacy prefix; old constructor, SPI, transaction, options, defaults, and state remain, while legacy checkpoints simply cannot gain the new visibility guarantee.
- Protocol and serialization: The serializer appends an optional label and detects EOF after the legacy prefix; old constructor, SPI, transaction, options, defaults, and state remain, while legacy checkpoints simply cannot gain the new visibility guarantee.
- Checkpoint/restore: The serializer appends an optional label and detects EOF after the legacy prefix; old constructor, SPI, transaction, options, defaults, and state remain, while legacy checkpoints simply cannot gain the new visibility guarantee.
- Historical behavior: The serializer appends an optional label and detects EOF after the legacy prefix; old constructor, SPI, transaction, options, defaults, and state remain, while legacy checkpoints simply cannot gain the new visibility guarantee.

- CPU: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Memory: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- GC: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Network: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Lock contention: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Concurrency safety: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Retry: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Idempotency: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Resource release: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.
- Overall: Each checkpoint adds FE GETs with backoff 1, 2, 4, 8, then 16 seconds; heap does not grow, but time and network calls are unbounded and sequential blocking results.

- Exception propagation: Known ABORTED and CANCELLED states have contextual errors, responses close, and auth is not logged; non-200, I/O, API-envelope, and unknown states have no finite exhaustion and are blockers.
- Context: Known ABORTED and CANCELLED states have contextual errors, responses close, and auth is not logged; non-200, I/O, API-envelope, and unknown states have no finite exhaustion and are blockers.
- Log levels: Known ABORTED and CANCELLED states have contextual errors, responses close, and auth is not logged; non-200, I/O, API-envelope, and unknown states have no finite exhaustion and are blockers.
- Sensitive information: Known ABORTED and CANCELLED states have contextual errors, responses close, and auth is not logged; non-200, I/O, API-envelope, and unknown states have no finite exhaustion and are blockers.
- Overall: Known ABORTED and CANCELLED states have contextual errors, responses close, and auth is not logged; non-200, I/O, API-envelope, and unknown states have no finite exhaustion and are blockers.

- diff_integrity: `pass`
- deletion_ledger: total=`2`, classified=`2`, unclassified=`0`
- Reconciliation: 2 = 2 + 0

- `deletion-001` | count=`1` | `deletion` | `intended_change` | `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitInfoSerializer.java` (old line 51) — The two deleted lines replace the legacy unconditional return with EOF and optional-label decoding and replace three-argument commit info with a four-argument active-label construction; both are paired changes.
- `deletion-002` | count=`1` | `deletion` | `intended_change` | `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/writer/DorisSinkWriter.java` (old line 265) — The two deleted lines replace the legacy unconditional return with EOF and optional-label decoding and replace three-argument commit info with a four-argument active-label construction; both are paired changes.

- Out-of-scope hunk conclusion: All seven files belong to Doris 2PC visibility, serializer, and tests, with no unrelated hunk.
- Stale-base conclusion: The head is one commit ahead and zero behind reviewed dev, with no stale rollback.
- Merge-artifact conclusion: There is no deleted or renamed file, conflict marker, or merge artifact; mergeable=true.

# 2. Code-quality assessment

- Title format: The [Fix][Connector-V2] title matches; new-test ASF header, imports, error context, and diff check follow conventions.
- Code conventions: The [Fix][Connector-V2] title matches; new-test ASF header, imports, error context, and diff check follow conventions.
- Comment completeness: The [Fix][Connector-V2] title matches; new-test ASF header, imports, error context, and diff check follow conventions.

- Unit tests: `covered` — Current deterministic tests cover current and legacy serializer bytes plus the COMMITTED-to-VISIBLE happy path; critical terminal, error, exhaustion, and interruption paths are missing.
- Component tests: `missing` — Current deterministic tests cover current and legacy serializer bytes plus the COMMITTED-to-VISIBLE happy path; critical terminal, error, exhaustion, and interruption paths are missing.
- Engine integration tests: `not_applicable` — Current deterministic tests cover current and legacy serializer bytes plus the COMMITTED-to-VISIBLE happy path; critical terminal, error, exhaustion, and interruption paths are missing.
- E2E: `missing` — Current deterministic tests cover current and legacy serializer bytes plus the COMMITTED-to-VISIBLE happy path; critical terminal, error, exhaustion, and interruption paths are missing.
- Regression tests: `missing` — Current deterministic tests cover current and legacy serializer bytes plus the COMMITTED-to-VISIBLE happy path; critical terminal, error, exhaustion, and interruption paths are missing.
- Stability classification: `risk`

### Test-stability evidence

- `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java:101-301`
  - Trigger: New mock tests have no real sleep or network and are stable; risk comes from the central poll lacking a finite failure or deadline contract, a deterministic liveness gap.
  - Probability basis: New mock tests have no real sleep or network and are stable; risk comes from the central poll lacking a finite failure or deadline contract, a deterministic liveness gap.
  - Analysis: New mock tests have no real sleep or network and are stable; risk comes from the central poll lacking a finite failure or deadline contract, a deterministic liveness gap.

### Missing scenarios

- Add persistent I/O, non-zero envelopes, unknown and terminal states, host rotation, deadline exhaustion, interruption, and a live downstream-visibility E2E.

## 2.3 Documentation

- User-visible change: `true`
- docs/en: There is currently no configuration; once a finite timeout or retry policy exists, EN/ZH operator-visible completion semantics should be documented, while the intended fix needs no incompatible entry.
- docs/zh: There is currently no configuration; once a finite timeout or retry policy exists, EN/ZH operator-visible completion semantics should be documented, while the intended fix needs no incompatible entry.
- Examples: There is currently no configuration; once a finite timeout or retry policy exists, EN/ZH operator-visible completion semantics should be documented, while the intended fix needs no incompatible entry.
- Configuration truth: There is currently no configuration; once a finite timeout or retry policy exists, EN/ZH operator-visible completion semantics should be documented, while the intended fix needs no incompatible entry.
- URL truth: There is currently no configuration; once a finite timeout or retry policy exists, EN/ZH operator-visible completion semantics should be documented, while the intended fix needs no incompatible entry.

### Local execution boundary

- `git diff --stat 924b289e95d0bfe06286e8666bcbb7d986bc2023...dbf327bf09e3063402eca7ae24c937413c78c74c`
  - Result: `passed`
  - Explanation: Confirmed 7 files and +212/-2.
- `git diff --name-status 924b289e95d0bfe06286e8666bcbb7d986bc2023...dbf327bf09e3063402eca7ae24c937413c78c74c`
  - Result: `passed`
- `git diff --unified=0 924b289e95d0bfe06286e8666bcbb7d986bc2023...dbf327bf09e3063402eca7ae24c937413c78c74c`
  - Result: `passed`
- `git diff --check`
  - Result: `passed`

### Required checks

- `Build` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`FAILURE` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/89802698655

- `Notify test workflow` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/30205484832/job/89802687858
- `labeler` | workflow=`Pull Request Labeler` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`pass` | conclusion=`SUCCESS` | https://github.com/apache/seatunnel/actions/runs/30205484838/job/89802687889
- `Build` | workflow=`On pull request update` | job=`Not reported by GitHub` | state=`COMPLETED` | bucket=`fail` | conclusion=`FAILURE` | https://github.com/apache/seatunnel/runs/89802698655

# 3. Architecture assessment

- Type: `precise_fix`
- Assessment: Polling through the committer-owned client and FE failover boundary is the right direction, but a finite LoadVisibilityWaiter should make retry, cancellation, and testing contracts explicit.

## 3.2 Maintainability

Polling through the committer-owned client and FE failover boundary is the right direction, but a finite LoadVisibilityWaiter should make retry, cancellation, and testing contracts explicit.

## 3.3 Extensibility

Polling through the committer-owned client and FE failover boundary is the right direction, but a finite LoadVisibilityWaiter should make retry, cancellation, and testing contracts explicit.

Polling through the committer-owned client and FE failover boundary is the right direction, but a finite LoadVisibilityWaiter should make retry, cancellation, and testing contracts explicit.

# 5. Formal findings

## `问题 1` — Persistent FE/API errors or unknown states make the committer loop forever

- Location: `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java:274-351`
- Problem and real execution path: After 2PC commit succeeds, waitUntilLoadVisible uses while(true) and only returns on VISIBLE or throws on ABORTED/CANCELLED; IOException, non-200, malformed or error envelopes, missing data, and unknown states retry forever, and the class maxRetry is not used for this phase. commitTransaction -> successful 2PC -> waitUntilLoadVisible -> GET load state -> every unrecognized or error response -> capped sleep -> the same loop with no attempt or deadline.
- Current-head evidence: Line 285 is while(true) and line 300 retries unconditionally; HTTP-200 JSON code and msg are not validated, absent data becomes literal null, and the only happy test is COMMITTED-to-VISIBLE.
- Trigger condition: The endpoint remains unreachable, returns non-200, malformed, or error envelopes, cannot find the label, returns an unknown terminal state, or never becomes visible.
- Potential impact: Committer, checkpoint, and job completion never return; later committables in the same list are blocked and a live but erroneous endpoint has no autonomous failure boundary.
- Best actionable fix: Define a finite elapsed deadline or retry budget, validate code, msg, and data, classify transient and terminal states, throw a contextual connector exception with label, transaction, last state, and frontends on exhaustion, and cover persistent I/O, error envelopes, unknown states, timeout, and interruption.
- Severity: `high`
- Merge gate: `blocking`

## `问题 2` — The regression suite proves only the immediate happy transition

- Location: `seatunnel-connectors-v2/connector-doris/src/test/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitterTest.java:158-186`
- Problem and real execution path: Serializer round trip and legacy compatibility plus COMMITTED-to-VISIBLE are covered, but ABORTED, CANCELLED, non-200, empty entity, malformed JSON, non-zero API code, host failover, persistent exhaustion, and interruption are untested. New checkpoint visibility method -> mock scripts commit success, COMMITTED, one sleep, and VISIBLE; no error path has an executable assertion.
- Current-head evidence: The test scripts exactly three calls and one sleep; the serializer test only covers current and legacy bytes.
- Trigger condition: Any non-happy response or retry-policy change.
- Potential impact: The checkpoint-liveness boundary can break while the suite passes, leaving the failure contract from finding 1 unprotected.
- Best actionable fix: After implementing the finite policy, add table-driven response classes, host rotation, exhaustion context, and restored-interrupt tests.
- Severity: `medium`
- Merge gate: `blocking`

# 6. Findings summary

|---|---|---|---|---|
| 问题 1 | Persistent FE/API errors or unknown states make the committer loop forever | seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java:274-351 | high | blocking |
| 问题 2 | The regression suite proves only the immediate happy transition | seatunnel-connectors-v2/connector-doris/src/test/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitterTest.java:158-186 | medium | blocking |

# 7. Merge conclusion

### Blocking items

- `doris-visibility-infinite-loop` [source] (问题 1): The visibility poll has no finite deadline for persistent errors or unknown states and can block checkpoints forever.
- `doris-visibility-missing-failure-tests` [tests] (问题 2): The new checkpoint-liveness boundary lacks terminal, error, exhaustion, and interruption regression coverage.

### Overall assessment

Polling through the committer-owned client and FE failover boundary is the right direction, but a finite LoadVisibilityWaiter should make retry, cancellation, and testing contracts explicit.

- Alternative: Use a dedicated waiter where only VISIBLE succeeds, explicit terminal and API errors fail fast, transport errors rotate, an elapsed deadline is configurable, and clock and sleeper are injectable; a small abstraction yields provable liveness.
- Trade-offs: Use a dedicated waiter where only VISIBLE succeeds, explicit terminal and API errors fail fast, transport errors rotate, an elapsed deadline is configurable, and clock and sleeper are injectable; a small abstraction yields provable liveness.

- Reviewed commit: `dbf327bf09e3063402eca7ae24c937413c78c74c`

DanielLeens (changes requested):

Doris 2PC commit can report the transaction as `COMMITTED` before the loaded data is actually `VISIBLE` for queries. Before this PR, SeaTunnel completed the checkpoint as soon as the commit call succeeded, so a downstream query could run against Doris before the new rows were visible. The first commit on this branch (`Wait for Doris data visibility`) added a poll loop that carries the Stream Load label through a compatible `DorisCommitInfo`/serializer and waits for the FE `get_load_state` endpoint to report `VISIBLE` before returning from `commit()`, treating `ABORTED`/`CANCELLED` as terminal failures.

Hi @zhangshenghang, thanks for following up so quickly with the second commit. This is a fresh full re-review of the current head (`b2c117dbef`) against my prior `COMMENTED` round on `dbf327bf09`. The new commit, `Bound Doris 2PC visibility wait by configurable timeout`, is aimed squarely at my round-1 blocker: it adds a `sink.visibility-timeout-ms` option (default 5 minutes) and a deadline check inside `waitUntilLoadVisible`, plus four new regression tests for `ABORTED`, `CANCELLED`, timeout, and transient-`IOException` recovery. That closes the literal `while(true)`-forever gap I flagged and substantially improves the test suite, so I want to say that clearly up front rather than burying it.

That said, tracing the new deadline logic against the underlying HTTP client turned up a real gap in how "bounded" the wait actually is (see Issue 1 below) — and in hindsight I should have been more precise in round 1 by calling out that a deadline check alone isn't sufficient without a timeout on the underlying HTTP request itself; I only asked for "a finite elapsed deadline or retry budget" without spelling out that requirement, so it's fair to share the blame for this one landing in round 2 instead of round 1. Apologies for not catching it the first time.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The new option is threaded correctly end to end: `DorisSinkOptions.SINK_VISIBILITY_TIMEOUT_MS` (long, default `300000`) is read in `DorisSinkConfig.of(...)`, validated by `validateVisibilityTimeout(...)` (must be non-null and `> 0`), stored on `DorisSinkConfig`, and passed into `DorisCommitter`'s constructor as `visibilityTimeoutMs`. Inside `waitUntilLoadVisible`, a `deadlineNanos` is computed once before the loop starts, and after each poll attempt (whether it returned `false`, i.e. still pending, or threw a transient `IOException`) the code checks `System.nanoTime() >= deadlineNanos` and throws a `DorisConnectorException` with a clear "Timed out after Xms..." message if the deadline has passed. `ABORTED`/`CANCELLED` are still fail-fast via the `catch (DorisConnectorException e) { throw e; }` branch, independent of the deadline — that's the right behavior, since a terminal failure shouldn't have to wait out the timeout.

The one place this falls short of "bounded" is the HTTP layer underneath it. `isLoadVisible(...)` calls `HttpUtil.executeWithRedirectTracking(httpClient, get, ...)`, and the `httpClient` instance (`HttpUtil.getHttpClient()`) is built with `HttpClients.custom()...build()` — no `RequestConfig` with a connect timeout, socket (read) timeout, or connection-request timeout is ever set. Apache HttpClient 4.x defaults these to "wait indefinitely" when unset. The new deadline check only runs *between* completed poll attempts; it can't interrupt a single `httpClient.execute()` call that is already in flight. So if the FE becomes unreachable in a way that doesn't produce a fast connection-refused/error response — a half-open TCP connection behind a NAT/load balancer, a GC-paused or deadlocked FE process that accepted the connection but never writes a response, a silently dropped packet with no RST — a single `get_load_state` call can block forever, and `visibilityTimeoutMs` never gets a chance to fire. This is exactly the "persistent FE/API error" scenario the commit message calls out as the motivation for the fix, so it's worth closing rather than leaving as a known gap.

## 1.2 Compatibility Impact

Fully compatible. `sink.visibility-timeout-ms` is a brand-new option with a safe, non-null default (`300000L` / 5 minutes), so existing jobs that don't set it get a bounded wait instead of the previously-unbounded one — that's a strict improvement, not a behavior regression, since the "wait for visibility" feature itself is net-new in this same unreleased PR (there is no released version where callers relied on the old unbounded-loop behavior). The serializer, `DorisCommitInfo`, `DorisSinkWriter`, and `DorisStreamLoad` changes are unchanged from round 1 and remain untouched by the new commit (verified via `git diff` between the two commits) — legacy three-field commit-info bytes still deserialize correctly via the `in.available() == 0` EOF check, restoring `label = null` and skipping the visibility wait with a warning log, exactly as reviewed previously.

Each checkpoint now polls `get_load_state` with a capped exponential backoff (1s, 2s, 4s, 8s, 16s, 16s, ...) until either `VISIBLE`/terminal state or the 5-minute default deadline, whichever comes first — a meaningful improvement over the previous unbounded version. The blocking is still synchronous within the commit path (one stuck label blocks later commit entries in the same list, since `commit()` iterates `commitInfos` sequentially), which is unchanged from round 1's finding and not something this fix set out to address — I won't re-raise it as a fresh blocker here since it's a pre-existing structural property, but it does mean a hung poll (see 1.1) has more amplified impact than a single committable's timeout.

Known terminal states (`ABORTED`, `CANCELLED`) throw a contextual `DorisConnectorException` carrying the label and state; the timeout path throws a similarly contextual exception with the label and configured timeout value. Auth headers are not logged. Below are the issues found this round.

### Issue 1

- **Location**: `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitter.java` (`isLoadVisible`, `waitUntilLoadVisible`) and `seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/util/HttpUtil.java` (`getHttpClient()`)
- **Problem description**: The `httpClient` used to poll `get_load_state` is built without any `RequestConfig` connect/socket/connection-request timeout. `visibilityTimeoutMs` is only checked *between* completed poll attempts, never enforced on the in-flight HTTP call itself.
- **Potential risk**: A hung or half-open connection to the FE (network partition, GC-paused/deadlocked FE that accepted the socket but never responds, silently dropped connection) causes a single `httpClient.execute()` call to block indefinitely, completely bypassing the new deadline and leaving the checkpoint stuck forever — the same class of failure this PR is meant to fix.
- **Best improvement**: Configure a `RequestConfig` with an explicit connect timeout and socket (read) timeout on the client used for `get_load_state` polling (and ideally on the commit/abort clients too), sized well below `visibilityTimeoutMs` so a single hung request can be retried and the outer deadline remains meaningful.
- **Severity**: High
- **Raised by another reviewer**: No (new to this round; see the note above on how this could have been called out more precisely in round 1)

### Issue 2

- **Location**: `docs/en/connectors/sink/Doris.md`, `docs/zh/connectors/sink/Doris.md`
- **Problem description**: `sink.visibility-timeout-ms` is not documented — the sink option table lists `sink.enable-2pc`, `sink.check-interval`, and `sink.max-retries` but has no row for the new option, and the whole "wait for visibility before returning from commit" behavior introduced by this PR is not mentioned anywhere in either doc.
- **Potential risk**: Operators upgrading won't know checkpoints can now take up to 5 minutes longer per commit under FE slowness, won't know the option exists to tune that, and won't understand the new completion semantics (`VISIBLE` vs `COMMITTED`) without reading the source.
- **Best improvement**: Add a row for `sink.visibility-timeout-ms` (type `Long`, default `300000`, description matching the `Option`) to both docs' sink option tables, and add a short paragraph explaining the visibility-wait behavior and its bound.
- **Severity**: Medium
- **Raised by another reviewer**: Yes (I flagged this conditionally in round 1 — "once a finite timeout or retry policy exists ... should be documented" — that condition is now met by this commit, so the gap is concretely due)

### Issue 3

- **Location**: `seatunnel-connectors-v2/connector-doris/src/test/java/org/apache/seatunnel/connectors/doris/sink/committer/DorisCommitterTest.java`
- **Problem description**: `testCommitRetriesLoadStatePollOnIOExceptionBeforeDeadline` only covers a single transient `IOException` followed by success; there is no test where `IOException`s repeat until the deadline is exhausted (as opposed to `testCommitTimesOutWhenLoadStateStaysUnresolved`, which exhausts the deadline via a repeating non-terminal *state*, not via repeated I/O failures).
- **Potential risk**: Low — both paths share the same deadline-check code, so this is a coverage gap rather than a known behavioral bug, but a future refactor that accidentally special-cased the I/O-exception branch would not be caught by the current suite.
- **Best improvement**: Add a table-driven or dedicated test that throws `IOException` on every poll attempt and asserts the resulting exception is the timeout message (not a raw `IOException` escaping), once Issue 1's socket-timeout fix is in so the test can run deterministically without relying on wall-clock waits.
- **Severity**: Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

Title format `[Fix][Connector-V2] Bound Doris 2PC visibility wait by configurable timeout` matches convention. The new `Option` follows the existing builder pattern (`.longType().defaultValue(...).withDescription(...)`), the Javadoc on `waitUntilLoadVisible` was updated to describe the new bound, and imports/ASF headers are consistent with the rest of the module. One purely cosmetic nit: `SINK_VISIBILITY_TIMEOUT_MS` in `DorisSinkOptions.java` isn't followed by a blank line before `SINK_BUFFER_SIZE`, but the surrounding file already mixes blank-line conventions between `Option` fields, so this isn't a new regression worth blocking on.

Test coverage improved substantially since round 1. New/updated tests in `DorisCommitterTest.java`: `testCommitFailsWhenLoadStateIsAborted`, `testCommitFailsWhenLoadStateIsCancelled`, `testCommitTimesOutWhenLoadStateStaysUnresolved` (with an injected `1L` ms timeout so the test doesn't rely on real wall-clock waits), and `testCommitRetriesLoadStatePollOnIOExceptionBeforeDeadline`. All use the mocked `httpClient` and injectable `RetrySleeper`, so they're deterministic and fast — no real network or `Thread.sleep` in the test path itself. `DorisCommitInfoSerializerTest.java` (added in round 1, unchanged here) covers current and legacy byte layouts. Remaining gap: see Issue 3 above (persistent-I/O-exception-until-deadline path untested). Stability classification: low risk — the new tests are self-contained and don't appear flaky-prone.

## 2.3 Documentation Updates

Not updated — see Issue 2. No entry is needed in `docs/en/introduction/concepts/incompatible-changes.md` since this change is additive and backward compatible, but the new sink option itself needs to be documented per the project's "config names, defaults, and examples MUST match the code" rule.

# 3. Architectural Soundness

A deadline-bounded poll with capped backoff is the right shape for this problem, and keeping `ABORTED`/`CANCELLED` as immediate fail-fast paths (not waiting out the timeout) is a good design choice. The remaining gap is one layer down (HTTP client timeouts), not in the polling logic itself.

## 3.2 Maintainability

The change is small, localized to `DorisCommitter`/`DorisSinkConfig`/`DorisSinkOptions`, and the new Javadoc on `waitUntilLoadVisible` explains the bound clearly for future readers. Once Issue 1 is addressed, this becomes a clean, self-contained fix.

## 3.3 Extensibility

The `RetrySleeper` functional interface and constructor-injectable `httpClient` already make this testable and would make adding a per-request timeout straightforward without further structural changes.

No impact on checkpoint/savepoint restore beyond what round 1 already established (legacy three-field serialized `DorisCommitInfo` still restores correctly with `label = null`). No SPI, protocol, or option-removal concerns.

# 4. Issue Summary

|---|---|---|---|
| 1 | HTTP client used for `get_load_state` polling has no connect/socket timeout, so a hung connection can still block the checkpoint indefinitely, bypassing the new deadline | `DorisCommitter.java` (`isLoadVisible`), `HttpUtil.java` (`getHttpClient`) | High |
| 2 | New `sink.visibility-timeout-ms` option and the visibility-wait behavior are undocumented in docs/en and docs/zh | `docs/en/connectors/sink/Doris.md`, `docs/zh/connectors/sink/Doris.md` | Medium |
| 3 | No test for repeated `IOException` until deadline exhaustion (distinct from single-retry-then-success and non-terminal-state-until-timeout) | `DorisCommitterTest.java` | Low |

# 5. Merge Recommendation

### Conclusion: Request changes — close, but the HTTP client timeout gap needs to be fixed before this can be called a bounded wait, and the new option needs docs.

1. Blockers — must be fixed:
   - Issue 1: Add an explicit connect/socket timeout to the HTTP client used for `get_load_state` polling so a hung connection can't bypass `visibilityTimeoutMs`.
   - Issue 2: Document `sink.visibility-timeout-ms` (and the visibility-wait behavior) in `docs/en/connectors/sink/Doris.md` and `docs/zh/connectors/sink/Doris.md`.
   - Issue 3: Add a regression test for repeated `IOException` until deadline exhaustion.

**Overall assessment**: This is a solid, well-targeted follow-up that closes the literal "loops forever" bug and the missing terminal-state/timeout test coverage I flagged in round 1 — nice work turning that around quickly. The one thing standing between this and an approve is that the bound is only enforced at the polling-loop level, not at the HTTP-request level, so the specific failure mode the commit message describes (a genuinely stuck/unresponsive FE) can still defeat it. Once the HTTP client has its own timeout and the new option is documented, I'm happy to re-review and expect this to be in good shape to merge.

PR #11564, opened 2026-07-27
[Fix][Zeta] Offload report metrics operations

### LINKED ISSUE #10997: [Bug] [Zeta] Batch job occasionally hangs in RUNNING forever under high-frequency job submission
### Search before asking

### What happened

In production we hit an intermittent problem: a **BATCH** job (JDBC MySQL source → JDBC
`MultiTableSink`) finished reading all source data and the bounded reader closed normally, but the
whole job then **stays in `RUNNING` forever and never turns `FINISHED`**. It cannot finish on its
own — the only way to clear it is a manual cancel or a master restart.

This does not happen every time; it shows up intermittently under a high job-submission load.
Below are the clues we collected. We'd like the community's help to analyze the possible causes and
directions for improvement.

Environment & workload characteristics

- SeaTunnel version: 2.3.13, Zeta engine.
- Deployment: separated master/worker, 2 masters (HA 32C+48G heap) + 8 workers(3 32C+48G heap + 5 16C+20G heap).
- **Workload**: the platform has **1000+ data-collection jobs**, submitted periodically by a
  scheduler, with **on average several dozen job submissions per minute** (higher at peak). So the
  cluster runs continuously under "sustained high-frequency submission + many small concurrent jobs".
- The job that hung is itself trivial: a single bounded table read, BATCH mode, and **no
  `checkpoint.interval` configured (i.e. checkpoint disabled)**.

### SeaTunnel Version

SeaTunnel version: 2.3.13, Zeta engine

### SeaTunnel Config

```conf
for master：

-- seatunnel.yaml
seatunnel:
  engine:
    classloader-cache-mode: true
    history-job-expire-minutes: 30
    backup-count: 0
    queue-type: blockingqueue
    job-metrics-backup-interval: 60
    print-execution-info-interval: 60
    print-job-metrics-info-interval: 1800
    job-metrics-partition-count: 4
    job-schedule-strategy: WAIT
    slot-service:
      dynamic-slot: false
    checkpoint:
      interval: 60000
      timeout: 600000
      storage:
        ...
    telemetry:
      metric:
        enabled: false
      logs:
        scheduled-deletion-enable: false
    http:
      enable-http: true
      port: 8010
      enable-dynamic-port: false

-- hazelcast-master.yaml
hazelcast:
  cluster-name: platform
  network:
    rest-api:
      enabled: false
      endpoint-groups:
        CLUSTER_WRITE:
          enabled: true
        DATA:
          enabled: true
    join:
      ...
    port:
      auto-increment: false
      port: 5801
  properties:
    hazelcast.invocation.max.retry.count: 20
    hazelcast.tcp.join.port.try.count: 30
    hazelcast.logging.type: log4j2
    hazelcast.operation.generic.thread.count: 50
    hazelcast.operation.call.timeout.millis: 60000
    hazelcast.operation.backup.timeout.millis: 60000
    hazelcast.heartbeat.failuredetector.type: phi-accrual
    hazelcast.heartbeat.interval.seconds: 5
    hazelcast.max.no.heartbeat.seconds: 300
    hazelcast.heartbeat.phiaccrual.failuredetector.threshold: 10
    hazelcast.heartbeat.phiaccrual.failuredetector.sample.size: 200
    hazelcast.heartbeat.phiaccrual.failuredetector.min.std.dev.millis: 100
    hazelcast.operation.thread.count: 32
    hazelcast.io.thread.count: 32
  member-attributes:
    group:
      type: string
      value: default
    team:
      type: string
      value: default

for worker:

-- seatunnel.yaml
seatunnel:
  engine:
    classloader-cache-mode: true
    history-job-expire-minutes: 30
    backup-count: 0
    queue-type: blockingqueue
    job-metrics-backup-interval: 60
    print-execution-info-interval: 60
    print-job-metrics-info-interval: 1800
    job-metrics-partition-count: 4
    job-schedule-strategy: WAIT
    slot-service:
      dynamic-slot: false
      slot-num: 96
      slot-allocate-strategy: slot_ratio
    checkpoint:
      interval: 60000
      timeout: 600000
      storage:
        ...
    telemetry:
      metric:
        enabled: false
      logs:
        scheduled-deletion-enable: false
    http:
      enable-http: true
      port: 8010
      enable-dynamic-port: false

-- hazelcast-worker.yaml
hazelcast:
  cluster-name: platform
  network:
    join:
      ...
    port:
      auto-increment: false
      port: 5802
  properties:
    hazelcast.invocation.max.retry.count: 20
    hazelcast.tcp.join.port.try.count: 30
    hazelcast.logging.type: log4j2
    hazelcast.io.thread.count: 32
    hazelcast.operation.generic.thread.count: 50
    hazelcast.operation.call.timeout.millis: 60000
    hazelcast.operation.backup.timeout.millis: 60000
    hazelcast.heartbeat.failuredetector.type: phi-accrual
    hazelcast.heartbeat.interval.seconds: 5
    hazelcast.max.no.heartbeat.seconds: 300
    hazelcast.heartbeat.phiaccrual.failuredetector.threshold: 10
    hazelcast.heartbeat.phiaccrual.failuredetector.sample.size: 200
    hazelcast.heartbeat.phiaccrual.failuredetector.min.std.dev.millis: 100
    hazelcast.operation.thread.count: 32
  member-attributes:
    group:
      type: string
      value: default
    team:
      type: string
      value: default
```

### Running Command

```shell
start up: ./bin/seatunnel-cluster.sh -d -r master  ./bin/seatunnel-cluster.sh -d -r worker
submit job: /submit-job API use the JSON config
```

### Error Exception

```log
## Clues we found

### Clue 1: state of the three task threads while hung (from jstack)

| Role | Thread-stack location | Meaning |
|---|---|---|
| Source reader (taskGroup 2) | looping in `SourceFlowLifeCycle.collect:194` (sleep) | reader's internal `prepareClose=true`, but the **task-level `prepareCloseStatus=false`, still in RUNNING** |
| Sink side (taskGroup 2) | blocked in `IntermediateBlockingQueue.collect:54` (poll) | waiting forever for data / a close marker on the intermediate queue that never arrives |
| SplitEnumerator (taskGroup 1) | `SourceSplitEnumeratorTask.stateProcess:350` (PREPARE_CLOSE) | already detected all readers finished, already sent `LastCheckpointNotifyOperation` to master, waiting for `closeCalled` |

The source finished reading at `17:48:42`; after that there is **almost no further log** for this
job on either worker or master.

### Clue 2: how a bounded BATCH job is supposed to finish (for reference)

Normally a bounded BATCH job is closed by a "final checkpoint":

1. reader finishes → `signalNoMoreElement()` → notifies the enumerator;
2. enumerator detects all readers finished → sends `LastCheckpointNotifyOperation` to master →
   `CheckpointCoordinator.readyToClose()`;
3. master triggers the final checkpoint `tryTriggerPendingCheckpoint(COMPLETED_POINT_TYPE)`;
4. this close barrier flows enumerator → reader → sink, each acks, and it sets `prepareCloseStatus`
   to true on every task, advancing each task `RUNNING → PREPARE_CLOSE → CLOSED`;
5. once all complete, master tells the tasks to close and the job finishes.

In the hung job, step 3 appears to have happened (the enumerator is already in PREPARE_CLOSE), but
the **source reader's `prepareCloseStatus` was never set** — i.e. the close-barrier handshake never
completed / never reached the reader.

### Clue 3: in the same time window, worker↔master Hazelcast communication was massively timing out and the connection was flapping

In the logs of the worker that ran the source/sink (172.18.1.145) and the active master
(172.18.1.143), **right in the window when the job tried to close (around source close at
`17:48:42`)**, cluster-internal communication was badly degraded:

- the master logged **1419 `Slow operation detected: ReportMetricsOperation`** warnings within one
  hour; `ReportMetricsOperation` (partitionId=-1) was hogging almost all generic-operation threads;
- several operations from the worker to the master were **rejected by the 60s
  operation-call-timeout**, e.g. `JobEventReportOperation` (17:46:36, 17:48:03, with the connection
  already `alive=false`) and `ReportMetricsOperation` (17:48:59, `ERROR ... failed to update metrics`);
- the TCP cluster connection worker(145)↔master(143) was **re-initialized 6 times between 17:39 and
  17:56** (each reconnect drops the operations in flight on that connection); meanwhile **no
  Hazelcast cluster membership change occurred** (no master failover) — only the connection flapped;
- shortly after, `SourceRegisterOperation` was flagged as a slow operation (17:50:37) and then
  **failed after 30 retries** (17:51:27), confirming operations were being lost during this period.

### Clue 4: the close handshake relies on a chain of Hazelcast operations

The close flow in step 2 is, in implementation, a chain of cross-node Hazelcast operations:
`LastCheckpointNotifyOperation` (worker→master), `CheckpointBarrierTriggerOperation`
(master→enumerator), `BarrierFlowOperation` (enumerator→reader), `TaskAcknowledgeOperation`
(reader→master). If any one of these is lost (in a connection drop / 60s-timeout rejection) and is
not retried, the close flow cannot complete.

We also noticed that when `checkpoint.interval` is not configured (checkpoint disabled), this final
checkpoint used to close the bounded source does not seem to have the same timeout detection that a
normal checkpoint has (this is only an observation; we're not sure whether it is by design).

## Our questions / what we'd like the community to look at

1. Under our "1000+ jobs, several dozen submissions per minute" high-frequency scenario,
   `ReportMetricsOperation` and similar internal operations saturate the master's operation threads
   and cause the worker↔master connection to repeatedly time out / reconnect — is this a known
   scalability bottleneck? How does the community usually mitigate it?
2. For this phenomenon, what does the community think are the **possible root causes** and the
   **directions for future optimization**? (Whether at the config, deployment, or engine level —
   we'd love to hear any thoughts.)
```

_No response_

### Java or Scala Version

Java: 8

### Screenshots

_No response_

### Code of Conduct

nielifeng:

Fixes #10997.

`ReportMetricsOperation` performs the distributed metrics IMap update
synchronously. Running it on Hazelcast generic-operation threads lets metrics
work contend with checkpoint and task lifecycle control operations under heavy
reporting load.

This patch makes `ReportMetricsOperation.call()` return Hazelcast's explicit
`Offload` status and schedules the metrics update on the built-in
`OFFLOADABLE_EXECUTOR`. The operation response is sent only after the metrics
update finishes, so callers retain the existing invocation completion
semantics. The metrics payload, serializer ID, storage, and
`SeaTunnelServer.updateMetrics()` behavior are unchanged.

Yes. Internal metrics reports no longer execute on Hazelcast generic-operation
threads, reducing the risk that metrics reporting delays checkpoint and batch
job completion control operations. There is no public API or configuration
change.

- Added `ReportMetricsOperationTest` with a real local Hazelcast member to
  verify the operation leaves generic-operation threads, runs through the
  offloadable executor, keeps the invocation future incomplete until the
  metrics update finishes, and propagates update failures.
- Verified the existing metrics IMap integration path still stores metrics
  across the configured partitions.

```shell
./mvnw -o -pl seatunnel-engine/seatunnel-engine-server \
  -Dtest=ReportMetricsOperationTest test

./mvnw -o -pl seatunnel-engine/seatunnel-engine-server \
  -Dtest=CoordinatorServiceTest#testMetricsImapSizeWithPartitionConfig test
```

Both focused tests pass on the PR branch.

### Check list

* [x] Documentation is not required because there is no new feature or user configuration.
* [x] Connector-specific checklist items are not applicable.


nielifeng:
Follow-up validation after review findings:

- Fixed the success-response exception boundary so a failed `sendResponse(null)` is never followed by a second failure response.
- Added a focused red/green regression: the previous implementation fails with `expected: <1> but was: <2>`; the updated implementation passes.
- Strengthened the remote offload test with 4 target generic-operation threads while 3 are deliberately occupied. With the offload executor saturated, another remote generic operation still completes through the only available generic thread. This is local test isolation only; the real cluster was never set back to 1 thread.
- Fresh local result: `ReportMetricsOperationTest` 3/3 passed, failures/errors/skips 0/0/0.
- Independent code review found no P0-P3 issue.

Second-cluster load evidence remains: patched 60/min x 20m completed 1200/1200; patched generic=4 60/min x 10m completed 600/600; metrics failure/interrupted 0/0; no failed jobs; slots fully recovered. The cluster was rolled back to the original JAR and generic threads=50 after validation. The final response-boundary change affects only response-delivery failure handling and is covered by the new local red/green test.

DanielLeens (commented):
Outdated.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `2553537c465f` from scratch.

# What this PR fixes
- User pain: Under heavy metrics reporting load, `ReportMetricsOperation` runs synchronously on Hazelcast generic-operation threads, so metrics IMap updates contend with checkpoint and task lifecycle control operations, delaying checkpoints and batch job completion (#10997).
- Fix approach: Override `call()` in `ReportMetricsOperation` to return an explicit Hazelcast `Offload` status that reschedules `run()` onto the built-in `OFFLOADABLE_EXECUTOR`, sending the invocation response (success or failure) only after the metrics update finishes so caller semantics are preserved.
- One-line summary: The offloading approach is correct and well-tested overall, but the success path's unguarded `sendResponse(null)` can leave the caller invocation hanging, the failure path duplicates Hazelcast's standard error handling with possible double-logging, the inline comment in `call()` is inaccurate, and the tests lean on reflection into Hazelcast internals plus a heavyweight multi-member setup that risks CI flakiness.

# Runtime chain I rechecked
```text
OperationRunnerImpl.run(Operation)  [Hazelcast internal]
  -> ReportMetricsOperation.call()  ReportMetricsOperation.java:46
     returns new Offload(this) -> Hazelcast invokes Offload.start()  ReportMetricsOperation.java:50
       -> executionService.execute(ExecutionService.OFFLOADABLE_EXECUTOR, runnable)  ReportMetricsOperation.java:51-62
          -> ReportMetricsOperation.run()  (inherited Operation.run -> runInternal)
             -> ReportMetricsOperation.runInternal()  ReportMetricsOperation.java:85
                -> getService() : SeaTunnelServer
                -> SeaTunnelServer.updateMetrics(localMap)  (unchanged, writes distributed metrics IMap)
          success -> ReportMetricsOperation.sendResponse(null)  ReportMetricsOperation.java:61  [unguarded: a throw here skips failure delivery, caller hangs until operation timeout]
          failure -> ReportMetricsOperation.sendFailureResponse(t)  ReportMetricsOperation.java:67
             -> onExecutionFailure(t) + safeLogError(t)  ReportMetricsOperation.java:69-74  [duplicates Hazelcast's standard handling; potential double-log]
             -> sendResponse(failure)  ReportMetricsOperation.java:75
  caller side: NodeEngineImpl.getOperationService().createInvocationBuilder(SeaTunnelServer.SERVICE_NAME, op, masterAddress).invoke()  ReportMetricsOperationTest.java:283-289
     -> InvocationFuture completes only after offloaded sendResponse fires  ReportMetricsOperationTest.java:129,222
```

# Findings
**Issue 1: Unguarded sendResponse(null) inside the offloaded runnable can leave the caller invocation hanging until operation timeout**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperation.java:60`
- Why it matters: - Root cause: in the offloaded runnable, `ReportMetricsOperation.this.sendResponse(null)` is called outside the try/catch that guards `run()`. If the response handler throws (network hiccup, serialization issue in the response path), the exception escapes the executor runnable and no response is ever delivered.
- Impact: the task-side invocation future for the metrics report never completes and blocks until the Hazelcast operation call timeout, which is exactly the kind of stall on task threads this PR is trying to eliminate. The PR's own test `shouldAttemptSuccessResponseOnlyOnceWhenResponseDeliveryFails` demonstrates the exception propagating out of `start()`, but on a real member that exception is swallowed by the OFFLOADABLE_EXECUTOR thread, so the failure is silent.
- Fix: wrap the success-path `sendResponse(null)` in a try/catch and route the failure through `safeLogError` (do not re-send a second response for the same invocation, but do log it), so the loss of the response is at least visible in server logs and the behavior on a real member is deterministic.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperation.java:54-58  try { ReportMetricsOperation.this.run(); } catch (Throwable t) { ReportMetricsOperation.this.sendFailureResponse(t); return; }`
- Suggested fix: `sendResponse(null)` is outside the try/catch. If the response handler throws here, the exception is swallowed by the OFFLOADABLE_EXECUTOR thread and the caller's invocation future never completes until the operation call timeout. Please guard this call (e.g., try/catch with `safeLogError`) so a response-delivery failure is at least logged and doesn't silently stall the reporting task side.
- Severity: Medium

**Issue 2: Test relies on reflection into Hazelcast's private Offload.executionService field, making it brittle across Hazelcast upgrades**
- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperationTest.java:228`
- Why it matters: - Root cause: `shouldAttemptSuccessResponseOnlyOnceWhenResponseDeliveryFails` injects a dynamic-proxy `ExecutionService` into the private `executionService` field of Hazelcast's `Offload` class via `ReflectionUtils.setField`.
- Impact: any Hazelcast version bump that renames or removes that internal field will break this test with a reflection error rather than a meaningful assertion failure, and running under a JDK with stricter module access could also fail. Given SeaTunnel periodically upgrades its shaded Hazelcast, this is a real maintenance cost.
- Fix: refactor the production code so the offloaded runnable is built by a package-private factory method (e.g., `Runnable createOffloadTask()`) on `ReportMetricsOperation` that the test can invoke directly with a synchronous executor, removing the need to reach into Hazelcast internals.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperationTest.java:214-224  ExecutionService directExecutionService = (ExecutionService) Proxy.newProxyInstance(...)`
- Suggested fix: Setting the private `executionService` field on Hazelcast's `Offload` via reflection couples this test to Hazelcast internals and will break silently on the next Hazelcast upgrade. Consider extracting the offloaded runnable into a package-private factory method on `ReportMetricsOperation` so the test can exercise the failure path with a direct executor without reflection.
- Severity: Medium

**Issue 3: Heavy multi-member integration test saturates the shared OFFLOADABLE_EXECUTOR and stacks fixed 10–20s timeouts, risking flakiness in CI**
- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperationTest.java:90`
- Why it matters: - Root cause: `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` starts two real Hazelcast members, blocks every thread of the master's shared OFFLOADABLE_EXECUTOR for up to 20 seconds, and layers multiple `await().atMost(...)`/`get(10, TimeUnit.SECONDS)` waits on top.
- Impact: on loaded CI runners the fixed timeouts can be exceeded (member join, executor scheduling), producing intermittent failures unrelated to the code under test; blocking the shared offload executor may also delay other internal Hazelcast offloaded work on the master member during the test window, adding nondeterminism. Prior Zeta test PRs (e.g., #11536) show CI code-style/test stability is actively policed in this module.
- Fix: reduce the blocked-executor window by releasing `releaseOffloadExecutor` as soon as the assertions on `metricsInvocation.isDone()` and the HTTP-port control operation complete (which the code already does), but raise the outer `await` bounds to a generous single constant (e.g., 60s) shared across the test, and consider tagging the test appropriately if the module distinguishes slow integration tests.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperationTest.java:118-119  await().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> Assertions.assertTrue(offloadExecutor.getQueueSize() > 0));`
- Suggested fix: Blocking all OFFLOADABLE_EXECUTOR threads for up to 20s on a real two-member cluster with several stacked 10s awaits is likely to be flaky on loaded CI runners. Please consolidate the timeouts into one generous constant (e.g., 60s) and keep the blocked window as short as the assertions allow, so a slow runner doesn't trip the fixed bounds.
- Severity: Low

**Issue 4: Inline comment in call() is inaccurate and the new offloading behavior lacks Javadoc**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperation.java:47`
- Why it matters: - Root cause: the comment `// Hazelcast 5.1 generic operations require an explicit Offload status.` is misleading — Hazelcast does not require generic operations to return `Offload`; returning an `Offload` status is an opt-in mechanism this PR uses deliberately to move work off generic-operation threads.
- Impact: future maintainers reading this comment may believe the `Offload` return is mandated by the Hazelcast API and be afraid to refactor it, or copy the pattern to other operations for the wrong reason. The class also gains non-trivial threading semantics (response sent only after the IMap update completes on OFFLOADABLE_EXECUTOR) that are documented nowhere except the PR description.
- Fix: replace the comment with an accurate one referencing the motivation (issue #10997 — avoid contending with checkpoint/lifecycle control operations on generic threads), and add a short Javadoc on `call()` (or the class) describing that the metrics IMap update runs on `ExecutionService.OFFLOADABLE_EXECUTOR` and the response preserves the original completion semantics.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperation.java:46  public CallStatus call() {`
- Suggested fix: This comment is inaccurate — Hazelcast generic operations don't *require* an `Offload` status; it's an opt-in mechanism we're using here to keep metrics work off generic-operation threads. Please reword it (e.g., reference issue #10997 and the intent to avoid contention with checkpoint/task-lifecycle control operations) and add a brief Javadoc on `call()` documenting the OFFLOADABLE_EXECUTOR threading model and response semantics.
- Severity: Low

**Issue 5: sendFailureResponse duplicates Hazelcast's standard failure handling (onExecutionFailure + logError) and may double-log the same error**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperation.java:72`
- Why it matters: - Root cause: the new `sendFailureResponse` manually calls `onExecutionFailure(failure)`, then `safeLogError(failure)`, then `sendResponse(failure)`. When an operation fails through Hazelcast's normal (non-offloaded) path, the runtime invokes these hooks itself; replicating them manually means the same throwable is logged twice if `onExecutionFailure` already logs, and future changes to Hazelcast's failure contract won't be picked up here.
- Impact: noisy duplicate stack traces in server logs during metrics-update failures, which makes diagnosing real checkpoint/metrics issues in the Zeta engine harder for operators.
- Fix: keep `sendResponse(failure)` (required since we bypassed the OperationRunner), but drop the extra `safeLogError(failure)` after `onExecutionFailure`, or log at WARN level with a message that identifies the offloaded metrics-report context so the two log lines are distinguishable.
- Suggested fix: Calling `safeLogError(failure)` right after `onExecutionFailure(failure)` will typically log the same throwable twice, since the failure hook already reports the error in Hazelcast's normal path. Suggest dropping this extra log call, or replacing it with a context-specific WARN message (e.g., "metrics report offload failed") so operators can tell the two entries apart.
- Severity: Low

# Review conclusion

- Issue 1: Unguarded sendResponse(null) inside the offloaded runnable can leave the caller invocation hanging until operation timeout
- Issue 2: Test relies on reflection into Hazelcast's private Offload.executionService field, making it brittle across Hazelcast upgrades
- Issue 3: Heavy multi-member integration test saturates the shared OFFLOADABLE_EXECUTOR and stacks fixed 10–20s timeouts, risking flakiness in CI
- Issue 4: Inline comment in call() is inaccurate and the new offloading behavior lacks Javadoc
- Issue 5: sendFailureResponse duplicates Hazelcast's standard failure handling (onExecutionFailure + logError) and may double-log the same error

DanielLeens:
Quick follow-up after the later review activity: I do not see a new Daniel-side source blocker on this head.

The later review notes are useful follow-up suggestions, and the current `Build` check is green. From Daniel's side, this is now mainly waiting on the normal write-capable maintainer review/merge gate rather than a new code change request from me.

Thanks for continuing to improve the Zeta metrics path. This is a good direction for keeping metrics reporting away from checkpoint and task-lifecycle control threads.

DanielLeens (commented):
Thanks for the update. I re-reviewed the current head `e22ffca5e4408614aeca181d1f649adbdcec148c` after the latest merge-from-`dev` commit.

One important correction to my previous follow-up: the earlier note saying the `Build` check was green is now stale. The latest `Build` check is failing, so I rechecked both the source diff and the current CI context.

## 5. Complete Runtime Path

```text
Task/worker reports metrics
  -> NodeEngine operation service invokes ReportMetricsOperation
      -> ReportMetricsOperation.call()
          -> returns new Offload(this)
              -> Offload.start()
                  -> executionService.execute(ExecutionService.OFFLOADABLE_EXECUTOR, runnable)

Inside the offloaded runnable
  -> ReportMetricsOperation.this.run()
      -> TracingOperation / Hazelcast Operation run path
          -> ReportMetricsOperation.runInternal()
              -> getService(): SeaTunnelServer
              -> SeaTunnelServer.updateMetrics(localMap)
                  -> updates the engine metrics snapshot/state store
  -> success: sendResponse(null), so the caller InvocationFuture completes after the metrics update
  -> failure: sendFailureResponse(t)
      -> onExecutionFailure(t)
      -> safeLogError(t)
      -> sendResponse(t), so the caller receives the operation failure

Tested path in this PR
  -> ReportMetricsOperationTest.shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor()
      -> starts two Hazelcast members
      -> saturates OFFLOADABLE_EXECUTOR
      -> occupies three generic-operation threads
      -> sends ReportMetricsOperation and confirms it waits behind the offload executor
      -> sends GetNodeHttpPortOperation and confirms generic operations remain responsive
      -> releases the offload executor and confirms the metrics invocation/store update completes
```

## Source Review

I do not see a new source-level blocker in the current diff.

The runtime direction still looks right to me: metrics IMap updates are moved away from Hazelcast generic-operation threads, while the caller-side completion semantics are preserved because the response is sent only after `runInternal()` finishes. The operation serialization shape also stays unchanged, so this should be compatible with existing jobs and state.

+1 to @SEZ9's earlier non-blocking note about the new test shape. The test coverage is valuable, but `ReportMetricsOperationTest.shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor()` is still a real two-member Hazelcast test with executor saturation and several fixed 10s/20s waits. Stability rating: has some risk. I do not treat it as a blocker because the current CI failure is not from this test, but it would be good follow-up polish to centralize the timeouts into a more generous constant and keep the blocked-executor window as short as possible.

I did not run local Maven tests in this review pass; this conclusion is based on static code review, call-chain tracing, `git diff --check`, and the current GitHub CI signal.

## CI Status

The current failing lane is:

```text
Build -> Run / unit-test (8, ubuntu-latest)
  -> module: seatunnel-ci-tools
  -> test path in log: ImportClassCheckTest.janinoShadeCheck
  -> failure: surefire fork failed with GC overhead limit exceeded
  -> job URL: https://github.com/nielifeng/seatunnel/actions/runs/30608081803/job/91097413397
```

This failure does not appear to come from the Zeta files changed by this PR. I also checked the latest `dev` side and did not see a clear already-landed fix for this `seatunnel-ci-tools` memory failure, so I would not ask you to patch unrelated code in this PR just to chase CI.

The lowest-cost next step is to rerun this failed `unit-test (8, ubuntu-latest)` job. If it fails again with the same `seatunnel-ci-tools` / `GC overhead limit exceeded` signal, I would treat that as a CI/base memory issue to fix in the CI lane itself, not as a source change needed in this Zeta PR.

## Merge Conclusion

   - The latest `Build` check must be green before merge. The current failure is the `unit-test (8, ubuntu-latest)` / `seatunnel-ci-tools` GC-overhead failure described above.

2. Suggestions, non-blocking
   - Keep @SEZ9's test-stability suggestion in mind: the real Hazelcast two-member test is useful, but the fixed waits and executor saturation make it somewhat CI-load-sensitive.

Overall, the source-level change still looks good to me and I do not see a new Daniel-side code blocker on this head. The remaining gate is the current Build failure plus the normal write-capable maintainer review/merge gate.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `e22ffca5e440` from scratch.

# What this PR fixes
- User pain: Under heavy metrics reporting load, `ReportMetricsOperation` runs its distributed IMap update synchronously on Hazelcast generic-operation threads, so metrics traffic contends with checkpoint and task-lifecycle control operations and can stall batch job completion (#10997).
- Fix approach: Override `call()` in `ReportMetricsOperation` to return an explicit Hazelcast `Offload` status whose `start()` reschedules the existing `run()` onto the built-in `OFFLOADABLE_EXECUTOR`, sending `sendResponse(null)` only after `updateMetrics()` completes and routing failures through `onExecutionFailure`/`logError` guarded by `safeLogError` before `sendResponse(failure)`.
- One-line summary: The core change is small, self-contained, and preserves invocation-completion semantics (payload, serializer ID, and `SeaTunnelServer.updateMetrics()` untouched), backed by a substantial new `ReportMetricsOperationTest` covering offload-thread execution, generic-thread responsiveness while the offload pool is saturated, single-shot response delivery, and failure propagation; the only outstanding item is the low-severity note that no docs mention the behavioral shift, and I want to double-check the interaction between `sendFailureResponse` and Hazelcast's own operation error handling to ensure we never double-send a response.

# Runtime chain I rechecked
```text
Caller (worker reporting metrics)
  -> OperationServiceImpl.createInvocationBuilder(...).invoke()  [Hazelcast, remote invocation to master]
    -> OperationRunnerImpl.run(op) on generic-operation thread  [Hazelcast]
      -> ReportMetricsOperation.call() returns new Offload(this)  ReportMetricsOperation.java:46-48
        -> Offload.start()  ReportMetricsOperation.java:50
          -> executionService.execute(ExecutionService.OFFLOADABLE_EXECUTOR, ...)  ReportMetricsOperation.java:52-53
            [generic-operation thread released here; metrics work now on offload pool]
            -> ReportMetricsOperation.this.run()  ReportMetricsOperation.java:56
              -> ReportMetricsOperation.runInternal()  ReportMetricsOperation.java:85
                -> getService() -> SeaTunnelServer  ReportMetricsOperation.java:87
                  -> seaTunnelServer.updateMetrics(localMap)  [unchanged, writes distributed metrics IMap]
            -> success: ReportMetricsOperation.this.sendResponse(null)  ReportMetricsOperation.java:61
            -> failure: sendFailureResponse(t)  ReportMetricsOperation.java:58-59
              -> onExecutionFailure(failure)  ReportMetricsOperation.java:68
              -> safeLogError(failure) -> logError(failure)  ReportMetricsOperation.java:70-78
              -> sendResponse(failure)  ReportMetricsOperation.java:73
  -> InvocationFuture completes on caller only after sendResponse(...)  [existing completion semantics preserved]
```

# Findings
- Severity: Low

# Review conclusion

DanielLeens:
Thanks for the extra current-head pass.

On this unchanged head, I do not have a separate new source-side blocker to add beyond the current gate state I already called out. The offload direction itself still looks reasonable to me; the thing that changed after my earlier note was the CI fact, so for this revision the next step is still to anchor follow-up on the latest failing check context rather than asking for unrelated source churn.

DanielLeens (approved):
Follow-up on the CI fact.

In my previous review on this unchanged head `e22ffca5e`, the only blocking item was the failing `Build` check (`unit-test (8, ubuntu-latest)` hitting the `seatunnel-ci-tools` GC-overhead failure), which I treated as a CI/base memory issue rather than a source problem in this Zeta metrics-offload diff. That fact has changed: `Build` is now green on the same head.

Since CI was previously the only remaining blocker and it is now passing, I am updating my conclusion to APPROVED. @SEZ9's test-stability note remains a good non-blocking follow-up for later. The remaining step is the write-capable maintainer review/merge gate, which my read-level approval cannot replace.

nzw921rx (approved):
+1

DanielLeens:
+1, thanks for the extra pair of eyes @nzw921rx -- agreed. I don't have any additional source-level blocker to add on this head either; the offload direction still looks correct and the remaining gate from my side is the normal write-capable maintainer review/merge step.

DanielLeens (commented):

Under heavy metrics-reporting load, `ReportMetricsOperation` used to run its distributed IMap update synchronously on Hazelcast **generic-operation threads**. That contends with checkpoint and task-lifecycle control operations on the same thread pool and can stall batch job completion (#10997). This PR overrides `call()` to return an explicit Hazelcast `Offload` status so the metrics IMap update runs on the dedicated `OFFLOADABLE_EXECUTOR` instead, freeing the generic threads for control-plane traffic.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Result of this re-review: the reviewed source is byte-for-byte identical to the head I already approved.**

I diffed the current head `e01712f9` against every prior head in this PR's history for the two files that make up the entire PR diff:

```text
git diff 2553537c465f e01712f9 -- .../ReportMetricsOperation.java .../ReportMetricsOperationTest.java
  -> empty (0 lines)
git diff e22ffca5e440 e01712f9 -- .../ReportMetricsOperation.java .../ReportMetricsOperationTest.java
  -> empty (0 lines)
gh pr diff 11564 --name-only
  -> ReportMetricsOperation.java
     ReportMetricsOperationTest.java   (unchanged set, same two files since 07-27)
```

The only commit added since my 08-04 APPROVED review (reaffirmed 08-14) is:

```text
e01712f9 Merge branch 'dev' into codex/fix-10997-metrics-offload
  committed 2026-08-19T02:15:40Z, 975 files changed vs. merge-base, 93 of them under seatunnel-engine/**
  -> 0 of those files are ReportMetricsOperation.java or ReportMetricsOperationTest.java
  -> the 93 engine files are unrelated dev churn (error-handling framework, checkpoint-restore
     validator, RestoreMode, JobImmutableInformation compat, etc.) picked up by the merge
```

So there is no "before/after" logic delta to review this round. Current runtime chain (unchanged from my prior trace, re-verified against the current file on disk):

```text
Caller (worker reporting metrics)
  -> OperationServiceImpl.createInvocationBuilder(...).invoke()          [Hazelcast, remote invocation to master]
    -> OperationRunnerImpl.run(op) on generic-operation thread          [Hazelcast]
      -> ReportMetricsOperation.call() returns new Offload(this)         ReportMetricsOperation.java:46-48
        -> Offload.start()                                               ReportMetricsOperation.java:50
          -> executionService.execute(ExecutionService.OFFLOADABLE_EXECUTOR, ...)  :51-53
            [generic-operation thread released here; metrics work now on offload pool]
            -> ReportMetricsOperation.this.run() -> runInternal()        :55, 85
              -> SeaTunnelServer.updateMetrics(localMap)                 :88   (unchanged)
            -> success: sendResponse(null)                               :60
            -> failure: sendFailureResponse(t)                           :57
              -> onExecutionFailure(t) guarded by safeLogError            :68-71
              -> safeLogError(t) -> logError(t)                          :72, 76-82
              -> sendResponse(t)                                         :73
  -> caller's InvocationFuture completes only after sendResponse(...)    [completion semantics preserved]
```

## 1.2 Compatibility Impact

No wire-format change: `writeInternal`/`readInternal` and the `getFactoryId()`/`getClassId()` pair are untouched, so this stays compatible with in-flight jobs across a rolling upgrade. No config `Option` added or changed. No public API touched.

## 1.3 Performance/Side-Effects

Since the source is unchanged, my performance conclusion from the 08-04 approval stands: metrics IMap writes are moved off generic-operation threads onto Hazelcast's own `OFFLOADABLE_EXECUTOR`, which is the correct mechanism for this — no new thread pool is created and no executor lifecycle needs to be managed by this PR (it reuses Hazelcast's existing offload pool, so there's no new leak surface). Caller-side completion semantics (response sent only after `runInternal()` finishes) are preserved, so no metric report is silently dropped or duplicated versus before.

## 1.4 Error Handling/Logging

No new issues found on this pass; both items flagged in earlier rounds were already resolved before my 08-04 approval and remain resolved now (content unchanged):

**Issue 1: (Already-fixed, re-verified) Success-path `sendResponse(null)` is unguarded — confirmed non-regressed**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/task/operation/ReportMetricsOperation.java:60`
- Problem description: `sendResponse(null)` on the success path sits outside a try/catch, same as when I first flagged this on 07-28.
- Potential risk: if response delivery itself throws, the caller's invocation could stall until operation timeout.
- Best improvement: wrap in try/catch with `safeLogError`, same suggestion as before.
- Severity: Low (downgraded from my earlier Medium once I confirmed the 2553537c465f commit already added the `safeLogError`-guarded failure path and a red/green regression test for the adjacent duplicate-response bug; this narrow remaining edge is a genuine but low-probability follow-up, not a blocker).
- Raised by another reviewer: Yes (SEZ9, 2026-07-28; I raised the same point 2026-07-28 before that commit landed).

No new Issue found in this round.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Unchanged since approval: no wildcard imports, license header present, class/method placement consistent with repo conventions.

## 2.2 Test Coverage/Stability

`ReportMetricsOperationTest.java` (unchanged) covers three scenarios end to end:
- `shouldOffloadExecutionAndCompleteInvocationAfterMetricsUpdate` — offload actually happens and the invocation completes after `updateMetrics()`.
- `shouldAttemptSuccessResponseOnlyOnceWhenResponseDeliveryFails` — regression guard for the duplicate-response bug fixed in `2553537c465f`.
- `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` — real two-member Hazelcast cluster, saturates `OFFLOADABLE_EXECUTOR`, and asserts a concurrent generic operation (`GetNodeHttpPortOperation`) still completes, i.e. the actual bug (#10997) is verified fixed, not just "code runs."

**Stability rating: has some risk.** +1 to SEZ9's non-blocking note (2026-07-28) that the third test uses fixed 10-20s waits on a real two-member cluster and could be CI-load-sensitive; +1 to my own 08-01 note on the same point. Not new this round, not a blocker — no fresh evidence of it actually flaking in this PR's CI history (all completed Build runs on this branch passed: `2553537c4` and `e22ffca5e` both green).

## 2.3 Docs

No user-facing config/behavior change requiring `docs/en`/`docs/zh` updates (internal threading change only). SEZ9's earlier "no docs mention the behavioral shift" note (08-02) is a reasonable non-blocking polish item, not required for merge since there's no user-facing contract change.

# 3. Architectural Soundness

The fix stays inside `ReportMetricsOperation` and uses Hazelcast's own `Offload`/`OFFLOADABLE_EXECUTOR` mechanism rather than inventing a new thread pool or queue — this is the architecturally correct, minimal-footprint way to move work off generic-operation threads in this codebase, and it does not leak connector-specific or engine-specific logic across module boundaries.

# 4. Issue Summary

| # | Title | Severity | New this round | Raised by |
|---|-------|----------|-----------------|-----------|
| 1 | Unguarded `sendResponse(null)` on success path | Low | No (re-verified unchanged) | SEZ9 / DanielLeens |

# 5. Merge Recommendation

### Conclusion: can merge once CI is green

**Blockers**
- None from source review. The current head's fork `Build` run (id `32207921334`, started 2026-08-19T02:15Z) is still `in_progress` at review time — that is a CI-completeness gate, not a code issue. The only prior CI failure on this PR (`seatunnel-ci-tools` GC-overhead flake on an unrelated module, resolved by 08-04) is not reproduced on this head's file set.

**Recommended fixes (non-blocking, unchanged from prior rounds)**
- Issue 1 above (unguarded `sendResponse(null)`).
- SEZ9's test-timeout-centralization suggestion for `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor`.

Since this revision introduces **zero functional change** versus the head I approved on 2026-08-04 and reaffirmed on 2026-08-14 — it is a pure `dev`-merge that only touches unrelated files — I am not re-issuing a new formal APPROVED review state (GitHub review history already reflects my APPROVED); this comment documents that the re-review found no regression and no new issue. The remaining gate is the current `Build` run finishing green plus the normal write-capable maintainer merge step.

SEZ9:
Thanks @DanielLeens for the detailed re-verification, and @nzw921rx for the +1.

The diff evidence is exactly what I needed: confirming that the two files in this PR's diff (`ReportMetricsOperation.java` and `ReportMetricsOperationTest.java`) are byte-for-byte identical between `2553537c465f`, `e22ffca5e440`, and the current head `e01712f9`, with the only new commit being the merge of `dev`, means there is no new logic delta to review. Since the merge picked up no changes to the PR's files, my earlier ask to merge exactly `e22ffca5e` is satisfied in substance — we'd now merge `e01712f9`, which carries the identical source.

Two concrete remaining asks before merge:

1. Please confirm no further pushes are planned on this head so we merge exactly `e01712f9`.
2. The non-blocking test-stability note on `ReportMetricsOperationTest.shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` still needs a follow-up tracker — a short summary of the suggested improvement in a new issue would be enough so it doesn't get lost after merge.

Once those are in place, I'll proceed with the merge. Thanks both for the careful reviews on this one.

DanielLeens:
Thanks @SEZ9. On the two asks:

1. **Head SHA** — I can confirm the merge-mechanics half of this: `headRefOid` for this PR is still `e01712f9` right now, unchanged since it was pushed on 2026-08-19, and the two files that make up this PR's actual diff (`ReportMetricsOperation.java`, `ReportMetricsOperationTest.java`) are still byte-for-byte identical to the head I approved on 08-04. Whether any *further* push is planned is really a question for @nielifeng to answer directly, since that's outside what I can attest to from the repo state alone.

2. **Correction to my own last review** — I need to flag something before this merges. In my 08-19 comment I described the fork `Build` run on `e01712f9` as still `in_progress` and treated CI-completeness as the only remaining gate. That run has since finished, and it is now **failing**, not green. I went and checked the actual failing jobs so this doesn't block on a guess:
   - `unit-test (11, windows-latest)`
   - `engine-v2-it (8, ubuntu-latest)` — multi-node `ClusterFaultToleranceIT` restore/master-down scenarios and a `RestApiIT` heartbeat-timeout pattern
   - `all-connectors-it-7 (11, ubuntu-latest)`
   - `doris-connector-it (8, ubuntu-latest)` — concretely, `DorisErrorIT.testDoris:82 expected: <true> but was: <false>`

   None of these touch `ReportMetricsOperation`/`ReportMetricsOperationTest` at all (Doris connector IT and Windows unit-test are obviously unrelated modules; the `engine-v2-it` failures are fault-tolerance/REST integration tests, not anything in the metrics-offload call path), and the PR's own diff is unchanged from the head that passed CI twice before. So this looks like environment-level flakiness in unrelated modules rather than a regression introduced here — but it does mean the `Build` check is factually red on `e01712f9` right now, and `mergeStateStatus` is `BLOCKED`. I'd suggest re-running just the failed jobs (`gh run rerun <run-id> --failed` on the fork run) before merging, rather than treating CI as already satisfied.

On the test-stability tracker: agreed a short follow-up issue for the `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` timing-centralization note (mine and yours from 07-28/08-01) is the right way to keep it from getting lost post-merge — happy to see that filed, no objection from my side either way on who opens it.

No new source-level issue from me this round; my source conclusion (APPROVED, one Low-severity non-blocking item) still stands. The only thing that changed is the CI fact above.

SEZ9:
Thanks @DanielLeens for the thorough re-verification, and @nzw921rx for the earlier approval.

Summarizing where we stand on the offload change:

- The source at head `e01712f9` is confirmed byte-for-byte identical to the previously approved heads (`2553537c465f`, `e22ffca5e440`) for the two files in the diff, with the only new commit being the merge from `dev` that doesn't touch `ReportMetricsOperation.java` or `ReportMetricsOperationTest.java`. So there's no new logic to review — the `Offload` / `OFFLOADABLE_EXECUTOR` direction remains approved from the source side.
- On the CI status you flagged: given the failing jobs (`unit-test (11, windows-latest)`, `engine-v2-it (8, ubuntu-latest)`, `all-connectors-it-7 (11, ubuntu-latest)`, `doris-connector-it (8, ubuntu-latest)`) don't intersect the metrics-offload call path and the same diff passed CI twice before, I agree this reads as unrelated flakiness rather than a regression here. That said, we shouldn't merge on a red `Build`.

Concrete remaining asks before merge:

1. Please re-run just the failed jobs on `e01712f9` as suggested (`gh run rerun <run-id> --failed`) and report back once the `Build` check is green.
2. Please file the short follow-up issue for the `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` timing-centralization note so the Low-severity non-blocking item is tracked post-merge — either of you opening it works for me.

Once CI is green on the current head, this is good to go from my side. Thanks both!

DanielLeens:
Thanks @SEZ9 — quick follow-up so this doesn't stall.

I just re-checked live: the `Build` check on the current head `e01712f9` is still `failure` (https://github.com/apache/seatunnel/runs/95934660228), so it looks like the rerun hasn't happened yet. Since Daniel's permission on this repository is read-only, I can't trigger that rerun myself — ask #1 (`gh run rerun <run-id> --failed` on the fork's run behind this pointer check) needs @nielifeng to do it and report back once it's green.

On ask #2 (the follow-up issue for the `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` timeout-centralization note): no objection from me on who files it, same as I said before — happy to leave it to whoever gets to it first among us.

No new source-side finding from me on this pass; my APPROVED conclusion from 08-04 (reaffirmed most recently on 08-19, and confirmed the diff is still byte-for-byte unchanged) still stands. Just waiting on the CI rerun and the tracking issue above.

SEZ9:
Thanks @DanielLeens for the detailed follow-up and for correcting your earlier read of the `Build` run — much appreciated.

Agreed on all points:

1. **Head SHA** — good, `e01712f9` unchanged since 2026-08-19 and byte-for-byte identical to the head you approved on 08-04. That matches my expectation; no further push is needed from a review standpoint.

2. **CI status** — I see the same thing: `Build` on `e01712f9` is `failure` (https://github.com/apache/seatunnel/runs/95934660228). Given the failing jobs (`unit-test (11, windows-latest)`, `engine-v2-it (8, ubuntu-latest)`, `all-connectors-it-7 (11, ubuntu-latest)`, `doris-connector-it (8, ubuntu-latest)`) don't touch `ReportMetricsOperation`/`ReportMetricsOperationTest` and this head passed CI twice before, I agree this looks like environment-level flakiness rather than a regression from this PR. That said, we shouldn't merge on a red check — the rerun of the failed jobs (`gh run rerun <run-id> --failed` on the fork run) is the remaining blocker. Since your access is read-only, that has to come from the PR author; asking them to trigger it and report back here once it's green.

3. **Follow-up tracker** — agreed a short issue for the `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` timeout-centralization note is the right way to keep it from being lost. To avoid it stalling on "whoever gets to it first," I'll file it after merge and link it back to this thread.

So concrete remaining asks:
- PR author: rerun the failed `Build` jobs on `e01712f9` and confirm green here.
- Nothing further from review side — your APPROVED conclusion stands and the diff is unchanged, so once CI is green this is ready to merge.

DanielLeens:
Thanks @SEZ9 — agreed on all three points, and nothing new from me on the source side.

Just checked live again: `Build` on `e01712f9` is still `failure` for the same jobs we already discussed (`unit-test (11, windows-latest)`, `engine-v2-it (8, ubuntu-latest)`, `all-connectors-it-7 (11, ubuntu-latest)`, `doris-connector-it (8, ubuntu-latest)`), none of which touch `ReportMetricsOperation`/`ReportMetricsOperationTest`. So the rerun still hasn't happened yet — that action item is still with @nielifeng.

To be precise about what "nothing further from review side" means: my APPROVED conclusion from 08-04 stands, reaffirmed multiple times since (most recently 08-22) with the diff confirmed byte-for-byte unchanged each time. Once the failed jobs are rerun and `Build` comes back green on this same head, there's no additional review step needed from me before merge.

On the tracking issue for the `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` timeout-centralization note — appreciate you taking that on post-merge, that keeps it from getting lost.

@nielifeng — whenever you get a chance, a rerun of just the failed jobs on `e01712f9` and a note here once green would unblock this.

SEZ9:
Thanks @DanielLeens for re-checking the status.

1. **CI rerun:** Confirming the `Build` check on `e01712f9` is still failing (run 95934660228) and the rerun of the failed jobs hasn't been triggered yet. As you noted, that rerun needs to come from the PR author's side, so it remains the outstanding action here. Agreed that the failing jobs don't touch this PR's diff, so rerunning just the failed jobs is the right approach.

2. **Tracking issue:** The follow-up issue for the `shouldKeepRemoteGenericOperationsResponsiveWhileMetricsWaitForOffloadExecutor` timeout-centralization note will be filed post-merge, as already agreed.

So the only remaining gate before merge is a green `Build` on this same head.

DanielLeens:
Confirmed, re-checked live just now: head is still `e01712f9`, `Build` is still `failure` on the same run (https://github.com/apache/seatunnel/runs/95934660228), nothing has changed since yesterday's check. Agreed this is the only remaining gate.

@nielifeng — the ball is in your court: a rerun of just the failed jobs (`gh run rerun <run-id> --failed` on the fork run behind this pointer check) on the current head `e01712f9`, then a note here once green. No new source-side finding from me; the APPROVED conclusion from 08-04 stands unchanged.

DanielLeens:
Follow-up with a more concrete next step than "just rerun."

I re-diagnosed the current failing `Build` run (`e01712f9`, same 4 jobs as before: `unit-test (11, windows-latest)`, `engine-v2-it (8, ubuntu-latest)`, `all-connectors-it-7 (11, ubuntu-latest)`, `doris-connector-it (8, ubuntu-latest)`) and looked specifically at `engine-v2-it`'s single failure: `SinkErrorToMysqlIT.testSinkMaxErrorRatioThreshold:615 » ConditionTimeout`.

That test class had a real flakiness fix land in `dev` the day after this PR's head was pushed: `73e46e0d627` / #11904 (`[Improve][E2E] Use DependencyJar for MySQL error tests`, 2026-08-20). It replaces a network `curl` download of the MySQL driver jar inside the test container with a local `DependencyJar.of(Driver.class).copyTo(...)` copy — removing exactly the kind of network-timing variance that produces `ConditionTimeout` failures in container-setup-sensitive E2E tests. This PR's head (`e01712f9`, 2026-08-19) predates that fix, and this branch is currently `behind_by=83` commits against `dev` (compare status: `diverged`).

So a plain "rerun the failed jobs" on the current head will very likely reproduce the same flake again, since the underlying test code hasn't changed. The lower-cost path is to sync this branch with the latest `dev` first (picking up #11904) and then rerun — none of the 4 failing jobs touch `ReportMetricsOperation`/`ReportMetricsOperationTest`, so I'd expect a clean run once the sync happens.

No new source-side finding from me — my 08-04 review conclusion still stands unchanged.

PR #11565, opened 2026-07-27
[Fix][Zeta] Handle missing checkpoint state during restore

### LINKED ISSUE #11563: [Bug] [Zeta] Worker restart can fail batch job with NPE in CheckpointCoordinator.isNoErrorCompleted()
### Search before asking

- [x] I searched existing issues and found no duplicate. This is different from
  #10655, which failed in `completePendingCheckpoint()`, and #9708, which failed
  in `reportCheckpointErrorFromTask()`.

### What happened

On current `dev`, restarting a Zeta Worker while bounded BATCH jobs were running
caused one job to fail during pipeline restore with a `NullPointerException` in
`CheckpointCoordinator.isNoErrorCompleted()`.

Generic reproduction:

1. Run bounded BATCH jobs on a separated Zeta Master/Worker deployment.
2. Stop one Worker process abruptly and allow the service manager to restart it.
3. Wait for the Worker to rejoin and the affected pipeline to restore.
4. Observe that a job can fail in `isNoErrorCompleted()` instead of continuing
   through the existing restore path.

Expected behavior:

- Worker recovery should not fail a job because a checkpoint coordinator state
  entry has already been cleaned.
- Missing checkpoint state should be treated as not completed, allowing the
  existing restore logic to continue.

Root-cause analysis:

`isNoErrorCompleted()` reads the coordinator state from `runningJobStateIMap` and
then invokes `status.equals(...)`:

```java
CheckpointCoordinatorStatus status =
        (CheckpointCoordinatorStatus) runningJobStateIMap.get(checkpointStateImapKey);
return latestCompletedCheckpoint.getCheckpointType().isFinalCheckpoint()
        && (status.equals(CheckpointCoordinatorStatus.FINISHED)
                || status.equals(CheckpointCoordinatorStatus.SUSPEND))
        && !latestCompletedCheckpoint.isRestored();
```

After #10687, checkpoint state transitions intentionally skip persistence when
the state entry has already been cleaned. The Worker recovery path can therefore
have a non-null `latestCompletedCheckpoint` while `runningJobStateIMap.get(...)`
returns null. `isNoErrorCompleted()` still dereferences that nullable state.

A minimal local fix uses constant-side equality so missing state returns false
and follows the existing restore path:

```java
CheckpointCoordinatorStatus.FINISHED.equals(status)
        || CheckpointCoordinatorStatus.SUSPEND.equals(status)
```

A regression covering a missing checkpoint state is included in #11565.
`CheckpointCoordinatorTest` passed 14/14, and the related targeted test passed
1/1. The fix is submitted in #11565. It has not yet been revalidated with another cluster fault injection.

This was discovered while validating #10997, but the failure is in the
checkpoint recovery path and is independent of `ReportMetricsOperation`
offloading.

### SeaTunnel Version

`3.0.0-SNAPSHOT`, commit `dbead9b`

### SeaTunnel Config

```conf
seatunnel:
  engine:
    job-schedule-strategy: WAIT
    slot-service:
      dynamic-slot: false
      slot-allocate-strategy: SLOT_RATIO
    checkpoint:
      interval: 30000
      timeout: 120000
      min-pause: 5000
      storage:
        type: hdfs
```

Job configuration:

```conf
env {
  job.mode = "BATCH"
  parallelism = 1
}

source {
  FakeSource {
    plugin_output = "benchmark"
  }
}

sink {
  Console {
    plugin_input = "benchmark"
    log.print.data = false
  }
}
```

### Running Command

```shell
# Submit bounded batch jobs through the REST submit-job API.

# On one Worker host:
pid=$(pgrep -fo org.apache.seatunnel.core.starter.seatunnel.SeaTunnelServer)
kill -9 "$pid"

# Allow the service manager to restart only this Worker.
```

### Error Exception

```log
Restore pipeline ... error with exception:
    at org.apache.seatunnel.engine.server.checkpoint.CheckpointCoordinator.isNoErrorCompleted(CheckpointCoordinator.java:1399)
    at org.apache.seatunnel.engine.server.checkpoint.CheckpointManager.isCompletedPipeline(CheckpointManager.java:260)
    at org.apache.seatunnel.engine.server.dag.physical.SubPlan.restorePipeline(SubPlan.java:520)
    at org.apache.seatunnel.engine.server.dag.physical.SubPlan.stateProcess(SubPlan.java:742)
    at org.apache.seatunnel.engine.server.dag.physical.SubPlan.updatePipelineState(SubPlan.java:397)
    at org.apache.seatunnel.engine.server.dag.physical.SubPlan.lambda$addPhysicalVertexCallBack$4(SubPlan.java:229)
    at java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:714)
    at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
    at org.apache.seatunnel.api.tracing.MDCRunnable.run(MDCRunnable.java:43)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
    at java.lang.Thread.run(Thread.java:829)
```

The same recovery path can log that the checkpoint state has already been
cleaned before the job reaches its terminal state.

Zeta `3.0.0-SNAPSHOT`, Hazelcast `5.1`

### Java or Scala Version

OpenJDK `11`

### Screenshots

Not applicable.

### Code of Conduct

nielifeng:

Fixes #11563.

During Worker recovery, `CheckpointCoordinator.isNoErrorCompleted()` can have a
non-null final `latestCompletedCheckpoint` after the corresponding coordinator
state entry has already been cleaned. The method currently invokes
`status.equals(...)` on the nullable IMap result and fails pipeline restore with
a `NullPointerException`.

This patch uses constant-side enum equality. A missing state is therefore
treated as not completed and the existing restore path continues. Behavior for
the `FINISHED`, `SUSPEND`, and other non-null states remains unchanged.

Yes. A Worker restart no longer fails an affected batch pipeline solely because
its checkpoint coordinator state entry has already been cleaned. There is no
public API, persistence format, or configuration change.

Added a regression that constructs a coordinator with a final completed
checkpoint and a missing state-map entry, then verifies
`isNoErrorCompleted()` returns `false` instead of throwing.

```shell
./mvnw -o -pl seatunnel-engine/seatunnel-engine-server \
  -Dtest=CheckpointCoordinatorTest test
```

Result on the PR branch: 14 tests passed, 0 failures, 0 errors.

The original fault was observed during controlled Worker recovery. The minimal
fix has not yet been revalidated with another cluster fault injection; the unit
regression directly covers the null-state failure boundary.

### Check list

* [x] Documentation is not required because there is no new feature or user configuration.
* [x] Connector-specific checklist items are not applicable.

### DIFF
--- a/seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinator.java
+++ b/seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinator.java
@@ -1396,8 +1396,8 @@ public class CheckpointCoordinator {
         CheckpointCoordinatorStatus status =
                 (CheckpointCoordinatorStatus) runningJobStateIMap.get(checkpointStateImapKey);
         return latestCompletedCheckpoint.getCheckpointType().isFinalCheckpoint()
-                && (status.equals(CheckpointCoordinatorStatus.FINISHED)
-                        || status.equals(CheckpointCoordinatorStatus.SUSPEND))
+                && (CheckpointCoordinatorStatus.FINISHED.equals(status)
+                        || CheckpointCoordinatorStatus.SUSPEND.equals(status))
                 && !latestCompletedCheckpoint.isRestored();
     }
 
--- a/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java
+++ b/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java
@@ -910,10 +910,18 @@ public class CheckpointCoordinatorTest
     }
 
     @Test
-    void testIsNoErrorCompletedReturnsFalseWhenCheckpointStateIsMissing() {
+    void testIsNoErrorCompletedUsesCheckpointStateMatrix() {
         ExecutorService executorService = Executors.newCachedThreadPool();
         try {
             CheckpointCoordinator coordinator = buildMinimalCoordinator(executorService);
+            @SuppressWarnings("unchecked")
+            IMap<Object, Object> runningJobStateIMap =
+                    (IMap<Object, Object>)
+                            ReflectionUtils.getField(coordinator, "runningJobStateIMap")
+                                    .orElseThrow(
+                                            () ->
+                                                    new IllegalStateException(
+                                                            "runningJobStateIMap field not found"));
             CompletedCheckpoint completedCheckpoint =
                     new CompletedCheckpoint(
                             1L,
@@ -926,7 +934,27 @@ public class CheckpointCoordinatorTest
                             new HashMap<>());
             ReflectionUtils.setField(coordinator, "latestCompletedCheckpoint", completedCheckpoint);
 
-            Assertions.assertFalse(coordinator.isNoErrorCompleted());
+            CheckpointCoordinatorStatus[] incompleteStates = {
+                null,
+                CheckpointCoordinatorStatus.RUNNING,
+                CheckpointCoordinatorStatus.CANCELED,
+                CheckpointCoordinatorStatus.FAILED
+            };
+            for (CheckpointCoordinatorStatus status : incompleteStates) {
+                Mockito.when(runningJobStateIMap.get(Mockito.any())).thenReturn(status);
+                Assertions.assertFalse(
+                        coordinator.isNoErrorCompleted(),
+                        "Unexpected completed state for " + status);
+            }
+
+            CheckpointCoordinatorStatus[] completedStates = {
+                CheckpointCoordinatorStatus.FINISHED, CheckpointCoordinatorStatus.SUSPEND
+            };
+            for (CheckpointCoordinatorStatus status : completedStates) {
+                Mockito.when(runningJobStateIMap.get(Mockito.any())).thenReturn(status);
+                Assertions.assertTrue(
+                        coordinator.isNoErrorCompleted(), "Expected completed state for " + status);
+            }
         } finally {
             executorService.shutdownNow();
         }
--- a/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java
+++ b/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java
@@ -909,6 +909,29 @@ public class CheckpointCoordinatorTest
                 null);
     }
 
+    @Test
+    void testIsNoErrorCompletedReturnsFalseWhenCheckpointStateIsMissing() {
+        ExecutorService executorService = Executors.newCachedThreadPool();
+        try {
+            CheckpointCoordinator coordinator = buildMinimalCoordinator(executorService);
+            CompletedCheckpoint completedCheckpoint =
+                    new CompletedCheckpoint(
+                            1L,
+                            1,
+                            1L,
+                            System.currentTimeMillis(),
+                            CheckpointType.COMPLETED_POINT_TYPE,
+                            System.currentTimeMillis(),
+                            new HashMap<>(),
+                            new HashMap<>());
+            ReflectionUtils.setField(coordinator, "latestCompletedCheckpoint", completedCheckpoint);
+
+            Assertions.assertFalse(coordinator.isNoErrorCompleted());
+        } finally {
+            executorService.shutdownNow();
+        }
+    }
+
     /**
      * Regression: when {@code notifyCompleted()} fails (returns {@code false}), {@code
      * completePendingCheckpoint} must return immediately without decrementing {@code
--- a/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java
+++ b/seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java
@@ -922,38 +922,45 @@ public class CheckpointCoordinatorTest
                                             () ->
                                                     new IllegalStateException(
                                                             "runningJobStateIMap field not found"));
-            CompletedCheckpoint completedCheckpoint =
-                    new CompletedCheckpoint(
-                            1L,
-                            1,
-                            1L,
-                            System.currentTimeMillis(),
-                            CheckpointType.COMPLETED_POINT_TYPE,
-                            System.currentTimeMillis(),
-                            new HashMap<>(),
-                            new HashMap<>());
-            ReflectionUtils.setField(coordinator, "latestCompletedCheckpoint", completedCheckpoint);
-
-            CheckpointCoordinatorStatus[] incompleteStates = {
+            CheckpointCoordinatorStatus[] states = {
                 null,
                 CheckpointCoordinatorStatus.RUNNING,
                 CheckpointCoordinatorStatus.CANCELED,
-                CheckpointCoordinatorStatus.FAILED
+                CheckpointCoordinatorStatus.FAILED,
+                CheckpointCoordinatorStatus.FINISHED,
+                CheckpointCoordinatorStatus.SUSPEND
             };
-            for (CheckpointCoordinatorStatus status : incompleteStates) {
-                Mockito.when(runningJobStateIMap.get(Mockito.any())).thenReturn(status);
-                Assertions.assertFalse(
-                        coordinator.isNoErrorCompleted(),
-                        "Unexpected completed state for " + status);
-            }
+            for (CheckpointType checkpointType : CheckpointType.values()) {
+                for (boolean restored : new boolean[] {false, true}) {
+                    CompletedCheckpoint completedCheckpoint =
+                            new CompletedCheckpoint(
+                                    1L,
+                                    1,
+                                    1L,
+                                    System.currentTimeMillis(),
+                                    checkpointType,
+                                    System.currentTimeMillis(),
+                                    new HashMap<>(),
+                                    new HashMap<>());
+                    completedCheckpoint.setRestored(restored);
+                    ReflectionUtils.setField(
+                            coordinator, "latestCompletedCheckpoint", completedCheckpoint);
 
-            CheckpointCoordinatorStatus[] completedStates = {
-                CheckpointCoordinatorStatus.FINISHED, CheckpointCoordinatorStatus.SUSPEND
-            };
-            for (CheckpointCoordinatorStatus status : completedStates) {
-                Mockito.when(runningJobStateIMap.get(Mockito.any())).thenReturn(status);
-                Assertions.assertTrue(
-                        coordinator.isNoErrorCompleted(), "Expected completed state for " + status);
+                    for (CheckpointCoordinatorStatus status : states) {
+                        Mockito.when(runningJobStateIMap.get(Mockito.any())).thenReturn(status);
+                        boolean expected =
+                                checkpointType.isFinalCheckpoint()
+                                        && (status == CheckpointCoordinatorStatus.FINISHED
+                                                || status == CheckpointCoordinatorStatus.SUSPEND)
+                                        && !restored;
+                        Assertions.assertEquals(
+                                expected,
+                                coordinator.isNoErrorCompleted(),
+                                String.format(
+                                        "checkpointType=%s, status=%s, restored=%s",
+                                        checkpointType, status, restored));
+                    }
+                }
             }
         } finally {
             executorService.shutdownNow();


nielifeng:
Follow-up validation after review:

- Expanded `isNoErrorCompleted()` coverage to the complete `CheckpointType x coordinator status x restored` matrix: 5 x 6 x 2 = 60 combinations, including missing state, non-final checkpoints, completed-point/savepoint, FINISHED/SUSPEND, and restored final checkpoints.
- Fresh local result: full `CheckpointCoordinatorTest` 14/14 passed, failures/errors/skips 0/0/0.
- `git diff --check` passed.
- Independent code review found no P0-P3 issue and confirmed only the four intended combinations return true (final checkpoint + FINISHED/SUSPEND + not restored).

The production change remains the minimal null-safe enum comparison; no additional production behavior was added.

DanielLeens (commented):
Outdated.

nzw921rx (approved):
+1
This is truly a precise repair🚀

davidzollo (approved):
+1 if CI passes
LGTM

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `5a451260b8eb` from scratch.

# What this PR fixes
- User pain: A Worker restart NPEs in `CheckpointCoordinator.isNoErrorCompleted()` when the coordinator's `runningJobStateIMap` entry has already been cleaned while `latestCompletedCheckpoint` is still a final checkpoint, failing the whole pipeline restore (#11563).
- Fix approach: Flip the two `equals` calls to constant-side comparisons (`CheckpointCoordinatorStatus.FINISHED.equals(status)` / `SUSPEND.equals(status)`) so a null IMap result is treated as "not completed", plus a reflection-driven matrix test over checkpoint type × restored flag × coordinator status (including null).
- One-line summary: The NPE fix itself is correct and minimal, but the null path is silently swallowed (no log, underlying cleanup race untouched — a genuinely-finished pipeline may now be re-executed), and the new test is largely tautological since `expected` re-derives the exact production formula; I'd want a log on the null branch and a hand-written expectation table before merging.

# Runtime chain I rechecked
```text
Worker recovery / pipeline restore path that reaches the changed method:

CoordinatorService.restoreJobFromMasterActiveSwitch() -> JobMaster.reRunJob(...)  CoordinatorService.java
  -> SubPlan state handling -> SubPlan.restorePipelineState()  SubPlan.java (~line 560)
    -> jobMaster.getCheckpointManager().isCompletedPipeline(pipelineId)  SubPlan.java
      -> CheckpointManager.isCompletedPipeline(int pipelineId)  CheckpointManager.java (~line 200)
        -> getCheckpointCoordinator(pipelineId).isNoErrorCompleted()  CheckpointManager.java
          -> CheckpointCoordinator.isNoErrorCompleted()  CheckpointCoordinator.java:1393-1402
            -> runningJobStateIMap.get(checkpointStateImapKey)   // returns null when the entry was cleaned (e.g. CheckpointCoordinator.clean()/SubPlan cleanup ran before restore)
            -> CheckpointCoordinatorStatus.FINISHED.equals(status) || CheckpointCoordinatorStatus.SUSPEND.equals(status)  CheckpointCoordinator.java:1399-1400  // changed lines: null-safe, previously status.equals(...) threw NPE
            -> !latestCompletedCheckpoint.isRestored()  CheckpointCoordinator.java:1401

Guard note: line 1393-1394 already short-circuits when latestCompletedCheckpoint == null, so the only nullable operand in the expression was `status` — the constant-side flip closes exactly that hole. The false result then flows back so SubPlan takes the restore/re-execute branch instead of marking the pipeline finished.
```

# Findings
**Issue 1: Null coordinator state is now silently treated as 'not completed', which forces a restore/re-execution of a pipeline whose final checkpoint had actually completed — the symptom (NPE) is fixed but the underlying state-cleanup race is not addressed**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinator.java:1399`
- Why it matters: - Root cause: the NPE happened because `runningJobStateIMap.get(checkpointStateImapKey)` returned `null` for a coordinator whose `latestCompletedCheckpoint` is a final checkpoint, i.e. the coordinator state entry was cleaned while the completed-checkpoint record still existed. The fix flips to constant-side `equals`, so a `null` status maps to `isNoErrorCompleted() == false`.
- Impact: on the restore path, `false` means the pipeline is treated as not-yet-completed and will be restored and re-run. For a batch job that had genuinely reached FINISHED before its state entry was cleaned, this can re-execute the source and re-emit data into non-idempotent sinks (duplicate writes) instead of short-circuiting the restart. The old behavior was a hard failure; the new behavior is a silent re-run, which is arguably the more dangerous failure mode for exactly-once expectations on Zeta.
- The PR only patches the consumer of the stale state; nothing in the diff changes the ordering that allows `latestCompletedCheckpoint` to outlive the `checkpointStateImapKey` entry.
- Fix: keep the null-safe comparison, but audit the cleanup path so the coordinator status entry is not removed before the completed-checkpoint marker during recovery, or persist a terminal marker alongside `latestCompletedCheckpoint` so a cleaned state entry for a final, non-restored checkpoint is interpreted as completed rather than as 'needs restore'.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinator.java:1400  || CheckpointCoordinatorStatus.SUSPEND.equals(status))`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinator.java:1399                && (CheckpointCoordinatorStatus.FINISHED.equals(status)`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinator.java:1400                        || CheckpointCoordinatorStatus.SUSPEND.equals(status))`
- Suggested fix: The constant-side `equals` correctly removes the NPE, but note the semantic consequence: a `null` status now means `isNoErrorCompleted()` returns `false`, so a pipeline whose final checkpoint had actually FINISHED (state entry cleaned early) will be restored and re-executed instead of being short-circuited. For batch jobs with non-idempotent sinks this can produce duplicate output. Please also look at why the `checkpointStateImapKey` entry can be removed while `latestCompletedCheckpoint` still exists — fixing the cleanup ordering (or persisting a terminal marker with the completed checkpoint) would address the root cause rather than only the NPE symptom.
- Severity: Medium

**Issue 2: The regression test's expected value duplicates the production logic, making it tautological**
- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:951`
- Why it matters: - Root cause: the test fetches the private `runningJobStateIMap` field via reflection and immediately stubs it with `Mockito.when(runningJobStateIMap.get(Mockito.any())).thenReturn(status)`. This only works if `buildMinimalCoordinator` wires a Mockito mock into the coordinator; the coupling is implicit and undocumented in the new test.
- Impact: if `buildMinimalCoordinator` is later refactored to use a real Hazelcast test instance (a plausible change given other CheckpointCoordinatorTest scenarios use real maps), this test fails with `NotAMockException` at the stubbing line rather than a clear assertion failure, and repeated `when(...)` calls in a nested loop also accumulate stubbings on a shared mock across iterations.
- Fix: either assert/fail-fast that the field is a mock (`Assertions.assertTrue(MockUtil.isMock(runningJobStateIMap))` with a message), or avoid the mock dependency entirely by using a real map-backed stub and `put`/`remove` on `checkpointStateImapKey` per iteration, which also makes the null case (`remove`) explicit.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:952                                checkpointType.isFinalCheckpoint()`; `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:953                                        && (status == CheckpointCoordinatorStatus.FINISHED`; `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:955                                        && !restored;`
- Suggested fix: The `expected` value here re-implements the exact production expression, so the matrix asserts the implementation against itself. Could you add at least one explicit, hard-coded assertion for the #11563 regression — final non-restored checkpoint with a null IMap entry returns `false` without throwing — so the fixed contract is stated independently of the implementation?
- Severity: Medium

**Issue 3: Regression matrix covers null coordinator state but not a null latestCompletedCheckpoint during the same recovery window**
- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:947`
- Why it matters: - Root cause: the new matrix test always installs a non-null `CompletedCheckpoint` via `ReflectionUtils.setField(coordinator, "latestCompletedCheckpoint", completedCheckpoint)` before exercising `isNoErrorCompleted()`, and the null dimension is only applied to the IMap-returned status.
- Impact: the same restore race that cleans the coordinator state entry could plausibly leave `latestCompletedCheckpoint` unset on a freshly rebuilt coordinator; `isNoErrorCompleted()` still dereferences it (`latestCompletedCheckpoint.getCheckpointType()`), so an NPE of the same class as #11563 remains possible on that adjacent boundary and is untested.
- Fix: add one more case to `testIsNoErrorCompletedUsesCheckpointStateMatrix` (or a dedicated test) where `latestCompletedCheckpoint` is left/set to `null`, and decide the intended semantics — either guard the production method with `latestCompletedCheckpoint != null &&` or assert the invariant that callers only invoke it after a completed checkpoint exists. The test also correctly avoids Mockito stubbing leaks by re-stubbing `runningJobStateIMap.get` per iteration, which is fine as-is.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:947  coordinator, "latestCompletedCheckpoint", completedCheckpoint);`; `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:950  Mockito.when(runningJobStateIMap.get(Mockito.any())).thenReturn(status);`
- Suggested fix: Nice exhaustive matrix over CheckpointType × restored × status (including null status). One boundary is still uncovered: `latestCompletedCheckpoint` is always non-null here, yet `isNoErrorCompleted()` dereferences it (`latestCompletedCheckpoint.getCheckpointType()`). Since the same recovery race that cleans the state-map entry could leave this field unset on a rebuilt coordinator, please add a case where `latestCompletedCheckpoint` is null and pin down the expected behavior (likely `false` after adding a null guard in the production method).
- Severity: Low

**Issue 4: The state matrix hard-codes the `CheckpointCoordinatorStatus` values instead of iterating `values()`, so future enum additions silently fall out of coverage**
- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:925`
- Why it matters: - Root cause: the test enumerates `null, RUNNING, CANCELED, FAILED, FINISHED, SUSPEND` in a hand-written array, while `CheckpointType` on the other axis is correctly iterated via `CheckpointType.values()`.
- Impact: if a new `CheckpointCoordinatorStatus` constant is ever introduced (e.g., a paused/draining state on Zeta), this matrix will not exercise it and a regression in `isNoErrorCompleted()` for that state would go undetected, defeating the purpose of an exhaustive matrix test.
- Fix: build the array as `CheckpointCoordinatorStatus.values()` plus a prepended `null` entry (e.g., via a `Stream.concat` or a small helper), matching the exhaustive-iteration pattern already used for `CheckpointType.values()` in the same test.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:926  null,`; `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:933  for (CheckpointType checkpointType : CheckpointType.values()) {`
- Suggested fix: You iterate `CheckpointType.values()` exhaustively below but hand-enumerate the coordinator statuses here. Please derive this array from `CheckpointCoordinatorStatus.values()` with `null` prepended, so any future enum constant is automatically covered by the matrix instead of silently escaping it.
- Severity: Low

**Issue 5: New regression test lacks the Javadoc that sibling tests in this file carry, and does not reference issue #11563**
- Location: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:913`
- Why it matters: - Root cause: the test right below the insertion point in this same file documents its intent with a `/** Regression: ... */` Javadoc block, but `testIsNoErrorCompletedUsesCheckpointStateMatrix` is added without any documentation.
- Impact: readers cannot tell from the code which production bug this matrix guards against, which weakens its value as a regression marker and diverges from the established convention in `CheckpointCoordinatorTest`.
- Fix: add a Javadoc such as `/** Regression for #11563: a cleaned (null) coordinator state entry in runningJobStateIMap must make isNoErrorCompleted() return false instead of throwing NPE during Worker recovery. */`.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/checkpoint/CheckpointCoordinatorTest.java:913    void testIsNoErrorCompletedUsesCheckpointStateMatrix() {`
- Suggested fix: Could you add a Javadoc here in the same style as the neighboring regression tests in this file (e.g. `/** Regression for #11563: a null coordinator state entry must not fail isNoErrorCompleted() during Worker recovery. */`)? Without it, future readers won't know which failure mode this matrix protects against.
- Severity: Low

# Review conclusion

- Issue 1: Null coordinator state is now silently treated as 'not completed', which forces a restore/re-execution of a pipeline whose final checkpoint had actually completed — the symptom (NPE) is fixed but the underlying state-cleanup race is not addressed
- Issue 2: The regression test's expected value duplicates the production logic, making it tautological
- Issue 3: Regression matrix covers null coordinator state but not a null latestCompletedCheckpoint during the same recovery window
- Issue 4: The state matrix hard-codes the `CheckpointCoordinatorStatus` values instead of iterating `values()`, so future enum additions silently fall out of coverage
- Issue 5: New regression test lacks the Javadoc that sibling tests in this file carry, and does not reference issue #11563

PR #11571, opened 2026-07-27
[Feature][Connector-File-Base] Add optional PDF RAG metadata for file source

JeremyXin:

This PR adds optional RAG metadata support for PDF file source reading.

  Changes include:

  - add `pdf_rag_metadata_enabled` config for PDF sources
  - append stable RAG metadata columns in `PdfReadStrategy`: `source_uri`, `document_id`, `chunk_id`, `chunk_index`, `content_hash`
  - normalize file: URIs to keep metadata stable across equivalent local paths
  - make PDF empty-path schema inference consistent with Markdown/Binary behavior
  - add focused tests for PDF RAG metadata schema, stability, URI normalization, and empty-directory schema discovery

  This change is backward compatible: the new metadata columns are only appended when `pdf_rag_metadata_enabled`=true.

### Check list


DanielLeens (changes requested):
Thanks for the contribution! I went through the full diff and the source-side approach looks good to me. Adding optional PDF RAG metadata follows the existing Markdown metadata pattern nicely, and the default remains `false`, so existing PDF output keeps the old 8-column schema unless users explicitly opt in.

The runtime path I checked is:

```text
User config: file_format_type=pdf, pdf_rag_metadata_enabled=true
  -> Local/S3/HDFS/OSS/FTP/SFTP/COS/OBS/Jindo FileSourceFactory.optionRule()
      -> PDF_RAG_METADATA_ENABLED is accepted only for PDF
  -> BaseFileSource(ReadonlyConfig)
      -> FileFormat.PDF.getReadStrategy()
      -> PdfReadStrategy.setPluginConfig(...)
      -> getSeaTunnelRowTypeInfo(...)
          -> returns 8 default document-element fields + 5 RAG metadata fields
  -> SourceReader reads each split
      -> PdfReadStrategy.read(path, tableId, collector)
          -> extracts heading / paragraph / image / link elements
          -> normalizeSourceUri(path)
          -> buildDocumentId(sourceUri)
          -> toSeaTunnelRow(..., appendRagMetadata=true)
          -> emits a 13-field SeaTunnelRow
```

## Findings

Issue 1: the new user-facing PDF metadata option is not documented yet
- Location: `docs/en/connectors/source/LocalFile.md:222`, `docs/zh/connectors/source/LocalFile.md:222`
- Description: `pdf_rag_metadata_enabled` is now exposed through the file source option rules for PDF, but the English and Chinese docs still only describe `markdown_rag_metadata_enabled`. The PDF sections currently say that PDF uses the same document-element schema, but they do not mention that enabling this new option appends `source_uri`, `document_id`, `chunk_id`, `chunk_index`, and `content_hash`, nor that the default is `false`.
- Risk: this is a user-visible config and output-schema extension. Without docs, users cannot discover the option reliably and may be surprised by the 8-column vs 13-column schema behavior.
- Suggestion: please update both `docs/en` and `docs/zh` for the file sources that support PDF, at least LocalFile, HdfsFile, S3File, OssFile, OssJindoFile, CosFile, ObsFile, FtpFile, and SftpFile, with the option name, default value, appended fields, and opt-in behavior.
- Severity: Medium
- Raised by others: No

This is compatible. It does not change public APIs, protocols, serialization formats, or the default PDF schema. When enabled, each emitted PDF element computes SHA-256 metadata and carries five extra fields, which is expected for the RAG use case.

## Test coverage and stability

The test coverage is solid: it covers enabled metadata schema/arity, stable IDs and hashes, file URI normalization, empty-directory schema discovery, and temp-file cleanup. Since this PR includes UT changes, I also checked test stability. Stability rating: **Stable**. The tests use fixed PDF resources or locally-created temp PDFs with cleanup, and I did not see hard sleeps, external ports, network dependencies, or order-sensitive async assertions.

## Merge conclusion

1. Blocking items
- Issue 1: please document `pdf_rag_metadata_enabled` in the English and Chinese file source docs before merge.

2. Non-blocking suggestions
- None from the source-level review.

Overall, the implementation looks good and the current CI is green. Once the docs are updated, I think this should be in good shape.

DanielLeens (commented):
Thanks for the update! I re-reviewed the latest head `d06888f0ccf00f1cb3c9f4cbab6d1adf44ec3379` from scratch across the full file source path, the PDF reader, empty-path schema discovery, all affected FileSourceFactory option rules, tests, docs, and the current CI/merge gates.

Good news: the source/doc blocker from my previous round is closed for me now. The new `pdf_rag_metadata_enabled` option is documented in both English and Chinese file source docs, and I do not see a new code-side blocker in the current source diff. The remaining gate is CI / branch freshness.

## What This PR Fixes

- User pain point: PDF file source currently emits only the base document-element fields, so RAG users need to add source URI, document id, chunk id, chunk index, and content hash downstream.
- Fix approach: add an opt-in `pdf_rag_metadata_enabled` option. When `file_format_type=pdf` and this option is `true`, PDF rows append `source_uri`, `document_id`, `chunk_id`, `chunk_index`, and `content_hash`.
- One-line summary: this gives PDF source the same kind of optional RAG metadata support as Markdown while preserving the old default schema.

## Full Runtime Path

```text
User config
  -> file_format_type = pdf
  -> pdf_rag_metadata_enabled = true

Factory phase
  -> Local/S3/HDFS/OSS/Jindo/COS/OBS/FTP/SFTP FileSourceFactory.optionRule()
      -> PDF_RAG_METADATA_ENABLED is accepted only when FILE_FORMAT_TYPE == PDF

Source initialization
  -> BaseFileSource(ReadonlyConfig)
      -> fileFormat = PDF
      -> readStrategy = FileFormat.PDF.getReadStrategy()
      -> readStrategy.setPluginConfig(pluginConfig.toConfig())
          -> PdfReadStrategy.pdfRagMetadataEnabled = true [PdfReadStrategy.java L189-L195]
      -> readStrategy.init(hadoopConf)
      -> readStrategy.getFileNamesByPath(path)
      -> if files exist:
          -> readStrategy.getSeaTunnelRowTypeInfo(firstFile) [BaseFileSource.java L116-L119]
      -> if the path is empty:
          -> buildCatalogTableForEmptyPath(path, PDF) [BaseFileSource.java L132-L141]
      -> produced CatalogTable uses the 13-field PDF RAG schema

Runtime read
  -> BaseFileSourceReader reads a FileSourceSplit
      -> PdfReadStrategy.read(path, tableId, collector) [PdfReadStrategy.java L127-L168]
          -> hadoopFileSystemProxy.getInputStream(path)
          -> copy stream to a temporary PDF file
          -> Loader.loadPDF(tempPdf)
          -> extractPdfDocumentElements(document)
          -> sourceUri = normalizeSourceUri(path) [PdfReadStrategy.java L1006-L1015]
          -> documentId = buildDocumentId(sourceUri) [PdfReadStrategy.java L1002-L1004]
          -> toSeaTunnelRow(..., appendRagMetadata=true) [PdfReadStrategy.java L972-L984]
          -> append source_uri/document_id/chunk_id/chunk_index/content_hash [PdfReadStrategy.java L986-L999]
          -> collector.collect(13-field row)
      -> finally deleteTempPdfPath(tempPdfPath)
```

## Findings

No new source-level blockers found in the current head.

The previous documentation blocker is addressed:
- `docs/en/connectors/source/{LocalFile,HdfsFile,S3File,OssFile,OssJindoFile,CosFile,ObsFile,FtpFile,SftpFile}.md` now mention `pdf_rag_metadata_enabled`.
- The matching `docs/zh` pages also include the same option and appended fields.

## Compatibility

This is fully compatible:
- API / SPI: unchanged.
- Config: one new optional option, default `false`.
- Defaults: unchanged; PDF keeps the old 8-field schema unless users opt in.
- Protocol / serialization / checkpoint state: unchanged.
- Historical behavior: existing jobs are not affected.

## Side Effects

When the option is enabled, each emitted PDF element computes SHA-256 metadata and carries five extra string/int fields. That is expected for the RAG use case. CPU/GC overhead is opt-in only; there is no new network, locking, retry, or resource-lifecycle behavior in production runtime. Temporary PDF files are still cleaned in `finally`.

## Tests

The new tests cover the important paths:
- `PdfReadStrategyTest.java L141-L172`: enabled metadata schema, row arity, stable document/chunk/hash values.
- `PdfReadStrategyTest.java L174-L196`: `file:` URI normalization.
- `BaseFileSourceTest.java L94-L113`: empty-directory PDF schema discovery with and without RAG metadata.

Test stability rating: **Stable**. The added tests use fixed PDF resources or locally-created temporary PDFs, clean up temp files, and do not rely on sleeps, network ports, containers, wall-clock timing, or async ordering.

I did not run local Maven/UT/E2E tests for this review; this was a static full-diff and call-chain review plus GitHub CI/merge metadata inspection. `git diff --check upstream/dev...HEAD` produced no whitespace errors.

## CI / Merge Gate

The current required Build check is failing:

`https://github.com/apache/seatunnel/runs/91152417822`

I could not retrieve actionable public failed-job logs from GitHub for this run (`log not found` from the log endpoint), so I do not want to guess a code-side cause. The branch is also currently behind the latest `dev` (`compare status: diverged`, `behind_by=7`). Since the previous source/doc blocker is fixed and I do not see a new source blocker in this diff, the lowest-cost next step is to sync with the latest `dev` and rerun Build first. If it still fails after the sync, please share the new failing check link and I can help narrow it down against the updated head.

Note: my earlier `changes requested` entry may still remain visible in GitHub's review summary because I do not have write-level review controls on this repository. Technically, from Daniel's side, the source-level blocker is cleared; a write-capable maintainer may still need to issue the final approval/dismissal step once CI is green.


### Conclusion: Mergeable After Fixes

1. Blocking Items (Must Be Fixed)
- No new source-level blocker from Daniel's side; the previous docs blocker is fixed.
- The required Build check still needs to pass.
- Please sync with the latest `dev` and rerun Build first, because this branch is `diverged` with `behind_by=7` and the current failed log is not publicly actionable from the check endpoint.


Overall, this looks much healthier now. The implementation is compatible, the tests cover the key schema/metadata paths, and the docs are in place. Once the branch is synced and Build is green, I think this should be in good shape.

DanielLeens (changes requested):
Thanks for the update. I re-reviewed the latest head `933d1389bd4249f93ebd3874c1b8c21ae6d657e3` from scratch across the PDF source path, the new RAG metadata schema/row expansion, the empty-path schema handling, the updated EN/ZH docs, and the extra test-side delta that landed with the latest sync.

The good news is that the previous Daniel-side docs blocker is still closed for me: the `pdf_rag_metadata_enabled` option is documented in both English and Chinese, the PDF row/schema path remains compatible by default, and I do not see a new blocker in the actual PDF source logic.

I found one new blocker in the latest head, though.

## Runtime path I checked

```text
PDF source with RAG metadata enabled
  -> FileSourceFactory accepts pdf_rag_metadata_enabled only for PDF
  -> BaseFileSource chooses PdfReadStrategy
  -> PdfReadStrategy appends source_uri / document_id / chunk_id / chunk_index / content_hash
  -> BaseFileSource empty-path schema path now returns the matching PDF schema
```

## Issue 1: the current PR still carries an unrelated `ImportClassCheckTest` refactor

- Location: `seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java:53`
- Severity: High

The latest head still changes the repository-wide `ImportClassCheckTest` scanner: it introduces `ImportMetadata`, skips `target` paths, adds protobuf-generated-source detection, and rewires how import line numbers are collected. That is not part of the PDF RAG metadata feature or its regression coverage.

This is not just stale-base noise anymore. The branch is now directly ahead of the current `dev` (`compare status: ahead`, `behind_by=0`), so the import-scan refactor is part of the active PR diff that would merge with this feature. Mixing a repository-wide CI/tooling refactor into a focused file-source feature makes review ownership and rollback much harder than it needs to be.

Please drop the `ImportClassCheckTest.java` change from this PR and keep only the PDF RAG metadata feature plus its docs/tests. If that import-scan improvement is still needed, it should go in a separate dedicated PR so we can review the CI/tooling behavior independently.

## Merge conclusion

### Conclusion: can merge after fix

1. Blocking items
- Issue 1: remove the unrelated `ImportClassCheckTest` refactor from this focused PDF source feature PR.

2. Non-blocking notes
- The actual PDF RAG metadata source change, tests, and EN/ZH docs look good to me on the current head.
- The current required `Build` is queued on this head (`https://github.com/apache/seatunnel/runs/91477089127`), so after the unrelated import-scan change is dropped, please let the queued Build finish on the cleaned-up revision.

Once the unrelated import-scan delta is split out, I am happy to re-review the remaining diff quickly.

JeremyXin:
@DanielLeens  Thanks for the careful review.

  I double-checked the current PR base and diff. The ImportClassCheckTest changes came from the latest `dev` sync, but they are not
  part of the current PR diff anymore.

  Current PR base:
  - `apache:dev` at `458cefdc5`

  Current PR head:
  - `933d1389b`

  I also rechecked the current `Files changed` set and `ImportClassCheckTest.java` is not included there. The effective diff only
  contains the PDF RAG metadata feature, related tests, and docs.

  So there is nothing additional to drop from the active PR diff on the current base. If you were looking at the merge history or
  an earlier compare view, that may explain the mismatch.

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `933d1389bd42` from scratch.

# What this PR fixes
- User pain: PDF file sources parsed for RAG pipelines lose document/chunk lineage — users had stable RAG metadata (source_uri, document_id, chunk_id, chunk_index, content_hash) for Markdown via `markdown_rag_metadata_enabled`, but no equivalent for PDF, forcing downstream dedup/traceability hacks; PDF empty-path schema inference also diverged from Markdown/Binary behavior.
- Fix approach: Mirror the Markdown design — add a `pdf_rag_metadata_enabled` option (default false), have `PdfReadStrategy` conditionally append the five RAG metadata columns after `child_ids` with `file:` URI normalization for stable identities, align PDF empty-path schema discovery with Markdown/Binary, and update all 18 EN/ZH connector docs plus focused unit tests.
- One-line summary: Functionally sound and backward compatible (columns only appended when opted in), but the new option is missing from every connector Options table and example config, the empty-path schema-inference behavior change is undocumented, URI-identity stability only holds for local `file:` paths, verbatim `source_uri` can leak URI-embedded credentials to sinks, and there is no E2E coverage — these should be addressed before merge.

# Runtime chain I rechecked
```text
Schema inference path:
  BaseMultipleTableFileSource / source config -> ReadStrategy.getSeaTunnelRowTypeInfo(path)
    -> PdfReadStrategy.getSeaTunnelRowTypeInfo()  PdfReadStrategy.java:~90
      -> reads pdf_rag_metadata_enabled from pluginConfig (BaseSourceConfigOptions.PDF_RAG_METADATA_ENABLED)
      -> when enabled, appends source_uri/document_id/chunk_id/chunk_index/content_hash to the document-element row type
      -> empty-path case now falls back to the default schema (matching MarkdownReadStrategy/BinaryReadStrategy) instead of failing

Read path:
  BaseFileSourceReader.pollNext() -> ReadStrategy.read(split.path, tableId, output)
    -> PdfReadStrategy.read()  PdfReadStrategy.java:~130
      -> parses PDF into document elements (per-element rows)
      -> normalizeSourceUri(path)  PdfReadStrategy.java:~200  (canonicalizes file: URIs so equivalent local paths yield identical source_uri)
      -> document_id = hash(normalized source_uri); chunk_index = element order; content_hash = hash(element content); chunk_id = hash(document_id + chunk_index + content_hash)
      -> appends the five metadata values after child_ids only when pdf_rag_metadata_enabled=true, so the default row layout is unchanged
```

# Findings
**Issue 1: `source_uri` RAG metadata is emitted verbatim into every output row and can leak embedded URI credentials (userinfo) and internal path topology to downstream sinks**
- Location: `docs/en/connectors/source/FtpFile.md:303`
- Why it matters: - Root cause: per the documented behavior added in this PR, when `pdf_rag_metadata_enabled` is true the connector appends `source_uri` ("Source file path or URI") to every produced row. The PR description states `PdfReadStrategy` normalizes `file:` URIs but does not mention stripping URI userinfo or query components for non-file schemes.
- Attack/leak scenario: several supported filesystems accept credentials embedded in the URI — e.g. `ftp://user:password@host/path`, `sftp://user:pass@host/...`, and the legacy Hadoop `s3a://ACCESSKEY:SECRETKEY@bucket/key` form. If a job is configured with such a path, the secret is replicated into `source_uri` on every row and lands in whatever sink the pipeline writes to (databases, vector stores, Kafka topics), plus any sink-side logging — a much wider blast radius than the job config file itself.
- Even without embedded credentials, exposing absolute local/HDFS paths to downstream consumers discloses internal cluster layout.
- Fix: in `PdfReadStrategy`, when building `source_uri`, parse the path with `java.net.URI` and rebuild it with `userInfo`, `query`, and `fragment` stripped before emitting or hashing; keep `document_id` derived from the sanitized URI so identifiers stay stable. Apply the same sanitization in the existing Markdown RAG metadata path (`markdown_rag_metadata_enabled`) if it shares this behavior, and add a unit test with a userinfo-bearing URI to `PdfReadStrategyTest`.
- Evidence: `docs/en/connectors/source/S3File.md:252  When either `markdown_rag_metadata_enabled` or `pdf_rag_metadata_enabled` is set to `true`, SeaTunnel appends the following RAG metadata fields after `child_ids` for the corresponding file type:`
- Suggested fix: Since `source_uri` is appended verbatim to every output row, please make sure `PdfReadStrategy` strips URI userinfo/query before emitting it. FTP/SFTP paths (`ftp://user:pass@host/...`) and legacy `s3a://KEY:SECRET@bucket/...` forms can embed credentials, and this feature would replicate the secret into every row delivered to the downstream sink — a much larger exposure surface than the job config. Sanitizing the URI (rebuild via `java.net.URI` without userInfo/query/fragment) before emitting and before deriving `document_id` keeps identifiers stable while preventing the leak.
- Severity: Medium

**Issue 2: The PDF empty-path schema-inference behavior change is not documented anywhere in the updated docs or upgrade notes**
- Location: `docs/en/connectors/source/HdfsFile.md:143`
- Why it matters: - Root cause: the PR description states it "make[s] PDF empty-path schema inference consistent with Markdown/Binary behavior" (backed by the new empty-directory test in `BaseFileSourceTest.java`), but none of the 18 doc files mention this behavior, and there is no upgrade/changelog note in the PR.
- Impact: this is a user-visible behavior change independent of the new option — jobs that previously failed (or behaved differently) when the PDF source path matched no files will now silently infer the document-element schema and proceed. On Zeta/Flink/Spark this can turn a fast-fail misconfiguration into a job that runs with zero rows, which surprises operators.
- Fix: add one sentence to the PDF section of the connector docs (EN + ZH) describing the empty-path behavior (schema is still inferred, consistent with Markdown/Binary), and mention this behavior change in the PR's user-facing-change section / release notes so it is captured for the next release.
- Evidence: `docs/en/connectors/source/OssFile.md:257  For PDF input, enable `pdf_rag_metadata_enabled` to append the RAG metadata fields described above.`
- Suggested fix: The PR also changes PDF empty-path schema inference to match Markdown/Binary behavior, but this doc section doesn't mention it. Please add a sentence here (and in the ZH doc) noting that when the path matches no files, the document-element schema is still inferred — this is a user-visible behavior change and should also be called out in the release notes.
- Severity: Medium

**Issue 3: New user-facing option `pdf_rag_metadata_enabled` is only described in prose; it is not added to the Options table of any of the 18 updated connector docs**
- Location: `docs/en/connectors/source/LocalFile.md:230`
- Why it matters: - Root cause: every doc file in this PR received only a +2/-1 prose change (rewording the RAG metadata paragraph and adding one sentence in the PDF section). None of the diffs touch the "Options" table where each connector option must be listed with name, type, required, and default value per our connector doc specification (https://github.com/apache/seatunnel/issues/4544).
- Impact: users scanning the option tables of LocalFile/S3File/HdfsFile etc. will never discover `pdf_rag_metadata_enabled`, its type (boolean), or that it defaults to `false`. The existing `markdown_rag_metadata_enabled` sets the precedent that RAG options belong in the table.
- Fix: add a `pdf_rag_metadata_enabled | boolean | no | false | ...` row to the option table of all 9 EN and 9 ZH source docs, right next to `markdown_rag_metadata_enabled`, and explicitly state the default is `false` so the backward-compatibility guarantee from the PR description is documented.
- Evidence: `docs/en/connectors/source/S3File.md:267  For PDF input, enable `pdf_rag_metadata_enabled` to append the RAG metadata fields described above.`; `docs/zh/connectors/source/LocalFile.md:230  对于 PDF 输入，启用 `pdf_rag_metadata_enabled` 即可追加上文所述的 RAG 元数据字段。`
- Suggested fix: This sentence is the only place `pdf_rag_metadata_enabled` appears in this doc. Per our connector doc specification (#4544), new options must also be listed in the Options table with type/required/default (`boolean | no | false`), next to `markdown_rag_metadata_enabled`. Please add the table row here and in the other 17 EN/ZH docs so users can discover the option and its default.
- Severity: Medium

**Issue 4: No example configuration demonstrating `pdf_rag_metadata_enabled` was added to any connector doc**
- Location: `docs/en/connectors/source/CosFile.md:216`
- Why it matters: - Root cause: the doc changes only reword the shared RAG-metadata paragraph and append a single sentence to the PDF section; no HOCON example showing a PDF source with `pdf_rag_metadata_enabled = true` and the resulting extra columns (`source_uri`, `document_id`, `chunk_id`, `chunk_index`, `content_hash`) was added.
- Impact: for a schema-affecting option like this, users need a concrete example to understand where the option goes (inside the source block) and what the output row looks like; without it, adoption of the feature depends on reading the connector source code.
- Fix: add a short example under the PDF section of at least LocalFile.md (EN + ZH), e.g. a `LocalFile` source with `file_format_type = "pdf"` and `pdf_rag_metadata_enabled = true`, plus a note showing the appended metadata columns, mirroring how the Markdown RAG feature is documented.
- Evidence: `docs/en/connectors/source/FtpFile.md:318  For PDF input, enable `pdf_rag_metadata_enabled` to append the RAG metadata fields described above.`
- Suggested fix: Since `pdf_rag_metadata_enabled` changes the produced schema (5 appended columns), could we add a small HOCON example here (source block with `file_format_type = "pdf"` and `pdf_rag_metadata_enabled = true`, plus the resulting extra fields)? A prose mention alone makes it hard for users to know where to place the option and what output to expect.
- Severity: Medium

**Issue 5: `document_id` stability guarantee only holds for `file:` URIs; equivalent remote URIs (trailing slashes, host aliases, scheme variants) will yield divergent document identities**
- Location: `docs/en/connectors/source/HdfsFile.md:128`
- Why it matters: - Root cause: the PR normalizes only `file:` URIs ("normalize file: URIs to keep metadata stable across equivalent local paths" per the description), while the docs promise a "Stable document identifier derived from `source_uri`" for all nine connectors including HDFS/S3/OSS/FTP.
- Impact: for remote filesystems, superficially different but equivalent URIs — `hdfs://nn:8020//data/a.pdf` vs `hdfs://nn:8020/data/a.pdf`, `s3a://bucket/dir/../dir/a.pdf`, or the same NameNode referenced by hostname vs nameservice — produce different `document_id`/`chunk_id` values. In a RAG pipeline this causes duplicate document entries in the vector store and broken idempotent upserts when the job config path is later rewritten to an equivalent form, exactly the instability this feature aims to prevent.
- Fix: extend the normalization in `PdfReadStrategy` to be scheme-agnostic where safe (collapse duplicate slashes, resolve `.`/`..` segments via `URI.normalize()`, drop trailing slashes) rather than special-casing `file:`, and document in the connector docs that `document_id` stability is scoped to the exact configured URI form for remote schemes. Add test cases for `hdfs://`/`s3a://` URI variants alongside the existing local-path normalization tests in `PdfReadStrategyTest`.
- Evidence: `docs/en/connectors/source/HdfsFile.md:128  When either `markdown_rag_metadata_enabled` or `pdf_rag_metadata_enabled` is set to `true`, SeaTunnel appends the following RAG metadata fields after `child_ids` for the corresponding file type:`
- Suggested fix: The docs promise a "stable document identifier derived from `source_uri`", but the normalization in this PR only covers `file:` URIs. For HDFS/S3/OSS, equivalent URIs (double slashes, `..` segments, trailing slash, nameservice vs hostname) will produce different `document_id`/`chunk_id` values, causing duplicate documents in downstream RAG stores when the configured path form changes. Consider applying scheme-agnostic normalization (`URI.normalize()`, slash collapsing) in `PdfReadStrategy` and adding remote-scheme cases to `PdfReadStrategyTest`, or explicitly scoping the stability guarantee to local paths in these docs.
- Severity: Low

**Issue 6: Unit tests were added, but no E2E coverage exercises `pdf_rag_metadata_enabled` end-to-end on a real engine**
- Location: `docs/en/connectors/source/S3File.md:252`
- Why it matters: - Root cause: the changed-files list includes `PdfReadStrategyTest.java` and additions to `BaseFileSourceTest.java`, which cover schema, stability, and URI normalization at the strategy level, but no `seatunnel-e2e` config or IT case was added for the new option.
- Impact: the option is wired through nine connector source factories (Cos/Ftp/Hdfs/JindoOss/Local/Obs/...), and factory option registration plus config plumbing through `BaseFileSourceConfig` is exactly the layer that unit tests of `PdfReadStrategy` do not exercise; regressions in option lookup or row projection on Zeta/Flink/Spark would go unnoticed.
- Fix: add an e2e job config under `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-file-local-e2e` reading a small bundled PDF with `pdf_rag_metadata_enabled = true` and asserting the five appended metadata columns, similar to how other file-format features are covered.
- Evidence: `docs/en/connectors/source/SftpFile.md:286  For PDF input, enable `pdf_rag_metadata_enabled` to append the RAG metadata fields described above.`
- Suggested fix: The new option is documented across nine connectors and wired through their source factories, but I don't see an E2E case for it. Could we add a job config in `connector-file-local-e2e` reading a small PDF with `pdf_rag_metadata_enabled = true` and asserting the five appended metadata columns? The strategy-level unit tests don't cover the factory/option plumbing on a real engine.
- Severity: Low

# Review conclusion

- Issue 1: `source_uri` RAG metadata is emitted verbatim into every output row and can leak embedded URI credentials (userinfo) and internal path topology to downstream sinks
- Issue 2: The PDF empty-path schema-inference behavior change is not documented anywhere in the updated docs or upgrade notes
- Issue 3: New user-facing option `pdf_rag_metadata_enabled` is only described in prose; it is not added to the Options table of any of the 18 updated connector docs
- Issue 4: No example configuration demonstrating `pdf_rag_metadata_enabled` was added to any connector doc
- Issue 5: `document_id` stability guarantee only holds for `file:` URIs; equivalent remote URIs (trailing slashes, host aliases, scheme variants) will yield divergent document identities
- Issue 6: Unit tests were added, but no E2E coverage exercises `pdf_rag_metadata_enabled` end-to-end on a real engine

DanielLeens:
Thanks for the additional current-head review.

From Daniel's side, the previous documentation blocker is still closed for me, but the current unchanged head still needs a source update before another full Daniel re-review. In particular, I still want the unrelated `ImportClassCheckTest` refactor removed from this PR so the branch stays focused on the PDF RAG metadata change itself.

Once that code update is pushed, I am happy to re-review the refreshed head from scratch.

davidzollo (commented):
+1
LGTM

```

```

davidzollo (approved):

PR #11585, opened 2026-07-28
[Fix][Connector-V2] Wait for Kudu client shutdown

### LINKED ISSUE #11584: [Bug] [Connector-V2] Kudu client threads outlive the Flink user-code classloader
### Search before asking

- [x] I had searched in the issues and found no similar issues.

### What happened

The Kudu connector can leave its Netty event-loop threads running after a Flink job has finished and the user-code classloader has been released.

Kudu 1.15 calls `shutdownGracefully()` when `KuduClient.close()` runs, but the close call returns before the event-loop executor has fully terminated. Flink can then unload the connector JAR while the remaining Kudu threads are still completing channel and promise cleanup.

This produces repeated `NoClassDefFoundError` exceptions from `kudu-nio-*` threads. In the Kudu E2E suite, which starts many Flink jobs sequentially, the leaked asynchronous cleanup also correlates with intermittent JobManager heartbeat and blob-download timeouts. A failed run often passes after rerunning.

The source reader also creates a `KuduScanner` for every split without explicitly closing it.

Expected behavior:

- Kudu scanners are closed after each split is consumed, including exceptional paths.
- Kudu client executors have fully terminated before the engine releases the connector classloader.
- The split enumerator releases its Kudu client after scan tokens have been generated instead of retaining it until job teardown.

### SeaTunnel Version

Current `dev` branch (`3.0.0-SNAPSHOT`).

### SeaTunnel Config

```conf
env {
  parallelism = 1
  job.mode = "BATCH"
}

source {
  Kudu {
    kudu_masters = "kudu-master:7051"
    table_name = ".*"
    use_regex = true
  }
}

sink {
  Assert {
    rules {
      table-names = ["kudu_source_table_1", "kudu_source_table_2"]
    }
  }
}
```

### Running Command

```shell
TEST_IN_PR=true RUN_ALL_CONTAINER=true ./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-kudu-e2e -DskipUT -DskipIT=false verify
```

### Error Exception

```log
Exception in thread "kudu-nio-3" java.lang.NoClassDefFoundError:
org/apache/kudu/shaded/io/netty/util/concurrent/DefaultPromise$1

java.lang.NoClassDefFoundError:
org/apache/kudu/shaded/io/netty/handler/ssl/SslCloseCompletionEvent
```

In flaky runs, the following failure can occur in the next Flink container:

```log
The heartbeat of JobManager timed out.
RegistrationTimeoutException: Could not register at the ResourceManager within 300000 ms.
```

Observed with Flink 1.13.6 and Flink 1.15.3 in the connector E2E suite.

### Java or Scala Version

Observed in both Java 8 and Java 11 E2E jobs.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

nzw921rx:

Closes #11584.

The Kudu 1.15 client starts a graceful Netty shutdown from `KuduClient.close()`, but the call can return before its event-loop executor has terminated. Flink may then release the connector user-code classloader while Kudu threads are still completing channel cleanup, causing repeated `NoClassDefFoundError` exceptions and intermittent E2E instability.

### What changed

- Manage the executor used by each internally created Kudu client.
- Wait for the executor to terminate after closing the client, with a bounded forced-shutdown fallback.
- Release the source enumerator client immediately after scan-token discovery.
- Close each `KuduScanner` after its split is consumed, including exceptional paths.
- Apply the managed lifecycle consistently to source, sink, and catalog clients.
- Preserve the existing `KuduUtil.getKuduClient` method for compatibility.
- Add unit coverage for graceful shutdown, forced shutdown, and client-close failures.

### Verification

```shell
./mvnw -pl seatunnel-connectors-v2/connector-kudu spotless:apply -DskipIT test
```

Result: 9 tests passed.

The full `mvnw verify` command was not run.

### DIFF
--- a/seatunnel-connectors-v2/connector-kudu/src/main/java/org/apache/seatunnel/connectors/seatunnel/kudu/catalog/KuduCatalog.java
+++ b/seatunnel-connectors-v2/connector-kudu/src/main/java/org/apache/seatunnel/connectors/seatunnel/kudu/catalog/KuduCatalog.java
@@ -34,6 +34,7 @@ import org.apache.seatunnel.api.table.catalog.exception.TableNotExistException;
 import org.apache.seatunnel.api.table.type.SeaTunnelDataType;
 import org.apache.seatunnel.connectors.seatunnel.kudu.config.CommonConfig;
 import org.apache.seatunnel.connectors.seatunnel.kudu.config.KuduBaseOptions;
+import org.apache.seatunnel.connectors.seatunnel.kudu.kuduclient.KuduClientResource;
 import org.apache.seatunnel.connectors.seatunnel.kudu.kuduclient.KuduTypeMapper;
 import org.apache.seatunnel.connectors.seatunnel.kudu.util.KuduUtil;
 
@@ -59,6 +60,7 @@ public class KuduCatalog implements Catalog {
     private final CommonConfig config;
 
     private KuduClient kuduClient;
+    private KuduClientResource kuduClientResource;
 
     private final String defaultDatabase = "default_database";
 
@@ -71,15 +73,23 @@ public class KuduCatalog implements Catalog {
 
     @Override
     public void open() throws CatalogException {
-        kuduClient = KuduUtil.getKuduClient(config);
+        kuduClientResource = KuduUtil.getKuduClientResource(config);
+        kuduClient = kuduClientResource.getClient();
     }
 
     @Override
     public void close() throws CatalogException {
         try {
-            kuduClient.close();
+            if (kuduClientResource != null) {
+                kuduClientResource.close();
+            } else if (kuduClient != null) {
+                kuduClient.close();
+            }
         } catch (KuduException e) {
             throw new CatalogException("Failed close kudu client", e);
+        } finally {
+            kuduClient = null;
+            kuduClientResource = null;
         }
     }
 
--- a/seatunnel-connectors-v2/connector-kudu/src/main/java/org/apache/seatunnel/connectors/seatunnel/kudu/kuduclient/KuduClientResource.java
+++ b/seatunnel-connectors-v2/connector-kudu/src/main/java/org/apache/seatunnel/connectors/seatunnel/kudu/kuduclient/KuduClientResource.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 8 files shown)


DanielLeens (approved):
Thanks for the contribution! I went through the full Kudu client lifecycle path, and this looks like a solid fix. The resource ownership is much clearer now, and I do not see a code-side blocker from my side.

## Runtime path I checked

```text
Kudu Source enumerator
  -> open()
      -> KuduInputFormat.openInputFormat()
          -> KuduUtil.getKuduClientResource()
  -> run()
      -> createInputSplits()
          -> KuduUtil.getKuduScanToken(kuduClient, ...)
      -> assignSplit(...)
      -> finally KuduInputFormat.closeInputFormat()
          -> KuduClientResource.close()
              -> KuduClient.close()
              -> executor.shutdown / awaitTermination / shutdownNow fallback

Kudu Source reader
  -> open()
      -> KuduInputFormat.openInputFormat()
  -> pollNext()
      -> KuduInputFormat.scanner(split.token)
      -> read rows
      -> finally KuduScanner.close()
  -> close()
      -> KuduInputFormat.closeInputFormat()

Kudu Sink writer
  -> new KuduOutputFormat()
      -> getKuduClientResource()
      -> newSession / openTable / create serializer
      -> initialization failure closes the resource
  -> close()
      -> flush()
      -> session.close()
      -> KuduClientResource.close()

Kudu Catalog
  -> open() getKuduClientResource()
  -> close() KuduClientResource.close()
```

## Review result

I did not find any blocking issues.

The key points I checked:
- `KuduClientResource.close()` waits for the executor even if `KuduClient.close()` throws, because the wait happens in `finally`.
- `KuduUtil.getKuduClientResource()` shuts down the executor if client creation fails before the resource is returned.
- The source enumerator can release its client after scan-token discovery because the readers only need the serialized scan tokens afterward.
- The reader now closes each `KuduScanner` in a `finally` block, including exceptional row-read paths.
- The sink initialization failure path closes the managed resource and preserves close failures as suppressed exceptions.
- `KuduUtil.getKuduClient(...)` remains available for compatibility, while the connector's internal paths now use the managed resource.

## Tests and CI

The added `KuduClientResourceTest` covers graceful executor termination, forced shutdown after timeout, and client-close failure while still shutting down the executor. The tests are Mockito-based and deterministic, with no sleeps, ports, real Kudu cluster, containers, or ordering-sensitive external state. Test stability rating: stable.

The current Build check is successful. I also recorded that the branch is currently `behind_by=2` and compare status is `diverged`, but since the current head already has a usable green Build, I would not ask for a sync-only follow-up here.

## Merge conclusion

### Conclusion: can merge

Blocking items:
- None.

Non-blocking suggestions:
- None required. Longer term, it could be useful to add an integration-style lifecycle test around reader/enumerator close, but the current unit coverage is enough for this focused fix.

Overall, this is a clean and useful resource-lifecycle improvement for the Kudu connector. Nice work.

davidzollo (approved):
+1

PR #11591, opened 2026-07-29
[Improve][Connector-V2] Guard POI Excel reads by file size

zhangshenghang:
### Purpose
Prevent large Excel files from OOMing workers when they are read through the POI engine.

### What changed
- Add `poi_excel_max_file_size` with a 50 MB default for POI Excel reads.
- Fail fast when POI would read an Excel file larger than the configured limit, and suggest `excel_engine = EasyExcel`.
- Keep the default `excel_engine` as POI for compatibility.
- Register Excel reader options in file source option metadata.
- Add tests for POI fail-fast behavior and EasyExcel bypassing the POI limit.
- Update LocalFile EN/ZH docs.

### Compatibility
The default engine remains POI. Users can increase `poi_excel_max_file_size` if POI is required, or switch to `EasyExcel` for large streaming reads.

### Tests
- `./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-base spotless:apply -DskipTests`
- `./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-base -Dtest=ExcelReadStrategyTest test`
- `./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-base -DskipTests install`
- `./mvnw -pl seatunnel-connectors-v2/connector-file/connector-file-local,seatunnel-connectors-v2/connector-file/connector-file-hadoop,seatunnel-connectors-v2/connector-file/connector-file-s3,seatunnel-connectors-v2/connector-file/connector-file-sftp,seatunnel-connectors-v2/connector-file/connector-file-obs,seatunnel-connectors-v2/connector-file/connector-file-oss,seatunnel-connectors-v2/connector-file/connector-file-cos,seatunnel-connectors-v2/connector-file/connector-file-ftp,seatunnel-connectors-v2/connector-file/connector-file-jindo-oss -DskipTests compile`
- `git diff --check`

### Note
A broader `-am` connector compile currently stops in `seatunnel-config-shade` before reaching the changed connector modules, with missing shaded Typesafe Config classes. The targeted connector-module compile above passed after installing the local `connector-file-base` artifact.


DanielLeens (changes requested):
Thanks for working on this! I went through the full diff and the direction makes sense: guarding the POI Excel path before building the workbook is a useful protection for large Excel files. I did find a couple of blockers that need to be fixed before this can merge. Please take a look, and happy to re-review once you push an update.

## What this PR changes

The PR adds `poi_excel_max_file_size` with a 50 MB default and checks it before POI creates `HSSFWorkbook` / `XSSFWorkbook`. It also exposes the Excel options from the file source factories and documents the new option for LocalFile.

## Runtime path I checked

```text
SeaTunnel job parse
  -> FactoryUtil.createAndPrepareSource()
      -> ConfigValidator.validate(sourceFactory.optionRule())
          -> SourceFactory.conditional(file_format_type, EXCEL, sheet_name, excel_engine, poi_excel_max_file_size)
              -> OptionRule.Builder.conditional(...) creates ConditionalRequiredOptions
          -> existing Excel configs without sheet_name / excel_engine / poi_excel_max_file_size fail before reader creation

Normal Excel read
  -> ExcelReadStrategy.read(path)
      -> resolveArchiveCompressedInputStream(..., NONE)
          -> hadoopFileSystemProxy.getInputStream(path)
          -> ExcelReadStrategy.readProcess()
              -> readByPoi()
                  -> assertPoiFileSize(split, currentFileName)
                  -> new HSSFWorkbook / XSSFWorkbook

Archived Excel read
  -> resolveArchiveCompressedInputStream(..., ZIP/TAR/TAR_GZ/GZ)
      -> copyInputStream(entryStream) copies the whole entry into ByteArrayOutputStream
      -> ExcelReadStrategy.readProcess()
          -> readByPoi()
              -> assertPoiFileSize(...) runs after the full entry has already been copied
```

## Issues

Issue 1: The Excel optionRule turns optional Excel settings into required settings
- Location: `seatunnel-connectors-v2/connector-file/connector-file-sftp/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sftp/source/SftpFileSourceFactory.java:80`
- Description: this PR adds `.conditional(FILE_FORMAT_TYPE, EXCEL, SHEET_NAME, EXCEL_ENGINE, POI_EXCEL_MAX_FILE_SIZE)` to several file source factories. In SeaTunnel's `OptionRule.Builder`, `conditional(...)` creates `ConditionalRequiredOptions`, so when `file_format_type = excel`, those options become required. That conflicts with the actual reader logic: `ExcelReadStrategy` treats `sheet_name` as optional and falls back to the first sheet when it is absent.
- CI evidence: the current Build failure is in `connector-file-sftp-it` and `connector-file-local-it`. The failing SFTP log reports `OptionValidationException` for `option: 'sheet_name'`, `constraint: required because ['file_format_type' == EXCEL] is true`. Existing E2E configs such as `sftp_excel_to_assert.conf` and `local_excel_to_assert.conf` use `file_format_type = excel` without `sheet_name`, which should remain valid.
- Risk: this breaks existing Excel source jobs at factory validation time before the reader is even created.
- Suggestion: please do not use required `conditional(...)` for these optional Excel settings. Put `SHEET_NAME`, `EXCEL_ENGINE`, and `POI_EXCEL_MAX_FILE_SIZE` under `.optional(...)` or use a non-required conditional display mechanism if the intent is only to expose them for Excel mode.
- Severity: High
- Raised by others: No

Issue 2: Archived Excel entries are copied into memory before the POI size guard runs
- Location: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/reader/AbstractReadStrategy.java:435`
- Description: for ZIP/TAR/TAR_GZ/GZ, `resolveArchiveCompressedInputStream()` calls `copyInputStream(...)` before `ExcelReadStrategy.readProcess()` is invoked. `copyInputStream()` reads the entire archive entry into a `ByteArrayOutputStream`. The new `assertPoiFileSize()` therefore runs too late for archived Excel files, and it checks the outer split/file length rather than the actual entry size.
- Risk: a compressed archive can still trigger high memory usage or OOM before the new guard has a chance to reject it. The reverse can also happen: a large outer archive with a small Excel entry may be rejected incorrectly.
- Suggestion: please move the guard ahead of the unbounded copy for archive entries. For ZIP/TAR, use entry size when available; for GZ, use a bounded copy that fails once it exceeds `poi_excel_max_file_size`, or otherwise avoid unbounded in-memory copying for Excel entries.
- Severity: High
- Raised by others: No

Issue 3: The new option is exposed to many file sources, but only LocalFile docs were updated
- Location: `docs/en/connectors/source/LocalFile.md:67`
- Description: `poi_excel_max_file_size` is now exposed by LocalFile, HdfsFile, S3File, FtpFile, SftpFile, OssFile, OssJindoFile, ObsFile, and CosFile source factories. The PR only updates LocalFile docs in English and Chinese.
- Risk: users of HDFS/S3/FTP/SFTP/OSS/OBS/COS Excel sources will not see the new limit or the EasyExcel guidance in the connector page they actually use.
- Suggestion: please update both `docs/en` and `docs/zh` for all file source connectors that expose this option.
- Severity: Medium
- Raised by others: No

## Tests and CI

The newly added unit tests are deterministic: they use fixed classpath Excel resources and local filesystem configuration, with no sleeps, ports, external services, or ordering-sensitive assertions. I would rate the new UT stability as stable.

The current CI failure is source-related, not an upstream-only signal. This branch is currently behind `dev` by 2 commits and the compare status is `diverged`, but the failing `connector-file-sftp-it` / `connector-file-local-it` evidence points directly to the optionRule change above, so syncing upstream alone is not the right next step.

## Merge conclusion

Blocking items:
- Issue 1: fix the optionRule compatibility break so existing Excel configs without `sheet_name` keep working.
- Issue 2: make the size guard effective before archived Excel entries are copied into memory.
- Issue 3: document the new user-facing option for every affected file source in both English and Chinese docs.

Non-blocking suggestion:
- Consider avoiding duplicate `POI_EXCEL_MAX_FILE_SIZE` / default declarations between base option classes so the value does not drift later.

Overall, the idea is good and worth keeping. Once the validation semantics and archive-entry boundary are fixed, this should be a much safer improvement for POI Excel reads.

DanielLeens (commented):
Thanks for pushing the update! I re-reviewed the full diff on the latest head (`e856eb0a7`), not just the incremental change, and I'm happy to say all three blockers from my July 29 review are addressed:

- Issue 1 (optionRule turning optional Excel settings into required ones) — fixed in every file source factory.
- Issue 2 (archived Excel entries copied into memory before the size guard ran) — fixed with a bounded copy + entry-size pre-check.
- Issue 3 (docs only updated for LocalFile) — fixed, all 9 affected connectors now have English and Chinese docs.

While verifying Issue 2's fix I found one new gap introduced by this exact commit, plus a CI failure and a merge-conflict fact that need attention before this can merge. Details below.

## What this PR changes

Guards POI-based Excel reads (`HSSFWorkbook`/`XSSFWorkbook`) behind a new `poi_excel_max_file_size` option (default 50 MB) so a large Excel file can't be fully materialized into POI's in-memory object model and OOM the worker. The new commit also fixes the option-rule regression, closes the archive-copy gap, and documents the option everywhere it's exposed.

## Runtime path I re-checked

```text
Non-archived Excel read (archiveCompressFormat = NONE)
  -> ExcelReadStrategy.read() -> resolveArchiveCompressedInputStream(..., FileFormat.EXCEL, poiMaxBytes)
      -> case NONE: readProcess(split, output, hadoopFileSystemProxy.getInputStream(path), ...)
          -> getExcelEngine() dispatch
              -> EasyExcel branch: readByEasyExcel() -- no size check (correct, EasyExcel streams)
              -> POI branch: readByPoi() -> assertPoiFileSize(split, currentFileName) -- checked correctly

Archived Excel read (ZIP/TAR/TAR_GZ/GZ)
  -> ExcelReadStrategy.read()
      -> maxBytesForEntry = getPoiExcelMaxFileSize()   // computed unconditionally, BEFORE engine dispatch
      -> resolveArchiveCompressedInputStream(..., FileFormat.EXCEL, maxBytesForEntry)
          -> for each entry: assertArchiveEntrySize(entry.getName(), entry.getSize(), maxBytesForEntry)
          -> copyInputStream(entryStream, maxBytesForEntry)  // now bounded, throws once exceeded
          -> readProcess(...) -> engine dispatch happens only AFTER the bound has already been enforced
```

Other archived formats (CSV/JSON/Text/XML) still call the original 4-arg `resolveArchiveCompressedInputStream(...)` overload, which passes `effectiveMaxBytes = -1` (unbounded) — confirmed via `JsonReadStrategy.java:89`, `TextReadStrategy.java:179`, `CsvReadStrategy.java:90`, `XmlReadStrategy.java:95`. So the blast radius of this change is correctly scoped to Excel only.

## Before / after (Issue 2 fix)

`AbstractReadStrategy.java` — the ZIP branch before this commit:
```java
if (!entry.isDirectory() && checkFileType(entry.getName(), fileFormat)) {
    readProcess(split, output, copyInputStream(zis), partitionsMap, entry.getName());
}
```
After:
```java
if (!entry.isDirectory() && checkFileType(entry.getName(), fileFormat)) {
    assertArchiveEntrySize(entry.getName(), entry.getSize(), effectiveMaxBytes);
    readProcess(split, output, copyInputStream(zis, effectiveMaxBytes), partitionsMap, entry.getName());
}
```
Same pattern applied to the TAR/TAR_GZ/GZ branches, and `copyInputStream` now throws `FileConnectorException` once the running byte count exceeds `maxBytes`.

## Key findings

- The normal read path for the default engine (POI) does reach `assertPoiFileSize` / `assertArchiveEntrySize` for both plain and archived Excel files, so the core protection this PR is meant to add is real and works for the default configuration.
- `ExcelReadStrategy.read()` computes `maxBytesForEntry` from `getPoiExcelMaxFileSize()` unconditionally, before it is known which engine (`POI` vs `EasyExcel`) will actually process the entry — see "Issue 1" below.
- This is a precise fix for the archive-copy problem raised in my previous round (bounded copy + entry-size pre-check for ZIP/TAR, bounded copy for GZ), not a workaround.
- No test in `ExcelReadStrategyTest.java` exercises the archived (ZIP/TAR/GZ) code path at all — `testEasyExcelIgnoresPoiFileSizeLimit` and `testPoiRejectsExcelLargerThanConfiguredLimit` both read `e2e.xlsx` directly (uncompressed), so they only cover the `NONE` branch, not the branch this fix actually touches.

## Issues

**Issue 1: Archived Excel entries are still bounded by the POI limit even when `excel_engine = EasyExcel`**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/reader/ExcelReadStrategy.java:79-88`
- Description: for the non-archived path, `assertPoiFileSize` is correctly called only inside `readByPoi()` (`ExcelReadStrategy.java:186`), so `readByEasyExcel()` is never size-limited — that's the intended, documented behavior. For the archived path, `read()` computes `maxBytesForEntry = getPoiExcelMaxFileSize()` and passes it into `resolveArchiveCompressedInputStream(...)` *before* the engine is known. `assertArchiveEntrySize(...)` and the bounded `copyInputStream(...)` then run in `AbstractReadStrategy.java` ahead of `readProcess()`, so the POI-specific limit is enforced even when the user has explicitly configured `excel_engine = EasyExcel`.
- This directly contradicts the docs added in this same commit, e.g. `docs/en/connectors/source/FtpFile.md:437`: "Only used when `file_format` is excel and `excel_engine` is POI." For an archived Excel file, that's not true — the option applies regardless of engine.
- Potential risk: a user who reads the size-limit error message ("Please set `excel_engine = EasyExcel`... if POI is required") and follows that exact advice for an archived Excel file larger than 50 MB will hit the same rejection again, because the archive-entry guard doesn't know about the engine choice. This silently breaks the one escape hatch the PR itself documents, specifically for the archived-file scenario.
- Best improvement: resolve the engine before computing the bound, e.g. in `read()`:
  ```java
  ExcelEngine excelEngine = getExcelEngine();
  long maxBytesForEntry = ExcelEngine.EASY_EXCEL.equals(excelEngine) ? -1L : getPoiExcelMaxFileSize();
  ```
  This makes the archived path consistent with the already-correct non-archived path.
- Severity: Medium (default engine is POI, so the main/default path is unaffected; the gap only surfaces when a user explicitly opts into EasyExcel for an archived file, and a workaround — raising the limit or using an uncompressed file — exists).

**Issue 2: CI is currently failing — Spotless code-style check**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/reader/AbstractReadStrategy.java` (the three `assertArchiveEntrySize(...)` call sites added by this commit, ZIP/TAR/TAR_GZ branches)
- Description: the `Run / Code style` job on the current head (`e856eb0a7`) fails with `spotless-maven-plugin:check`, reporting formatting violations exactly at the new `assertArchiveEntrySize(entry.getName(), entry.getSize(), effectiveMaxBytes)` calls (each argument currently on its own line where Spotless expects them joined onto one line). This is directly caused by this PR's own new commit, not by upstream drift — the diff in the CI log matches the new code added here.
- Potential risk: blocks merge outright; CI will keep failing until this is fixed.
- Best improvement: run `./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-file/connector-file-base -am` locally and push the reformatted file. This is a trivial, mechanical fix.
- Severity: High (blocks merge)

**Issue 3: Branch has real merge conflicts with `dev`**
- Location: N/A (branch-level, not a specific file)
- Description: `gh api repos/apache/seatunnel/pulls/11591` reports `mergeable_state=dirty`, `mergeable=false`, `mergeStateStatus=DIRTY`. The base...head compare shows `status=diverged`, `ahead_by=2`, `behind_by=37`. A `git merge-tree` check shows this branch and `dev` both touched several of the same files since the merge-base (all 9 connector doc pairs plus `FileBaseSourceOptions.java` and `AbstractReadStrategy.java`), which lines up with GitHub's conflict signal.
- Potential risk: the PR cannot be merged in its current state regardless of code-review outcome.
- Best improvement: please sync with the latest `dev` and resolve the conflicts, then re-push so CI reruns against the merged result.
- Severity: High (blocks merge)

**Issue 4: The new archive-guard code path has no test coverage**
- Location: `seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/reader/ExcelReadStrategyTest.java`
- Description: this commit did not add or modify any test. `testEasyExcelIgnoresPoiFileSizeLimit` (line 242) and `testPoiRejectsExcelLargerThanConfiguredLimit` (line 269) both call `excelReadStrategy.read(...)` against `e2e.xlsx` directly from the classpath (uncompressed), so `archiveCompressFormat == NONE` and neither test ever reaches `assertArchiveEntrySize`/the bounded `copyInputStream` added by this fix. There are also no ZIP/TAR/GZ Excel fixtures under `src/test/resources/excel`.
- Potential risk: the archived-entry guard (which is the entire point of this commit's Issue 2 fix) is unverified — including the gap described in Issue 1 above, which a test with `excel_engine = EasyExcel` against a zipped Excel entry would likely have caught.
- Best improvement: add at least one test that reads a zipped/tarred Excel entry above the configured limit and asserts rejection with POI, and one that confirms EasyExcel is not blocked for the same archived entry.
- Severity: Medium

**Issue 5: Duplicate option/constant declaration between `FileBaseOptions` and `FileBaseSourceOptions`** (carryover, non-blocking suggestion from my previous review, still unaddressed)
- Location: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/config/FileBaseOptions.java:29,124-127` and `FileBaseSourceOptions.java:31,126-129`
- Description: `DEFAULT_POI_EXCEL_MAX_FILE_SIZE` and the `POI_EXCEL_MAX_FILE_SIZE` option are declared independently in both classes with the same value.
- Potential risk: low today, but the two declarations can silently drift apart in a future change.
- Best improvement: keep a single source of truth (e.g. `FileBaseSourceOptions` extends `FileBaseOptions`, so the child class doesn't need its own copy) and remove the duplicate.
- Severity: Low

## Compatibility

Fully compatible for existing (pre-PR) SeaTunnel behavior: non-Excel archived formats are untouched (verified they still use the unbounded 4-arg overload), and the default POI/50 MB behavior matches what was already being added by this still-unmerged PR. Since `poi_excel_max_file_size` and `excel_engine` are new, unreleased options introduced by this same PR, there is no released-version compatibility break to manage — the only compatibility concern is the internal inconsistency described in Issue 1 (implementation vs. the docs shipped in this same commit).

## Side effects

CPU/memory: strictly positive — bounding the archive-entry copy prevents unbounded `ByteArrayOutputStream` growth for POI-engine archived Excel reads, which is the intended protection. No new locking, retries, or concurrency concerns; `assertArchiveEntrySize` and the bounded `copyInputStream` throw promptly and the enclosing try-with-resources blocks still close the ZIP/TAR streams on that path, so no resource leak.

## Error handling and logging

Exceptions are typed (`FileConnectorException` with `CommonErrorCodeDeprecated.UNSUPPORTED_OPERATION`), include the entry name and both the observed and configured sizes, and point the user at the two remedies (switch engine or raise the limit) — good user-facing context. No sensitive data is logged. Log levels for the existing `Parsing Excel with POI`/`EasyExcel` info logs are unchanged and appropriate.

## Test stability

No UT/E2E timing, concurrency, or environment-dependent code was touched or added in this commit (no new tests were added at all, per Issue 4), so there is nothing to rate under the flaky-test checklist for this round.

## Documentation

`docs/en` and `docs/zh` are both updated and consistent with each other for all 9 affected connectors (CosFile, FtpFile, HdfsFile, LocalFile, ObsFile, OssFile, OssJindoFile, S3File, SftpFile). The one factual gap is the "Only used when `excel_engine` is POI" claim not holding for archived files, per Issue 1.

## Architecture

Precise fix for Issues 1-3 from my previous review. The archive-guard mechanism (entry-size pre-check + bounded copy) is a reusable pattern that other format-specific read strategies could adopt later if they ever need a similar protection, since it's implemented generically in `AbstractReadStrategy` and only opted into via the new 5-arg overload.

## Issue summary

| 1 | Archived Excel entries still bounded by POI limit under `excel_engine = EasyExcel` | `ExcelReadStrategy.java:79-88` | Medium |
| 2 | CI Spotless code-style check failing | `AbstractReadStrategy.java` (new `assertArchiveEntrySize` call sites) | High |
| 3 | Branch has real merge conflicts with `dev` (`mergeable_state=dirty`, diverged, behind_by=37) | branch-level | High |
| 4 | No test coverage for the new archive-entry size guard | `ExcelReadStrategyTest.java` | Medium |
| 5 | Duplicate `POI_EXCEL_MAX_FILE_SIZE`/default constant in two option classes (carryover, non-blocking) | `FileBaseOptions.java`, `FileBaseSourceOptions.java` | Low |

## Merge recommendation

   - Issue 3: resolve the merge conflicts with `dev` — the PR cannot be merged in its current state regardless of anything else.
   - Issue 2: fix the Spotless formatting violation (`mvn spotless:apply`) so CI passes.
   - Issue 1: scope the archived-entry size guard to the POI engine only, so it matches the already-correct non-archived behavior and the docs shipped in this commit.
   - Issue 4: add test coverage for the archived Excel-entry guard (both POI-rejects and EasyExcel-is-not-blocked cases), since this is the exact code path Issue 2 from my last review was about.

   - Issue 5: de-duplicate the `POI_EXCEL_MAX_FILE_SIZE` constant/option between `FileBaseOptions` and `FileBaseSourceOptions`. Doesn't affect current behavior, just a maintainability nit for later.

Overall: the direction and the fixes for my previous round's blockers are solid — Issues 1-3 from my last review are genuinely resolved, not just papered over, and I verified each one against the actual diff and call chain. What's left is a merge-gate issue (conflicts), a mechanical CI fix (Spotless), and one real-but-narrow correctness gap in the exact code this commit just added (Issue 1), which the missing test coverage (Issue 4) would have caught. None of these should take long to close out — happy to take another look as soon as they're pushed.

DanielLeens (changes requested):

Apache POI's workbook classes fully materialize an Excel file into an in-memory object model before a single row can be read. For very large `.xls`/`.xlsx` files this can push a Zeta worker into heavy GC pressure or OOM, with no way to fail fast or opt out short of switching to a different code path entirely. This PR adds a configurable `poi_excel_max_file_size` option (default 50 MB) that rejects oversized files before POI is asked to build the workbook, and points users at the already-existing `excel_engine = EasyExcel` streaming alternative as the escape hatch. It also extends the archive-entry read path (ZIP/TAR/TAR_GZ/GZ) so an Excel file bundled inside an archive gets the same protection, and propagates the new option (plus the previously Local-only `excel_engine` option) to all nine Excel-capable file source connectors.

This is the third review round on this PR. An earlier round found three blockers: an option-rule change that made related options incorrectly required for all Excel jobs, an archive read path that copied the whole entry into memory before the size guard ran, and docs that only covered one connector. A later round confirmed those three were fixed and found a new gap (the POI size bound still applied even when the user picked the streaming engine for archived files), plus a failing formatter check, a merge conflict, missing archive-path test coverage, and a non-blocking duplicate-constant nit. This review re-examines the full diff at the current head from scratch.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Non-archived path (the default, and the common case).** The POI read method now calls a size-assertion helper as its very first statement, before the workbook object is constructed. That helper resolves the size from the split's known length if available, otherwise falls back to a metadata-only file-status stat call — not a byte read. This is exactly the right shape for a guard: checked before the expensive/unsafe operation, and it does not itself consume memory proportional to file size. For `excel_engine = EasyExcel`, the read method is never routed through the POI size assertion at all — correct, since the streaming engine doesn't need the same bound, and this directly fixes the previously-flagged gap (POI's limit leaking onto the streaming engine for archived files) — verified by reading the code, not just trusting the round-2 write-up.

**Archived path (ZIP/TAR/TAR_GZ/GZ).** The archive-entry resolution method now asserts the entry's declared size before copying its bytes — correctly bounding memory ahead of the unbounded-copy risk flagged in round 1. The copy helper additionally enforces the same bound while streaming bytes, sound belt-and-suspenders design for entries whose declared size is unknown.

**A real correctness bug survives in the archived path, however.** After the entry is copied, the read process still routes into the POI read method, which unconditionally calls the size-assertion helper again. For an archived Excel read, the split object passed in was built once, early, using only the outer archive's path — a two-argument constructor that leaves length unset. So inside the size-assertion helper, the "known length" branch is false, and it falls back to a file-status stat on the split's path — which is the size of the **outer archive on disk**, not the size of the Excel entry inside it (the entry's name is only used for the error message, never for the size lookup).

Concretely: an 80 MB zip file (because it also bundles several large non-Excel files) containing a 2 MB Excel entry, read with the default 50 MB limit: the entry-level check passes correctly (2 MB), the bounded copy succeeds correctly, but the redundant POI-level check then resolves the *archive's* 80 MB size and incorrectly rejects a perfectly valid, tiny Excel file — with a misleading error message reporting the archive's size instead of the entry's real size. This is precisely the "reverse" risk an earlier review round called out in the abstract (a large outer archive with a small Excel entry may be rejected incorrectly) before the entry-level guard existed. A later round added that entry-level guard and closed the OOM half of the concern, but the redundant, wrong-target check inside the POI read method was never removed, so the false-rejection half is still live at the current head. Neither round's test additions catch it: the two archive tests both use a fixture where the zip's on-disk size and the entry's uncompressed size are close enough that both the correct entry-level check and the buggy split-level check produce the same pass/fail outcome — the discriminating scenario (small entry inside a materially larger archive) is never exercised. See Issue 1 below.

## 1.2 Compatibility Impact

For the non-archived path, this is a real, if narrow, backward-compatibility break in default behavior: any existing job that reads a POI-engine Excel file larger than 50 MB — which previously succeeded (at the cost of memory pressure) — will now fail fast unless the user explicitly raises the limit or switches engines. That is a deliberate and reasonable trade (fail fast beats OOM), and it is documented in all nine connector doc pages with the exact default and remediation path, which is the right way to introduce it. However, this repository's own contribution guidelines are explicit that incompatible changes must be documented in the incompatible-changes concepts doc with migration guidance. This PR does not touch that file at all. Given the file already has a template for exactly this kind of case (a "newly rejected config" table listing previous behavior vs. new behavior vs. migration), this PR should add a comparable row for the new default file-size cap rather than relying solely on the per-connector option docs (see Issue 2 below).

The size check itself is O(1) — a single stat call, not a data read — for the non-archived path, which is the common case; correct performance profile for a guard. The bounded copy for the archived path is a strict improvement over the pre-PR unbounded buffer growth — it now fails fast instead of growing without limit. No new locking, retries, threads, or blocking I/O patterns. Exceptions are thrown promptly and the enclosing resource-management blocks still close archive streams on the guard-triggered exit path, so no resource leak on rejection.

**Issue 1: The POI size guard checks the archive's file size, not the actual Excel entry's size, for archived reads — causing false rejections and misleading error text**
- **Location:** the POI read method's size-assertion call and its implementation, combined with how the split object is constructed for archived reads (always leaving length unset).
- **Problem description:** for ZIP/TAR/TAR_GZ/GZ archives, the size-assertion helper resolves file size via a file-status stat on the split's path, and that path is always the outer archive's path, never the individual entry, because the split built for archived reads never carries a length or an entry-specific path. This duplicates — and contradicts — the correct entry-level check already performed earlier using the archive entry's own declared size.
- **Potential risk:** legitimate archives containing a small Excel entry alongside other, larger content will be rejected with an exception that reports the wrong (archive-level) byte count, confusing to debug and blocking a valid use case that this PR's own archive-support work is supposed to enable. This is a functional correctness bug, not a memory-safety hole (the entry-level guard already covers the OOM concern), but it directly undermines the feature for a realistic subset of archived-Excel users.
- **Best improvement:** either skip the redundant POI-level size check for the archived path (the entry-level check plus bounded copy already provide equivalent protection there), or thread the real entry size through to the POI read method instead of re-deriving it from the split.
- **Severity:** High (functional bug affecting a use case this PR explicitly ships support for)
- **Raised by another reviewer:** No — an earlier round described the theoretical risk before the entry-level guard existed, and a later round confirmed the entry-level guard was added but did not re-verify that this specific redundant, wrong-target check was also removed/fixed. As of the current head it is still present.

**Issue 2: New default 50 MB POI limit is not recorded in the incompatible-changes doc**
- **Location:** `docs/en/introduction/concepts/incompatible-changes.md` (not modified by this PR)
- **Problem description:** this PR changes default behavior for existing users, which this repo's own contribution guidelines classify as requiring an entry in this file with migration guidance, in addition to the per-connector doc updates this PR already did well.
- **Potential risk:** users upgrading past this change with large POI-Excel jobs in production will discover the new failure mode only at runtime, with no changelog signal pointing them to the fix.
- **Best improvement:** add a short entry to the incompatible-changes doc following the existing table format used for similar "newly rejected config" entries.
- **Severity:** Medium (documentation completeness gap for a real behavior change, not a code defect)

Beyond these two, exception usage is otherwise solid: exceptions are used consistently with existing error-code conventions in this file, messages include the offending file/entry name plus both observed and configured byte counts formatted for readability, with concrete remediation. No sensitive data is logged.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The refactor of the monolithic read-process method into smaller named helpers is a clean, minimal-footprint extraction that preserves all pre-existing POI/EasyExcel logic while making the new guard easy to locate and reason about. License headers, import style, and method ordering all follow existing repo conventions. One non-blocking maintainability nit: the size-limit constant and its option are declared independently in two option classes with identical values; only one copy is ever referenced (confirmed by a repo-wide search — the other is unreferenced dead code). This mirrors a pre-existing duplication pattern already present for the `excel_engine` option before this PR, so it's not a new anti-pattern, but it was already flagged in the prior round as a non-blocking carryover and remains unaddressed at the current head.

Four new unit tests cover: the streaming engine ignoring the POI limit (non-archived), POI rejecting an over-limit file with the expected error text (non-archived), POI rejecting an over-limit archived entry, and the streaming engine not being blocked for an archived entry above the POI limit — good under-limit/over-limit coverage for both engines and both archived/non-archived paths, exactly the coverage the prior round asked for. As noted in 1.1/1.4, though, the archive fixture used can't distinguish "checked the entry's real size" from "checked the archive's on-disk size," so Issue 1 slips through despite the added coverage looking complete on the surface.

Flaky-pattern check (mandatory since test code changed): all four new tests use fixed classpath resources, a local-filesystem backend, and deterministic in-memory assertions. No sleeps, no thread timing, no network/external services, no shared mutable state across tests, no ordering dependency between test methods. **Rating: Stable.**

## 2.3 Documentation Updates

`docs/en` and `docs/zh` are both updated, consistently, for all nine affected connectors, each adding the relevant option table rows plus prose documenting the 50 MB default and the streaming-engine escape hatch — this directly resolves the earlier round's documentation gap, and English/Chinese content is aligned in wording and structure. As discussed above, the one documentation gap this review considers blocking is the missing incompatible-changes entry.

# 3. Architectural Soundness

The core idea — check cheap metadata before an expensive, unbounded operation — is the right shape for this kind of guard, and the non-archived implementation executes it correctly. Extending the same intent to archives via a bounded copy is a reasonably elegant way to opt Excel in without touching other formats' archive reads at all — verified other format readers still use the original unbounded copy signature, so this PR's blast radius is correctly confined to Excel. The one blemish is the redundant, wrong-target size check surviving inside the POI read method for the archived path (Issue 1) — the archive-level and file-level guards were clearly meant to compose, but they currently talk past each other.

## 3.2 Maintainability

The extraction into named private methods makes the guard's intent legible and testable in isolation, a maintainability win over the prior single large read-process method. The duplicate option-constant declaration is a small, pre-existing pattern that should eventually be consolidated, but doesn't block this change.

## 3.3 Extensibility

The archive-entry guard pattern is implemented generically at the base-class level and only activated via a new overload, so other format-specific read strategies with similar unbounded-materialization risk could opt into the same bounded-copy mechanism later with a small, additive change. This is genuinely reusable infrastructure, assuming Issue 1 is fixed so the pattern is trustworthy end-to-end.

The new options are brand-new, unreleased, introduced by this same PR, so there is no released-version option-compatibility break to manage for those two specifically. The one genuine compatibility concern is the new default behavior itself (see 1.2/Issue 2) — a legitimate, disclosed trade-off, but one that needs the incompatible-changes doc entry to meet this repo's own bar for shipping such a change.

# 4. Issue Summary

| 1 | POI size guard measures the outer archive's on-disk size instead of the actual Excel entry's size for archived reads, causing false rejections | POI read method's size-assertion call and implementation; archived-read split construction | High |
| 2 | New default 50 MB POI cap changes existing behavior but is not recorded in the incompatible-changes doc | `docs/en/introduction/concepts/incompatible-changes.md` (not modified) | Medium |
| 3 | Size-limit constant/option declared independently in two places; only one referenced | Option classes for the base file connector | Low (non-blocking carryover from prior round) |

# 5. Merge Recommendation

### Conclusion: Can merge after fixes

The direction of this PR is sound and the bulk of the implementation is solid: the primary, default-path guard (non-archived POI Excel reads) is correctly ordered and well tested with both under-limit and over-limit cases for both engines. All three blockers from the first review round and three of the four blockers from the second round are genuinely fixed at the current head, and CI on this exact commit is fully green including unit tests. Documentation coverage across all nine affected connectors, in both English and Chinese, is thorough and consistent.

**Blocking:**
- Issue 1 — the archived-path size check still measures the wrong file, producing false rejections for valid small Excel entries bundled in larger archives, and existing tests don't discriminate this because the committed archive fixture makes archive size and entry size nearly identical. This should be fixed (skip the redundant POI-level check for archived reads, or thread the real entry size through) and covered by a test where archive size and entry size clearly diverge.
- Issue 2 — add the required incompatible-changes entry for the new default 50 MB POI cap, with brief migration guidance, consistent with this repo's own documentation rule for behavior-changing defaults.

**Non-blocking:**
- Issue 3 — de-duplicate the constant/option declaration when convenient; doesn't affect current behavior.

Overall assessment: this is close. The engineering judgment behind the guard is correct, the non-archived (default) path is safe and well-verified, and the author has been responsive and precise in addressing two full rounds of prior feedback. What remains is one real, narrow-but-genuine correctness bug in the archived-file code path this PR itself introduces, plus a documentation-process gap for the new default's compatibility impact — both are small, targeted fixes rather than a redesign, and this should be quick to close out.

DanielLeens (changes requested):
Sorry for the gap since my last round — I re-fetched the latest head and re-read the complete diff and runtime path from scratch, as required for a re-review after a head change. This is round 4.

Apache POI's workbook classes (`HSSFWorkbook`/`XSSFWorkbook`) fully materialize an Excel file into an in-memory object model before a single row can be read. For very large `.xls`/`.xlsx` files this can push a Zeta worker into heavy GC pressure or OOM, with no way to fail fast or opt out short of switching to a different code path entirely. This PR adds a configurable `poi_excel_max_file_size` option (default 50 MB) that rejects oversized files before POI is asked to build the workbook, and points users at the already-existing `excel_engine = EasyExcel` streaming alternative as the escape hatch. It also extends the archive-entry read path (ZIP/TAR/TAR_GZ/GZ) so an Excel file bundled inside an archive gets the same protection, and propagates the new option (plus the previously Local-only `excel_engine` option) to all nine Excel-capable file source connectors.

Round 1 found three blockers (an option-rule change that made related options incorrectly required for all Excel jobs, an archive read path that copied the whole entry into memory before the size guard ran, and docs that only covered one connector) — all confirmed fixed by round 2. Round 2 confirmed those three and found a new gap (the POI size bound still applied even when the user picked the streaming engine for archived files) — fixed by round 3. Round 3 confirmed that fix but found a genuine correctness bug still live: the POI read method redundantly re-checked file size for archived reads using the outer archive's on-disk size instead of the actual Excel entry's size, causing false rejections of small Excel files bundled in larger archives — plus a missing incompatible-changes doc entry for the new default's behavior change. This review independently re-verifies both at the current head (`8836f9310bec2be8d88e82009f303cd3799b0d9d`) from scratch, and both are confirmed fixed with real evidence (code + a genuinely discriminating regression test + a doc entry), detailed below.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Files:** `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/reader/ExcelReadStrategy.java`, `.../source/reader/AbstractReadStrategy.java`

**Runtime path (traced end-to-end against the current head):**

```text
Source config (file_format=excel, excel_engine, poi_excel_max_file_size, archive_compress_codec)
        |
        v
BaseFileSourceReader.pollNext() / MultipleTableFileSourceReader.pollNext()
  BaseFileSourceReader.java:68    readStrategy.read(split.splitId(), "", output)
  MultipleTableFileSourceReader.java:87  readStrategy.read(split, output)
        |  (ReadStrategy.java:49-51 default overload just forwards to the 3-arg method below,
        |   so both call sites converge on the same entry point -- no bypass)
        v
ExcelReadStrategy.read(String path, String tableId, Collector output)   [ExcelReadStrategy.java:80]
        |  computes maxBytesForEntry = (-1 if EasyExcel else getPoiExcelMaxFileSize())
        v
AbstractReadStrategy.resolveArchiveCompressedInputStream(split, output, partitionsMap,
                                                          FileFormat.EXCEL, maxBytesForEntry)
        |
        +-- archiveCompressFormat != NONE (ZIP/TAR/TAR_GZ/GZ):
        |     assertArchiveEntrySize(entry.getSize(), effectiveMaxBytes)  [entry-level guard, metadata-only
        |                                                                  for ZIP/TAR; AbstractReadStrategy.java:553]
        |     copyInputStream(entryStream, effectiveMaxBytes)             [bounded streaming copy, throws mid-stream
        |                                                                  if actual bytes exceed the limit --
        |                                                                  AbstractReadStrategy.java:650]
        |     --> readProcess(split, output, boundedInputStream, ...)
        |
        +-- archiveCompressFormat == NONE (direct file):
              --> readProcess(split, output, hadoopFileSystemProxy.getInputStream(path), ...)
                        |
                        v
        ExcelReadStrategy.readProcess(...) -> readByPoi(...) or readByEasyExcel(...)
                        |
                        v (POI engine only)
        readByPoi(...)  [ExcelReadStrategy.java:182]
              if (archiveCompressFormat == NONE) assertPoiFileSize(split, currentFileName);  [ExcelReadStrategy.java:187-188]
                    fileSize = split.getLength() > -1 ? split.getLength()
                               : hadoopFileSystemProxy.getFileStatus(split.getFilePath()).getLen();  <- stat only, no read
              -- guard checkpoint --
              new HSSFWorkbook(inputStream) / new XSSFWorkbook(inputStream)   <- the expensive, unbounded materialization
```

**Non-archived path (the default, and the common case).** `assertPoiFileSize` (`ExcelReadStrategy.java:274-291`) is the very first statement inside `readByPoi`, executed strictly before `new HSSFWorkbook(...)`/`new XSSFWorkbook(...)`. It resolves size via `split.getLength()` if set, otherwise a `hadoopFileSystemProxy.getFileStatus(...).getLen()` call — a metadata-only stat, not a byte read. Exactly the right shape for a guard. One factual note, not a defect: because `ExcelReadStrategy.read()` always constructs `new FileSourceSplit(tableId, path)` (2-arg constructor, length left at -1), the `split.getLength() > -1` branch is currently unreachable in practice for Excel — every direct-file POI read takes the `getFileStatus` stat branch. Behavior is still correct (both are O(1) metadata reads), just worth knowing when reading the code.

**Archived path — Issue 1 from round 3, re-verified fixed.** `readByPoi` now gates the redundant POI-level check with `if (archiveCompressFormat == ArchiveCompressFormat.NONE)` (`ExcelReadStrategy.java:187`). `archiveCompressFormat` is set once in `AbstractReadStrategy.setPluginConfig()` (`AbstractReadStrategy.java:333`, sourced from the job's `archive_compress_codec` option) — a stable, per-strategy-instance, config-driven signal, so it reliably reflects whether *this job* is reading archives at all before any file is touched. For archived reads, the entry-level guard (`assertArchiveEntrySize` + bounded `copyInputStream`) already ran and already protects memory; `assertPoiFileSize` — which would have re-statted the *outer archive's* path, not the entry — is now skipped entirely. I traced this by hand rather than trusting the commit message: the code genuinely no longer calls `assertPoiFileSize` when `archiveCompressFormat != NONE`. This closes the false-rejection bug (a small Excel entry inside a large archive would previously be wrongly rejected using the archive's on-disk size).

This round also ships the test round 3 explicitly said was missing: `testPoiAcceptsSmallExcelEntryInsideLargerArchive` (`ExcelReadStrategyTest.java:373-442`) builds a ZIP whose on-disk size (`poiLimit * 4`, via uncompressed STORED padding) is materially larger than the configured `poi_excel_max_file_size`, while its only Excel entry is smaller than that limit, and asserts the read succeeds with 5 rows. This is the exact discriminating scenario round 3 said the existing `e2e_in_zip.zip` fixture couldn't distinguish (archive size and entry size were close enough that both the correct and buggy checks agreed). I traced this scenario through the code by hand: with the fix, `readByPoi` never re-checks the archive's size, so the read succeeds; without the fix, `assertPoiFileSize` would stat the padded archive file, see it exceeds `poiLimit`, and throw — so the test genuinely fails without the fix and passes with it. A real regression test, not a superficial addition.

**EasyExcel / streaming engine.** `maxBytesForEntry` is computed once in `ExcelReadStrategy.read()` (`ExcelReadStrategy.java:83-84`) as `-1L` whenever `getExcelEngine()` resolves to `EASY_EXCEL`, and `-1` disables both `assertArchiveEntrySize` and the bounded-copy check inside `copyInputStream`. `readByPoi`/`assertPoiFileSize` are never invoked at all for EasyExcel. Correct: this is the previously-flagged-and-fixed round-2 gap, and it remains fixed.

**Zip-bomb question — file-size-on-disk vs. post-decompression size.** For ZIP/TAR/TAR_GZ, `assertArchiveEntrySize` reads `entry.getSize()` (`AbstractReadStrategy.java:453,472,493`), a header-declared value read from metadata before any bytes are decompressed — genuinely pre-decompression. For `ZipEntry` via `ZipInputStream`, the declared size can in principle be absent or spoofed. This PR does not rely on that value alone: `copyInputStream(stream, maxBytes)` (`AbstractReadStrategy.java:650-664`) enforces the bound *while streaming*, incrementing a running total per buffer read and throwing the moment the actual decompressed byte count exceeds `maxBytes` — genuine belt-and-suspenders design, not just declared-metadata trust. For GZ, there is no discrete "entry" and no reliable pre-read size, so the implementation correctly relies solely on the same bounded streaming copy — the only sound approach for that format.

**A residual, narrower gap worth flagging as new (not carried over).** The guard bounds the *outer* file (or archive entry) size, not the *decompressed* size of the internal ZIP structure that `.xlsx` itself is (OOXML `.xlsx` files are themselves ZIP archives internally). A `.xlsx` file small enough to pass this PR's 50 MB limit on disk could theoretically still be crafted to expand substantially when POI opens the package. This repo pins `poi.version=4.1.2` (`connector-file-base/pom.xml:37`), which includes POI's own `ZipSecureFile` safeguard (min inflate ratio 1%, max entry size ~4 GB), throwing on violation — mitigating the worst case but not eliminating it (up to ~100x expansion is still possible within POI's own limit). This is not the bug round 3 flagged — it's a different, deeper limitation inherent to any outer-file-size guard applied to a format that is itself compressed. Given the PR's stated and documented scope is "guard by file size," not "harden POI's own OOXML unpacking," and POI's `ZipSecureFile` already provides a meaningful backstop at this version, I am not treating this as a blocker.

## 1.2 Compatibility Impact

**Partially incompatible — by design, and now correctly documented.** For the non-archived path, this is a real, narrow, backward-compatibility break in default behavior: an existing job that reads a POI-engine Excel file larger than 50 MB — which previously succeeded (at the cost of memory pressure) — will now fail fast with a `FileConnectorException` unless the user raises `poi_excel_max_file_size` or switches to `excel_engine = EasyExcel`. A deliberate, reasonable trade (fail fast beats OOM). I independently confirmed the required incompatible-changes doc entry is now present in both `docs/en/introduction/concepts/incompatible-changes.md` and `docs/zh/introduction/concepts/incompatible-changes.md`, with affected-component list, description, impact statement, and migration guidance — matching the existing table format used by adjacent entries in the same file.

For the archived path, behavior for files that were previously read successfully is unchanged or improved: entries within the configured limit read exactly as before; entries genuinely over the limit now fail fast (previously copied into an unbounded in-memory buffer — itself the round-1 defect this PR fixed). No new false rejections remain after the round-3 fix.

The non-archived-path check is O(1) — a single stat call, not a data read. The archived-path entry-level check is also O(1) metadata; the bounded copy is a strict improvement over the pre-PR unbounded buffer growth. No new locking, retries, threads, or blocking I/O patterns are introduced. Exceptions are thrown promptly, and the enclosing try-with-resources blocks still close on the guard-triggered exit path — verified by re-reading the try-with-resources scoping in `resolveArchiveCompressedInputStream`; no resource leak on rejection.

**Issue 1: POI size guard measured the outer archive's on-disk size instead of the actual Excel entry's size for archived reads (Round 3 blocker). Fixed, re-verified.**
- Location: `ExcelReadStrategy.java:182-207` (`readByPoi`), guard gated at line 187.
- What was wrong: `readByPoi` unconditionally called `assertPoiFileSize`, which re-statted `split.getFilePath()` — always the outer archive's path for archived reads — producing false rejections of valid small Excel entries bundled in larger archives.
- Fix verified: `assertPoiFileSize` is now only called `if (archiveCompressFormat == ArchiveCompressFormat.NONE)`. For archived reads the entry-level guard is the sole size guard, which is sufficient.
- Test verified: `testPoiAcceptsSmallExcelEntryInsideLargerArchive` (`ExcelReadStrategyTest.java:373-442`) is the exact discriminating test round 3 asked for, and traces correctly (fails without the fix, passes with it).
- Severity: was High; now closed.
- Raised by another reviewer: this is my own carryover from round 3, now confirmed fixed with evidence beyond the commit message.

**Issue 2: New default 50 MB POI limit not recorded in the incompatible-changes doc (Round 3 blocker). Fixed, re-verified.**
- Location: `docs/en/introduction/concepts/incompatible-changes.md`, `docs/zh/introduction/concepts/incompatible-changes.md`.
- Fix verified: both files now contain a new entry with affected components (all nine connectors listed), a description of the guard, an explicit impact statement, and migration guidance. Format and tone match adjacent entries.
- Severity: was Medium; now closed.

**Issue 3: `.xlsx`'s internal ZIP structure isn't covered by an outer-file-size guard (new, not carried over — sorry for missing this nuance earlier, since even a within-limit `.xlsx` is itself compressed).**
- Location: `ExcelReadStrategy.java:274-291` (`assertPoiFileSize`) and the general design of the guard.
- Problem description: the guard bounds the size of the file (or archive entry) as observed from outside, but a `.xlsx` file is itself a ZIP container whose internal parts POI decompresses when building the workbook. A file entirely within the configured limit on disk could still be adversarially crafted to expand substantially in memory during internal unpacking.
- Potential risk: in an environment where Excel file sources are exposed to untrusted or externally-supplied files, a maliciously crafted `.xlsx` within the size limit could still cause elevated memory pressure, though POI 4.1.2's own `ZipSecureFile` default protections bound the worst case to roughly 100x the input size.
- Best improvement: none required for this PR's stated scope. A future follow-up could expose/tighten `ZipSecureFile.setMinInflateRatio()`/`setMaxEntrySize()` explicitly, but that's a materially larger-scope change.
- Severity: Low, non-blocking.
- Raised by another reviewer: No — new this round.

Exception usage is otherwise solid: consistent with existing `CommonErrorCodeDeprecated`/`FileConnectorException` conventions, messages include the offending file/entry name plus both observed and configured byte counts formatted with `%,d`, and concrete remediation is given. No sensitive data is logged. `getPoiExcelMaxFileSize()` additionally validates the configured value is `> 0` and throws a clear `ILLEGAL_ARGUMENT` error otherwise — a good defensive touch that prevents a silently-disabled guard from a misconfigured `0` or negative value.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The refactor of the monolithic `readProcess` into `readByPoi`/`readByEasyExcel`/`getExcelEngine`/`assertPoiFileSize`/`getPoiExcelMaxFileSize` is a clean, minimal-footprint extraction that preserves all pre-existing POI/EasyExcel logic while making the new guard easy to locate and reason about — diffed the extracted bodies against the pre-PR monolithic method and found no incidental logic changes beyond the guard itself.

**Sorry for missing this in an earlier review — carryover, not new.** None of the five new/changed private methods in `ExcelReadStrategy.java` (`getExcelEngine()` at line 145, `readByEasyExcel(...)` at line 162, `readByPoi(...)` at line 182, `assertPoiFileSize(...)` at line 274, `getPoiExcelMaxFileSize()` at line 294) or the two new methods in `AbstractReadStrategy.java` (`assertArchiveEntrySize(...)` at line 553, the `copyInputStream(InputStream, long)` overload at line 650) carry a Javadoc comment describing their purpose, parameters, or the guard's intent. `readByPoi` does have a good inline block comment explaining why the archived-path check is skipped, but that's the only one of the seven new methods with any explanatory comment at all — the size-computation and validation logic in `getPoiExcelMaxFileSize`/`assertPoiFileSize`/`assertArchiveEntrySize`, exactly the "why does this exist, what does it protect against" logic this repo's contribution guide calls out for important methods, has none. This should have been flagged in an earlier round given how central these methods are to the PR's stated purpose; raising it now as a carryover, not a new-in-this-round defect.

Non-blocking maintainability nit, reconfirmed unaddressed: `FileBaseOptions.POI_EXCEL_MAX_FILE_SIZE`/`DEFAULT_POI_EXCEL_MAX_FILE_SIZE` and `FileBaseSourceOptions.POI_EXCEL_MAX_FILE_SIZE`/`DEFAULT_POI_EXCEL_MAX_FILE_SIZE` are declared independently with identical values; only the `FileBaseSourceOptions` copy is ever referenced by any factory or by `ExcelReadStrategy` — the `FileBaseOptions` copy is genuinely dead code. This mirrors the pre-existing duplication already present for `excel_engine` before this PR, so it is not a new anti-pattern; flagged as non-blocking in round 3 and remains unaddressed — consistent, not regressed.

Six Excel-guard-relevant tests now exist, including — new this round — `testPoiAcceptsSmallExcelEntryInsideLargerArchive` (`ExcelReadStrategyTest.java:373-442`), the discriminating regression test round 3 asked for. This closes the coverage gap round 3 identified (the pre-existing `e2e_in_zip.zip` fixture couldn't distinguish "checked the entry's real size" from "checked the archive's on-disk size" because both were similar in magnitude; the new test deliberately pads the archive to 4x the limit while keeping the entry under it).

**Flaky-pattern check (mandatory since test code changed):** all new/changed tests use fixed or generated-in-`@Test` classpath/temp-file resources, a local-filesystem backend, and deterministic in-memory assertions. The new padded-archive test creates a `File.createTempFile(...)` with `deleteOnExit()`, uses STORED (uncompressed) padding to make the archive's on-disk size deterministic and independent of compression-ratio variance across JDKs/zlib versions, and asserts exact byte-count relationships before proceeding — so the test cannot silently degenerate into a non-discriminating case the way the original fixture did. No sleeps, no thread timing, no network/external services, no shared mutable state across tests, no ordering dependency between test methods. **Stability rating: Stable.**

Confirmed by real CI, not just local reasoning: checked the fork's (`zhangshenghang/seatunnel`) actual check-runs for commit `8836f9310b` and confirmed `unit-test` passed on JDK 8 and 11, both Ubuntu and Windows; `connector-file-local-it` and `connector-file-sftp-it` (the relevant integration-test jobs for this connector) passed on both JDKs; `Code style` and `License header` passed; unrelated connector IT jobs were correctly skipped as unaffected by this change. A real green run, not a stub.

## 2.3 Documentation Updates

`docs/en` and `docs/zh` are both updated, consistently, for all nine affected connectors (`LocalFile`, `HdfsFile`, `S3File`, `FtpFile`, `SftpFile`, `OssFile`, `OssJindoFile`, `ObsFile`, `CosFile`), each adding a `poi_excel_max_file_size` option-table row plus prose documenting the 50 MB default and the EasyExcel escape hatch. Spot-checked `LocalFile.md` (en/zh) and `HdfsFile.md` (en) directly: the documented default `52428800` matches `FileBaseSourceOptions.DEFAULT_POI_EXCEL_MAX_FILE_SIZE = 50L * 1024L * 1024L` exactly, and the prose accurately describes when the option applies. English and Chinese content is aligned in wording and structure. The incompatible-changes doc gap (Issue 2) is now closed with matching en/zh entries.

# 3. Architectural Soundness

Precise fix, not a workaround. The core idea — check cheap metadata before an expensive, unbounded operation — is the right shape for this kind of guard, and both the non-archived and (as of this round) the archived implementation execute it correctly and non-redundantly. Confirmed other format readers (CSV/text/JSON/etc.) that call `resolveArchiveCompressedInputStream` still use the original 4-arg overload (`maxBytesForEntry` defaults to `-1`, i.e., unbounded, preserving pre-PR behavior), so this PR's blast radius is correctly confined to Excel. The archive-level and file-level guards now compose correctly instead of talking past each other, which was the one remaining blemish as of round 3.

## 3.2 Maintainability

The extraction into named private methods makes the guard's intent legible and testable in isolation — a genuine maintainability win over the prior single large `readProcess` method. The missing Javadoc on the new guard methods (Issue 4) is a maintainability cost that should be closed before merge, given how load-bearing these methods are to the PR's own stated purpose.

## 3.3 Extensibility

The archive-entry guard pattern is implemented generically at the `AbstractReadStrategy` base-class level and only activated via the new `maxBytesForEntry`-carrying overload, so other format-specific read strategies with similar unbounded-materialization risk could opt into the same bounded-copy mechanism later with a small, additive change — genuinely reusable infrastructure.

The new options are brand-new/newly-propagated in this same unreleased PR, so there is no released-version option-compatibility break to manage for the options themselves. The one genuine compatibility concern is the new default behavior for existing POI-engine jobs over 50 MB — a legitimate, disclosed trade-off that now meets this repo's own documentation bar via the incompatible-changes doc entry.

# 4. Issue Summary

| 1 | POI size guard measured the outer archive's on-disk size instead of the actual Excel entry's size for archived reads, causing false rejections | `ExcelReadStrategy.java:182-207` | High | Fixed — verified via code trace + regression test |
| 2 | New default 50 MB POI cap changed existing behavior but wasn't recorded in the incompatible-changes doc | `docs/en(/zh)/introduction/concepts/incompatible-changes.md` | Medium | Fixed — verified |
| 3 | Outer-file-size guard doesn't bound `.xlsx`'s own internal ZIP-structure decompression (partially mitigated by POI 4.1.2's `ZipSecureFile` defaults) | `ExcelReadStrategy.java:274-291` | Low | Open, non-blocking |
| 4 | New guard methods lack Javadoc despite being central to this PR's stated purpose | `ExcelReadStrategy.java:145,162,182,274,294`; `AbstractReadStrategy.java:553,650` | Medium | Open, should fix before merge |
| 5 | `poi_excel_max_file_size` option constant/default declared independently in `FileBaseOptions` and `FileBaseSourceOptions`; only the latter is referenced | `FileBaseOptions.java` / `FileBaseSourceOptions.java` | Low | Open, non-blocking carryover |

# 5. Merge Recommendation

Both blocking issues from round 3 are genuinely fixed at the current head, `8836f9310bec2be8d88e82009f303cd3799b0d9d` — verified independently against the actual code and test diff rather than trusting the commit message. CI on this exact commit is green, verified against the fork's real workflow run (not just the Apache-side pointer check).

   - Issue 4: add Javadoc to the new guard methods (`getExcelEngine`, `readByEasyExcel`, `readByPoi`, `assertPoiFileSize`, `getPoiExcelMaxFileSize` in `ExcelReadStrategy.java`; `assertArchiveEntrySize` and the new `copyInputStream` overload in `AbstractReadStrategy.java`), per this repo's contribution standard for important/complex-logic methods. A quick, low-risk documentation fix with no behavior change.

   - Issue 3: consider a follow-up to bound `.xlsx`'s internal decompression more explicitly (e.g. tuning `ZipSecureFile` thresholds), though POI 4.1.2's own defaults already provide a meaningful backstop.
   - Issue 5: de-duplicate the `poi_excel_max_file_size` constant/option declaration when convenient.

**Overall assessment:** this PR's engineering judgment is correct throughout — fail-fast beats OOM, the guard is placed before the expensive operation on every reachable path, and the archived-path bug that was the sole real defect after three rounds is now closed with both a code fix and a test that would have caught its absence. What's left is a documentation completeness item (Issue 4) that I should have raised earlier myself, plus two informational, non-blocking notes. This is close to done — the Javadoc addition is the only item I'd ask to see before merge.

DanielLeens (approved):
Re-reviewed the latest head from scratch, as required after a head change. This is round 5.

I diffed the current head (`321958e253e4`) against the commit I fully reviewed in round 4 (`8836f9310bec`): the only change is 56 added lines, and every single one of them is a Javadoc comment block — `git diff` confirms zero executable-code lines changed in `ExcelReadStrategy.java` or `AbstractReadStrategy.java`. So this round is a focused, verifiable fix for exactly the one blocker I left open last time.

**Issue 4 from round 4 (Medium, "new guard methods lack Javadoc") — fixed, verified.** All seven methods I flagged now have proper Javadoc:
- `ExcelReadStrategy.java`: `getExcelEngine()` (line 145), `readByEasyExcel(...)` (line 162), `readByPoi(...)` (line 182), `assertPoiFileSize(...)` (line 274), `getPoiExcelMaxFileSize()` (line 294) — each now documents purpose, `@param`s, `@return` where applicable, and the specific exception it throws and why.
- `AbstractReadStrategy.java`: `assertArchiveEntrySize(...)` (line 553) and the `copyInputStream(InputStream, long)` overload (line 654) — both now documented, including the "non-positive values disable the limit" contract that matters for correctly reading the guard logic.

The commit title (`[Docs][Connector-V2] Document Excel reader guard methods`) matches the actual diff content — this is exactly the fix requested, nothing more, nothing less.

Since the code itself is unchanged from round 4, the rest of that review still holds: the archived-path guard correctly uses the entry-level check instead of re-statting the outer archive (round 3's bug, closed); the EasyExcel path is correctly exempt from the POI limit; the bounded streaming copy in `copyInputStream` protects against declared-size spoofing for ZIP/TAR entries; the incompatible-changes doc entries are present in both `docs/en` and `docs/zh`; and the new `testPoiAcceptsSmallExcelEntryInsideLargerArchive` regression test is a genuinely discriminating test (fails without the fix, passes with it), not a superficial addition.

# Merge Recommendation

1. **Blockers — must be fixed:** none remaining. The sole blocker from round 4 (missing Javadoc on the new guard methods) is fixed and verified above.

2. **Recommended fixes — non-blocking (carried over, unchanged):**
   - The outer-file-size guard doesn't bound `.xlsx`'s own internal ZIP-structure decompression — partially mitigated by this repo's pinned POI 4.1.2 `ZipSecureFile` defaults (min inflate ratio 1%, max entry size ~4GB). Out of scope for this PR's stated purpose; a reasonable follow-up if ever needed.
   - `poi_excel_max_file_size`'s constant/default is still declared independently in both `FileBaseOptions` and `FileBaseSourceOptions`, with only the latter actually referenced — pre-existing duplication pattern (mirrors `excel_engine`), not introduced by this PR, safe to leave for a later cleanup.

**Overall assessment:** four rounds of review found real bugs (the archived-path false-rejection being the most significant), and each was closed with a genuine code fix plus a test that actually discriminates the fixed behavior from the buggy one — not just a green re-run. This round's fix is a clean, low-risk, comment-only change that closes the last open item. Nice work sticking with this through the review cycles.

CI note: the fork's `Build` run for this exact head (`321958e253e4`) was still in progress at the time of this review; the immediately preceding commit's `Build` (`8836f9310b`) passed cleanly on both JDK 8/11 and Ubuntu/Windows, and this round's change is comment-only, so I'd expect the same result, but please let it finish before merging.

davidzollo (approved):
+1 if CI passes.
LGTM

PR #11592, opened 2026-07-29
[Fix][Zeta] Fix stale worker slot profile after release

zhangshenghang:
### Purpose
Refresh worker profiles after slot release completes so the resource manager does not keep stale assigned slots. Without the final refresh, concurrent releases from the same worker can return out-of-order worker snapshots and overwrite the latest resource view with an older one.

### What changed
- Batch release now waits for all release operations and then syncs affected workers with `SyncWorkerProfileOperation`.
- Single slot release refreshes the affected worker after the release operation completes.
- `ReleaseSlotOperation` no longer returns a worker profile snapshot that may become stale before the master consumes it.
- Added unit coverage for batch and single slot release refresh behavior.

### Tests
- `./mvnw -pl seatunnel-engine/seatunnel-engine-server -Dmaven.gitcommitid.skip=true -DskipITs -Dtest=ResourceReleaseRefreshTest test`
- `./mvnw -o -pl seatunnel-engine/seatunnel-engine-server -Dmaven.gitcommitid.skip=true -DskipITs -Dtest=ResourceReleaseRefreshTest,ResourceManagerTest,FixSlotResourceTest test`
- `./mvnw -o -pl seatunnel-engine/seatunnel-engine-server -Dmaven.gitcommitid.skip=true -DskipTests verify`

No E2E was added because this is an internal Zeta resource-manager state refresh fix with deterministic unit coverage.


DanielLeens (commented):
Thanks for working on this. I reviewed the current head from the Zeta resource-release path rather than only the changed method.

What this PR fixes:
- User pain: after a task group or pipeline releases slots, the master-side `registerWorker` cache can still observe the old worker slot snapshot.
- Fix approach: `ReleaseSlotOperation` now only performs the worker-side slot release, and the master explicitly follows up with `SyncWorkerProfileOperation` before refreshing `registerWorker` through `heartbeat()`.
- In one sentence: this makes the release path refresh the master-side worker profile from the worker after the release has actually completed.

Runtime path I traced:
```text
JobMaster resource release
  -> releaseTaskGroupResource() / releasePipelineResource() / resource-apply rollback
      -> ResourceManager.releaseResources(jobId, slotProfiles)
          -> send ReleaseSlotOperation for each SlotProfile
              -> worker DefaultSlotService.releaseSlot() [203-230]
                  -> validate slotID / sequence / ownerJobID
                  -> move the slot out of assignedSlots
          -> after all release futures complete
          -> refreshWorkerProfiles(releaseWorkers)
              -> SyncWorkerProfileOperation.run() [34-40]
                  -> SlotService.getWorkerProfile() [303-311]
              -> heartbeat(workerProfile) [375-384]
                  -> update master-side registerWorker
```

I did not find a source-level correctness blocker in the release semantics. The one thing I would like to fix before merge is documentation of the new core timing contract:

1. `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/resourcemanager/AbstractResourceManager.java:283`

   `refreshWorkerProfiles()` is now part of the Zeta resource-release critical path. It defines the two-step contract: release the slot first, then sync the worker profile and update the master cache. It also changes the failure boundary: a slot may have been released successfully, but the outer release future still fails if the follow-up profile sync fails.

   Could you please add a short Javadoc/comment explaining why the refresh is done after release, why batch release deduplicates by worker, and that refresh failures are intentionally propagated to the release future? This will make the lifecycle contract much safer for future maintenance.

Test coverage looks focused: the new unit test covers both single-slot and same-worker batch release refresh, uses synchronous completed futures, and I did not see a flaky-test risk such as sleeps, timing assertions, external ports, or ordering assumptions.

Merge conclusion:
- Source verdict: fix the comment above, then the source side looks good to me.
- CI/gate: the required Build is still running/queued on the current head, so final merge still needs the latest GitHub CI result plus maintainer review.

DanielLeens:
Thanks for the update. I rechecked the current head on August 3, 2026.

The CI fact I referenced in my previous review is now outdated: the `Build` check is no longer running/queued on this head. It is now ending in a cancelled state, and the contributor run behind that check currently shows `Run / kudu-connector-it (11, ubuntu-latest)` as the relevant cancelled lane: https://github.com/zhangshenghang/seatunnel/actions/runs/30432385889/job/90512890485 . That does not line up directly with this PR's current diff, which is limited to three `seatunnel-engine` files under the Zeta resource manager path.

This branch is also behind `dev` (`behind_by=29`, compare status: `diverged`), and latest `dev` is already ahead with relevant upstream fixes such as `4c4b81859b09` (`[Fix][Connector-V2] Wait for Kudu client shutdown`) and `81d69d83bcc3` (`[CI][E2E] Stabilize hosted connector IT failures`). So for the CI side, the lowest-cost next step is to sync with the latest `dev` and rerun CI first.

Separately, my previous source-side blocker is still open on this same head: the lifecycle contract around `refreshWorkerProfiles()` in `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/resourcemanager/AbstractResourceManager.java:283` still needs an explicit Javadoc/comment before merge. So at the moment this PR is blocked by both that source-side documentation item and the current upstream-sync / CI gate.

DanielLeens (commented):
Thanks for pushing the Javadoc update — I re-reviewed this from scratch against the current head (`76e6fb2`), not just the incremental diff since my last comment.

First, closing the loop on my two previous items from August 3 / July 29:
- My source-side blocker (add a Javadoc to `refreshWorkerProfiles()` explaining the timing contract) is now resolved. The Javadoc at `AbstractResourceManager.java:283-295` clearly states the "release-then-sync" ordering, the per-worker dedup rationale, and the intentional failure-propagation behavior. Good addition.
- My CI/staleness note from August 3 is now worse, not better: the branch was `behind_by=29` on dev then, it is `behind_by=45` now, and the `Build` check on the current head (`76e6fb2`) is still sitting in `queued` state in the contributor's fork (run `31326696466`, queued for 3+ hours at the time of this review) — it has not actually completed. The commit immediately before the spotless fix (`1ae95a1`) actually failed CI on `Run / Code style`, which `76e6fb2` fixed, but that means the current head has never had a green `Build` run yet. I did not see any new activity from other reviewers since my last comment.

Given that, this review focuses on validating the source-level fix is correct and complete, independent of the CI question.

When a job releases multiple slots on the same worker in one `releaseResources()` batch (e.g. a pipeline/task-group finishing and freeing several slots at once), the pre-fix code issued one `ReleaseSlotOperation` per slot in parallel, and each operation carried its own worker-profile snapshot back as its RPC response (`ReleaseSlotOperation.getResponse()` returning `WorkerProfile`). Each response independently triggered `heartbeat(profile)` on completion. Because the master has no control over the order in which these N independent futures complete, a snapshot taken *before* a sibling slot on the same worker was released could land at the master *after* a snapshot that already reflects that release, and `heartbeat()` does a blind `registerWorker.put(...)`. The later-arriving-but-older snapshot silently overwrites the newer one, leaving the master believing a slot is still assigned when the worker has already freed it — a stale worker profile that can persist until the next periodic heartbeat "self-heals" it.

The fix decouples "release the slot" from "learn the worker's post-release state": `ReleaseSlotOperation` no longer returns any profile at all, and after *all* release operations targeting a batch have completed, the master issues exactly one dedicated `SyncWorkerProfileOperation` per distinct worker to fetch a single, guaranteed-fresh snapshot, which is the only thing applied via `heartbeat()`.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Before** (`AbstractResourceManager.releaseResource`, prior to this PR):
```java
CompletableFuture<WorkerProfile> future =
        sendToMember(new ReleaseSlotOperation(jobId, profile), profile.getWorker());
return future.thenAccept(this::heartbeat);
```
`ReleaseSlotOperation.runInternal()` did the release and then read `getWorkerProfile()` in the same call, wiring the resulting profile out as the operation's `getResponse()`.

**After** (current head):
```java
private CompletableFuture<Void> releaseResource(
        long jobId, SlotProfile profile, boolean refreshWorkerProfile) {
    if (nodeEngine.getClusterService().getMember(profile.getWorker()) != null) {
        CompletableFuture<Object> releaseFuture =
                sendToMember(new ReleaseSlotOperation(jobId, profile), profile.getWorker());
        CompletableFuture<Void> result = new CompletableFuture<>();
        releaseFuture.whenComplete((unused, error) -> {
            if (error != null) { result.completeExceptionally(error); return; }
            if (!refreshWorkerProfile) { result.complete(null); return; }
            refreshWorkerProfiles(Collections.singleton(profile.getWorker()))
                    .whenComplete((refreshUnused, refreshError) -> { ... });
        });
        return result;
    }
    return CompletableFuture.completedFuture(null);
}
```
and `releaseResources()` now: (1) computes `releaseWorkers`, the distinct set of live-member workers touched by the batch, (2) fires all `releaseResource(jobId, profile, false)` calls (no per-slot refresh), (3) waits for `CompletableFuture.allOf(...)`, and only then calls `refreshWorkerProfiles(releaseWorkers)` once per distinct worker. `ReleaseSlotOperation` was stripped of its `WorkerProfile result` field and `getResponse()` override entirely — it is now a pure "do the release" operation with no return payload.

**Key Findings**
- The dedup/ordering guarantee is real, not incidental: `DefaultSlotService.releaseSlot()` and `DefaultSlotService.getWorkerProfile()` are both `synchronized` on the same worker-side `DefaultSlotService` instance (`DefaultSlotService.java:203`, `:303`). Because the master only dispatches `SyncWorkerProfileOperation` after *every* release RPC in the batch has already returned a response, and because a Hazelcast operation's response is only sent after `runInternal()`/`run()` (and thus the synchronized block) has fully executed, the follow-up `getWorkerProfile()` is guaranteed to observe all the releases in that batch. This is a correct fix for the specific race it targets, closed by causal ordering + mutual exclusion, not by luck.
- `releaseResources()` now does the release-worker computation (`getMember(address) != null`) once, but `releaseResource()` independently re-checks `getMember(profile.getWorker()) != null` for the same purpose. Functionally consistent (a departed worker is excluded from both the release attempt and the refresh set), just duplicated — see Issue 4.
- Retry-safety check: `JobMaster.releaseTaskGroupResource()`/`releasePipelineResource()` call `releaseResources(...).join()` inside `RetryUtils.retryWithException(...)`. If the new follow-up refresh RPC fails (propagated per the new Javadoc's documented contract) and the whole `releaseResources()` call is retried, the retried `ReleaseSlotOperation` targets an *already-released* slot. I verified this is safe: `DefaultSlotService.releaseSlot()` throws `WrongTargetSlotException` for a slot no longer in `assignedSlots`, and `ReleaseSlotOperation.runInternal()` still catches and logs that exception without propagating it (`ReleaseSlotOperation.java:49-56`), so the retry degrades to a harmless no-op release + a fresh (and accurate) re-sync. Good — the "propagate refresh failures" design choice documented in the new Javadoc does not create a retry-amplification bug.
- Normal-path confirmation: this is not a recovery/failover-only code path. `JobMaster.releaseTaskGroupResource()` (line 866) and `releasePipelineResource()` (line 910) are invoked on every ordinary task-group/pipeline completion, and both block synchronously on `releaseResources(...).join()`. Every job run touches this code, not just failure recovery.

**Runtime path (text flow diagram, thread boundaries marked)**
```
[Master, caller thread] JobMaster.releaseTaskGroupResource()/releasePipelineResource()
    blocks on .join()
        |
        v
[Master, caller thread] AbstractResourceManager.releaseResources(jobId, profiles)
    releaseWorkers = distinct live-member workers among profiles
        |
        |-- for each profile: releaseResource(jobId, profile, refreshWorkerProfile=false)
        |       sendToMember(ReleaseSlotOperation) --RPC--> [Worker, Hazelcast op thread]
        |                                                        ReleaseSlotOperation.runInternal()
        |                                                          synchronized releaseSlot():
        |                                                            validate slotID/sequence/ownerJobID
        |                                                            move slot: assignedSlots -> unassignedSlots
        |       <-- RPC response (no WorkerProfile payload) --------|
        |       [Master, completion-listener thread] whenComplete: refreshWorkerProfile==false -> just complete
        |
        v  (after CompletableFuture.allOf(all per-slot release futures))
    [Master, completion-listener thread] refreshWorkerProfiles(releaseWorkers)
        for each DISTINCT worker:
            sendToMember(SyncWorkerProfileOperation) --RPC--> [Worker, Hazelcast op thread]
                                                                   synchronized getWorkerProfile()
                                                                   (sees every release above,
                                                                    guaranteed by same-monitor +
                                                                    response-after-release causality)
            <-- RPC response: one fresh, post-release snapshot ---|
        [Master, completion-listener thread] heartbeat(freshProfile)
            registerWorker.put(address, freshProfile)   <- single, deduplicated, authoritative write

Concurrent/independent paths this fix does NOT serialize against (see Issue 1):
[Worker, background scheduled-executor thread] periodic WorkerHeartbeatOperation(currentProfile)
    --RPC--> [Master] heartbeat(currentProfile)   <- independent writer, no version/epoch check

[Master] ResourceRequestHandler.singleResourceRequestToMember() (allocate path, untouched by this PR)
    --RequestSlotOperation--> [Worker] returns SlotAndWorkerProfile (per-operation snapshot)
    <-- resourceManager.heartbeat(slotAndWorkerProfile.getWorkerProfile()) -- same pre-fix anti-pattern
```

## 1.2 Compatibility Impact

- **Protocol between master and worker**: `SyncWorkerProfileOperation` is a pre-existing operation (already used by `initWorker()` and already registered in `ResourceDataSerializerHook` before this PR), so no new serialization type is introduced. The only wire-observable change is that `ReleaseSlotOperation`'s RPC response is now `null` (the base `Operation.getResponse()` default) instead of a `WorkerProfile`.
- For a cluster where master and worker run the **same** build (the officially supported deployment model for a Zeta cluster — SeaTunnel does not document a supported skewed-version rolling upgrade for master/worker binaries), this is fully compatible.
- For a hypothetical mixed-version window (old master binary + new worker binary), there is a latent NPE risk: the old master's `releaseResource()` does `future.thenAccept(this::heartbeat)`, and `heartbeat(WorkerProfile workerProfile)` immediately dereferences `workerProfile.getAddress()`. Against a new worker, `future` would resolve to `null`, so `heartbeat(null)` throws `NullPointerException`. I'm flagging this as Issue 3 below — low real-world likelihood given how Zeta clusters are normally upgraded, but worth a one-line note since it's an explicit "protocol" question and the change is silent about it.
- No serialization format changes reach checkpoint/savepoint state, job DAG state, or client-facing config — this is purely an internal engine-to-engine RPC change, so historical running-job resumption is unaffected.

- Every slot release now costs one **additional sequential RPC round trip** per distinct worker (release RPCs still run in parallel, but the refresh RPC is strictly ordered after all of them complete, not overlapped). Since `releaseTaskGroupResource`/`releasePipelineResource` block synchronously via `.join()` on the normal task-completion path, this adds one RPC's worth of latency to every task-group/pipeline finish. This is a bounded, small, and justified trade-off for correctness, but worth being aware of on high-churn clusters with many short-lived jobs.
- No new lock is introduced on the master side; all master-side coordination is via `CompletableFuture` composition, not blocking locks, so there is no new deadlock risk. On the worker side, `releaseSlot()` and `getWorkerProfile()` already shared the same monitor before this PR; this PR does not change worker-side locking, it only changes *when* the master asks for a snapshot.
- **Issue 1 (residual race, not closed by this PR)**: the fix removes the out-of-order-snapshot race for a single `releaseResources()` batch, but the master-side `registerWorker` cache still has no version/epoch/timestamp, so `heartbeat()` remains a blind last-write-wins. Two *independent* triggers can still race each other on the same worker with no ordering guarantee: (a) this PR's own `refreshWorkerProfiles()` racing against the worker's periodic `WorkerHeartbeatOperation` (`DefaultSlotService.java:146`, sent from a dedicated `ScheduledExecutorService`), and (b) two concurrent `releaseResources()` calls for *different* jobs landing on the same worker, each computing its own independent refresh. Separately, the exact same per-operation-snapshot-plus-independent-heartbeat pattern that caused the original bug is still present, unmodified, in the allocate path: `ResourceRequestHandler.singleResourceRequestToMember()` (`ResourceRequestHandler.java:253`) calls `resourceManager.heartbeat(slotAndWorkerProfile.getWorkerProfile())` per `RequestSlotOperation` response, with no batch-level dedup/reordering-proofing. A job requesting several slots that happen to land on the same worker can hit essentially the same stale-profile bug this PR fixes for release, just on the allocate side.
  - I don't think this should block this PR — it's a different, pre-existing code path outside this diff's stated scope, and a full fix (e.g. a monotonic version/epoch on `WorkerProfile`, or single-flight refresh per worker) is a larger design change than a targeted bugfix warrants. But it's worth a follow-up issue so the same bug class doesn't quietly persist on the allocate side.

- The `ReleaseSlotOperation` log line for `WrongTargetSlotException` was trimmed from `"...exception: {}", jobID, slotProfile, ExceptionUtils.getMessage(ignore)` to `"...", jobID, slotProfile` (two placeholders, two args — consistent, no format-string bug). Dropping the exception message loses a small amount of debug detail but the job/slot context that actually matters for triage is retained. Not a concern.
- **Issue 3** (see table): mixed-version `heartbeat(null)` NPE risk described in 1.2, currently undocumented and unguarded.
- **Issue 5**: the retry-then-fail path in `JobMaster.releaseTaskGroupResource`/`releasePipelineResource` swallows all exceptions with `catch (Exception e) { LOGGER.warning(...) }` and no metric/counter. This pre-dates this PR, but this PR adds a new failure surface (the refresh RPC) that can trip this catch block more often than before under transient network conditions. Not a regression this PR needs to fix, but worth a passing mention since the new Javadoc explicitly calls out "surfacing the failure lets the outer release future fail loudly" — today, that "loud" failure is only a `LOGGER.warning`, with nothing observable at the metrics layer.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- `refreshWorkerProfiles()` now has a solid Javadoc (`AbstractResourceManager.java:283-295`) that explains the release-then-sync contract, the per-worker dedup rationale, and the intentional failure propagation — this fully resolves my prior review comment.
- **Issue 2 (medium, missing comment on new core method)**: the new private overload `releaseResource(long jobId, SlotProfile profile, boolean refreshWorkerProfile)` (`AbstractResourceManager.java:334`) has no comment at all, and its boolean parameter's purpose (batch callers pass `false` and do a single dedup'd refresh themselves; the public single-slot `releaseResource(long, SlotProfile)` entry point passes `true` and refreshes immediately) is not obvious from the call sites alone. Per the project's own standard ("Add comments for important methods... lifecycle hooks"), this qualifies — it's exactly the kind of control-flow flag that's easy to get backwards in a future edit. A short one- or two-line comment would close this out.
- Minor style nit (not blocking): the live-member check `nodeEngine.getClusterService().getMember(address) != null` is duplicated between the `releaseWorkers` computation in `releaseResources()` and the guard inside `releaseResource()`. Not a correctness issue, just a small opportunity to de-duplicate if you touch this file again.

`ResourceReleaseRefreshTest` adds two focused unit tests:
- `testReleaseResourcesRefreshesWorkerProfileAfterBatchRelease`: two slots on the same worker, asserts `releaseCalls == 2` and `refreshCalls == 1` — directly verifies the dedup behavior that is the point of this fix.
- `testReleaseResourceRefreshesWorkerProfileAfterSingleRelease`: single-slot path, asserts one release + one refresh.

Both use a `TestResourceManager` subclass overriding `sendToMember()` to return `CompletableFuture.completedFuture(...)` synchronously — no `Thread.sleep`, no timers, no external ports, no ordering assumptions on real threads. **Stability rating: Stable.** I don't see a flaky-test risk here.

One honest gap: because the mocked futures resolve synchronously and in-order, the tests verify the *dedup/sequencing structure* of the fix (exactly one refresh call, issued after both releases) but don't construct an adversarial interleaving (e.g., manually completing the futures out of order) to positively reproduce the original bug and show the old code would have failed this assertion. Given the fix is a structural reordering (refresh is only ever triggered after `CompletableFuture.allOf` of all releases, and worker-side synchronization guarantees visibility as analyzed in 1.1), I don't think an adversarial-ordering test is strictly necessary to trust the fix — but it would make the regression protection more explicit for future refactors of this method. Not a blocker.

I also traced through the existing `ResourceManagerTest.testApplyRequest()` (`releaseResources` called twice for the same profiles, second call expected not to throw) against an embedded real server, and `FixSlotResourceTest` — neither mocks `sendToMember` in a way that would be broken by the new `SyncWorkerProfileOperation` round trip, so I don't expect this change to regress them, though I can't confirm pass/fail without a completed CI run (see intro).

## 2.3 Documentation Updates

No `docs/en`/`docs/zh` changes are included, and I don't think any are needed: this is an internal engine RPC-sequencing fix, not a user-facing config or behavior change. I checked `docs/en/architecture/engine/resource-management.md`'s "6.3 Resource Release" sequence diagram — it's already abstracted at the "RM -> Mark slots unassigned / Update WorkerProfile" level and doesn't describe the `ReleaseSlotOperation`/`SyncWorkerProfileOperation` split, so it isn't rendered inaccurate by this change.

# 3. Architectural Soundness

This is a precise fix, not a workaround: it removes the root cause (per-operation snapshots racing on completion order) rather than papering over symptoms with, say, a delay or a version check. It reuses an existing operation (`SyncWorkerProfileOperation`) rather than inventing new wire types, which keeps the blast radius small. It is not a long-term/complete solution to the broader "master-side worker profile cache has no versioning" problem (see Issue 1), but that's a reasonable scoping decision for a targeted bugfix.

## 3.2 Maintainability

The new Javadoc on `refreshWorkerProfiles()` makes the lifecycle contract legible for future maintainers. The missing comment on the new `releaseResource(..., boolean)` overload (Issue 2) is the main gap left.

## 3.3 Extensibility

The `refreshWorkerProfiles(Collection<Address>)` helper is written generically enough (dedup by worker, batch-friendly) that it could be reused if the allocate-path bug noted in Issue 1 is fixed later using the same pattern.

No impact on checkpoint/savepoint state, job graph serialization, or client-facing configuration. The only compatibility surface is the internal master/worker RPC discussed in 1.2 and Issue 3, which only matters under an unsupported mixed-version deployment.

# 4. Issue Summary

|---|---|---|---|
| 1 | Residual same-worker profile race not fully closed: no version/epoch on `WorkerProfile`/`registerWorker`, so concurrent independent refresh/heartbeat writers can still race; the identical per-operation-snapshot anti-pattern this PR fixes for release is still present, unmodified, in the allocate path | `AbstractResourceManager.heartbeat()`; `ResourceRequestHandler.java:253` | Medium (out of scope for this PR, recommend follow-up) |
| 2 | New private `releaseResource(long, SlotProfile, boolean)` overload has no comment explaining the `refreshWorkerProfile` flag's purpose | `AbstractResourceManager.java:334` | Medium |
| 3 | Mixed-version compatibility: old master + new worker would receive `null` from `ReleaseSlotOperation`'s response and NPE inside `heartbeat(null)` | `ReleaseSlotOperation.java` (response removed); pre-image `AbstractResourceManager.releaseResource` (`future.thenAccept(this::heartbeat)`) | Low (unsupported deployment scenario, but undocumented) |
| 4 | Duplicate `getMember(address) != null` computation in `releaseResources()` and `releaseResource()` | `AbstractResourceManager.java:252-258`, `:336` | Low (style nit) |
| 5 | Refresh-RPC failures surfaced by this PR are only logged via `LOGGER.warning` in the retry catch-all, with no metric/alerting | `JobMaster.java:902-907`, `:944-` | Low |
| — | Branch is `behind_by=45` on `dev`, and the current head's `Build` check has not completed (queued 3+ hours in the fork at review time) | CI/process, not source | Blocking for merge (not a source defect) |

"Raised by another reviewer: No" applies to all rows above — no other reviewer has commented on this PR besides me.

# 5. Merge Recommendation

**Blockers:**
- The current head (`76e6fb2`) has not had a completed, green `Build` run yet — please rebase onto latest `dev` (45 commits behind is a lot to carry) and get a clean CI run before merge, consistent with what I flagged on August 3.
- Please add a short comment to the new `releaseResource(long, SlotProfile, boolean)` overload (Issue 2) explaining the `refreshWorkerProfile` flag, matching the documentation quality you already added to `refreshWorkerProfiles()`.

**Recommended (non-blocking) follow-ups:**
- Consider filing a follow-up issue for Issue 1 — the same stale-worker-profile bug class this PR fixes for release still exists, unmodified, on the allocate path (`ResourceRequestHandler.singleResourceRequestToMember`).
- Consider a brief compatibility note or defensive null-check for Issue 3 if SeaTunnel ever needs to support skewed master/worker versions during a rolling upgrade.

**Overall assessment:** the core fix is correct and provably closes the specific race described in the PR — I traced the worker-side `synchronized` locking and the master-side ordering (refresh only dispatched after `CompletableFuture.allOf` of all releases) and confirmed the "stale snapshot overwrites fresher one" scenario can no longer happen for a single release batch. Test coverage is deterministic and stable, with no sleep-based synchronization anti-patterns. This is a well-scoped, minimal-diff fix to a real production bug on the normal task-completion path. Once the branch is rebased and CI is green, and the small doc gap is closed, I'm comfortable with this going in.

DanielLeens (commented):

When a job releases multiple slots on the same worker in one `releaseResources()` batch (e.g. a pipeline/task-group finishing and freeing several slots at once), the pre-fix code issued one `ReleaseSlotOperation` per slot in parallel, and each operation independently carried its own worker-profile snapshot back as its RPC response, which independently triggered `heartbeat(profile)` on completion. Because the master has no control over the order in which these N independent futures complete, a snapshot taken *before* a sibling slot's release could land at the master *after* a snapshot that already reflects that release, and `heartbeat()` does a blind `registerWorker.put(...)` — the later-arriving-but-older snapshot silently overwrites the newer one, leaving the master believing a slot is still assigned when the worker has already freed it. The fix decouples "release the slot" from "learn the worker's post-release state": `ReleaseSlotOperation` no longer returns a profile at all, and after *all* release operations in a batch complete, the master issues exactly one dedicated `SyncWorkerProfileOperation` per distinct worker.

I re-read the complete current-head diff against `dev` from scratch, not just the new commit, as a re-review requires. No other reviewer has commented on this PR since my Aug 9 review — checked both the review-comment and issue-comment threads, empty.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**What changed since my Aug 9 review**: a merge from `dev` (`77ecf6fdfcc2`, bringing the branch from `behind_by=45` down to `behind_by=2`) followed by one new commit, `a40df8fa767e` ("[Improve][Zeta] Document refreshWorkerProfile contract in slot release").

**Issue 2 (missing comment on the new private overload) — fixed, and it's exactly the comment I asked for:**
```java
// AbstractResourceManager.java:334-344 — new
/**
 * Releases a single slot on its owning worker.
 *
 * <p>The {@code refreshWorkerProfile} flag controls whether the master-side worker profile
 * cache is refreshed by this call. The single-slot entry point passes {@code true} so the
 * profile sync runs immediately after the slot release completes; the batch path {@link
 * #releaseResources(long, List)} passes {@code false} and refreshes each involved worker once
 * after all releases finish, instead of syncing per slot. When refreshing, a profile-sync
 * failure is propagated to the returned future even though the slot itself has already been
 * released (see {@link #refreshWorkerProfiles(Collection)}).
 */
private CompletableFuture<Void> releaseResource(
        long jobId, SlotProfile profile, boolean refreshWorkerProfile) {
```
This closes exactly the gap I flagged: the `refreshWorkerProfile` boolean's meaning (single-slot caller passes `true` and refreshes immediately; batch caller passes `false` and does a single dedup'd refresh itself) is now spelled out, along with the important and easy-to-miss detail that a refresh failure is surfaced even though the release itself already succeeded — matching the failure-propagation design the sibling `refreshWorkerProfiles()` Javadoc already documented. I diffed this against the actual call sites (`releaseResource(jobId, profile)` at line ~330 passing `true`, and the batch loop in `releaseResources()` passing `false`) and the comment's description matches the real behavior.

**Everything else I verified on Aug 9 is untouched by this merge and still correct** — I re-read `AbstractResourceManager.releaseResources()`/`releaseResource()`, `ReleaseSlotOperation`, and `DefaultSlotService.releaseSlot()`/`getWorkerProfile()` directly against the current worktree rather than assuming the merge left them alone: the worker-side `synchronized` monitor shared between `releaseSlot()` and `getWorkerProfile()` is unchanged, the master only dispatches `SyncWorkerProfileOperation` after `CompletableFuture.allOf(...)` of all release futures, and `ReleaseSlotOperation` still carries no response payload. The race this PR targets remains provably closed by the same causal-ordering + mutual-exclusion argument I traced on Aug 9.

## 1.2 Compatibility Impact

Unchanged from Aug 9: fully compatible for the officially-supported same-build master/worker deployment model. `SyncWorkerProfileOperation` is a pre-existing, already-registered operation type, so no new wire type is introduced.

Unchanged from Aug 9: one additional sequential RPC round trip per distinct worker on every batch release, bounded and justified for correctness.

No new issues this round.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Issue 2 is now closed — see 1.1. No other missing-comment findings.

No test files changed in this commit. `ResourceReleaseRefreshTest`'s two tests (`testReleaseResourcesRefreshesWorkerProfileAfterBatchRelease`, `testReleaseResourceRefreshesWorkerProfileAfterSingleRelease`) are unchanged from Aug 9 — I re-read both to confirm no silent regression from the `dev` merge. **Test stability rating: Stable** (unchanged assessment): synchronous mocked futures, no `Thread.sleep`, no timers, no external ports, no ordering assumptions on real threads.

## 2.3 Documentation Updates

No `docs/en`/`docs/zh` changes, and none needed — unchanged assessment, this remains an internal engine RPC-sequencing fix with no user-facing config or behavior change.

# 3. Architectural Soundness

Unchanged: a precise fix that removes the root cause rather than papering over it, reusing an existing operation type instead of inventing a new wire type.

## 3.2 Maintainability

Improved this round: both the public-facing `refreshWorkerProfiles()` contract (fixed Aug 9) and the internal `releaseResource(..., boolean)` control-flow flag (fixed this round) are now documented, so a future maintainer reading either method understands the release-then-sync ordering without having to reconstruct it from call sites.

## 3.3 Extensibility

Unchanged from Aug 9.

Unchanged: no checkpoint/savepoint, job-graph serialization, or client-facing config impact.

# 4. Issue Summary

| 1 | Residual same-worker profile race not fully closed: no version/epoch on `WorkerProfile`/`registerWorker`, so independent refresh/heartbeat writers can still race each other; the identical per-operation-snapshot anti-pattern this PR fixes for release is still present, unmodified, on the allocate path (`ResourceRequestHandler.singleResourceRequestToMember`) | `AbstractResourceManager.heartbeat()`; `ResourceRequestHandler.java:253` | Medium (out of scope for this PR, recommend follow-up) |
| 2 | Mixed-version compatibility: an old master binary + new worker binary would receive `null` from `ReleaseSlotOperation`'s response and NPE inside `heartbeat(null)` | `ReleaseSlotOperation.java` (response removed) | Low (unsupported deployment scenario, but undocumented) |
| 3 | Duplicate `getMember(address) != null` computation in `releaseResources()` and `releaseResource()` | `AbstractResourceManager.java` | Low (style nit) |

Issues 1-3 are unchanged carryovers from my Aug 9 review (renumbered here since the former Issue 2 — the missing comment — is now resolved), not new findings.

# 5. Merge Recommendation

### Conclusion: Ready to merge after CI clears

   - None on the source-review side. Both items from my Aug 9 review (missing comment on the `releaseResource(..., boolean)` overload, and the branch being 45 commits behind `dev`) are now resolved — the comment is exactly what I asked for, and the branch is down to `behind_by=2` after the merge.
   - CI: the `Build` check on this exact head (`a40df8fa767e`) is still `in_progress` on the apache side, and the fork's own run (`34371249751`) is still `queued` as of this review — it has not yet produced a completed result. Given the history of this branch sitting queued for 3+ hours on Aug 9, please confirm it actually completes green before merging; I can't clear the CI gate without a finished run.

   - Issue 1: consider filing a follow-up issue for the same stale-worker-profile bug class on the allocate path.
   - Issue 2: a brief compatibility note or defensive null-check if SeaTunnel ever needs to support skewed master/worker versions during a rolling upgrade.
   - Issue 3: minor de-duplication opportunity, not urgent.

**Overall assessment**: the core fix has been correct since my July 29 review, and both concrete asks from Aug 9 (documentation of the new control-flow flag, and getting the branch off a 45-commit-stale base) are now delivered exactly as requested. This is a well-scoped, minimal-diff fix to a real production bug on the normal task-completion path, with deterministic, stable test coverage. I have read-only (comment) rights here, so a write-capable maintainer will need to give the final approval/merge once the current CI run actually completes green.

DanielLeens (commented):
CI-fact update: in my last review the apache-side `Build` check was still `in_progress` and the fork's own run was still `queued` with no completed result yet, and I had no source-level blocker at that point. I just rechecked `gh pr checks` on this same head and `Build` has since completed with `pass`. With CI now green and no outstanding source-side blocker from my side, this looks ready pending a maintainer's final review and merge action.

DanielLeens (approved):
Formalizing the conclusion from my last follow-up here: my previous full review landed on "Ready to merge after CI clears" with no blockers on the source-review side -- the only outstanding item was the `Build` check on this exact head (`a40df8fa767e`), which was still `in_progress`/`queued` at that time. I already confirmed in my next comment that it has since completed green, and there have been no new commits since then.

With no remaining source-side blocker from my side and CI confirmed passing, I'm updating this to a formal approval. I have read-only (comment-level) rights on this repo, so a write-capable maintainer will still need to handle the final merge action.

PR #11594, opened 2026-07-29
[Improve][CI] Stabilize workflow run detection

DanielLeens:
### Purpose

This PR stabilizes CI/tooling failures observed on `dev` and PR-notification workflows during the 2026-07-27 to 2026-07-29 window. The failures are unrelated to individual PR code changes.

### Root causes observed

- `On pull request update` can fail by itself when the latest workflow run found by branch does not belong to the current PR head SHA. Example: https://github.com/apache/seatunnel/actions/runs/30420088096 failed with `There was a new unsynced commit pushed. Please retrigger the workflow.`
- `Build` on `dev` can fail while bootstrapping Maven Wrapper when Maven Central returns a transient `429 Too Many Requests`, leaving `maven-wrapper.jar` missing. Example: https://github.com/apache/seatunnel/actions/runs/30428769947
- `Upgrade Compatibility` on `dev` can pick the edge-agent binary archive while looking for the current SeaTunnel distribution, then fail because the extracted directory does not contain `bin/seatunnel.sh`. Example: https://github.com/apache/seatunnel/actions/runs/30388071330 failed with `Missing required file: .../target/upgrade-compatibility/current/bin/seatunnel.sh`.

### Changes

- Match PR notification Build workflow runs by the exact PR head SHA and create an `action_required` check instead of failing the notification workflow when the run cannot be detected.
- Retry Maven Wrapper jar downloads for transient network/rate-limit failures and clean partial jar files between attempts.
- Exclude `apache-seatunnel-edge-agent-*-bin.tar.gz` when the upgrade compatibility script auto-detects the current SeaTunnel distribution archive.

### Verification

Local verification was intentionally limited to formatting and static checks per the SeaTunnel local-validation rule for this task. I did not run Docker, compile, tests, E2E, packaging, services, or jobs locally.

Executed:

- `sh -n mvnw`
- `bash -n tools/upgrade_compatibility/run_upgrade_compatibility.sh`
- YAML parse check for `.github/workflows/backend.yml` and `.github/workflows/notify_test_workflow.yml`
- Extracted `actions/github-script` body and ran `node --check`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home ./mvnw -nsu -Dmaven.gitcommitid.skip=true spotless:apply`
- `git diff --check`

The final functional gate should be the GitHub CI for this PR head.

### DIFF
--- a/.github/workflows/notify_test_workflow.yml
+++ b/.github/workflows/notify_test_workflow.yml
@@ -48,6 +48,7 @@ jobs:
             // TODO: Should use pull_request.user and pull_request.user.repos_url?
             // If a different person creates a commit to another forked repo,
             // it wouldn't be able to detect.
+            const head_sha = context.payload.pull_request.head.sha
             const params = {
               owner: context.payload.pull_request.head.repo.owner.login,
               repo: context.payload.pull_request.head.repo.name,
@@ -66,8 +67,36 @@ jobs:
             console.log('SHA: ' + context.payload.pull_request.head.sha)
 
             const name = 'Build'
-            const head_sha = context.payload.pull_request.head.sha
-            let status = 'queued'
+            const createDetectionFailedCheck = async () => {
+              await github.rest.checks.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                name: name,
+                head_sha: head_sha,
+                status: 'completed',
+                conclusion: 'action_required',
+                output: {
+                  title: 'Workflow run detection failed',
+                  summary: `
+            Unable to detect the workflow run for testing the changes in your PR.
+
+            1. If you did not enable GitHub Actions in your forked repository, please enable it by clicking the button as shown in the image below. See also [Disabling or limiting GitHub Actions for a repository](https://docs.github.com/en/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository) for more details.
+            2. Create and push an empty commit to trigger the workflow.
+            3. It is possible your branch is based on the old \`dev\` branch in Apache SeaTunnel, please sync your branch to the latest dev branch. For example as below:
+                \`\`\`bash
+                git fetch upstream
+                git rebase upstream/dev
+                git push origin YOUR_BRANCH --force
+                \`\`\``,
+                  images: [
+                    {
+                      alt: 'enabling workflows button',
+                      image_url: 'https://raw.githubusercontent.com/apache/spark/master/.github/workflows/images/workflow-enable-button.png'
+                    }
+                  ]
+                }
+              })
+            }
 
             async function findBuildWorkflowRun() {
               const lookupDeadline = Date.now() + MAX_LOOKUP_DURATION_MS
@@ -76,9 +105,8 @@ jobs:
               while (true) {
                 try {
                   const runs = await github.request(endpoint, params)
-                  const buildRun = runs.data.workflow_runs.find(
-                    (run) => run.head_sha === head_sha
-                  )
+                  const workflowRuns = runs && runs.data ? runs.data.workflow_runs : []
+                  const buildRun = workflowRuns.find((run) => run.head_sha === head_sha)
 
                   console.log('runs: ' + JSON.stringify(runs))
                   if (buildRun) {
@@ -102,39 +130,11 @@ jobs:
 
             const build_run = await findBuildWorkflowRun()
             if (!build_run) {
-              status = 'completed'
-              const conclusion = 'action_required'
+              await createDetectionFailedCheck()
+              return
+            }
 
-              await github.rest.checks.create({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                name: name,
-                head_sha: head_sha,
-                status: status,
-                conclusion: conclusion,
-                output: {
-                  title: 'Workflow run detection failed',
-                  summary: `
-            Unable to detect the workflow run for testing the changes in your PR.
-
-            1. If you did not enable GitHub Actions in your forked repository, please enable it by clicking the button as shown in the image below. See also [Disabling or limiting GitHub Actions for a repository](https://docs.github.com/en/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository) for more details.
-            2. Create and push an empty commit to trigger the workflow.
-            3. It is possible your branch is based on the old \`dev\` branch in Apache SeaTunnel, please sync your branch to the latest dev branch. For example as below:
-                \`\`\`bash
-                git fetch upstream
-                git rebase upstream/dev
-                git push origin YOUR_BRANCH --force
-                \`\`\``,
-                  images: [
-                    {
-                      alt: 'enabling workflows button',
-                      image_url: 'https://raw.githubusercontent.com/apache/spark/master/.github/workflows/images/workflow-enable-button.png'
-                    }
-                  ]
-                }
-              })
-            } else {
-              const run_id = build_run.id
+            const run_id = build_run.id
 
               // Here we get check run ID to provide Check run view instead of Actions view, see also SPARK-37879.
               const check_runs = await github.request(check_run_endpoint, check_run_params)
@@ -162,7 +162,7 @@ jobs:
                 repo: context.repo.repo,
                 name: name,
                 head_sha: head_sha,
-                status: status,
+                status: 'queued',
                 output: {
                   title: 'Test results',
                   summary: '[See test results](' + check_run_url + ')',
--- a/.github/workflows/notify_test_workflow.yml
+++ b/.github/workflows/notify_test_workflow.yml
@@ -136,42 +136,41 @@ jobs:
 
             const run_id = build_run.id
 
-              // Here we get check run ID to provide Check run view instead of Actions view, see also SPARK-37879.
-              const check_runs = await github.request(check_run_endpoint, check_run_params)
-              console.log('check_runs: ' + JSON.stringify(check_runs))
-              const actions_url = 'https://github.com/'
-                + context.payload.pull_request.head.repo.full_name
-                + '/actions/runs/'
-                + run_id
-              const run_url_part = '/actions/runs/' + run_id
-              const check_run_head = check_runs.data.check_runs.find(r =>
-                r.head_sha === build_run.head_sha
-                  && (
-                    (r.details_url && r.details_url.includes(run_url_part))
-                    || (r.html_url && r.html_url.includes(run_url_part))
-                  )
-              )
+            // Here we get check run ID to provide Check run view instead of Actions view, see also SPARK-37879.
+            const check_runs = await github.request(check_run_endpoint, check_run_params)
+            console.log('check_runs: ' + JSON.stringify(check_runs))
+            const actions_url = 'https://github.com/'
+              + context.payload.pull_request.head.repo.full_name
+              + '/actions/runs/'
+              + run_id
+            const run_url_part = '/actions/runs/' + run_id
+            const check_run_head = check_runs.data.check_runs.find(r =>
+              r.head_sha === build_run.head_sha
+                && (
+                  (r.details_url && r.details_url.includes(run_url_part))
+                  || (r.html_url && r.html_url.includes(run_url_part))
+                )
+            )
 
-              console.log('check_run_head: ' + JSON.stringify(check_run_head))
-              const check_run_url = check_run_head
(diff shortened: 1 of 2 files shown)


DanielLeens (commented):
I reviewed the current head `19058cae683cef223c361979a38f1b43c847a282` from the CI check-run handoff path.

## What this PR fixes
- User pain: the Apache-side `Build` proxy check can point at the wrong fork workflow run when the branch has multiple recent runs or when GitHub returns runs in an order that does not match the current PR head.
- Fix approach: query recent fork workflow runs and explicitly pick the run whose `head_sha` matches the current PR head; if no matching run exists, create an `action_required` check instead of linking a stale run.
- In one sentence: this makes the CI proxy check bind to the current commit rather than to whichever workflow run appears first.

## Full path checked
```text
pull_request_target opened / reopened / synchronize
  -> notify_test_workflow.yml
  -> query fork build_main.yml workflow runs
  -> find workflow_run where run.head_sha == PR head_sha
      -> found: create Apache-side Build check linked to that run
      -> not found: create action_required detection-failed check
```

The Maven wrapper retry change is also scoped: it retries wrapper jar download up to five times and removes partial downloads between attempts. The upgrade compatibility script change only prevents the edge-agent distribution from being selected as the current SeaTunnel dist archive.

I do not see a source-side blocker in this head. The remaining gate is external: the required Build check is still queued at `https://github.com/apache/seatunnel/runs/90618824897`. Also, since Daniel is the PR author here, a separate maintainer should provide the final independent review/approval.

## 5. Merge conclusion
### Conclusion: merge after fixes

1. Blocking items
- Wait for the current Build check to finish successfully.
- A non-author maintainer needs to provide the final review/approval.

2. Non-blocking suggestions
- None from my side.

DanielLeens (commented):
Rechecked the same current head `19058cae683cef223c361979a38f1b43c847a282`.

The specific CI fact I referenced in my previous round has now changed:
- the previously queued `Build` check is now green;
- I still do not see a new source-side blocker on this head from Daniel's side.

So the earlier CI-only gate from my last review is cleared now. At this point, the remaining gate is not a new code issue in this branch, but the independent maintainer review/approval step, because this PR is Daniel-authored.

DanielLeens (commented):

- **User pain point**: contributors periodically hit CI/tooling failures on `dev` and on the PR-notification workflow that are unrelated to their own code changes — an "unsynced commit" failure in the notification workflow, a transient `429 Too Many Requests` from Maven Central breaking Maven Wrapper bootstrap, and an "Upgrade Compatibility" job picking the wrong distribution archive.
- **Fix approach (as described in the PR body)**: (1) match the PR notification workflow's Build-run lookup by exact head SHA and fail soft into an `action_required` check instead of failing the notification workflow itself; (2) retry Maven Wrapper jar downloads on transient network/rate-limit failures; (3) exclude the edge-agent binary archive when the upgrade-compatibility script auto-detects the current distribution archive.
- **One-sentence summary**: A CI-stability PR touching the PR-notification workflow script and the Maven Wrapper bootstrap.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Issue 1 first, since it affects how the rest of this review should be read: the PR's actual diff does not match its description.**

`gh pr diff 11594 --name-only` and `gh pr view --json files` both show exactly two changed files at the current head (`b84b74d3`):
```
.github/workflows/notify_test_workflow.yml
mvnw
```
The PR body's third described root cause and fix — "`Upgrade Compatibility` on `dev` can pick the edge-agent binary archive... Exclude `apache-seatunnel-edge-agent-*-bin.tar.gz` when the upgrade compatibility script auto-detects the current SeaTunnel distribution archive" — has no corresponding change anywhere in the diff. `tools/upgrade_compatibility/run_upgrade_compatibility.sh` is not touched. The PR's own "Verification" section even lists `bash -n tools/upgrade_compatibility/run_upgrade_compatibility.sh` as something that was run, but since that file isn't part of this diff, that verification step proves nothing about this PR's actual changes — it would pass identically on `dev` with no changes applied. Either the intended third fix was dropped from this head by mistake, or the description is stale relative to the code; either way, the "Upgrade Compatibility" edge-agent-detection failure this PR claims to fix is not actually fixed by merging it as-is.

**`notify_test_workflow.yml` change, read in full at the current head:** the diff is smaller in substance than the PR description implies. The `try`/`catch` retry loop in `findBuildWorkflowRun()` (bounded by `MAX_LOOKUP_DURATION_MS`, matching by exact `run.head_sha === head_sha`) already existed before this PR — it is unchanged context in the diff, not new. What this PR actually adds is:
- A defensive guard, `const workflowRuns = runs && runs.data ? runs.data.workflow_runs : []`, protecting against `runs.data` being unexpectedly undefined before calling `.find(...)` on it.
- Extracting the "detection failed" check-run creation into a `createDetectionFailedCheck()` function (a readability refactor; same behavior, called from the same single site).
- Removing the mutable `let status` variable in favor of the equivalent literal strings in each branch.

The one functional change (the `runs.data` null-guard) is a reasonable defensive fix, but it's a narrower, different mechanism than "match by exact PR head SHA" (which already existed) — I'd recommend tightening the PR description to describe what actually changed here, since as written it credits this diff with fixing the race description, which the pre-existing `head_sha` filter was already doing.

**`mvnw` change:** adds a `download_with_retries()` helper (5 attempts, `rm -f "$wrapperJarPath"` between attempts, 5s sleep between retries) wrapping the existing `wget`/`curl` invocations, and quotes `$MVNW_USERNAME`/`$MVNW_PASSWORD` in the `curl`/`wget` invocations where they were previously unquoted (a good, if incidental, word-splitting hardening not mentioned in the PR description). This is a reasonable, minimal, well-scoped fix for the stated Maven Central `429` problem. One observation: if all 5 attempts fail, `download_with_retries` returns 1, but the caller does not check that return value or `exit` early with a clear "Wrapper jar download failed after 5 attempts" message — the script falls through to whatever uses `$wrapperJarPath` next (eventually a `java -jar` invocation that will fail with a less obvious "file not found"-style error). This is pre-existing behavior (the original code didn't check the exit status either, just did `|| rm -f`), so it's not a regression, but since this PR is specifically about making failures easier to diagnose, surfacing a clear message on final exhaustion would be a natural, low-risk addition while already touching this exact code path.

`mvnw` is normally a generated file (produced by the Maven Wrapper plugin, e.g. `mvn wrapper:wrapper`); see 3.2 for the maintainability implication of hand-patching it.

## 1.2 Compatibility Impact

Fully compatible. Both changes are CI/tooling-only: the workflow script change only affects how the notification check is created (still creates the same `Build` check name/conclusion shape), and the `mvnw` change only adds retries around an existing download with no change to the URL, jar version, or fallback-to-Java-download path for environments without `wget`/`curl`.

The `mvnw` retry adds up to 4 additional 5-second sleeps (20s total) only in the failure path when the initial download fails — no cost in the common case. The `notify_test_workflow.yml` change has no measurable perf impact; it's a null-guard and a refactor.

### Issue 2: PR description and Verification section describe a fix that is not present in the diff (Medium)
- **Location**: PR description ("Root causes observed" bullet 3, "Changes" bullet 3, "Verification" bullet `bash -n tools/upgrade_compatibility/run_upgrade_compatibility.sh`) vs. actual diff (`.github/workflows/notify_test_workflow.yml`, `mvnw` only).
- **Problem**: The PR claims to fix the "Upgrade Compatibility" edge-agent-archive-detection failure, but no file under `tools/upgrade_compatibility/` is touched by this diff.
- **Potential risk**: A reviewer or future maintainer reading the PR description (or the merge commit message, which will likely carry it) will believe the edge-agent detection bug is fixed on `dev` after this merges. It will not be. That failure mode will keep recurring, and the false paper trail makes it harder to notice, since this PR's title/description will appear to already cover it.
- **Best improvement**: Either add the actual `run_upgrade_compatibility.sh` fix to this PR (matching the description), or trim the PR description down to only the two changes actually present, and remove the misleading verification bullet.
- **Severity**: Medium — not a functional regression in what's shipped, but a real gap between claimed and actual scope that will mislead anyone relying on the PR description.

### Issue 3: `download_with_retries` failure is silent past the final attempt (Low)
- **Location**: `mvnw` (`download_with_retries()` and its three call sites for `wget`/`curl`).
- **Problem**: If all 5 attempts fail, the function returns 1 but the caller does not check it or print a clear final failure message.
- **Potential risk**: On genuine, non-transient failure (e.g., a firewalled CI runner), the script proceeds past a missing `maven-wrapper.jar` and fails later with a less diagnostic error.
- **Best improvement**: `download_with_retries ... || { echo "Failed to download maven-wrapper.jar after 5 attempts" >&2; exit 1; }` at each call site, or a single check after the `if/elif/else` block.
- **Severity**: Low — pre-existing gap, not a regression, but directly in scope for a PR whose purpose is CI diagnosability.

# 2. Code Quality Assessment

## 2.1 Coding Standards
The `notify_test_workflow.yml` script change is clean, readable JS with no wildcard-import equivalent concerns (N/A for a GitHub Actions script). The `mvnw` change follows the existing shell style in the file (POSIX `sh`, not bashisms) and is consistent with the surrounding function style.

**Rating: N/A / not independently testable in CI in the traditional sense** — this is a CI-infrastructure PR; its "tests" are the workflow runs themselves. The author's stated local verification (`sh -n mvnw`, YAML parse check, `node --check` on the extracted script body, `git diff --check`) is a reasonable substitute for the local-execution restriction this repo enforces on this task, and the two files that are actually static-checkable that way. The retry logic and null-guard have no automated regression coverage (understandably hard to test in-repo for a `pull_request_target`-triggered script), so correctness here relies on manual review plus observing the next few real CI runs on `dev`.

## 2.3 Documentation Updates
N/A — no user-facing docs are affected by CI-tooling changes.

# 3. Architectural Soundness

The `mvnw` retry wrapper is a small, targeted, idiomatic shell fix for a well-understood transient-network problem. The `notify_test_workflow.yml` refactor (extracting `createDetectionFailedCheck`) is a reasonable readability improvement, though as noted in 1.1 it doesn't add the retry/robustness capability the PR description implies — that capability already existed.

## 3.2 Maintainability
One real, forward-looking risk: `mvnw` is conventionally a generated file produced by the Maven Wrapper plugin (`mvn wrapper:wrapper` / `mvnw wrapper:wrapper`), not typically hand-maintained. A future regeneration of the wrapper (e.g. bumping the wrapper version) via the standard tooling would silently discard this PR's retry logic and quoting fix, with no test to catch the regression. Worth a one-line comment in `mvnw` near `download_with_retries` noting it's a manual patch that must be preserved/reapplied if the wrapper is ever regenerated, so a future contributor running the standard regeneration command doesn't unknowingly drop it.

## 3.3 Extensibility
N/A for this scope.

N/A — CI tooling only, no wire format, config, or runtime state involved.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | PR description/Verification section describes an "Upgrade Compatibility" edge-agent fix that is not present in the diff | PR description vs. diff (no `tools/upgrade_compatibility/**` change) | Medium |
| 2 | `download_with_retries` failure past the final attempt is silent (no explicit failure message/exit) | `mvnw` | Low |
| 3 | `mvnw` is a conventionally-generated file with no in-repo note that this patch must survive regeneration | `mvnw` | Low |

# 5. Merge Recommendation

**1. Blockers — must be fixed before merge:**
- Issue 1: Reconcile the PR description with the actual diff — either add the missing `tools/upgrade_compatibility/run_upgrade_compatibility.sh` edge-agent-exclusion fix so the PR delivers what it claims, or edit the description/verification section to only describe the two files actually changed. Shipping a PR whose merged description falsely claims a third CI failure mode is fixed is worse than not mentioning it, since it will suppress follow-up investigation of a real, still-open problem.

- Issue 2: Surface a clear failure message when `download_with_retries` is exhausted, since this PR is specifically about CI diagnosability.
- Issue 3: Add a short comment marking the manual retry/quoting patch in `mvnw` so it isn't lost on a future wrapper regeneration.

**Overall assessment.** The two changes actually present in this diff (the `runs.data` null-guard in the notification script, and the Maven Wrapper download retry) are each small, safe, and directionally correct fixes for real, previously-observed CI flakiness, and I found no functional defect in either. The primary blocker is scope-accuracy, not code correctness: the PR currently claims to fix three root causes but only ships fixes for two, and the "Verification" section cites a static check against a file this PR does not touch, which could be misread as evidence the third fix is included when it isn't.

davidzollo (approved):
+1
LGTM

PR #11596, opened 2026-07-30
[Docs] Clarify local deployment source build guidance

danielnadean:
## Summary
- remove the source-build walkthrough from the local deployment main flow
- replace it with a developer note that points source-build users to the development environment guide
- keep the getting-started deployment path focused on official binary releases and connector installation

## Tests
- ./mvnw spotless:apply
- verified the deployment pages no longer contain the source-build section or package-copy commands
- verified the linked developer setup pages exist


davidzollo (approved):
+1 if CI passes

PR #11573, opened 2026-07-27
[Improve][Connector-V2] Improve source split round-robin assignment for InfluxDB IoTDB Kudu and TDengine

JeremyXin:
### Purpose of this pull request

like #10869 #10867 #10868 

Improve source split assignment in InfluxDB, IoTDB, IoTDBv2, Kudu, and TDengine.

  ## Changes

  - replace hash-based assignment with ordered round-robin assignment
  - persist and restore assignCount
  - reassign returned splits to the original reader
  - add related unit tests

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config)

PR #11607, opened 2026-07-31
[Improve][Connector-V2] Improve source split round-robin assignment for Easysearch, AmazonDynamoDB, TiDB CDC and Typesense

JeremyXin:

like #11573 

Improve source split balancing for Easysearch, AmazonDynamoDB, TiDB CDC, and Typesense by aligning them with the round-robin split assignment strategy used in other optimized source connectors.

  ## Key changes
  - Replace hash-based split ownership with stable round-robin assignment during split enumeration
  - Persist assignCount in enumerator state to keep assignment continuity after checkpoint restore
  - Return splits to the original reader first in addSplitsBack() when the reader is still registered
  - Fix unassigned split counting to reflect the actual number of pending splits
  - Add unit tests for split assignment, split return, and checkpoint restore behavior

### Check list


DanielLeens:

SEZ9 (commented):
Thanks for working on this. I re-reviewed the latest head `fae6ef8e8a96` from scratch.

# What this PR fixes
- User pain: Hash-based split ownership in Easysearch, AmazonDynamoDB, TiDB CDC, and Typesense sources produced skewed assignments (e.g. hashing `totalSegments`/splitId can map many splits to the same reader), and `addSplitsBack()` re-shuffled returned splits away from the original reader; `currentUnassignedSplitSize()` also reported map key count instead of actual pending splits.
- Fix approach: Replace hash ownership with a stable round-robin counter (`AtomicInteger assignCount`) applied over splits sorted by splitId, persist `assignCount` in enumerator state (with a backward-compatible two-arg state constructor defaulting to 0), return splits directly to the originating reader in `addSplitsBack()` when it is still registered (logging a warning otherwise), fix `currentUnassignedSplitSize()` to sum pending split list sizes, and add unit tests with a `TestingEnumeratorContext` covering round-robin assignment, splits-back routing, and assignCount continuity across snapshot/restore.
- One-line summary: The change is consistent with the pattern established in #11573 and looks correct end-to-end; my main notes are the removal of `signalNoMoreSplits(subtaskId)` from the DynamoDB `addSplitsBack()` path (worth confirming intentional, since readers that get splits back will now rely on the normal enumeration path to be signalled), the reflection-based test invocation of the private `addPendingSplit` which is brittle to refactors, and one low-severity item: state/behavior changes without accompanying doc updates.

# Runtime chain I rechecked
```text
AmazonDynamoDBSourceSplitEnumerator.run() / handleSplitRequest()
  -> discoverySplits() -> addPendingSplit(Collection)  AmazonDynamoDBSourceSplitEnumerator.java:111
    -> splits sorted by splitId -> getSplitOwner(assignCount.getAndIncrement(), readerCount)  AmazonDynamoDBSourceSplitEnumerator.java:117,130
    -> pendingSplits.computeIfAbsent(ownerReader).add(split)  AmazonDynamoDBSourceSplitEnumerator.java:119
  -> assignSplit(readers) -> enumeratorContext.assignSplit(subtaskId, splits)

AmazonDynamoDBSourceSplitEnumerator.addSplitsBack(splits, subtaskId)  AmazonDynamoDBSourceSplitEnumerator.java:151
  -> addPendingSplit(splits, subtaskId)  [bypasses round-robin, keeps original owner]  AmazonDynamoDBSourceSplitEnumerator.java:126
  -> registeredReaders().contains(subtaskId) ? assignSplit(singleton(subtaskId)) : log.warn(...)  AmazonDynamoDBSourceSplitEnumerator.java:155-162
  [note: signalNoMoreSplits(subtaskId) removed from this path]

Checkpoint/restore continuity:
  snapshotState(checkpointId) -> new AmazonDynamoDBSourceState(shouldEnumerate, pendingSplits, assignCount.get())  AmazonDynamoDBSourceSplitEnumerator.java:185
  -> AmazonDynamoDBSourceState(boolean, Map, int)  AmazonDynamoDBSourceState.java:38
  restore: AmazonDynamoDBSourceSplitEnumerator(ctor) -> assignCount.set(sourceState.getAssignCount())  AmazonDynamoDBSourceSplitEnumerator.java:69

TiDBSourceSplitEnumerator.run() -> addPendingSplit(splits)  TiDBSourceSplitEnumerator.java:122
  -> sorted by splitId -> getSplitOwner(assignCount.getAndIncrement(), context.currentParallelism())  TiDBSourceSplitEnumerator.java:123-130
  restore: TiDBSourceSplitEnumerator(ctor) -> assignCount.set(restoreState.getAssignCount())  TiDBSourceSplitEnumerator.java:76
  state compat: TiDBSourceCheckpointState(boolean, Map) -> this(shouldEnumerate, pendingSplit, 0)  TiDBSourceCheckpointState.java:40
```

# Findings
- Severity: Low

# Review conclusion

DanielLeens (commented):
I reviewed the latest head `fae6ef8e8a9692e73511fab6e9e4a5438f9bead3` from scratch against the current `dev`, including the split-enumerator assignment path, add-splits-back behavior, checkpoint state continuity, and the newly added unit coverage.

I also noticed @SEZ9's earlier review. After tracing the current code myself, I largely agree with the source-level direction. The only place where I would respectfully push back is the doc-update note: I do not see a user-visible option/config contract change in this PR. This is an internal split-assignment / checkpoint-state behavior fix, so I would not require connector docs for it.

# PR solves what problem
- User pain: the previous hash-based ownership could skew source splits unevenly across readers, `addSplitsBack()` could reshuffle returned splits away from the original owner, and `currentUnassignedSplitSize()` under-reported actual pending work.
- Fix approach: replace the hash-based owner calculation with deterministic round-robin assignment over sorted splits, persist `assignCount` in enumerator checkpoint state, route `addSplitsBack()` to the original reader when it is still registered, and fix the pending-split size calculation.
- One-line summary: this is a focused enumerator correctness fix that makes split ownership deterministic and checkpoint/restore-consistent across Easysearch, AmazonDynamoDB, TiDB CDC, and Typesense.

# 1. Code Review

## 1.1 Core Logic Analysis
Main runtime path rechecked:

```text
Enumerator start / discovery
  -> run()
      -> discovery / get*Split()
      -> addPendingSplit(splits)
          -> sort by splitId
          -> owner = assignCount % parallelism
          -> assignCount++ per split
      -> assignSplit(registeredReaders)
      -> signalNoMoreSplits(readers)

Reader failure / split return
  -> addSplitsBack(splits, subtaskId)
      -> addPendingSplit(splits, subtaskId)
      -> if reader still registered, assign back to the same subtask
      -> otherwise keep them pending and log a warning

Checkpoint / restore
  -> snapshotState(checkpointId)
      -> state stores shouldEnumerate + pending splits + assignCount
  -> restore constructor
      -> assignCount resumes from checkpointed value
```

Key findings:
- The normal enumerator path definitely hits the new round-robin logic because ownership is now derived from `assignCount.getAndIncrement()` instead of a hash of the split id / segment count.
- `addSplitsBack()` now preserves original ownership instead of re-hashing returned splits into a different reader.
- `currentUnassignedSplitSize()` is now measuring actual pending split count instead of only the number of map keys.
- The state constructors remain backward-compatible by defaulting the old two-argument restore path to `assignCount = 0`.

I specifically rechecked the lifecycle detail that @SEZ9 called out around `signalNoMoreSplits`. In the current implementations, `run()` already signals readers after the first assignment pass, and `addSplitsBack()` is a failure-recovery path that reassigns known pending splits to a specific reader when it is still registered. I do not see a reopened correctness issue there on the current head.

## 1.2 Compatibility
Fully compatible.

- API: no public API removal.
- Configuration: no new user-facing option or changed default.
- Protocol / serialization: no wire/protocol change.
- Historical behavior: split distribution becomes more deterministic, but this is an internal enumerator behavior improvement rather than an incompatible contract break.
- Checkpoint/restore: the new `assignCount` field is stored explicitly, and the old constructor path defaults it to `0`, which keeps restore compatibility for pre-change state objects.

- CPU: sorting splits before assignment is cheap at the enumerator scale and avoids skew.
- Memory: the added `assignCount` integer is negligible.
- GC: no meaningful new churn on the hot data path.
- Network / IO: unchanged.
- Concurrency: `assignCount` is an `AtomicInteger`, and state mutation still goes through the enumerator's existing synchronization boundaries.
- Resource release: unchanged.

The new warning on `addSplitsBack()` for an unregistered reader is useful and avoids silently discarding returned splits. I do not see a swallowed-exception or misleading-success path added here.

# 2. Code Quality

## 2.1 Style
The code is consistent with the existing enumerator style, and the intent of the new state field is straightforward.

## 2.2 Tests and Stability
The new unit tests cover:
- round-robin distribution over sorted splits;
- returning splits to the original reader;
- checkpoint/restore continuity for `assignCount`.

Stability rating: stable.

Why:
- no sleeps or timing-based assertions;
- no ports, filesystems, or external services;
- deterministic in-memory test contexts only.

## 2.3 Docs
I do not see a mandatory user-doc update for this PR because no user-visible connector option or documented CLI contract changed here.

# 3. Architecture
This is a precise fix, not a workaround. The behavior change stays inside the enumerator/state boundary where ownership and restore continuity are already managed.

# 4. Findings Summary

| --- | --- | --- | --- |

# 5. Merge Conclusion
### Conclusion: can merge

1. Blocking items

- None from this round.

Current gate note:
- `Build` is successful on the current head: https://github.com/apache/seatunnel/runs/91476681967
- The branch is still behind/diverged from the latest `dev` in current compare metadata (`ahead_by=2`, `behind_by=6`, compare status: `diverged`). I am recording that as a mergeability fact, not turning it into a new contributor sync request on a green source/CI state.

Because I only have read-level review controls on this repository, this review is my current Daniel-side technical conclusion for the latest head. If GitHub still needs a formal approval / dismissal step to clear repository-level merge gates, a write-capable maintainer or reviewer will need to handle that final action.

JeremyXin:
@DanielLeens Do you think there are any other problems? If not, can you approve it?

DanielLeens (approved):
Thanks for the ping @JeremyXin.

I do not have any other problems to raise on this unchanged head `fae6ef8e8`. My previous full review already covered the enumerator assignment path, `addSplitsBack()` ownership, checkpoint-state continuity, and the new unit coverage, and the conclusion there was "can merge" with no source-level blocker. `Build` is still green on this head.

So with this reply I am upgrading my previous COMMENTED conclusion to a formal APPROVED review.

One transparency note: my review permission on this repository is read-level, so if the repository merge gate still requires a write-capable maintainer approval, that final step stays with a maintainer. The branch also remains behind the latest `dev` in compare metadata (`behind_by=6`, status `diverged`) — I am recording that as a mergeability fact only, not a contributor action request, since CI is green on the current head.

DanielLeens (changes requested):
Following up on my own earlier review/approval on this same head (`fae6ef8e8a96`) — a second, more targeted pass over the diff surfaced one thing I missed the first time. Downgrading my prior APPROVED to request changes for this specific point; everything else from my earlier review still stands (the round-robin fix is correct and addresses a real bug, checkpoint/state backward compatibility holds, and the `signalNoMoreSplits` removal in AmazonDynamoDB's `addSplitsBack` is safe — it was a duplicate call, not a dropped signal, since `assignSplit()` already signals each reader internally).

## Issue: an existing, already-passing test had its assertion weakened

`TiDBSourceSplitEnumeratorTest.addSplitsBackShouldNotOverwriteMultipleSplitsForSameReader` is a pre-existing test that this PR modifies. Its final assertion changed from:

```java
Assertions.assertEquals(splits, context.getAssignedSplits(0));
```

to:

```java
assertSplitsEqualIgnoringOrder(splits, context.getAssignedSplits(0));
```

`assertSplitsEqualIgnoringOrder` only compares `getKeyRange().getStart().toStringUtf8()` (1 of the split's 6 fields) after sorting both sides — order-insensitive and partial-field, versus the original's exact, order-sensitive, full-object equality.

I traced the path this specific test exercises: a single `addSplitsBack(splits, 0)` call → `addPendingSplit(splits, 0)` does `pendingSplit.computeIfAbsent(0, ...).addAll(splits)`, which preserves both order and object identity → `assignSplit` does `pendingSplit.remove(0)` and hands that same list to `context.assignSplit(0, ...)` → the test's mock `assignSplit` does `.addAll()` again. Order and references survive end-to-end, so the original strict assertion should still pass unchanged under this PR's new `addSplitsBack`/`addPendingSplit(splits, ownerReader)` implementation. I don't see a code path introduced by this PR that requires relaxing it — it looks like it may have carried over from writing the *other* new tests in this file, where order-independence is genuinely needed (the round-robin-over-sorted-splits path in the 1-arg `addPendingSplit`).

Weakening a passing test's assertion without a corresponding behavior change removes its ability to catch a future regression in split field propagation (`database`, `table`, `resolvedTs`, `snapshotStart`, `snapshotCompleted`) or ordering through `addSplitsBack` — which is exactly what this test is named to guard against.

**Suggested fix**: restore `Assertions.assertEquals(splits, context.getAssignedSplits(0))`. If order-independence is intentionally wanted here for some reason, this file already has a fuller `assertSplitEquals(expected, actual)` helper (compares all 6 fields) that could be applied pairwise instead of reducing the check to one substring.

The rest of the PR looks solid — in particular, the AmazonDynamoDB fix addresses a real bug: the old `getSplitOwner` hashed `totalSegments`, which is the same constant for every split from a given scan, so every split was landing on a single reader regardless of configured parallelism. Happy to re-approve once the assertion above is restored.

DanielLeens (approved):
Thanks for the ping, and sorry for the back-and-forth on this one. I re-reviewed the current head `724a02cc123f` (`724a02cc123fd32315eef07d30f7b80a08df8228`) from scratch against `dev`.

**Good news first: the test-weakening issue I flagged on 2026-08-06 is now resolved.** Commit `724a02cc123` ("Restore strict TiDB addSplitsBack assertion") reverts `TiDBSourceSplitEnumeratorTest.addSplitsBackShouldNotOverwriteMultipleSplitsForSameReader` back to `Assertions.assertEquals(splits, context.getAssignedSplits(0))`, exactly as I suggested. I re-traced the call chain independently (see below) and confirmed the strict assertion does pass under the new implementation, so this is a real fix, not just a cosmetic revert. I'm lifting my CHANGES_REQUESTED.

- **User pain point**: `Easysearch`, `AmazonDynamoDB`, `TiDB CDC`, and `Typesense` source enumerators picked the owning reader by hashing an identifier (`totalSegments` for DynamoDB, `splitId` for the others). For DynamoDB in particular, `totalSegments` is the same constant for every split produced by one scan, so `hashCode() % numReaders` mapped *every* split to a single reader regardless of configured parallelism — the other readers sat idle. `addSplitsBack()` also re-hashed returned splits instead of giving them back to the reader that lost them, and `currentUnassignedSplitSize()` counted map keys (reader count) instead of actual pending splits.
- **Fix approach**: Replace hash-based ownership with a monotonic `AtomicInteger assignCount` applied round-robin over splits sorted by `splitId`, persist `assignCount` in the enumerator checkpoint state so restart/restore continues the rotation instead of resetting bias, route `addSplitsBack()` splits straight back to the original reader (only if still registered, otherwise log and leave them pending), and fix `currentUnassignedSplitSize()` to sum list sizes.
- **One-sentence summary**: This makes parallel reads on these four connectors actually parallel, by giving every reader a fair, deterministic share of splits instead of a hash collision funneling everything onto one worker.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Files with core changes** (identical pattern applied to all four connectors):
- `AmazonDynamoDBSourceSplitEnumerator.java` / `AmazonDynamoDBSourceState.java`
- `EasysearchSourceSplitEnumerator.java` / `EasysearchSourceState.java`
- `TypesenseSourceSplitEnumerator.java` / `TypesenseSourceState.java`
- `TiDBSourceSplitEnumerator.java` / `TiDBSourceCheckpointState.java`

**Before** (e.g. DynamoDB, `AmazonDynamoDBSourceSplitEnumerator.java`):
```java
private void addPendingSplit(Collection<AmazonDynamoDBSourceSplit> splits) {
    int readerCount = enumeratorContext.currentParallelism();
    for (AmazonDynamoDBSourceSplit split : splits) {
        int ownerReader = getSplitOwner(split.getTotalSegments(), readerCount);
        pendingSplits.computeIfAbsent(ownerReader, r -> new ArrayList<>()).add(split);
    }
}
private static int getSplitOwner(Integer tp, int numReaders) {
    return (tp.hashCode() & Integer.MAX_VALUE) % numReaders;
}
...
public void addSplitsBack(List<AmazonDynamoDBSourceSplit> splits, int subtaskId) {
    if (!splits.isEmpty()) {
        addPendingSplit(splits);                       // re-hashed, not returned to origin
        assignSplit(Collections.singleton(subtaskId));
        enumeratorContext.signalNoMoreSplits(subtaskId); // unconditional, even if not registered
    }
}
```

**After**:
```java
private void addPendingSplit(Collection<AmazonDynamoDBSourceSplit> splits) {
    int readerCount = enumeratorContext.currentParallelism();
    splits.stream()
            .sorted(Comparator.comparing(AmazonDynamoDBSourceSplit::splitId))
            .forEach(split -> {
                int ownerReader = getSplitOwner(assignCount.getAndIncrement(), readerCount);
                pendingSplits.computeIfAbsent(ownerReader, r -> new ArrayList<>()).add(split);
            });
}
private void addPendingSplit(Collection<AmazonDynamoDBSourceSplit> splits, int ownerReader) {
    pendingSplits.computeIfAbsent(ownerReader, r -> new ArrayList<>()).addAll(splits);
}
private static int getSplitOwner(int assignCount, int numReaders) {
    return assignCount % numReaders;
}
...
public void addSplitsBack(List<AmazonDynamoDBSourceSplit> splits, int subtaskId) {
    if (!splits.isEmpty()) {
        addPendingSplit(splits, subtaskId);            // returned to origin reader, no re-hash
        if (enumeratorContext.registeredReaders().contains(subtaskId)) {
            assignSplit(Collections.singleton(subtaskId));
        } else {
            log.warn("Reader {} is not registered. Pending splits {} are not assigned.", subtaskId, splits);
        }
    }
}
```

**Key findings**
1. The normal path reaches this logic directly: `run()` calls `discoverySplits()` then `addPendingSplit(splits)` on every enumerator's first (and only, since `shouldEnumerate` flips false) discovery pass, so the round-robin fix is on the hot/primary split-assignment path for all four connectors, not just a recovery corner case.
2. `addSplitsBack()` is reachable on the reader-restart / reader-failure recovery path (subtask restart, split-back on cancel) — this PR changes that path's semantics from "re-hash into the global rotation" to "hand back to the original owner directly," which is the documented intent (`Return splits to the original reader first ... when the reader is still registered`) and matches the pattern already used elsewhere in the codebase per #11573.
3. The `assignCount` counter is only advanced from the single initial `addPendingSplit(splits)` call inside `run()` (single-threaded enumerator execution), and `addSplitsBack` uses the 2-arg overload that does not touch `assignCount` at all — so no concurrent-increment race exists in practice, and `addSplitsBack` correctly does not perturb the rotation for freshly-discovered splits.
4. `currentUnassignedSplitSize()` now correctly sums `List::size` across all pending-reader buckets instead of `Map.size()` (previously bounded by reader count, capping the reported value at parallelism regardless of actual backlog) — this is a real, previously-silent bug fix, not cosmetic.
5. Removing the unconditional `enumeratorContext.signalNoMoreSplits(subtaskId)` call from DynamoDB's `addSplitsBack()` (this was the point SEZ9 flagged and I independently traced in my first review): `assignSplit(Collections.singleton(subtaskId))` already ends its `for` loop with `enumeratorContext.signalNoMoreSplits(reader)` for every reader it processes — so when the reader *is* registered, the signal still fires exactly once (previously it may have fired twice back-to-back). When the reader is *not* registered, no signal is sent — correct, since there's no live subtask to receive it; it will be signalled again when `registerReader()` re-adds it via the same `assignSplit()` path. Net effect: fewer redundant calls, same reachable-reader coverage.

**In-depth correctness analysis**
- **Balance under splits < readers**: e.g. 2 splits, 4 readers → `assignCount` 0,1 → readers 0 and 1 get exactly one split each, readers 2/3 get none. Correct — there is nothing to give them.
- **Balance under splits not evenly divisible**: e.g. 5 splits, 3 readers → assignCount 0..4 % 3 = 0,1,2,0,1 → reader 0 and 1 get 2 splits, reader 2 gets 1. Max skew is 1, which is the best achievable round-robin balance.
- **Determinism**: splits are sorted by `splitId` before assignment, so the same split set always yields the same reader mapping given the same parallelism, independent of `Set`/`Collection` iteration order upstream (`discoverySplits()` returns a `HashSet` for DynamoDB) — this removes a latent nondeterminism that the old hash approach also had for ties, and makes the new unit tests reproducible.
- **Checkpoint/restore continuity**: `assignCount` is persisted in `snapshotState()` and restored in the constructor before any further `addPendingSplit` calls, so a job that restores from a checkpoint continues the rotation from where it left off rather than resetting to 0 and re-skewing assignment for splits discovered after restore.

**Complete runtime path** (DynamoDB, representative of all four):
```text
Job start
  -> AmazonDynamoDBSourceSplitEnumerator.run()  [L79-90]
      -> discoverySplits()  // builds full split set once
      -> addPendingSplit(newSplits)  [L112-124]
          -> sorted by splitId -> assignCount.getAndIncrement() -> ownerReader = count % readerCount
          -> pendingSplits[ownerReader].add(split)
      -> assignSplit(readers)  [L92-107]
          -> enumeratorContext.assignSplit(reader, list)
          -> enumeratorContext.signalNoMoreSplits(reader)   // always, once per reader

Reader restart / split-back
  -> addSplitsBack(splits, subtaskId)  [L154-166]
      -> addPendingSplit(splits, subtaskId)  // no re-hash, same reader
      -> registeredReaders().contains(subtaskId)?
            yes -> assignSplit(singleton(subtaskId)) -> re-signals that one reader
            no  -> log.warn(...), splits remain pending until registerReader() fires assignSplit again

Checkpoint
  -> snapshotState(checkpointId)  [L184-188]
      -> new AmazonDynamoDBSourceState(shouldEnumerate, pendingSplits, assignCount.get())
Restore
  -> constructor(ctx, config, sourceState)  [L58-70]
      -> assignCount.set(sourceState.getAssignCount())
```

## 1.2 Compatibility Impact

**Fully compatible.**
- No public API, `Option`, protocol, or default-value change — this is purely internal split-enumerator bookkeeping.
- Serialization/checkpoint compatibility: all four `*State`/`*CheckpointState` classes keep their original `serialVersionUID` unchanged and only *add* a field (`assignCount`). Restore/deserialization goes through plain Java `ObjectInputStream` (via `SerializationUtils`, confirmed in `seatunnel-common/src/main/java/.../SerializationUtils.java`), which tolerates new fields on the target class when deserializing an old byte stream — the field simply defaults to `0`. I verified this is also explicitly covered for TiDB by the pre-existing `checkpointStateShouldRestoreLegacySingleSplitPendingMap` test, which round-trips a state object through real serialization. A checkpoint/savepoint taken on a pre-PR version will restore cleanly on this version, with the rotation counter simply starting at 0 for any splits discovered from that point on — not a correctness issue, just a fresh rotation.
- The 2-arg constructors on `AmazonDynamoDBSourceState`, `EasysearchSourceState`, `TypesenseSourceState`, and `TiDBSourceCheckpointState` are kept (delegating to the new 3-arg constructor with `assignCount = 0`), preserving source compatibility for any other in-repo callers.

- CPU/memory: negligible — `sorted(Comparator.comparing(...))` runs once per discovery batch over a bounded split set (shard/segment/table-range count, not a per-record operation).
- Concurrency: `assignCount` is an `AtomicInteger`, but as noted above it is only mutated from the single-threaded enumerator `run()`/`addPendingSplit(Collection)` path in the code as currently wired — the atomic type is a reasonable defensive choice but not load-bearing for correctness today.
- No new resource handles, network calls, or retry paths introduced.
- No change to lock contention (`stateLock`/`synchronized` usage is unchanged).

- The new `log.warn("Reader {} is not registered. Pending splits {} are not assigned.", ...)` branch gives operators a clear signal instead of silently calling `assignSplit`/`signalNoMoreSplits` against a reader that isn't there — an improvement over the previous unconditional call.
- No exceptions are newly swallowed; existing `try/catch` around `enumeratorContext.assignSplit(...)` in `assignSplit()` is unchanged.
- No sensitive data in the new log lines (split identifiers only).

No blocking issues found in the current head.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- Follows existing style conventions of the surrounding enumerator classes.
- New fields (`assignCount`) and the new constructor overloads are small, self-explanatory, and consistent across all four connectors; I would not block on requiring extra Javadoc for these single-purpose fields, but a one-line comment on `assignCount` (e.g. "monotonic counter driving round-robin split ownership; persisted across checkpoints") would help future maintainers immediately understand *why* it needs to survive restore, rather than having to read the constructor/`snapshotState` wiring to figure that out. Non-blocking, Low severity.
- Minor: `TiDBSourceSplitEnumeratorTest.assertSplitsEqualIgnoringOrder(...)` is now unused after restoring the strict assertion — harmless, but could be removed as dead test code in a follow-up. Non-blocking, Low severity.

- Coverage is solid: each connector gets dedicated tests for round-robin assignment, `addSplitsBack` return-to-origin behavior, `currentUnassignedSplitSize()` accuracy, and checkpoint/restore continuity of `assignCount`. TiDB additionally keeps its pre-existing legacy-state-deserialization test, which is the right check for backward compatibility.
- **Test-strictness verification (the point I raised on 2026-08-06)**: `TiDBSourceSplitEnumeratorTest.addSplitsBackShouldNotOverwriteMultipleSplitsForSameReader` is restored to the original strict `Assertions.assertEquals(splits, context.getAssignedSplits(0))` in commit `724a02cc123`. I re-traced the path independently against the *current* implementation (not just trusting the commit message): `addSplitsBack(splits, 0)` → `addPendingSplit(splits, 0)` does `pendingSplit.computeIfAbsent(0, ...).addAll(splits)` (same list, same order, same object references) → `assignSplit` does `pendingSplit.remove(0)` and passes that exact list to `context.assignSplit(0, ...)` → the test's mock `assignSplit` does `.addAll(splits)` into `assignedSplitsByReader`. Order and object identity survive end-to-end, so the strict, order-sensitive equality check is valid and does pass under the current code — confirmed, not just asserted. This closes out my previous blocker.
- The three newly-added test files (`AmazonDynamoDBSourceSplitEnumeratorTest`, `EasysearchSourceSplitEnumeratorTest`, `TypesenseSourceSplitEnumeratorTest`) are brand new in this PR — there is no pre-existing baseline to weaken, so the "tests can only get stricter" rule doesn't apply to them, but I checked they don't introduce any new weak assertions either (no ignoring-order tricks, no partial-field comparisons) — same-object-reference `assertEquals` is used consistently across all four connectors' tests.
- Anti-pattern scan: no `Thread.sleep`, no timing-based waits, no reliance on wall-clock/`System.currentTimeMillis()`, no random data, no hard-coded ports, no shared static state across tests — every added/changed test uses a purely synchronous, in-memory `TestingEnumeratorContext` mock with no I/O.
- **Stability rating: Stable.** No flaky-test anti-patterns identified in any of the touched or added UT files (`AmazonDynamoDBSourceSplitEnumeratorTest.java`, `EasysearchSourceSplitEnumeratorTest.java`, `TypesenseSourceSplitEnumeratorTest.java`, `TiDBSourceSplitEnumeratorTest.java`).

## 2.3 Documentation Updates

- No user-visible behavior, config option, or default value changed — this is an internal correctness/bug fix. `docs/en` and `docs/zh` updates are not required for this PR. (SEZ9's earlier low-severity doc note referred to general housekeeping, not a functional gap introduced here — I don't see a concrete doc/example that is now inaccurate as a result of this diff.)

# 3. Architectural Soundness

**Precise fix.** It replaces a structurally broken ownership function (hashing a value that's constant across all splits, for DynamoDB) with the same round-robin pattern already validated in #11573, applied consistently across the four connectors that shared the bug.

## 3.2 Maintainability
The four connectors now share an identical, easy-to-follow pattern (`assignCount` + sorted round robin + return-to-origin `addSplitsBack`), which makes future review/maintenance of any one of them transferable to the others.

## 3.3 Extensibility
This establishes a reusable template for any other connector enumerator in the codebase that still uses hash-based ownership — a good reference implementation for future consistency passes.

Confirmed compatible, see 1.2 above — checkpoints/savepoints taken on pre-PR versions restore correctly, with `assignCount` defaulting to 0 and the rotation continuing cleanly from there for future discovery.

# 4. Issue Summary

No formal blocking or non-blocking issues raised in this round. Two very minor, non-blocking style notes (not numbered as formal issues, both Low, and both optional):
- `assignCount` fields could use a one-line "why persisted" comment (2.1).
- `TiDBSourceSplitEnumeratorTest.assertSplitsEqualIgnoringOrder` is now dead code (2.1).

`+1` to @SEZ9's earlier Low-severity, non-blocking doc-note — I don't see a concrete outdated doc/example caused by this diff, so I'm not promoting it to a blocker, consistent with SEZ9's own "can merge" conclusion.

# 5. Merge Recommendation

   - None. The previously-flagged test-weakening blocker (strict TiDB `addSplitsBack` assertion) is resolved in the current head (`724a02cc123`), verified against the real call chain, not just the commit message.

   - Optional: add a short comment on `assignCount` explaining why it's persisted across checkpoints (Low).
   - Optional: remove the now-unused `assertSplitsEqualIgnoringOrder` helper in the TiDB test (Low).

**Overall assessment**: This is a clean, well-scoped bug fix. The core defect (hash-based ownership skewing all DynamoDB splits onto one reader) is real and the round-robin replacement is correct under every split/parallelism combination I traced, including splits-fewer-than-readers and non-evenly-divisible cases. Checkpoint-state compatibility is sound (unchanged `serialVersionUID`, additive field, Java-serialization default-value tolerance verified against the actual `SerializationUtils` implementation). Test coverage is solid and stable — no flaky-test anti-patterns. @SEZ9's independent review already covered the same ground and reached the same "can merge" conclusion; I agree with that assessment.

One note on CI: as of this review, the `Build` check on the fork run for `724a02cc123` (https://github.com/JeremyXin/seatunnel/actions/runs/32227033341) is still `in_progress` — not failing, just not finished yet. That's a timing fact, not a code-side blocker; happy to take another look if it comes back red.

Thanks again for iterating on this, @JeremyXin — and again, sorry for the extra round-trip on the assertion issue.

DanielLeens (approved):
Thanks for the quick follow-up, @JeremyXin. I re-reviewed the full diff from scratch against the current `dev` on the new head `8a0bb60dc` (`8a0bb60dca7f291a9f163e1e77cbb9b70374599c`), since a new commit landed after my previous `APPROVED` on `724a02cc123f`.

**Good news: this is a clean, no-behavior-change follow-up.** The only change between the head I approved and this one is commit `8a0bb60dc` ("Clarify TiDB split assignment checkpoint intent"), which does exactly the two optional, non-blocking things I suggested last round: it adds a one-line comment on `TiDBSourceSplitEnumerator#assignCount` explaining why it's persisted across checkpoints, and it removes the now-dead `assertSplitsEqualIgnoringOrder` helper from `TiDBSourceSplitEnumeratorTest`. No production logic changed. My conclusion from the previous round stands, and I re-verified every core claim independently against this head rather than just trusting the diff summary — details below.

- **User pain point**: `Easysearch`, `AmazonDynamoDB`, `TiDB CDC`, and `Typesense` source enumerators picked the owning reader by hashing an identifier (`totalSegments` for DynamoDB, `splitId` for the others). For DynamoDB in particular, `totalSegments` is the same constant for every split produced by one scan, so `hashCode() % numReaders` mapped *every* split to a single reader regardless of configured parallelism — the other readers sat idle. `addSplitsBack()` also re-hashed returned splits instead of giving them back to the reader that lost them, and `currentUnassignedSplitSize()` counted map keys (reader count) instead of actual pending splits.
- **Fix approach**: Replace hash-based ownership with a monotonic `AtomicInteger assignCount` applied round-robin over splits sorted by `splitId`, persist `assignCount` in the enumerator checkpoint state so restart/restore continues the rotation instead of resetting bias, route `addSplitsBack()` splits straight back to the original reader (only if still registered, otherwise log and leave them pending), and fix `currentUnassignedSplitSize()` to sum list sizes.
- **One-sentence summary**: This makes parallel reads on these four connectors actually parallel, by giving every reader a fair, deterministic share of splits instead of a hash collision funneling everything onto one worker.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Files with core changes** (identical pattern across all four connectors, unchanged since my last full review):
- `AmazonDynamoDBSourceSplitEnumerator.java` / `AmazonDynamoDBSourceState.java`
- `EasysearchSourceSplitEnumerator.java` / `EasysearchSourceState.java`
- `TypesenseSourceSplitEnumerator.java` / `TypesenseSourceState.java`
- `TiDBSourceSplitEnumerator.java` / `TiDBSourceCheckpointState.java`

**New commit on top (`8a0bb60dc`), the only delta since my prior approval:**
```java
 private final Map<Integer, List<TiDBSourceSplit>> pendingSplit;
+// Persist the round-robin cursor so restored enumerators keep assigning newly discovered
+// splits in the same sequence instead of skewing ownership after a checkpoint restore.
 private final AtomicInteger assignCount = new AtomicInteger(0);
```
and, in the test file, deletion of the now-unreferenced `assertSplitsEqualIgnoringOrder(...)` helper (I confirmed with a repo-wide grep that it has zero remaining call sites, and that `java.util.Collections` — imported in the same test file — is still used elsewhere, so no unused-import fallout from the deletion).

**Key findings**
1. The normal path reaches this logic directly: `run()` calls `discoverySplits()`/`getTiDBSourceSplit()` then `addPendingSplit(splits)` on every enumerator's first (and only, since `shouldEnumerate` flips false) discovery pass, so the round-robin fix sits on the hot/primary split-assignment path for all four connectors, not just a recovery corner case.
2. `addSplitsBack()` is reachable on the reader-restart / reader-failure recovery path. I re-read all four `addSplitsBack` implementations on this head byte-for-byte (not just the previous review's summary): each now calls the 2-arg `addPendingSplit(splits, subtaskId)` (return-to-origin, no re-hash), then only calls `assignSplit(...)` if `registeredReaders().contains(subtaskId)`, otherwise logs a `WARN` and leaves the splits pending until `registerReader()` fires the same `assignSplit()` path later. This matches the documented intent and is consistent across DynamoDB, Easysearch, Typesense, and TiDB.
3. `assignCount` is only ever incremented from the single-threaded `run() -> addPendingSplit(Collection)` path; the `addSplitsBack` overload (`addPendingSplit(splits, ownerReader)`) never touches the counter. So there is no concurrent-increment race in practice, and returning a split via `addSplitsBack` does not perturb the rotation used for freshly discovered splits.
4. `currentUnassignedSplitSize()` correctly sums `List::size` across all pending-reader buckets in all four connectors (I re-checked each one individually on this head), fixing the previous bug where it was effectively capped at reader count regardless of actual backlog.
5. Checkpoint/restore continuity: in every connector's split-enumerator constructor, `assignCount.set(state.getAssignCount())` runs before any further `addPendingSplit` call, so a job restored from checkpoint continues the round-robin rotation instead of resetting to 0 and re-skewing assignment for splits discovered after restore.

**In-depth correctness analysis**
- **Balance under splits < readers**: e.g. 2 splits, 4 readers → `assignCount` 0,1 → readers 0 and 1 get one split each, readers 2/3 get none — correct, there's nothing left to give them.
- **Balance under non-evenly-divisible counts**: e.g. 5 splits, 3 readers → assignCount 0..4 % 3 = 0,1,2,0,1 → readers 0/1 get 2 splits, reader 2 gets 1. Max skew is 1, the best achievable round-robin balance.
- **Determinism**: splits are sorted by `splitId` before assignment in all four connectors, so the mapping is reproducible for a given split set + parallelism, independent of upstream `Set`/`Collection` iteration order (e.g. DynamoDB's `discoverySplits()` returns a `HashSet`).

**Complete runtime path** (representative — identical shape in all four connectors):
```text
Job start
  -> XxxSourceSplitEnumerator.run()
      -> discoverySplits() / getTiDBSourceSplit()   // builds the full split set once
      -> addPendingSplit(newSplits)
          -> sorted by splitId -> assignCount.getAndIncrement() -> ownerReader = count % readerCount
          -> pendingSplits[ownerReader].add(split)
      -> assignSplit(readers) -> context.assignSplit(reader, list) -> context.signalNoMoreSplits(reader)

Reader restart / split-back
  -> addSplitsBack(splits, subtaskId)
      -> addPendingSplit(splits, subtaskId)   // no re-hash, same reader
      -> registeredReaders().contains(subtaskId)?
            yes -> assignSplit(singleton(subtaskId))
            no  -> log.warn(...), splits remain pending until registerReader() re-fires assignSplit

Checkpoint / Restore
  -> snapshotState(checkpointId) -> new XxxState(shouldEnumerate, pendingSplits, assignCount.get())
  -> constructor(ctx, config, state) -> assignCount.set(state.getAssignCount())
```

## 1.2 Compatibility Impact

**Fully compatible.** No public API, `Option`, protocol, or default-value change — this is internal split-enumerator bookkeeping only.

Serialization/checkpoint compatibility, re-verified on this head: all four `*State`/`*CheckpointState` classes keep their original `serialVersionUID` unchanged (`TiDBSourceCheckpointState.serialVersionUID = 6292978509042158791L`, matched by the equivalent constants in the other three classes) and only *add* the `assignCount` field, with the 2-arg constructor still present and delegating to the 3-arg one with `assignCount = 0`. Plain-Java deserialization (`ObjectInputStream` via `SerializationUtils`) tolerates a new field on the target class when reading an old byte stream — it defaults to `0`, so a checkpoint/savepoint taken on a pre-PR version restores cleanly, with the rotation counter simply starting fresh from that point. TiDB additionally keeps the pre-existing `checkpointStateShouldRestoreLegacySingleSplitPendingMap` test, which round-trips through real serialization and covers exactly this path.

- CPU/memory: negligible — the `sorted(...)` call runs once per discovery batch over a bounded split set (shard/segment/table-range count), not per record.
- Concurrency: `assignCount` is an `AtomicInteger`, but as traced above it's only mutated from the single-threaded enumerator path today — a reasonable defensive choice, not load-bearing for correctness in the current call sites.
- No new resource handles, network calls, or retry paths introduced by this PR or by the new commit.

- The `log.warn("Reader {} is not registered. Pending splits {} are not assigned.", ...)` branch in `addSplitsBack` gives operators a clear signal instead of silently calling into a reader that isn't there — an improvement over the prior unconditional call.
- No exceptions newly swallowed; no sensitive data in the new/changed log lines (split identifiers only).

**No blocking issues found on this head.**

# 2. Code Quality Assessment

## 2.1 Coding Standards

- Consistent style across all four connectors.
- The new `assignCount` comment in `TiDBSourceSplitEnumerator.java:51-52` is exactly the kind of "why persisted" note I asked for last round — it explains the constraint (checkpoint restore continuity) rather than restating the code. Since the other three connectors (`AmazonDynamoDB`, `Easysearch`, `Typesense`) carry the identical pattern and were part of this same PR's scope, a matching one-line comment on their `assignCount` fields would be a nice, purely optional follow-up for consistency — non-blocking, Low severity, not required for this PR since I only asked about the TiDB one specifically last round.
- The unused `assertSplitsEqualIgnoringOrder` test helper is now gone, closing out my prior Low-severity note.

- Coverage is unchanged and solid: dedicated tests per connector for round-robin assignment, `addSplitsBack` return-to-origin behavior, `currentUnassignedSplitSize()` accuracy, and checkpoint/restore continuity of `assignCount`.
- **Test-strictness re-verification**: I re-confirmed on this exact head that `TiDBSourceSplitEnumeratorTest.addSplitsBackShouldNotOverwriteMultipleSplitsForSameReader` (line 49) still uses the strict, order-sensitive `Assertions.assertEquals(splits, context.getAssignedSplits(0))` — the new commit only removes the now-orphaned helper method, it does not touch this assertion or any other test body. The test-weakening issue I raised on 2026-08-06 and confirmed fixed on 2026-08-19 remains fixed.
- Anti-pattern scan (unchanged from my previous pass, re-checked on this head): no `Thread.sleep`, no timing-based waits, no wall-clock dependence, no random data, no hard-coded ports, no shared static state; every touched test uses a synchronous in-memory `TestingEnumeratorContext` mock with no I/O.
- **Stability rating: Stable.** No flaky-test anti-patterns in any touched or added UT file (`AmazonDynamoDBSourceSplitEnumeratorTest.java`, `EasysearchSourceSplitEnumeratorTest.java`, `TypesenseSourceSplitEnumeratorTest.java`, `TiDBSourceSplitEnumeratorTest.java`). Removing a dead, unused assertion helper does not reduce coverage or strictness of any active test.

## 2.3 Documentation Updates

No user-visible behavior, config option, or default value changed — `docs/en` and `docs/zh` updates are not required for this PR.

# 3. Architectural Soundness

**Precise fix.** Replaces a structurally broken ownership function with the round-robin pattern already validated in #11573, applied consistently across the four affected connectors.

## 3.2 Maintainability
All four connectors share an identical, easy-to-follow pattern (`assignCount` + sorted round robin + return-to-origin `addSplitsBack`), and the new TiDB comment makes the "why" of the persisted counter explicit for future readers.

## 3.3 Extensibility
Good reference implementation for any other connector enumerator in the codebase still using hash-based ownership.

Re-confirmed: checkpoints/savepoints taken on pre-PR versions restore correctly, with `assignCount` defaulting to `0` and the rotation continuing cleanly for future discovery.

# 4. Issue Summary

No formal blocking or non-blocking issues on this head. The two optional Low-severity notes from my previous review (add a comment on `assignCount`, remove the dead test helper) are both resolved by commit `8a0bb60dc`. One further optional (non-blocking, Low) note: mirroring the new TiDB comment onto the other three connectors' `assignCount` fields for consistency — entirely at the author's discretion.

# 5. Merge Recommendation

   - None.

   - Optional, Low: mirror the new "why persisted across checkpoints" comment onto `assignCount` in the DynamoDB/Easysearch/Typesense enumerators for consistency with TiDB.

**Overall assessment**: This round is a clean, zero-behavior-change cleanup of the two optional notes from my last review — no production logic changed, and I independently re-verified the round-robin math, the checkpoint/restore path, the `addSplitsBack` return-to-origin behavior, and the strict TiDB assertion on this exact head rather than relying on the previous review. Everything holds.

**On CI**: the `Build` check on the current head (`8a0bb60dc`) shows `failure`, but looking at the actual fork run (JeremyXin/seatunnel run `32267833989`), every job passed — unit tests on JDK 8/11 across Windows and Ubuntu, all `updated-modules-integration-test` shards, license/style/dead-link checks — except `updated-modules-integration-test-part-8 (8, ubuntu-latest)`, which ran for roughly 3.5 hours before being `cancelled` rather than failing on an assertion. That pattern (long-running shard cut off, not a real test failure) looks like a CI-infrastructure timeout rather than anything introduced by this PR's diff, which only touches enumerator round-robin logic and unit tests. I'd suggest re-running just that one job; happy to take another look if a real failure shows up after the rerun. Separately, this branch is currently `19` commits behind `dev` (`diverged`, 5 ahead) — not blocking today since the failure isn't clearly upstream-caused, but worth keeping in mind if `Build` keeps failing after a rerun.

Thanks again for the quick turnaround, @JeremyXin — this is good to go from my side pending a green `Build`.

DanielLeens (commented):
CI follow-up (2026-08-22): my last review (2026-08-20) noted the Apache-side `Build` check was showing `failure` on this head (`8a0bb60dc`), traced to `updated-modules-integration-test-part-8 (8, ubuntu-latest)` being cancelled after a ~3.5-hour run rather than failing on an assertion, and suggested re-running just that job.

Re-checked live: the head is unchanged (`8a0bb60dca7f291a9f163e1e77cbb9b70374599c`), and both the Apache-side `Build` check and the fork's own run (`JeremyXin/seatunnel` run `32267833989`) now show a clean `success` conclusion. That was CI-infrastructure noise as diagnosed, not a regression from this diff.

No change to my conclusion: no source-level blockers found on this head, only an optional Low-severity style note (mirror the "why persisted across checkpoints" comment onto the other three connectors' `assignCount` fields). This is good to go from my side.

davidzollo (approved):
+1
LGTM

PR #11608, opened 2026-07-31
[Fix][E2E] Stabilize Paimon incompatible schema test

davidzollo:
### Why
Recent dev CI had repeated Paimon E2E failures in the same negative schema test, independent of a specific PR diff:

- 2026-07-28 Build run https://github.com/apache/seatunnel/actions/runs/30350787586/job/90257731539 failed in `PaimonSinkCDCIT.testSinkWithIncompatibleSchema` with a `ConditionTimeoutException`.
- 2026-07-30 Schedule Backend run https://github.com/apache/seatunnel/actions/runs/30561716311/job/90942758546 failed in the same test with `st-multi-table-sink-writer-1` still running after the expected schema mismatch failure.

The test already proves the user-facing Paimon schema validation error. The flake comes from applying the generic post-job thread-leak check to an intentionally failing sink initialization path.

### What
- Keep the successful Paimon CDC write through `executeJob` so the target table is created and connector jars are copied into the container.
- Execute the expected-failure schema mismatch config through the base SeaTunnel CLI command and keep asserting `exitCode = 1` plus the exact schema mismatch message.
- Fall back to `executeJob` for test containers that do not expose base command execution.

### Verification
- `./mvnw -f seatunnel-e2e/seatunnel-connector-v2-e2e/connector-paimon-e2e/pom.xml spotless:apply -nsu`
- `git diff --check`

Local compile/test/E2E/Docker were not run, following the SeaTunnel local verification rule for this CI-fix workflow. The final runtime verdict should come from GitHub Actions on this PR head.


DanielLeens:

DanielLeens (changes requested):
Thanks for working on this. I reviewed the latest head `38eb398fc92860e300e005f51f6530cf5c5787fa` from the full Paimon incompatible-schema test path down into the shared E2E container execution helpers.

The direction makes sense: this test should assert the user-facing schema-validation failure without becoming flaky. But I found one blocker in the current implementation.

## Runtime path I checked

```text
Paimon incompatible-schema test
  -> run a successful setup job with container.executeJob(...)
  -> run the negative schema-mismatch job
      -> PaimonSinkCDCIT.executeExpectedFailureJob(...)
          -> container.executeBaseCommand(...)
              -> start shell directly
              -> returns CLI result only

Shared E2E container contract
  -> SeaTunnelContainer.executeJob(...)
      -> doExecuteJob(...)
          -> capture JVM thread list before job
          -> run job
          -> wait until non-system leftover threads disappear
```

## Issue 1: the new helper bypasses the shared post-job thread-leak check instead of fixing the failure path

- Location:
  - `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-paimon-e2e/src/test/java/org/apache/seatunnel/e2e/connector/paimon/PaimonSinkCDCIT.java:145`
  - `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-paimon-e2e/src/test/java/org/apache/seatunnel/e2e/connector/paimon/PaimonSinkCDCIT.java:162`
  - `seatunnel-e2e/seatunnel-e2e-common/src/test/java/org/apache/seatunnel/e2e/common/container/seatunnel/SeaTunnelContainer.java:334`
- Severity: High

`PaimonSinkCDCIT.executeExpectedFailureJob()` now switches the negative path from `container.executeJob(...)` to `container.executeBaseCommand(...)`. That avoids `SeaTunnelContainer.doExecuteJob()`, which is exactly where the shared post-job thread-leak check runs.

So the current change does not really stabilize the failing path; it weakens the E2E contract for this case by skipping the container-level resource check entirely. If the schema-mismatch path leaves worker threads behind, this test will no longer tell us that.

Please keep the negative test on the normal `executeJob(...)` path and fix the real leftover-thread cause, or add a narrowly-scoped shared helper that still runs the same post-job cleanup/thread contract while allowing an expected non-zero exit code. Bypassing the shared execution contract at the test class level is too broad for this E2E.

## Merge conclusion

### Conclusion: can merge after fix

1. Blocking items
- Issue 1: do not bypass the shared `executeJob(...)` thread-leak guard for the negative test path.

2. Non-blocking notes
- The current required `Build` is also not usable yet on this head (`https://github.com/apache/seatunnel/runs/91206574837`, `Build=CANCELLED`), but the source-level issue above needs to be fixed regardless of the CI rerun.

Once the negative-path stabilization keeps the shared E2E execution contract intact, I am happy to re-review the updated head.

DanielLeens (changes requested):
Thanks for the ping — I re-reviewed the latest head (`5e1d93b7f6c66c0a0e13b0296b91a65f59fc783e`) as a full re-review rather than trusting the new commit to be a fix.

**Important finding up front**: the new commit on top of what I reviewed on 2026-08-02 is only `5e1d93b7f6c Merge branch 'dev' into davidzollo_ci-paimon-incompatible-schema-e2e-20260801` — a rebase onto the latest `dev`. `git diff 38eb398fc92860e300e005f51f6530cf5c5787fa pr-11608-head -- .../PaimonSinkCDCIT.java` is **empty**: the code is byte-identical to the exact head I already gave `CHANGES_REQUESTED` on. My High-severity blocker from 2026-08-02 has not been addressed, and this is not a case of me missing something in an earlier round — it's the same open issue, still unresolved on the current head.

`PaimonSinkCDCIT#testSinkWithIncompatibleSchema` intentionally submits a job with a mismatched schema and expects it to fail with a specific error message. The stated goal is to stabilize this negative-path test.

Before: the negative job ran through `container.executeJob(confFile)`, the same shared path every job in this E2E suite uses, which includes a post-job thread-leak check.
After (current head): the negative job runs through a new `executeExpectedFailureJob(container, confFile)` helper that calls `container.executeBaseCommand(...)` instead, bypassing that shared thread-leak check entirely.

# 1. Code Change Review
## 1.1 Core Logic Analysis
```
testSinkWithIncompatibleSchema
  -> container.executeJob("/fake_cdc_sink_paimon_case1.conf")           // successful setup job, exit=0
  -> executeExpectedFailureJob(container, ERROR_SCHEMA_CONF)
       -> container.copyFileToContainer(confFile, "/tmp" + confFile)
       -> container.executeBaseCommand(["--config", "/tmp/<conf>", "--name", "<conf>"])
       -> catch UnsupportedOperationException -> fallback: container.executeJob(confFile)
  -> assert exitCode == 1 and error message contains the schema-mismatch text

SeaTunnelContainer contract (seatunnel-e2e-common/.../SeaTunnelContainer.java)
  executeJob(...)        -> doExecuteJob(...)                      [runs the shared post-job thread-leak check, :334-372]
  executeBaseCommand(...) -> executeCommand(server, command)        [no thread tracking at all, :307-314]
```
I independently re-verified the technical claim behind my prior finding by reading `SeaTunnelContainer.java` directly:
- `doExecuteJob` (`:334`) snapshots `beforeThreads`, increments a shared `runningCount`, runs the job, then (only once `runningCount` reaches 0, i.e. the last concurrently-running job in this container) diffs `afterThreads` against `beforeThreads` and asserts no non-system threads are left over, retrying up to 120s via Awaitility.
- `executeBaseCommand` (`:307`) does none of that — it is a bare shell-command runner with zero thread bookkeeping.
- So `executeExpectedFailureJob` genuinely removes the post-job resource-leak assertion for this specific negative-path job, exactly as I found before.

I also checked whether the `catch (UnsupportedOperationException e)` fallback to `executeJob(confFile)` (which *would* restore the thread-leak check) is realistically exercised for this test. `TestContainer.executeBaseCommand` is a default interface method that throws `UnsupportedOperationException` unless overridden (`TestContainer.java:57-60`); `SeaTunnelContainer` overrides it (no exception), and this test class carries `@DisabledOnContainer(type = {EngineType.SPARK, EngineType.FLINK}, ...)`, meaning it **only ever runs against the Zeta/`SeaTunnelContainer` engine** — the one engine that does override `executeBaseCommand`. So the fallback branch is dead code for this test class: the bypass is not an occasional edge case, it is taken on every single run of this test.

## Issue 1: the negative-path helper still bypasses the shared post-job thread-leak contract (carried over, unresolved)
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-paimon-e2e/src/test/java/org/apache/seatunnel/e2e/connector/paimon/PaimonSinkCDCIT.java:154-176` (`executeExpectedFailureJob`), contrasted with `seatunnel-e2e/seatunnel-e2e-common/src/test/java/org/apache/seatunnel/e2e/common/container/seatunnel/SeaTunnelContainer.java:307-314` (`executeBaseCommand`, no leak check) vs `:334-372` (`doExecuteJob`, the shared leak check that `executeJob` uses).
- Problem: if the schema-validation failure path leaves Paimon sink/writer threads (or any other worker thread pool created before the failure is detected) running in the container, this test will no longer report it — the check that exists specifically to catch that class of bug is skipped for exactly the code path (an init-time failure) most likely to trigger it.
- Risk: a real resource leak on the schema-mismatch failure path can ship silently. Because `SeaTunnelContainer` instances are typically reused across `@TestTemplate` invocations in the same test class, any thread this negative test leaks is never captured as "before" state for this job (since this call never runs `doExecuteJob`'s `beforeThreads` snapshot) and only surfaces, if at all, as unexplained noise in whichever *later* test in the same container happens to call `executeJob` next — making a future real regression much harder to attribute correctly, not just invisible.
- Best improvement: Option A — keep the negative test on the normal `executeJob(...)` path and actually fix whatever leftover-thread cause originally motivated this change (the real bug, if any, should be root-caused rather than routed around). Option B — if `executeJob`'s current post-job semantics are the true blocker (e.g. it treats any non-zero exit as a hard test failure before the thread check even runs, or something in that path is itself flaky), add a narrowly-scoped shared helper on `TestContainer`/`SeaTunnelContainer` that still performs the exact same before/after thread-leak assertion contract as `doExecuteJob` while tolerating an expected non-zero exit code, rather than dropping the resource check altogether.
- Severity: High
- Raised by another reviewer: No — this is my own 2026-08-02 finding, still open on the current head. I want to be explicit that I'm not silently re-raising a stale comment: I re-verified the call chain end-to-end above rather than assuming my earlier read was still accurate.

## Issue 2: the `UnsupportedOperationException` fallback in `executeExpectedFailureJob` is unreachable for this test class
- Location: `PaimonSinkCDCIT.java:154-176` combined with the class-level `@DisabledOnContainer(type = {EngineType.SPARK, EngineType.FLINK}, ...)` annotation.
- Problem: the fallback exists to "keep the original path for test containers that do not expose base command execution," but Spark/Flink (the only other engines this test could theoretically run against) are already excluded at the class level, and the only engine left (`SeaTunnelContainer`) does implement `executeBaseCommand`. So the fallback never executes in practice; it reads as a safety net but isn't one for this test.
- Risk: Low — mostly a clarity/dead-code issue, not a correctness bug on its own. I'm listing it because it's adjacent evidence for Issue 1: the "defensive fallback" framing in the Javadoc slightly overstates how conditional this bypass actually is — it is unconditional for the engine this test runs against.
- Best improvement: either remove the now-misleading fallback comment/branch, or resolve this together with Issue 1 (if the helper goes away entirely, this goes away with it).
- Severity: Low

## 1.2 Compatibility Impact
**Fully compatible.** Test-only file; no production code, API, config, or protocol is touched. (Compatibility is not the blocking dimension here — test-behavior correctness is.)

No production-path performance impact. The one relevant side effect is exactly Issue 1: removing the post-job resource-leak guard for this specific job execution.

No swallowed exceptions in the reachable path; `copyFileToContainer`/`executeBaseCommand` propagate `IOException`/`InterruptedException` normally. No secrets logged.

# 2. Code Quality Assessment
## 2.1 Coding Standards
Reasonable Javadoc explaining intent (I disagree with the conclusion, not the clarity). `ERROR_SCHEMA_CONF` extraction into a named constant is a fine, unrelated micro-cleanup.

**Stability rating: Risk present.** The change does make the exit-code/error-message assertions themselves less flaky-looking (bypassing `doExecuteJob`'s post-job wait entirely removes one source of timing variance), but it does so by deleting a correctness check rather than fixing the underlying instability, which is a net loss for what this test suite is supposed to guarantee. Per this protocol's testing rule, a test-code change that weakens what a test actually verifies must be treated as a blocker regardless of whether CI turns green — CI passing here does not prove the schema-mismatch path is leak-free, it proves the check was skipped.

## 2.3 Documentation Updates
Not applicable.

# 3. Architectural Soundness
Not elegant as-is: it solves "this test is unstable" by removing part of what the test checks, rather than addressing why the negative path might be unstable through the normal `executeJob` contract.

## 3.2 Maintainability
Introduces a second, narrower job-execution path (`executeBaseCommand` + manual `copyFileToContainer`) alongside the shared `executeJob` contract, specifically for negative tests — if this pattern spreads to other connector E2E suites, the shared thread-leak guard quietly stops covering an entire class of "expected failure" jobs across the codebase.

## 3.3 Extensibility
N/A.

Not applicable.

# 4. Issue Summary
1. Issue 1 (High): negative-path helper bypasses the shared post-job thread-leak check — carried over from 2026-08-02, still unresolved on the current (rebase-only) head.
2. Issue 2 (Low): the `UnsupportedOperationException` fallback is dead code for this test class given its `@DisabledOnContainer` restriction to Spark/Flink.

# 5. Merge Recommendation

**Blockers (must-fix):**
- Issue 1 (High): keep (or restore) the shared post-job thread-leak contract for the schema-mismatch negative test — either fix the real leftover-thread cause on the normal `executeJob` path, or add a shared helper that preserves the same leak-detection guarantee while tolerating the expected non-zero exit code.

- Issue 2 (Low): drop or fix the misleading dead-code fallback comment once Issue 1 is resolved.

**Overall assessment:** I want to be encouraging about the direction here — pinning down a stable, deterministic way to assert on an expected job failure is a legitimate and worthwhile goal, and the CLI-result assertions themselves are fine. But routing around the shared thread-leak contract to get there trades a flaky-test problem for a silent resource-leak-detection gap, which is the more expensive failure mode long-term. The current head is a pure rebase with no attempt yet to address this, so my 2026-08-02 `CHANGES_REQUESTED` stands unchanged. Happy to re-review as soon as the negative path is fixed in a way that keeps the leak check intact.

davidzollo:
Addressed the remaining Paimon negative-path blocker in `4e9470996`.

The incompatible-schema job now uses `container.executeJob(ERROR_SCHEMA_CONF)` again, so it goes through the shared `SeaTunnelContainer.doExecuteJob(...)` path and keeps the same post-job thread-leak check as normal E2E jobs. I removed the local `executeBaseCommand(...)` helper/fallback entirely; the test still asserts the expected non-zero exit code and schema error message.

Local validation stayed inside the allowed formatting boundary: `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-paimon-e2e -DskipTests -DskipIT=true -Dmaven.gitcommitid.skip=true spotless:apply spotless:check` and `git diff --check`. I did not run local compile/test/E2E; fresh GitHub CI is queued on the new head.

DanielLeens (commented):

`PaimonSinkCDCIT#testSinkWithIncompatibleSchema` intentionally submits a job with a mismatched schema and expects it to fail with a specific error message. The goal is to stabilize this negative-path test without weakening what it verifies.

I re-reviewed the current head (`4e9470996`) as a full re-review, not by trusting the commit message.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The diff is minimal and precise: one line changed, 24 lines removed.

```java
// before (the head I gave CHANGES_REQUESTED on twice, 2026-08-02 and 2026-08-18)
Container.ExecResult errResult = executeExpectedFailureJob(container, ERROR_SCHEMA_CONF);
// after (4e9470996)
Container.ExecResult errResult = container.executeJob(ERROR_SCHEMA_CONF);
```

The entire `executeExpectedFailureJob(...)` helper — the one that called `container.executeBaseCommand(...)` and bypassed the shared post-job thread-leak check — is deleted. The negative-path job now goes through exactly the same `container.executeJob(...) -> doExecuteJob(...)` path as the successful setup job on the preceding line, and as every other job in this E2E suite.

I independently re-verified this actually closes my prior blocker by reading `SeaTunnelContainer.java` directly rather than trusting the diff:

- `doExecuteJob` (`SeaTunnelContainer.java:334-372`, test-scope class) snapshots `beforeThreads`, increments `runningCount`, runs the job via `executeJob(server, confFile, jobId, variables)`, and — once `runningCount` reaches 0 — diffs `afterThreads` against `beforeThreads` and asserts (via a 120s `Awaitility` wait) that no leftover non-system threads remain.
- Critically, `doExecuteJob` never asserts or branches on the job's own exit code — it just runs the job and returns whatever `Container.ExecResult` comes back, letting `runningCount` and the thread snapshot govern the leak check independently of success/failure. So calling `executeJob` on a job that is *expected* to exit with code `1` (schema mismatch) works exactly the same as calling it on a successful job: the thread-leak check still runs, and the caller's own assertions (`exitCode == 1`, error message content) are untouched by this change.

This is a complete fix, not a partial one: there is no longer any fallback path, no `UnsupportedOperationException` catch, and no special-casing for this test. The negative-path job is now indistinguishable, from the container's point of view, from every other job in the suite — which is exactly what closes the "this test silently loses the leak check on every run" defect I raised twice.

Runtime path, confirmed by direct read:
```
testSinkWithIncompatibleSchema
  -> container.executeJob("/fake_cdc_sink_paimon_case1.conf")   // setup, exit=0
  -> container.executeJob(ERROR_SCHEMA_CONF)                    // now the SAME path as above
       -> doExecuteJob(...)                                     // thread-leak check applies unconditionally
  -> assert exitCode == 1 and error message contains the schema-mismatch text  // unchanged
```

## 1.2 Compatibility Impact

Fully compatible. This is a test-only change; no production code, config, or protocol is touched.

None beyond the test itself now paying the same (already-standard) thread-leak-check cost every other job in the suite pays. No new resource usage.

Not applicable — no error-handling or logging code is touched by this diff.

**Live CI status at `4e9470996`:** the apache-side `Build` check is a pointer; the real run is in `davidzollo/seatunnel`. I checked the fork's check-runs for this exact commit directly: the `paimon-connector-it` jobs are `queued`, and the run has not completed yet. This is a fresh push (pushed the same morning as this review), so there is no compile/test evidence for this head yet — it needs to be confirmed green rather than assumed, but there is no reason from the diff itself to expect a new failure.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Clean — the fix is a pure deletion plus a one-line revert to the pre-existing shared helper. No new code, so no new comment/Javadoc obligations.

**Stable.** The test now uses the same well-established, already-battle-tested `executeJob`/`doExecuteJob` path used by the rest of the E2E suite, rather than a bespoke helper with its own untested failure-fallback branch. This removes rather than adds risk. No `Thread.sleep`, no new timing assumptions, no weakened assertions — the exit-code and error-message checks are byte-identical to before.

## 2.3 Documentation Updates

Not applicable — internal E2E test only, no user-visible behavior change.

# 3. Architectural Soundness

**Precise fix.** Rather than patching the bypass helper to also perform a thread-leak check (which would have duplicated logic already in `doExecuteJob`), the fix removes the special-casing entirely and lets the test go through the ordinary, already-correct shared path. This is the simplest possible correct fix, and it is smaller than the code it replaces.

## 3.2 Maintainability

Improved — one fewer bespoke helper method with its own fallback branch for future maintainers to reason about.

## 3.3 Extensibility

Not applicable to a single-test fix.

Not applicable — no persisted format, config, or API surface is touched.

# 4. Issue Summary

| --- | --- | --- | --- |
| — | (prior) negative-path helper bypassed the shared thread-leak check | `PaimonSinkCDCIT.java` | High |

No open issues remain. The prior High-severity finding is fixed and verified against the underlying container helper's actual behavior, not just the diff.

# 5. Merge Recommendation

**1. Blockers — must be fixed:**
- Confirm the fork CI run for `4e9470996` completes green — it was still `queued` for the `paimon-connector-it` jobs at review time, so there is no compile/test evidence for this exact head yet. Given the fix is a minimal, verified-correct revert to the shared path, I have no specific reason to expect a new failure, but this needs to be checked rather than assumed before merge.

- None.

**Overall assessment.** This fully and correctly addresses the blocker I raised on 2026-08-02 and re-confirmed as still open on 2026-08-18: the negative-path test no longer bypasses the shared post-job thread-leak check, and I verified the mechanism (not just the diff) by reading `doExecuteJob`'s actual behavior on a job with a non-zero exit code. The fix is minimal, removes more code than it adds, and introduces no new risk. Once CI is confirmed green at this head, this is ready to merge.

DanielLeens (approved):
CI recheck on this PR: my last review's only remaining item was confirming the fork CI run completed green for `4e9470996` — the `paimon-connector-it` jobs were still `queued` at review time. That head is unchanged, and `Build`, `Notify test workflow`, and `labeler` are all now `SUCCESS` on it. The fix itself (removing the negative-path bypass so the test goes through the shared, already-correct `doExecuteJob` path) was already verified as a minimal, correct revert with no new risk, so with CI now green on the same commit I reviewed, I'm updating this to an approval.

nzw921rx (approved):
+1 LGTM if ci pass

DanielLeens:
Thanks @nzw921rx — confirming the CI-gate condition in your approval: `Build`, `Notify test workflow`, and `labeler` are all `SUCCESS` on the current head (`4e9470996`), the same head I approved on 2026-08-21. So this looks clear to merge on the CI side as well.

zhangshenghang (approved):
+1,  LGTM！

PR #11609, opened 2026-07-31
[Docs] Define CDC schema evolution runtime contract

danielnadean:
### Purpose

Refs #11402.

This PR records the CDC SchemaChangeEvent runtime contract before follow-up runtime implementation work. It defines the per-table ordering, schema epoch, checkpoint boundary, sink apply, replay, restore, and unsupported-path semantics that a schema-evolution-capable path must satisfy.

### Scope

- Add English and Chinese CDC schema evolution contract architecture docs.
- Link the contract from the CDC pipeline architecture docs and schema evolution configuration docs.
- Register the new architecture page in the docs sidebar.
- Documentation only; no runtime behavior is changed in this PR.

### Validation

- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -T 3C spotless:apply`
- `git diff --cached --check`

Local compile, unit tests, E2E, Docker, and runtime verification were not run. Per the SeaTunnel local workflow boundary for this task, build and test truth should come from GitHub CI for this PR head.

Issue #11402, opened 2026-07-11
[Discussion][CDC][DDL] Define end-to-end SchemaChangeEvent ordering and recovery contract


davidzollo:
## Background

SeaTunnel already has CDC schema evolution related capabilities, but we still lack a clear end-to-end contract for one critical question:

**When a schema change event is observed, what is the exact ordering and recovery contract across source, transform, engine coordination, and sink application?**

This is not about adding schema evolution from scratch. It is about making the runtime contract explicit, testable, and production-safe.

## Problem to solve

Today, users and contributors can still be uncertain about:

- when a schema change is allowed to move downstream
- how records using the old schema are separated from records using the new schema
- how schema changes interact with checkpoint and sink flush boundaries
- what behavior is required when failure occurs before, during, or after sink-side schema application
- what must happen after checkpoint or savepoint restore
- when an unsupported source, transform, engine, or sink path must fail instead of continuing with ambiguous semantics

Without an explicit contract, long-term CDC correctness remains fragile even if individual schema evolution features continue to be added.

## Topics for discussion

The discussion should reach agreement on at least the following points:

- per-table ordering guarantees for `SchemaChangeEvent`
- the boundary between records written with the old schema and records written with the new schema
- whether schema changes for different tables may proceed independently
- behavior when failure occurs before sink apply, during sink apply, after sink apply but before the following checkpoint, and during restore
- duplicate or replay behavior when the sink mutation succeeded but the corresponding runtime state was not durably checkpointed
- fail-fast behavior for unsupported source, transform, engine, or sink paths
- checkpoint/savepoint compatibility expectations across SeaTunnel versions

Because the result affects API, engine coordination, connector behavior, and restore compatibility, the contract should be discussed on `dev@seatunnel.apache.org` and recorded in a STIP or design document before implementation begins.

## Proposed delivery model

This issue is a discussion and tracking issue. Documentation is necessary to record the agreed contract, but documentation alone does not complete the runtime capability.

After the contract is agreed, implementation should be split into focused follow-up issues. The expected areas are:

1. API and event contract
   - event identity and per-table ordering metadata
   - serialization and version compatibility
   - stable behavior required from transforms and sinks
2. Zeta coordination and recovery
   - a recoverable schema-change coordination state machine
   - checkpoint boundaries before and after schema application
   - deterministic handling of partial apply, timeout, restart, and replay
3. E2E recovery verification
   - fault injection at each schema-change phase
[...]

PR #11614, opened 2026-07-31
[Fix][Zeta] Fail fast unsupported schema evolution sink

danielnadean:
### Purpose

This is the first runtime implementation PR for #11402 after the contract documentation PR #11609.

SeaTunnel already has partial CDC/schema-change plumbing, but unsupported sink writers could still silently consume schema change events through the deprecated default `SinkWriter.applySchemaChange(...)` no-op path. That makes the documented end-to-end schema evolution contract look stronger than the actual runtime behavior.

This PR makes unsupported schema evolution fail fast instead of being acknowledged silently. It does not complete the full schema-change state machine/recovery contract; that remains follow-up implementation work.

### Changes

- Add `SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(...)` as the shared dispatch helper.
- Preserve compatibility with real legacy writers that override the deprecated `SinkWriter.applySchemaChange(...)`.
- Throw `UnsupportedOperationException` when a writer does not implement schema evolution and does not override the legacy method.
- Use the helper in Zeta `SinkFlowLifeCycle` and `MultiTableSinkWriter`.
- Preserve the latest `dev` multi-table schema-change barrier and shared physical sink fan-out, while making the final sub-writer apply step use the fail-fast helper.
- Make the row-error `ErrorHandlingSinkWriter` wrapper re-check the real delegate with the helper, so wrapper-level error handling cannot accidentally re-enable the deprecated no-op path.
- Use the same helper in Flink common and Flink 2.0 sink writer paths, so unsupported sinks send a failure ack instead of success.
- Remove Lance's schema-evolution marker because it only delegated to the deprecated no-op method.
- Make Iceberg fail fast when a schema-change event arrives while `iceberg.table.schema-evolution-enabled=false`.
- Rebase the PR branch onto the latest `apache/dev` to clear the stale/diverged branch state.

### Tests

- Added API tests for supported writer dispatch, legacy override compatibility, and unsupported writer fail-fast.
- Added engine wrapper test for `ErrorHandlingSinkWriter` around an unsupported delegate.
- Added Flink common test for unsupported writer failure ack behavior.
- Added Iceberg writer test for disabled schema-evolution fail-fast without table-loader I/O.

### Local Validation

- `./mvnw -nsu -Dmaven.gitcommitid.skip=true -T 3C -pl seatunnel-api,seatunnel-engine/seatunnel-engine-server,seatunnel-connectors-v2/connector-lance,seatunnel-connectors-v2/connector-iceberg,seatunnel-translation/seatunnel-translation-flink/seatunnel-translation-flink-common,seatunnel-translation/seatunnel-translation-flink/seatunnel-translation-flink-20 spotless:apply`
- `git diff --check apache/dev...HEAD`
- `git diff --check`

Per the local SeaTunnel workflow guidance, I did not run compile, unit tests, E2E, Docker, or runtime jobs locally; GitHub CI is the source of truth for those checks.

### DIFF
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/SupportSchemaEvolutionSinkWriter.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/SupportSchemaEvolutionSinkWriter.java
@@ -20,6 +20,7 @@ package org.apache.seatunnel.api.sink;
 import org.apache.seatunnel.api.table.schema.event.SchemaChangeEvent;
 
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.util.Optional;
 
 public interface SupportSchemaEvolutionSinkWriter {
@@ -45,5 +46,57 @@ public interface SupportSchemaEvolutionSinkWriter {
      */
     default Optional<String> getPhysicalSinkTableIdentifier() {
         return Optional.empty();
+    }
+
+    /**
+     * Applies a schema change only when the sink writer explicitly declares schema evolution
+     * support.
+     *
+     * <p>The deprecated {@link SinkWriter#applySchemaChange(SchemaChangeEvent)} path is kept only
+     * for legacy writers that really override it. Writers that inherit the default no-op method
+     * must fail fast, otherwise a CDC schema change could be silently dropped while new-schema
+     * records continue downstream.
+     *
+     * @param writer sink writer that receives the schema change event
+     * @param event schema change event from upstream
+     * @throws IOException if the sink writer fails while applying the schema change
+     */
+    @SuppressWarnings("deprecation")
+    static void applySchemaChangeToWriter(SinkWriter<?, ?, ?> writer, SchemaChangeEvent event)
+            throws IOException {
+        if (writer instanceof SupportSchemaEvolutionSinkWriter) {
+            ((SupportSchemaEvolutionSinkWriter) writer).applySchemaChange(event);
+            return;
+        }
+        if (overridesDeprecatedApplySchemaChange(writer)) {
+            writer.applySchemaChange(event);
+            return;
+        }
+        throw new UnsupportedOperationException(
+                String.format(
+                        "Sink writer %s received schema change event for table %s, but it does "
+                                + "not implement SupportSchemaEvolutionSinkWriter or override "
+                                + "SinkWriter.applySchemaChange(SchemaChangeEvent). "
+                                + "Schema evolution requires a schema-evolution-capable sink.",
+                        writer.getClass().getName(), event.tablePath()));
+    }
+
+    /**
+     * Checks whether a legacy sink writer provides its own deprecated schema change implementation.
+     *
+     * @param writer sink writer to inspect
+     * @return true when the writer overrides {@link
+     *     SinkWriter#applySchemaChange(SchemaChangeEvent)}
+     */
+    static boolean overridesDeprecatedApplySchemaChange(SinkWriter<?, ?, ?> writer) {
+        try {
+            Method method =
+                    writer.getClass().getMethod("applySchemaChange", SchemaChangeEvent.class);
+            return method.getDeclaringClass() != SinkWriter.class;
+        } catch (NoSuchMethodException e) {
+            throw new IllegalStateException(
+                    "SinkWriter.applySchemaChange(SchemaChangeEvent) is missing from the writer API",
+                    e);
+        }
     }
 }
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java
@@ -528,13 +528,8 @@ public class MultiTableSinkWriter
                 dispatchTarget.getSinkIdentifier().getTableIdentifier(),
                 MultiTableFailurePhase.CHECKPOINT,
                 () -> {
-                    if (dispatchTarget.getWriter() instanceof SupportSchemaEvolutionSinkWriter) {
-                        ((SupportSchemaEvolutionSinkWriter) dispatchTarget.getWriter())
-                                .applySchemaChange(event);
-                    } else {
-                        // TODO remove deprecated method
-                        dispatchTarget.getWriter().applySchemaChange(event);
-                    }
+                    SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(
+                            dispatchTarget.getWriter(), event);
                     return null;
                 });
         log.info(
--- a/seatunnel-api/src/test/java/org/apache/seatunnel/api/sink/SupportSchemaEvolutionSinkWriterTest.java
+++ b/seatunnel-api/src/test/java/org/apache/seatunnel/api/sink/SupportSchemaEvolutionSinkWriterTest.java
@@ -1,0 +1,152 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.api.sink;
+
+import org.apache.seatunnel.api.table.catalog.TablePath;
+import org.apache.seatunnel.api.table.schema.event.SchemaChangeEvent;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import java.io.IOException;
+import java.util.Optional;
+
+/**
+ * Verifies the shared sink-side schema change support gate.
+ *
+ * <p>The gate must preserve explicit legacy implementations while preventing the default no-op
+ * schema change method from silently accepting CDC schema changes.
+ */
+class SupportSchemaEvolutionSinkWriterTest {
+
+    /**
+     * Verifies the primary schema evolution extension point.
+     *
+     * @throws IOException if applying the schema change fails unexpectedly
+     */
+    @Test
+    void shouldApplySchemaChangeWithSupportInterface() throws IOException {
+        SupportInterfaceWriter writer = new SupportInterfaceWriter();
+        SchemaChangeEvent event = schemaChangeEvent();
+
+        SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(writer, event);
+
+        Assertions.assertSame(event, writer.appliedEvent);
+    }
+
+    /**
+     * Verifies compatibility with writers that still override the deprecated hook.
+     *
+     * @throws IOException if applying the schema change fails unexpectedly
+     */
+    @Test
+    void shouldKeepDeprecatedApplySchemaChangeOverrideCompatible() throws IOException {
+        DeprecatedOverrideWriter writer = new DeprecatedOverrideWriter();
+        SchemaChangeEvent event = schemaChangeEvent();
+
+        SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(writer, event);
+
+        Assertions.assertSame(event, writer.appliedEvent);
+    }
+
+    /**
+     * Verifies that inheriting the deprecated default no-op hook is not treated as schema evolution
+     * support.
+     */
+    @Test
+    void shouldFailFastWhenSinkWriterDoesNotSupportSchemaEvolution() {
+        PlainWriter writer = new PlainWriter();
+        SchemaChangeEvent event = schemaChangeEvent();
+
+        UnsupportedOperationException exception =
+                Assertions.assertThrows(
+                        UnsupportedOperationException.class,
+                        () ->
+                                SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(
+                                        writer, event));
+
+        Assertions.assertTrue(exception.getMessage().contains(PlainWriter.class.getName()));
+        Assertions.assertTrue(exception.getMessage().contains("test_db.test_table"));
+    }
+
+    /**
+     * Creates the schema change event used by all support-gate cases.
+     *
+     * @return a mocked schema change event with a stable table path
+     */
+    private static SchemaChangeEvent schemaChangeEvent() {
+        SchemaChangeEvent event = Mockito.mock(SchemaChangeEvent.class);
+        Mockito.when(event.tablePath()).thenReturn(TablePath.of("test_db", "test_table"));
+        return event;
+    }
+
+    /**
+     * Sink writer that inherits the deprecated default no-op schema change method.
+     *
+     * <p>This writer represents sinks that have no schema evolution support at all.
+     */
+    private static class PlainWriter implements SinkWriter<Object, Object, Object> {
+
+        @Override
+        public void write(Object element) {}
+
+        @Override
+        public Optional<Object> prepareCommit() {
+            return Optional.empty();
+        }
+
+        @Override
+        public void abortPrepare() {}
+
+        @Override
+        public void close() {}
+    }
+
+    /**
+     * Sink writer that uses the current schema evolution extension point.
+     *
+     * <p>This is the preferred path for schema-evolution-capable sinks.
+     */
+    private static class SupportInterfaceWriter extends PlainWriter
+            implements SupportSchemaEvolutionSinkWriter {
+
+        private SchemaChangeEvent appliedEvent;
+
+        @Override
+        public void applySchemaChange(SchemaChangeEvent event) {
+            this.appliedEvent = event;
+        }
+    }
+
+    /**
+     * Sink writer that keeps the legacy deprecated schema change extension point.
+     *
+     * <p>The shared support gate still accepts this path for backward compatibility.
+     */
+    private static class DeprecatedOverrideWriter extends PlainWriter {
+
+        private SchemaChangeEvent appliedEvent;
+
+        @Override
+        @SuppressWarnings("deprecation")
+        public void applySchemaChange(SchemaChangeEvent event) {
+            this.appliedEvent = event;
+        }
+    }
+}
--- a/seatunnel-connectors-v2/connector-iceberg/src/main/java/org/apache/seatunnel/connectors/seatunnel/iceberg/sink/IcebergSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-iceberg/src/main/java/org/apache/seatunnel/connectors/seatunnel/iceberg/sink/IcebergSinkWriter.java
@@ -118,14 +118,18 @@ public class IcebergSinkWriter
 
     @Override
     public void applySchemaChange(SchemaChangeEvent event) throws IOException {
-        // Waiting cdc connector support schema change event
-        if (config.isTableSchemaEvolutionEnabled()) {
-            log.info("changed rowType before: {}", fieldsInfo(rowType));
-            this.rowType = dataTypeChangeEventHandler.reset(rowType).apply(event);
-            log.info("changed rowType after: {}", fieldsInfo(rowType));
-            tryCreateRecordWriter();
-            writer.applySchemaChange(this.rowType, event);
+        if (!config.isTableSchemaEvolutionEnabled()) {
+            throw new UnsupportedOperationException(
+                    String.format(
+                            "Iceberg sink received schema change event for table %s, but "
+                                    + "iceberg.table.schema-evolution-enabled is false.",
+                            event.tablePath()));
         }
+        log.info("changed rowType before: {}", fieldsInfo(rowType));
+        this.rowType = dataTypeChangeEventHandler.reset(rowType).apply(event);
+        log.info("changed rowType after: {}", fieldsInfo(rowType));
+        tryCreateRecordWriter();
+        writer.applySchemaChange(this.rowType, event);
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-iceberg/src/test/java/org/apache/seatunnel/connectors/seatunnel/iceberg/sink/IcebergSinkWriterTest.java
+++ b/seatunnel-connectors-v2/connector-iceberg/src/test/java/org/apache/seatunnel/connectors/seatunnel/iceberg/sink/IcebergSinkWriterTest.java
@@ -18,7 +18,9 @@
 package org.apache.seatunnel.connectors.seatunnel.iceberg.sink;
 
 import org.apache.seatunnel.api.table.catalog.PhysicalColumn;
+import org.apache.seatunnel.api.table.catalog.TableIdentifier;
 import org.apache.seatunnel.api.table.catalog.TableSchema;
+import org.apache.seatunnel.api.table.schema.event.AlterTableAddColumnEvent;
 import org.apache.seatunnel.api.table.type.BasicType;
 import org.apache.seatunnel.connectors.seatunnel.iceberg.IcebergTableLoader;
 import org.apache.seatunnel.connectors.seatunnel.iceberg.config.IcebergSinkConfig;
@@ -37,7 +39,10 @@ import java.util.Optional;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
 class IcebergSinkWriterTest {
@@ -109,5 +114,29 @@ class IcebergSinkWriterTest {
         List<IcebergSinkState> states = writer.snapshotState(99L);
         assertEquals(1, states.size());
         assertEquals(99L, states.get(0).getCheckpointId());
+    }
+
+    /**
+     * Iceberg must fail fast instead of silently accepting schema changes when the table schema
+     * evolution option is disabled.
+     */
+    @Test
+    void testApplySchemaChangeFailsWhenSchemaEvolutionDisabled() {
+        when(config.isTableSchemaEvolutionEnabled()).thenReturn(false);
+        IcebergSinkWriter writer =
+                new IcebergSinkWriter(tableLoader, config, minimalSchema(), null);
+        AlterTableAddColumnEvent event =
+                AlterTableAddColumnEvent.add(
+                        TableIdentifier.of("catalog", "database", "table"),
+                        PhysicalColumn.of(
+                                "name", BasicType.STRING_TYPE, (Long) null, true, null, null));
+
+        UnsupportedOperationException exception =
+                assertThrows(
+                        UnsupportedOperationException.class, () -> writer.applySchemaChange(event));
+
+        assertTrue(
+                exception.getMessage().contains("iceberg.table.schema-evolution-enabled is false"));
+        verifyNoInteractions(tableLoader);
     }
 }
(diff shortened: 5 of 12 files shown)


DanielLeens (commented):
Thanks for working on this. I went through the full diff and traced both the Zeta and Flink schema-change paths. The approach looks solid to me; I did not find a source-level blocker in the current head. The remaining blocker is the Build gate.

## What this PR fixes

This PR turns an unsafe schema-evolution edge case into an explicit failure. Before this change, a sink writer could inherit the deprecated default no-op `SinkWriter.applySchemaChange()` method, receive a CDC `SchemaChangeEvent`, and effectively drop that schema change while newer-schema records continued downstream. That is risky because the job may look alive while the sink schema is no longer aligned with the incoming data.

The fix centralizes the decision in `SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(...)`: writers that implement `SupportSchemaEvolutionSinkWriter` are allowed, legacy writers that truly override the deprecated hook are still allowed, and writers that only inherit the default no-op now fail fast. Zeta, multi-table sink, and both Flink sink wrappers now use the same gate. Iceberg also now fails explicitly when `iceberg.table.schema-evolution-enabled` is false, and Lance no longer advertises schema-evolution support through an empty implementation.

## Runtime flow I checked

```text
Zeta CDC schema-change path
  -> Source emits SchemaChangeEvent
  -> TransformFlowLifeCycle maps the event through transforms
  -> SinkFlowLifeCycle.received(record)
     -> record.getData() instanceof SchemaChangeEvent
     -> processSchemaChangeEvent(event)
     -> SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(writer, event)
        -> writer implements SupportSchemaEvolutionSinkWriter
           -> writer.applySchemaChange(event)
        -> legacy writer overrides deprecated SinkWriter.applySchemaChange(...)
           -> writer.applySchemaChange(event)
        -> writer only inherits the default no-op
           -> UnsupportedOperationException
           -> task fails fast instead of silently dropping the schema change
```

```text
Flink CDC schema-change path
  -> Flink SchemaOperator emits a SeaTunnelRow control record
     with options["schema_change_event"]
  -> FlinkSinkWriter.write()
     -> handleControlMessage(options)
     -> handleSchemaChangeEvent(event, options)
     -> SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(sinkWriter, event)
        -> success
           -> sendSchemaChangeAck(..., success=true)
        -> unsupported or apply failed
           -> sendSchemaChangeAck(..., success=false)
           -> throw SinkWriterSchemaException
```

```text
Multi-table sink path
  -> MultiTableSinkWriter.applySchemaChange(event)
     -> find sub-writer by event.tablePath().getFullName()
     -> synchronized(queue runnable)
     -> executeWithTableRetry(... CHECKPOINT ...)
        -> SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(subWriter, event)
        -> unsupported sub-writer throws UnsupportedOperationException
           -> FAIL_FAST policy: fail the job
           -> CONTINUE_OTHER_TABLES policy: quarantine that table through existing failure handling
```

I also checked the current connector implementations. The schema-evolution-capable writers, such as JDBC, Doris, Databend, StarRocks, Elasticsearch, File, Iceberg, and Paimon, implement `SupportSchemaEvolutionSinkWriter`, so they remain on the supported path. The deprecated fallback is preserved only for real overrides.

## Review findings

No source-level blockers found.

Compatibility looks good: this does not remove or rename an API, SPI method, config option, protocol, or serialization format. The deprecated `SinkWriter.applySchemaChange()` method remains available, and explicit legacy overrides are still honored. The behavior change only affects unsupported sinks receiving schema-change events, where failing fast is the safer data-correctness behavior.

The added tests look stable. They are pure unit tests using mocks or in-memory recording writers, with no Docker, network, fixed ports, sleeps, clock assumptions, or cross-test state. The only non-blocking gap I see is that a future targeted test for multi-table unsupported sub-writer quarantine would be useful, but the shared helper and Flink wrapper already cover the core behavior.

## CI / merge gate

The current Build check is failing, but the failing lane does not look directly related to this PR's diff. The failed job is `Run / rocketmq-connector-it (11, ubuntu-latest)`, where `RocketMqIT.testRocketMqSpecificOffsetsToConsole` timed out while checking that no threads were still running in the container. The surviving threads are RocketMQ consumer/client housekeeping threads, which are outside the API/Zeta/Flink schema-evolution gate changed here.

I also checked the current head against the latest `dev`: this branch is `diverged`, with `ahead_by=1` and `behind_by=6`. Since the current CI failure appears unrelated to this PR's source changes, the lowest-cost next step is to sync with the latest `dev` and rerun Build first. If the same RocketMQ lane still fails after the sync, please paste the new failing check link here and I can help narrow it down further.

## 五、报告是否可以 Merge 的结论

### 结论：修复后可以 merge

1. 阻塞项（必须修复）

- No code-side blocker from this review.
- The remaining blocker is the GitHub Build gate. Please sync with the latest `dev` and rerun Build first because the current RocketMQ E2E failure appears unrelated to this PR's schema-evolution changes and the branch is behind/diverged.

2. 建议修复（非阻塞）

- Optional follow-up: add a targeted multi-table unsupported sub-writer unit test for the `CONTINUE_OTHER_TABLES` quarantine path. I would not block this PR on that because the shared helper and Flink wrapper already cover the core fail-fast behavior.

Overall, this is a good correctness improvement. It keeps the compatible legacy path while closing the dangerous default no-op path, which is the right direction for CDC schema evolution.

DanielLeens (approved):
Re-review at the current head `61be2cf60`. This branch has been rebased since my 2026-08-01 review (old head `f6c15414e` is no longer reachable via a normal ancestry diff — I fetched it explicitly to compare). The branch is now exactly one commit ahead of `origin/dev` and zero commits behind, which resolves the staleness I'd have otherwise had to account for.

Unchanged from my last review: a sink writer could previously inherit the deprecated no-op default `SinkWriter.applySchemaChange()`, receive a CDC `SchemaChangeEvent`, and silently drop it while newer-schema records kept flowing downstream. This PR centralizes the decision in `SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(...)`: writers that implement the interface are allowed, legacy writers that genuinely override the deprecated hook are still allowed, and writers that only inherit the no-op now fail fast instead of silently dropping the event.

# 1. Code Change Review

## 1.1 Core Logic Analysis

I isolated the PR's actual payload via `git diff origin/dev...HEAD` (rather than diffing against the old, now-unreachable head, which would otherwise be swamped by ~19 days of unrelated `dev` history) and confirmed it is exactly 12 files, matching what I reviewed on 2026-08-01 file-for-file and line-for-line where I spot-checked:

- `SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(...)` — unchanged: implements-interface -> call directly; legacy override detected via `getMethod(...).getDeclaringClass() != SinkWriter.class` -> call directly; neither -> `UnsupportedOperationException`. This reflection check runs once per schema-change event (a rare, DDL-driven event), not per record, so there is no hot-path cost.
- `SinkFlowLifeCycle.processSchemaChangeEvent` (Zeta), `ErrorHandlingSinkWriter.applySchemaChange`, and `FlinkSinkWriter.applySchemaChangeEvent` (Flink) all now route through the single `applySchemaChangeToWriter` choke point — confirmed identical to the 08-01 diff.
- `IcebergSinkWriter.applySchemaChange` now throws `UnsupportedOperationException` explicitly when `iceberg.table.schema-evolution-enabled` is false, instead of silently no-op-ing.
- `LanceSinkWriter` no longer implements `SupportSchemaEvolutionSinkWriter` at all (it previously advertised support via an empty override), so it now correctly falls through to the fail-fast path in the choke point rather than claiming false capability.

Runtime path (re-traced, unchanged from 08-01):
```
Zeta: Source emits SchemaChangeEvent -> SinkFlowLifeCycle.received() -> processSchemaChangeEvent()
  -> SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(writer, event)
Flink: SchemaOperator emits control row -> FlinkSinkWriter.write() -> handleSchemaChangeEvent()
  -> SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(sinkWriter, event)
MultiTableSinkWriter.applySchemaChange() -> per-table sub-writer -> same choke point,
  gated by FAIL_FAST / CONTINUE_OTHER_TABLES policy (unchanged, pre-existing)
```

## 1.2 Compatibility Impact

**Fully compatible**, unchanged assessment from 08-01: no API/SPI method, config option, protocol, or serialization format is removed or renamed. The deprecated `SinkWriter.applySchemaChange()` path remains available and legacy overrides are still honored. The only behavior change is that a writer which previously silently no-op'd a schema change now fails the job explicitly — a data-correctness improvement, not a compatibility break, since a silently-dropped schema change was never a documented or reliable contract to begin with.

No new thread pools, executors, or connections. The added reflection call in `overridesDeprecatedApplySchemaChange` is bounded to once per schema-change event.

Consistent, no issues found: the `UnsupportedOperationException` messages name the writer class and the affected table, which is exactly the context needed to diagnose a fail-fast schema-evolution rejection.

**Live CI status at `61be2cf60`:** checked the fork's (`danielnadean/seatunnel`) check-runs directly for this exact commit: 58 success, 20 still pending, 10 skipped, and exactly **one** failure — `all-connectors-it-7 (8, ubuntu-latest)`. I pulled that job's raw log rather than trusting the summary: the failure is entirely inside `connector-couchbase-e2e`, with `com.couchbase.client.core.error.UnambiguousTimeoutException: WaitUntilReady timed out in stage WAIT_FOR_CONFIG` after repeated `AuthenticationFailureException` connect attempts to the Couchbase test container. This PR does not touch Couchbase, Spark, or anything in that module — it is the known, already-tracked Couchbase container-bootstrap flake affecting `dev` itself since 2026-08-15 (a fix already exists upstream). This is environmental, not PR-caused, and rerunning it is unlikely to be productive; I would not block on it. The rest of the run should be allowed to finish before final merge, but nothing so far implicates this PR's own code.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Unchanged from 08-01: AOSP formatting, ASF headers, no wildcard imports, standard Javadoc on the new interface methods explaining the fail-fast rationale.

`SupportSchemaEvolutionSinkWriterTest` (152 lines), `IcebergSinkWriterTest` (+29), `ErrorHandlingSinkWriterTest` (+23), `FlinkSinkWriterTest` (+36) — all pure unit tests using mocks or in-memory records, matching what I confirmed stable on 08-01. No `Thread.sleep`, no shared mutable state, no order dependence. **Stable.**

## 2.3 Documentation Updates
No user-facing config/option change, so no `docs/en`/`docs/zh` update is required for this fix.

# 3. Architectural Soundness

**Precise fix**, unchanged assessment: centralizing the decision in one static choke point (`applySchemaChangeToWriter`) rather than duplicating the implements-check across Zeta, Flink, and `MultiTableSinkWriter` is the right shape, and it is what makes this fix trustworthy — there is exactly one place to audit for correctness rather than three independently-drifting call sites.

## 3.2 Maintainability
Improved — future sink writers only need to implement (or not implement) `SupportSchemaEvolutionSinkWriter`; the fail-fast behavior is automatic and centrally enforced rather than something each engine integration has to remember to check.

## 3.3 Extensibility
Unchanged — the interface-based capability check composes cleanly with `MultiTableSinkWriter`'s existing `FAIL_FAST`/`CONTINUE_OTHER_TABLES` per-table policy without needing any new plumbing.

No persisted format, checkpoint state, or serialization contract is touched by this change.

# 4. Issue Summary

No open issues. My 08-01 review found no source-level blockers, and this round's content is verified identical.

# 5. Merge Recommendation

**1. Blockers — must be fixed:**
- None at the source level.
- Let the in-flight fork CI run finish (20 jobs still pending at review time). The one failure so far (`connector-couchbase-e2e` container bootstrap) is a known, unrelated, already-tracked infrastructure flake on `dev` itself, not something this PR's diff touches or could plausibly cause.

- None.

**Overall assessment.** This PR's substance has not changed since my 08-01 review, where I found the approach solid and no source-level blockers — I independently re-verified that by isolating the true PR diff against current `dev` and confirming it matches file-for-file. The branch is now fully rebased (0 commits behind `dev`), which removes the staleness concern from before. The one CI failure I found is a well-understood, unrelated Couchbase container flake with an existing upstream fix, not a regression from this change. Once the remaining CI jobs finish (and assuming they come back clean, as everything so far suggests), this is ready to merge.

DanielLeens:
Thanks for keeping this moving. I want to open with a correction: on 2026-08-01 I reviewed this PR (COMMENTED) and again on 2026-08-20 I re-reviewed the current head and APPROVED it as "Fully compatible" with "no source-level blocker." Doing a completely fresh, independent pass today, I found a real compatibility gap that both of my earlier rounds missed. I'm sorry for approving past it — here's the honest re-analysis.

Before this PR, a sink writer could inherit the deprecated, default no-op `SinkWriter.applySchemaChange(SchemaChangeEvent)` method. If it then received a CDC `SchemaChangeEvent` (e.g. a source-side `ALTER TABLE ADD COLUMN`), the event was silently swallowed: no exception, no log warning at the call site, nothing — the job just kept running while its sink schema quietly drifted out of sync with the incoming data.

This PR closes that gap by routing every `applySchemaChange` call through a single new static helper, `SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(...)`, which allows the call only when the writer either (a) implements `SupportSchemaEvolutionSinkWriter`, or (b) genuinely overrides the deprecated method (detected via reflection). Any other writer now gets an explicit `UnsupportedOperationException` instead of a silent no-op. Iceberg's writer is also changed to throw explicitly when `iceberg.table.schema-evolution-enabled=false` instead of silently ignoring the event, and Lance no longer falsely advertises schema-evolution support via an empty override.

One-sentence summary: this PR turns a previously silent, hard-to-detect schema-drift failure mode into an explicit, fail-fast exception across Zeta, Flink, and multi-table sink dispatch.

# 1. Code Change Review

## 1.1 Core Logic Analysis

I independently re-verified the diff against the current PR head (`61be2cf60d75141a06817c2128eda3165e72e02c`) via `git diff` against the merge-base with `dev` — 12 files, matching what both of my prior rounds reported file-for-file.

Core logic, read directly from `seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/SupportSchemaEvolutionSinkWriter.java`:

```java
static void applySchemaChangeToWriter(SinkWriter<?, ?, ?> writer, SchemaChangeEvent event)
        throws IOException {
    if (writer instanceof SupportSchemaEvolutionSinkWriter) {
        ((SupportSchemaEvolutionSinkWriter) writer).applySchemaChange(event);
        return;
    }
    if (overridesDeprecatedApplySchemaChange(writer)) {
        writer.applySchemaChange(event);
        return;
    }
    throw new UnsupportedOperationException(...);
}
```

`overridesDeprecatedApplySchemaChange` uses `writer.getClass().getMethod("applySchemaChange", SchemaChangeEvent.class).getDeclaringClass() != SinkWriter.class` to detect a real legacy override. Since the interface method is implicitly public, any override must also be public, so `getMethod` (public-only lookup) is sufficient here — I don't see a case where a legitimate override would be missed. This runs once per schema-change event (a rare, DDL-driven event), not per record, so there's no hot-path cost.

**Simple before/after example**, traced myself against the real call chain rather than assumed: say a CDC pipeline writes MySQL into Iceberg, and the user never sets `iceberg.table.schema-evolution-enabled` (documented default is `false` — confirmed in `IcebergSinkOptions.java:47` and `docs/en/connectors/sink/Iceberg.md:76`). A DBA runs `ALTER TABLE ... ADD COLUMN` on the source.
- **Before this PR**: `IcebergSinkWriter.applySchemaChange` just returns early when the option is `false`. The job keeps running; the new column's data is effectively dropped downstream. Silent, but alive.
- **After this PR**: the same event now throws `UnsupportedOperationException("Iceberg sink received schema change event for table %s, but iceberg.table.schema-evolution-enabled is false.")`.

I traced where that exception actually goes, since that determines real-world blast radius:

```
Zeta direct (non-multi-table) sink path
  CDC source emits SchemaChangeEvent
  -> SinkFlowLifeCycle.received(record)                      (SinkFlowLifeCycle.java:268)
     -> record.getData() instanceof SchemaChangeEvent
     -> processSchemaChangeEvent(event)                       (SinkFlowLifeCycle.java:825)
        -> SupportSchemaEvolutionSinkWriter.applySchemaChangeToWriter(writer, event)
           -> unsupported writer -> throw UnsupportedOperationException
  -> received()'s own catch block:
        catch (RuntimeException e) { throw e; }                (SinkFlowLifeCycle.java:291-292)
  -> propagates out of received() uncaught -> sink task fails -> job fails
```

I read `received()` directly (lines 268-293) to confirm this: it catches `RuntimeException` and rethrows it as-is, and wraps any other `Exception` in a new `RuntimeException`. There is no "log and continue" branch for a schema-change failure at this layer. So for a single-table sink using the default config, this is a genuine job-crashing change, not a cosmetic one.

For **multi-table** sinks, I also checked `MultiTableSinkWriter.applySchemaChange` — the diff only swaps the old inline instanceof-check for the same shared helper inside the pre-existing `executeWithTableRetry(..., MultiTableFailurePhase.CHECKPOINT, ...)` wrapper, so the pre-existing `MultiTableFailurePolicy` (FAIL_FAST vs. CONTINUE_OTHER_TABLES) still governs blast radius here — that dispatch logic itself isn't touched by this PR. Worth knowing: I checked `MultiTableSinkWriter.java:126`, and one of the writer's constructor overloads defaults `failurePolicy` to `MultiTableFailurePolicy.FAIL_FAST`, so unless a job explicitly configured `CONTINUE_OTHER_TABLES`, the multi-table path also fails the whole sink, not just the affected table.

For **Flink**, I read the full current `FlinkSinkWriter.handleSchemaChangeEvent` (both the `-common` and `-20` copies are structurally identical). This is where I found something genuinely good that neither of my prior rounds called out: the **old** code returned early on an unsupported writer without ever calling `sendSchemaChangeAck(...)`:
```java
// before
if (!(sinkWriter instanceof SupportSchemaEvolutionSinkWriter)) {
    log.warn(...);
    return;   // no ack sent to LocalSchemaCoordinator at all
}
```
The **new** code always goes through a `try { applySchemaChangeToWriter(...) } catch (Exception e) { log.error(...) } finally { sendSchemaChangeAck(..., success) }` block, so the coordinator is now notified of failure even for unsupported writers, instead of potentially waiting forever for an ack that the old code path would never send. That's a real, independently-verified correctness improvement on top of the headline fail-fast change, and it's not mentioned in either of my earlier reviews.

**Key findings:**
- The single dispatch choke point is correctly wired into all five call sites (`SinkFlowLifeCycle`, `MultiTableSinkWriter`, `ErrorHandlingSinkWriter`, and both `FlinkSinkWriter` copies) — I read each one directly, not just the diff hunks.
- The reflection-based legacy-override detection is correct and bounded to a rare event type, not the record hot path.
- The Flink wrapper's ack-on-failure fix (via the `finally` block) closes a real coordinator-hang risk that existed before this PR, independent of the headline fail-fast behavior.
- The actual production-facing risk of this PR is not in its logic — it's that the resulting behavior change is a real backward-incompatibility for a specific, common historical configuration, and it isn't documented as one. See 1.2.

## 1.2 Compatibility Impact

**Partially incompatible** — this is a correction to my prior two rounds, both of which said "Fully compatible."

No API, SPI method, config option, or serialization format is removed or renamed, and `iceberg.table.schema-evolution-enabled`'s default value is unchanged (`false`). So in the narrow "did you rename/remove a contract" sense, my earlier assessment wasn't wrong.

But this project's own `CLAUDE.md` treats backward compatibility as covering "historical behavior," not just the config/API surface, and explicitly requires: *"Any incompatible change MUST be explicitly documented... in `docs/en/introduction/concepts/incompatible-changes.md`... include migration guidance... be clearly explained in the PR description."* I checked `docs/en/introduction/concepts/incompatible-changes.md` at this PR's head: it already has a dedicated (currently empty) `### Engine Behavior Changes` section, and this PR's change is exactly the kind of entry that section exists for — but no entry was added, and the PR description doesn't flag this as a breaking change either (it frames it purely as closing a documentation-vs-runtime gap).

The concrete production impact: any currently-running job that (a) uses a sink writer that doesn't implement `SupportSchemaEvolutionSinkWriter` and doesn't override the deprecated method (this includes Iceberg with the default `schema-evolution-enabled=false`, and any third-party/custom sink not on the small supported list), and (b) has a source that ever emits a `SchemaChangeEvent` — will go from "keeps running with a silently stale schema" to "job fails" after upgrading to this version, with **zero config or code change on the user's side**, and **no way to opt back into the old lenient behavior even temporarily** (I looked for an escape hatch — there isn't one). That's the right long-term direction (silent data drift is worse than a loud failure), but it's a real "surprise on upgrade" that this project's own policy says must be documented with migration guidance, and currently isn't.

No new thread pools, executors, connections, or locks. The added reflection call in `overridesDeprecatedApplySchemaChange` runs once per schema-change event (DDL-driven, rare), not per record — no measurable hot-path cost. No new resource-release concerns; `close()`/writer lifecycle paths are untouched by this diff.

No sensitive data in the new exception messages — they include the writer's class name and the table path, which is exactly the context needed to diagnose a fail-fast rejection, and nothing credential-like. See Issue 1 below for the one substantive gap I found (documentation of the behavior change, not the error-handling code itself, which is sound).

# 2. Code Quality Assessment

## 2.1 Coding Standards

AOSP formatting, ASF license headers present on the new test file, no wildcard imports. The two new static methods on `SupportSchemaEvolutionSinkWriter` both carry proper multi-line Javadoc explaining intent and the legacy-compat rationale — no missing-comment gap here.

I read the full content of all four new/changed test files in the diff (`SupportSchemaEvolutionSinkWriterTest` — 152 lines, new; `IcebergSinkWriterTest` — +29 lines; `ErrorHandlingSinkWriterTest` — +23 lines; `FlinkSinkWriterTest` — +36 lines). All are pure JUnit 5 + Mockito unit tests: no `Thread.sleep`, no Docker/network/fixed ports, no shared static mutable state between tests, no order dependence, deterministic assertions (exact exception type + message-contains checks). **Stability rating: Stable.**

Coverage is good for the headline behavior (supported-writer pass-through, legacy-override pass-through, unsupported-writer fail-fast, Iceberg-disabled fail-fast, Flink unsupported-writer failure-ack-then-exception) — I confirmed each of these by reading the actual assertions, not just the test names. One non-blocking gap, carried over from my own 2026-08-01 review and still true today: `MultiTableSinkWriter`'s unsupported-sub-writer fail-fast path (the diff at `MultiTableSinkWriter.java`) has no dedicated unit test of its own; it's only exercised indirectly through the shared helper's own tests. I wouldn't block on this — the shared helper is the actual logic under test, and the multi-table wiring is a thin pass-through — but it would be a good addition.

## 2.3 Documentation Updates

This is where Issue 1 actually lives (see 1.2): `docs/en/introduction/concepts/incompatible-changes.md` needs a new entry under its existing `### Engine Behavior Changes` heading, in the same `**[BREAKING]**` / Impact / Migration Guide format the file already uses for other entries (I read several existing entries in that file to confirm the expected format). `docs/en/connectors/sink/Iceberg.md`'s description of `iceberg.table.schema-evolution-enabled` (currently just "Setting to true enables Iceberg tables to support schema evolution...") should also gain a sentence on what happens when it's `false` and a schema change event arrives now (job failure, not silent ignore). The `docs/zh` counterparts need the same updates per this repo's own contribution guide.

# 3. Architectural Soundness

**Precise fix** for the stated problem. Centralizing the decision in one static choke point rather than duplicating the implements-check at each of the five call sites is the right shape — there's exactly one place to audit, and I confirmed all five call sites actually route through it rather than re-implementing the check locally.

## 3.2 Maintainability

Improved. New sink writers only need to decide whether to implement `SupportSchemaEvolutionSinkWriter`; the fail-fast behavior is automatic and centrally enforced rather than something every engine integration has to remember to re-check.

## 3.3 Extensibility

Good — the interface-based capability check composes cleanly with `MultiTableSinkWriter`'s existing `FAIL_FAST`/`CONTINUE_OTHER_TABLES` per-table policy without new plumbing, and with the Flink coordinator's existing ack protocol.

No persisted checkpoint/savepoint format, state schema, or serialization contract is touched. The compatibility concern here (see 1.2) is about *runtime behavior* for historically-configured jobs on upgrade, not about state-restore or serialization compatibility — those remain unaffected.

# 4. Issue Summary

|---|---|---|---|
| 1 | Fail-fast behavior for sinks that previously silently no-op'd a schema change (including any sink with `iceberg.table.schema-evolution-enabled=false`, the documented default) is a real behavior change for existing production jobs on upgrade, with no opt-out, but is not documented as an incompatible/breaking change in `docs/en/introduction/concepts/incompatible-changes.md` (which has an empty `Engine Behavior Changes` section ready for exactly this), nor called out as a migration item in the PR description or in `docs/en/connectors/sink/Iceberg.md`'s option description | `seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/SupportSchemaEvolutionSinkWriter.java`; `seatunnel-connectors-v2/connector-iceberg/.../IcebergSinkWriter.java:118-131`; `docs/en/introduction/concepts/incompatible-changes.md` (missing entry) | Medium |
| 2 | `MultiTableSinkWriter`'s unsupported-sub-writer fail-fast path has no dedicated unit test | `seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java` | Low |

# 5. Merge Recommendation

**1. Blockers — must be fixed:**
- Issue 1 (Medium): add the missing incompatible-change documentation entry (and the accompanying Iceberg doc-option sentence) before this leaves draft. This is a project policy requirement (`CLAUDE.md`'s backward-compatibility rules), not a code defect — the underlying behavior change itself is the right direction, it just needs to be written down with migration guidance so operators aren't surprised by a job failing after a routine upgrade.
- CI: live-checked just now against the fork (`danielnadean/seatunnel`) run for head `61be2cf60`: 76 success, 10 skipped, and 3 non-success — `paimon-connector-it (8, ubuntu-latest)` failed, `all-connectors-it-7 (8, ubuntu-latest)` failed, `kudu-connector-it (11, ubuntu-latest)` cancelled. None of those modules line up with what this diff touches (`seatunnel-api`, `seatunnel-engine-server`, `connector-iceberg`, `connector-lance`, `seatunnel-translation-flink`), which is consistent with the unrelated-IT-flake pattern seen in earlier rounds of this PR (rocketmq, then couchbase). I want to flag honestly, though: I was not able to pull the actual failure logs this round (the Actions log-streaming API returned a local stream error both times I tried), so this diagnosis is based on module/job-name correlation rather than a confirmed root-cause read — weaker evidence than my earlier rounds. Please paste the failing job link(s) here, or just resync and rerun (see next point) and I'll take another look.
- Branch sync: live-checked via `gh api repos/apache/seatunnel/compare/dev...61be2cf60...` just now — still `ahead_by=1`, `behind_by=16`, `diverged`. That's worse than what my 2026-08-20 review reported ("zero commits behind") — `dev` has moved on since then and this branch hasn't been resynced. Please sync with the latest `dev` and rerun CI; that will also refresh the two IT lanes above in case they've already been fixed upstream.

- Issue 2 (Low): add a targeted unit test for `MultiTableSinkWriter`'s unsupported sub-writer fail-fast path.

**Overall assessment.** The core approach is sound and I independently re-verified it end to end rather than trusting my own prior conclusion: one shared dispatch choke point, correctly wired into all five call sites, with a genuine bonus fix in the Flink wrapper (failure acks are now always sent, closing what looks like a real coordinator-hang gap in the old code). The tests are stable and cover the headline behavior well. The one real gap is that this PR changes historical runtime behavior for a common configuration (any sink without explicit schema-evolution support, including Iceberg's own documented default) without the incompatible-change documentation this project's own rules require — that's a quick, doc-only fix, not a sign the implementation is wrong, and I'd be glad to take another look as soon as it's added and CI is green against a resynced `dev`.

DanielLeens (commented):
Quick correction to my own last comment, since the CI fact I asserted there has since changed: I wrote "assuming they come back clean, as everything so far suggests, this is ready to merge" while 20 jobs were still in flight. That run has since finished, and it did not come back clean — the fork `Build` on this exact head (`61be2cf60d`) is now `FAILURE`, from two jobs:

- `all-connectors-it-7 (8, ubuntu-latest)` — the same Couchbase container-bootstrap flake I already identified and dismissed as unrelated in my last review (`UnambiguousTimeoutException` / `WAIT_FOR_CONFIG`, a known `dev`-wide environmental issue since 2026-08-15, already tracked with an upstream fix in flight).
- `paimon-connector-it (8, ubuntu-latest)` — new since my last check (this job was still pending when I reviewed). I pulled the raw log rather than trusting the summary: `PaimonSinkDynamicBucketIT.testParallelismBucketCount` fails with `Container exited with code 127` / `Inconsistency detected by ld.so: ../elf/dl-tls.c: 538: _dl_allocate_tls_init: Assertion \`listp->slotinfo[cnt].gen <= GL(dl_tls_generation)' failed!` inside the `seatunnelhub/openjdk:8u342` test container during startup — a glibc/dynamic-linker crash in the container image itself, before any Paimon-specific test logic runs. This PR's diff doesn't touch `connector-paimon` or anything in its dependency path (it only touches `seatunnel-api`, `seatunnel-engine-server`, `connector-iceberg`, `connector-lance`, and the Flink translation module), so this is not a regression caused by this change either.

Net effect: both failures are environmental/infrastructure flakes unrelated to this PR's schema-evolution fail-fast change, not a regression it introduced. My technical conclusion from the 2026-08-20 review (approved, no source-level blocker) stands unchanged. Concrete next step: please re-run just those two failed jobs (`all-connectors-it-7` and `paimon-connector-it`, both on JDK 8) rather than the full matrix — once they're green (or confirmed as the same known flakes on a rerun), this should be clear to merge from my side.

PR #11619, opened 2026-08-01
[Test][Core] Exclude build outputs from import class check

DanielLeens:
## What changed

- Exclude `target` build output directories from `ImportClassCheckTest`s repository-wide JavaParser scan.

## Why

- The latest fork CI failure on August 1, 2026 came from `seatunnel-ci-tools` on `unit-test (8, windows-latest)`.
- `ImportClassCheckTest` walks `..` and parses every `.java` file it finds.
- In the full `clean verify` unit-test workflow, earlier modules have already produced many generated Java files under `target`, so this test ends up parsing checked-in sources and build outputs together.
- That inflates the JavaParser workload and can trigger `GC overhead limit exceeded` in the Surefire fork JVM on hosted Windows JDK 8 runners.

## Impact

- Keep the import check focused on checked-in repository sources.
- Avoid repeatedly parsing generated build outputs that are not part of the source tree contract.

## Validation

- `./mvnw -f seatunnel-ci-tools/pom.xml spotless:apply`
- `git diff --check`
- Independent maintainer review of the final diff (static review only)

### DIFF
--- a/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java
+++ b/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java
@@ -30,6 +30,7 @@ import com.github.javaparser.ast.ImportDeclaration;
 import com.github.javaparser.ast.NodeList;
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


davidzollo:
Closing this PR because it was opened from the wrong maintainer fork by mistake. It is superseded by the daniel2-authored PR that carries the same fix.

PR #11621, opened 2026-08-01
[CI][Core] Exclude build outputs from import class check

danielnadean:
## What changed

- Exclude `target` build output directories from `ImportClassCheckTest`s repository-wide JavaParser scan.

## Why

- The latest fork CI failure on August 1, 2026 came from `seatunnel-ci-tools` on `unit-test (8, windows-latest)`.
- `ImportClassCheckTest` walks `..` and parses every `.java` file it finds.
- In the full `clean verify` unit-test workflow, earlier modules have already produced generated Java files under `target`, so this test ends up parsing checked-in sources and build outputs together.
- That inflates the JavaParser workload and can trigger `GC overhead limit exceeded` in the Surefire fork JVM on hosted Windows JDK 8 runners.

## Impact

- Keep the import check focused on checked-in repository sources.
- Avoid repeatedly parsing generated build outputs that are not part of the source tree contract.

## Validation

- `./mvnw -f seatunnel-ci-tools/pom.xml spotless:apply`
- `git diff --check`
- Independent maintainer review of the final diff (static review only)

### DIFF
--- a/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java
+++ b/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java
@@ -30,6 +30,7 @@ import com.github.javaparser.ast.ImportDeclaration;
 import com.github.javaparser.ast.NodeList;
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.FileVisitOption;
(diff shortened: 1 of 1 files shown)


davidzollo (approved):
+1
Good job

DanielLeens (commented):
Thanks for the contribution. I re-reviewed the latest head `6e79e412ccb9` locally against the real `ImportClassCheckTest` execution path.

I also noticed @davidzollo already left an approval on this PR. After my own pass, I largely agree with the source-side direction: I do not see a new code-side blocker from Daniel's side on the current revision.

## Runtime path I rechecked

```text
CI unit-test job
  -> build modules
      -> generate sources under */target/*
  -> ImportClassCheckTest.beforeAll()
      -> Files.walk("..")
      -> keep *.java files
      -> now skip paths containing /target/
  -> parse checked-in Java sources only
  -> run shaded/import rule assertions
```

## Current finding

### Issue 1: the current Build is still in progress, so this head still needs its final CI signal

- latest Build: `https://github.com/apache/seatunnel/runs/91394712546`

From the code side, this looks like a focused fix: the test is supposed to scan checked-in Java sources, not build outputs under `target/`. Excluding generated sources from the walk keeps the import check aligned with that intent and avoids repeatedly parsing compiled/generated artifacts during full unit-test jobs.

So I do not see a source-side blocker here. The remaining gate for me is simply to let the current Build finish on this exact head.

## Merge conclusion

1. Blocking items
- Let the current Build finish on the exact head.

2. Non-blocking suggestions
- No new source-side change is required from Daniel in this round.

Overall, this looks like a small and reasonable CI-test-scope fix. If the current Build stays green, I would not expect another source-level blocker from Daniel on this PR.

davidzollo (approved):
+1
Good job

PR #11623, opened 2026-08-02
[Test][API] Avoid ImportClassCheckTest OOM on generated sources

danielnadean:
## Why
- on Sunday, August 2, 2026, `apache/seatunnel#11198` is still failing in the fork `unit-test (8, ubuntu-latest)` lane with `java.lang.OutOfMemoryError: GC overhead limit exceeded` while running `seatunnel-ci-tools` `ImportClassCheckTest`
- `apache/seatunnel#11621` already excluded `target` build outputs, but the test can still parse checked-in generated protobuf sources and keep all JavaParser import nodes for the whole repository in memory

## What
- skip protobuf-generated Java sources detected by the `@@protoc_insertion_point` marker before parsing
- store lightweight import metadata instead of retaining JavaParser import nodes for every file

## Validation
- `./mvnw -pl seatunnel-ci-tools -DspotlessFiles=seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java spotless:apply`
- `git diff --check`

### DIFF
--- a/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java
+++ b/seatunnel-ci-tools/src/test/java/org/apache/seatunnel/api/ImportClassCheckTest.java
@@ -230,7 +230,8 @@ public class ImportClassCheckTest {
                                 imports.stream()
                                         .filter(
                                                 importMetadata -> {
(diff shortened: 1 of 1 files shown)


davidzollo (approved):
+1

PR #11634, opened 2026-08-03
[Fix][Format-JSON] Apply column defaultValue when JSON field is null or missing (#11632)

### LINKED ISSUE #11632: [Bug][Connector-JSON] Schema defaultValue is ignored during JSON deserialization (fields remain null)
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/seatunnel/issues?q=is%3Aissue+label%3A%22bug%22) and found related but **different-root-cause** reports:
  - [#8759](https://github.com/apache/seatunnel/issues/8759) — Oracle CDC → Kafka: Debezium struct default value used when column is null (Debezium deserialization path).
  - [PR #7950](https://github.com/apache/seatunnel/commit/3b432125a) — fixed the Debezium path via `struct.getWithoutDefault(...)`; the **JSON format** deserializer (`seatunnel-format-json`) is a separate code path and was NOT touched.
  - [PR #6247](https://github.com/apache/seatunnel/pull/6247) — fixed `Object.class` option value parsing so `defaultValue` config can be read correctly.
  - No existing issue reports the JSON format deserializer ignoring `schema.columns[].defaultValue`.

### What happened

When a source connector using the **JSON format** (Kafka, Pulsar, HTTP, MongoDB, Elasticsearch, …) configures `schema { columns = [ { name = "GROUPID", type = "int", nullable = false, defaultValue = 0 } ] }`, the `defaultValue` is **parsed and stored on the `Column`** but **never applied** during deserialization: if the JSON message is missing the field or contains an explicit `null`, the row field stays `null` instead of the configured default value.

Expected: `defaultValue = 0` is used when the JSON field is missing or `null`.
Actual: the field is `null`; the `defaultValue` config is silently ignored.

### SeaTunnel Version

2.3.13 (verified locally). **Also reproducible on current upstream `dev`** (checked 2026-08-03, `6bf786bee7`).

### SeaTunnel Config

```hocon
source {
  Kafka {
    bootstrap.servers = "localhost:9092"
    topic = "test"
    format = json
    schema {
      columns = [
        { name = "GROUPID", type = "int", nullable = false, defaultValue = 0 }
        { name = "name", type = "string", nullable = false, defaultValue = "unknown" }
      ]
    }
  }
}

sink {
  Console {}
}
```

### Running Command

```shell
./bin/seatunnel.sh --config ./config/xxx.conf
```

### Error Exception

No error. The job runs, but fields configured with `defaultValue` come out as `null` when the JSON message omits them or sets them to `null` — a silent data-integrity bug.

### Root cause (confirmed by code inspection on upstream dev)

The `defaultValue` is lost on the way from `CatalogTable` to the row converter:

```
CatalogTable (contains Column metadata incl. defaultValue)
    ↓ getSeaTunnelRowType()
SeaTunnelRowType (fieldNames[] + fieldTypes[] only) ← defaultValue dropped here
    ↓
JsonToRowConverters (no access to Column[] / defaultValue)
    ↓
convertField(): field == null → returns null (defaultValue never consulted)
```

Concretely, on upstream `dev` (same as 2.3.13):

1. `JsonToRowConverters.convertField(...)` (lines ~474-486) — for a missing field it only checks `failOnMissingField` and otherwise returns `null`; it has **no `Column[]` metadata** and never checks `getDefaultValue()`:

```java
private Object convertField(
        JsonToObjectConverter fieldConverter, String fieldName, JsonNode field) {
    if (field == null) {
        if (failOnMissingField) {
            throw new IllegalArgumentException(...);
        } else {
            return null; // defaultValue never applied
        }
    } else {
        return fieldConverter.convert(field, fieldName);
    }
}
```

2. `JsonDeserializationSchema` (both constructors, lines ~82 and ~103) builds `new JsonToRowConverters(failOnMissingField, ignoreParseErrors)` and never passes `catalogTable.getTableSchema().getColumns()` to the converters.

3. The configuration side works correctly: `ReadonlyConfigParser` (line ~139) reads `columnConfig.get(ConnectorCommonOptions.DEFAULT_VALUE)` into `PhysicalColumn.of(...)`, and `Column.getDefaultValue()` exists in `seatunnel-api`. The API contract explicitly supports default values — the deserialization side just never uses them.

### Proposed fix (implemented & verified locally on 2.3.13)

Minimal, backward-compatible, root-cause fix:

- `JsonToRowConverters`: add a `Column[] columns` field and a new constructor `JsonToRowConverters(boolean failOnMissingField, boolean ignoreParseErrors, Column[] columns)`; add a `convertField(converter, fieldName, field, fieldIndex)` overload that consults `columns[fieldIndex].getDefaultValue()` when the field is missing or `null`, falling back to the original `failOnMissingField`/`null` behavior when no default value is configured.
- `JsonDeserializationSchema`: extract `Column[]` from `catalogTable.getTableSchema().getColumns()` (with null-safety) and pass it to the new constructor.
- New test `JsonDefaultValueTest`: missing field → defaultValue; explicit `null` → defaultValue; no defaultValue → `null` (backward compatibility); string-typed default values.

Verification on 2.3.13: 3 new unit tests + full `seatunnel-format-json` module (46 tests) all pass; `spotless:apply` clean.

zeta (format module is engine-agnostic; affects all of them)

### Java or Scala Version

Java 8 / 11 (both OK)

### Screenshots

_No response_

- [X] Yes I am willing to submit a PR! (fix already implemented and verified locally; will port to `dev`)

waterWang:
### What problem does this PR solve?

Issue Number: close #11632

### Problem Summary

When a source connector using the JSON format (Kafka, Pulsar, HTTP, etc.) configures `schema { columns = [ { name = "GROUPID", type = "int", defaultValue = 0 } ] }`, the `defaultValue` is parsed and stored on the `Column` but never applied during deserialization: if the JSON message is missing the field or contains an explicit `null`, the row field stays `null` instead of the configured default value.

### Root Cause

The `defaultValue` is lost on the way from `CatalogTable` to the row converter:

1. `CatalogTable.getSeaTunnelRowType()` calls `toPhysicalRowDataType()`, which creates a `SeaTunnelRowType` with only `fieldNames[]` and `fieldTypes[]` — the `defaultValue` is dropped.
2. `JsonToRowConverters` builds converters from `SeaTunnelRowType` and has no access to `Column` metadata.
3. When `convertField()` encounters a `null` or missing field, it returns `null` without consulting the default value.

### What Changed

1. **`JsonToRowConverters`**: Added an optional `defaultValues` array parameter. When a field is `null`/missing and a non-null default value exists for that column, the default value is used instead of `null`.

2. **`JsonDeserializationSchema`**: Extracts default values from `CatalogTable.getTableSchema().getColumns()` (filtering to physical columns to match `toPhysicalRowDataType()` behavior) and passes them to the converters.

### How was this tested?

- Existing tests pass without modification (the `defaultValues` array is `null` by default, preserving backward compatibility).
- Manually verified that the scenario from the issue (Kafka JSON source with `defaultValue = 0` / `defaultValue = "unknown"`) now correctly applies defaults when fields are missing or `null`.

### DIFF
--- a/seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonDeserializationSchema.java
+++ b/seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonDeserializationSchema.java
@@ -29,6 +29,7 @@ import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.node.NullNode;
 import org.apache.seatunnel.api.serialization.DeserializationSchema;
 import org.apache.seatunnel.api.source.Collector;
 import org.apache.seatunnel.api.table.catalog.CatalogTable;
+import org.apache.seatunnel.api.table.catalog.Column;
 import org.apache.seatunnel.api.table.catalog.TablePath;
 import org.apache.seatunnel.api.table.type.CompositeType;
 import org.apache.seatunnel.api.table.type.SeaTunnelDataType;
@@ -40,6 +41,7 @@ import org.apache.seatunnel.common.exception.CommonErrorCodeDeprecated;
 import org.apache.seatunnel.format.json.exception.SeaTunnelJsonFormatException;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Optional;
 
 import static org.apache.seatunnel.shade.com.google.common.base.Preconditions.checkNotNull;
@@ -99,8 +101,24 @@ public class JsonDeserializationSchema implements DeserializationSchema<SeaTunne
         this.rowType = checkNotNull(catalogTable.getSeaTunnelRowType());
         this.failOnMissingField = failOnMissingField;
         this.ignoreParseErrors = ignoreParseErrors;
+
+        // Extract default values from the catalog table schema.
+        // Only physical columns have defaults; toPhysicalRowDataType() filters
+        // non-physical columns, so we apply the same filter here.
+        Object[] defaultValues = null;
+        List<Column> columns = catalogTable.getTableSchema().getColumns();
+        if (columns != null) {
+            defaultValues = new Object[columns.size()];
+            int physIdx = 0;
+            for (int i = 0; i < columns.size(); i++) {
+                if (columns.get(i).isPhysical()) {
+                    defaultValues[physIdx++] = columns.get(i).getDefaultValue();
+                }
+            }
+        }
+
         this.runtimeConverter =
-                new JsonToRowConverters(failOnMissingField, ignoreParseErrors)
+                new JsonToRowConverters(failOnMissingField, ignoreParseErrors, defaultValues)
                         .createRowConverter(checkNotNull(rowType));
 
         if (hasDecimalType(rowType)) {
--- a/seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonToRowConverters.java
+++ b/seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonToRowConverters.java
@@ -78,11 +78,20 @@ public class JsonToRowConverters implements Serializable {
     /** Flag indicating whether to ignore invalid fields/rows (default: throw an exception). */
     private final boolean ignoreParseErrors;
 
+    /** Default values for each column, indexed by physical column position. */
+    private final Object[] defaultValues;
+
     public Map<String, DateTimeFormatter> fieldFormatterMap = new HashMap<>();
 
     public JsonToRowConverters(boolean failOnMissingField, boolean ignoreParseErrors) {
+        this(failOnMissingField, ignoreParseErrors, null);
+    }
+
+    public JsonToRowConverters(
+            boolean failOnMissingField, boolean ignoreParseErrors, Object[] defaultValues) {
         this.failOnMissingField = failOnMissingField;
         this.ignoreParseErrors = ignoreParseErrors;
+        this.defaultValues = defaultValues;
     }
 
     /** Creates a runtime converter which is null safe. */
@@ -411,6 +420,12 @@ public class JsonToRowConverters implements Serializable {
                             fieldName = rowFieldName + "." + fieldName;
                         }
                         Object convertedField = convertField(fieldConverters[i], fieldName, field);
+                        if (convertedField == null
+                                && defaultValues != null
+                                && i < defaultValues.length
+                                && defaultValues[i] != null) {
+                            convertedField = defaultValues[i];
+                        }
                         row.setField(i, convertedField);
                     } catch (Throwable t) {
                         throw CommonError.jsonOperationError(


DanielLeens (commented):
# What Problem This PR Solves
- User pain point
  When a JSON record omits a field or sends `null`, the current JSON deserializer leaves the SeaTunnel row field as `null` even if the catalog column declares a default value.
- Fix approach
  This PR reads `Column.defaultValue` from the catalog schema and applies it when the converted JSON field result is `null`.
- One-sentence summary
  The goal is to make JSON source deserialization honor catalog column defaults for missing or null fields.

# I. Code Change Review

## 1.1 Core Logic Analysis

### Exact change description
The core changes are in:
- `seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonDeserializationSchema.java:105-122`
- `seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonToRowConverters.java:81-95`
- `seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonToRowConverters.java:422-428`

### Before / after snippets
Before:

```java
this.runtimeConverter =
        new JsonToRowConverters(failOnMissingField, ignoreParseErrors)
                .createRowConverter(checkNotNull(rowType));
```

After:

```java
Object[] defaultValues = ... columns.get(i).getDefaultValue();
this.runtimeConverter =
        new JsonToRowConverters(failOnMissingField, ignoreParseErrors, defaultValues)
                .createRowConverter(checkNotNull(rowType));
```

Before:

```java
Object convertedField = convertField(fieldConverters[i], fieldName, field);
row.setField(i, convertedField);
```

After:

```java
Object convertedField = convertField(fieldConverters[i], fieldName, field);
if (convertedField == null && defaultValues != null && i < defaultValues.length
        && defaultValues[i] != null) {
    convertedField = defaultValues[i];
}
row.setField(i, convertedField);
```

### Key findings
- The normal JSON-to-row path will hit this change immediately in connectors that construct `JsonDeserializationSchema` from a `CatalogTable`.
- The PR injects catalog defaults after conversion, but it does not convert the default value through the field's runtime converter.
- Existing API evidence already shows catalog defaults are not always stored in the final SeaTunnel runtime type.
- So the fix only works when `Column.defaultValue` already happens to be in the target runtime type; it is incorrect for typed defaults such as dates/timestamps represented as strings.

### Deep logic analysis
The execution path here is:

```text
JSON source connector
  -> new JsonDeserializationSchema(catalogTable, ...)
      -> JsonDeserializationSchema extracts Column.defaultValue from catalog schema
      -> passes raw defaultValues[] into JsonToRowConverters

Row conversion
  -> JsonToRowConverters.createRowConverter()
      -> convertField(fieldConverters[i], fieldName, field)
      -> if result is null, directly assign defaultValues[i]
      -> SeaTunnelRow.setField(i, convertedField)
```

The bug is in the last step: `defaultValues[i]` is assigned as-is, without going through the same type-specific conversion logic used for JSON field values.

That is observable from existing parser coverage in `seatunnel-api/src/test/java/org/apache/seatunnel/api/table/catalog/schema/ReadonlyConfigParserTest.java:263-268`, where a date default is still stored as the string `"2020-01-01"` rather than a `LocalDate`. With this PR, a JSON date field that is null/missing would therefore put a `String` into a `DATE` column slot of `SeaTunnelRow`, which violates the row type contract and can break downstream operators that expect the runtime type.

### Full execution path

```text
Connector source reads JSON
  -> JsonDeserializationSchema(CatalogTable, failOnMissingField, ignoreParseErrors)
      -> JsonDeserializationSchema.java:109-118 collects Column.defaultValue
      -> JsonToRowConverters(..., defaultValues).createRowConverter(rowType)

Per-field conversion
  -> JsonToRowConverters.createRowConverter()
      -> jsonNode.get(fieldName)
      -> convertField(fieldConverters[i], fieldName, field)
          -> normal path converts JSON into runtime field type
      -> PR-added fallback at JsonToRowConverters.java:423-428
          -> if convertedField == null, assign raw defaultValues[i]

Risk window
  -> Column.defaultValue can still be a config-level literal
      -> ReadonlyConfigParserTest.java:268 shows a DATE default stored as \"2020-01-01\"
  -> PR bypasses type conversion for that fallback
      -> SeaTunnelRow gets a String in a DATE/TIMESTAMP/other typed slot
```

Local verification results used for this review:
- `gh pr view 11634 --json ...`: completed, used to confirm the latest head (`c070511e06f3d2ef6585383501f1064287e2a93f`) and current checks.
- `gh api repos/apache/seatunnel/compare/dev...c070511e06f3d2ef6585383501f1064287e2a93f`: completed, current compare status is `ahead`, `ahead_by=2`, `behind_by=0`.
- `gh api repos/apache/seatunnel/commits/c070511e06f3d2ef6585383501f1064287e2a93f/check-runs`: completed. The `Build` check is `action_required` with title `Workflow run detection failed`.
- `git diff upstream/dev...seatunnel-review-11634 -- <files>`: completed for the full changed set.
- Build / unit tests / E2E execution: not run locally. Per this batch's SeaTunnel review rule, I relied on source analysis and current GitHub check state.

## 1.2 Compatibility Impact
Judgment: **Partially incompatible**

Impacts:
- API: no public API signature change.
- Configuration: no config key change.
- Default behavior: yes, but the new behavior is not type-safe for every column type.
- Protocol: no protocol change.
- Serialization format: no direct byte-format change, but the runtime `SeaTunnelRow` field type can become inconsistent with the declared schema.
- Historical behavior: downstream code that assumes typed row fields can now receive raw config literals when defaults are used.

- CPU / memory / GC: negligible overhead from the fallback lookup itself.
- Network: no direct impact.
- Concurrency safety: unchanged.
- Retry / idempotency: unchanged.
- Resource release: unchanged.
- Main side effect: schema-typed row fields can now silently carry raw config literal objects instead of runtime-converted values.

Issue 1: Default values are injected without type conversion
- Location:
  - `seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonDeserializationSchema.java:108-122`
  - `seatunnel-formats/seatunnel-format-json/src/main/java/org/apache/seatunnel/format/json/JsonToRowConverters.java:422-428`
  - supporting evidence: `seatunnel-api/src/test/java/org/apache/seatunnel/api/table/catalog/schema/ReadonlyConfigParserTest.java:263-268`
- Problem description:
  The PR collects `Column.defaultValue` from the catalog schema and assigns it directly when a field converts to `null`, but it never passes that fallback through the field's runtime converter. In current SeaTunnel code, catalog defaults are not guaranteed to already be stored in the final runtime type. For example, the existing parser test shows a date default still stored as the string `"2020-01-01"`. That means this fix can place a raw string into a `DATE` or `TIMESTAMP` field of `SeaTunnelRow`.
- Potential risk:
  Downstream transforms/sinks can observe row values whose runtime Java type no longer matches the declared schema. That is a real correctness issue on the main deserialization path.
  Apply the default through the same field-specific converter path instead of assigning the raw catalog object. For typed literals, that likely means converting from the config literal representation into the row runtime type before `row.setField(i, ...)`.
- Severity: High

Issue 2: The current Build gate is blocked because no runnable workflow was detected for this fork head
- Location:
  - Build check: `https://github.com/apache/seatunnel/runs/91706443832`
- Problem description:
  The current `Build` check is `action_required` with the GitHub message `Workflow run detection failed`, which means the PR still does not have a runnable current-head backend workflow result.
- Potential risk:
  Even after the source fix above, the PR still cannot merge until the contributor-side workflow gate is restored.
  Please enable GitHub Actions in the fork if they are currently disabled, then push an empty commit to retrigger the workflow. Since this branch is not behind `dev` (`behind_by=0`), the immediate next step is the workflow enable/retrigger path rather than a sync-first instruction.
- Severity: High

# II. Code Quality Assessment

## 2.1 Coding Standards
The added comments in `JsonDeserializationSchema` explain the physical-column alignment clearly. The style is consistent overall.

I do not see any new regression test covering the key typed-default path on `JsonDeserializationSchema(CatalogTable, ...)`, even though this PR changes a deserialization contract. Existing tests already exercise that constructor (`JsonRowDataSerDeSchemaTest.java:186-188`), so this change should add coverage there for at least one typed default such as `DATE` or `TIMESTAMP`.

- No UT/E2E code was changed in this PR, so I did not find a new flaky-test pattern introduced here.

## 2.3 Documentation Updates
No user-facing documentation update is needed if the implementation is fixed in place, because this is a bug fix to existing default-value behavior rather than a new feature contract.

# III. Architecture Rationality

The intent is precise, but the current implementation stops one step too early: it adds a fallback value without integrating it into the existing type-conversion path.

## 3.2 Maintainability
The fallback hook is small and localized, which is good. The maintainability issue is correctness: future readers could assume `defaultValues[i]` is already safe for the row runtime type, but the current API does not guarantee that.

## 3.3 Extensibility
Once fixed through the existing converters, this can become a reusable default-value mechanism for more JSON-based sources. In its current raw-assignment form, it is brittle across more data types.

## 3.4 Historical Version Compatibility
This should remain backward compatible as a bug fix, but only after the default fallback is converted into the declared runtime type.

# IV. Issue Summary

#### Blocker Exhaustiveness Self-Check
- Full changed diff reviewed.
- Main deserialization call chain traced from `JsonDeserializationSchema(CatalogTable, ...)` into `JsonToRowConverters.createRowConverter()`.
- Current CI gate reviewed separately from source findings.
- No second source-side blocker beyond the raw default-type injection was found in this version.

| --- | --- | --- | --- |
| 1 | Default values are injected without type conversion | `JsonDeserializationSchema.java:108-122`; `JsonToRowConverters.java:422-428` | High |
| 2 | Current Build gate is still `action_required` for this fork head | `https://github.com/apache/seatunnel/runs/91706443832` | High |

# V. Merge Recommendation
### Conclusion: Can be merged after fixes

1. Blocking items (must fix)
- Issue 1: convert catalog defaults into the field runtime type before inserting them into `SeaTunnelRow`.
- Issue 2: enable/retrigger the contributor-side workflow so the current head gets a runnable Build result.

2. Suggested improvements (non-blocking)
- Add a regression test in `JsonRowDataSerDeSchemaTest` that covers typed defaults on the `CatalogTable` constructor path.

Overall assessment:
The direction makes sense, but the current implementation is not yet safe for typed defaults, and the PR still lacks a runnable current-head Build workflow.

- Reuse the existing field-specific conversion path for default literals instead of assigning the raw `Column.defaultValue` object.

davidzollo:

```

```

PR #11636, opened 2026-08-03
[FIX][CI] increase time and memory management

srijan-singh:
### Purpose

This pull request addresses CI failures by updating memory consumption and increasing timeout. It also updates the Maven command for running unit tests.

### Changes Made

* Updated Maven command to use OS-based settings:
	+ Added separate tasks for Linux/macOS and Windows
	+ Set MAVEN_OPTS and SUREFIRE_JVM_ARGS environment variables
* Increased timeout-minutes for `all-connectors-it-3` from 210 to 240
* Updated POM file with `surefire.jvm.args` configuration

### Testing and Validation

* This patch was tested as part of commit https://github.com/apache/seatunnel/pull/11198/commits/eb5ced99596e7b88f38740c8d9ac6a0ccd7cc631
* A build was triggered and completed successfully: https://github.com/srijan-singh/seatunnel/actions/runs/30786984078

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -473,9 +473,17 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
       - name: run all modules unit test
-        run: ./mvnw -B -T 1 clean verify -DskipUT=false -DskipIT=true -D"license.skipAddThirdParty"=true -D"skip.ui"=true -DargLine="-Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8" --no-snapshot-updates
+        run: ./mvnw -B -T 1 clean verify -DskipUT=false -DskipIT=true -D"license.skipAddThirdParty"=true -D"skip.ui"=true "-Dsurefire.jvm.args=$SUREFIRE_JVM_ARGS" --no-snapshot-updates
         env:
           MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8
+          SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
+        if: runner.os != 'Windows'
+      - name: run all modules unit test (Windows)
+        run: ./mvnw -B -T 1 clean verify -DskipUT=false -DskipIT=true -D"license.skipAddThirdParty"=true -D"skip.ui"=true "-Dsurefire.jvm.args=$env:SUREFIRE_JVM_ARGS" --no-snapshot-updates
+        env:
+          MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8
+          SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
+        if: runner.os == 'Windows'
 
   updated-modules-integration-test-part-1:
     needs: [ changes, sanity-check ]
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -470,10 +470,9 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
       - name: run all modules unit test
-        run: |
-          ./mvnw -B -T 1 clean verify -DskipUT=false -DskipIT=true -D"license.skipAddThirdParty"=true -D"skip.ui"=true --no-snapshot-updates
+        run: ./mvnw -B -T 1 clean verify -DskipUT=false -DskipIT=true -D"license.skipAddThirdParty"=true -D"skip.ui"=true -DargLine="-Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8" --no-snapshot-updates
         env:
-          MAVEN_OPTS: -Xmx4096m -Dfile.encoding=UTF-8
+          MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8
 
   updated-modules-integration-test-part-1:
     needs: [ changes, sanity-check ]
@@ -940,7 +939,7 @@ jobs:
       matrix:
         java: [ '8', '11' ]
         os: [ 'ubuntu-latest' ]
-    timeout-minutes: 210
+    timeout-minutes: 240
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK ${{ matrix.java }}
--- a/pom.xml
+++ b/pom.xml
@@ -670,6 +670,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven-surefire-plugin.version}</version>
                     <configuration>
+                        <argLine>${argLine}</argLine>
                         <skip>${skipUT}</skip>
                         <systemPropertyVariables>
                             <jacoco-agent.destfile>${project.build.directory}/jacoco.exec</jacoco-agent.destfile>
@@ -701,6 +702,7 @@
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <version>${maven-failsafe-plugin.version}</version>
                     <configuration>
+                        <argLine>${argLine}</argLine>
                         <skip>${skipIT}</skip>
                     </configuration>
                     <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,7 @@
         <jakarta.servlet-api>4.0.4</jakarta.servlet-api>
         <hugegraph.client.version>1.5.0</hugegraph.client.version>
         <!-- Option args -->
+        <surefire.jvm.args>-Dfile.encoding=UTF-8</surefire.jvm.args>
         <skipUT>false</skipUT>
         <skipIT>true</skipIT>
         <elasticsearch>7</elasticsearch>
@@ -670,7 +671,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven-surefire-plugin.version}</version>
                     <configuration>
-                        <argLine>${argLine}</argLine>
+                        <argLine>${surefire.jvm.args}</argLine>
                         <skip>${skipUT}</skip>
                         <systemPropertyVariables>
                             <jacoco-agent.destfile>${project.build.directory}/jacoco.exec</jacoco-agent.destfile>
@@ -702,7 +703,7 @@
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <version>${maven-failsafe-plugin.version}</version>
                     <configuration>
-                        <argLine>${argLine}</argLine>
+                        <argLine>${surefire.jvm.args}</argLine>
                         <skip>${skipIT}</skip>
                     </configuration>
                     <executions>


DanielLeens (commented):
# What Problem This PR Solves
- User pain point
  The backend unit-test workflow can hit memory pressure on GitHub-hosted runners, which makes CI less stable.
- Fix approach
  This PR lowers `MAVEN_OPTS`, moves a larger heap setting into `-DargLine`, and extends the containerized test timeout.
- One-sentence summary
  The goal is to stabilize backend CI memory behavior.

# I. Code Change Review

## 1.1 Core Logic Analysis

### Exact change description
The change is limited to `.github/workflows/backend.yml`, mainly:
- `backend.yml:475-479`
- `backend.yml:943-946`

### Before / after snippets
Before:

```yaml
run: |
  ./mvnw ... clean verify ...
env:
  MAVEN_OPTS: -Xmx4096m -Dfile.encoding=UTF-8
```

After:

```yaml
run: ./mvnw ... clean verify ... -DargLine="-Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8"
env:
  MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8
```

### Key findings
- The intended memory increase for test JVMs is not wired through the current root Maven test-plugin configuration.
- The root `maven-surefire-plugin` and `maven-failsafe-plugin` configuration does not consume `${argLine}` at all.
- So this PR lowers Maven's own heap from 4G to 1.5G without reliably increasing the test-process heap for most modules.
- The timeout bump to 240 minutes is straightforward, but the heap change is not functionally connected to the actual test JVMs on the main backend job.

### Deep logic analysis
The backend job runs:

```text
.github/workflows/backend.yml
  -> run all modules unit test
      -> ./mvnw -B -T 1 clean verify ... -DargLine="..."
      -> MAVEN_OPTS now only gives Maven itself 1.5G heap
```

For the `-DargLine` part to help, the repository-wide test plugin setup must pass that property into Surefire/Failsafe. But the root configuration in `pom.xml:670-700` and `pom.xml:701-713` does not define `<argLine>${argLine}</argLine>` or any equivalent handoff. The only repo hits for `argLine` are connector-specific Lance module poms, not the global backend test path.

That means the main unit-test workflow does not actually gain the promised 3G test JVM heap from this change, while Maven itself loses heap headroom immediately.

### Full execution path

```text
Backend CI workflow
  -> backend.yml:475-479
      -> Maven verify process runs with MAVEN_OPTS=-Xmx1536m
      -> PR passes -DargLine=... on the command line

Root test-plugin configuration
  -> pom.xml:670-700
      -> maven-surefire-plugin config has no <argLine>${argLine}</argLine>
  -> pom.xml:701-713
      -> maven-failsafe-plugin config also does not consume ${argLine}

Result
  -> most modules on the backend job still do not receive the extra test-JVM heap
  -> Maven/compilation process itself now has less memory than before
  -> the claimed memory-management fix is ineffective on the primary path
```

Local verification results used for this review:
- `gh pr view 11636 --json ...`: completed, used to confirm the latest head (`da34d085bc80b2c9c7c23f3ebf87df09fc524dc7`) and current queued Build check.
- `gh api repos/apache/seatunnel/compare/dev...da34d085bc80b2c9c7c23f3ebf87df09fc524dc7`: completed, current compare status is `ahead`, `ahead_by=2`, `behind_by=0`.
- `git diff upstream/dev...seatunnel-review-11636 -- .github/workflows/backend.yml`: completed.
- `rg -n "argLine" ...` and `sed -n '660,725p' pom.xml`: completed, used to verify the root Surefire/Failsafe config does not consume `${argLine}`.
- Build / unit tests / E2E execution: not run locally. Per this batch's SeaTunnel review rule, I relied on source analysis and current GitHub check state.

## 1.2 Compatibility Impact
Judgment: **Fully compatible**

Impacts:
- API: none.
- Configuration: none for end users.
- Default behavior: CI behavior only.
- Protocol / serialization / historical runtime behavior: none directly.

- CPU / memory / GC: on the intended path, this PR does not reliably increase test-JVM heap, but it definitely reduces Maven process heap from 4G to 1.5G.
- Network: none.
- Concurrency safety: none.
- Retry / idempotency: none.
- Resource release: none.
- Main side effect: the primary workflow may become more memory-constrained rather than less.

Issue 1: The new `-DargLine` heap setting is not consumed by the root backend test path
- Location:
  - `.github/workflows/backend.yml:475-479`
  - `pom.xml:670-700`
  - `pom.xml:701-713`
- Problem description:
  The PR moves the larger heap budget into `-DargLine`, but the repository-wide Surefire/Failsafe configuration does not pass `${argLine}` into test JVMs. As a result, the main backend workflow does not actually receive the added test heap from this change.
- Potential risk:
  The job keeps the same effective test-JVM memory behavior while Maven itself loses heap headroom (`4096m -> 1536m`). That can make OOMs or memory-related instability worse instead of better.
  Either wire the desired JVM args through the actual root Surefire/Failsafe configuration used by this job, or keep the memory budget in the environment variable that the current backend workflow really consumes.
- Severity: High

# II. Code Quality Assessment

## 2.1 Coding Standards
The workflow YAML remains readable and scoped, and the timeout bump is straightforward.

This PR does not add UT/E2E code. The missing piece is verification of the actual Maven/Surefire handoff path: the workflow change is not backed by repository configuration that would make `${argLine}` effective on the main backend job.

- No UT/E2E test code was changed, so there is no new flaky-test pattern introduced here.

## 2.3 Documentation Updates
No documentation change is needed because this is CI-only work.

# III. Architecture Rationality

The timeout increase is fine. The memory change is not elegant because it targets a property that the main workflow path does not actually consume.

## 3.2 Maintainability
As written, future maintainers may assume the backend workflow now gives test JVMs 3G heap, but that assumption is false on the current root Maven setup.

## 3.3 Extensibility
Once wired through the real plugin configuration, the idea could be reusable. In its current form, it is brittle and misleading.

## 3.4 Historical Version Compatibility
No historical runtime compatibility concern for users; the problem is workflow effectiveness.

# IV. Issue Summary

#### Blocker Exhaustiveness Self-Check
- Full changed diff reviewed.
- Workflow execution path traced into the repository's root Surefire/Failsafe configuration.
- Current CI queue state checked separately from source findings.
- No second independent source blocker was found beyond the ineffective memory handoff.

| --- | --- | --- | --- |
| 1 | The new `-DargLine` heap setting is not consumed by the root backend test path | `backend.yml:475-479`; `pom.xml:670-700`; `pom.xml:701-713` | High |

# V. Merge Recommendation
### Conclusion: Can be merged after fixes

1. Blocking items (must fix)
- Issue 1: wire the heap setting through the actual root test-plugin configuration, or keep the memory change on the path that this job already consumes.

2. Suggested improvements (non-blocking)
- None beyond the blocker above. The timeout increase itself looks reasonable.

Overall assessment:
The motivation makes sense, but the current implementation does not actually apply the intended memory increase to the main backend test path.

- Update the root Surefire/Failsafe configuration to consume the intended JVM argument property, then keep the workflow change aligned with that contract.

DanielLeens (changes requested):
Thanks for coming back so quickly on this, and welcome — I can see this is one of your first PRs here. I re-reviewed the latest head `36e92bc6` from scratch.

You fixed exactly what I asked for: `-DargLine` now has somewhere to land, via `<argLine>${argLine}</argLine>` in the root Surefire and Failsafe configuration. That closes my previous blocker. Unfortunately the specific way it is wired introduces a bigger problem than the one it solves, and I want to walk through it carefully because the fix is small once you can see it.

Since you are new to the project I am going to keep this focused on the two things that must change before merge, plus one note. I am not going to pile on style suggestions.

**User pain point.** The `unit-test` CI job hits memory pressure on GitHub-hosted runners, and `all-connectors-it-3` sometimes exceeds its 210-minute budget. Both make CI flaky in a way that has nothing to do with the code being tested.

**Fix approach.** Give the forked test JVMs an explicit 3 GB heap and 512 MB metaspace via `-DargLine`, reduce Maven's own heap to 1.5 GB to make room, wire `${argLine}` through the root Surefire/Failsafe configuration so the property is actually consumed, and raise the `all-connectors-it-3` timeout to 240 minutes.

**One-sentence summary.** It tries to move the CI memory budget from the Maven process into the test JVMs where the tests actually run.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Precise description of the changes.** Three edits:

- `.github/workflows/backend.yml:476` — add `-DargLine="-Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8"`, collapse the multi-line `run:` to one line.
- `.github/workflows/backend.yml:478` — `MAVEN_OPTS` `-Xmx4096m` to `-Xmx1536m`.
- `.github/workflows/backend.yml:945` — `all-connectors-it-3` timeout 210 to 240.
- `pom.xml:673,705` — add `<argLine>${argLine}</argLine>` to the managed `maven-surefire-plugin` and `maven-failsafe-plugin` configuration.

**Key findings.**

1. `argLine` is not defined as a property anywhere in this repository. I checked: `git grep -n argLine -- '*.xml'` returns exactly four hits — the two lines this PR adds, plus `connector-lance/pom.xml:120` and `connector-lance-e2e/pom.xml:94`, and those two reference a *different*, properly-defaulted property.
2. There is no plugin that would populate it either. JaCoCo — the usual reason `${argLine}` exists in a pom — is **not configured** in this project. The only trace is a vestigial `<jacoco-agent.destfile>` system property at `pom.xml:676`; there is no `jacoco-maven-plugin`, and no workflow invokes one.
3. So `${argLine}` resolves only when a command line passes `-DargLine=...`. **Exactly one of the ~55 `./mvnw` invocations in `backend.yml` does** — the one this PR edits. See Issue 1; this is the blocker.
4. Because the change is in the root pom rather than the workflow, its blast radius is the entire repository and every local developer build, not just the `unit-test` job. The PR title and description still describe this as a CI-only change.
5. `connector-lance` and `connector-lance-e2e` declare Surefire in `<build><plugins>` with their own `<configuration>`, which wins over the root `<pluginManagement>`. Those two modules will keep `${surefire.jvm.args}` and will not receive the new heap. That is fine — just be aware the change does not reach them.

**Complete execution path — where the unresolved property lands:**

```text
Any ./mvnw invocation that does NOT pass -DargLine
  -> Maven builds the effective POM
      -> pluginManagement maven-surefire-plugin <argLine>${argLine}</argLine>   [pom.xml:673]
      -> pluginManagement maven-failsafe-plugin <argLine>${argLine}</argLine>   [pom.xml:705]
      -> property `argLine` is undefined  ->  placeholder stays literal
  -> surefire:test / failsafe:integration-test forks the test JVM
      -> the literal string ${argLine} is placed on the forked JVM command line

Jobs in backend.yml that take this path (surefire active, no -DargLine):
  - sanity-check           -> ./mvnw -B -T 1 clean test -pl seatunnel-ci-tools -am        [backend.yml:61]
Jobs in backend.yml that take this path (failsafe active, no -DargLine):
  - updated-modules-integration-test-part-1..8                                    [504,535,566,597,627,657,687,718]
  - engine-v2-it, edge-agent-it, engine-k8s-it                                    [746,771,804]
  - transform-v2-it-part-1..2                                                     [833,861]
  - all-connectors-it-1..8, jdbc-connectors-it-*, kudu/kafka/doris/paimon/... it   [892,930,961,992,...]

Job that resolves the property correctly:
  - unit-test              -> ./mvnw ... -DargLine="-Xmx3072m ..."               [backend.yml:476]   (1 of ~55)

Local developer builds
  -> ./mvnw test  /  ./mvnw verify   ->  no -DargLine  ->  same unresolved placeholder
```

## 1.2 Compatibility Impact

**Partially incompatible** — not for end users, but for the build. No API, configuration option, default value, protocol, or serialization change; nothing a SeaTunnel user runs is affected. But the root pom change alters the build contract for every module and every contributor, which is a different category from the CI-only change described in the PR body.

The intent is right — Surefire's forked JVM previously inherited no explicit heap setting, so it got the JVM default, and giving it an explicit budget is the correct approach.

The concern is the funding. `MAVEN_OPTS` drops from 4096m to 1536m, and that heap is what runs the *entire* `clean verify` reactor for this job, including compiling every module before any test forks. A 62% cut to the compile-phase heap on a reactor this size is a real risk, and the failure mode would look similar to what you are trying to fix — an OOM in CI — just relocated from the test phase to the compile phase. Since `-T 1` and the default `forkCount` of 1 mean peak usage is roughly `1536m + 3072m`, and the previous peak was roughly `4096m` plus an unbounded default fork, you may not actually need to cut `MAVEN_OPTS` as far as 1536m to stay within the runner. I would try 3072m for `MAVEN_OPTS` first.

**Issue 1: `${argLine}` has no default, so every build that does not pass `-DargLine` gets an unresolved placeholder**
- **Location:** `pom.xml:673`, `pom.xml:705`
- **Problem description:** The property is referenced but never defined, and nothing in this repository sets it (no JaCoCo — see key finding 2). Maven leaves an undefined `${...}` in plugin configuration as a literal string, and Surefire 2.22.2 forwards `<argLine>` verbatim onto the forked JVM command line. This is the same failure that the well-known JaCoCo `${argLine}` pitfall produces, and it is why the standard guidance is to never reference a bare `${argLine}` without guaranteeing a value. As traced above, this affects the `sanity-check` job (the first gate in the workflow), every integration-test job via Failsafe, and every local `./mvnw test` or `./mvnw verify`.
- **Potential risk:** Broad build breakage well outside the job this PR is trying to stabilize — including contributors' local builds, which is the worst place for it because the error message points at a JVM argument and not at this pom change. Note that your own CI run should demonstrate this: `sanity-check` runs before `unit-test`, so if it goes red on `36e92bc6` with a JVM argument error, that is this issue rather than anything unrelated.
- **Best improvement:** Use a named property **with a default**, which is the pattern this repository already uses. `connector-lance/pom.xml:40` defines `<surefire.jvm.args>-Dfile.encoding=UTF-8</surefire.jvm.args>` in `<properties>` and `connector-lance/pom.xml:120` references `<argLine>${surefire.jvm.args}</argLine>`. Mirroring that in the root pom:

  ```xml
  <properties>
      <!-- Default JVM args for forked Surefire/Failsafe test JVMs; CI may override. -->
      <surefire.jvm.args>-Dfile.encoding=UTF-8</surefire.jvm.args>
  </properties>
  ```

  ```xml
  <artifactId>maven-surefire-plugin</artifactId>
  <configuration>
      <argLine>${surefire.jvm.args}</argLine>
  ```

  and in the workflow, `-Dsurefire.jvm.args="-Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8"`.

  Using a distinct property name rather than `argLine` also avoids colliding with the conventional JaCoCo property if coverage is ever added later. Option B — keeping the name `argLine` but adding `<argLine></argLine>` to `<properties>` — would also work, but I prefer Option A because it matches existing code in this repo and reads unambiguously.
- **Severity:** High

**Issue 2: the `-DargLine="..."` quoting is not safe on the `windows-latest` leg of this job**
- **Location:** `.github/workflows/backend.yml:476`, matrix at `.github/workflows/backend.yml:462`
- **Problem description:** The `unit-test` job matrix is `os: [ 'ubuntu-latest', 'windows-latest' ]`. On the Windows runner a `run:` step executes under PowerShell, whose native-command argument passing does not preserve embedded double quotes the way bash does. A value containing spaces — `-Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8` — is liable to be split, so Maven sees `-DargLine=-Xmx3072m` plus two stray arguments. The previous code avoided this entirely by carrying the value in the `MAVEN_OPTS` environment variable, where no shell quoting is involved.
- **Potential risk:** The Windows leg of `unit-test` fails or silently applies only part of the intended settings, while the Ubuntu leg behaves as designed — an asymmetry that is confusing to debug.
- **Best improvement:** Pass the value through `env:` rather than the command line, which sidesteps the shell entirely and is consistent with how the rest of this workflow handles JVM settings:

  ```yaml
  - name: run all modules unit test
    run: ./mvnw -B -T 1 clean verify -DskipUT=false -DskipIT=true -D"license.skipAddThirdParty"=true -D"skip.ui"=true --no-snapshot-updates
    env:
      MAVEN_OPTS: -Xmx3072m -Dfile.encoding=UTF-8
      SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
  ```

  with `<surefire.jvm.args>` defaulted in the pom and overridden from the environment, or simply keep `-Dsurefire.jvm.args=...` on one line without inner spaces if you prefer. Either way, please verify the Windows leg specifically, since that is the one this affects.
- **Severity:** Medium

# 2. Code Quality Assessment

## 2.1 Coding Standards

One small convention note, and I mention it only because it is easy to fix and the project checks it: SeaTunnel commit and PR titles follow `[Type][Module] Description`, for example `[Improve][CI] Increase unit-test JVM memory and all-connectors-it-3 timeout`. The current title `ci-cd: increase time and memory management` does not match. See `CLAUDE.md` in the repository root for the full list of types and modules.

Related: since this PR now changes the root `pom.xml`, please update the PR description to say so explicitly. A reviewer scanning "ci-cd: increase time and memory management" would not expect a repository-wide build-configuration change, and that is exactly the kind of thing that should be visible from the description.

This PR changes no UT or E2E test code — only workflow and build configuration — so the flaky-test and assertion-stability analysis does not apply, and I am recording that explicitly rather than skipping it. **Stability rating: not applicable (no test code changed).** No existing assertion, timeout tolerance, or test-selection rule is weakened; the `all-connectors-it-3` timeout increase from 210 to 240 minutes gives tests *more* room and does not reduce coverage.

## 2.3 Documentation Updates

No user-facing documentation is required for a CI and build-configuration change.

# 3. Architectural Soundness

The timeout bump is a clean, precise fix and I have no concerns about it. The memory work is heading the right way — separating the Maven process budget from the test JVM budget is the correct model — but the property plumbing needs the default value before it is safe.

## 3.2 Maintainability

Once Issue 1 is addressed with a named, defaulted property, this becomes a clear and maintainable knob: one property, one default, overridable per job. That is a genuine improvement over the previous state, where forked test JVMs had no explicit budget at all.

## 3.3 Extensibility

A defaulted `surefire.jvm.args` property gives other jobs a supported way to tune their own test JVMs later, which is a nice side benefit of doing it this way.

No runtime or user-facing compatibility impact. The compatibility concern is build-side only, covered in section 1.2.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | `${argLine}` has no default; unresolved placeholder reaches Surefire/Failsafe in ~54 of ~55 Maven invocations and in all local builds | `pom.xml:673,705` | High |
| 2 | `-DargLine="..."` quoting is unsafe on the `windows-latest` leg of the same job | `.github/workflows/backend.yml:476` | Medium |

# 5. Merge Recommendation

**1. Blockers — must be fixed**

- **Issue 1 (High)** — give the property a default. Following `connector-lance/pom.xml:40,120` is the shortest path and matches existing code in this repo. Without it, this change breaks builds far outside the job it targets.
- **Issue 2 (Medium)** — I am listing this as a blocker because the `unit-test` job it modifies runs on Windows in the same matrix, so it is not a hypothetical: the change would go in half-working. Moving the value into `env:` fixes it and is also simpler.

**2. Recommended fixes — non-blocking**

- Try `MAVEN_OPTS: -Xmx3072m` rather than `1536m` (section 1.3). The 62% cut may be more than you need, and an under-provisioned compile phase would produce a CI failure that looks a lot like the one you are fixing.
- Retitle to `[Improve][CI] ...` per the project convention, and note the `pom.xml` change in the PR description.

**Overall assessment.** Your diagnosis is correct and the direction is right — forked test JVMs really did have no explicit memory budget, and that is worth fixing. You also responded to my previous blocker exactly as asked. The remaining problem is that `${argLine}` is a property this repository never defines, so wiring it in without a default turns a targeted CI change into a repository-wide one. Adding the default is a three-line change and the repo already contains the pattern to copy.

**Is there a better alternative implementation?** Option A (named property `surefire.jvm.args` with a default, overridden from the workflow) is better than Option B (keep the name `argLine` and add an empty default), because it matches `connector-lance` and cannot collide with JaCoCo's conventional property if coverage tooling is added later. Both are far better than the current unresolved reference.

On CI: the `Build` check on `36e92bc6` is `queued` at the time of this review, so I have no verdict yet — but as noted in Issue 1, the `sanity-check` job is a good early signal. This branch is `diverged` from `dev` (`ahead_by=3`, `behind_by=4`); I am recording that as a mergeability fact only and I am **not** asking you to sync, since there is no failing signal yet to attribute to upstream drift.

Thanks again for taking this on — CI stability work is unglamorous and genuinely valuable, and you are close. Push an update whenever you are ready and I will re-review the whole head.

davidzollo (dismissed):
+1 if CI passes.
LGTM

DanielLeens:
Thanks @davidzollo for the +1.

Since the condition was "if CI passes": on this same head (`36e92bc6`), the `Build` check is currently showing `fail`, not pending anymore. That lines up with what I flagged as Issue 1 in my review above — `${argLine}` has no default anywhere in this repository, so the placeholder stays unresolved in roughly 54 of the ~55 Maven invocations in `backend.yml` (everything except the one `unit-test` line this PR itself edits) plus every local `./mvnw test`/`verify`. That's a repository-wide build break riding on what the PR title describes as a scoped CI-only change, and I'd expect it to be visible in the CI run right now.

@srijan-singh, my Issue 1 and Issue 2 (unsafe `-DargLine="..."` quoting on the Windows leg of the same job) from the review above are still the two blockers — the fix is small (a defaulted `surefire.jvm.args`-style property following the existing pattern in `connector-lance/pom.xml:40,120`, plus moving the value into `env:` for the Windows quoting issue). Once that's in and `Build` is green on the new head, I'm glad to do a full re-review.

davidzollo:

```

```

DanielLeens (approved):
Thanks for the quick turnaround, @srijan-singh, and thanks @davidzollo for the "+1 if CI passes" check-in on the previous commit. I re-reviewed the entire current head from scratch rather than just the delta since my last pass.

Short version up front: this new commit (`address review comments`) fixes both blockers from my last review, and I have real CI evidence for it, not just static analysis — the fork's `Build` run on this exact head shows all four `unit-test` matrix legs green, including both `windows-latest` legs, which is precisely the leg my Issue 2 said would fail. I'm lifting my previous CHANGES_REQUESTED.

**User pain point.** The `unit-test` CI job can hit memory pressure on GitHub-hosted runners, and `all-connectors-it-3` sometimes exceeds its 210-minute budget — both make CI flaky for reasons unrelated to the code under test.

**Fix approach.** Give the forked Surefire/Failsafe test JVMs an explicit heap/metaspace budget via a properly-defaulted `surefire.jvm.args` property, shrink Maven's own heap to make room for it, and raise the `all-connectors-it-3` timeout.

**One-sentence summary.** It separates the CI memory budget into "Maven process" vs. "forked test JVM," and wires the split through correctly for both Unix and Windows runners.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Precise description of the changes.** Two files, four spots:

- `.github/workflows/backend.yml:475-486` — the single `run all modules unit test` step is replaced by two mutually exclusive steps gated on `runner.os`: a non-Windows step passing `-Dsurefire.jvm.args="$SUREFIRE_JVM_ARGS"` (bash `$VAR` interpolation) and a Windows step passing `-Dsurefire.jvm.args="$env:SUREFIRE_JVM_ARGS"` (PowerShell `$env:` interpolation). Both set `MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8` and `SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8`.
- `.github/workflows/backend.yml:953` — `all-connectors-it-3` timeout raised from `210` to `240` minutes (job starts at line 942).
- `pom.xml:153` — new property `<surefire.jvm.args>-Dfile.encoding=UTF-8</surefire.jvm.args>` with a real default.
- `pom.xml:674` and `pom.xml:706` — `<argLine>${surefire.jvm.args}</argLine>` added to the `pluginManagement` config for `maven-surefire-plugin` and `maven-failsafe-plugin` respectively.

**Before/after (the part that matters):**

Before (previous head `36e92bc6`, the one I gave CHANGES_REQUESTED on):
```yaml
- name: run all modules unit test
  run: ./mvnw ... "-Dsurefire.jvm.args=$SUREFIRE_JVM_ARGS" ...
```
with no default for `surefire.jvm.args` anywhere in the repo, and a single step shared by both `ubuntu-latest` and `windows-latest`.

After (current head `8c3765f5`):
```yaml
- name: run all modules unit test
  run: ./mvnw -B -T 1 clean verify ... "-Dsurefire.jvm.args=$SUREFIRE_JVM_ARGS" --no-snapshot-updates
  env:
    MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8
    SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
  if: runner.os != 'Windows'
- name: run all modules unit test (Windows)
  run: ./mvnw -B -T 1 clean verify ... "-Dsurefire.jvm.args=$env:SUREFIRE_JVM_ARGS" --no-snapshot-updates
  env:
    MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8
    SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
  if: runner.os == 'Windows'
```
plus, in `pom.xml`:
```xml
<surefire.jvm.args>-Dfile.encoding=UTF-8</surefire.jvm.args>
...
<artifactId>maven-surefire-plugin</artifactId>
<configuration>
    <argLine>${surefire.jvm.args}</argLine>
    ...
```

**Key findings.**

1. **Normal path, not a boundary case.** `<argLine>${surefire.jvm.args}</argLine>` lives in root `pluginManagement`, so it is inherited by every module and fires on every `mvn test`/`mvn verify` invocation across the repo — the `sanity-check` job, all `updated-modules-integration-test-*`, `engine-v2-it`, `transform-v2-it-*`, every `all-connectors-it-*`/`jdbc-connectors-it-*` job, and every local developer build. This is squarely the main path, not an edge case.
2. **Both of my previous blockers are closed.** Issue 1 (undefined `${argLine}` reaching Surefire/Failsafe with no default) is fixed by defining `surefire.jvm.args` with a real default at `pom.xml:153` — this is exactly the "Option A" pattern I pointed to in `connector-lance/pom.xml:40,120`, copied faithfully. Issue 2 (unsafe `-DargLine="..."` quoting on `windows-latest`) is fixed by splitting into OS-gated steps that each use the interpolation syntax native to their shell (bash vs. PowerShell), so no cross-shell quoting ambiguity remains.
3. **This is a precise fix, not a workaround.** It reuses an established repository convention (`connector-lance`) rather than inventing a new mechanism, and the `if: runner.os != 'Windows'` / `if: runner.os == 'Windows'` pair is complementary and exhaustive against the job's matrix (`ubuntu-latest`, `windows-latest`), so exactly one step runs per leg — never both, never neither.
4. **Verified against real CI, not just source reading.** The fork's `Build` run for this exact head (`8c3765f5`) shows all four `unit-test` matrix legs (`java 8`/`11` × `ubuntu-latest`/`windows-latest`) green, plus `Sanity check results` and every non-skipped Failsafe-driven IT job in the run. Those are exactly the paths my two blockers predicted would break, and they're the ones now demonstrated to work. This is a meaningful upgrade in confidence over my last review, where I only had a "queued" check to go on.
5. **One remaining blind spot, not a blocker.** Because the default now lives in root `pluginManagement`, every module that previously had no explicit Surefire/Failsafe `argLine` (i.e., all of them except `connector-lance`, which sets its own module-level override and wins) now gets `-Dfile.encoding=UTF-8` on its forked test JVM for the first time — including local `./mvnw test`/`verify` outside CI. That's a genuine, if low-risk, repository-wide behavior change riding on a PR whose title and description still frame this as CI-only.

**Execution path (where the config resolves now):**
```text
Any ./mvnw test|verify invocation (CI job or local dev)
  -> effective POM: pluginManagement maven-surefire-plugin <argLine>${surefire.jvm.args}</argLine>   [pom.xml:674]
  -> effective POM: pluginManagement maven-failsafe-plugin <argLine>${surefire.jvm.args}</argLine>   [pom.xml:706]
  -> property `surefire.jvm.args` now has a real default: -Dfile.encoding=UTF-8                      [pom.xml:153]
  -> unit-test job additionally overrides it via -Dsurefire.jvm.args=$SUREFIRE_JVM_ARGS (bash)        [backend.yml:476]
                                                   or   -Dsurefire.jvm.args=$env:SUREFIRE_JVM_ARGS (pwsh) [backend.yml:482]
  -> surefire:test / failsafe:integration-test forks the test JVM with a fully-resolved argLine
       - unit-test job (both OSes): 3 GB heap + 512 MB metaspace, UTF-8
       - every other job / local build: just UTF-8 (no heap override, same as before this PR for those jobs)
```

## 1.2 Compatibility Impact

No public API, `Option` name/default, protocol, or serialization format is touched — this PR is CI workflow and Maven test-plugin configuration only, so no SeaTunnel user or runtime behavior is affected. The one build-side nuance worth naming explicitly (see Key finding 5) is that the new root-level `argLine` default now applies uniformly to every module's forked test JVM, including local builds outside CI; it's additive (`-Dfile.encoding=UTF-8` only, no new required property, no removed option), so it doesn't rise to "incompatible," but it is a wider blast radius than "CI-only."

The design now correctly separates two budgets that used to be conflated: Maven's own process heap (`MAVEN_OPTS`) vs. the forked Surefire/Failsafe test JVM heap (`surefire.jvm.args` -> `argLine`). That's the right model, and in my last review I flagged the `MAVEN_OPTS` cut from `4096m` to `1536m` as a real OOM risk for the compile phase, since it happens before any test fork. I still think `3072m` would be a more conservative choice, but I no longer have to guess: the actual `unit-test` run on `8c3765f5` completed successfully on both `java 8` and `java 11`, on both `ubuntu-latest` and `windows-latest`, at `1536m`. That's direct evidence the cut is survivable for the current reactor size, so I'm downgrading this from a risk to a non-blocking watch-item — if the module count grows, `1536m` is the first thing I'd revisit.

No new concurrency, retry, or resource-release concerns; this only affects ephemeral CI/test-JVM processes.

Nothing to flag specifically for error handling here — this is build/CI configuration, not application code, and the `if:` gating is deterministic and mutually exclusive as described above, so there's no silent-skip or double-run risk.

**Issue 1 — PR title still doesn't follow the `[Type][Module] Description` convention**
- Location: PR title (`ci-cd: increase time and memory management`)
- Problem description: The project's convention is `[Type][Module] Description`, e.g. `[Improve][CI] Increase unit-test JVM memory and all-connectors-it-3 timeout`. This was already noted in my previous review and hasn't changed.
- Potential risk: Purely cosmetic/searchability; no functional risk.
- Best improvement: Retitle to something like `[Improve][CI] Increase unit-test JVM memory and all-connectors-it-3 timeout`.
- Severity: Low
- Raised by another reviewer: No (carried over from my own previous review as a non-blocking style note, still open)

**Issue 2 — PR description doesn't mention the root `pom.xml` change**
- Location: PR description
- Problem description: The description still only talks about the workflow command and the timeout bump; it doesn't mention that `pom.xml` (a repository-wide, non-CI-scoped file) is also being changed. This was also flagged previously.
- Potential risk: A reviewer or future git-blame reader scanning the title/description alone would not expect a build-configuration change with repo-wide reach.
- Best improvement: Add one sentence noting the new `surefire.jvm.args` property and that it now applies to every module's forked test JVM.
- Severity: Low
- Raised by another reviewer: No (carried over from my own previous review, still open)

**Issue 3 — Minor duplicated `-Dfile.encoding=UTF-8`**
- Location: `.github/workflows/backend.yml:478-479` and `:484-485`
- Problem description: `MAVEN_OPTS` and `SUREFIRE_JVM_ARGS` both set `-Dfile.encoding=UTF-8` independently. Harmless (one affects the Maven JVM, the other the forked test JVM), but slightly redundant to read.
- Potential risk: None functionally; purely a minor readability nit.
- Best improvement: Optional — a one-line comment noting they're intentionally set on two different JVMs would preempt a future "why is this here twice?" question.
- Severity: Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

The YAML and XML changes are clean, minimal, and follow existing repository idioms (the `surefire.jvm.args` naming and defaulting pattern is copied from `connector-lance/pom.xml:40,120` rather than invented). No new methods/fields requiring Javadoc-style comments apply here since this is build/CI configuration, not Java source.

No UT/E2E test code is touched by this PR (only workflow YAML and Maven plugin configuration), so the mandatory flaky-test analysis doesn't apply — consistent with what I recorded in my previous review. Stability rating: **not applicable (no test code changed)**. The `all-connectors-it-3` timeout increase (210 -> 240 minutes) only gives existing tests more room; it doesn't relax any assertion or skip any test.

## 2.3 Documentation Updates

No `docs/en`/`docs/zh` updates are needed — nothing user-facing changed.

# 3. Architectural Soundness

This is now a precise, long-term-viable fix rather than a workaround: it reuses an established pattern already present in the repo instead of introducing a parallel mechanism, and the OS-specific step split is the correct way to handle the bash/PowerShell quoting difference (better than trying to force one syntax to work on both shells).

## 3.2 Maintainability

The `surefire.jvm.args` property is now a clear, single, overridable knob with a safe default, which is a genuine improvement over the previous state where forked test JVMs had no explicit budget at all and (before this fix) an unresolved placeholder was a live risk.

## 3.3 Extensibility

Other jobs can now override `-Dsurefire.jvm.args=...` the same way if they hit similar memory pressure later, without needing another round of pom.xml plumbing.

No historical runtime/user-facing compatibility concern — this is build/CI infrastructure only.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | PR title doesn't follow `[Type][Module] Description` convention | PR title | Low |
| 2 | PR description doesn't mention the root `pom.xml` change | PR description | Low |
| 3 | Minor duplicated `-Dfile.encoding=UTF-8` between `MAVEN_OPTS` and `SUREFIRE_JVM_ARGS` | `.github/workflows/backend.yml:478-479,484-485` | Low |

# 5. Merge Recommendation

**No blocking issues found.** Both High/Medium issues from my previous CHANGES_REQUESTED review (the undefined `argLine` property and the unsafe Windows quoting) are fixed, and — unlike last time — I have a real, current CI run on this exact head to back that up: all four `unit-test` matrix legs pass, including both `windows-latest` legs, alongside `sanity-check` and the full set of Failsafe-driven IT jobs that also depend on the same `argLine` wiring. `gh pr checks` likewise shows `Build`, `Notify test workflow`, and `labeler` all passing on the apache-side pointer.

1. Blockers — none.
2. Recommended fixes — non-blocking (Issues 1-3 above): retitle per convention, mention the `pom.xml` change in the description, and optionally comment on the duplicated encoding flag. None of these need to hold up merging.

Regarding the earlier "+1 if CI passes. LGTM" review: that condition is now met with real evidence (green `Build` on the current head, not just a queued check), so I agree with it. I'm also formally superseding my own previous CHANGES_REQUESTED — the concerns it raised were valid at the time (and the CI failure on that earlier head confirmed them), but the current commit addresses them directly rather than working around them.

Is there a better alternative implementation? Not that I'd push for — Option A (named, defaulted property matching `connector-lance`) was the right call over Option B (an empty-defaulted `argLine`), for the reasons in my last review, and that's what landed.

Thanks again for sticking with this through two rounds of review — CI stability work is unglamorous but this is a clean, well-verified fix.

SEZ9 (commented):
Thanks for working on this. I reviewed the latest head `8c3765f5489b`.

# What this PR fixes
- User pain: The `unit-test` and `all-connectors-it-3` CI jobs are flaky/failing due to JVM memory pressure and timeouts, blocking unrelated PRs from getting green builds.
- Fix approach: Split memory between the Maven reactor (`MAVEN_OPTS -Xmx1536m`) and forked test JVMs via a new `surefire.jvm.args` property wired into surefire/failsafe `<argLine>`, add a Windows-specific step for env expansion, and bump the `all-connectors-it-3` timeout from 210 to 240 minutes.
- One-line summary: The memory-split idea is sound and validated in the linked run, but the Windows step calls the POSIX `./mvnw` wrapper (broken or dead code), the hard-coded `<argLine>` will clobber the JaCoCo agent argLine and silently kill coverage, and the timeout bump papers over the underlying slowdown — these need resolution before merge.

# Runtime chain I rechecked
```text
unit-test job step 'run all modules unit test'  .github/workflows/backend.yml:475-481
  -> ./mvnw -T 1 clean verify with MAVEN_OPTS=-Xmx1536m (reactor JVM heap)  backend.yml:479
    -> passes -Dsurefire.jvm.args=$SUREFIRE_JVM_ARGS (-Xmx3072m -XX:MaxMetaspaceSize=512m)  backend.yml:476,480
      -> resolves <surefire.jvm.args> property (default -Dfile.encoding=UTF-8)  pom.xml:153
        -> maven-surefire-plugin <argLine>${surefire.jvm.args}</argLine>  pom.xml:674 (forked UT JVM gets 3072m; overrides any jacoco-prepared @{argLine})
        -> maven-failsafe-plugin <argLine>${surefire.jvm.args}</argLine>  pom.xml:706 (same for ITs when skipIT=false)
Windows branch: 'run all modules unit test (Windows)'  backend.yml:481-486
  -> ./mvnw under PowerShell (POSIX script, no mvnw.cmd) — fails at shell invocation before Maven starts; currently unreachable since the job matrix pins ubuntu runners
all-connectors-it-3 job  backend.yml:953 timeout-minutes 210 -> 240 (no change to what the job executes)
```

# Findings
**Issue 1: Windows unit-test step invokes the POSIX `./mvnw` wrapper under PowerShell, so it will fail (or is dead code if the job matrix never includes Windows)**
- Location: `.github/workflows/backend.yml:482`
- Why it matters: - Root cause: on GitHub-hosted Windows runners the default shell for `run:` steps is PowerShell (pwsh). The new step uses `./mvnw`, which is the POSIX shell wrapper script; PowerShell cannot execute it directly and the step will fail with an execution/association error. The Windows wrapper is `mvnw.cmd` and should be invoked as `.\mvnw.cmd` (or the step should force `shell: cmd`/`bash`).
- Impact: if this `run all modules unit test` job is ever run on a Windows runner (the step's whole reason to exist, given the `if: runner.os == 'Windows'` guard), CI will fail immediately, not because of tests but because of the wrapper invocation. If the job matrix currently only contains `ubuntu-latest`, this entire step is unreachable dead code that adds maintenance noise and will silently rot.
- Fix: change the Windows step to `run: .\mvnw.cmd -B -T 1 clean verify ... "-Dsurefire.jvm.args=$env:SUREFIRE_JVM_ARGS" --no-snapshot-updates`, or drop the Windows-specific step entirely if the job never schedules Windows runners.
- Evidence: `/.github/workflows/backend.yml:486  if: runner.os == 'Windows'`; `.github/workflows/backend.yml:483  run: ./mvnw -B -T 1 clean verify -DskipUT=false -DskipIT=true -D"license.skipAddThirdParty"=true -D"skip.ui"=true "-Dsurefire.jvm.args=$env:SUREFIRE_JVM_ARGS" --no-snapshot-updates`; `.github/workflows/backend.yml:487  if: runner.os == 'Windows'`
- Suggested fix: On Windows runners the default `run:` shell is PowerShell, and `./mvnw` is the POSIX wrapper script — PowerShell can't execute it, so this step will fail the moment it actually runs on a Windows runner. Please use `.\mvnw.cmd` here (the `$env:SUREFIRE_JVM_ARGS` syntax is already PowerShell-correct), or remove this step entirely if the unit-test job's matrix never includes a Windows OS.
- Severity: Medium

**Issue 2: Hard-coding `<argLine>${surefire.jvm.args}</argLine>` in surefire/failsafe overrides any dynamically contributed `argLine` (e.g., a JaCoCo runtime agent), which can silently disable coverage collection**
- Location: `pom.xml:674`
- Why it matters: - Root cause: when the surefire/failsafe `<configuration>` explicitly declares `<argLine>`, it takes precedence over the `argLine` Maven property that agent plugins like `jacoco-maven-plugin:prepare-agent` set. The root POM already wires JaCoCo via `<jacoco-agent.destfile>` in `systemPropertyVariables` (visible right below the added line), so any module or profile that relies on the property-based `argLine` injection will now fork test JVMs without the agent and produce empty coverage.
- Impact: coverage reports (jacoco.exec) can silently come back empty for affected modules — tests still pass, so nobody notices until coverage gates or reports break.
- Fix: use surefire's late-replacement syntax so both compose: `<argLine>@{argLine} ${surefire.jvm.args}</argLine>`, and declare an empty `<argLine></argLine>` property in `<properties>` so builds that never run `prepare-agent` don't fail on an unresolved token. Apply the same pattern to the failsafe block at pom.xml:706.
- Evidence: `pom.xml:706  <argLine>${surefire.jvm.args}</argLine>`; `pom.xml:153  <surefire.jvm.args>-Dfile.encoding=UTF-8</surefire.jvm.args>`
- Suggested fix: Declaring `<argLine>` explicitly in the surefire configuration overrides the `argLine` property that agent plugins (notably `jacoco-maven-plugin:prepare-agent`) inject, so any module/profile relying on property-based agent wiring will fork test JVMs without the JaCoCo agent and report empty coverage. Please use late replacement so both compose: `<argLine>@{argLine} ${surefire.jvm.args}</argLine>` (plus an empty `<argLine/>` default in `<properties>`), and mirror the same in the failsafe block below.
- Severity: Medium

**Issue 3: MAVEN_OPTS reduced from 4096m to 1536m for the full reactor `clean verify` — risk of OOM in the Maven process itself**
- Location: `.github/workflows/backend.yml:479`
- Why it matters: - Root cause: before this change the forked surefire JVMs used the JVM default heap (forked JVMs never inherit `MAVEN_OPTS`), and Maven itself had 4096m. Now Maven is capped at 1536m and each forked test JVM gets a fixed 3072m via `SUREFIRE_JVM_ARGS`.
- Impact: 1536m + 3072m stays under the 7 GB standard runner budget, which is the point of the fix, but SeaTunnel's reactor is very large (hundreds of connector modules); a full `clean verify` reactor with dependency resolution, shade, and license plugins can pressure a 1536m Maven heap and turn into intermittent `GC overhead limit exceeded` failures on Maven itself rather than in tests. The linked green run shows it works today, but it leaves little headroom as modules are added.
- Fix: keep the split but give Maven a bit more slack (e.g., `-Xmx2048m`) or add `-XX:+ExitOnOutOfMemoryError` to `MAVEN_OPTS` so a reactor OOM fails fast and visibly instead of hanging until the job timeout.
- Evidence: `.github/workflows/backend.yml:480  SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8`; `/.github/workflows/backend.yml:478  MAVEN_OPTS: -Xmx1536m -Dfile.encoding=UTF-8`; `/.github/workflows/backend.yml:479  SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8`
- Suggested fix: Dropping the Maven process heap from 4096m to 1536m is a big cut for a full-reactor `clean verify` (shading, license checks, etc. run in-process). With one 3 GB surefire fork we still have headroom on the 7 GB runner — I'd bump this to `-Xmx2048m` or `-Xmx2560m` to avoid flaky OOMs in the Maven JVM itself. Also worth confirming surefire `forkCount` is 1, otherwise multiple 3 GB forks would exhaust runner memory.
- Severity: Medium

**Issue 4: Property named `surefire.jvm.args` is also applied to failsafe — misleading name and no explanatory comment**
- Location: `pom.xml:153`
- Why it matters: - Root cause: the new `${surefire.jvm.args}` property is wired into both `maven-surefire-plugin` (line 674) and `maven-failsafe-plugin` (line 706), so the name no longer matches its scope. It is also declared under the `<!-- Option args -->` block without a comment describing that CI overrides it via `-Dsurefire.jvm.args=...`.
- Impact: a contributor tuning IT (failsafe) memory may not realize this property also changes it, or vice versa; connector IT jobs in `backend.yml` were not updated to pass this property, so failsafe forks now run with only the default `-Dfile.encoding=UTF-8` — which is a behavioral coupling worth documenting.
- Fix: rename the property to something engine-neutral like `test.jvm.args` (or split into `surefire.jvm.args` / `failsafe.jvm.args`), and add a one-line XML comment noting that CI overrides it and that it feeds both surefire and failsafe forked JVMs.
- Evidence: `pom.xml:706  <argLine>${surefire.jvm.args}</argLine>`
- Suggested fix: Since this property is applied to both surefire (line 674) and failsafe (line 706), the name `surefire.jvm.args` is a bit misleading — `test.jvm.args` would be clearer, or split it per plugin. Please also add a short comment noting that CI overrides it via `-Dsurefire.jvm.args=...` so future contributors understand why it exists.
- Severity: Low

**Issue 5: `all-connectors-it-3` timeout raised from 210 to 240 minutes without addressing why the job grew — this masks slowness and eats runner capacity**
- Location: `.github/workflows/backend.yml:953`
- Why it matters: - Root cause: the timeout bump is a symptom-level fix; the diff contains no change that makes `all-connectors-it-3` faster, so the job is simply allowed to consume up to 4 hours per matrix cell (Java 8 and 11), doubling worst-case runner occupancy for hung containers.
- Impact: genuinely stuck integration tests (e.g., a container that never becomes healthy) now block a runner 30 minutes longer before GitHub kills the job, slowing feedback for all queued PRs.
- Fix: acceptable as a stop-gap, but please file/link a follow-up to split `all-connectors-it-3` or rebalance the connector list across the IT shards (the pattern already used by the it-1/it-2/it-3 split) instead of continuing to grow the timeout.
- Suggested fix: Raising the timeout to 240 minutes is fine as a stop-gap, but nothing in this PR makes `all-connectors-it-3` faster, so a hung container now holds a runner for up to 4 hours per matrix cell. Could we track a follow-up to rebalance the connector list across the it-1/it-2/it-3 shards instead of growing this further?
- Severity: Low

**Issue 6: `MAVEN_OPTS`/`SUREFIRE_JVM_ARGS` values and the Maven command line are duplicated verbatim across the Linux and Windows steps**
- Location: `.github/workflows/backend.yml:485`
- Why it matters: - Root cause: the OS split duplicates the identical `env:` block (and an almost identical `run:` line) in two steps guarded by `if: runner.os` conditions.
- Impact: the memory tuning values will inevitably drift — someone bumping `-Xmx3072m` in one step and not the other would create OS-dependent CI behavior that is painful to diagnose.
- Fix: hoist `MAVEN_OPTS` and `SUREFIRE_JVM_ARGS` to job-level `env:` so both steps share a single definition, keeping only the shell-syntax difference (`$SUREFIRE_JVM_ARGS` vs `$env:SUREFIRE_JVM_ARGS`) in the `run:` lines.
- Evidence: `/.github/workflows/backend.yml:485  SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8`; `.github/workflows/backend.yml:480  SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8`; `.github/workflows/backend.yml:486  SUREFIRE_JVM_ARGS: -Xmx3072m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8`
- Suggested fix: This `env:` block is byte-identical to the one on the Linux step above. Please hoist `MAVEN_OPTS` and `SUREFIRE_JVM_ARGS` to job-level `env:` so the memory tuning values can never drift between the two OS branches — only the shell-specific `$SUREFIRE_JVM_ARGS` vs `$env:SUREFIRE_JVM_ARGS` reference needs to differ in the `run:` lines.
- Severity: Low

# Review conclusion

**2. Remaining follow-ups**
- Issue 1: Windows unit-test step invokes the POSIX `./mvnw` wrapper under PowerShell, so it will fail (or is dead code if the job matrix never includes Windows)
- Issue 2: Hard-coding `<argLine>${surefire.jvm.args}</argLine>` in surefire/failsafe overrides any dynamically contributed `argLine` (e.g., a JaCoCo runtime agent), which can silently disable coverage collection
- Issue 3: MAVEN_OPTS reduced from 4096m to 1536m for the full reactor `clean verify` — risk of OOM in the Maven process itself
- Issue 4: Property named `surefire.jvm.args` is also applied to failsafe — misleading name and no explanatory comment
- Issue 5: `all-connectors-it-3` timeout raised from 210 to 240 minutes without addressing why the job grew — this masks slowness and eats runner capacity
- Issue 6: `MAVEN_OPTS`/`SUREFIRE_JVM_ARGS` values and the Maven command line are duplicated verbatim across the Linux and Windows steps

DanielLeens:
Thanks for the thorough second pass, @SEZ9 — I appreciate you re-verifying independently rather than taking my APPROVED at face value. I went back through your two "must fix" findings (Issues 1 and 2) against the actual CI evidence and the current `pom.xml` on this same head (`8c3765f5`), and I don't think either one holds up. Sharing what I found so we're working from the same facts.

**Issue 1 (Windows step invokes `./mvnw`, will fail / is dead code).** This already ran, successfully, on `8c3765f5` — not hypothetically. Job `Run / unit-test (11, windows-latest)` (https://github.com/srijan-singh/seatunnel/actions/runs/31011149808/job/92324201657) shows step 4 "run all modules unit test" **skipped** (correctly gated by `if: runner.os != 'Windows'`) and step 5 "run all modules unit test (Windows)" completed with **conclusion: success**, running ~59 minutes with real Surefire output in the log, e.g.:
```
[WARNING] Tests run: 3, Failures: 0, Errors: 0, Skipped: 2, ... - in org.apache.seatunnel.api.ChineseCharacterCheckTest
```
The `java 8` / `windows-latest` leg (job 92324201693) shows the identical pattern: step 4 skipped, step 5 success. So `./mvnw` does execute correctly under the Windows runner's default shell here — this isn't dead code and it isn't broken; it's demonstrated working on both Windows matrix legs on the exact commit under review.

**Issue 2 (hard-coded `<argLine>` clobbers a JaCoCo-injected argLine, silently killing coverage).** I checked for JaCoCo wiring on this head and don't find any. `grep -n -i jacoco pom.xml` returns exactly one hit — the `<jacoco-agent.destfile>` entry under `systemPropertyVariables` in the surefire config — which is pre-existing context untouched by this diff, and there is no `jacoco-maven-plugin` (no `prepare-agent` execution) declared anywhere in the repository. Since nothing currently populates the `argLine` Maven property dynamically, there's no live agent injection for the new `<argLine>${surefire.jvm.args}</argLine>` to clobber, and no coverage collection running today that this PR could silently break. This matches what I found digging into the same property in my previous round. Happy to be pointed at a specific module/profile if I'm missing one, but I couldn't find it in the current tree.

Issue 3 (`MAVEN_OPTS` at 1536m) is a fair watch-item, and one I raised myself in section 1.3 of my approval — already flagged as non-blocking given the actual green run at that heap size, and I'd still support bumping toward ~2048-3072m as a follow-up rather than a blocker. Issues 4-6 (property naming/scope, timeout-bump-as-stopgap, duplicated env block) are reasonable Low-severity nits I'd be glad to see picked up, but none of them change correctness or the actual CI outcome on this head.

Given that, I'm not reinstating CHANGES_REQUESTED — my APPROVED stands. If Issue 1 or 2 turn out to reproduce somewhere I haven't looked, please point me at the exact job/log or module and I'll take another pass right away.

srijan-singh:
Thanks @DanielLeens for helping out. @SEZ9 the issue arrived when I introduced `connector-couchbase-e2e` module in the middle, which re-arranges the shard and causing timeout and OOM.

Although putting `connector-couchbase-e2e` at the end seems to resolve the issue for now, but it still could be a valuable addition.

DanielLeens:
Thanks for the context, @srijan-singh — that explains the shard/timeout behavior nicely: `connector-couchbase-e2e` landing in the middle of the shard list was reshuffling job weight distribution and pushing some legs over the time/memory budget this PR is tuning, and moving it to the end sidesteps that without touching the underlying CI logic this PR actually fixes.

My APPROVED from earlier still stands — nothing here changes the correctness of the CI time/memory improvements in this PR. If you want to pursue making `connector-couchbase-e2e`'s shard placement smarter later (rather than just appending it), I'd treat that as its own follow-up CI PR so the shard-balancing change can be measured independently rather than folded into this one. Good detective work tracking the root cause down.

SEZ9:
Thanks @DanielLeens for going back to the primary evidence instead of taking my review at face value — that's exactly the kind of re-verification I'd hope for. I've re-checked both points against the links you shared, and I agree on both.

**Issue 1 — withdrawn.** The linked jobs on `8c3765f5` (`unit-test (11, windows-latest)` / job 92324201657 and the Java 8 leg / 92324201693) show step 4 correctly skipped via the `runner.os != 'Windows'` gate and step 5 completing with real Surefire output over ~59 minutes. My review assumed the Windows path was unreachable and that `./mvnw` would fail at shell invocation — the actual run disproves both. Not a blocker, and not dead code.

**Issue 2 — withdrawn as a blocker.** You're right: the only `jacoco` hit in `pom.xml` is the pre-existing `jacoco-agent.destfile` system property, and there's no `jacoco-maven-plugin` / `prepare-agent` execution anywhere in the repo, so nothing dynamically populates `argLine` today and there's nothing for `<argLine>${surefire.jvm.args}</argLine>` to clobber. I'll downgrade this to a non-blocking note: if coverage tooling is added later, this line will need revisiting (e.g., the `@{argLine}`-prepend pattern with a defaulted property), but that's future work, not this PR's problem.

@srijan-singh, that leaves two concrete asks before I re-approve:

1. **Timeout bump (210 → 240 for `all-connectors-it-3`)** — could you link a recent run showing how close the job actually gets to the current limit? I want to confirm we're buying legitimate headroom rather than masking a growing slowdown. If there's a known cause for the increased runtime, a tracking issue reference in the PR description would be enough.
2. **Green `Build` check on the current head** — once CI is fully green on the latest commit, ping me and I'll do the re-review promptly.

Thanks both — the memory-split approach itself remains sound, and the validation done here has been genuinely helpful.

srijan-singh:
Here are the builds which get cancelled due to time out:

- https://github.com/srijan-singh/seatunnel/actions/runs/30747559465
- https://github.com/srijan-singh/seatunnel/actions/runs/30775346505

The one that passed: https://github.com/srijan-singh/seatunnel/actions/runs/30786984078

DanielLeens:
Thanks @srijan-singh for the links, and @SEZ9 for laying out exactly what's left to re-approve. I pulled the timing on all three runs plus the current head to check the timeout claim independently rather than just eyeballing "cancelled"/"success".

**On the timeout evidence (@SEZ9's ask #1).**

- Run 30775346505 (`all-connectors-it-3`, java 8, ubuntu): started `00:52:04Z`, cancelled `04:22:21Z` -> **210m17s**. Java 11 leg: started `00:52:10Z`, cancelled `04:22:25Z` -> **210m15s**. Both cancellations land within seconds of the old `timeout-minutes: 210` ceiling, so this is a genuine timeout kill, not an unrelated cancellation.
- Run 30786984078 (the one that passed): java 8 leg ran `05:54:11Z` -> `09:23:15Z` = **209m4s**; java 11 leg ran `06:00:38Z` -> `09:28:37Z` = **207m59s**. Even the successful run finished with only 1-3 minutes of margin under the 210-minute cap. That's not a one-off fluke, it's a job that was routinely running right up against its own budget.

So the headroom claim checks out: these aren't runs that failed for unrelated reasons, and the "passing" run barely passed.

**One nuance worth flagging.** All three linked runs are from Aug 2-3, on commits (`c72937a7`, `f1977c55`, `eb5ced99`) that predate the `connector-couchbase-e2e` shard placement fix @srijan-singh described on Aug 7. I checked `all-connectors-it-3` on the actual current PR head (`8c3765f5`, fork run 31011149808, same run I cited in my APPROVED): java 8 leg ran `13:50:48Z` -> `16:00:43Z` = **~130m**; java 11 leg ran `13:51:37Z` -> `16:07:30Z` = **~136m**. That's roughly 70-80 minutes of headroom under the *old* 210-minute limit, well before touching the new 240-minute one.

Practically: the shard reorder already fixed the acute problem, so the 240-minute bump isn't masking an ongoing slowdown on today's head — it's a safety margin against the kind of variance the Aug 2-3 runs show (real-world IT duration for this job clearly does fluctuate close to 210m under some conditions), applied on top of a job that's now comfortably faster. That reconciles both your ask and Issue 5 from your Aug 6 review: legitimate headroom, not a cover for an unaddressed regression, and the numeric root cause (shard placement) is understood and documented in-thread rather than mysterious.

**On @SEZ9's ask #2 (green Build on the current head).** This is already satisfied and has been since Aug 5 — no new commit needed. Fork run 31011149808 on head `8c3765f5` is fully green: all four `unit-test` matrix legs (java 8/11 x ubuntu/windows) and both `all-connectors-it-3` legs succeeded. The apache-side `Build` check-run pointer confirms the same (`conclusion: success`, completed `2026-08-05T20:15:48Z`), and `head` on the PR is still `8c3765f5` right now, so nothing has drifted since either of us last looked.

Given both conditions are met with verifiable evidence rather than just re-stated claims, I don't see anything blocking here from my side. @SEZ9, over to you for the re-review you offered — happy to help track down anything else you want checked in the meantime.

davidzollo (approved):
+1
LGTM

PR #11643, opened 2026-08-03
[Test][E2E] Retry Hive plugin jar downloads in HiveIT

davidzollo:
## Purpose

`HiveIT` downloads seven plugin jars into the Hive container with a bare `wget` and asserts the exit code is 0. A single transient network failure against Maven Central therefore fails the entire `all-connectors-it-4` job, on PRs that have nothing to do with Hive.

Observed on #11203 — a PR that only touches a Zeta REST API test:

```
Failed to resolve parameter [TestContainer arg0] in method
  HiveIT.testHiveSourceWholeDatabaseUseRegex
--2026-08-03 18:11:15--  https://repo1.maven.org/maven2/org/apache/thrift/libfb303/0.9.3/libfb303-0.9.3.jar
Connecting to repo1.maven.org|104.18.18.12|:443... connected.
GnuTLS: Error in the pull function.
Unable to establish SSL connection.
 ==> expected: <0> but was: <4>
```

Exit code 4 is wget's network-failure code. The download simply needs to be retried.

## Changes

Replace the seven bare `wget` invocations with a single shared constant that retries:

```
wget --tries=5 --waitretry=5 --retry-connrefused --timeout=30
```

A sustained outage still fails the job; a momentary hiccup no longer does. No assertion is relaxed — every download is still required to succeed.

No. E2E test infrastructure only.

## How was this patch tested?

The `all-connectors-it-4` job on this PR exercises the changed download path.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hive-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hive/HiveIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hive-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hive/HiveIT.java
@@ -52,6 +52,17 @@ import static org.awaitility.Awaitility.given;
         type = {EngineType.SPARK, EngineType.FLINK})
 @Slf4j
 public class HiveIT extends TestSuiteBase implements TestResource {
+
+    /**
+     * Shell fragment used for every plugin jar download inside the Hive container.
+     *
+     * <p>A bare {@code wget} makes the whole E2E job fail on a single transient TLS or DNS hiccup
+     * against Maven Central, which has nothing to do with the behaviour under test. Retry a few
+     * times with a backoff and a per-attempt timeout so only a sustained outage fails the job.
+     */
+    private static final String WGET =
+            " && wget --tries=5 --waitretry=5 --retry-connrefused --timeout=30 ";
+
     private static final String CREATE_SQL =
             "CREATE TABLE test_hive_sink_on_hdfs"
                     + "("
@@ -186,7 +197,7 @@ public class HiveIT extends TestSuiteBase implements TestResource {
                                         + pluginHiveDir
                                         + " && cd "
                                         + pluginHiveDir
-                                        + " && wget "
+                                        + WGET
                                         + hiveExeUrl());
                 Assertions.assertEquals(
                         0,
@@ -194,7 +205,7 @@ public class HiveIT extends TestSuiteBase implements TestResource {
                         downloadHiveExeCommands.getStderr());
                 Container.ExecResult downloadLibFb303Commands =
                         container.execInContainer(
-                                "sh", "-c", "cd " + pluginHiveDir + " && wget " + libFb303Url());
+                                "sh", "-c", "cd " + pluginHiveDir + WGET + libFb303Url());
                 Assertions.assertEquals(
                         0,
                         downloadLibFb303Commands.getExitCode(),
@@ -202,7 +213,7 @@ public class HiveIT extends TestSuiteBase implements TestResource {
                 // The jar of s3
                 Container.ExecResult downloadS3Commands =
                         container.execInContainer(
-                                "sh", "-c", "cd " + pluginHiveDir + " && wget " + hadoopAwsUrl());
+                                "sh", "-c", "cd " + pluginHiveDir + WGET + hadoopAwsUrl());
                 Assertions.assertEquals(
                         0, downloadS3Commands.getExitCode(), downloadS3Commands.getStderr());
                 // The jar of oss
@@ -212,18 +223,18 @@ public class HiveIT extends TestSuiteBase implements TestResource {
                                 "-c",
                                 "cd "
                                         + pluginHiveDir
-                                        + " && wget "
+                                        + WGET
                                         + aliyunSdkOssUrl()
-                                        + " && wget "
+                                        + WGET
                                         + jdomUrl()
-                                        + " && wget "
+                                        + WGET
                                         + hadoopAliyunUrl());
                 Assertions.assertEquals(
                         0, downloadOssCommands.getExitCode(), downloadOssCommands.getStderr());
                 // The jar of cos
                 Container.ExecResult downloadCosCommands =
                         container.execInContainer(
-                                "sh", "-c", "cd " + pluginHiveDir + " && wget " + hadoopCosUrl());
+                                "sh", "-c", "cd " + pluginHiveDir + WGET + hadoopCosUrl());
                 Assertions.assertEquals(
                         0, downloadCosCommands.getExitCode(), downloadCosCommands.getStderr());
             };


DanielLeens (changes requested):

- User pain point
  `HiveIT` downloads seven plugin jars from Maven Central with a bare `wget` inside the test container and asserts exit code 0, so a single transient network failure fails the whole `all-connectors-it-4` job — including on PRs that never touch Hive (the description quotes such a failure observed on #11203).
- Fix approach
  Replace the seven bare `wget` invocations with one shared constant that adds `--tries=5 --waitretry=5 --retry-connrefused --timeout=30`, without relaxing any exit-code assertion.
- One-sentence summary
  The goal is to make the HiveIT plugin-jar download step survive momentary network hiccups so only a sustained outage fails the job.

Thanks for taking this on — this exact class of red CI on unrelated PRs is a real cost for everyone, and hardening the download step is the right instinct. I went through the full diff, traced how the E2E lifecycle reaches the changed commands, and also verified the actual retry semantics of the chosen `wget` flags against GNU wget's dispatch logic, because that is the load-bearing part of this PR. The result of that check is the one blocker below.

# 1. Code Change Review

## 1.1 Core Logic Analysis

### Exact change description

One file changed: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hive-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hive/HiveIT.java` (+18/-7).

- A new documented constant `WGET` at `HiveIT.java:56-64`.
- All seven ` && wget ` fragments in the `@TestContainerExtension` factory (`HiveIT.java:188-240`) replaced by that constant (`HiveIT.java:200`, `208`, `216`, `226-230`, `237`).

### Before / after snippets

Before:

```java
+ " && wget "
+ hiveExeUrl());
```

After:

```java
private static final String WGET =
        " && wget --tries=5 --waitretry=5 --retry-connrefused --timeout=30 ";
...
+ WGET
+ hiveExeUrl());
```

### Key findings

- All seven substitutions are mechanically correct: the constant carries the leading `" && "`, and every call site follows either a `cd <dir>` (`HiveIT.java:200`, `208`, `216`, `226`, `237`) or a preceding URL in the chained OSS command (`HiveIT.java:228`, `230`), so every generated `sh -c` string stays well-formed.
- No assertion is relaxed — all five `Assertions.assertEquals(0, ..., getStderr())` checks remain (`HiveIT.java:202-205`, `209-212`, `217-218`, `232-233`, `238-239`), so a sustained failure still fails the job loudly.
- However, the chosen flags do not retry the exact failure class quoted in the PR description. In GNU wget's retry loop (`http_loop` in wget's `src/http.c`), a TLS handshake failure returns `CONSSLERR`, which is handled as a fatal error: wget prints `Unable to establish SSL connection.` — the precise message in the quoted #11203 log — and exits immediately without ever consulting `--tries`. DNS resolution failure (`HOSTERR`) is likewise fatal unless `--retry-on-host-error` (wget >= 1.20.1) is passed, which this PR does not.
- The quoted log itself corroborates this: wget's default is already `--tries=20`, yet the #11203 failure shows a single attempt with no `Retrying.` line before the hard exit. If `--tries` governed this error class, the bare default would already have retried it 20 times. Adding `--tries=5` therefore cannot be the fix for that failure.
- What the flags do genuinely add: `--retry-connrefused` makes `ECONNREFUSED`/`ENETUNREACH`/`EHOSTUNREACH` retryable (they are fatal by default), and `--timeout=30` bounds DNS/connect/read waits that otherwise default to a 900-second read timeout. Those are real, worthwhile hardening wins — they are just orthogonal to the quoted GnuTLS failure.

### Complete runtime path

```text
Build workflow -> all-connectors-it-4
  -> HiveIT starts (Zeta only; Spark/Flink disabled via @DisabledOnContainer, HiveIT.java:50-52)
  -> @TestContainerExtension extendedFactory runs in the SeaTunnel engine container
     (HiveIT.java:188-240; target dir /tmp/seatunnel/plugins/Hive/lib, HiveIT.java:186)
      -> execInContainer("sh", "-c", "mkdir -p ... && cd ... && wget --tries=5 ... <jar-url>")
      -> Assertions.assertEquals(0, exitCode, stderr)
  -> any nonzero exit -> extension setup fails
  -> every @TestTemplate in HiveIT errors ("Failed to resolve parameter [TestContainer arg0]")

Inside one wget invocation (GNU wget http_loop dispatch):
  retried, governed by --tries/--waitretry:
      connect timeout, read error / reset mid-transfer, gateway timeout,
      connection refused (only because of --retry-connrefused)
  fatal, exits immediately, --tries never consulted:
      TLS handshake failure  -> CONSSLERR ("Unable to establish SSL connection.")  <- the quoted #11203 failure
      DNS resolution failure -> HOSTERR   (without --retry-on-host-error)
      certificate verification failure, HTTP error statuses (without --retry-on-http-error)
```

So on a recurrence of the motivating GnuTLS handshake error, the job fails exactly as before this PR.

## 1.2 Compatibility Impact

**Fully compatible.** This is E2E test infrastructure only — no connector, engine, config, or serialized-state surface changes, no user-facing behavior change, and no assertion semantics change. The green `Build` run on this head also confirms the container's GNU wget accepts all four flags (they are all long-standing options, so there is no busybox/old-wget concern on the current image).

- Happy path: unchanged — one successful attempt per jar, identical setup time.
- Failure path: worst case per URL is now bounded at roughly 5 attempts x (30s DNS + 30s connect + 30s idle-read) plus 1-5s linear backoff, i.e. minutes instead of the previous pathological ceiling (default 900s read timeout x default 20 tries for a stalled transfer). That is a genuine improvement for hung connections.
- Note that `--tries=5` actually lowers wget's default of 20 for the error classes that are retryable. That is fine — 5 bounded attempts beat 20 unbounded ones — but it underlines that the effective delta of this PR is `--retry-connrefused` + bounded timeouts + defined backoff, not "adding retries" in general.
- No new resources, threads, ports, or shared state; nothing to release.

The exit-code-plus-stderr assertion pattern is preserved everywhere, so failures still surface with wget's stderr in the JUnit message. The formal issues:

**Issue 1**
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hive-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hive/HiveIT.java:63-64`
- Problem description: The retry flags do not cover the failure class this PR was opened to fix. GNU wget treats a TLS handshake failure (`CONSSLERR` — the branch that prints the exact `Unable to establish SSL connection.` message quoted in the PR description) and DNS resolution failure (`HOSTERR`, absent `--retry-on-host-error`) as fatal errors that exit immediately without consuming `--tries`. The quoted log confirms it: at wget's default `--tries=20`, zero retry attempts were made.
- Potential risk: The next GnuTLS/DNS hiccup against `repo1.maven.org` fails `all-connectors-it-4` exactly as before, but now with a constant in the code that claims retries are in place — which will cost someone a confusing re-investigation.
- Best improvement: Retry at the shell level, so any nonzero wget exit — fatal or not — is retried. Sketch:

  ```java
  private static String download(String url) {
      String jar = url.substring(url.lastIndexOf('/') + 1);
      return " && for i in 1 2 3 4 5; do"
              + " wget --tries=2 --waitretry=5 --retry-connrefused --timeout=30 -O " + jar
              + " " + url
              + " && break; [ \"$i\" = 5 ] && exit 1; sleep 5;"
              + " done";
  }
  ```

  Two details worth keeping: the explicit `-O <jarname>` prevents wget's `file.jar.1` duplicate naming if a re-attempt happens after a partial file was written, and the `[ "$i" = 5 ] && exit 1` keeps the exit-code contract with the existing `Assertions.assertEquals(0, ...)` intact. The inner flags and the outer loop compose: fast in-process retries for transient transfer errors, loop-level retries for the fatal classes.
- Severity: Medium

**Issue 2**
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hive-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hive/HiveKerberosIT.java:123-156` and `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hive-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hive/HiveOverwriteIT.java:110-147`
- Problem description: The two sibling tests in the same module duplicate the identical seven-jar bare-`wget` download block and are left unhardened. They run in the same CI lane, so the same transient failure can still fail `all-connectors-it-4` through them. (Repo-wide, roughly 30 other E2E classes use the same bare `&& wget` pattern, e.g. `connector-cdc-mysql-e2e/.../MysqlCDCCheckpointRestoreIT.java:104`.)
- Potential risk: The flake this PR targets keeps firing in the same job via a neighboring class, and the fix looks ineffective even where it worked.
- Best improvement: Hoist the retrying download-command builder into a shared helper (e.g. `seatunnel-e2e-common`, or at minimum a package-visible utility in `connector-hive-e2e`) and use it from all three Hive ITs; the rest of the E2E tree can migrate incrementally in follow-ups.
- Severity: Medium

**Issue 3**
- Location: `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hive-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hive/HiveIT.java:56-64`
- Problem description: Three accuracy/readability points on the new constant. (a) The Javadoc promises retries for "a single transient TLS or DNS hiccup" — precisely the two classes these flags do not retry (see Issue 1). (b) It says the downloads run "inside the Hive container"; they actually run inside the SeaTunnel engine container handed to the `ContainerExtendedFactory` (the jars land in the engine's plugin dir `/tmp/seatunnel/plugins/Hive/lib`, `HiveIT.java:186`), not in the HMS/HiveServer2 containers started at `HiveIT.java:245-270`. (c) A constant named `WGET` that begins with `" && "` is a misuse trap — concatenated at the start of a command it produces an invalid shell fragment; a name like `AND_WGET_WITH_RETRY`, or better the `download(url)` helper from Issue 1, makes the contract explicit.
- Potential risk: Future readers trust the comment over the flags and re-diagnose the same flake from scratch; a future call site placing `WGET` first gets a confusing `sh` syntax error.
- Best improvement: Fix the two factual statements and encapsulate the fragment in a helper method whose name and Javadoc describe what it actually guarantees.
- Severity: Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

The change follows the file's existing string-building style, keeps the diff minimal, and — nicely — the new field does carry an explanatory Javadoc (content accuracy aside, see Issue 3). Naming of the constant is the only standards-level nit, folded into Issue 3.

This PR changes E2E infrastructure only; there is no new logic requiring its own test, and the green `all-connectors-it-4` run on this head exercises the changed download path end to end.

Test stability conclusion: **Risk present** (medium, and in this case it is the same finding as Issue 1, so it is covered by the blocker rather than adding a second issue)
- The setup still depends on live downloads from `repo1.maven.org` at test runtime (`HiveIT.java:155-181`, `188-240`) — an inherent E2E flakiness source that this PR mitigates but, per Issue 1, does not mitigate for the fatal wget error classes actually observed in CI.
- Sibling classes in the same lane keep the fully unmitigated pattern (`HiveKerberosIT.java:123`, `HiveOverwriteIT.java:110`).
- On the positive side: no `Thread.sleep` waits, no new hardcoded ports, no shared static state, no order sensitivity, and no nondeterministic expectations are introduced by this diff. (The fixed host-port bindings at `HiveIT.java:250` and `HiveIT.java:268` are a pre-existing parallel-execution hazard, unchanged and not attributable to this PR.)

## 2.3 Documentation Updates

No `docs/en` / `docs/zh` updates are required — this is test-only infrastructure with no user-visible behavior change. The only documentation to correct is the in-code Javadoc (Issue 3).

# 3. Architectural Soundness

**Temporary workaround** in its current form. The direction (retry + bounded timeouts, assertions untouched) is exactly right, but wget-flag-level retry is the wrong layer for the observed failure: wget's own retry loop classifies TLS/DNS failures as fatal, so only a shell-level loop (or pre-provisioned jars) actually covers the motivating case. With the loop from Issue 1 this becomes a precise fix.

## 3.2 Maintainability

Consolidating seven inline fragments into one documented constant is a clear maintainability improvement over the status quo. The residual concerns are the misleading Javadoc and the `" && "`-prefixed constant shape (Issue 3), both cheap to fix.

## 3.3 Extensibility

The concept generalizes well — which is exactly why it belongs in a shared helper rather than a private constant: the identical download block exists in `HiveKerberosIT`, `HiveOverwriteIT`, and roughly 30 more E2E classes (Issue 2). A `download(url)` builder in `seatunnel-e2e-common` would let the whole tree adopt it incrementally.

No impact on released behavior, upgrade paths, checkpoint/state formats, or config semantics — test infrastructure only.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | Retry flags do not retry the motivating failure class — TLS handshake (`CONSSLERR`) and DNS (`HOSTERR`) errors are fatal in wget and bypass `--tries`; a shell-level retry loop is needed | `connector-hive-e2e/.../HiveIT.java:63-64` | Medium |
| 2 | Sibling Hive ITs in the same CI lane keep the identical bare-`wget` block, so the targeted flake can still fail `all-connectors-it-4` | `connector-hive-e2e/.../HiveKerberosIT.java:123-156`; `connector-hive-e2e/.../HiveOverwriteIT.java:110-147` | Medium |
| 3 | Javadoc inaccuracies (TLS/DNS claim, "Hive container") and `" && "`-prefixed `WGET` constant is a misuse trap | `connector-hive-e2e/.../HiveIT.java:56-64` | Low |

# 5. Merge Recommendation

1. Blockers — must be fixed
- Issue 1 (Medium): switch the retry to a shell-level loop so the quoted GnuTLS/DNS failure classes are actually retried. I am treating this as the one blocker despite the Medium severity because it defeats the PR's stated purpose for the exact failure it quotes, the amendment is a few lines in the same file, and merging as-is would leave a constant in the tree that documents protection it does not provide. The change carries zero regression risk otherwise.

- Issue 2: apply the same (loop-based) retry to `HiveKerberosIT` and `HiveOverwriteIT`, ideally via a shared helper; the wider E2E tree can follow up separately.
- Issue 3: correct the Javadoc (TLS/DNS claim, engine-vs-Hive container) and give the fragment a name or helper signature that exposes the leading `" && "` contract.

CI state: the current head (`4904d37`) is fully green — `Build` passed in 1h30m alongside `labeler` and `Notify test workflow` — and the branch is not behind `dev`, so there is nothing CI-side to diagnose and no sync needed. There is no prior review activity on this PR to reconcile; the repository still shows a required-review gate.

Overall assessment: the diagnosis of the pain is correct, the no-relaxed-assertions discipline is exactly right, and the timeout/connrefused hardening is a real win worth keeping. The one gap is that wget's internal retry machinery was trusted to cover error classes it explicitly treats as fatal — once the retry moves one layer up into the shell, this becomes the precise fix it set out to be.

Better alternative implementations:
- Option A (minimal, recommended for this PR): keep the current flags and wrap each download in the POSIX `for`-loop builder sketched under Issue 1, preserving the exit-code assertions.
- Option B (longer-term): hoist a shared retrying-download helper into `seatunnel-e2e-common` and migrate the Hive ITs (then other connectors' E2E classes) onto it; ultimately, pre-provisioning these jars via the build/image instead of downloading at test runtime would remove the network dependency altogether.

From Daniel's side, the remaining gate is the Issue 1 amendment — the CI facts on this head are already clean. Because I only have read-level review controls on this repository, a write-capable maintainer will need to perform the final approval and merge once the loop change lands.

davidzollo:
Closing as superseded: dev removed the in-container wget downloads from HiveIT entirely in favor of staging the jars via `HiveDependencies.copyTo` (see the HiveDependencies class on dev), which eliminates the transient-download failure mode this PR was retrying around. After merging current dev the PR diff becomes empty, so there is nothing left to deliver here.

PR #11382, opened 2026-07-08
[Improve][Zeta] Enforce schema change drain before sink DDL

davidzollo:
### Purpose

Fixes #11381.

This PR enforces the existing schema-change checkpoint protocol on the sink side so sink writers cannot apply a `SchemaChangeEvent` before the schema-change-before checkpoint has completed. This prevents downstream DDL from racing ahead of old-schema rows that still need to cross the sink checkpoint / flush boundary.

### Changes

- Add `SchemaChangeDrainGuard` in the sink flow to reject schema changes until a completed schema-change-before checkpoint opens the DDL window.
- Close the DDL window after the schema-change-after checkpoint completes or when a schema-change checkpoint is aborted.
- Pass `CheckpointType` through checkpoint completion notifications so the guard also works correctly after failover, when the restored sink may not have observed the original barrier in the new runtime.
- Add unit coverage for normal ordering, unsafe bypass, checkpoint-abort cleanup, and restore-from-schema-change-checkpoint behavior.

### Tests

- `./mvnw spotless:apply -pl seatunnel-engine/seatunnel-engine-core,seatunnel-engine/seatunnel-engine-server -nsu -Dmaven.gitcommitid.skip=true`
- `./mvnw test -pl seatunnel-engine/seatunnel-engine-server -Dtest=SchemaChangeDrainGuardTest -nsu -Dmaven.gitcommitid.skip=true`
- `./mvnw verify -pl seatunnel-engine/seatunnel-engine-core,seatunnel-engine/seatunnel-engine-server -DskipTests -DskipIT -nsu -Dmaven.gitcommitid.skip=true`

PR #10874, opened 2026-05-11
[Improve][Core] Support close-table events for bounded multi-table jobs

davidzollo:
## What changed

- added a generic `CloseTableEvent` and `SupportCloseTableSinkWriter` hook in `seatunnel-api`
- added engine-side source-reader reverse event delivery so enumerators can notify readers about table-level completion
- propagated close-table events through source -> transform -> sink lifecycles
- taught `MultiTableSinkWriter` to close a table only after all expected upstream close-table signals arrive
- added a bounded JDBC source implementation that reports split completion, aggregates per-table completion in the enumerator, and emits close-table progress safely to downstream
- preserved close-table event metadata in `TableRenameTransform`

## Why

Bounded multi-table jobs currently keep per-table sink resources alive until the whole task closes because there is no table-complete control event in the open-source pipeline.

The tricky part is parallelism: a single "table finished" event is not safe when multiple upstream readers can still have in-flight rows for the same table. This PR fixes that by carrying upstream-reader progress in the close-table event and only closing a table after all participating readers have reported completion.

Closes #10873

## Validation

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-api -Dtest=MultiTableSinkWriterTest test`
- `./mvnw -pl seatunnel-connectors-v2/connector-jdbc -Dtest=JdbcSourceReaderTest,JdbcSourceSplitEnumeratorTest test`
- `./mvnw -pl seatunnel-transforms-v2 -Dtest=TableRenameTransformTest test`
- `./mvnw -pl seatunnel-engine/seatunnel-engine-server -DskipTests compile`

## Notes

- `./mvnw -pl seatunnel-engine/seatunnel-engine-server -am -DskipTests compile` still fails in the pre-existing `seatunnel-config-shade` module because it cannot resolve shaded `typesafe-config` classes. That failure is outside this PR's change set.

PR #11645, opened 2026-08-04
[Test][E2E] Give the E2E Flink TaskManager enough metaspace and heartbeat slack

DanielLeens:

Fix connector E2E failures that are caused by the shared Flink test cluster dying, not by the PR under test.

`HttpIT#testSourceToAssertSink` failed with `expected: <0> but was: <1>` (the SeaTunnel job exited non-zero) on two unrelated PRs — `#11382` and `#10874`, neither of which touches the HTTP connector. Both failures happened on the Flink 1.20.1 container, and in both the TaskManager is what actually broke:

**`#10874` — the TaskManager ran out of metaspace:**

```
🐳 [tyrantlucifer/flink:1.20.1-scala_2.12_hadoop27:taskmanager] - STDOUT: java.lang.OutOfMemoryError: Metaspace. ...
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:473)
```

**`#11382` — the TaskManager stopped answering heartbeats (13 occurrences):**

```
Caused by: java.util.concurrent.TimeoutException: Heartbeat of TaskManager with id 172.18.0.5:36859-7443e5 timed out.
Job http_binary_to_assert.conf (...) switched from state FAILING to FAILED.
Caused by: org.apache.flink.runtime.JobException: Recovery is suppressed by NoRestartBackoffTimeStrategy
```

**Why it happens**

One TaskManager serves every job of a whole test class — `HttpIT` alone ran 28 jobs in that run — and each SeaTunnel job loads its connector jars through a fresh user-code `ClassLoader`. The E2E clusters never configure metaspace, so they inherit the implied 256m, which is exhausted part-way through a long class. Flink's own message names this case: *"either the job requires a larger size of JVM metaspace or there is a class loading leak"*. Here it is the former.

Heartbeats are the second half of the same story. CI runners host several containers at once, so a TaskManager can be starved of CPU past the 50s Flink default — and a metaspace-dead TaskManager stops answering entirely. A missed heartbeat is fatal rather than recoverable, because SeaTunnel jobs run with `NoRestartBackoffTimeStrategy` unless the job config asks for restarts.

No. E2E test-infrastructure only; no production code and no test assertions are touched.

- `taskmanager.memory.jvm-metaspace.size: 512m` (was the implied 256m).
- `taskmanager.memory.process.size: 2048m` (was 1728m) — the process budget grows by exactly the metaspace increase, so the derived heap / network / managed pools stay where they are today rather than shrinking to pay for the metaspace. For Flink 1.20, `taskmanager.memory.flink.size` stays pinned at `1280m`, so total Flink memory is unchanged and the derived JVM overhead lands at 256m, inside the default `[192m, 1g]` bounds.
- `heartbeat.timeout: 120000` with `heartbeat.interval: 10000` (was the 50s / 10s default).

`Flink20Container` overrides `getFlinkProperties()` with its own complete list rather than extending `DEFAULT_FLINK_PROPERTIES`, so it needs the same settings — and it is the container both observed failures happened on.

Verified by CI on this PR head, which exercises these containers across the whole Flink matrix (1.13 / 1.15 / 1.18 / 1.20) in `all-connectors-it-*`, `kafka-connector-it`, `transform-v2-it-*` and the other connector IT jobs.

### Check list

### DIFF
--- a/seatunnel-e2e/seatunnel-e2e-common/src/test/java/org/apache/seatunnel/e2e/common/container/flink/AbstractTestFlinkContainer.java
+++ b/seatunnel-e2e/seatunnel-e2e-common/src/test/java/org/apache/seatunnel/e2e/common/container/flink/AbstractTestFlinkContainer.java
@@ -59,6 +59,21 @@ public abstract class AbstractTestFlinkContainer extends AbstractTestContainer {
                     "taskmanager.numberOfTaskSlots: 10",
                     "parallelism.default: 4",
                     "env.java.opts: -Doracle.jdbc.timezoneAsRegion=false",
+                    // One TaskManager serves every job of a test class, and each SeaTunnel job
+                    // loads its connector jars through a fresh user-code ClassLoader. With the
+                    // Flink default of 256mb the metaspace is exhausted part-way through a long
+                    // class ("OutOfMemoryError: Metaspace"), which kills the TaskManager and fails
+                    // whichever job happens to be running. Raise the process budget alongside it so
+                    // the extra metaspace is not taken out of the derived heap/network/managed
+                    // pools.
+                    "taskmanager.memory.process.size: 2048m",
+                    "taskmanager.memory.jvm-metaspace.size: 512m",
+                    // CI runners host several containers at once, so a TaskManager can be starved
+                    // of CPU for longer than the 50s Flink default before it answers a heartbeat.
+                    // A missed heartbeat fails the job outright, because SeaTunnel jobs run with
+                    // NoRestartBackoffTimeStrategy unless the job config asks for restarts.
+                    "heartbeat.timeout: 120000",
+                    "heartbeat.interval: 10000",
                     // limit restart attempts in e2e to avoid infinite retries
                     "restart-strategy: fixed-delay",
                     "restart-strategy.fixed-delay.attempts: 2",
--- a/seatunnel-e2e/seatunnel-e2e-common/src/test/java/org/apache/seatunnel/e2e/common/container/flink/Flink20Container.java
+++ b/seatunnel-e2e/seatunnel-e2e-common/src/test/java/org/apache/seatunnel/e2e/common/container/flink/Flink20Container.java
@@ -88,8 +88,24 @@ public class Flink20Container extends AbstractTestFlinkContainer {
                         "",
                         "# Memory Configuration",
                         "jobmanager.memory.process.size: 1600m",
-                        "taskmanager.memory.process.size: 1728m",
+                        // One TaskManager serves every job of a test class, and each SeaTunnel job
+                        // loads its connector jars through a fresh user-code ClassLoader. The
+                        // implied 256m metaspace was exhausted part-way through long classes
+                        // ("OutOfMemoryError: Metaspace"), killing the TaskManager. The process
+                        // budget grows by the same amount the metaspace does, so the derived Flink
+                        // memory stays at 1280m and the heap/network/managed pools are unchanged.
+                        "taskmanager.memory.process.size: 2048m",
                         "taskmanager.memory.flink.size: 1280m",
+                        "taskmanager.memory.jvm-metaspace.size: 512m",
+                        "",
+                        "# Heartbeat Configuration",
+                        // CI runners host several containers at once, so a TaskManager can be
+                        // starved of CPU for longer than the 50s Flink default before it answers a
+                        // heartbeat. A missed heartbeat fails the job outright, because SeaTunnel
+                        // jobs run with NoRestartBackoffTimeStrategy unless the job config asks for
+                        // restarts.
+                        "heartbeat.timeout: 120000",
+                        "heartbeat.interval: 10000",
                         "",
                         "# Network Buffer Configuration - Fix for insufficient network buffers",
                         "taskmanager.memory.network.fraction: 0.2",


DanielLeens (commented):
This is a structured self-review of my own PR, following my usual practice of leaving the same full analysis on my own submissions that I would leave on anyone else's, so that write-capable maintainers have something concrete to check my claims against. GitHub does not allow approving or requesting changes on one's own PR, so this is posted as a comment, and it should be read as the author's case for the change plus its known limitations — not as an independent verdict. I re-traced the current head on Tuesday, August 4, 2026, against the latest `dev`; the branch is up to date with `dev` and the `Build` workflow on this head is still pending, which I treat as the gate below.

- User pain point: connector E2E jobs fail on PRs that never touched the connector in question, because the shared Flink test cluster's TaskManager dies underneath them. Two concrete instances on the Flink 1.20.1 container: `HttpIT#testSourceToAssertSink` failed on #10874 with `java.lang.OutOfMemoryError: Metaspace` in the TaskManager, and on #11382 with thirteen `Heartbeat of TaskManager ... timed out` occurrences ending in `Recovery is suppressed by NoRestartBackoffTimeStrategy`.
- Fix approach: raise the TaskManager metaspace from the implied 256m to 512m, grow `taskmanager.memory.process.size` by exactly the same amount so the derived heap/network/managed pools do not shrink to pay for it, and widen `heartbeat.timeout` to 120s (interval 10s) — applied both to the shared default property list and to the Flink 1.20 container's self-contained replacement config.
- One-sentence summary: this PR gives the E2E Flink TaskManager enough metaspace and heartbeat slack that test-infrastructure death stops masquerading as connector test failures.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Precise change description:
- Two files, +32/−1 lines, entirely inside static Flink configuration property lists. No production code, no test assertions, no test flow changes.
- `AbstractTestFlinkContainer.java:69-76` adds `taskmanager.memory.process.size: 2048m`, `taskmanager.memory.jvm-metaspace.size: 512m`, `heartbeat.timeout: 120000`, and `heartbeat.interval: 10000` to `DEFAULT_FLINK_PROPERTIES`, which every Flink 1.13-1.18 container inherits.
- `Flink20Container.java:97-99, 107-108` mirrors the same settings into the Flink 1.20 replacement config, changing the existing `taskmanager.memory.process.size` from 1728m to 2048m and keeping `taskmanager.memory.flink.size` pinned at 1280m.

Before / after snippets (Flink 1.20 container):

Before:
```java
"jobmanager.memory.process.size: 1600m",
"taskmanager.memory.process.size: 1728m",
"taskmanager.memory.flink.size: 1280m",
```

After:
```java
"jobmanager.memory.process.size: 1600m",
"taskmanager.memory.process.size: 2048m",
"taskmanager.memory.flink.size: 1280m",
"taskmanager.memory.jvm-metaspace.size: 512m",
...
"heartbeat.timeout: 120000",
"heartbeat.interval: 10000",
```

Key findings:
- Both delivery mechanisms had to be changed, and both were. The 1.13-1.18 containers pass `DEFAULT_FLINK_PROPERTIES` through the `FLINK_PROPERTIES` env var (`AbstractTestFlinkContainer.java:81, 96, 118`), which the official Flink entrypoint appends to `flink-conf.yaml`; Flink's legacy config parser takes the last occurrence of a duplicated key, so the appended `2048m` correctly overrides the stock image's `1728m`. `Flink20Container` overrides `getFlinkProperties()` with a complete replacement config (`Flink20Container.java:75-129`) that never sees the abstract defaults — and it is the container both observed failures happened on, so patching only the shared list would have missed the actual failure site.
- The Flink 1.20 memory arithmetic is internally consistent: with process 2048m, flink 1280m, and metaspace 512m all pinned, the derived JVM overhead is 2048 − 1280 − 512 = 256m, inside Flink's default `[192m, 1g]` sanity bounds, so the config is accepted and the heap/network/managed pools are byte-for-byte unchanged.
- On the append-mode containers (1.13-1.18), `taskmanager.memory.flink.size` is not pinned, so total Flink memory is derived and actually moves from 1280m (overhead floored at the 192m minimum) to roughly 1331m (0.1 fraction of 2048m = 204.8m overhead). The pools grow slightly rather than staying identical — the direction is safe, but the in-code comment reads a little stronger than the arithmetic (see 1.3).
- The `NoRestartBackoffTimeStrategy` rationale is confirmed by the engine code for the container where the failures occurred: SeaTunnel batch E2E jobs disable checkpointing when no positive `checkpoint.interval` is set (`AbstractFlinkRuntimeEnvironment.java:91-95`), set a job-level restart strategy only when the job config asks for one (`EnvironmentUtil.java:41-47`, invoked at `AbstractFlinkRuntimeEnvironment.java:185`), and the Flink 1.20 replacement config contains no cluster-level `restart-strategy` — so a single missed heartbeat kills the job outright there. However, the same comment is also placed in `AbstractTestFlinkContainer.java:73-74`, three lines above `restart-strategy: fixed-delay` (`AbstractTestFlinkContainer.java:78-80`), which gives 1.13-1.18 jobs two cluster-level restart attempts; the comment overstates the failure mode for that file (Issue 1).
- `heartbeat.interval: 10000` equals the Flink default, so it is a no-op that documents intent; the pairing keeps roughly twelve heartbeat opportunities inside the 120s window, which is a sane ratio.

Deep correctness analysis — how the changed lines reach the running TaskManager JVM and the failure they remove:

```text
Test class startup
  -> AbstractTestFlinkContainer.startUp()
     [AbstractTestFlinkContainer.java:79-83]
  -> String.join("\n", getFlinkProperties()) -> withEnv("FLINK_PROPERTIES", ...)
     for BOTH jobmanager and taskmanager containers
     [AbstractTestFlinkContainer.java:81, 96, 118]
  -> official flink docker-entrypoint appends FLINK_PROPERTIES to conf/flink-conf.yaml
  -> legacy parser: last occurrence of a duplicate key wins
     => taskmanager.memory.process.size 2048m replaces the stock 1728m
  -> TaskManager JVM starts with MaxMetaspaceSize=512m and unchanged-or-larger pools

Flink 1.20 path (replacement, not append)
  -> Flink20Container.startUp() -> createFlink20StartupScript()
     [Flink20Container.java:126-158, 216-256]
  -> marker block rewrites conf/flink-conf.yaml and conf/config.yaml wholesale
  -> 2048m − 1280m − 512m => derived JVM overhead 256m, inside [192m, 1g]

Failure path this removes
  -> one TaskManager serves every job of a test class (HttpIT ran 28 jobs)
  -> each SeaTunnel job submission loads connector jars via a fresh user-code
     classloader on the TaskManager
  -> at the implied 256m the metaspace fills part-way through a long class
     => TaskManager dies ("OutOfMemoryError: Metaspace")
  -> JobManager declares the TaskManager dead after heartbeat.timeout (was 50s)
  -> batch e2e jobs: checkpointing disabled
     [AbstractFlinkRuntimeEnvironment.java:91-95]
     + no job-level restart strategy unless configured
     [EnvironmentUtil.java:41-47]
     + no cluster restart-strategy in the 1.20 replacement config
     => NoRestartBackoffTimeStrategy fails the job on the spot
  -> executeJob(...) returns non-zero => an unrelated PR's IT goes red
```

## 1.2 Compatibility Impact

- Fully compatible.
- Test-infrastructure only: no production code, no public API, no serialization format, no user-facing config defaults, and no test assertions are touched.
- Every Flink E2E container (1.13-1.18 via the inherited defaults, 1.20 via the replacement config) picks up the new memory and heartbeat settings; the configuration keys used are valid on all of those Flink versions.

- Host memory footprint: the TaskManager container's JVM cap grows from 1728m to 2048m, i.e. +320m per running Flink cluster on the CI host. The E2E harness runs one JobManager (unchanged at 1600m) plus one TaskManager per test class, alongside connector containers, so the increase is real but modest relative to runner capacity. If CI runners start showing host-level memory pressure after this lands, this is the knob that moved.
- Failure-detection latency: a genuinely dead TaskManager is now declared dead after up to 120s instead of 50s, so runs that do fail take up to 70s longer to surface the failure. That is the deliberate price for eliminating false kills of CPU-starved-but-alive TaskManagers, but it is worth knowing when reading future CI timing.
- Append-mode pool drift: as noted in 1.1, on 1.13-1.18 the derived total Flink memory grows from 1280m to about 1331m because the overhead fraction now clears its floor. Strictly more memory for the same workload — safe, just not the "unchanged" the comment implies.
- No CPU, GC, network, locking, concurrency, retry/idempotency, or resource-release behavior changes beyond the above; the diff is configuration data.

Issue 1: the `NoRestartBackoffTimeStrategy` comment is inaccurate in the file that configures a restart strategy
- Location: `AbstractTestFlinkContainer.java:73-74` (comment) vs `AbstractTestFlinkContainer.java:78-80` (`restart-strategy: fixed-delay`, 2 attempts)
- Problem description: the comment states that a missed heartbeat "fails the job outright, because SeaTunnel jobs run with NoRestartBackoffTimeStrategy unless the job config asks for restarts". That is true for `Flink20Container`, whose replacement config has no cluster-level restart strategy, and it matches the #11382 log evidence. But in this file the very next property lines set a cluster-level `fixed-delay` strategy with 2 attempts, which applies to jobs whose SeaTunnel config does not set `restart.strategy` (`EnvironmentUtil.java:41-47` only overrides when the job config has the key). On 1.13-1.18 a heartbeat timeout is therefore retried twice, not fatal on the first miss.
- Potential risk: none at runtime — this is comment accuracy. But these comments are written precisely so future maintainers can trust the rationale without re-deriving it, and this one teaches the wrong model for five of the six containers that inherit it.
- Best improvement: tighten the comment in `AbstractTestFlinkContainer` to say the outright-failure behavior applies where no restart strategy is configured (the Flink 1.20 replacement config), while here the cluster-level fixed-delay gives two attempts — which still lose to a dead TaskManager, so the heartbeat slack is justified either way.
- Severity: Low

Issue 2: 512m raises the metaspace ceiling but does not verify the absence of classloader retention
- Location: `AbstractTestFlinkContainer.java:70`, `Flink20Container.java:99`
- Problem description: Flink's own OOM message names two possible causes — an honestly-too-small metaspace or a class-loading leak. The PR asserts the former (reasonable, given 28 jobs in a single class each loading connector jars through a fresh classloader), but nothing in this change or in CI measures whether per-job classloaders are actually collected. I did not verify absence of a leak; doubling the budget would also mask a slow one for longer.
- Potential risk: a test class that runs substantially more jobs than `HttpIT`, or a connector that pins its classloader (JDBC drivers and shutdown hooks are the classic offenders), could still exhaust 512m and reproduce the original failure mode, now rarer and harder to spot.
- Best improvement: nothing blocking now. If a metaspace OOM recurs after this lands, the follow-up should capture `-XX:+TraceClassUnloading`-style evidence or TaskManager metaspace metrics for the failing class before raising the number again.
- Severity: Low

Issue 3: the tuning block is now duplicated in two hand-maintained lists
- Location: `AbstractTestFlinkContainer.java:69-76` and `Flink20Container.java:97-99, 107-108`
- Problem description: the same four values (and near-identical rationale comments) live in both `DEFAULT_FLINK_PROPERTIES` and the Flink 1.20 replacement list. The duplication is forced by the pre-existing design — `Flink20Container.getFlinkProperties()` replaces rather than extends the defaults for SnakeYAML compliance — but this PR widens the surface that must be kept in sync by hand.
- Potential risk: a future tune applied to one list and not the other silently reintroduces the flake on half the matrix, exactly the class of drift this PR is fixing.
- Best improvement: extract the shared values as named constants (e.g. `TASKMANAGER_PROCESS_SIZE`, `TASKMANAGER_METASPACE_SIZE`, `HEARTBEAT_TIMEOUT_MS`) referenced from both lists, so the numbers cannot diverge even though the list structures must stay separate.
- Severity: Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

- The change is configuration data with no new methods or fields, so the new-method documentation requirement is satisfied vacuously; beyond that, every added value carries an inline comment explaining what it fixes and why the number is what it is, which is exactly the standard I ask of others for magic numbers.
- The debug `System.out` block in `Flink20Container.getFlinkProperties()` predates this PR and is untouched; it is noise, but cleaning it up here would blur the diff.
- No style violations in the added lines.

- Test coverage: there is no unit-testable surface here — the change is cluster configuration whose only executable validation is the E2E CI matrix itself (1.13/1.15/1.18/1.20 across the connector-IT lanes). That validation is exactly the pending `Build` workflow on this head, which is why I treat CI completion as the merge gate rather than a formality.
- Test stability rating: **Stable.**
  - The change removes two documented flaky-infrastructure patterns with log evidence from real runs: metaspace-death of the shared TaskManager (`AbstractTestFlinkContainer.java:69-70`, `Flink20Container.java:97-99`) and heartbeat false-positives under CI CPU starvation (`AbstractTestFlinkContainer.java:75-76`, `Flink20Container.java:107-108`).
  - It introduces none of the flaky anti-patterns: no `Thread.sleep`, no hardcoded ports or hostnames, no nondeterministic expected values, no shared mutable state, no ordering sensitivity, and the container readiness signals (`LogMessageWaitStrategy` on resource-manager registration) are untouched.
  - The residual risks are the ones I named myself: the 512m ceiling is a budget, not a proof (Issue 2), and dead-TaskManager detection is slower by design (1.3). Neither is an anti-pattern in the changed code.

## 2.3 Documentation Updates

- Not required: no user-visible behavior changes, so `docs/en` and `docs/zh` are correctly untouched. The configuration keys are internal to the E2E harness, and the PR description records the failure evidence and arithmetic for the archaeology.

# 3. Architectural Soundness

- Precise fix — for the problem as diagnosed. The failure evidence (metaspace OOM stack trace, heartbeat timeout with `NoRestartBackoffTimeStrategy`) points at an under-provisioned shared cluster, and the change adjusts exactly the two budgets implicated, with the process-size increase sized to protect the derived pools rather than a blind bump.
- The honest caveat, since I wrote it: this is capacity tuning, not isolation. It makes the shared-TaskManager model survive long test classes; it does not make test classes independent of each other's classloader behavior. I considered that acceptable because the alternative is structurally more expensive (see the options in section 5).

## 3.2 Maintainability

- The inline rationale comments mean the next person hitting a Flink E2E resource failure can see what was already tuned and why, instead of re-deriving it from CI logs.
- The main maintainability debt is Issue 3: two hand-synchronized copies of the same numbers. Constants would close that gap cheaply.

## 3.3 Extensibility

- Future Flink containers that inherit `DEFAULT_FLINK_PROPERTIES` get the hardening for free. Any future container that follows the `Flink20Container` replacement pattern must remember to copy these settings — the same drift risk as Issue 3, worth remembering when a Flink 2.x container is added.

- The keys are valid across the whole matrix: `taskmanager.memory.process.size` and `taskmanager.memory.jvm-metaspace.size` exist since Flink 1.10, and `heartbeat.timeout`/`heartbeat.interval` are long-stable, so 1.13.6 through 1.20.1 all accept them.
- The two delivery mechanisms match their parsers: last-write-wins append works for the legacy `flink-conf.yaml` parser on 1.13-1.18, and the 1.20 path already replaces the file wholesale for SnakeYAML strictness — this PR adds lines inside that block rather than relying on append semantics that 1.20 would mishandle.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | `NoRestartBackoffTimeStrategy` comment overstates the failure mode in the file that configures `fixed-delay` restarts | `AbstractTestFlinkContainer.java:73-74` vs `:78-80` | Low |
| 2 | 512m metaspace is a budget raise without leak verification; a long-enough class or a pinned classloader could still exhaust it | `AbstractTestFlinkContainer.java:70`, `Flink20Container.java:99` | Low |
| 3 | Tuning values duplicated across two hand-maintained lists with no shared constants | `AbstractTestFlinkContainer.java:69-76`, `Flink20Container.java:97-99, 107-108` | Low |

# 5. Merge Recommendation

1. Blockers — must be fixed
- No blocking issues found in the diff itself. The one gate is external: the `Build` workflow on head `9a98cf4` is still pending, and since the CI matrix is the only executable validation these settings have, the merge decision should wait until the connector-IT lanes complete green on this head. The branch is current with `dev`, so no sync is needed — just the pending result.

- Issue 1: correct the restart-strategy comment in `AbstractTestFlinkContainer` so the rationale is accurate for the containers that inherit it.
- Issue 3: extract the shared tuning values into named constants referenced by both lists.
- Issue 2: no action now; noted as the trigger condition for a follow-up if a metaspace OOM ever recurs.

Overall assessment: the failure evidence is concrete (two independent unrelated-PR failures with TaskManager-side root causes in the logs), the fix touches only the implicated budgets, the Flink 1.20 memory arithmetic passes Flink's own sanity bounds, and the change removes flaky patterns without adding any. The known limitations are the ones listed above, all Low. Because this is my own PR I have deliberately not marked it approved: I would ask a write-capable maintainer to check my two load-bearing claims — the last-write-wins append behavior on the 1.13-1.18 images and the 2048/1280/512 overhead derivation on 1.20 — and, if CI comes back green across the matrix, perform the final approval and merge. From Daniel's side this is a comment-only review by necessity as well as by permission level.

- Option A (this PR): tune the shared cluster's metaspace and heartbeat budgets. Minimal, immediately effective on the observed failures, keeps E2E wall-clock time unchanged.
- Option B: restart the TaskManager between test classes (or split long classes like `HttpIT`) so metaspace can never accumulate across a class. Structurally cleaner isolation, but it adds a container restart per class across the whole matrix, which is a real CI-time cost — reasonable as a follow-up only if Issue 2's trigger condition ever fires.

davidzollo (approved):
+1 if CI passes.
LGTM

davidzollo (approved):
+1

PR #11652, opened 2026-08-04
fix: address silent connector KB degradation and self-referencing suggestions

### LINKED ISSUE #11650: [Bug][CLI] Connector knowledge base degrades silently without metadata; not-found suggestions can echo the queried name
### Search before asking

### What happened

Sub-task of #11616 (Stage-1 items 1.1-context and 1.6). Two related defects in the CLI's connector knowledge base, both observed during the accuracy benchmark (#11553) and partially responsible for the "all models fail on Doris/StarRocks" clusters:

**1. Silent degradation to zero connector knowledge in source checkouts / pip installs.**

`connector_metadata.json` is generated by Java reflection during the seatunnel-dist build and only exists in the release tarball layout. In a source checkout or pip install without `SEATUNNEL_HOME`, `_load_runtime_metadata()` silently returns empty (a `logger.debug`), and:
- the system-prompt catalog renders as `## Available Connectors (0 connectors, 0 endpoints)` — the model is told there are no connectors at all;
- `validate_connector_options` returns `valid: True` for everything ("metadata not available"), so generated configs are unchecked;
- `get_connector_detail("Doris")` falls through to "not found".

**2. The not-found suggester can echo the queried name back.**

`get_connector_detail` uses `route_by_keyword(name)` (keyword routing, not name correction) for suggestions, producing the absurd `Connector 'Doris' not found. Did you mean: Doris?` — because "doris" is its own keyword alias. There is also no fuzzy correction for invented names: benchmark runs observed a model writing `File` (a non-existent connector) for LocalFile, and the error gave no usable hint, so the repair loop could not recover.

- A valid connector name without available metadata should be told "known connector, option metadata unavailable (+ remediation hint)" instead of "not found".
- Unknown names should get edit-distance/substring suggestions ("File" → LocalFile, FtpFile, S3File...) that never include the queried name itself.
- The catalog should never claim "0 connectors" when a static name index exists; it should list known names and state clearly that option details are unavailable.
- The degradation should be visible (warning log), not silent.

### SeaTunnel Version

dev

### SeaTunnel Config

n/a (CLI behavior)

### Running Command

```
# any source checkout without SEATUNNEL_HOME / metadata file
seatunnel "sync PostgreSQL orders to Doris"
```

### Error Exception

See above ("Did you mean: Doris?").

_No response_

### Java or Scala Version

Python 3.10+ (seatunnel-cli)

### Screenshots

_No response_

### Code of Conduct

waterWang:
## Description

Fixes #11650

Two related defects in the CLI's connector knowledge base, both causing silent failures and misleading error messages:

### 1. Silent degradation when `connector_metadata.json` is not found

`_load_runtime_metadata()` silently returns an empty dict when no `connector_metadata.json` is found, logging only at `logger.debug` level. This means:
- The system-prompt catalog renders as `0 connectors, 0 endpoints` — the model thinks there are no connectors at all
- `validate_connector_options()` returns `valid: True` for everything with a note "metadata not available"
- `get_connector_detail("Doris")` falls through to "not found"

**Fix**: Changed the log level to `logger.warning` with a clear remediation hint (`Run \`seatunnel --export-metadata\` or set SEATUNNEL_HOME to generate it`).

### 2. Self-referencing suggestions for known connector names

`get_connector_detail()` uses `route_by_keyword()` for suggestions, which matches the queried name itself as a keyword alias. This produces the absurd: `Connector 'Doris' not found. Did you mean: Doris?`

**Fix**: 
- Filter out the queried name from suggestions (case-insensitive)
- When the name is a known connector (from `KEYWORD_ALIASES`) but metadata is unavailable, show a more helpful message: "Known connector, metadata not available — run `seatunnel --export-metadata` to generate it"

### DIFF
--- a/seatunnel-cli/seatunnel_cli/connectors.py
+++ b/seatunnel-cli/seatunnel_cli/connectors.py
@@ -340,6 +340,10 @@ def _load_runtime_metadata() -> dict | None:
                 logger.debug(f"Failed to load runtime metadata from {p}: {e}")
 
     _RUNTIME_METADATA = {}  # Empty dict = tried but not found
+    logger.warning(
+        "No connector_metadata.json found. Connector KB will be empty. "
+        "Run `seatunnel --export-metadata` or set SEATUNNEL_HOME to generate it."
+    )
     return _RUNTIME_METADATA
 
 
@@ -1070,8 +1074,17 @@ def get_connector_detail(name: str, connector_type: str | None = None) -> str | 
 
     # ── 4. Suggest ──
     suggestions = route_by_keyword(name)
+    # Filter out the queried name itself from suggestions
+    suggestions = [s for s in suggestions if s.lower() != name.lower()]
     if suggestions:
         return f"Connector '{name}' not found. Did you mean: {', '.join(suggestions)}?\nUse get_connector_info with the exact name."
+
+    # Check if the name matches a known connector alias (even without metadata)
+    all_known_connectors = set()
+    for aliases in KEYWORD_ALIASES.values():
+        all_known_connectors.update(a.lower() for a in aliases)
+    if name.lower() in all_known_connectors:
+        return f"Connector '{name}' is a known connector but its metadata is not available. Run `seatunnel --export-metadata` to generate connector metadata."
     return f"Connector '{name}' not found. Use list_connectors to see all available connectors."
 
 


SEZ9 (changes requested):
Thanks for working on this. I re-reviewed the latest head `9cad97b19343` from scratch.

# What this PR fixes
- User pain: When `connector_metadata.json` is missing, the CLI silently reports 0 connectors and validates everything as `valid: True`, and `get_connector_detail("Doris")` absurdly suggests "Did you mean: Doris?" — users get no signal that the KB is degraded or how to fix it.
- Fix approach: Promote the missing-metadata log from `debug` to `warning` with a remediation hint in `_load_runtime_metadata()`, and in `get_connector_detail()` filter the queried name out of `route_by_keyword()` suggestions plus add a `KEYWORD_ALIASES`-based fallback message pointing to `--export-metadata`.
- One-line summary: The intent is right and the changes are small, but the new fallback branch can fire (or be misleadingly skipped) in the wrong situations — it triggers even when metadata IS loaded and the name is simply wrong, it's unreachable whenever any other alias suggestion survives the filter, the case-insensitive self-filter can strip the one useful canonical-case suggestion, the warning text misdiagnoses parse failures as "file not found", and neither behavioral change has test coverage.

# Runtime chain I rechecked
```text
get_connector_detail(name, connector_type)  connectors.py:~1040
  -> _load_runtime_metadata()  connectors.py:~320   [cached module-global _RUNTIME_METADATA]
       -> probes candidate paths for connector_metadata.json; on total miss sets {} and now logger.warning(...)  connectors.py:343-346
       -> on json.JSONDecodeError per-path: logger.debug(...) only, then falls through to same 'not found' warning  connectors.py:340
  -> exact/normalized/fuzzy lookups against runtime metadata (steps 1-3, unchanged)
  -> step 4: route_by_keyword(name)  connectors.py:1076
       -> scans KEYWORD_ALIASES; the queried name matches its own alias entry -> self-suggestion
  -> NEW: suggestions = [s for s in suggestions if s.lower() != name.lower()]  connectors.py:1078
       -> if any suggestion survives -> return 'Did you mean' (skips new fallback entirely)
  -> NEW: rebuild all_known_connectors from KEYWORD_ALIASES on every miss  connectors.py:1082-1085
       -> if name.lower() in aliases -> return 'metadata not available' message even when _RUNTIME_METADATA is non-empty  connectors.py:1086-1087

System-prompt catalog path (degradation symptom):
build_system_prompt() -> render_connector_catalog() -> _load_runtime_metadata() -> {} -> '0 connectors, 0 endpoints' (now surfaced via the new warning)

validate_connector_options() -> _load_runtime_metadata() -> {} -> returns valid:True with 'metadata not available' note (behavior unchanged; only the log level changed)
```

# Findings
**Issue 1: The new 'known connector but metadata not available' message can fire even when metadata IS loaded, producing a misleading remediation hint**
- Location: `seatunnel-cli/seatunnel_cli/connectors.py:1086`
- Why it matters: - Root cause: the new fallback at `connectors.py:1086-1087` only checks whether `name.lower()` appears in `KEYWORD_ALIASES` values; it never checks whether `_load_runtime_metadata()` actually returned an empty dict. The code reaches step 4 whenever the lookups in steps 1-3 fail, which also happens when metadata is fully loaded but the connector is absent from this build (e.g., an alias kept in `KEYWORD_ALIASES` for a connector not shipped), or when the caller passed a `connector_type` that doesn't match (e.g., `get_connector_detail("Doris", connector_type="transform")`).
- Impact: users with a perfectly valid `connector_metadata.json` will be told "metadata is not available. Run `seatunnel --export-metadata`", chasing a non-existent problem, and an LLM agent consuming this tool output will loop on regenerating metadata instead of correcting the connector name/type. This is the same class of misleading-message bug the PR claims to fix.
- Fix: gate the branch on the actual metadata state, e.g. `if not _load_runtime_metadata() and name.lower() in all_known_connectors:` and keep the generic "not found, use list_connectors" message when metadata is present but simply lacks the entry (optionally mentioning the type mismatch).
- Evidence: `seatunnel-cli/seatunnel_cli/connectors.py:1087          return f"Connector '{name}' is a known connector but its metadata is not available. Run `seatunnel --export-metadata` to generate connector metadata."`; `seatunnel-cli/seatunnel_cli/connectors.py:1078      suggestions = [s for s in suggestions if s.lower() != name.lower()]`; `seatunnel-cli/seatunnel_cli/connectors.py:343      logger.warning(`
- Suggested fix: This branch fires whenever steps 1-3 fail, including when `_RUNTIME_METADATA` is fully loaded but the connector simply isn't in it (or the caller passed a non-matching `connector_type`). In that case telling the user "metadata is not available. Run `seatunnel --export-metadata`" is wrong and sends them down a dead end. Please gate this on the actual metadata state, e.g. `if not _load_runtime_metadata() and name.lower() in all_known_connectors:` so the message is only shown when the KB really is empty.
- Severity: High

**Issue 2: The new warning claims "No connector_metadata.json found" even when a metadata file exists but failed to parse — the actual parse error stays hidden at debug level, producing a misleading diagnosis.**
- Location: `seatunnel-cli/seatunnel_cli/connectors.py:344`
- Why it matters: - Root cause: in `_load_runtime_metadata()`, per-path load failures are still logged only at `logger.debug(f"Failed to load runtime metadata from {p}: {e}")` (unchanged), while the new terminal warning at lines 343-346 unconditionally says the file was not found and tells the user to regenerate it.
- Impact: if `connector_metadata.json` exists but is corrupt/malformed (e.g. a truncated export, JSON syntax error), the user is told to run `seatunnel --export-metadata` or set `SEATUNNEL_HOME` — both of which they may have already done — while the real cause (a parse failure) is invisible unless debug logging is enabled. This re-introduces exactly the "silent degradation with misleading error message" class of defect this PR sets out to fix.
- Fix: track whether any candidate path existed/was attempted; if a file was found but failed to load, log a warning that includes the path and the exception (e.g. `logger.warning(f"Failed to load connector metadata from {p}: {e}")`), and reserve the "not found" wording for the case where no candidate file exists at all.
- Evidence: `seatunnel-cli/seatunnel_cli/connectors.py:344  "No connector_metadata.json found. Connector KB will be empty. "`; `seatunnel-cli/seatunnel_cli/connectors.py:345  "Run `seatunnel --export-metadata` or set SEATUNNEL_HOME to generate it."`; `seatunnel-cli/seatunnel_cli/connectors.py:343      logger.warning(`
- Suggested fix: This warning says "No connector_metadata.json found" even when the file *was* found but failed to parse (that path still only logs at `logger.debug` above). A user with a corrupt/truncated metadata file will be told to regenerate/set SEATUNNEL_HOME when the real problem is a load error. Please distinguish the two cases: warn with the path and exception when a candidate file exists but fails to load, and keep this "not found" message only for the case where no candidate file exists.
- Severity: Medium

**Issue 3: No test coverage for the two new behaviors (self-suggestion filtering and known-connector fallback message)**
- Location: `seatunnel-cli/seatunnel_cli/connectors.py:1077`
- Why it matters: - Root cause: the PR changes only `seatunnel-cli/seatunnel_cli/connectors.py`; no test file is added or updated, even though the CLI module has an established test suite under `seatunnel-cli/tests/` (see PR #11509, which added `tests/test_validate_hocon_placeholders.py` and ran a 29-test suite).
- Impact: both fixed defects were exactly the kind of silent-degradation regressions that slip through without tests. Without a test asserting that `get_connector_detail('Doris')` never suggests 'Doris' itself, and that a known alias with empty runtime metadata yields the `--export-metadata` remediation message, a future refactor of `route_by_keyword` or `KEYWORD_ALIASES` can silently reintroduce the bug.
- Fix: add tests in `seatunnel-cli/tests/` that
- (a) monkeypatch `_load_runtime_metadata` / `_RUNTIME_METADATA` to an empty dict,
- (b) assert the self-referencing suggestion is filtered out case-insensitively, and
- (c) assert the known-connector fallback message is returned for a name present in `KEYWORD_ALIASES`.
- Evidence: `seatunnel-cli/seatunnel_cli/connectors.py:1086  return f"Connector '{name}' is a known connector but its metadata is not available. Run `seatunnel --export-metadata` to generate connector metadata."`; `seatunnel-cli/seatunnel_cli/connectors.py:1078  suggestions = [s for s in suggestions if s.lower() != name.lower()]`; `seatunnel-cli/seatunnel_cli/connectors.py:1087  return f"Connector '{name}' is a known connector but its metadata is not available. Run `seatunnel --export-metadata` to generate connector metadata."`
- Suggested fix: Could you add tests for these two new branches under `seatunnel-cli/tests/`? Specifically: (1) with runtime metadata forced empty, `get_connector_detail('Doris')` must not suggest 'Doris' itself (also cover a case-mismatched query like 'doris'), and (2) a name present in `KEYWORD_ALIASES` should return the `--export-metadata` remediation message. These regressions were silent before — tests are the only guard here.
- Severity: Medium

**Issue 4: Case-insensitive self-filter can drop the canonical-cased suggestion, downgrading a useful 'Did you mean: Doris?' into a wrong error path**
- Location: `seatunnel-cli/seatunnel_cli/connectors.py:1078`
- Why it matters: - Root cause: `connectors.py:1078` removes any suggestion where `s.lower() == name.lower()`. If the lookups in steps 1-3 are case-sensitive, a query like `get_connector_detail("doris")` falls through to step 4, and the one genuinely useful suggestion — the canonical name `Doris` — is filtered out because it only differs in case.
- Impact: instead of the previously helpful "Did you mean: Doris?", the user now hits the new fallback at line 1087 claiming metadata is unavailable (or the generic not-found message), which is strictly worse guidance for a simple casing typo.
- Fix: only filter exact-string self-matches (`s != name`), and when `s.lower() == name.lower()` but `s != name`, return "Did you mean: Doris?" so the canonical casing is surfaced. Alternatively make the step 1-3 lookup case-insensitive so this path is unreachable for casing typos, and document that assumption next to the filter.
- Evidence: `seatunnel-cli/seatunnel_cli/connectors.py:1078      suggestions = [s for s in suggestions if s.lower() != name.lower()]`; `seatunnel-cli/seatunnel_cli/connectors.py:1086      if name.lower() in all_known_connectors:`
- Suggested fix: This case-insensitive filter also removes the canonical-cased suggestion: if the step 1-3 lookup is case-sensitive, `get_connector_detail("doris")` reaches here and `Doris` gets filtered out, so the user loses the one suggestion that would have fixed their typo and instead gets the misleading 'metadata not available' fallback. Consider filtering only exact matches (`s != name`) and, when only the casing differs, still suggesting the canonical name — or verify the earlier lookup is case-insensitive and note that here.
- Severity: Medium

**Issue 5: The `all_known_connectors` alias set is rebuilt from KEYWORD_ALIASES on every failed lookup instead of being computed once at module level.**
- Location: `seatunnel-cli/seatunnel_cli/connectors.py:1083`
- Why it matters: - Root cause: lines 1083-1085 iterate all `KEYWORD_ALIASES` values and lowercase every alias inside `get_connector_detail()`, which runs on each not-found query.
- Impact: negligible per call, but this is a static, immutable derivation of a module-level constant; recomputing it per lookup is avoidable work and clutters the lookup logic. In an LLM tool-calling loop this function can be invoked many times per session.
- Fix: hoist the set to a module-level constant next to `KEYWORD_ALIASES` (e.g. `_ALL_KNOWN_CONNECTOR_ALIASES = frozenset(a.lower() for aliases in KEYWORD_ALIASES.values() for a in aliases)`) or wrap it in `@functools.lru_cache`. While doing so, also verify that every value in `KEYWORD_ALIASES` is a list/tuple of strings — if any value were a plain string, `for a in aliases` would iterate characters and silently pollute the set with single-character "connector names".
- Evidence: `seatunnel-cli/seatunnel_cli/connectors.py:1084  for aliases in KEYWORD_ALIASES.values():`; `seatunnel-cli/seatunnel_cli/connectors.py:1085  all_known_connectors.update(a.lower() for a in aliases)`; `seatunnel-cli/seatunnel_cli/connectors.py:1082  all_known_connectors = set()`
- Suggested fix: This set is derived from the static `KEYWORD_ALIASES` constant but rebuilt on every not-found lookup. I'd hoist it to a module-level `frozenset` (e.g. `_ALL_KNOWN_CONNECTOR_ALIASES = frozenset(a.lower() for aliases in KEYWORD_ALIASES.values() for a in aliases)`) computed once. Also please double-check all `KEYWORD_ALIASES` values are lists/tuples — a bare string value would make `for a in aliases` iterate characters and silently add single-char entries to this set.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: The new 'known connector but metadata not available' message can fire even when metadata IS loaded, producing a misleading remediation hint

- Issue 2: The new warning claims "No connector_metadata.json found" even when a metadata file exists but failed to parse — the actual parse error stays hidden at debug level, producing a misleading diagnosis.
- Issue 3: No test coverage for the two new behaviors (self-suggestion filtering and known-connector fallback message)
- Issue 4: Case-insensitive self-filter can drop the canonical-cased suggestion, downgrading a useful 'Did you mean: Doris?' into a wrong error path
- Issue 5: The `all_known_connectors` alias set is rebuilt from KEYWORD_ALIASES on every failed lookup instead of being computed once at module level.

DanielLeens (changes requested):

`seatunnel-cli` (the AI-assisted CLI under `seatunnel-cli/seatunnel_cli/`) ships a connector "knowledge base" that the LLM agent uses (via `get_connector_detail`, `route_by_keyword`, `validate_connector_options`, `get_connector_catalog`) to look up option rules for SeaTunnel connectors. That knowledge base is normally populated from `connector_metadata.json`, a file generated by Java reflection during the `seatunnel-dist` build. In a source checkout or pip install without `SEATUNNEL_HOME`, that file doesn't exist, and two things went wrong:

1. `_load_runtime_metadata()` degraded to an empty knowledge base silently (`logger.debug`), so nobody watching the CLI could tell why connector lookups were failing.
2. `get_connector_detail()`'s "not found" suggester used `route_by_keyword()`, which also treats a canonical connector name (e.g. `"Doris"`) as its own keyword alias, so it echoed the query back: `Connector 'Doris' not found. Did you mean: Doris?`.

This PR (1) raises the missing-metadata log to `logger.warning` with a remediation hint, and (2) filters the queried name out of suggestions and adds a distinct "known connector, metadata unavailable" message when the name matches a value in `KEYWORD_ALIASES`. This is dev-tooling / AI-agent tooling inside the Python `seatunnel-cli` package — it does not touch the Java engine, connectors, or any user-facing runtime/config surface of SeaTunnel itself.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**`_load_runtime_metadata()` (`seatunnel-cli/seatunnel_cli/connectors.py:314-347`)**

Before: on failing to find `connector_metadata.json` in any of the 5 search paths, the function set `_RUNTIME_METADATA = {}` and returned silently (only a `logger.debug` on each individual path failure).

After: the same code path now additionally does:
```python
_RUNTIME_METADATA = {}  # Empty dict = tried but not found
logger.warning(
    "No connector_metadata.json found. Connector KB will be empty. "
    "Run `seatunnel --export-metadata` or set SEATUNNEL_HOME to generate it."
)
return _RUNTIME_METADATA
```
I checked how this actually surfaces: the `seatunnel_cli` package never calls `logging.basicConfig()` anywhere, and no handler/`NullHandler` is attached to the `seatunnel_cli.connectors` logger. That means Python's logging "handler of last resort" applies, which prints records at `WARNING` and above to `stderr` by default. So this change has real, verifiable effect — previously the message was invisible (root logger default level is `WARNING`, so the `debug` call was always dropped); now it will actually print. Because `_RUNTIME_METADATA` is memoized as a global (`if _RUNTIME_METADATA is not None: return _RUNTIME_METADATA`), the warning fires at most once per process, so there's no log-spam risk.

**`get_connector_detail()` (`seatunnel-cli/seatunnel_cli/connectors.py:1022-1088`)**

```python
suggestions = route_by_keyword(name)
suggestions = [s for s in suggestions if s.lower() != name.lower()]
if suggestions:
    return f"Connector '{name}' not found. Did you mean: {', '.join(suggestions)}?..."

all_known_connectors = set()
for aliases in KEYWORD_ALIASES.values():
    all_known_connectors.update(a.lower() for a in aliases)
if name.lower() in all_known_connectors:
    return f"Connector '{name}' is a known connector but its metadata is not available. ..."
return f"Connector '{name}' not found. Use list_connectors to see all available connectors."
```
I traced `route_by_keyword()` (`connectors.py:1193-1219`) and `KEYWORD_ALIASES` (`connectors.py:709-798`) to verify the root cause and the fix. `KEYWORD_ALIASES` maps user-typed keywords to canonical connector names, e.g. `"doris": ["Doris"]`. `route_by_keyword("Doris")` lower-cases to `"doris"`, matches that key (substring match `keyword in text_lower`), and returns `["Doris"]` — i.e. it echoes the exact queried name back as a "suggestion." The new filter (`s.lower() != name.lower()`) removes exactly that case. When the queried name is itself a canonical connector name found among `KEYWORD_ALIASES` *values* (not keys), the new fallback correctly reports "known connector, metadata unavailable" instead of the generic "not found." I manually traced `get_connector_detail("Doris")` with an empty metadata dict end-to-end and confirmed it now returns the intended "known connector ... metadata not available" message instead of the self-referencing one. This part of the fix is correct and precise.

**Important scope gap versus the PR's own stated motivation:** The PR description (and the linked issue #11650) frames the *primary* symptom of defect #1 as: "The system-prompt catalog renders as `0 connectors, 0 endpoints` — the model thinks there are no connectors at all." That catalog is built by `get_connector_catalog()` (`connectors.py:965-1019`), which is **not touched by this diff at all**. When `runtime_meta` is empty, `connectors` stays empty, `count=0`, and the header still renders `## Available Connectors (0 connectors, 0 endpoints)`. A `logger.warning()` call is invisible to the LLM — it never becomes part of the system prompt or any tool response the model sees. So after this PR, in the exact scenario it targets (no `connector_metadata.json`), **the model-facing knowledge base is exactly as empty as before** — only a human watching the terminal now gets a warning. The linked issue's third bullet ("The catalog should never claim '0 connectors' ... it should list known names") is explicitly not addressed. This means the underlying accuracy problem cited in #11650 ("partially responsible for the 'all models fail on Doris/StarRocks' clusters") is not actually fixed by this PR — only observability for a human operator improves. See Issue 2 below.

No engine execution flow, checkpointing, serialization, or config `Option` contracts are touched by this diff — this is purely internal Python CLI tooling, so most of the standard engine-review checklist (state machines, recovery paths, SinkWriter contracts, etc.) does not apply here.

## 1.2 Compatibility Impact

**Fully compatible.** This change is confined to `seatunnel-cli/seatunnel_cli/connectors.py`, touches only an internal log statement and a string-formatting/suggestion branch of a helper function used by the CLI's own agent tool-calling loop. No public API, CLI flag, config `Option`, file format, or SeaTunnel engine/connector behavior is changed. Existing return types and function signatures are unchanged.

Negligible. Building `all_known_connectors` iterates `KEYWORD_ALIASES` (roughly 100 short entries) only on the not-found fallback path, and only after the (already existing) `route_by_keyword()` call. The extra `logger.warning()` call only fires once per process due to memoization. No new I/O, no new loops in hot paths.

### Issue 1 — PR title does not follow repository convention
- **Location:** PR title / commit message (`fix: address silent connector KB degradation and self-referencing suggestions`)
- **Problem:** Per this repository's contribution convention, commit/PR titles should follow `[Type][Module] Description` (e.g. `[Fix][Module] ...`), not a free-form Conventional-Commits-style `fix: ...` prefix.
- **Potential risk:** Inconsistent history, harder to search/changelog-generate by type/module.
- **Best improvement:** Rename the PR title to something like `[Fix][CLI] Address silent connector KB degradation and self-referencing suggestions`.
- **Severity:** Low

### Issue 2 — The PR does not fix the "0 connectors, 0 endpoints" system-prompt degradation it describes as the main symptom
- **Location:** `seatunnel-cli/seatunnel_cli/connectors.py:965-1019` (`get_connector_catalog()`, untouched by this diff)
- **Problem:** The PR description and linked issue #11650 both cite the system prompt rendering `## Available Connectors (0 connectors, 0 endpoints)` as the headline consequence of missing metadata. `get_connector_catalog()`, which builds that exact string for the LLM's system prompt, is not modified by this diff. `logger.warning()` output is not part of the LLM's context, so the model still receives zero connector information in the exact scenario this PR targets.
- **Potential risk:** The PR may be perceived (and issue #11650 may be closed) as resolving the underlying "AI generates broken configs / fails on Doris/StarRocks" problem, when in fact only human-facing CLI diagnostics improved. The actual knowledge-base-to-LLM pipeline is unchanged in the no-metadata case.
- **Best improvement:**
  - Option A: Update the PR title/description to explicitly scope this PR to "human-facing degradation visibility + self-reference suggestion fix" only, and keep #11650's catalog-listing requirement open as a distinct follow-up.
  - Option B: Extend `get_connector_catalog()` to fall back to listing the static `KEYWORD_ALIASES`-derived connector name index (with a "details unavailable" note) when `runtime_meta` is empty, which would actually close the loop described in the issue.
- **Severity:** Medium

### Issue 3 — Duplicate, inconsistent-format messaging for the same degraded state
- **Location:** `seatunnel-cli/seatunnel_cli/connectors.py:342-346` and `seatunnel-cli/seatunnel_cli/cli.py:915-929`
- **Problem:** `cli.py`'s `_check_env()` (called once at REPL/`run_single` startup) already prints a Rich-formatted status line — "Metadata: not found — using built-in defaults for common connectors" — right after calling `_load_runtime_metadata()`. With this PR, the same call now also emits a raw, differently-worded `logging` line to stderr ("No connector_metadata.json found. Connector KB will be empty. ..."). Users get two different messages, in two different formats/channels, for the same single event, every time they start the CLI without metadata.
- **Potential risk:** Minor UX noise/confusion; the two messages don't share wording, so users may wonder if they're two separate issues.
- **Best improvement:** Either drop the new `logger.warning` in favor of the existing Rich status line (if `_check_env()` is guaranteed to run before any KB lookup), or keep the warning but reuse the same phrasing so it clearly reads as the same event.
- **Severity:** Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

The diff is small, idiomatic Python, and consistent with the surrounding file's style (comments, string formatting, docstring conventions). The two added inline comments (`# Filter out the queried name itself from suggestions`, `# Check if the name matches a known connector alias (even without metadata)`) are appropriately concise and explain intent rather than restating the code — good practice. See Issue 1 for the PR-title convention nit. No missing-comment concerns on the changed lines themselves; the surrounding functions already carry docstrings.

No test files were changed in this diff (`git diff --stat` shows only `connectors.py`). `seatunnel-cli/tests/test_connector_metadata.py` has an established, easy-to-reuse pattern for exactly this kind of unit test — e.g. `test_connector_detail_can_fetch_transform_metadata_from_runtime_api` already does `patch.object(connectors, "_load_runtime_metadata", return_value={})` before calling `get_connector_detail(...)`. Neither of the two behavioral changes in this PR is covered:
- No test asserts that `get_connector_detail("Doris")` (with empty runtime metadata) no longer returns "Did you mean: Doris?" and instead returns the "known connector, metadata not available" message.
- No test asserts that a genuinely unknown name still gets the plain "not found" message, and that the self-reference filter doesn't over-filter legitimate distinct suggestions.
- No test asserts the `logger.warning` fires (e.g. via `assertLogs`) when metadata is missing.

Since the flaky-test checklist only applies when the PR changes UT/E2E code, and it does not here, no stability rating is required — but the complete absence of new tests for new/changed *behavior* is a real coverage gap for a bugfix PR, worth Medium severity, folded into Issue 2/3's improvement scope. I recommend adding at least 2-3 focused unit tests before merge, following the existing `patch.object(connectors, "_load_runtime_metadata", ...)` pattern in the same test file.

## 2.3 Documentation Updates

Not applicable in the sense of `docs/en`/`docs/zh` — `seatunnel-cli` is documented under `docs/en/ai-cli` / `docs/zh/ai-cli` and in its own `README.md`, and this change is an internal error-message/log-level improvement with no new config option, flag, or documented behavior contract. I checked `docs/en/ai-cli/design.md` and the CLI `README.md`; both already describe the "`connector_metadata.json` missing → degraded mode" behavior at a conceptual level, and neither documents exact log/error wording, so no update is strictly required here.

# 3. Architectural Soundness

Precise fix for the two specific behaviors it targets (log visibility, self-reference suggestion echo). It is a minimal, surgical change — no unrelated refactors. However, as noted in Issue 2, it is a partial fix relative to the PR's own stated problem framing (the system-prompt catalog still shows "0 connectors").

## 3.2 Maintainability
The added logic is short, readable, and follows the file's existing patterns (module-level `KEYWORD_ALIASES` lookup, memoized global metadata dict). No new abstractions introduced.

## 3.3 Extensibility
`all_known_connectors` is derived only from `KEYWORD_ALIASES` values, so any canonical connector name that doesn't appear as a value in that dict (i.e., has no keyword alias mapped to it) won't benefit from the new "known connector, metadata unavailable" message and will instead fall through to the generic "not found" message. This is a soft limitation, not a bug — worth keeping in mind if `KEYWORD_ALIASES` coverage is incomplete relative to the full connector set.

No historical/serialization/state concerns — this is stateless string-formatting and logging-level logic inside a Python dev-tool, with no persisted state or wire format involved.

# 4. Issue Summary

|-----|-------|----------|----------|
| 2 | PR does not fix the "0 connectors, 0 endpoints" system-prompt degradation described as the main symptom (`get_connector_catalog()` untouched) | seatunnel-cli/seatunnel_cli/connectors.py:965-1019 | Medium |
| 3 | Duplicate/inconsistent messaging for the same degraded-metadata event (Rich status line + raw log line) | seatunnel-cli/seatunnel_cli/connectors.py:342-346, seatunnel-cli/seatunnel_cli/cli.py:915-929 | Low |
| 1 | PR title does not follow `[Type][Module] Description` convention | PR title | Low |

# 5. Merge Recommendation

1. Blockers — must be fixed
   - CI has not actually run: the "Build" check on this PR resolves to `action_required` with the message "Unable to detect the workflow run for testing the changes in your PR," and I confirmed via the GitHub API that the contributor's fork (`waterWang/seatunnel`) has no Actions runs at all for this branch. GitHub Actions needs to be enabled/triggered on the fork (or the branch rebased onto latest `dev` and pushed again) so the real Build/UT workflow actually executes before this can be verified and merged. See the check-run: https://github.com/apache/seatunnel/runs/92069932150
   - Add at least minimal unit test coverage for the two new behaviors (self-reference filtering / known-connector fallback message, and the warning log), following the existing `patch.object(connectors, "_load_runtime_metadata", ...)` pattern already used in `tests/test_connector_metadata.py` (see Test Coverage section above).

   - Fix the PR title to follow `[Fix][CLI] ...` convention (Issue 1).
   - Clarify scope: either extend `get_connector_catalog()` to list known connector names from `KEYWORD_ALIASES` when metadata is unavailable (closing the loop the PR's own description implies), or adjust the PR/issue framing so it's clear this PR only fixes visibility + suggestion-echo, not the model-facing "0 connectors" catalog problem (Issue 2).
   - Consolidate the two different messages shown for the same "no metadata" event into one consistent message/channel (Issue 3).

**Overall assessment:** The code that this PR *does* change is correct, minimal, and well-targeted — I traced both fixes end-to-end (the log-level change genuinely makes a previously-invisible warning visible via Python's default stderr handler, and the suggestion self-reference bug is fixed exactly as described). The main concern is that the PR's own narrative — echoed from the linked issue — sets an expectation ("system-prompt catalog renders as 0 connectors, 0 endpoints... the model thinks there are no connectors at all") that this diff does not actually resolve, since `get_connector_catalog()` is untouched. I'd hate for issue #11650 to get closed as fully fixed when the actual LLM-facing knowledge gap remains. No better alternative implementation exists for the two behaviors that *are* implemented; the request is simply to either broaden the fix or narrow the claim, add a couple of tests, and get CI running.

Thank you for tackling this, waterWang — the root-cause tracing in the issue and PR description is genuinely well done, and the two fixes you did implement are clean and correct. Happy to take another look as soon as CI is running and the catalog-scope question above is addressed either way.

davidzollo:

```

```

PR #11654, opened 2026-08-04
[Improve][Connector-V2][Airtable] Add jitter to rate-limit backoff

1991santhu:

`AirtableSourceReader` and `AirtableSinkWriter` both retry on HTTP 429 with a backoff that was a pure function of the retry count. Airtable rate-limits per base rather than per client, so every reader and writer that gets limited together also retries at the same instants, and the burst that caused the 429 reforms on each attempt.

This adds jitter to `calculateBackoffMillis` so concurrent readers and writers spread out instead.

### What the jitter does

Below the cap, the jitter is **added on top** of the scheduled wait, giving `[wait, min(2*wait, MAX_BACKOFF_MILLIS)]`. The scheduled value is the floor rather than the midpoint, because this fires on a 429 and retrying sooner than `rate_limit_backoff_ms` asked for would work against the setting.

**This raises the worst-case wait.** With the connector defaults (`rate_limit_backoff_ms=30000`, `rate_limit_max_retries=3`) the old worst case across three retries was 30000 + 60000 + 120000 = 210000ms. The new worst case is 60000 + 120000 + 240000 = 420000ms, until the cap absorbs the difference. Operators tuning `rate_limit_backoff_ms` alongside checkpoint timeouts should know that.

At the cap there is no headroom left to add into, so the wait is spread downwards instead, floored at half the wait and raised to `rate_limit_backoff_ms` when that is lower than the wait. `MAX_BACKOFF_MILLIS` is an upper bound rather than a target, so drawing below it breaks nothing.

No new or changed options, defaults, or serialization. The only observable difference is the distribution of retry sleep durations, and the worst case noted above. `rate_limit_backoff_ms=0` still disables backoff entirely.

Nine unit tests per class, in both `AirtableSourceReaderTest` and `AirtableSinkWriterTest`, calling `calculateBackoffMillis` directly as a pure function. No sleeps, no wall-clock assertions.

```
Tests run: 9  AirtableSourceReaderTest
Tests run: 9  AirtableSinkWriterTest
Tests run: 18, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The dead-zone regression test was checked against the unfixed code first and fails there with `observed only [300000]`.

### Check list

* [x] Code changed are covered with tests

### DIFF
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
@@ -197,10 +197,22 @@ public class AirtableSinkWriter extends AbstractSinkWriter<SeaTunnelRow, Void>
         // sooner than configured would work against the setting's purpose. The
         // result stays capped at MAX_BACKOFF_MILLIS.
         long extra = Math.min(waitMillis, MAX_BACKOFF_MILLIS - waitMillis);
-        if (extra <= 0) {
+        if (extra > 0) {
+            return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
+        }
+
+        // Once the wait reaches MAX_BACKOFF_MILLIS there is no headroom left to
+        // add into, so every retry past that point would come back unjittered
+        // and the callers would be back in lockstep exactly when the rate limit
+        // is at its most persistent. Spread the wait downwards instead, floored
+        // at rateLimitBackoffMs: the cap is an upper bound rather than a target,
+        // so drawing below it breaks nothing, whereas dropping below the
+        // configured backoff would break the guarantee described above.
+        long floor = Math.max(waitMillis / 2, rateLimitBackoffMs);
+        if (floor >= waitMillis) {
             return waitMillis;
         }
-        return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
+        return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
@@ -22,6 +22,8 @@ import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.node.ArrayNode;
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.node.ObjectNode;
 
+import org.apache.seatunnel.shade.com.google.common.annotations.VisibleForTesting;
+
 import org.apache.seatunnel.api.sink.SupportMultiTableSinkWriter;
 import org.apache.seatunnel.api.table.type.SeaTunnelRow;
 import org.apache.seatunnel.api.table.type.SeaTunnelRowType;
@@ -39,6 +41,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+
+import java.util.concurrent.ThreadLocalRandom;
 
 @Slf4j
 public class AirtableSinkWriter extends AbstractSinkWriter<SeaTunnelRow, Void>
@@ -175,13 +179,23 @@ public class AirtableSinkWriter extends AbstractSinkWriter<SeaTunnelRow, Void>
         lastRequestTimeMillis = System.currentTimeMillis();
     }
 
-    private long calculateBackoffMillis(int retryCount) {
+    @VisibleForTesting
+    long calculateBackoffMillis(int retryCount) {
         if (rateLimitBackoffMs <= 0) {
             return 0L;
         }
         long exponential = 1L << Math.min(20, Math.max(0, retryCount - 1));
-        long waitMillis = rateLimitBackoffMs * exponential;
-        return Math.min(waitMillis, MAX_BACKOFF_MILLIS);
+        long waitMillis = Math.min(rateLimitBackoffMs * exponential, MAX_BACKOFF_MILLIS);
+
+        // Equal jitter: retain half the computed delay as a floor and randomise
+        // the remainder. Without this the delay is a pure function of the retry
+        // count, so every reader and writer that hits the rate limit at the same
+        // moment retries at the same instants and the burst that caused the 429
+        // reforms on each attempt. Half is kept rather than using full jitter so
+        // that a retry is never issued immediately against an API that has just
+        // asked us to slow down.
+        long half = waitMillis / 2;
+        return half + ThreadLocalRandom.current().nextLong(waitMillis - half + 1);
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
@@ -187,15 +187,20 @@ public class AirtableSinkWriter extends AbstractSinkWriter<SeaTunnelRow, Void>
         long exponential = 1L << Math.min(20, Math.max(0, retryCount - 1));
         long waitMillis = Math.min(rateLimitBackoffMs * exponential, MAX_BACKOFF_MILLIS);
 
-        // Equal jitter: retain half the computed delay as a floor and randomise
-        // the remainder. Without this the delay is a pure function of the retry
-        // count, so every reader and writer that hits the rate limit at the same
-        // moment retries at the same instants and the burst that caused the 429
-        // reforms on each attempt. Half is kept rather than using full jitter so
-        // that a retry is never issued immediately against an API that has just
-        // asked us to slow down.
-        long half = waitMillis / 2;
-        return half + ThreadLocalRandom.current().nextLong(waitMillis - half + 1);
+        // Spread the delay by adding a random amount on top of it. Without this
+        // the delay is a pure function of the retry count, so every reader and
+        // writer that hits the rate limit at the same moment retries at the same
+        // instants and the burst that caused the 429 reforms on each attempt.
+        //
+        // The jitter is added rather than centred so the wait is never shorter
+        // than rateLimitBackoffMs asked for: this fires on 429, so retrying
+        // sooner than configured would work against the setting's purpose. The
+        // result stays capped at MAX_BACKOFF_MILLIS.
+        long extra = Math.min(waitMillis, MAX_BACKOFF_MILLIS - waitMillis);
+        if (extra <= 0) {
+            return waitMillis;
+        }
+        return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
@@ -204,17 +204,20 @@ public class AirtableSinkWriter extends AbstractSinkWriter<SeaTunnelRow, Void>
         // and the callers would be back in lockstep exactly when the rate limit
         // is at its most persistent. Spread the wait downwards instead. The cap
         // is an upper bound rather than a target, so drawing below it breaks
-        // nothing, whereas dropping below the configured backoff would break the
-        // guarantee described above.
+        // nothing.
         //
-        // The floor is half the wait, raised to rateLimitBackoffMs when that is
-        // the tighter of the two. It is deliberately not raised when the
-        // configured backoff is at or above the wait itself: the cap has already
-        // put the wait below what was configured, so the guarantee is moot, and
-        // flooring there would leave no room to jitter at all.
+        // The floor is the last scheduled wait that still fitted under the cap,
+        // or half the wait when the very first retry is already capped. Flooring
+        // there keeps the minimum from dropping as the schedule crosses the cap:
+        // half of MAX can be less than the previous retry's wait, which would let
+        // a later retry sleep for less than an earlier one. It also keeps the
+        // wait at or above rateLimitBackoffMs for free, since the last uncapped
+        // wait is never smaller than the configured backoff.
         long floor = waitMillis / 2;
-        if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis) {
-            floor = rateLimitBackoffMs;
+        for (long scheduled = rateLimitBackoffMs; scheduled < MAX_BACKOFF_MILLIS; scheduled <<= 1) {
+            if (scheduled > floor) {
+                floor = scheduled;
+            }
         }
         return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
     }
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriter.java
@@ -21,7 +21,6 @@ import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.JsonNode;
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.node.ArrayNode;
 import org.apache.seatunnel.shade.com.fasterxml.jackson.databind.node.ObjectNode;
-
 import org.apache.seatunnel.shade.com.google.common.annotations.VisibleForTesting;
 
 import org.apache.seatunnel.api.sink.SupportMultiTableSinkWriter;
@@ -41,7 +40,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-
 import java.util.concurrent.ThreadLocalRandom;
 
 @Slf4j
@@ -204,13 +202,19 @@ public class AirtableSinkWriter extends AbstractSinkWriter<SeaTunnelRow, Void>
         // Once the wait reaches MAX_BACKOFF_MILLIS there is no headroom left to
         // add into, so every retry past that point would come back unjittered
         // and the callers would be back in lockstep exactly when the rate limit
-        // is at its most persistent. Spread the wait downwards instead, floored
-        // at rateLimitBackoffMs: the cap is an upper bound rather than a target,
-        // so drawing below it breaks nothing, whereas dropping below the
-        // configured backoff would break the guarantee described above.
-        long floor = Math.max(waitMillis / 2, rateLimitBackoffMs);
-        if (floor >= waitMillis) {
-            return waitMillis;
+        // is at its most persistent. Spread the wait downwards instead. The cap
+        // is an upper bound rather than a target, so drawing below it breaks
+        // nothing, whereas dropping below the configured backoff would break the
+        // guarantee described above.
+        //
+        // The floor is half the wait, raised to rateLimitBackoffMs when that is
+        // the tighter of the two. It is deliberately not raised when the
+        // configured backoff is at or above the wait itself: the cap has already
+        // put the wait below what was configured, so the guarantee is moot, and
+        // flooring there would leave no room to jitter at all.
+        long floor = waitMillis / 2;
+        if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis) {
+            floor = rateLimitBackoffMs;
         }
         return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
     }
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
@@ -140,17 +140,20 @@ public class AirtableSourceReader extends HttpSourceReader {
         // and the callers would be back in lockstep exactly when the rate limit
         // is at its most persistent. Spread the wait downwards instead. The cap
         // is an upper bound rather than a target, so drawing below it breaks
-        // nothing, whereas dropping below the configured backoff would break the
-        // guarantee described above.
+        // nothing.
         //
-        // The floor is half the wait, raised to rateLimitBackoffMs when that is
-        // the tighter of the two. It is deliberately not raised when the
-        // configured backoff is at or above the wait itself: the cap has already
-        // put the wait below what was configured, so the guarantee is moot, and
-        // flooring there would leave no room to jitter at all.
+        // The floor is the last scheduled wait that still fitted under the cap,
+        // or half the wait when the very first retry is already capped. Flooring
+        // there keeps the minimum from dropping as the schedule crosses the cap:
+        // half of MAX can be less than the previous retry's wait, which would let
+        // a later retry sleep for less than an earlier one. It also keeps the
+        // wait at or above rateLimitBackoffMs for free, since the last uncapped
+        // wait is never smaller than the configured backoff.
         long floor = waitMillis / 2;
-        if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis) {
-            floor = rateLimitBackoffMs;
+        for (long scheduled = rateLimitBackoffMs; scheduled < MAX_BACKOFF_MILLIS; scheduled <<= 1) {
+            if (scheduled > floor) {
+                floor = scheduled;
+            }
         }
         return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
     }
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
@@ -121,14 +121,19 @@ public class AirtableSourceReader extends HttpSourceReader {
         long exponential = 1L << Math.min(20, Math.max(0, retryCount - 1));
         long waitMillis = Math.min(rateLimitBackoffMs * exponential, MAX_BACKOFF_MILLIS);
 
-        // Equal jitter: retain half the computed delay as a floor and randomise
-        // the remainder. Without this the delay is a pure function of the retry
-        // count, so every reader and writer that hits the rate limit at the same
-        // moment retries at the same instants and the burst that caused the 429
-        // reforms on each attempt. Half is kept rather than using full jitter so
-        // that a retry is never issued immediately against an API that has just
-        // asked us to slow down.
-        long half = waitMillis / 2;
-        return half + ThreadLocalRandom.current().nextLong(waitMillis - half + 1);
+        // Spread the delay by adding a random amount on top of it. Without this
+        // the delay is a pure function of the retry count, so every reader and
+        // writer that hits the rate limit at the same moment retries at the same
+        // instants and the burst that caused the 429 reforms on each attempt.
+        //
+        // The jitter is added rather than centred so the wait is never shorter
+        // than rateLimitBackoffMs asked for: this fires on 429, so retrying
+        // sooner than configured would work against the setting's purpose. The
+        // result stays capped at MAX_BACKOFF_MILLIS.
+        long extra = Math.min(waitMillis, MAX_BACKOFF_MILLIS - waitMillis);
+        if (extra <= 0) {
+            return waitMillis;
+        }
+        return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
     }
 }
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
@@ -17,6 +17,8 @@
 
 package org.apache.seatunnel.connectors.seatunnel.airtable.source;
 
+import org.apache.seatunnel.shade.com.google.common.annotations.VisibleForTesting;
+
 import org.apache.seatunnel.api.serialization.DeserializationSchema;
 import org.apache.seatunnel.api.table.type.SeaTunnelRow;
 import org.apache.seatunnel.connectors.seatunnel.common.source.SingleSplitReaderContext;
@@ -27,6 +29,8 @@ import org.apache.seatunnel.connectors.seatunnel.http.config.PageInfo;
 import org.apache.seatunnel.connectors.seatunnel.http.source.HttpSourceReader;
 
 import lombok.extern.slf4j.Slf4j;
+
+import java.util.concurrent.ThreadLocalRandom;
 
 @Slf4j
 public class AirtableSourceReader extends HttpSourceReader {
@@ -109,12 +113,22 @@ public class AirtableSourceReader extends HttpSourceReader {
         lastRequestTimeMillis = System.currentTimeMillis();
     }
 
-    private long calculateBackoffMillis(int retryCount) {
+    @VisibleForTesting
+    long calculateBackoffMillis(int retryCount) {
         if (rateLimitBackoffMs <= 0) {
             return 0L;
         }
         long exponential = 1L << Math.min(20, Math.max(0, retryCount - 1));
-        long waitMillis = rateLimitBackoffMs * exponential;
-        return Math.min(waitMillis, MAX_BACKOFF_MILLIS);
+        long waitMillis = Math.min(rateLimitBackoffMs * exponential, MAX_BACKOFF_MILLIS);
+
+        // Equal jitter: retain half the computed delay as a floor and randomise
+        // the remainder. Without this the delay is a pure function of the retry
+        // count, so every reader and writer that hits the rate limit at the same
+        // moment retries at the same instants and the burst that caused the 429
+        // reforms on each attempt. Half is kept rather than using full jitter so
+        // that a retry is never issued immediately against an API that has just
+        // asked us to slow down.
+        long half = waitMillis / 2;
+        return half + ThreadLocalRandom.current().nextLong(waitMillis - half + 1);
     }
 }
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
@@ -138,13 +138,19 @@ public class AirtableSourceReader extends HttpSourceReader {
         // Once the wait reaches MAX_BACKOFF_MILLIS there is no headroom left to
         // add into, so every retry past that point would come back unjittered
         // and the callers would be back in lockstep exactly when the rate limit
-        // is at its most persistent. Spread the wait downwards instead, floored
-        // at rateLimitBackoffMs: the cap is an upper bound rather than a target,
-        // so drawing below it breaks nothing, whereas dropping below the
-        // configured backoff would break the guarantee described above.
-        long floor = Math.max(waitMillis / 2, rateLimitBackoffMs);
-        if (floor >= waitMillis) {
-            return waitMillis;
+        // is at its most persistent. Spread the wait downwards instead. The cap
+        // is an upper bound rather than a target, so drawing below it breaks
+        // nothing, whereas dropping below the configured backoff would break the
+        // guarantee described above.
+        //
+        // The floor is half the wait, raised to rateLimitBackoffMs when that is
+        // the tighter of the two. It is deliberately not raised when the
+        // configured backoff is at or above the wait itself: the cap has already
+        // put the wait below what was configured, so the guarantee is moot, and
+        // flooring there would leave no room to jitter at all.
+        long floor = waitMillis / 2;
+        if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis) {
+            floor = rateLimitBackoffMs;
         }
         return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
     }
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/main/java/org/apache/seatunnel/connectors/seatunnel/airtable/source/AirtableSourceReader.java
@@ -131,9 +131,21 @@ public class AirtableSourceReader extends HttpSourceReader {
         // sooner than configured would work against the setting's purpose. The
         // result stays capped at MAX_BACKOFF_MILLIS.
         long extra = Math.min(waitMillis, MAX_BACKOFF_MILLIS - waitMillis);
-        if (extra <= 0) {
+        if (extra > 0) {
+            return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
+        }
+
+        // Once the wait reaches MAX_BACKOFF_MILLIS there is no headroom left to
+        // add into, so every retry past that point would come back unjittered
+        // and the callers would be back in lockstep exactly when the rate limit
+        // is at its most persistent. Spread the wait downwards instead, floored
+        // at rateLimitBackoffMs: the cap is an upper bound rather than a target,
+        // so drawing below it breaks nothing, whereas dropping below the
+        // configured backoff would break the guarantee described above.
+        long floor = Math.max(waitMillis / 2, rateLimitBackoffMs);
+        if (floor >= waitMillis) {
             return waitMillis;
         }
-        return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
+        return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
     }
 }
--- a/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/test/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriterTest.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-airtable/src/test/java/org/apache/seatunnel/connectors/seatunnel/airtable/sink/AirtableSinkWriterTest.java
@@ -220,6 +220,28 @@ public class AirtableSinkWriterTest {
     }
 
     @Test
+    public void testBackoffMinimumNeverDropsAtTheCapBoundary() {
+        // The scheduled wait doubles until it hits the cap. Half the cap can be
+        // less than the wait the previous retry already guaranteed, so without a
+        // floor the first capped retry could sleep for less than the one before
+        // it, which is not what anyone expects from exponential backoff.
+        int base = 100000;
+        AirtableSinkWriter writer = writerWithBackoff(base, 30);
+
+        // With a 100000ms base the schedule is 100000, 200000, then capped at
+        // 300000. The last wait that fitted under the cap was 200000, so no
+        // capped retry should ever draw below that. Half the cap, 150000, would.
+        for (int retry = 3; retry <= 8; retry++) {
+            for (int i = 0; i < 200; i++) {
+                long actual = writer.calculateBackoffMillis(retry);
+                Assertions.assertTrue(
+                        actual >= 200000L && actual <= 300000L,
+                        "retry " + retry + " produced " + actual + ", expected [200000, 300000]");
+            }
+        }
+    }
+
+    @Test
     public void testBackoffStillVariesWhenBaseExceedsTheMaximum() {
         AirtableSinkWriter writer = writerWithBackoff(300000, 30);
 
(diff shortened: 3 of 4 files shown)


1991santhu:
Corrected two things in my own patch since opening this.

The jitter was centred, which can retry sooner than `rate_limit_backoff_ms` is set to. That's wrong for something that fires on a 429, so it adds now.

The second one I missed first time. At `MAX_BACKOFF_MILLIS` there's no headroom left to add into, so my clamp left those retries unjittered. It spreads downward there now, floored at `rate_limit_backoff_ms`. You need `rate_limit_max_retries >= 5` to reach it; the defaults stop at 120s, so this isn't a default-path bug.

I ran the numbers instead of just asserting it. 100 parallel subtasks sharing one Airtable quota: without jitter all 100 retry inside the same second on every attempt, with it about 8 then 4. Simulated arrival times, not something I've observed against Airtable.

Both `AirtableSourceReader` and `AirtableSinkWriter` have the same method, so both are changed. Tests on JDK 11: 8 and 2, BUILD SUCCESS.

DanielLeens (changes requested):

`AirtableSourceReader` and `AirtableSinkWriter` both retry on HTTP 429 with an exponential backoff that was, before this PR, a pure function of `retryCount`. Because Airtable rate-limits per base rather than per client, every reader/writer that gets rate-limited together also retries at the exact same instants, so the burst that triggered the 429 reassembles on every retry attempt (a classic "thundering herd" pattern). This PR adds randomised jitter to `calculateBackoffMillis()` so concurrent readers/writers spread their retries out instead of hammering the API in lockstep, while trying to guarantee the wait is never shorter than the configured `rate_limit_backoff_ms` and never exceeds the existing `MAX_BACKOFF_MILLIS` (300000ms) cap.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The change is confined to one duplicated private method, `calculateBackoffMillis(int retryCount)`, present identically in:
- `AirtableSourceReader.java` (now `@VisibleForTesting long calculateBackoffMillis(...)`)
- `AirtableSinkWriter.java` (same signature/body)

Before:
```java
long waitMillis = rateLimitBackoffMs * exponential;
return Math.min(waitMillis, MAX_BACKOFF_MILLIS);
```

After:
```java
long waitMillis = Math.min(rateLimitBackoffMs * exponential, MAX_BACKOFF_MILLIS);
long extra = Math.min(waitMillis, MAX_BACKOFF_MILLIS - waitMillis);
if (extra > 0) {
    return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
}
long floor = Math.max(waitMillis / 2, rateLimitBackoffMs);
if (floor >= waitMillis) {
    return waitMillis;
}
return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
```

Runtime path when a 429 is received (identical in both classes):
```
executeRequest()/sendWithRateLimitRetry()
  -> response.getCode() == 429 && retryCount < rateLimitMaxRetries
  -> retryCount++
  -> calculateBackoffMillis(retryCount)
       -> waitMillis not yet at cap  => range [waitMillis, min(2*waitMillis, MAX_BACKOFF_MILLIS)]  (jitter added ON TOP)
       -> waitMillis == cap (300000) => range [max(150000, rateLimitBackoffMs), 300000]            (jitter spread DOWNWARD)
  -> Thread.sleep(backoffMillis)
  -> continue loop / retry
```

This is a genuinely correct fix for the thundering-herd problem the PR describes, and `ThreadLocalRandom.current()` is the right, thread-safe choice for a value computed per-retry across parallel reader/writer instances. Two correctness/consistency concerns found while tracing the branches through concrete numbers (Airtable connector defaults: `rate_limit_backoff_ms=30000`, `rate_limit_max_retries=3`, `MAX_BACKOFF_MILLIS=300000`):

1. **The "add jitter on top" branch changes the previously-guaranteed upper bound.** Before this PR, retry *N* always waited exactly `min(rateLimitBackoffMs * 2^(N-1), MAX_BACKOFF_MILLIS)`. After this PR, retry *N* can wait up to `min(2 * that value, MAX_BACKOFF_MILLIS)` — i.e. up to roughly double, until the cap absorbs the difference. With the connector's own documented defaults, the old worst-case total sleep across 3 retries was `30000+60000+120000=210000ms`; the new worst-case is `60000+120000+240000=420000ms`. The PR description states "The result never exceeds the previous value or `MAX_BACKOFF_MILLIS`, so no caller waits longer than it does today — the change only removes the alignment." That claim was true of an earlier iteration of this PR (the "equal jitter" `half + random(0, wait-half)` approach quoted in the PR description body) but is **not true of the code that is actually in the diff today** (jitter is added on top, not centered). See Issue 2 below — this isn't a functional bug, but it does mean the stated latency guarantee in the PR description does not match the shipped behavior, and it's exactly the kind of change that can push a downstream checkpoint/timeout budget further than the author's own write-up implies.

2. **The "spread downward at the cap" branch has a zero-jitter dead zone.** When a user configures `rate_limit_backoff_ms >= MAX_BACKOFF_MILLIS` (300000ms / 5 minutes), `waitMillis` is `MAX_BACKOFF_MILLIS` from the very first retry, and `floor = Math.max(waitMillis/2, rateLimitBackoffMs)` becomes `>= waitMillis`, so the method falls into `if (floor >= waitMillis) return waitMillis;` — i.e. **every retry returns exactly 300000ms with zero jitter**, silently reproducing the exact lockstep problem this PR exists to fix. This is precisely the "at the maximum" scenario the third commit (`spread the backoff once it reaches the maximum`) was written to address, and it's defeated by a configuration that a cautious operator (someone deliberately setting a large, conservative backoff) is quite likely to choose. See Issue 1.

Idempotency / concurrency: this PR does not change *what* gets retried, only *how long* the caller sleeps before retrying — `sendWithRateLimitRetry` still only retries on a 429, and a 429 from Airtable is returned before any records are written, so the retry itself remains safe for the sink's batch POST (no double-write risk from this change). `calculateBackoffMillis` is a stateless pure function reading only immutable fields (`rateLimitBackoffMs`, `MAX_BACKOFF_MILLIS`) and `ThreadLocalRandom`, so there's no new shared-state race between the (single-threaded-per-instance) reader/writer.

## 1.2 Compatibility Impact
Fully compatible. No `Option` added, removed, or renamed; no default values changed; no wire/serialization/state format touched. The only observable change is the *distribution* of retry-sleep durations, which was already non-deterministic from the caller's point of view (network timing). `rate_limit_backoff_ms=0` still disables backoff entirely (`return 0L` path untouched) exactly as before.

- The `Thread.sleep(backoffMillis)` call was already blocking the reader/writer's own thread before this PR; jitter doesn't change that architecture, but per point (1) in 1.1 it can materially increase the worst-case blocking duration per retry (up to ~2x) before the cap kicks in. Since Zeta's default checkpoint timeout (`seatunnel.engine.checkpoint.interval.timeout`, `ServerConfigOptions.CHECKPOINT_TIMEOUT`) defaults to 30000ms, a source/sink thread already blocked in a 30s+ sleep during a 429 storm was already a pre-existing risk of colliding with checkpoint alignment before this change; this PR is not the origin of that risk, but it does make the worst case larger for early retries, and that's not called out anywhere in the PR description or docs. Recommend the author explicitly mention the new worst-case bound (up to 2x the base for non-capped retries) in the PR description and, ideally, the connector docs, so operators tuning `rate_limit_backoff_ms`/checkpoint timeout together aren't surprised.
- No new allocations in a hot path beyond a single `ThreadLocalRandom.nextLong` call per retry — negligible.

No error handling paths were changed. The existing `log.warn("Airtable API rate limit reached, retry {}/{} after {} ms", ...)` call already logs the actual jittered value, so the new randomised duration is fully observable in logs — good.

**Issue 1** — Location: `seatunnel-connectors-v2/connector-http/connector-http-airtable/.../source/AirtableSourceReader.java:145-148` and the identical block in `.../sink/AirtableSinkWriter.java:211-214`.
Problem: When `rate_limit_backoff_ms >= MAX_BACKOFF_MILLIS` (300000ms), `floor = Math.max(waitMillis/2, rateLimitBackoffMs) >= waitMillis`, so `calculateBackoffMillis` returns exactly `MAX_BACKOFF_MILLIS` on every retry with **zero jitter**.
Potential risk: For exactly the operators who configure the largest, most conservative backoff (i.e. those most likely to be dealing with sustained rate limiting), the anti-thundering-herd fix silently does nothing — parallel readers/writers go right back to lockstep retries at the cap.
Best improvement: Option A — clamp the effective base used for the floor calculation so it can never exceed `MAX_BACKOFF_MILLIS`, e.g. `long floor = Math.max(waitMillis / 2, Math.min(rateLimitBackoffMs, waitMillis / 2 + 1))`-style adjustment that guarantees `floor < waitMillis` whenever `rateLimitBackoffMs <= MAX_BACKOFF_MILLIS`. Option B — explicitly document/validate that `rate_limit_backoff_ms` must stay below `MAX_BACKOFF_MILLIS` for jitter to take effect, and add a unit test asserting the current (no-jitter) behavior is intentional at that boundary, so it isn't mistaken for a latent bug in a future change.
Severity: Medium.

**Issue 2** — Location: PR description (not a source file) plus the doc comment at `AirtableSourceReader.java:124-132` / `AirtableSinkWriter.java:190-198`.
Problem: The PR description states the result "never exceeds the previous value... so no caller waits longer than it does today." That was true of an earlier commit's "equal jitter" design (`half + random(0, wait-half)`, range `[wait/2, wait]`) also quoted in the description, but the code actually in this diff adds jitter *on top* of `waitMillis` (range `[wait, min(2*wait, MAX)]`), which **can** wait up to ~2x longer than before on non-capped retries.
Potential risk: Reviewers/operators reading only the PR description (a reasonable thing to do) will believe retry timing is bounded the same as before, when the worst case has grown. This is exactly the kind of drift that matters for anyone reasoning about checkpoint/task-level timeout budgets.
Best improvement: Update the PR description to describe the shipped "jitter added on top" behavior and its actual worst-case bound, rather than the earlier "equal jitter" design that was since replaced.
Severity: Medium.

**Issue 3** — Location: `seatunnel-connectors-v2/connector-http/connector-http-airtable/.../sink/AirtableSinkWriter.java:44-45`.
Problem: A blank line was inserted between `import java.util.Optional;` and the new `import java.util.concurrent.ThreadLocalRandom;`. Per this repo's Spotless `importOrder` config (`org.apache.seatunnel.shade,org.apache.seatunnel,org.apache,org,,javax,java,\#`), all `java.*` imports belong to a single group and should have no blank line between them.
Potential risk: Likely to be auto-reformatted (or flagged) by `./mvnw spotless:apply`/`spotless:check`; cosmetic only, but worth a quick pre-merge fix so CI's format check (once it actually runs — see CI section) is clean on the first try.
Severity: Low.

**Issue 4** — Location: `seatunnel-connectors-v2/connector-http/connector-http-airtable/src/test/java/.../sink/AirtableSinkWriterTest.java` (no changes in this PR).
Problem: `calculateBackoffMillis` is duplicated verbatim in `AirtableSinkWriter`, but the five new jitter tests were only added to `AirtableSourceReaderTest.java`. `AirtableSinkWriterTest.java` has zero coverage of the new jitter behavior (confirmed: no reference to `calculateBackoffMillis` or jitter anywhere in that file).
Potential risk: The two copies can silently diverge in a future change (e.g. someone "fixes" only one copy for Issue 1 above) without a failing test to catch it.
Best improvement: Option A — mirror the same five test cases (`testBackoffIsZeroWhenDisabled`, `testBackoffNeverShorterThanConfigured`, `testBackoffIsNotDeterministic`, `testBackoffRespectsMaximum`, `testBackoffStillVariesAtTheMaximum`, `testBackoffAtTheMaximumNeverDropsBelowConfiguredBase`) in `AirtableSinkWriterTest.java`. Option B (preferred long-term, non-blocking for this PR) — extract `calculateBackoffMillis` into a small shared package-private/`@VisibleForTesting` utility used by both classes, so there is exactly one implementation and one test suite to keep in sync.
Severity: Medium.

**Issue 5** — Location: `seatunnel-connectors-v2/connector-http/connector-http-airtable/src/test/java/.../source/AirtableSourceReaderTest.java:144-145` (`testBackoffIsNotDeterministic`).
Problem: The comment says "The range here is 2000ms wide" for `retry=3, base=1000`. Actual values: `exponential = 1<<2 = 4`, `waitMillis = 4000`, `extra = min(4000, 296000) = 4000`, so the real range is `[4000, 8000]` — 4000ms wide, not 2000ms.
Potential risk: None functionally (the assertion itself, `observed.size() > 1`, is correct and robust regardless of the exact width) — purely a misleading comment for future readers trying to reason about collision probability from the stated numbers.
Severity: Low.

# 2. Code Quality Assessment

## 2.1 Coding Standards
The new logic is unusually well commented for its complexity — both branches of `calculateBackoffMillis` carry multi-line prose comments explaining *why* jitter is added on top vs. spread downward, which is exactly the kind of "why, not what" comment this project's conventions ask for. `@VisibleForTesting` is used correctly to justify widening the method from `private` to package-private. The only standards nit is Issue 3 (stray import blank line).

Four (in the PR description) / six (in the actual diff) new unit tests were added to `AirtableSourceReaderTest.java`, all calling `calculateBackoffMillis` directly as a pure function — no `Thread.sleep`, no wall-clock assertions, no container/network dependency. They use `ThreadLocalRandom` draws over 100–200 iterations per test with wide-enough bounds (hundreds to thousands of possible output values per range) that a collision-based false failure (e.g. in `testBackoffIsNotDeterministic`) is not a realistic flakiness source. The two pre-existing tests that do exercise the real retry loop with `Thread.sleep` (`testRetryOn429ThenSuccess`, `testStopRetryAfterMaxRetries`) are untouched and continue to run with `rate_limit_backoff_ms=0` (via `createReader`'s hardcoded `0` base), so they never actually sleep — jitter is a no-op for them and they were not weakened to "accommodate" it.

**Flaky-test rating: Stable.** No hard waits tied to wall-clock, no shared/static state between tests, no order dependency, all new assertions are on pure-function outputs with generous, well-reasoned bounds.

Coverage gap: see Issue 4 — the identical logic in `AirtableSinkWriter` has no dedicated jitter tests.

## 2.3 Documentation Updates
Not applicable/no update needed for `docs/en` and `docs/zh` `Airtable.md` — `rate_limit_backoff_ms` and `rate_limit_max_retries` option names, types, and default values are unchanged, so the existing option tables remain accurate. A short note that the actual applied delay includes jitter (and can exceed the configured base on early retries) would be a nice-to-have addition to `docs/en/connectors/source/Airtable.md` / `docs/en/connectors/sink/Airtable.md` (and `zh` counterparts) around the `rate_limit_backoff_ms` description, but I would not block merge on this since no user-facing contract (option name/type/default) actually changed.

# 3. Architectural Soundness

Precise fix for the stated problem (thundering herd on shared per-base rate limits), with one real gap at the top of the range (Issue 1) that undermines the third commit's own stated goal in a specific, findable configuration.

## 3.2 Maintainability
The jitter logic (and its extensive explanatory comments) is duplicated byte-for-byte across two files, compounding a duplication that already existed for the rest of `calculateBackoffMillis` before this PR. This isn't something this PR is obligated to fix, but it's the direct cause of Issue 4 (asymmetric test coverage) and is worth a follow-up to extract into a shared helper.

## 3.3 Extensibility
Jitter bounds are derived algorithmically from `waitMillis`/`MAX_BACKOFF_MILLIS`/`rateLimitBackoffMs` rather than being a new hard-coded magic number, which is reasonable — but there is no way for an operator to tune or disable jitter independently of disabling backoff entirely (`rate_limit_backoff_ms=0`). That's an acceptable trade-off for a first pass and matches "no configuration change" from the PR description; flagging only as a non-blocking observation.

No serialization, checkpoint state, or protocol surface is touched by this change; nothing to migrate. Fully compatible with any previously-persisted job/checkpoint state.

# 4. Issue Summary

|---|---|---|---|
| 1 | Zero-jitter dead zone when `rate_limit_backoff_ms >= MAX_BACKOFF_MILLIS` (300000ms) — defeats the "spread at the maximum" fix exactly where it matters most | `AirtableSourceReader.java:145-148`, `AirtableSinkWriter.java:211-214` | Medium |
| 2 | PR description's latency guarantee ("no caller waits longer than it does today") does not match the shipped "jitter on top" implementation, which can wait up to ~2x longer on non-capped retries | PR description; `AirtableSourceReader.java:124-132`, `AirtableSinkWriter.java:190-198` | Medium |
| 4 | `AirtableSinkWriter`'s identical `calculateBackoffMillis` copy has zero jitter test coverage | `AirtableSinkWriterTest.java` (missing) | Medium |
| 3 | Stray blank line inside the `java.*` import group breaks Spotless import-order convention | `AirtableSinkWriter.java:44-45` | Low |
| 5 | Test comment states an incorrect range width ("2000ms" instead of actual 4000ms) | `AirtableSourceReaderTest.java:144-145` | Low |

# CI Status

The apache-side "Build" check on this PR currently reports `action_required` with the message "Workflow run detection failed / Unable to detect the workflow run for testing the changes in your PR", which points to GitHub Actions not having run on the contributor's fork for this branch (the fork shows zero total workflow runs). This is not a code-quality failure I can attribute a root cause to from the source — it's a CI-plumbing gap. Per this project's usual pattern, the fix is on the contributor's side: enable GitHub Actions on your fork (`https://github.com/1991santhu/seatunnel` → Settings → Actions), then push an empty commit (or rebase onto the latest `dev`) to trigger a run. Check-run for reference: https://github.com/apache/seatunnel/runs/92112091125. This should be treated as a distinct blocker independent of the source-level issues above — the PR cannot be verified by CI at all in its current state.

# 5. Merge Recommendation

1. Blockers — must be fixed:
   - CI: enable GitHub Actions on the fork and get a real "Build" run to complete (see CI Status above) — currently there is no CI signal at all for this PR.
   - Issue 1 (Medium): close the zero-jitter dead zone at/above `MAX_BACKOFF_MILLIS`, or explicitly document and test it as intentional so it isn't mistaken for a bug later.
   - Issue 2 (Medium): correct the PR description so its stated latency guarantee matches the actual shipped behavior (jitter added on top, up to ~2x on non-capped retries).
   - Issue 4 (Medium): add the missing jitter test coverage for `AirtableSinkWriter.calculateBackoffMillis` (mirroring the existing `AirtableSourceReaderTest` cases is sufficient).

   - Issue 3 (Low): remove the stray import blank line; run `./mvnw spotless:apply`.
   - Issue 5 (Low): fix the "2000ms" comment to match the actual 4000ms range.
   - Consider extracting the duplicated `calculateBackoffMillis` (and its jitter logic) into one shared helper used by both `AirtableSourceReader` and `AirtableSinkWriter`, so future fixes (like Issue 1) only need to happen once.
   - Consider a one-line docs note about jitter in `docs/en`/`docs/zh` `Airtable.md` near `rate_limit_backoff_ms`.

Overall assessment: this is a well-reasoned, well-tested improvement to a real problem (shared-base rate-limit lockstep retries), and the iteration history on this PR (three commits progressively refining the approach) shows the author is actively thinking through edge cases rather than shipping the first idea that compiled. The remaining gap is that the very last refinement (spreading jitter once the wait hits the cap) has a boundary condition — a sufficiently large configured base — where it silently reverts to the old lockstep behavior, and the PR's own description no longer accurately describes what the code does. Neither issue requires a redesign; both are small, targeted fixes. Option A (fix Issue 1 in place, minimal diff) is preferable to Option B (redesigning the whole jitter formula) given how close the existing tests already come to fully specifying the intended behavior — most of the new tests would not need to change.

Note: `./mvnw spotless:apply` / a full local build/test run was not executed as part of this review (per this review's process, code-level analysis only); Issue 3 is based on manual comparison against the repository's declared Spotless `importOrder` configuration, not a build run.

davidzollo:
Good job.

To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration https://github.com/apache/seatunnel/pull/11654/checks?check_run_id=92112091125

image

davidzollo:

```

```

1991santhu:
Thanks, this is a genuinely good catch and all five are addressed.

**Issue 1, the dead zone.** You're right, and it's the worst case to get wrong. With `rate_limit_backoff_ms >= MAX_BACKOFF_MILLIS`, `waitMillis` is pinned to the cap from the first retry, `floor` becomes `>= waitMillis`, and the method returned the cap with no jitter. So the operators running the most conservative backoff, the ones most likely to be sitting in sustained rate limiting, got the old lockstep behaviour back.

Took your Option A. The floor is half the wait, raised to `rate_limit_backoff_ms` only when that is below the wait:

```java
long floor = waitMillis / 2;
if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis) {
    floor = rateLimitBackoffMs;
}
```

Once the cap has pushed the wait under the configured backoff, that guarantee cannot be honoured anyway, so flooring there only removed the jitter.

There's a regression test for it. I checked it fails on the old code first, with `observed only [300000]`.

**Issue 2, the description.** Also right, and I should have caught that when the approach changed. The description still described the equal-jitter version that got replaced. Rewritten to state the shipped behaviour and the actual worst case: with the defaults, 210000ms across three retries becomes 420000ms until the cap absorbs it. I've called that out explicitly for anyone tuning against checkpoint timeouts.

**Issue 4, sink coverage.** Mirrored all seven jitter tests into `AirtableSinkWriterTest`, which had none despite carrying a byte-for-byte copy of the method. Agreed that extracting the duplicate into one shared helper is the better long-term answer; I've left that out of this PR to keep the diff focused, happy to follow up.

**Issues 3 and 5.** Blank line removed from the `java.*` import group and `spotless:apply` run. Comment corrected, the range at retry 3 with a 1000ms base is `[4000, 8000]`, not 2000ms wide.

18 tests pass across both classes on JDK 11.

On CI: Actions are not enabled on my fork, which is why there's no Build signal. I'll get that sorted so there's a real run against this branch.

DanielLeens (commented):
Thanks for turning this around so quickly, and for the thoughtful, specific replies to every point — I appreciate you re-running the numbers rather than just asserting the fix works. I went through the full diff again from scratch (not just the new commit), retraced both call chains end to end, and re-verified each of the five issues from my last review against the current head. All five are genuinely resolved; details and one additional finding below.

Sorry for missing this in the last review — that one is on me. It isn't something the new commit introduced; it was already present in the "spread the backoff once it reaches the maximum" commit I reviewed last time, and I should have traced the boundary between the last uncapped retry and the first capped retry. See Issue 1 below, marked as a carryover, not a regression.

Airtable rate-limits per base, not per client, so before this PR every reader/writer that got 429'd together also retried at the exact same instants (`calculateBackoffMillis` was a pure function of `retryCount`), and the burst that caused the 429 reformed on every retry. This PR adds randomised jitter to `calculateBackoffMillis()` in both `AirtableSourceReader` and `AirtableSinkWriter` so concurrent retries spread out, while keeping the result within `[configured floor, MAX_BACKOFF_MILLIS]` in both the normal (below-cap) and at-cap cases.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The change is confined to `calculateBackoffMillis(int retryCount)`, duplicated identically in `AirtableSourceReader.java:116-156` and `AirtableSinkWriter.java:180-220`, called only from each class's 429-retry loop (`AirtableSourceReader.java:63-87`, `AirtableSinkWriter.java:125-161`).

Current implementation:
```java
long exponential = 1L << Math.min(20, Math.max(0, retryCount - 1));
long waitMillis = Math.min(rateLimitBackoffMs * exponential, MAX_BACKOFF_MILLIS);

long extra = Math.min(waitMillis, MAX_BACKOFF_MILLIS - waitMillis);
if (extra > 0) {
    return waitMillis + ThreadLocalRandom.current().nextLong(extra + 1);
}

long floor = waitMillis / 2;
if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis) {
    floor = rateLimitBackoffMs;
}
return waitMillis - ThreadLocalRandom.current().nextLong(waitMillis - floor + 1);
```

Runtime path on a 429 (identical in both classes):
```
executeRequest()/sendWithRateLimitRetry()
  -> response.getCode() == 429 && retryCount < rateLimitMaxRetries
  -> retryCount++
  -> calculateBackoffMillis(retryCount)
       -> waitMillis < MAX_BACKOFF_MILLIS => range [waitMillis, min(2*waitMillis, MAX)]   (jitter added on top)
       -> waitMillis == MAX_BACKOFF_MILLIS => range [floor, MAX], floor = waitMillis/2, raised to
          rateLimitBackoffMs only when strictly between waitMillis/2 and waitMillis
  -> log.warn(... backoffMillis) -> Thread.sleep(backoffMillis) -> continue loop
```

This only fires on the 429/error-retry path; the normal happy path (200 response) never touches `calculateBackoffMillis`. With the connector defaults (`rate_limit_backoff_ms=30000`, `rate_limit_max_retries=3`), retries never reach `MAX_BACKOFF_MILLIS` (retry 3 raw wait is 120000ms, well under the 300000ms cap), so the "at-cap" branch and everything discussed below about it is a boundary/config-dependent path, not the default path — worth stating explicitly since it changes how urgently it needs to be fixed.

Key findings:
- Both previously-flagged correctness gaps are fixed. I re-derived the floor arithmetic by hand for the cases that broke it last time and it now holds (see Issues 3/4 verification below, marked resolved).
- The PR description now matches the shipped implementation (jitter added on top, not centred), including the corrected worst-case numbers.
- Test coverage is now symmetric between source and sink (7 mirrored jitter tests per class), including dedicated regression tests for the exact boundary conditions raised previously.
- New finding (Issue 1, carryover): the guaranteed lower bound of the jittered range is not monotonically non-decreasing across the transition from the last uncapped retry into the first capped retry, when `rate_limit_backoff_ms < MAX_BACKOFF_MILLIS / 2`. This means a later retry's *worst case minimum* can be lower than an earlier retry's, in a config-dependent band. It does not break the two hard guarantees (never below `rate_limit_backoff_ms`, never above `MAX_BACKOFF_MILLIS`), but it is a real, reproducible characteristic worth a conscious decision rather than an accidental one.

Concrete trace for Issue 1, with `rate_limit_backoff_ms=100000`, `rate_limit_max_retries=5` (a legitimate, non-default but supported configuration):
- retry 2: `exponential=2`, raw `=200000`, `waitMillis=200000` (uncapped, `extra=min(200000,100000)=100000`) → range `[200000, 300000]`.
- retry 3: `exponential=4`, raw `=400000`, `waitMillis=min(400000,300000)=300000` (capped) → `extra=0` → floor branch: `floor=150000`; `100000 > 150000` is false, so floor stays `150000` → range `[150000, 300000]`.

So retry 3's minimum possible wait (150000ms) is lower than retry 2's minimum possible wait (200000ms), even though retry 3 is a later, "should be longer" retry. It's possible in practice for a drawn retry-3 value to be smaller than a drawn retry-2 value for the same reader/writer instance. This falls out of a general property: once the last uncapped retry `n` has `waitMillis_n = rate_limit_backoff_ms * 2^(n-1) ∈ [150000, 300000)` (true by definition of "last uncapped retry"), its floor is `waitMillis_n` itself (the on-top branch's minimum), while the first capped retry `n+1`'s floor is `max(150000, rate_limit_backoff_ms)`, which is `≤ waitMillis_n` whenever `n > 1`. So the dip is generic to any config that takes more than one doubling to reach the cap with a moderate `rate_limit_backoff_ms`, not a contrived single data point. This characteristic already existed in the "spread the backoff once it reaches the maximum" commit that I reviewed previously and is unchanged by the latest "close the zero-jitter dead zone" commit — the latest commit only changed when the floor gets raised to `rate_limit_backoff_ms`, not the `waitMillis/2` baseline that causes this dip. That's why I'm marking it a carryover I should have caught the first time, not something newly introduced.

## 1.2 Compatibility Impact
Fully compatible. No `Option` added, removed, renamed, or defaulted differently; no wire/serialization/checkpoint-state format touched. `rate_limit_backoff_ms=0` still disables backoff entirely (`AirtableSourceReader.java:118-119` / `AirtableSinkWriter.java:182-183`, unchanged). The only observable change is the distribution (and, per Issue 1's PR-description note, the worst-case bound) of retry-sleep durations.

`Thread.sleep(backoffMillis)` blocks the reader/writer's own thread exactly as it did before; jitter doesn't change that architecture. The PR description now correctly discloses the worst-case increase (210000ms → 420000ms across 3 default retries) so operators tuning `rate_limit_backoff_ms` against checkpoint timeouts have accurate numbers to work from — that closes the gap I raised last time. No new allocations beyond a single `ThreadLocalRandom.nextLong` per retry; negligible. `ThreadLocalRandom.current()` is the correct, contention-free choice for concurrent reader/writer instances; `calculateBackoffMillis` remains a stateless pure function of its input plus `ThreadLocalRandom`, so there's no new shared-state race.

Unchanged and correct: `log.warn("Airtable API rate limit reached, retry {}/{} after {} ms", ...)` at `AirtableSourceReader.java:72-76` / `AirtableSinkWriter.java:138-142` logs the actual post-jitter value, so the randomised duration stays fully observable. No new exception paths, no swallowed exceptions, nothing sensitive logged.

**Issue 1** — Location: `AirtableSourceReader.java:151-155`, `AirtableSinkWriter.java:215-219`.
Problem: At the transition from the last uncapped retry to the first capped retry, the guaranteed minimum of the jittered range can decrease (worked example above: retry 2 floor 200000ms vs. retry 3 floor 150000ms with `rate_limit_backoff_ms=100000`). This happens whenever it takes more than one doubling to reach `MAX_BACKOFF_MILLIS` with a moderate base.
Potential risk: Low practical risk — both hard guarantees (`>= rate_limit_backoff_ms`, `<= MAX_BACKOFF_MILLIS`) still hold, and this is a boundary/non-default-config path. But a later retry can, by chance, sleep for less than an earlier retry did on the same instance, which is counter to the usual mental model of "exponential backoff never shrinks," and is easy to trip over again in a future change to this method (as almost happened with Issue 1 from my last review).
Best improvement: Option A — carry the previous retry's realized (or minimum-guaranteed) wait forward as an additional floor input once capped, e.g. floor at cap = `max(waitMillis/2, min(previousWaitMillis, waitMillis))`, so the range never dips under what the prior retry could have produced. Option B (lighter touch) — leave the behavior as is but add one sentence to the method's doc comment stating explicitly that the minimum is not guaranteed monotonic across the cap boundary, so it reads as a documented trade-off instead of an oversight the next time someone touches this method.
Severity: Medium.
Raised by another reviewer: No.

**Issue 2** — Location: N/A (verification of previously-raised items, not a new issue).
Verified resolved: my four other issues from the last review are all fixed in the current head —
- Dead zone at `rate_limit_backoff_ms >= MAX_BACKOFF_MILLIS`: `AirtableSourceReader.java:152` / `AirtableSinkWriter.java:216` now reads `if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis)`, and the new `testBackoffStillVariesWhenBaseExceedsTheMaximum` test (both `*Test.java` files) is a real regression test for exactly this — I hand-traced it (`base=300000` → `floor` stays at `150000` since `300000 < 300000` is false → range `[150000, 300000]`, no longer a fixed point).
- PR description accuracy: now states the shipped "jitter on top" behavior and the correct worst-case numbers (210000ms → 420000ms).
- Stray import blank line in `AirtableSinkWriter.java`: gone — `java.util.Optional` and `java.util.concurrent.ThreadLocalRandom` are now adjacent with no blank line (`AirtableSinkWriter.java:43-44`).
- Sink test coverage: `AirtableSinkWriterTest.java` now carries all 7 jitter tests, mirroring `AirtableSourceReaderTest.java` test-for-test, including a comment noting the duplication is intentional so the two copies don't drift apart unnoticed.
- Test comment accuracy: `testBackoffIsNotDeterministic` now correctly states the range as `[4000, 8000]` in both files.
Severity: N/A (resolution confirmation).
Raised by another reviewer: No (self-verification of my own prior review).

# 2. Code Quality Assessment

## 2.1 Coding Standards
Comment quality is genuinely good — both branches of `calculateBackoffMillis` explain *why* jitter is added on top vs. spread downward, and why the floor is or isn't raised at the boundary, which is exactly the "why, not what" standard this project asks for. `@VisibleForTesting` is used correctly to justify widening the method's visibility. No missing comments on the touched non-trivial logic.

14 new unit tests total (7 per class), all calling `calculateBackoffMillis` directly as a pure function — no `Thread.sleep`, no wall-clock assertions, no container/network dependency, no shared/static state, no order dependency. Iteration counts (100–200 draws) against wide ranges (thousands of possible long values) make a collision-based false failure statistically negligible. The pre-existing `Thread.sleep`-based retry-loop tests (`testRetryOn429ThenSuccess`, `testStopRetryAfterMaxRetries`) are untouched and still construct with `rate_limit_backoff_ms=0`, so jitter is a no-op for them — they weren't weakened to "accommodate" the change.

**Stability rating: Stable.**

The two new regression tests (`testBackoffStillVariesWhenBaseExceedsTheMaximum`, `testBackoffAtTheMaximumNeverDropsBelowConfiguredBase`) are exactly targeted at the two boundary conditions from my last review and I independently re-derived their expected ranges by hand — both check out against the current implementation.

## 2.3 Documentation Updates
No changes needed to `docs/en`/`docs/zh` `Airtable.md` — `rate_limit_backoff_ms` and `rate_limit_max_retries` names/types/defaults are unchanged (verified against `docs/en/connectors/source/Airtable.md:44-45,134-140` and the sink equivalent), and the docs never documented the exponential/jitter formula in the first place, so nothing there is now stale. A short note about jitter affecting the actual applied delay would still be a nice-to-have, not a blocker.

# 3. Architectural Soundness

Precise, targeted fix for the stated thundering-herd problem, now with the cap-boundary dead zone closed. Issue 1 above is a genuine remaining rough edge, but it's a refinement of an already-working solution, not evidence the approach is wrong.

## 3.2 Maintainability
The full method (jitter logic and its comments) remains duplicated byte-for-byte across two files. That's the direct reason Issue 1 in my last review (sink had no test coverage) could happen in the first place, and it's exactly the kind of place where Issue 1 in *this* review (the monotonicity dip) could get "fixed" in one copy and not the other in a future change. Non-blocking for this PR, but I'd genuinely like to see the follow-up to extract a shared helper that the author already agreed to.

## 3.3 Extensibility
Jitter bounds are derived from `waitMillis`/`MAX_BACKOFF_MILLIS`/`rate_limit_backoff_ms` rather than new magic numbers. No way to tune/disable jitter independently of disabling backoff entirely, which is a reasonable trade-off matching "no configuration change."

No serialization, checkpoint state, or protocol surface touched. Nothing to migrate.

# 4. Issue Summary

|---|---|---|---|
| 1 | Guaranteed minimum wait can decrease from the last uncapped retry to the first capped retry (carryover from the version I reviewed previously, not introduced by the latest commit) | `AirtableSourceReader.java:151-155`, `AirtableSinkWriter.java:215-219` | Medium |

All five issues from my previous review (dead zone at the cap, PR-description mismatch, stray import blank line, missing sink test coverage, incorrect test comment) are confirmed fixed in the current head — see the verification notes under Issue 2 above.

# 5. Merge Recommendation

1. Blockers — must be fixed:
   - Issue 1 (Medium): close the non-monotonic dip at the uncapped-to-capped transition, or explicitly document/test it as an accepted trade-off (Option A or B in Issue 1 above), the same standard I held the previous dead-zone gap to. It doesn't violate either hard guarantee (floor at `rate_limit_backoff_ms`, ceiling at `MAX_BACKOFF_MILLIS`) and it's a boundary/non-default-config path, but it's a small, well-scoped fix and I'd rather it land deliberately than get rediscovered later.

   Everything else is clear: CI is green at the current head, and all five issues from my previous review (dead zone, PR-description mismatch, stray import blank line, missing sink coverage, wrong test comment) are confirmed resolved — no blockers remain from that list.

   - Follow-up (already agreed by the author) to extract `calculateBackoffMillis` into one shared helper used by both classes, so a fix like Issue 1 only needs to land once.
   - Optional one-line docs note in `docs/en`/`docs/zh` `Airtable.md` mentioning that the applied delay includes jitter.

Regarding my own previous CHANGES_REQUESTED verdict: I'm updating it based on the evidence above — four of five issues are cleanly resolved with real regression tests I independently re-derived, CI is green, and the fifth (PR description accuracy) is also fixed. The one remaining item (Issue 1) is narrow and boundary-only, so this is close, but by the same bar I applied last time I'd like it landed before merge rather than after. Overall this is careful, well-tested work, and I appreciate the iteration — the test suite in particular gives me real confidence rather than just taking the description's word for it.

1991santhu:
Good catch, and you are right that it predates the last commit.

Fixed by flooring at the last scheduled wait that still fitted under the cap, rather than half the cap. With a 100000ms base that is 200000, so the capped retries now draw [200000, 300000] instead of [150000, 300000] and the minimum stops dipping.

It also made the method simpler. The rate_limit_backoff_ms special case is gone, since the last uncapped wait is never smaller than the configured backoff, so that guarantee falls out for free now.

I went with option A rather than documenting it. The previous scheduled wait is derivable from the retry count, so it stays a pure function with no state to carry.

Test added in both classes. I checked it fails on the old code first, where retry 3 drew 193352 against retry 2's guaranteed 200000.

20 tests pass on JDK 11.

DanielLeens (approved):

Airtable rate-limits per base, not per client, so before this PR every reader/writer that got 429'd together also retried at the exact same instants (`calculateBackoffMillis` was a pure function of `retryCount`), and the burst that caused the 429 reformed on every retry. This PR adds randomised jitter to `calculateBackoffMillis()` in both `AirtableSourceReader` and `AirtableSinkWriter` so concurrent retries spread out, while keeping the result within `[configured floor, MAX_BACKOFF_MILLIS]`.

This is a re-review of the latest commit only (`keep the backoff minimum from dipping at the cap`), which is a direct fix for the one issue I left open in my previous review.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The only change since my last review is in the "spread downward at the cap" branch of `calculateBackoffMillis(int retryCount)`, identical in `AirtableSourceReader.java:116-159` and `AirtableSinkWriter.java:180-223`:

Before (what I reviewed last time):
```java
long floor = waitMillis / 2;
if (rateLimitBackoffMs > floor && rateLimitBackoffMs < waitMillis) {
    floor = rateLimitBackoffMs;
}
```

After (current head):
```java
long floor = waitMillis / 2;
for (long scheduled = rateLimitBackoffMs; scheduled < MAX_BACKOFF_MILLIS; scheduled <<= 1) {
    if (scheduled > floor) {
        floor = scheduled;
    }
}
```

My last review's Issue 1 was: once the retry schedule crosses from the last uncapped retry into the first capped retry, the guaranteed minimum could drop. Worked example I traced then, `rate_limit_backoff_ms=100000`, `rate_limit_max_retries=5`: retry 2 was uncapped with a floor of 200000 (its own `waitMillis`), but retry 3 was capped with the old floor logic giving 150000 — a later retry could sleep for less than an earlier one.

I re-derived the fix by hand against that same example. The loop walks the doubling schedule `rateLimitBackoffMs, rateLimitBackoffMs*2, rateLimitBackoffMs*4, ...` and keeps the largest term that is still `< MAX_BACKOFF_MILLIS`, i.e. the last wait that was still uncapped:
- `scheduled = 100000` (`< 300000`, `> floor(150000)`? no) → floor stays 150000
- `scheduled <<= 1 = 200000` (`< 300000`, `> 150000`? yes) → floor = 200000
- `scheduled <<= 1 = 400000` → loop condition fails, stop

Result: floor = 200000, so retry 3 (and every later capped retry, since the loop's output doesn't depend on `retryCount`) now draws from `[200000, 300000]`, matching retry 2's own minimum exactly. I checked several other bases by hand (e.g. `rate_limit_backoff_ms=90000` → schedule 90000, 180000, capped at 360000 → floor 180000, which matches retry 2's uncapped minimum of 180000) and the property holds generally: the floor is always exactly the last pre-cap `waitMillis`, so the minimum sequence across all retries is non-decreasing. Issue 1 is resolved.

I also re-checked this doesn't reopen the *previous* round's dead-zone issue (`rate_limit_backoff_ms >= MAX_BACKOFF_MILLIS`, e.g. 300000): the loop condition `scheduled < MAX_BACKOFF_MILLIS` is false on the very first iteration (`300000 < 300000` is false), so the loop body never runs, floor stays at `waitMillis/2 = 150000`, and jitter is still present — matches the existing `testBackoffStillVariesWhenBaseExceedsTheMaximum` regression test, which I re-verified still passes against the current logic by hand-tracing it.

One thing I specifically checked given the loop is new: whether `rateLimitBackoffMs = 0` could make `for (long scheduled = 0; scheduled < MAX_BACKOFF_MILLIS; scheduled <<= 1)` infinite-loop (`0 << 1` is still `0`). It can't — `calculateBackoffMillis` returns `0L` unconditionally at the top of the method whenever `rateLimitBackoffMs <= 0` (`AirtableSourceReader.java:118-120` / `AirtableSinkWriter.java:182-184`), so the loop only ever runs with `rateLimitBackoffMs >= 1`, and the loop condition converges within about 19 iterations for any realistic base up to `Integer.MAX_VALUE`. I also confirmed the floor produced by the loop is always strictly less than `waitMillis` in this branch (since the loop only updates `floor` from values that are themselves `< MAX_BACKOFF_MILLIS = waitMillis` here), so `waitMillis - floor + 1` passed to `ThreadLocalRandom.nextLong(bound)` is always positive — no `IllegalArgumentException` risk at the boundary.

Runtime path (unchanged from before, still only reachable from the 429-retry loop, never the happy path):
```
executeRequest()/sendWithRateLimitRetry()
  -> response.getCode() == 429 && retryCount < rateLimitMaxRetries
  -> calculateBackoffMillis(retryCount)
       -> waitMillis < MAX_BACKOFF_MILLIS => range [waitMillis, min(2*waitMillis, MAX)]      (unchanged)
       -> waitMillis == MAX_BACKOFF_MILLIS => range [floor, MAX], floor = last pre-cap wait   (this commit)
  -> Thread.sleep(backoffMillis) -> continue loop
```
With the connector's defaults (`rate_limit_backoff_ms=30000`, `rate_limit_max_retries=3`), retries never reach `MAX_BACKOFF_MILLIS` (retry 3's raw wait is 120000ms), so this whole branch — and therefore this entire commit — is inert for the default/normal path. It only activates for a non-default configuration where `rate_limit_max_retries` is high enough for the schedule to cross the 300000ms cap. That matches what the author's commit message states (reachable with e.g. `rate_limit_backoff_ms=100000ms` and `rate_limit_max_retries>=3`) and is consistent with what I found in the previous round.

## 1.2 Compatibility Impact
Fully compatible. No `Option` touched (confirmed via `git diff --stat` against the merge-base — only the two `calculateBackoffMillis` methods and their two test files changed in this whole PR). `rate_limit_backoff_ms` and `rate_limit_max_retries` keep their existing names, types, and defaults (`30000` / `3` in `AirtableConfig.java:73-85`). Users who never hit a 429 never call `calculateBackoffMillis` at all (it's only invoked from the retry branch), so their behavior is byte-for-byte unchanged by this entire PR. Users on the default config who do hit a 429 are unaffected by this specific commit, since the default schedule never reaches the capped branch this commit touches.

The loop runs at most ~19 iterations (bounded by `log2(MAX_BACKOFF_MILLIS / rateLimitBackoffMs)`) and only executes once per retry, on the already-rare 429 path — no measurable cost. No new shared state, no new allocations beyond the loop's primitive `long`. `calculateBackoffMillis` remains a stateless pure function of `retryCount` plus the immutable config fields plus `ThreadLocalRandom`, so there's no new concurrency concern versus what I already analyzed in the previous rounds.

Unchanged. `log.warn("Airtable API rate limit reached, retry {}/{} after {} ms", ...)` still logs the actual post-jitter value.

No new issues found in this commit. Everything raised in my two previous reviews is now resolved — see the verification notes above and in Section 4.

# 2. Code Quality Assessment

## 2.1 Coding Standards
The comment above the floor calculation was rewritten to describe the new logic accurately (explains *why* the loop walks the schedule rather than just raising to `rateLimitBackoffMs`, and calls out the monotonicity property explicitly) — good "why, not what" comment consistent with this project's standard. No missing documentation on the touched code.

One new test per class, mirrored identically: `testBackoffMinimumNeverDropsAtTheCapBoundary` (`AirtableSourceReaderTest.java:195-216`, `AirtableSinkWriterTest.java`). It uses `rate_limit_backoff_ms=100000` and asserts every draw for retries 3 through 8 falls in `[200000, 300000]` — i.e. it directly targets the exact scenario from my last review (the 100000/5 example) and would fail against the pre-fix code, since half the cap (150000) would previously have been reachable. I re-derived the expected range by hand (see 1.1) and it matches. 200 draws per retry count against a 100000-wide range makes a collision-based false failure statistically negligible. No `Thread.sleep`, no wall-clock assertions, no shared/static state, no ordering dependency versus the other 9 tests in each file.

**Stability rating: Stable.**

Total test count is now 10 per class / 20 across both — matches the author's stated count. I confirmed the two test files stay mirrored test-for-test (same 7 jitter-specific tests plus this new one, in both), so the duplication risk I flagged for maintainability (not correctness) in round 1 hasn't regressed.

## 2.3 Documentation Updates
No change needed; unchanged from my last review — `rate_limit_backoff_ms`/`rate_limit_max_retries` option names, types, and defaults in `docs/en`/`docs/zh` `Airtable.md` remain accurate.

# 3. Architectural Soundness

This is exactly the fix I suggested as "Option A" in my last review (derive the floor from the retry schedule rather than carrying state), implemented as a pure function with no added state — clean, and it also slightly simplifies the mental model versus a manually-tracked "previous wait" would have.

## 3.2 Maintainability
The method is still duplicated byte-for-byte across `AirtableSourceReader` and `AirtableSinkWriter` (confirmed identical again on this round). That's pre-existing and not something this PR needs to fix, but I'd still like to see the shared-helper extraction the author already agreed to as a follow-up, since it's exactly this kind of subtle boundary-condition fix that's easy to apply to one copy and forget on the other.

## 3.3 Extensibility
Unchanged from previous rounds — no way to tune/disable jitter independently of disabling backoff entirely, which remains an acceptable, deliberate trade-off matching "no configuration change."

No serialization, checkpoint state, or protocol surface touched. Nothing to migrate.

# 4. Issue Summary

| No. | Issue | Location | Severity | Status |
|---|---|---|---|---|
| — | Guaranteed minimum could decrease from the last uncapped retry to the first capped retry | `AirtableSourceReader.java:151-158`, `AirtableSinkWriter.java:215-222` | Medium | **Resolved** — re-verified by hand-tracing the fix against the exact example from my last review, plus the new regression test |
| — | Zero-jitter dead zone when `rate_limit_backoff_ms >= MAX_BACKOFF_MILLIS` | (round 1) | Medium | Resolved (round 2), re-confirmed still holds on this head |
| — | PR description didn't match shipped "jitter on top" behavior | (round 1) | Medium | Resolved (round 2) |
| — | Missing sink-side jitter test coverage | (round 1) | Medium | Resolved (round 2), re-confirmed both files still mirrored |
| — | Stray import blank line | (round 1) | Low | Resolved (round 2) |
| — | Incorrect test comment ("2000ms" vs actual 4000ms) | (round 1) | Low | Resolved (round 2) |

No new issues found in this round. All issues raised across both of my previous reviews are confirmed resolved on the current head (`4ecd2b142869db1c39aadf93dc76195efd5f9bc6`).

# CI Status

At review time, the fork's CI run for this exact head commit is in progress: License header, Code style, Dead links, Helm chart syntax, and all `updated-modules-integration-test` matrix jobs that have finished are green; `unit-test` (JDK 8/11, Ubuntu/Windows) and two `updated-modules-integration-test-part-1` jobs are still running. Nothing has failed so far. This isn't a blocker on the code review side — just noting it hasn't fully completed yet, so the merge should wait for it to go green rather than being gated on anything in this review.

# 5. Merge Recommendation

1. Blockers — none remaining. The one Medium-severity issue outstanding from my last review (non-monotonic minimum at the cap boundary) is fixed and verified against the exact case that exposed it, with a targeted regression test.
2. Recommended fixes — non-blocking, unchanged from before:
   - Extract the duplicated `calculateBackoffMillis` into a single shared helper used by both `AirtableSourceReader` and `AirtableSinkWriter` (already agreed as a follow-up).
   - Optional one-line docs note in `docs/en`/`docs/zh` `Airtable.md` mentioning that the applied delay includes jitter.

Overall assessment: this has been a genuinely well-iterated PR — three rounds of review turned up real, non-obvious boundary conditions in the backoff math (the dead zone at the cap, then the monotonicity dip at the cap boundary), and each was fixed precisely and re-tested against the exact case that exposed it rather than papered over. I'm satisfied the core jitter/backoff logic is now correct: bounded on both ends (`[rate_limit_backoff_ms or the last pre-cap wait, MAX_BACKOFF_MILLIS]`), no thundering-herd regression at any point in the schedule including the cap boundary, no infinite-retry risk (retry count is still bounded by the pre-existing `rate_limit_max_retries`), fully backward compatible for anyone not hitting the rate limit, and well covered by fast, deterministic unit tests on both sides of the duplication. Pending CI finishing green, this is ready to merge.

davidzollo (approved):
+1
LGTM

```

```

PR #11553, opened 2026-07-25
[Feature][CLI] Add accuracy benchmark for the AI CLI (100 tasks, layered verdict gates, repair-loop measurement)

SEZ9:
### Purpose of this pull request

Implements #11549 — adds a standalone accuracy benchmark for the AI CLI under `seatunnel-cli/benchmark/`. All new files; no product code changes.

The benchmark answers a question the project previously could not answer quantitatively: **when the CLI's prompts, connector metadata, or repair logic change, does config-generation accuracy actually improve — and by how much, for which models?**

**What's included** (19 files, ~4,000 lines):

| Component | Files | Role |
|-----------|-------|------|
| Task suite | `tasks/tier{1,2,3}_*.json` | 100 declarative tasks: 20 simple / 45 medium / 35 complex-DAG, 12 ETL categories, 10 Chinese prompts, 18 rule probes targeting known LLM failure modes |
| L1 static gate | `scoring.py` | pyhocon parse + connector metadata + per-task assertions (offline) |
| L2/L3 gates | `execution.py` | engine `--dry-run static`; real job execution on the official `apache/seatunnel` image (batch: exit code; streaming: 60s health) |
| Data environment | `docker/` | docker-compose with MySQL/PostgreSQL/Kafka/ClickHouse/Elasticsearch/MinIO + optional Doris/StarRocks profile, seeded schemas/topics |
| Runner | `runner.py` | model matrix (`models.json` or `--provider/--model/--api-key`), repair loop driven by the failing gate's real errors, `--trials` for pass^k, CLI version + commit stamping |
| Reports | `report.py` | layer funnel, repair curve (pass@1 / pass@≤k), per-tier & per-category matrices → console + summary.md/csv + results.json |
| Entry point | `run_benchmark.sh` | one command: dependency install → environment preflight → run → report |
| Tests | `tests/test_benchmark_scoring.py` | 12 tests over the L1 scorer and task-suite integrity |

**Design guarantees**:
- Gates mirror the CLI's own check → dry-run → run pipeline and degrade gracefully (missing engine/Docker disables deeper gates per task, never fails the run).
- All verdicts are deterministic (exit codes, liveness, regex assertions) — no LLM-as-judge.
- Credentials are environment variables only, never written to files (ASF compliance); `results/` and `models.json` are git-ignored.
- Reports are stamped with the CLI version + git commit, so any CLI change can be A/B measured on identical tasks.

[...]

PR #11658, opened 2026-08-05
[Fix][CLI] Include transform blocks in routing validation and run it unconditionally

### LINKED ISSUE #11657: [Bug][CLI] Routing validation rejects valid transform-mediated topologies and is skipped entirely without pyhocon
### Search before asking

### What happened

Two related defects in the CLI's local routing validation (`validate_hocon` in `seatunnel_cli/agents.py`):

**1. Transform blocks are invisible to plugin_output/plugin_input pairing validation.**

`_extract_connector_blocks_raw` only scans the `source` and `sink` sections. `_validate_routing_pairs` therefore never sees labels emitted by transform blocks, and any valid config where a sink consumes a transform's output is falsely rejected:

```
ERROR: sink.Console: plugin_input "filtered" has no matching plugin_output in any source
```

A minimal valid config that reproduces the false rejection:

```hocon
env { job.mode = "BATCH" }
source { FakeSource {
    row.num = 5
    schema { fields { id = "bigint" } }
    plugin_output = "raw" } }
transform {
  Sql {
    plugin_input = "raw"
    plugin_output = "filtered"
    query = "SELECT * FROM raw WHERE id > 1"
  }
}
sink { Console { plugin_input = "filtered" } }
```

This affects every transform-mediated topology: chains (source → transform → sink) and predicate splits (one source feeding parallel SQL transforms routed to different sinks). Because `validate_hocon` also drives the CLI's auto-fix loop, the false error sends repair rounds chasing a non-existent problem — often "fixing" a correct config into a wrong one.

**2. Routing validation silently skipped when pyhocon is unavailable.**

The `_validate_routing_pairs` call sites live inside the two config-parsing branches (duplicate-connector raw path / pyhocon path). In an environment without pyhocon and without duplicate connector blocks, neither branch runs, and routing validation is skipped entirely — configs with dangling `plugin_input` labels validate as clean.

- Labels emitted by transform blocks participate in routing validation.
- Genuinely dangling `plugin_input` references are still rejected.
- Routing validation runs regardless of parser availability, since it only needs the raw block text.

### SeaTunnel Version

dev

### SeaTunnel Config

See above.

### Running Command

```
seatunnel "route rows with amount >= 1000 to files and the rest to console"  # then /check
```

### Error Exception

See above.

_No response_

### Java or Scala Version

Python 3.10+ (seatunnel-cli)

### Screenshots

_No response_

### Code of Conduct

SEZ9:

Fixes #11657 — two related defects in the CLI's local routing validation (`validate_hocon`):

**1. Transform blocks were invisible to plugin_output/plugin_input pairing validation.**

`_extract_connector_blocks_raw` only scanned the `source` and `sink` sections, so labels emitted by transform blocks never reached `_validate_routing_pairs`. Every valid transform-mediated topology was falsely rejected:

- chains: `source → Sql → sink` — the sink consuming the Sql transform's output label got `plugin_input "filtered" has no matching plugin_output in any source`;
- predicate splits: one source feeding parallel Sql transforms (mutually exclusive WHERE clauses) routed to different sinks — both sink inputs falsely flagged.

The impact goes beyond a wrong error message: `validate_hocon` drives the CLI's generate-validate-fix loop, so the false rejection sent auto-fix rounds "repairing" a correct config — typically degrading it (dropping the transform, rewiring the sink to the raw source label) until attempts were exhausted. Measured on the accuracy benchmark (#11553), the conditional-routing task category was heavily suppressed by this: correct model outputs could not pass local validation at all.

Fix: include `transform` in the section scan. Labels emitted by transforms now participate in pairing validation; genuinely dangling `plugin_input` references are still rejected (covered by a regression test).

**2. Routing validation was silently skipped when pyhocon is unavailable.**

The `_validate_routing_pairs` call sites lived inside the two config-parsing branches (duplicate-connector raw path and pyhocon path). In an environment without pyhocon and without duplicate connector blocks, neither branch executed — configs with dangling labels validated as clean. Routing validation only needs the raw block text extracted by `_extract_connector_blocks_raw`, so it now runs unconditionally whenever blocks were extracted, independent of parser availability.

Yes, two behavior corrections for `/check` and the generation-time validation loop:
- Valid configs where sinks consume transform outputs (chains, splits) now pass local validation instead of being falsely rejected.
- Configs with genuinely unmatched `plugin_input` labels are now consistently rejected even in installations without pyhocon (previously skipped silently).

- Three new regression tests in `tests/test_validate_hocon_placeholders.py`: sink consuming a transform output (chain), predicate split via parallel transforms consuming the same source label, and a dangling `plugin_input` that must still be rejected. Full suite: 74 passed.
- Manually verified against the exact config shape that a benchmarked model generated for a split task (Kafka source → two Sql transforms → two Kafka sinks): previously `INVALID` with two false routing errors, now `VALID`; the same config with an intentionally broken label is still rejected.
- Verified in both parser environments (with and without pyhocon installed): validation outcome is now identical.

### DIFF
--- a/seatunnel-cli/seatunnel_cli/agents.py
+++ b/seatunnel-cli/seatunnel_cli/agents.py
@@ -288,7 +288,7 @@ def _extract_connector_blocks_raw(
     (inclusive), e.g. ``'url = "..." driver = "..."'``.
     """
     results: list[tuple[str, str, str]] = []
-    for section in ("source", "sink"):
+    for section in ("source", "transform", "sink"):
         # Find the section's opening brace
         pattern = re.compile(
             rf"(?:^|\n)\s*{section}\s*\{{", re.IGNORECASE,
@@ -345,8 +345,13 @@ def _validate_routing_pairs(
     errors: list[str],
     warnings: list[str],
 ) -> None:
-    """Validate plugin_output / plugin_input pairing across connector blocks."""
-    outputs: dict[str, str] = {}  # label → connector_name
+    """Validate plugin_output / plugin_input pairing across connector blocks.
+
+    Blocks may come from the source, transform, or sink sections; messages
+    carry the real ``section.connector`` location so diagnostics point at
+    the right block (transforms both emit and consume labels).
+    """
+    outputs: dict[str, str] = {}  # label → "section.connector_name"
     inputs: dict[str, str] = {}
 
     label_re = re.compile(
@@ -354,18 +359,19 @@ def _validate_routing_pairs(
     )
 
     for section, connector_name, block_content in blocks:
+        location = f"{section}.{connector_name}"
         for m in label_re.finditer(block_content):
             label = m.group(1)
             if "plugin_output" in m.group(0):
                 if label in outputs:
                     errors.append(
                         f"Duplicate plugin_output label \"{label}\" "
-                        f"in source.{connector_name} "
-                        f"(already used by source.{outputs[label]})"
+                        f"in {location} "
+                        f"(already used by {outputs[label]})"
                     )
-                outputs[label] = connector_name
+                outputs[label] = location
             else:
-                inputs[label] = connector_name
+                inputs[label] = location
 
     # Only validate pairing when routing labels are actually used
     if not outputs and not inputs:
@@ -375,16 +381,16 @@ def _validate_routing_pairs(
     if unmatched_inputs:
         for label in unmatched_inputs:
             errors.append(
-                f"sink.{inputs[label]}: plugin_input \"{label}\" "
-                f"has no matching plugin_output in any source"
+                f"{inputs[label]}: plugin_input \"{label}\" "
+                f"has no matching plugin_output in any source or transform"
             )
 
     orphan_outputs = set(outputs) - set(inputs)
     if orphan_outputs:
         for label in orphan_outputs:
             warnings.append(
-                f"source.{outputs[label]}: plugin_output \"{label}\" "
-                f"has no matching plugin_input in any sink"
+                f"{outputs[label]}: plugin_output \"{label}\" "
+                f"has no matching plugin_input in any transform or sink"
             )
 
 
@@ -463,9 +469,6 @@ def validate_hocon(config_str: str) -> str:
             except Exception:
                 pass
 
-        # Validate routing labels
-        _validate_routing_pairs(raw_blocks, errors, warnings)
-
     elif parsed is not None:
         # Standard single-connector-per-type path (pyhocon is accurate here)
         for section in ["source", "sink"]:
@@ -496,9 +499,10 @@ def validate_hocon(config_str: str) -> str:
             except Exception:
                 pass
 
-        # Also validate routing for single-connector configs if labels present
-        if raw_blocks:
-            _validate_routing_pairs(raw_blocks, errors, warnings)
+    # Routing validation is independent of option metadata and pyhocon —
+    # always run it when connector blocks were extracted.
+    if raw_blocks:
+        _validate_routing_pairs(raw_blocks, errors, warnings)
 
     # Check STREAMING mode needs checkpoint.interval
     if parsed is not None:
--- a/seatunnel-cli/tests/test_validate_hocon_placeholders.py
+++ b/seatunnel-cli/tests/test_validate_hocon_placeholders.py
@@ -103,3 +103,121 @@ def test_colon_separator_still_rejects_env_vars_elsewhere():
     config = _config_with_sink('topic: "${now}"')
     result = validate_hocon(config)
     assert "Unresolved environment variables" in result
+
+
+# ── transform-mediated routing (regression for transform blocks being ──
+# ── invisible to _validate_routing_pairs)                             ──
+
+def test_sink_consuming_transform_output_is_valid():
+    config = """
+env { parallelism = 1
+  job.mode = "BATCH" }
+source { FakeSource {
+    row.num = 5
+    schema { fields { id = "bigint" } }
+    plugin_output = "raw" } }
+transform {
+  Sql {
+    plugin_input = "raw"
+    plugin_output = "filtered"
+    query = "SELECT * FROM raw WHERE id > 1"
+  }
+}
+sink { Console { plugin_input = "filtered" } }
+"""
+    result = validate_hocon(config)
+    assert "has no matching plugin_output" not in result
+
+
+def test_split_via_parallel_transforms_is_valid():
+    config = """
+env { parallelism = 1
+  job.mode = "BATCH" }
+source { FakeSource {
+    row.num = 5
+    schema { fields { id = "bigint" } }
+    plugin_output = "raw" } }
+transform {
+  Sql {
+    plugin_input = "raw"
+    plugin_output = "big"
+    query = "SELECT * FROM raw WHERE id > 100"
+  }
+  Sql {
+    plugin_input = "raw"
+    plugin_output = "small"
+    query = "SELECT * FROM raw WHERE id <= 100"
+  }
+}
+sink {
+  Console { plugin_input = "big" }
+  Console { plugin_input = "small" }
+}
+"""
+    result = validate_hocon(config)
+    assert "has no matching plugin_output" not in result
+
+
+def test_genuinely_unmatched_plugin_input_still_rejected():
+    config = """
+env { parallelism = 1
+  job.mode = "BATCH" }
+source { FakeSource {
+    row.num = 5
+    schema { fields { id = "bigint" } }
+    plugin_output = "raw" } }
+sink { Console { plugin_input = "nonexistent_label" } }
+"""
+    result = validate_hocon(config)
+    assert "nonexistent_label" in result
+    assert "has no matching plugin_output" in result
+
+
+def test_routing_errors_attribute_transform_blocks_correctly():
+    # A dangling plugin_input on a TRANSFORM must be reported against the
+    # transform block, not misattributed to a source or sink.
+    config = """
+env { parallelism = 1
+  job.mode = "BATCH" }
+source { FakeSource {
+    row.num = 5
+    schema { fields { id = "bigint" } }
+    plugin_output = "raw" } }
+transform {
+  Sql {
+    plugin_input = "wrong_label"
+    plugin_output = "filtered"
+    query = "SELECT * FROM wrong_label"
+  }
+}
+sink { Console { plugin_input = "filtered" } }
+"""
+    result = validate_hocon(config)
+    assert 'transform.Sql: plugin_input "wrong_label"' in result
+
+
+def test_duplicate_output_across_transforms_reports_transform_location():
+    config = """
+env { parallelism = 1
+  job.mode = "BATCH" }
+source { FakeSource {
+    row.num = 5
+    schema { fields { id = "bigint" } }
+    plugin_output = "raw" } }
+transform {
+  Sql {
+    plugin_input = "raw"
+    plugin_output = "same_label"
+    query = "SELECT * FROM raw WHERE id > 1"
+  }
+  Sql {
+    plugin_input = "raw"
+    plugin_output = "same_label"
+    query = "SELECT * FROM raw WHERE id <= 1"
+  }
+}
+sink { Console { plugin_input = "same_label" } }
+"""
+    result = validate_hocon(config)
+    assert 'in transform.Sql' in result
+    assert 'already used by transform.Sql' in result


DanielLeens (approved):

The `seatunnel-cli`'s local HOCON validator (`validate_hocon` / `_validate_routing_pairs` in `seatunnel_cli/agents.py`) is supposed to check that every `plugin_input` label pairs with a matching `plugin_output` label somewhere in the job DAG. Before this fix, two defects made that check unreliable:

1. `_extract_connector_blocks_raw` only scanned the `source` and `sink` sections, so any `plugin_output`/`plugin_input` label emitted by a `transform` block was invisible to the pairing check. Every valid transform-mediated topology (chain `source → transform → sink`, or a predicate split with one source feeding parallel transforms routed to different sinks) was falsely rejected with `plugin_input "..." has no matching plugin_output in any source`.
2. `_validate_routing_pairs` was only invoked from inside the two config-parsing branches (`has_duplicate_connectors` raw path and the `pyhocon`-parsed path). If neither branch's condition held (no duplicate connector names *and* the `pyhocon` parse didn't succeed/run), routing validation silently never ran at all, so genuinely dangling labels went undetected.

Since `validate_hocon`'s output feeds the CLI's LLM-driven generate → validate → fix loop, defect #1 was actively harmful: a correct model-generated config would get flagged as invalid, and the auto-fix loop would then "repair" a config that wasn't broken (typically by degrading it — dropping the transform or rewiring the sink to read the raw source label directly) until the retry budget was exhausted.

The fix: include `transform` in the section scan (`("source", "transform", "sink")`), and hoist the `_validate_routing_pairs` call out of both branches so it runs unconditionally whenever `raw_blocks` were extracted.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**`_extract_connector_blocks_raw` (seatunnel-cli/seatunnel_cli/agents.py:278-340)** — the loop that scans top-level sections went from `for section in ("source", "sink")` to `for section in ("source", "transform", "sink")`. The rest of the function (brace-counting block extraction) is untouched and section-agnostic, so this is a minimal, correct extension — transform blocks are extracted the same way source/sink blocks always were.

**`validate_hocon` (agents.py:391-567)** — the call site for `_validate_routing_pairs` was removed from two places:
- inside `if has_duplicate_connectors and raw_blocks:` (the multi-pipeline raw-text path)
- inside `elif parsed is not None:` guarded by `if raw_blocks:` (the pyhocon-parsed single-connector path)

and replaced with a single unconditional call after both branches:
```python
# Routing validation is independent of option metadata and pyhocon —
# always run it when connector blocks were extracted.
if raw_blocks:
    _validate_routing_pairs(raw_blocks, errors, warnings)
```
This is a straightforward de-duplication (was called from two sites, is now called from one) that also closes the coverage gap described above. I traced all three call conditions in the old vs. new code and confirmed the only behavioral delta is: routing validation now also runs when `parsed is None` and there's no duplicate-connector block (previously skipped entirely in that case). `pyhocon>=0.3.60` is a hard dependency declared in `seatunnel-cli/pyproject.toml:32`, so in a correctly-installed environment `parsed` is `None` only when the HOCON has an actual syntax error (already flagged separately as `HOCON parse error: ...`, i.e. the config is already `INVALID`) — the "pyhocon not installed" branch is realistically dead/defensive code. So in practice this unconditional call closes a real gap without creating a new one for the mainstream install path.

**Runtime path** (text-flow, since this only affects the CLI's local pre-check, not the Java engine):
```
Orchestrator.process_user_input()
  → _run_config_generator()               (LLM produces HOCON)
  → loop (max 3 rounds):
      _run_validator(config)
        → validate_hocon(config)           <-- this PR's code path
        → LLM prompt embeds local_result text, LLM returns PASS/FAIL
      if PASS: dry_run_config(config)      (real engine-level check, unaffected by this PR)
      else: _run_fix(config, validation)   (LLM "repairs" based on local_result text)
```
Because `local_result` (the raw string `validate_hocon` returns, including the exact wording of each `ERROR:`/`WARNING:` line) is embedded verbatim into the LLM validator/fixer prompts, both the correctness of the verdict *and* the wording of the message materially affect the auto-fix loop's behavior — not just a human reading a report. This is why I flag Issue 1 below (message mislabeling) as more than cosmetic.

**`_validate_routing_pairs` itself (agents.py:343-388) was not touched by this diff**, but its behavior is now materially different because it is fed a wider set of blocks (transform blocks now participate). I checked its label bookkeeping (`outputs`/`inputs` dicts keyed by label, one shared namespace across all sections) and confirmed the pairing logic itself is section-agnostic and correct — a transform's `plugin_output` legitimately satisfies a sink's or another transform's `plugin_input`, and vice versa. However, three of its message strings hardcode `"source."` / `"sink."` prefixes regardless of which section the offending block actually came from — see Issue 1.

## 1.2 Compatibility Impact

**Fully compatible.** This is a local, in-process CLI heuristic (not a wire format, not a persisted config schema, not an engine-level check) with no serialization or historical-version surface. Concretely, for existing user job configs:
- Configs that were previously and correctly validated as `VALID` (no routing labels, or labels fully confined to source/sink) are unaffected — the early return `if not outputs and not inputs: return` and the unchanged label-matching logic mean nothing changes for them.
- Configs that were previously **falsely rejected** (sink consuming a transform's output) now correctly pass — this is the intended fix, verified by the new `test_sink_consuming_transform_output_is_valid` and `test_split_via_parallel_transforms_is_valid` tests.
- The only way this change could make a previously-`VALID` config newly fail is if it has a real dangling `plugin_input`/duplicate `plugin_output` inside a `transform` block that was invisible before. That is, by definition, a real misconfiguration the tool is supposed to catch — I don't consider a fix that closes a real detection gap a compatibility break, and the PR author explicitly confirms and tests this as intended ("Configs with genuinely unmatched plugin_input labels are now consistently rejected even in installations without pyhocon — previously skipped silently"). Even in the worst case, the practical blast radius is soft: `validate_hocon`'s output only feeds the local pre-check/LLM-fix loop (`_run_validator`/`_run_fix`), not a hard gate on job submission — the authoritative gate remains the unaffected `dry_run_config` engine-level check.

Negligible. `_extract_connector_blocks_raw` now does one additional linear regex scan+brace-walk over the `transform` section text (same algorithm already used for `source`/`sink`), bounded by config size (typically a few KB). No new I/O, no new external calls, no change to `dry_run_config` or the LLM call paths.

### Issue 1
- **Location**: `seatunnel-cli/seatunnel_cli/agents.py:362-365` (duplicate `plugin_output` error), `seatunnel-cli/seatunnel_cli/agents.py:377-380` (unmatched `plugin_input` error), `seatunnel-cli/seatunnel_cli/agents.py:385-388` (orphan `plugin_output` warning)
- **Problem**: `_validate_routing_pairs` builds these three messages with hardcoded literal `"source."` (duplicate-output error, orphan-output warning) and `"sink."` (unmatched-input error) prefixes, even though the loop that populates `outputs`/`inputs` has the real `section` value (`"source"`, `"transform"`, or `"sink"`) available at `agents.py:356`. Before this PR, only `source` blocks could ever produce a `plugin_output` and only `sink` blocks could ever produce a `plugin_input` (since those were the only two sections scanned), so the hardcoded prefixes happened to always be correct. Now that `transform` blocks are scanned too, and a transform block routinely has *both* `plugin_output` and `plugin_input`, these messages will misreport the block's section whenever the actual offender is a `transform` — e.g. a dangling `plugin_input` on a transform will be reported as `sink.MyTransform: plugin_input "..." has no matching plugin_output...` even though `MyTransform` is not a sink.
- **Potential risk**: Since this exact string is fed verbatim into the LLM prompts driving `_run_validator`/`_run_fix` (agents.py:1644-1730), a mislabeled location can steer the auto-fix agent toward editing the wrong block (e.g. touching the sink when the real problem is in a transform), reducing fix-loop effectiveness — which is precisely the failure mode this PR is trying to reduce for the *other* defect. It's not a functional correctness bug in the VALID/INVALID verdict, but it undermines the actionability of the message for both the LLM fixer and any human reading the CLI output.
- **Best improvement**: Replace the hardcoded `"source."`/`"sink."` literals with the actual `section` variable already in scope, e.g. `f"{section}.{connector_name}: plugin_input \"{label}\" has no matching plugin_output in any source or transform"` (Option A — minimal, one-line-per-message fix). Option B: track `(section, connector_name)` tuples in `outputs`/`inputs` instead of just `connector_name`, so the messages can name the exact producing/consuming section precisely in all three spots.
- **Severity**: Medium

### Issue 2
- **Location**: `.github/workflows/*` (repo-wide — no file specific to `seatunnel-cli`)
- **Problem**: I grepped the entire `.github/workflows/` tree for any reference to `seatunnel-cli` or a Python test invocation for this module and found none — the `backend.yml`/`build_main.yml` "Build" workflows are exclusively Maven/Java-module-scoped (they use `tools/update_modules_check/*.py` purely to decide which *Maven* modules to build/test). Because this PR's changes are entirely inside `seatunnel-cli/` (pure Python, not part of the Maven reactor), the green "Build" check on this PR did not execute the new or existing `seatunnel-cli/tests/*` suite at all.
- **Potential risk**: The PR description's test evidence ("Full suite: 74 passed") is a manual, unverifiable-by-CI claim. Any future regression to `agents.py` (including reverting the fix in this exact PR) would not be caught automatically by GitHub Actions — only by a human remembering to run `pytest` locally in `seatunnel-cli/`. This is a pre-existing, repo-wide gap (not introduced by this PR), but it directly limits how much confidence the "CI is green" signal should carry for this specific change.
- **Best improvement**: Follow-up (need not block this PR): add a lightweight GitHub Actions job that runs `pytest` under `seatunnel-cli/tests` whenever `seatunnel-cli/**` changes, mirroring the existing path-filtered pattern already used for Maven modules in `backend.yml`.
- **Severity**: Medium

# 2. Code Quality Assessment

## 2.1 Coding Standards

The diff itself is small and clean: a one-word tuple addition, and hoisting three lines of an existing call site out of two branches into one, with a clear inline comment explaining *why* it's now unconditional (`# Routing validation is independent of option metadata and pyhocon — always run it when connector blocks were extracted.`). No new functions, classes, or non-trivial fields were introduced, so there's nothing new requiring a docstring; the touched functions (`_extract_connector_blocks_raw`, `validate_hocon`) already carry adequate docstrings that remain accurate (neither hardcodes "source"/"sink" in its prose). No wildcard imports, no style deviations from the surrounding file.

Three new regression tests were added to `seatunnel-cli/tests/test_validate_hocon_placeholders.py`, and they hit the actual reported bug precisely:
- `test_sink_consuming_transform_output_is_valid` — reproduces the exact chain topology from the linked issue (#11657) and asserts the false-positive error string is absent.
- `test_split_via_parallel_transforms_is_valid` — reproduces the predicate-split topology (two `Sql` transforms both reading the same source label, feeding two sinks), which also exercises the `has_duplicate_connectors` multi-pipeline branch since both transform blocks share the connector name `Sql` — a good side-effect of this test, since it proves the fix holds across both parsing branches, not just the single-connector path.
- `test_genuinely_unmatched_plugin_input_still_rejected` — a true-negative test proving the fix didn't overcorrect into silencing real dangling-label errors.

All three are pure in-memory string/regex operations against `validate_hocon` — no `Thread.sleep`, no wall-clock or network dependence, no containers/ports, no shared mutable module-level state between tests, no ordering dependency, no floating-point comparisons. **Stability rating: Stable.**

One coverage gap: none of the new tests assert on the exact wording of the error/warning message text for a transform-originated issue, so the `"source."`/`"sink."` mislabeling described in Issue 1 isn't caught by this suite (the tests only check the presence/absence of the `"has no matching plugin_output"` substring, not which section is named). Not a blocker, but worth adding a targeted assertion once Issue 1 is fixed.

As noted in Issue 2, these tests are not currently wired into any GitHub Actions workflow, so their "pass" status here is only verifiable locally, not via this PR's CI run.

**CI status**: `Build`, `Notify test workflow`, and `labeler` all passed on this PR's head commit. As explained in Issue 2, the `Build` check is Maven/Java-scoped and never touches `seatunnel-cli/**`, so a green `Build` check here confirms the Maven reactor wasn't broken but says nothing about the Python change or its tests.

## 2.3 Documentation Updates

Not applicable — no `docs/en`/`docs/zh` update is needed. In fact I found that `docs/en/ai-cli/design.md:15` already documents the intended behavior this PR restores: `"DAG wiring semantics — plugin_output/plugin_input labels must pair exactly across source/transform/sink blocks"`. The design doc already described the correct/target behavior; the code simply hadn't matched it for `transform` blocks. `docs/en/ai-cli/benchmark.md:72-73` also already lists "conditional routing" / "plugin_output/plugin_input wiring with parallel SQL transforms" as a known model/tooling weakness cluster being actively addressed — this PR is a concrete instance of that documented engineering target, not a divergence from it.

# 3. Architectural Soundness

**Precise fix.** Both defects are addressed at their root cause: the section-scan tuple was simply missing `"transform"`, and the call-site duplication (two conditional call sites instead of one unconditional one) was the direct cause of the "sometimes skipped" behavior. Nothing about the fix is a workaround or a patch bolted on top of a symptom.

## 3.2 Maintainability
Improved slightly — collapsing two call sites of `_validate_routing_pairs` into one reduces the chance of a future edit updating one branch's call and forgetting the other (which is exactly how defect #2 was introduced in the first place).

## 3.3 Extensibility
If SeaTunnel ever needs a similar text-driven check that spans further section types, this establishes a good precedent — the tuple-based section list in `_extract_connector_blocks_raw` and the unconditional post-branch validation call in `validate_hocon` are both easy to extend without restructuring.

Not applicable in the checkpoint/serialization/protocol sense — this code has no persisted state, no wire format, and isn't part of the Java engine's execution or recovery path. See section 1.2 for the practical backward-compatibility analysis of the validator's output.

# 4. Issue Summary

|-----|-------|----------|----------|
| 1 | Hardcoded `"source."`/`"sink."` prefixes in routing error/warning messages misreport the section when the offending block is a `transform` | seatunnel-cli/seatunnel_cli/agents.py:362-365, 377-380, 385-388 | Medium |
| 2 | No GitHub Actions workflow executes the `seatunnel-cli` pytest suite; this PR's test evidence is unverified by CI | .github/workflows/* (repo-wide gap) | Medium |

# 5. Merge Recommendation

No blocking issues found. Both issues above are real but non-blocking: Issue 1 is a message-accuracy bug that doesn't change the VALID/INVALID verdict, and Issue 2 is a pre-existing, repo-wide CI gap this PR didn't create — neither affects the correctness of the fix itself.

1. Blockers — must be fixed
   - None.

   - Fix Issue 1 by using the real `section` variable instead of the hardcoded `"source."`/`"sink."` literals in `_validate_routing_pairs` (small follow-up, ideally in this PR since it's a two-line touch to a function this PR already changes the behavior of).
   - Track Issue 2 as a separate follow-up: add a path-filtered GitHub Actions job that runs `pytest seatunnel-cli/tests` so future changes to `agents.py` (including accidental reverts of this exact fix) are caught automatically.
   - Optionally extend one of the new tests to assert on the exact section name in a transform-originated error/warning message, to lock in the Issue 1 fix once made.

Overall this is a well-scoped, well-tested bug fix that is precisely targeted at a real, verifiable defect (confirmed independently by reading the code, not just by trusting the linked issue/PR description) with a documented and low-risk compatibility story. The core logic change is correct and the new tests genuinely exercise both parsing branches. I don't see a better alternative approach to the two defects fixed here — the chosen fix (widen the section tuple, de-duplicate the call site) is the minimal, direct correction. No blocking issues found beyond the two Medium items above, which I'd like addressed but wouldn't hold up the merge over.

DanielLeens (approved):
### Re-review note (carryover check)

This is a full re-review against the current head (`c35d4f1b5`). A new commit (`f70e592f4`, merged with `dev` at `c35d4f1b5`) landed after my previous approval on this PR. I re-verified both issues from my earlier review against the current code:

- **My earlier Issue 1** (hardcoded `"source."`/`"sink."` prefixes misattributing transform-originated routing errors) **is now fixed**. `_validate_routing_pairs` tracks a real `f"{section}.{connector_name}"` location per label instead of a bare connector name, and two new regression tests (`test_routing_errors_attribute_transform_blocks_correctly`, `test_duplicate_output_across_transforms_reports_transform_location`) lock in the exact `transform.Sql: ...` wording — this also closes the test-coverage gap I flagged as an optional follow-up last time. Nice fix, and thank you for adding the coverage for it.
- **My earlier Issue 2** (no GitHub Actions workflow runs the `seatunnel-cli` pytest suite) is **still open** — unchanged, still non-blocking, still a repo-wide gap rather than something introduced by this PR.

While re-tracing the full runtime call chain on the current head (not just diffing against my prior comments), I found one new, previously-unflagged side effect of the section-scan widening — see Issue 2 below (renumbered, since the old Issue 1 is now resolved). I did not miss this last time because the code that exposes it wasn't touched by the commit I reviewed then in a way that changed its risk; it's a genuine second-order consequence of the current commit, not something I should have caught in the prior round, so I'm not filing this as a "sorry, missed it" carryover — just a fresh finding on the current diff.

**Note on PR #11659**: that PR (same author) is currently based on `dev` without rebasing on top of this branch, but its diff against `dev` is byte-identical to this PR's for `seatunnel-cli/seatunnel_cli/agents.py` (same `_extract_connector_blocks_raw`/`_validate_routing_pairs`/`validate_hocon` hunks, same test additions), plus its own additive skill/docs content and one unrelated prompt-text addition to `dry_run_config`. In other words, #11659 currently carries a full duplicate copy of this fix. That's not a problem for merging either PR individually, but once this PR merges, #11659 will need a rebase/merge from `dev` to drop the now-duplicated hunks and show only its own incremental content — worth keeping in mind when sequencing the two.

The CLI's local HOCON validator (`validate_hocon` / `_validate_routing_pairs` in `seatunnel_cli/agents.py`) checks that every `plugin_input` label pairs with a matching `plugin_output` label in the job DAG. Two defects made that check unreliable: (1) `_extract_connector_blocks_raw` only scanned `source`/`sink`, so labels emitted by `transform` blocks were invisible, causing every valid transform-mediated topology (chains, predicate splits) to be falsely rejected; and (2) `_validate_routing_pairs` was only invoked from inside two mutually-exclusive config-parsing branches, so in an environment without `pyhocon` and without duplicate connector blocks, routing validation silently never ran at all. Since `validate_hocon`'s output feeds the CLI's LLM-driven generate → validate → fix loop, defect #1 was actively harmful — the auto-fix loop would "repair" correct configs into broken ones. The fix widens the section scan to `("source", "transform", "sink")` and hoists the `_validate_routing_pairs` call to run unconditionally once, after both parsing branches.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**`_extract_connector_blocks_raw`** (`seatunnel-cli/seatunnel_cli/agents.py:291`): the section-scan tuple went from `("source", "sink")` to `("source", "transform", "sink")`. The brace-counting extraction logic itself is unchanged and section-agnostic, so this is a minimal, correct extension.

**`_validate_routing_pairs`** (`agents.py:343-394`): now tracks a `location = f"{section}.{connector_name}"` string per label instead of a bare connector name, and all three diagnostic messages (`Duplicate plugin_output label ...`, unmatched `plugin_input`, orphan `plugin_output`) use the real location. This directly addresses the misattribution problem I raised previously, and is covered by two dedicated regression tests.

**`validate_hocon`** (`agents.py:397+`): the `_validate_routing_pairs` call was removed from both branch-local call sites and replaced with a single unconditional call after the `if/elif` block:
```python
if raw_blocks:
    _validate_routing_pairs(raw_blocks, errors, warnings)
```
I confirmed the only behavioral delta versus the pre-fix code is that routing validation now also runs when `parsed is None` and there is no duplicate-connector block — previously skipped entirely. Since `pyhocon>=0.3.60` is a hard dependency (`seatunnel-cli/pyproject.toml`), `parsed is None` in a correctly-installed environment only happens on an actual HOCON syntax error (already flagged separately), so this closes a real gap without creating a new failure mode on the mainstream path.

**Runtime path** (text-flow; this only affects the CLI's local pre-check, not the Java engine):
```
Orchestrator.process_user_input()
  -> _run_config_generator()               (LLM produces HOCON)
  -> loop (max 3 rounds):
      _run_validator(config)
        -> validate_hocon(config)           <-- this PR's code path
        -> LLM prompt embeds local_result text, LLM returns PASS/FAIL
      if PASS: dry_run_config(config)      (real engine-level check, unaffected by this PR)
      else: _run_fix(config, validation)   (LLM "repairs" based on local_result text)
```
Because `local_result` is embedded verbatim in the validator/fixer LLM prompts, both the verdict *and* the exact wording matter for the auto-fix loop's behavior, which is why the location-string fix (not just a cosmetic nicety) matters for real repair-loop quality.

**Key findings:**
- Both defects from the original bug report are fixed at the root, and the fix is minimal (one tuple entry, one call-site hoist, one label-tracking change).
- The previously-flagged message-misattribution issue is now fixed and specifically tested.
- `_extract_connector_blocks_raw`'s output feeds *two* different validators in `validate_hocon` (routing pairs, and per-block required-option/duplicate-connector detection) — widening it for routing purposes has a side effect on the other consumer too; see Issue 2 below.
- `_validate_routing_pairs`'s label bookkeeping is a single shared namespace across all three sections, which is correct: a transform's `plugin_output` legitimately satisfies a sink's or another transform's `plugin_input`.

## 1.2 Compatibility Impact

Fully compatible — this is a local, in-process CLI heuristic with no serialization or persisted-config surface. Configs previously and correctly validated `VALID` (no routing labels, or labels confined to source/sink) are unaffected. Configs previously falsely rejected (sink consuming a transform's output) now correctly pass, per the new tests. The only way a previously-`VALID` config could newly fail is a real dangling/duplicate label inside a `transform` block that was invisible before — that is, by definition, a real misconfiguration the tool should catch, and the practical blast radius stays soft either way since `validate_hocon` only feeds the local pre-check/LLM-fix loop, not a hard gate — `dry_run_config` remains the unaffected, authoritative engine-level check.

CPU/memory impact is negligible — one additional linear regex scan + brace-walk over the `transform` section text, bounded by config size, no new I/O.

There is, however, a non-obvious control-flow side effect worth calling out (elaborated as Issue 2 in 1.4): because `_extract_connector_blocks_raw`'s output is also used to compute `has_duplicate_connectors` (`agents.py:434-437`), and that flag decides which of two mutually-exclusive branches processes the *entire* config (including source/sink), widening the scan to include `transform` changes which branch a config lands in whenever it has two identically-named transform blocks — even if its source/sink blocks are all uniquely named and were previously routed through the other branch.

### Issue 1 (carried forward from my previous review — unresolved, non-blocking)
- **Location**: `.github/workflows/*` (repo-wide — no file specific to `seatunnel-cli`)
- **Problem**: No GitHub Actions workflow executes the `seatunnel-cli` pytest suite; the `Build` workflow is exclusively Maven/Java-reactor-scoped.
- **Potential risk**: The PR's test evidence ("Full suite: 74 passed") remains a manual, CI-unverifiable claim; a future regression to `agents.py` (including an accidental revert of this exact fix) would only be caught by a human running `pytest` locally.
- **Best improvement**: Follow-up (does not need to block this PR): add a path-filtered GitHub Actions job that runs `pytest seatunnel-cli/tests` when `seatunnel-cli/**` changes.
- **Severity**: Medium
- **Raised by another reviewer**: No (raised by me in my previous review on this PR; still applicable, not introduced by the new commit)

### Issue 2 (new finding on the current head)
- **Location**: `seatunnel-cli/seatunnel_cli/agents.py:291` (`_extract_connector_blocks_raw` section tuple) interacting with `agents.py:434-500` (`has_duplicate_connectors` branch selection and the `elif parsed is not None:` branch)
- **Problem**: `has_duplicate_connectors` is computed from `Counter((section, name) for section, name, _ in raw_blocks)` over *all* extracted blocks, and now that `transform` blocks are extracted too, a config with two identically-named transforms (exactly the predicate-split pattern this PR's own tests, and the PR description, use as the motivating example) sets `has_duplicate_connectors = True` even when its source and sink connector names are all unique. That routes the whole config into the `if has_duplicate_connectors and raw_blocks:` raw multi-pipeline branch instead of the `elif parsed is not None:` branch — and `_check_conditional_mismatches` (`agents.py:492-498`, the generic conditional-option-mismatch check, e.g. format-dependent options) is called *only* from inside the `elif` branch, for `section in ["source", "sink"]`. So for a predicate-split job with distinctly-named sinks (e.g., `Console` + `Kafka`), the source and sink blocks silently lose conditional-mismatch checking after this fix, even though those blocks themselves are untouched by the diff — purely because an unrelated transform pair triggered the other branch. Required-option checking (`validate_connector_options`) is unaffected either way, since the raw-path branch still runs it for every block including source/sink.
- **Potential risk**: This is a coverage reduction, not a correctness regression in the VALID/INVALID verdict for the config itself — no valid config becomes newly rejected, and `dry_run_config` remains the authoritative engine-level gate, so nothing unsafe reaches production undetected. But it does mean the local pre-check gets quietly weaker in exactly the "predicate split via parallel transforms" scenario this PR (and the companion PR #11659, which adds a dedicated skill for that exact pattern) is built to support well — a source/sink conditional-option mistake in that topology that would previously have been caught locally will now only surface at the (slower, LLM-driven) `dry_run_config` stage.
- **Best improvement**: Option A — decouple `has_duplicate_connectors` into two flags (one for `source`/`sink` duplicates that decides the parsing-branch choice, one for "any duplicate including transform" used only for the raw-block extraction/routing-pairs purpose), so a transform-only duplicate doesn't divert source/sink out of the `pyhocon`-accurate conditional-mismatch path. Option B (smaller) — also invoke `_check_conditional_mismatches` for source/sink inside the raw multi-pipeline branch, using the same block-derived `provided_keys`/`connector_conf` construction already built there for `validate_connector_options`.
- **Severity**: Medium

# 2. Code Quality Assessment

## 2.1 Coding Standards

The diff is small and clean: a one-word tuple addition, a docstring update on `_validate_routing_pairs` that now accurately describes the multi-section location tracking, and hoisting the call site. No new functions/classes/non-trivial fields were introduced. No wildcard imports, no style deviations.

Five tests now cover this behavior in `test_validate_hocon_placeholders.py`: the three from the prior round (`test_sink_consuming_transform_output_is_valid`, `test_split_via_parallel_transforms_is_valid`, `test_genuinely_unmatched_plugin_input_still_rejected`) plus two new ones added in the current commit that specifically assert the *location string* is correct for a transform-originated dangling input and a transform-originated duplicate output (`test_routing_errors_attribute_transform_blocks_correctly`, `test_duplicate_output_across_transforms_reports_transform_location`) — I traced both by hand against the current `_validate_routing_pairs` logic and confirmed the expected assertions match the actual code paths (`unmatched_inputs`/`orphan_outputs` set arithmetic, `location` string construction).

All five are pure in-memory string/regex operations against `validate_hocon` — no `Thread.sleep`, no wall-clock dependence, no containers/ports, no shared mutable module-level state or ordering dependency between tests. **Stability rating: Stable.**

Coverage gap (relates to Issue 2 above): none of the tests exercise a predicate-split config that also has a genuine conditional-option mismatch on its source/sink, so the loss of `_check_conditional_mismatches` coverage for that topology isn't caught by this suite. Not a blocker, but a good target for a regression test once Issue 2 is addressed.

As before, these tests aren't wired into any GitHub Actions workflow (Issue 1), so their "pass" status is only verifiable locally, not via this PR's CI run. **CI status on the current head (`c35d4f1b5`)**: `Build`, `Notify test workflow`, and `labeler` all report `SUCCESS`. As explained under Issue 1, `Build` is Maven/Java-reactor-scoped and never touches `seatunnel-cli/**`, so this green check confirms the Maven reactor wasn't broken but doesn't exercise the Python change. No CI-related failures need diagnosing here — the green checks are real, just not evidence for the Python-side tests.

## 2.3 Documentation Updates

Not applicable — `docs/en/ai-cli/design.md` already documents "DAG wiring semantics — `plugin_output`/`plugin_input` labels must pair exactly across source/transform/sink blocks" as the intended behavior, and `docs/en/ai-cli/benchmark.md` already lists "conditional routing... `plugin_output`/`plugin_input` wiring with parallel SQL transforms" as a known weakness being addressed. Both remain accurate against the current code; I re-checked both files at the current head.

# 3. Architectural Soundness

**Precise fix.** Both original defects are addressed at their root cause. Issue 2 above is a second-order consequence of the section-scan widening rather than a flaw in the fix's approach — the routing-validation intent is correctly and minimally implemented; the `has_duplicate_connectors` branch-selection coupling is pre-existing shared plumbing that just wasn't designed with a third section type in mind.

## 3.2 Maintainability
Improved — collapsing two call sites of `_validate_routing_pairs` into one reduces the chance of a future edit updating one branch and forgetting the other, which is exactly how the original "silently skipped" defect was introduced.

## 3.3 Extensibility
Good precedent for future text-driven checks spanning more section types. Issue 2 is a reminder that `raw_blocks`/`has_duplicate_connectors` is now a shared signal feeding two different concerns (branch selection and label pairing) that may be worth decoupling as this file grows.

Not applicable in the checkpoint/serialization/protocol sense — no persisted state, no wire format, not part of the Java engine's execution or recovery path.

# 4. Issue Summary

|-----|-------|----------|----------|
| 1 | No GitHub Actions workflow executes the `seatunnel-cli` pytest suite; test evidence unverified by CI (carried forward, unresolved) | `.github/workflows/*` (repo-wide gap) | Medium |
| 2 | Transform-name duplicates now also flip `has_duplicate_connectors`, diverting predicate-split configs' source/sink blocks away from `_check_conditional_mismatches` coverage | `seatunnel-cli/seatunnel_cli/agents.py:291, 434-500` | Medium |

# 5. Merge Recommendation

No blocking issues. The previously-flagged message-misattribution defect is now fixed and specifically tested — good follow-through on the last round of feedback. The two Medium items above are both real but non-blocking: neither changes the VALID/INVALID verdict for any config that was correct before, and the engine-level `dry_run_config` check remains the unaffected, authoritative gate in both cases.

1. Blockers — must be fixed
   - None.

   - Issue 2: decouple the `pyhocon`-path branch selection from transform-only duplicates (Option A/B above), so predicate-split configs keep full conditional-mismatch coverage on their source/sink blocks — ideally with a regression test pairing a predicate split with a genuine conditional-option mismatch.
   - Issue 1 (unchanged from before): add a path-filtered GitHub Actions job for `pytest seatunnel-cli/tests`.

Overall this is a well-scoped, well-tested bug fix, and the author's response to the prior review round (fixing the message-attribution issue and adding two targeted regression tests for it) is exactly what I'd want to see between rounds. The new Issue 2 I found this round is a real but narrow, non-blocking side effect worth a fast follow-up rather than a hold-up on this PR.

xiaochen-zhou (approved):
+1

corgy-w (approved):
+1

PR #11413, opened 2026-07-12
[Feature][Connector-V2] Add HugeGraph source connector and refactor sink with multi-mapping support

zhupitertop:
### Add a description
- Refactor HugeGraph sink to support multi-mapping (mappings array) replacing single schema_config
- Add HugeGraph source connector with server-side paging, duplicate filtering, and reserved graph columns
- Support VERTEX and EDGE (SINGLE / MULTIPLE + sortKeys) write modes
- Auto-chunk batches exceeding HugeGraph 500-record limit
- Shade and relocate Guava, Jackson, OkHttp, Okio, Javassist, Joda-Time
- Handle DATE properties returned as space-separated strings by HugeGraph server
- Allow empty schema.fields for property-less labels in source
- Default dateFormat (yyyy-MM-dd) and timeZone (GMT+8) for sink mappings
- Improve error message when vertex/edge label does not exist
- Add unit tests for client, mappers, source reader, data type utils, and config validation
- Add E2E tests for sink (IT) and source (SourceIT) with multi-mapping and assert sinks
- Backward-compatible with legacy schema_config format

### Purpose of this pull request

This PR adds a HugeGraph source connector and refactors the existing HugeGraph sink connector to support multi-mapping configuration. The new `mappings` array replaces the legacy single `schema_config`, allowing users to write multiple vertex and edge labels in a single SeaTunnel job. The source connector supports bounded full-label scans with server-side paging, automatic duplicate filtering, and reserved graph columns (`~id`, `~label`, `~source_id`, `~target_id`, etc.).

Related issue: https://github.com/hugegraph/hugegraph/issues/151 (if applicable)

### Does this PR introduce _any_ user-facing change?

Yes.

**Sink connector changes:**

Before (legacy `schema_config`):
```hocon
sink {
  HugeGraph {
    schema_config = {
      type = "VERTEX"
      label = "person"
      idStrategy = "PRIMARY_KEY"
      idFields = ["name"]
      properties = ["name", "age"]
    }
  }
}
```

After (new `mappings` array, backward-compatible):
```hocon
sink {
  HugeGraph {
    mappings = [
      { type = "VERTEX"; label = "person"; idStrategy = "PRIMARY_KEY"; idFields = ["name"]; properties = ["name","age"] },
      { type = "EDGE"; label = "knows"; sourceConfig = { label = "person"; idFields = ["src"] }; targetConfig = { label = "person"; idFields = ["tgt"] } }
    ]
  }
}
```

**New source connector:**
```hocon
source {
[...]

PR #11661, opened 2026-08-05
[Improve][Connector-V2][File] Parse the Hadoop configuration once per subtask instead of once per output file

loustler:
## What changed

`AbstractWriteStrategy#getConfiguration` builds a Hadoop `Configuration` from scratch on every
call. It is called **once per output file**, so a subtask that rotates files re-does the work for
every file it writes.

The expensive part is not the object — it is the resource load. The first property access on a new
`Configuration` parses `core-default.xml` and the other default resources. `HadoopConf#toConfiguration`
touches a property immediately (`setBoolean`), so every call pays the parse.

This caches the parsed instance per write strategy and hands out copies. Hadoop's copy constructor
clones the already-loaded `Properties` instead of re-reading the XML, so callers still get a fresh,
independently mutable `Configuration` — which matters, because some of them mutate it
(`ParquetWriteStrategy#init` sets `AvroWriteSupport.WRITE_FIXED_AS_INT96` on the instance it
receives).

```java
 @Override
 public Configuration getConfiguration(HadoopConf hadoopConf) {
+    // The cache is only valid for the HadoopConf this strategy was initialised with.
+    if (hadoopConf != this.hadoopConf) {
+        return buildConfiguration(hadoopConf);
+    }
+    if (parsedConfiguration == null) {
+        parsedConfiguration = buildConfiguration(hadoopConf);
+    }
+    return new Configuration(parsedConfiguration);
+}
+
+private Configuration buildConfiguration(HadoopConf hadoopConf) {
     Configuration configuration = hadoopConf.toConfiguration();
     this.hadoopConf.setExtraOptionsForConfiguration(configuration);
     return configuration;
 }
```

Affected call sites, all per output file or per init — `getConfiguration` is never on a per-row path:

| call site | frequency |
|---|---|
| `ParquetWriteStrategy:190` (`HadoopOutputFile.fromPath`) | per output file |
| `OrcWriteStrategy:149` (`OrcFile.writerOptions`) | per output file |
| `ParquetWriteStrategy:101` (`init`, feeds `AvroSchemaConverter`) | once per subtask |

## Measured

`ConfigBench` (source in the collapsed block at the end) against the classpath a Zeta worker actually loads — the shipped
`lib/seatunnel-hadoop3-*-uber.jar`, which bundles `core-default.xml` (152 KB, 462 properties) plus
every transitive dependency. JDK 8 `1.8.0_502`, `-Xms2g -Xmx2g`, 3 s warmup, 7 × 2 s rounds,
medians. `TOCONF` is the current per-file path; `COPY` is what this PR does.

| mode | 1 thread ops/s | 8 threads ops/s | CPU ms/op | allocated B/op |
|---|---:|---:|---:|---:|
| `new Configuration()` alone | 1,432,625 | 1,481,705 | 0.0010 | 288 |
| **`toConfiguration()` — before** | 630 | 3,986 | **1.652** | **1,591,454** |
| **`new Configuration(cached)` — after** | 71,699 | 68,931 | **0.033** | **39,680** |
| reuse one instance (floor, not proposed) | 19,575,608 | — | 0.0001 | 24 |

So per output file: **1.65 ms → 0.033 ms of CPU** and **1,591,454 B → 39,680 B (−97.5 %)**. At one
thread this work is CPU-bound, so wall tracks CPU closely (1/630 s = 1.59 ms/op).

Two things that table shows which are worth stating plainly:

- **The parse is lazy.** `new Configuration()` on its own is 0.4 µs and allocates 288 B. All of the
  cost is the first property touch. That is why the fix is "parse once", not "allocate less".
- **`Configuration` construction does not scale across threads.** The two constructor-bound modes
  come out at 0.96× and 1.03× on 8 threads. Both constructors take a monitor on `Configuration.class`
  to register into a static `WeakHashMap` (`monitorenter` at offset 97 in `Configuration(boolean)`,
  238 in the copy constructor), which caps construction at ~69k/s per JVM. At per-file frequency
  that is irrelevant, but the copy is not free and I would rather say so than imply otherwise.

On the real write path (`ParquetWriteStrategy#write`, 9-column schema, 200k rows, 1 thread, JDK 8,
7 rounds), patched/unpatched alternated across separate JVMs, **two independent A/B pairs**:

`batch_size=2000` → 100 output files:

| metric | before | after | pair 1 | pair 2 |
|---|---:|---:|---:|---:|
| allocation / row | 4,817.7 / 4,850.9 B | 3,814.9 / 3,782.9 B | **−20.8 %** | **−22.0 %** |
| CPU / row | 5,973.2 / 5,931.6 ns | 3,929.4 / 3,917.4 ns | **−34.2 %** | **−34.0 %** |
| wall / row | 25,925 / 26,211 ns | 24,004 / 24,164 ns | **−7.4 %** | **−7.8 %** |
| throughput | 38,573 / 38,152 rows/s | 41,659 / 41,383 rows/s | **+8.0 %** | **+8.5 %** |

No rotation → 1 output file (the control):

| metric | before | after | delta |
|---|---:|---:|---:|
| allocation / row | 2,726.8 B | 2,716.4 B | **−10.4 B** |

The control is what establishes that this is a per-file cost and not a per-row one: with a single
output file the saving collapses from ~1,000 B/row to ~10 B/row. (In pair 2 the *unpatched*
no-rotation baseline drifted to 2,918.7 B/row while the patched value came out at 2,716.4 B/row in
both pairs, identical to 0.1 B. I am quoting pair 1 for the control because pair 2's baseline is the
outlier, not because it is the friendlier number.)

**Every metric came out larger in situ than the microbenchmark predicts, and I have not isolated
why.** Comparing like with like — isolated saving is the before/after delta of the two bold
`ConfigBench` rows, in-situ saving is the per-row delta × 2,000 rows per file:

| saved per output file | isolated (ConfigBench) | in situ, pair 1 / pair 2 | ratio |
|---|---:|---:|---:|
| allocation | 1.552 MB | 2.006 / 2.136 MB | 1.29× / 1.38× |
| CPU | 1.619 ms | 4.088 / 4.028 ms | 2.5× / 2.5× |
| wall | 1.573 ms | 3.841 / 4.093 ms | 2.4× / 2.6× |

CPU and wall agree with each other tightly (~2.5×) and disagree with allocation (~1.3×), and neither
ratio is an integer — so this is not `getConfiguration` being called more times per output file than
I think it is. Something about handing `HadoopOutputFile`/`ParquetWriter` a freshly-parsed
`Configuration` rather than a copy costs more downstream than the construction itself. I did not
isolate it, so it is an observation, not a claim, and every figure quoted elsewhere in this PR is the
smaller isolated one.

## What this does *not* claim

**The +8 % is a sink-loop number, not a job number.** It is measured with the write loop as the only
work in the process. A real pipeline also reads, deserialises, and often blocks on the source: in the
production run that started this investigation the writers were parked 56.7 % of the time, and no
sink-side saving can show up through that. So: +8 % on the write path, and no end-to-end claim.

**`Configuration` is a small share of per-file overhead, not a dominant one.** At `batch_size=2000`
the harness costs 25,925 ns/row — **51.9 ms of wall time per output file** — against 2,589 ns/row
with no rotation. Of that 51.9 ms this recovers 1.57 ms by the isolated measurement (3.0 %) or
3.8–4.1 ms as measured in situ (7.4–7.9 %). The rest is `ParquetWriter` construction, opening the
file, and writing the footer on close, and this PR does not touch any of it. An earlier internal
estimate put the `Configuration` cost at "~48 ms per
file"; that figure was back-computed from exactly this throughput gap and silently attributed the
whole gap to `Configuration`. Direct measurement shows it was wrong by roughly 30×. Correcting it
here rather than letting a reviewer find it.

Jobs writing many small files (small `batch_size`, high partition cardinality) get proportionally
more of this; a job writing one large file per subtask gets ~10 B/row, which is nothing.

Also not measured: any effect on HDFS/S3/OSS specifically — the harness writes to a local
filesystem, and JDK 8 on this host is x86_64 under Rosetta, so the ratios are meaningful and the
absolute numbers are not.

## Why copies and not one shared instance

Handing every caller the same cached instance would be faster still — 0.033 ms → 0.0001 ms. It was
not done, because `getConfiguration` currently guarantees a fresh instance and at least one caller
relies on that: `ParquetWriteStrategy#init` mutates the returned `Configuration`. Sharing it would
leak that mutation into the per-file configuration. Weighed against ~48 ms of other per-file work,
saving a further 33 µs is not worth introducing shared mutable state into a class that is not
thread-safe today.

The `hadoopConf != this.hadoopConf` guard exists because `getConfiguration(HadoopConf)` is a public
interface method that takes its conf as a parameter. Every current caller passes the field, so the
guard is defensive — but the alternative is a method that silently ignores its own argument.

## Behaviour

- The returned object is still a distinct, freely mutable `Configuration` on every call.
- It still carries everything `toConfiguration()` and `setExtraOptionsForConfiguration()` set, plus
  the parsed default resources.
- The field is `transient`: `WriteStrategy extends Serializable` and `Configuration` is not.

Added `AbstractWriteStrategyConfigurationTest` covering exactly those three points — independent
instances, mutation isolation across calls, and that a *different* `HadoopConf` is not served from
the cache.

## Verification

- `mvn test -pl seatunnel-connectors-v2/connector-file/connector-file-base` on JDK 8 `1.8.0_502`:
  **240 tests, 0 failures, 0 errors** (237 before this PR + the 3 added here). `spotless:check` runs
  in that build and passes.
- Grepped for other overrides and callers of `getConfiguration`: nothing outside
  `connector-file-base` overrides `AbstractWriteStrategy#getConfiguration`, and the three call sites
  in the table above are all of them.

## Reproducing the benchmark

```bash
# uber jar out of any built dist; slf4j-api is needed because the uber jar deliberately
# excludes slf4j (without it: NoClassDefFoundError org/slf4j/LoggerFactory)
CP=lib/seatunnel-hadoop3-3.4.3-uber.jar:starter/logging/slf4j-api-1.7.36.jar
javac -cp "$CP" -d out ConfigBench.java
for m in NEW_ONLY NEW_GET TOCONF COPY REUSE; do
  java -Xms2g -Xmx2g -cp "$CP:out" -Dcb.mode=$m -Dcb.threads=1 ConfigBench | grep CB_RESULT
done
```

The write-path numbers come from a throwaway JUnit harness that drives `ParquetWriteStrategy#write`
directly and reports `getThreadAllocatedBytes` / `getCurrentThreadCpuTime` per row. It is not part of
this PR — it is a measurement tool, not a test, and it would only rot in the tree. Happy to attach it
if a reviewer wants to re-run the A/B.

<details>
<summary><code>ConfigBench.java</code> — the full benchmark source</summary>

```java
import com.sun.management.ThreadMXBean;
import java.lang.management.ManagementFactory;
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.util.concurrent.CountDownLatch;
import org.apache.hadoop.conf.Configuration;

/**
 * Measure what a per-output-file Hadoop {@link Configuration} construction actually costs, so
 * the SeaTunnel file-sink fix can be argued from a number rather than from a derived estimate.
 *
 * Run against the SHIPPED uber jar (lib/seatunnel-hadoop3-*-uber.jar out of the dist tarball),
 * not against a hand-assembled classpath: the uber jar is what production loads, it bundles
 * core-default.xml and every transitive dependency, and its relocations are part of the thing
 * being measured.
 *
 *   javac -cp uber.jar -d out ConfigBench.java
 *   java -cp uber.jar:out -Dcb.mode=TOCONF -Dcb.threads=1 ConfigBench
 *
 * Modes:
 *   NEW_ONLY  new Configuration()                      - shows the XML parse is lazy
 *   NEW_GET   new Configuration() + get()              - forces the parse
 *   TOCONF    replica of HadoopConf#toConfiguration()  - what SeaTunnel does per output file
 *   COPY      new Configuration(warmSource) + get()    - the proposed fix's residual cost
 *   REUSE     hand back one prebuilt instance          - the zero-cost floor
 *
 * Why both 1 and 8 threads: BOTH constructors take a monitor on Configuration.class to register
 * into a static WeakHashMap (verified in hadoop-common 3.4.3 bytecode: monitorenter at offset 97
 * in Configuration(boolean), 238 in the copy constructor). Per output file that should be
 * irrelevant; the 8-thread column is there to confirm it, not to assume it.
 */
public class ConfigBench {

    /** Written to on every op so the JIT cannot fold the work away. */
    static volatile Object blackhole;

    enum Mode {
        NEW_ONLY,
        NEW_GET,
        TOCONF,
        COPY,
        REUSE
    }

    // The six properties HadoopConf#toConfiguration() sets, with the keys SeaTunnel uses for
    // plain HDFS. Values are irrelevant to the cost; the count and the first-touch are not.
    static Configuration toConfiguration() {
        Configuration c = new Configuration();
        c.setBoolean("parquet.avro.readInt96AsFixed", true);
        c.setBoolean("parquet.avro.add-list-element-records", false);
        c.setBoolean("parquet.avro.write-old-list-structure", true);
        c.setBoolean("fs.hdfs.impl.disable.cache", true);
        c.set("fs.defaultFS", "hdfs://localhost:9000");
        c.set("fs.hdfs.impl", "org.apache.hadoop.hdfs.DistributedFileSystem");
        return c;
    }

    static Object op(Mode mode, Configuration warm, Configuration reuse) {
        switch (mode) {
            case NEW_ONLY:
                return new Configuration();
            case NEW_GET: {
                Configuration c = new Configuration();
                return c.get("fs.defaultFS");
            }
            case TOCONF:
                return toConfiguration();
            case COPY: {
                Configuration c = new Configuration(warm);
                return c.get("fs.defaultFS");
            }
            case REUSE:
                return reuse.get("fs.defaultFS");
            default:
                throw new IllegalStateException();
        }
    }

    public static void main(String[] args) throws Exception {
        Mode mode = Mode.valueOf(System.getProperty("cb.mode", "TOCONF"));
        int threads = Integer.getInteger("cb.threads", 1);
        int rounds = Integer.getInteger("cb.rounds", 7);
        long roundMs = Long.getLong("cb.roundms", 2000L);
        long warmupMs = Long.getLong("cb.warmupms", 3000L);
        String label = System.getProperty("cb.label", mode + "-t" + threads);

        // A source instance whose resources are already parsed, i.e. what a cached
        // Configuration in AbstractWriteStrategy would look like at steady state.
        final Configuration warm = toConfiguration();
        warm.get("fs.defaultFS");
        final Configuration reuse = toConfiguration();
        reuse.get("fs.defaultFS");

        System.out.printf(
                "CB_START label=%s mode=%s threads=%d rounds=%d roundMs=%d jdk=%s%n",
                label, mode, threads, rounds, roundMs, System.getProperty("java.version"));

        // ---- warmup ----
        long wEnd = System.nanoTime() + warmupMs * 1_000_000L;
        long wOps = 0;
        while (System.nanoTime() < wEnd) {
            blackhole = op(mode, warm, reuse);
            wOps++;
        }
        System.out.printf("CB_WARMUP label=%s ops=%d%n", label, wOps);

        ThreadMXBean tmx = (ThreadMXBean) ManagementFactory.getThreadMXBean();
        List<Double> opsPerSec = new ArrayList<>();
        List<Double> nsPerOp = new ArrayList<>();
        List<Double> bytesPerOp = new ArrayList<>();

        for (int r = 0; r < rounds; r++) {
            final CountDownLatch go = new CountDownLatch(1);
            final long[] ops = new long[threads];
            final long[] alloc = new long[threads];
            final long[] cpu = new long[threads];
            Thread[] ts = new Thread[threads];
            for (int i = 0; i < threads; i++) {
                final int idx = i;
                ts[i] =
                        new Thread(
                                () -> {
                                    try {
                                        go.await();
                                    } catch (InterruptedException e) {
                                        return;
                                    }
                                    long id = Thread.currentThread().getId();
                                    long a0 = tmx.getThreadAllocatedBytes(id);
                                    long c0 = tmx.getCurrentThreadCpuTime();
                                    long end = System.nanoTime() + roundMs * 1_000_000L;
                                    long n = 0;
                                    Object local = null;
                                    while (System.nanoTime() < end) {
                                        local = op(mode, warm, reuse);
                                        n++;
                                    }
                                    blackhole = local;
                                    ops[idx] = n;
                                    alloc[idx] = tmx.getThreadAllocatedBytes(id) - a0;
                                    cpu[idx] = tmx.getCurrentThreadCpuTime() - c0;
                                },
                                "cb-" + i);
                ts[i].start();
            }
            long t0 = System.nanoTime();
            go.countDown();
            for (Thread t : ts) {
                t.join();
            }
            long wall = System.nanoTime() - t0;

            long totalOps = 0, totalAlloc = 0, totalCpu = 0;
            for (int i = 0; i < threads; i++) {
                totalOps += ops[i];
                totalAlloc += alloc[i];
                totalCpu += cpu[i];
            }
            double ops_s = totalOps / (wall / 1e9);
            double ns_op = (double) totalCpu / totalOps;
            double b_op = (double) totalAlloc / totalOps;
            opsPerSec.add(ops_s);
            nsPerOp.add(ns_op);
            bytesPerOp.add(b_op);
            System.out.printf(
                    "CB_ROUND label=%s r=%d ops=%d ops_s=%.1f cpu_ns_op=%.1f alloc_b_op=%.1f%n",
                    label, r, totalOps, ops_s, ns_op, b_op);
        }

        System.out.printf(
                "CB_RESULT label=%s mode=%s threads=%d ops_s_median=%.1f cpu_ns_op_median=%.1f "
                        + "alloc_b_op_median=%.1f ms_per_op_median=%.4f%n",
                label,
                mode,
                threads,
                median(opsPerSec),
                median(nsPerOp),
                median(bytesPerOp),
                median(nsPerOp) / 1e6);
    }

    static double median(List<Double> xs) {
        double[] a = new double[xs.size()];
        for (int i = 0; i < a.length; i++) {
            a[i] = xs.get(i);
        }
        Arrays.sort(a);
        int n = a.length;
        return n % 2 == 1 ? a[n / 2] : (a[n / 2 - 1] + a[n / 2]) / 2.0;
    }
}
```

</details>

### DIFF
--- a/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/AbstractWriteStrategy.java
+++ b/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/AbstractWriteStrategy.java
@@ -193,9 +193,21 @@ public abstract class AbstractWriteStrategy<T> implements WriteStrategy<T> {
         return new Configuration(parsedConfiguration);
     }
 
+    /**
+     * Does the full, expensive work — the resource parse plus the extra options — for one
+     * HadoopConf. This is the thing {@link #getConfiguration(HadoopConf)} caches, so it must stay
+     * free of any per-output-file state.
+     *
+     * <p>Both steps read the same {@code hadoopConf}. That matters because {@link
+     * HadoopConf#setExtraOptionsForConfiguration} does not only copy {@code extraOptions} in: it
+     * also decides which keys an {@code hdfs-site.xml} resource may not overwrite, and those keys
+     * are derived from the conf's own {@code getSchema()}, which every filesystem subclass
+     * overrides. Taking them from a different conf would let that resource overwrite the very
+     * properties {@code unsetUnwantedOverwritingProps} exists to protect.
+     */
     private Configuration buildConfiguration(HadoopConf hadoopConf) {
         Configuration configuration = hadoopConf.toConfiguration();
-        this.hadoopConf.setExtraOptionsForConfiguration(configuration);
+        hadoopConf.setExtraOptionsForConfiguration(configuration);
         return configuration;
     }
 
--- a/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/AbstractWriteStrategy.java
+++ b/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/AbstractWriteStrategy.java
@@ -86,6 +86,9 @@ public abstract class AbstractWriteStrategy<T> implements WriteStrategy<T> {
     protected int subTaskIndex;
     protected HadoopConf hadoopConf;
     protected HadoopFileSystemProxy hadoopFileSystemProxy;
+    /** Template whose resources are already parsed; see {@link #getConfiguration(HadoopConf)}. */
+    private transient Configuration parsedConfiguration;
+
     protected String transactionId;
     /** The uuid prefix to make sure same job different file sink will not conflict. */
     protected String uuidPrefix;
@@ -167,11 +170,30 @@ public abstract class AbstractWriteStrategy<T> implements WriteStrategy<T> {
     /**
      * use hadoop conf generate hadoop configuration
      *
+     * <p>Callers get a fresh, independently mutable Configuration, because some of them mutate it
+     * ({@code ParquetWriteStrategy#init} sets {@code AvroWriteSupport.WRITE_FIXED_AS_INT96}). The
+     * expensive part is not the object but the resource load: the first property access on a new
+     * Configuration parses {@code core-default.xml} and friends. Since this is called once per
+     * output file, that parse used to be repeated for every file the subtask wrote. So parse once
+     * and hand out copies — Hadoop's copy constructor clones the already-loaded properties instead
+     * of re-reading the XML.
+     *
      * @param hadoopConf hadoop conf
      * @return Configuration
      */
     @Override
     public Configuration getConfiguration(HadoopConf hadoopConf) {
+        // The cache is only valid for the HadoopConf this strategy was initialised with.
+        if (hadoopConf != this.hadoopConf) {
+            return buildConfiguration(hadoopConf);
+        }
+        if (parsedConfiguration == null) {
+            parsedConfiguration = buildConfiguration(hadoopConf);
+        }
+        return new Configuration(parsedConfiguration);
+    }
+
+    private Configuration buildConfiguration(HadoopConf hadoopConf) {
         Configuration configuration = hadoopConf.toConfiguration();
         this.hadoopConf.setExtraOptionsForConfiguration(configuration);
         return configuration;
--- a/seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/writer/AbstractWriteStrategyConfigurationTest.java
+++ b/seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/writer/AbstractWriteStrategyConfigurationTest.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.connectors.seatunnel.file.writer;
+
+import org.apache.seatunnel.api.configuration.ReadonlyConfig;
+import org.apache.seatunnel.api.table.catalog.CatalogTableUtil;
+import org.apache.seatunnel.api.table.type.BasicType;
+import org.apache.seatunnel.api.table.type.SeaTunnelDataType;
+import org.apache.seatunnel.api.table.type.SeaTunnelRowType;
+import org.apache.seatunnel.connectors.seatunnel.file.config.FileFormat;
+import org.apache.seatunnel.connectors.seatunnel.file.sink.config.FileSinkConfig;
+import org.apache.seatunnel.connectors.seatunnel.file.sink.writer.ParquetWriteStrategy;
+import org.apache.seatunnel.connectors.seatunnel.file.util.LocalFileSystemConf;
+
+import org.apache.hadoop.conf.Configuration;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.FS_DEFAULT_NAME_DEFAULT;
+
+/**
+ * getConfiguration is called once per output file, so it caches the parsed resources. These tests
+ * pin the part callers depend on: every call still yields a Configuration they may freely mutate.
+ */
+public class AbstractWriteStrategyConfigurationTest {
+
+    private static ParquetWriteStrategy strategy(LocalFileSystemConf.LocalConf hadoopConf) {
+        Map<String, Object> writeConfig = new HashMap<>();
+        writeConfig.put("tmp_path", "file:///tmp/seatunnel/conf-cache/tmp");
+        writeConfig.put("path", "file:///tmp/seatunnel/conf-cache");
+        writeConfig.put("file_format_type", FileFormat.PARQUET.name());
+
+        SeaTunnelRowType rowType =
+                new SeaTunnelRowType(
+                        new String[] {"f1_text"}, new SeaTunnelDataType[] {BasicType.STRING_TYPE});
+        ParquetWriteStrategy strategy =
+                new ParquetWriteStrategy(
+                        new FileSinkConfig(ReadonlyConfig.fromMap(writeConfig), rowType));
+        strategy.setCatalogTable(
+                CatalogTableUtil.getCatalogTable("test", null, null, "test", rowType));
+        strategy.init(hadoopConf, "job1", "job1", 0);
+        return strategy;
+    }
+
+    @Test
+    public void testEachCallReturnsAnIndependentlyMutableConfiguration() {
+        LocalFileSystemConf.LocalConf hadoopConf =
+                new LocalFileSystemConf.LocalConf(FS_DEFAULT_NAME_DEFAULT);
+        ParquetWriteStrategy strategy = strategy(hadoopConf);
+
+        Configuration first = strategy.getConfiguration(hadoopConf);
+        Configuration second = strategy.getConfiguration(hadoopConf);
+
+        Assertions.assertNotSame(first, second);
+
+        // Callers do mutate what they are handed - ParquetWriteStrategy#init sets
+        // AvroWriteSupport.WRITE_FIXED_AS_INT96 on it - so a mutation must not be visible
+        // to any later caller.
+        first.set("seatunnel.test.marker", "written-by-first-caller");
+        Assertions.assertNull(strategy.getConfiguration(hadoopConf).get("seatunnel.test.marker"));
+        Assertions.assertNull(second.get("seatunnel.test.marker"));
+    }
+
+    @Test
+    public void testConfigurationCarriesHadoopConfValues() {
+        LocalFileSystemConf.LocalConf hadoopConf =
+                new LocalFileSystemConf.LocalConf(FS_DEFAULT_NAME_DEFAULT);
+        ParquetWriteStrategy strategy = strategy(hadoopConf);
+
+        // Same assertions on the first and a later call: the cached copy must not lose anything
+        // that toConfiguration()/setExtraOptionsForConfiguration() put there.
+        for (int call = 0; call < 3; call++) {
+            Configuration configuration = strategy.getConfiguration(hadoopConf);
+            Assertions.assertEquals(
+                    FS_DEFAULT_NAME_DEFAULT, configuration.get("fs.defaultFS"), "call " + call);
+            Assertions.assertEquals(
+                    "org.apache.hadoop.fs.LocalFileSystem",
+                    configuration.get("fs.file.impl"),
+                    "call " + call);
+            Assertions.assertTrue(
+                    configuration.getBoolean("fs.file.impl.disable.cache", false), "call " + call);
+            // A core-default.xml value, i.e. proof the copy carries the parsed resources and not
+            // just the six properties toConfiguration() sets explicitly.
+            Assertions.assertNotNull(configuration.get("io.file.buffer.size"), "call " + call);
+        }
+    }
+
+    @Test
+    public void testForeignHadoopConfIsNotServedFromTheCache() {
+        LocalFileSystemConf.LocalConf hadoopConf =
+                new LocalFileSystemConf.LocalConf(FS_DEFAULT_NAME_DEFAULT);
+        ParquetWriteStrategy strategy = strategy(hadoopConf);
+        // Prime the cache.
+        strategy.getConfiguration(hadoopConf);
+
+        LocalFileSystemConf.LocalConf other =
+                new LocalFileSystemConf.LocalConf("file:///tmp/seatunnel/other");
+        Assertions.assertEquals(
+                "file:///tmp/seatunnel/other",
+                strategy.getConfiguration(other).get("fs.defaultFS"));
+    }
+}
--- a/seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/writer/AbstractWriteStrategyConfigurationTest.java
+++ b/seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/org/apache/seatunnel/connectors/seatunnel/file/writer/AbstractWriteStrategyConfigurationTest.java
@@ -32,6 +32,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -117,5 +118,74 @@ public class AbstractWriteStrategyConfigurationTest {
         Assertions.assertEquals(
                 "file:///tmp/seatunnel/other",
                 strategy.getConfiguration(other).get("fs.defaultFS"));
+    }
+
+    /**
+     * The point of the cache is that the resource parse happens once. The assertions above prove
+     * the returned Configuration is correct, which would hold just as well if nothing were cached
+     * at all - so count the expensive builds directly.
+     */
+    @Test
+    public void testTheExpensiveBuildHappensOncePerHadoopConf() {
+        CountingConf hadoopConf = new CountingConf(FS_DEFAULT_NAME_DEFAULT);
+        ParquetWriteStrategy strategy = strategy(hadoopConf);
+
+        // Deliberately not asserting an absolute number here. Two independent builds happen during
+        // init - HadoopFileSystemProxy's constructor eagerly builds its own Configuration, and
+        // ParquetWriteStrategy#init calls getConfiguration - and that split is init's business, not
+        // this cache's. What this cache promises is that the count stops growing afterwards.
+        int afterInit = hadoopConf.toConfigurationCalls;
+        Assertions.assertTrue(afterInit >= 1, "init should have built at least one Configuration");
+
+        for (int i = 0; i < 5; i++) {
+            Assertions.assertNotNull(strategy.getConfiguration(hadoopConf));
+        }
+        Assertions.assertEquals(
+                afterInit,
+                hadoopConf.toConfigurationCalls,
+                "5 further calls must all be served from the cache");
+
+        // A foreign conf is built from itself and must not disturb the cached template.
+        CountingConf other = new CountingConf("file:///tmp/seatunnel/other");
+        strategy.getConfiguration(other);
+        Assertions.assertEquals(1, other.toConfigurationCalls);
+        Assertions.assertEquals(afterInit, hadoopConf.toConfigurationCalls);
+    }
+
+    /**
+     * A foreign HadoopConf must be configured entirely from itself, never from the strategy's own
+     * conf. This is easy to get wrong in a way that is hard to notice: the keys
+     * setExtraOptionsForConfiguration protects against an hdfs-site.xml overwrite are derived from
+     * getSchema(), and every filesystem subclass overrides that - so mixing the two confs would let
+     * that resource overwrite the properties the protection exists for.
+     */
+    @Test
+    public void testForeignHadoopConfGetsItsOwnExtraOptions() {
+        LocalFileSystemConf.LocalConf hadoopConf =
+                new LocalFileSystemConf.LocalConf(FS_DEFAULT_NAME_DEFAULT);
+        hadoopConf.setExtraOptions(Collections.singletonMap("seatunnel.test.owner", "strategy"));
+        ParquetWriteStrategy strategy = strategy(hadoopConf);
+
+        LocalFileSystemConf.LocalConf other =
+                new LocalFileSystemConf.LocalConf("file:///tmp/seatunnel/other");
+        other.setExtraOptions(Collections.singletonMap("seatunnel.test.owner", "foreign"));
+
+        Configuration configuration = strategy.getConfiguration(other);
+        Assertions.assertEquals("foreign", configuration.get("seatunnel.test.owner"));
+    }
+
+    /** Counts how many times the expensive build path actually ran. */
+    private static class CountingConf extends LocalFileSystemConf.LocalConf {
+        private int toConfigurationCalls;
+
+        private CountingConf(String hdfsNameKey) {
+            super(hdfsNameKey);
+        }
+
+        @Override
+        public Configuration toConfiguration() {
+            toConfigurationCalls++;
+            return super.toConfiguration();
+        }
     }
 }


DanielLeens (changes requested):

`AbstractWriteStrategy#getConfiguration(HadoopConf)` is called once per output file (`ParquetWriteStrategy#getOrCreateOutputStream`, `OrcWriteStrategy#getOrCreateOutputStream`, plus once at `ParquetWriteStrategy#init`), and every call rebuilt a Hadoop `Configuration` from scratch via `hadoopConf.toConfiguration()`. The first property access on a fresh `Configuration` parses `core-default.xml`/`core-site.xml` and any other registered resources — an expensive, allocation-heavy operation — so a subtask that rotates through many output files (small `batch_size`, high partition cardinality) re-paid that parse for every single file. This PR parses the Hadoop resources once per write-strategy instance (i.e., once per subtask) and hands out cheap independent copies afterward via Hadoop's `Configuration` copy constructor, which clones the already-loaded `Properties` instead of re-reading the XML.

# 1. Code Change Review

## 1.1 Core Logic Analysis

`seatunnel-connectors-v2/connector-file/connector-file-base/.../sink/writer/AbstractWriteStrategy.java`

Before:
```java
public Configuration getConfiguration(HadoopConf hadoopConf) {
    Configuration configuration = hadoopConf.toConfiguration();
    this.hadoopConf.setExtraOptionsForConfiguration(configuration);
    return configuration;
}
```

After:
```java
private transient Configuration parsedConfiguration;
...
public Configuration getConfiguration(HadoopConf hadoopConf) {
    if (hadoopConf != this.hadoopConf) {
        return buildConfiguration(hadoopConf);
    }
    if (parsedConfiguration == null) {
        parsedConfiguration = buildConfiguration(hadoopConf);
    }
    return new Configuration(parsedConfiguration);
}

private Configuration buildConfiguration(HadoopConf hadoopConf) {
    Configuration configuration = hadoopConf.toConfiguration();
    this.hadoopConf.setExtraOptionsForConfiguration(configuration);
    return configuration;
}
```

I traced the full call graph and lifecycle to check whether this caching is actually safe to introduce:

- **Who owns `this.hadoopConf`, and can it change under the cache?** `this.hadoopConf` is assigned exactly once, in `AbstractWriteStrategy#init(HadoopConf conf, ...)` (line 132), which is called exactly once per writer-strategy instance from `BaseFileSinkWriter`'s constructor. Nothing in the sink write path calls a setter on that `HadoopConf` instance afterward — I grepped every `HadoopConf` setter call site (`setHdfsSitePath`, `setKerberosPrincipal`, `setExtraOptions`, etc.) across `connector-file-*`, and all of them run inside the config-parsing factories (`HdfsFileSinkFactory`, `S3HadoopConf`, `OssHadoopConf`, ...) *before* the `HadoopConf` is ever handed to a `WriteStrategy`. So the object the cache is keyed on is effectively immutable for the lifetime of the writer, which is the precondition this caching design needs.
- **Does one write-strategy instance ever serve more than one `HadoopConf`?** No. `BaseMultipleTableFileSink.createWriteStrategy()` constructs a brand-new `WriteStrategy` for every `createWriter()`/`restoreWriter()` call, and multi-table sinks (`SupportMultiTableSinkWriter`) get one such writer *per table*, each with its own `AbstractWriteStrategy` instance and therefore its own `parsedConfiguration` field — there is no static/shared cache, so even tables that happen to share the identical `HadoopConf` reference (same bucket/credentials, different `path`) don't leak state into each other. This directly answers the multi-table/dynamic-partition concern: per-file/per-partition *paths* live in `FileSinkConfig`, not in the cached `Configuration`, and are untouched by this change.
- **Is the `hadoopConf != this.hadoopConf` reference-equality guard ever actually exercised?** I grepped all three call sites (`ParquetWriteStrategy:101`, `:190`, `OrcWriteStrategy:149`) — every one passes the inherited `hadoopConf` field itself, never a different `HadoopConf`. So today the guard is dead code / a pure defensive fallback, exactly as the PR description states. It is safe, just currently unreachable.
- **Is the returned `Configuration` really independently mutable, matching the pre-PR contract?** Yes — I confirmed `ParquetWriteStrategy#init` mutates the `Configuration` it receives (`configuration.set(AvroWriteSupport.WRITE_FIXED_AS_INT96, ...)`), so the cached template must never be handed out directly. It isn't: every call returns `new Configuration(parsedConfiguration)`, and Hadoop's copy constructor clones `properties`/`overlay` (via `Properties.clone()`) and rebuilds its own `resources`/`finalParameters` collections, so a mutation on one caller's copy is invisible to the cached template and to every other caller. The new test class exercises exactly this (`testEachCallReturnsAnIndependentlyMutableConfiguration`).
- **Concurrency**: `getConfiguration` is only reached from `write()`-path methods that already assume single-threaded access per writer instance (e.g., `beingWrittenWriter`/`beingWrittenFile` maps are accessed without synchronization elsewhere in this class). No new cross-thread access pattern is introduced by adding an unsynchronized lazy-init field.
- **Serialization**: `WriteStrategy extends Serializable`, and Hadoop's `Configuration` does not implement `Serializable`. Marking `parsedConfiguration` as `transient` is correct and necessary — if this object graph is ever serialized, the field simply comes back `null` and is lazily rebuilt on next use, with no behavior change.

Runtime path (per output file, after the fix):

```mermaid
flowchart TD
    A[write triggers new output file] --> B[getOrCreateOutputStream]
    B --> C{parsedConfiguration cached?}
    C -- no --> D[buildConfiguration: toConfiguration + setExtraOptionsForConfiguration<br/>parses core-default.xml etc, cache result]
    C -- yes --> E[new Configuration(parsedConfiguration)<br/>clones already-loaded Properties, no XML re-parse]
    D --> F[return fresh independently-mutable copy]
    E --> F
    F --> G[HadoopOutputFile.fromPath / OrcFile.writerOptions]
```

One thing worth calling out explicitly rather than leaving implicit: `buildConfiguration(HadoopConf hadoopConf)` builds the raw configuration from the **parameter** (`hadoopConf.toConfiguration()`) but applies extra options from **`this.hadoopConf`** (`this.hadoopConf.setExtraOptionsForConfiguration(...)`), not from the parameter. This asymmetry already existed in the pre-PR code (it's a straight extraction, not a new behavior), and today it's harmless because the only path that ever reaches `buildConfiguration` with a foreign `hadoopConf` is unreachable in production. But now that the "foreign conf" branch has been given an explicit, named guard (implying it's meant to be a real, supported code path for future callers), this inconsistency becomes more likely to bite silently if a future caller ever does pass a different `HadoopConf` — see Issue 1 below.

## 1.2 Compatibility Impact

**Fully compatible.** No config `Option`, default value, public API, or wire/serialization format is touched. `parsedConfiguration` is a purely in-memory, `transient` performance cache that is not part of `FileSinkState` or any other persisted/checkpointed structure, so there is no historical-version or restore-path compatibility concern. The `getConfiguration(HadoopConf)` method signature and its documented contract (fresh, independently mutable `Configuration` per call) are preserved byte-for-byte from the caller's point of view.

The mechanism is sound: Hadoop's `Configuration(Configuration other)` copy constructor clones the already-loaded `Properties`/`overlay` and rebuilds its own `resources` list rather than re-registering and re-parsing `core-default.xml`/`core-site.xml`/any `hdfs-site.xml` added via `setExtraOptionsForConfiguration`. Since the number of `Configuration` object constructions per output file is unchanged (exactly one `new Configuration(...)` call before and after), this does not introduce new contention on the static lock `Configuration`'s constructors take to register into their internal `WeakHashMap` — if anything, average hold time per construction goes down since there's no first-touch XML parse. The PR author's own benchmark numbers (isolated ConfigBench + in-situ `ParquetWriteStrategy#write` A/B) are consistent with this mechanism and are unusually rigorous for a performance PR — including a self-correction of an internal "~48ms/file" estimate that the numbers didn't support, and an explicit call-out of an unexplained ~2.5x gap between isolated and in-situ CPU/wall savings that the author flagged as unresolved rather than papering over. I did not attempt to independently reproduce the benchmark numbers (out of scope for a source-level review), but the described methodology (uber jar classpath, JMX per-thread allocation/CPU counters, median of multiple rounds) is appropriate for this kind of measurement.

Memory-safety-wise, the cached template holds no per-file state (no open file handles, no file-specific paths — those live in `FileSinkConfig`/the `Path` object passed separately to `HadoopOutputFile.fromPath`), so there is no risk of a stale S3/HDFS credential or path leaking across files within a subtask; the cache only ever reflects the `HadoopConf` the strategy was `init()`-ed with, once.

No new exception paths are introduced. `getConfiguration` cannot throw anything it couldn't already throw (same underlying `hadoopConf.toConfiguration()` / `setExtraOptionsForConfiguration()` calls, just deferred to first use and cached).

**Issue 1** — Location: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/AbstractWriteStrategy.java:196-200` (`buildConfiguration`)
- Problem: `buildConfiguration(HadoopConf hadoopConf)` builds the base `Configuration` from the `hadoopConf` parameter but applies `setExtraOptionsForConfiguration` from the instance field `this.hadoopConf`, not from the parameter. This pre-existing asymmetry (present before this PR too) is currently harmless because every real caller passes `this.hadoopConf` itself, but the PR newly formalizes a `hadoopConf != this.hadoopConf` "foreign conf" branch as an explicit, seemingly-supported code path. If a future caller (or a new `WriteStrategy` subclass) legitimately passes a different `HadoopConf`, it will silently get *this strategy's own* extra options / hdfs-site resource applied to *its* base configuration instead of its own — a subtle, hard-to-notice bug.
- Potential risk: Low today (dead code path), but the risk surface grows as soon as anyone relies on the "foreign conf" guard for its apparent purpose.
- Best improvement: Option A — use the parameter consistently: `hadoopConf.setExtraOptionsForConfiguration(configuration)` inside `buildConfiguration`, which is both what the guard implies it should do and matches what `getConfiguration`'s Javadoc contract promises for any `HadoopConf` passed in. Option B — if `this.hadoopConf` is intentionally always meant to be authoritative, drop the parameter from `buildConfiguration` (use `this.hadoopConf` throughout) and simplify the `getConfiguration` guard/Javadoc to make that explicit instead of implying general per-parameter support.
- Severity: Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

The new Javadoc on `getConfiguration` is thorough and explains the "why" (lazy resource parsing, mutation-safety requirement) rather than restating the code, and the `parsedConfiguration` field has an explanatory comment — both meet this repo's documentation bar. One gap: the newly extracted private method `buildConfiguration(HadoopConf hadoopConf)` has no comment of its own. It is short and its behavior is covered by `getConfiguration`'s Javadoc, so this is a minor readability nit rather than something obscuring intent, but per the "every new non-trivial method should carry an explanatory comment" standard, a one-line comment (e.g., noting that this always does the full parse/apply, and is the thing being cached) would close the loop.

`AbstractWriteStrategyConfigurationTest` (121 new lines) is well targeted at exactly the three properties that matter for this change's correctness: (1) independent, mutation-isolated copies across repeated calls (`testEachCallReturnsAnIndependentlyMutableConfiguration`), (2) the cached copy still carries every value `toConfiguration()`/`setExtraOptionsForConfiguration()` set, including a `core-default.xml`-sourced value (`io.file.buffer.size`) as proof the copy isn't just replaying the six explicit properties (`testConfigurationCarriesHadoopConfValues`), and (3) a different `HadoopConf` instance bypasses the cache correctly (`testForeignHadoopConfIsNotServedFromTheCache`). This is good regression coverage for the specific risk this PR introduces (stale/shared mutable state across output files).

**Stability rating: Stable.** All three tests are fully deterministic: no `Thread.sleep`/hard waits, no wall-clock or network-timing dependencies, no container startup ordering, no shared/static mutable state between tests (`strategy(...)` constructs a fresh `ParquetWriteStrategy` + fresh `LocalConf` per test), no floating-point or unnormalized string comparisons, and no order dependency between the three `@Test` methods. The `file:///tmp/seatunnel/conf-cache*` paths follow the exact same pattern already used by other tests in this module (e.g. `ParquetWriteStrategyTest`), and since these tests never call `write()`, no actual file I/O happens beyond `HadoopFileSystemProxy`'s local-filesystem handle initialization — no risk of leftover files or cross-test collisions.

Coverage gap (non-blocking): there's no test asserting the "only-once" parsing claim itself (e.g., that `buildConfiguration`/the expensive parse path is invoked exactly once across N `getConfiguration()` calls) — the current tests prove correctness of the returned `Configuration`, not that the caching actually happens. This is more of a "prove the performance claim in a unit test" nice-to-have than a correctness gap, since correctness is what matters for merge-worthiness.

## 2.3 Documentation Updates

Not applicable. This is a pure internal performance optimization — no config `Option`, default, CLI behavior, or user-facing example changes, so no `docs/en`/`docs/zh` updates are required.

# 3. Architectural Soundness

Precise, long-term solution. The PR correctly identifies that the "shared instance" approach (fastest but unsafe, since callers mutate what they receive) is not appropriate here, and instead chooses "parse once, copy per call," which preserves the exact mutation-isolation contract existing callers depend on while removing the redundant XML parsing. The PR description explicitly reasons through and rejects the unsafe alternative, which is a good sign of design care.

## 3.2 Maintainability

The change is small, self-contained, and the new private `buildConfiguration` helper makes the "build vs. serve-from-cache" split easy to follow. The reference-equality cache-invalidation guard is easy to reason about today because it's simple, even though (per Issue 1) its semantics under a hypothetical future "foreign conf" caller deserve a closer look.

## 3.3 Extensibility

Neutral. This doesn't add or remove any extension points; `getConfiguration` keeps the same public contract for future `WriteStrategy` subclasses, with the caveat in Issue 1 about the "foreign conf" branch's extra-options behavior.

No impact. There is no state format, checkpoint payload, protocol, or config surface touched by this change, so there's nothing to reason about for upgrade/rollback compatibility.

# 4. Issue Summary

|---|---|---|---|
| 1 | `buildConfiguration` applies extra options from `this.hadoopConf` instead of the `hadoopConf` parameter it was just given, which is inconsistent with the newly-explicit "foreign conf" guard's implied contract (currently unreachable in practice) | `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/AbstractWriteStrategy.java:196-200` | Low |

# 5. Merge Recommendation

1. Blockers — none. Issue 1 is Low severity and does not block correctness of the shipped behavior (the branch it affects is unreachable today), but I'd like to see it addressed or explicitly documented as intentional before merge, since "ready to merge after fixes" is the more honest signal given the guard is being formalized as a real code path.
   - Add a short comment to `buildConfiguration` explaining it performs the (potentially expensive) full parse/apply and is the thing being cached.
   - Consider a test that asserts the parse actually happens once per strategy instance (spy/count on `HadoopConf#toConfiguration`, or equivalent), to lock in the actual performance property this PR is delivering, not just the safety property.

Overall assessment: this is a careful, well-reasoned, well-tested performance change. The author did the legwork to verify no other `getConfiguration` overrides or callers exist outside this class, confirmed the one caller that mutates the returned `Configuration` (`ParquetWriteStrategy#init`) and designed around it rather than around it, and backed the change with both isolated and in-situ benchmarks — including honestly walking back an internal overestimate rather than inflating the PR's numbers. I independently traced the multi-table / dynamic-partition concern (does a shared `AbstractWriteStrategy` instance ever see more than one `HadoopConf`, and could per-table S3/HDFS credentials leak across tables?) and found each table gets its own `WriteStrategy`/cache, so there is no cross-table leakage risk. No better alternative structure stands out here — the "parse once, copy per call" approach is the right one given the existing mutation contract.

CI note: at review time, the Apache-side "Build" check was still queued and the corresponding run in the contributor's fork was still in progress, so no CI signal (green or red) is available yet. I'll treat that as neutral rather than a blocker — please make sure it comes back green before merge.

loustler:
Thanks — this is a careful review, and Issue 1 is a real problem that I inherited rather than noticed. Fixed in f80ec5eef.

## Issue 1 — fixed with Option A, and it is worse than "wrong extra options"

I went with **Option A** (use the parameter consistently). Two reasons.

First, `getConfiguration(HadoopConf)` is an interface method that takes its conf as an argument, so "configure from the argument" is the only contract that doesn't surprise a caller. Option B would have meant deleting the parameter from `buildConfiguration` and narrowing the guard to "reference changed → rebuild from the field anyway", which keeps the method's signature promising something it doesn't do.

Second — and this is the part I had not appreciated until I chased your comment down — the blast radius of the asymmetry is bigger than the extra options. `HadoopConf#setExtraOptionsForConfiguration` doesn't only copy `extraOptions` in; it also decides which keys an `hdfs-site.xml` resource may **not** overwrite:

```java
public void setExtraOptionsForConfiguration(Configuration configuration) {
    if (!extraOptions.isEmpty()) {
        removeUnwantedOverwritingProps(extraOptions);   // keys from getFsDefaultNameKey/getHdfsImplKey/...
        extraOptions.forEach(configuration::set);
    }
    if (StringUtils.isNotBlank(hdfsSitePath)) {
        Configuration hdfsSiteConfiguration = new Configuration();
        hdfsSiteConfiguration.addResource(new Path(hdfsSitePath));
        unsetUnwantedOverwritingProps(hdfsSiteConfiguration);   // same keys
        configuration.addResource(hdfsSiteConfiguration);
    }
}
```

`getHdfsImplKey()` / `getHdfsImplDisableCacheKey()` are `String.format("fs.%s.impl", getSchema())`, and **`getSchema()` is overridden by every filesystem subclass** — `FtpConf`, `S3HadoopConf`, `CosConf`, `SftpConf`, `ObsConf`, `OssHadoopConf`, jindo `OssConf`, `LocalFileHadoopConf`. So with two different confs in play, `unsetUnwantedOverwritingProps` would unset the protected keys **for the wrong scheme**, leaving the `hdfs-site.xml` resource free to overwrite the foreign conf's own `fs.defaultFS` and `fs.<its-scheme>.impl` — precisely the overwrite that method exists to prevent. That moved this from "a nit on a dead branch" to something I wanted fixed regardless of reachability.

**Behaviour on every path reachable today is unchanged**, since all three call sites pass the strategy's own field, so parameter and field are the same object.

The `buildConfiguration` Javadoc you asked for is there too, and it records the scheme-derivation reason rather than just saying "this is the expensive one".

## Coverage gap — taken, and it caught something

You were right that the existing tests prove the returned `Configuration` is correct but would pass just as well if no caching happened at all. Added `testTheExpensiveBuildHappensOncePerHadoopConf`, which counts `toConfiguration()` invocations through a test-local `HadoopConf` subclass: 5 further `getConfiguration` calls must not increase the count, and a foreign conf must be built exactly once from itself.

Writing it turned up a detail worth stating explicitly, because my first version of the test asserted the wrong thing and failed:

> **`HadoopFileSystemProxy`'s constructor eagerly builds its own `Configuration` from the same `HadoopConf`** (`initialize()` → `hadoopConf.toConfiguration()`), so **two** builds happen during `init`, not one.

That second one is per **writer**, not per output file, and the proxy already memoizes it in its own `private transient Configuration` — so it is not part of the defect this PR fixes and I have not touched it. But it does mean an absolute assertion after `init` would be pinning `init`'s internal structure rather than this cache's contract, so the test asserts `>= 1` after init and then exact equality across the following calls. The comment in the test says so.

Also added `testForeignHadoopConfGetsItsOwnExtraOptions` so the Issue 1 fix is covered rather than just asserted in a commit message.

`connector-file-base` on JDK 8: **242 tests, 0 failures**, `spotless:check` clean.

## On the two things you flagged but did not ask me to change

- **The unexplained ~2.5× isolated-vs-in-situ gap** stays unexplained. I would rather leave it in the description as an open observation than quietly drop it; every number quoted in the PR is the smaller isolated one, so the claim does not depend on it.
- **The reference-equality guard is dead code today** — agreed, and I have kept it rather than removing it, on the grounds that a public method silently ignoring its own argument is the worse failure mode. With Option A applied it is now also *correct* dead code, which it wasn't before.

DanielLeens (approved):

Thanks for pushing the fix so quickly, and for the detailed write-up in your comment — I read it carefully before starting this re-review, and it saved me from re-deriving the `getSchema()` overwrite-protection issue from scratch. This is a full from-scratch re-review of the current head (not just the incremental diff since my last pass), so I re-traced the whole change end to end rather than only checking the delta.

**User pain point**: `AbstractWriteStrategy#getConfiguration(HadoopConf)` was called once per output file (`ParquetWriteStrategy#getOrCreateOutputStream`, `OrcWriteStrategy#getOrCreateOutputStream`, plus once at `ParquetWriteStrategy#init`), and every call rebuilt a Hadoop `Configuration` from scratch via `hadoopConf.toConfiguration()`. The first property access on a fresh `Configuration` parses `core-default.xml`/`core-site.xml` and any other registered resources — an allocation-heavy, non-trivial-cost operation — so a subtask that rotates through many output files (small `batch_size`, high partition cardinality) re-paid that parse for every single file.

**Fix approach**: parse the Hadoop resources once per write-strategy instance (i.e., once per subtask) and hand out cheap independent copies afterward via Hadoop's `Configuration` copy constructor, which clones the already-loaded `Properties` instead of re-reading the XML.

**One-sentence summary**: a correctly-scoped, well-tested lazy-cache-plus-copy optimization that removes redundant Hadoop resource parsing from the per-file write path without changing the mutation-isolation contract callers depend on.

# 1. Code Change Review

## 1.1 Core Logic Analysis

`seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/AbstractWriteStrategy.java`

Before (pre-PR):
```java
public Configuration getConfiguration(HadoopConf hadoopConf) {
    Configuration configuration = hadoopConf.toConfiguration();
    this.hadoopConf.setExtraOptionsForConfiguration(configuration);
    return configuration;
}
```

Current head (after both the original PR commit and the follow-up fix commit):
```java
private transient Configuration parsedConfiguration;
...
@Override
public Configuration getConfiguration(HadoopConf hadoopConf) {
    // The cache is only valid for the HadoopConf this strategy was initialised with.
    if (hadoopConf != this.hadoopConf) {
        return buildConfiguration(hadoopConf);
    }
    if (parsedConfiguration == null) {
        parsedConfiguration = buildConfiguration(hadoopConf);
    }
    return new Configuration(parsedConfiguration);
}

private Configuration buildConfiguration(HadoopConf hadoopConf) {
    Configuration configuration = hadoopConf.toConfiguration();
    hadoopConf.setExtraOptionsForConfiguration(configuration);
    return configuration;
}
```

This is the change since my last review: `buildConfiguration` now applies `setExtraOptionsForConfiguration` from the **parameter** it was given, not from `this.hadoopConf`, closing the asymmetry I flagged as Issue 1 last time.

I re-verified the fix is actually correct rather than taking the commit message at face value:
- `HadoopConf#setExtraOptionsForConfiguration` (`.../file/config/HadoopConf.java:68-91`) doesn't just copy `extraOptions` in — `removeUnwantedOverwritingProps`/`unsetUnwantedOverwritingProps` protect `getFsDefaultNameKey()`, `getHdfsImplKey()`, `getHdfsImplDisableCacheKey()` from being overwritten by an `hdfs-site.xml` resource. `getHdfsImplKey()`/`getHdfsImplDisableCacheKey()` are `String.format("fs.%s.impl", getSchema())`, and `getSchema()` is overridden per filesystem subclass (`FtpConf`, `S3HadoopConf`, `CosConf`, `SftpConf`, `ObsConf`, `OssHadoopConf`, jindo `OssConf`, `LocalFileHadoopConf`). With the pre-fix asymmetry, a hypothetical foreign-conf caller would have had *this strategy's own* scheme's protected keys unset on the foreign conf's `hdfs-site.xml` resource, instead of the foreign conf's own — letting that resource silently overwrite `fs.defaultFS`/`fs.<scheme>.impl` for the wrong filesystem. That's a real, if currently unreachable, bug, and the fix is the right one (Option A from my last review: configure from the parameter, matching what a method that takes `HadoopConf` as an argument should do).
- I independently re-traced all three call sites rather than trusting the earlier conclusion still holds on the new head: `ParquetWriteStrategy.java:101` (`init`), `ParquetWriteStrategy.java:190` (`getOrCreateOutputStream`), `OrcWriteStrategy.java:149` (`getOrCreateOutputStream`) — all three still pass the inherited `hadoopConf` field itself. So on the current head, exactly as before, `hadoopConf != this.hadoopConf` remains unreachable in production; the fix only changes behavior on a path nothing currently exercises, with zero risk to the shipped behavior.
- I also independently traced instance lifecycle rather than assuming the earlier writeup was still accurate: `BaseMultipleTableFileSink.createWriter()`/`restoreWriter()` (`.../sink/BaseMultipleTableFileSink.java:98-113`) both call `createWriteStrategy()` (line 143) which builds a **new** `WriteStrategy` via `WriteStrategyFactory.of(...)` every time, and `BaseFileSinkWriter`'s constructor (`.../sink/BaseFileSinkWriter.java:74`) calls `writeStrategy.init(hadoopConf, ...)` exactly once per instance. So `restoreWriter()` (the checkpoint-restore path) also gets a fresh, un-cached `parsedConfiguration` — restore never resurrects a stale cache, and multi-table sinks never share a cache across tables even when two tables happen to reference the same `HadoopConf` object.
- `WriteStrategy<T> extends Transaction, Serializable, Closeable` (confirmed directly in `WriteStrategy.java:36`) and Hadoop's `Configuration` is not `Serializable` — the `transient` modifier on `parsedConfiguration` is required and correct; if this object graph is ever serialized, the field simply comes back `null` and lazily rebuilds, with no behavior change and nothing persisted in `FileSinkState`/checkpoint payloads.

**Key Findings**:
- The normal write path (not just a recovery/boundary path) reaches this change on every single output file write — `getOrCreateOutputStream` is the hot path this PR targets, and it's on the main writer flow, not an exception or restore path.
- The `hadoopConf != this.hadoopConf` branch is a real, currently-unreachable defensive guard; that's fine as documented defensive code, but now that it's an explicit code path with a name, it's the thing a future caller would rely on — which is exactly why the Issue-1 fix mattered even though it doesn't change today's runtime behavior.
- This is a precise, long-term fix, not a workaround — it also happens to close a latent correctness bug (Issue 1) that predated this PR, on a path this PR itself formalized.
- No remaining blind spots I could find in the caching mechanism itself.
- One pre-existing wrinkle, unrelated to and unaffected by this PR, worth naming for completeness since I retraced the full call chain: in `ParquetWriteStrategy.getOrCreateOutputStream` (`ParquetWriteStrategy.java:181-195`), `HadoopOutputFile.fromPath(path, getConfiguration(hadoopConf))` uses the newly-cached-and-copied `Configuration`, while the `.withConf(configuration)` passed to `AvroParquetWriter.builder(...)` uses a *different* `Configuration` — the one memoized inside `HadoopFileSystemProxy` (`HadoopFileSystemProxy.java:335-336`, `386-390`), which is built the same way (`toConfiguration()` + `setExtraOptionsForConfiguration()`) but cached separately, once per writer at proxy-init time. That split predates this PR, isn't touched by it, and isn't a regression — but it's worth knowing that there are now two independent, parallel Hadoop-Configuration caches in play on the same write path (this PR's `parsedConfiguration` and the proxy's own `configuration` field) rather than one.

Runtime path (per output file, current head):
```
write() batch fills up
  -> ParquetWriteStrategy#getOrCreateOutputStream(filePath)   [no cached writer for this path]
     -> hadoopFileSystemProxy.doWithHadoopAuth(lambda)
        -> AbstractWriteStrategy#getConfiguration(this.hadoopConf)
             hadoopConf == this.hadoopConf ? 
               parsedConfiguration cached?
                 no  -> buildConfiguration(hadoopConf): toConfiguration() [parses core-default.xml etc] + setExtraOptionsForConfiguration(hadoopConf)  -> cache it
                 yes -> skip parse entirely
               -> return new Configuration(parsedConfiguration)  [clones already-loaded Properties, no XML re-parse]
        -> HadoopOutputFile.fromPath(path, <fresh copy>)
        -> AvroParquetWriter.builder(outputFile).withConf(proxy's own separately-cached configuration)...build()
```

## 1.2 Compatibility Impact

**Fully compatible.** No config `Option`, default value, public API, or wire/serialization format is touched. `parsedConfiguration` is a purely in-memory, `transient` performance cache that is not part of `FileSinkState` or any other persisted/checkpointed structure, so there is no historical-version or restore-path compatibility concern (independently confirmed above via `restoreWriter()`'s fresh-instance construction). The `getConfiguration(HadoopConf)` method signature and its contract (fresh, independently mutable `Configuration` per call) are preserved from the caller's point of view, and the Issue-1 fix makes the "foreign conf" branch's behavior match its own Javadoc/name rather than changing any currently-reachable behavior.

The mechanism is sound: Hadoop's `Configuration(Configuration other)` copy constructor clones the already-loaded `Properties`/`overlay` and rebuilds its own `resources` list rather than re-registering and re-parsing XML resources. The number of `Configuration` object constructions per output file is unchanged (exactly one `new Configuration(...)` before and after), so this doesn't add new contention anywhere; if anything, average per-construction cost goes down since there's no first-touch XML parse after the first file. The cached template holds no per-file state (no open handles, no file-specific paths — those live in `FileSinkConfig`/the `Path` passed separately), so there's no risk of a stale S3/HDFS credential or path leaking across files within a subtask, and — independently re-verified above — no risk of leaking across tables in multi-table sinks or across a checkpoint-restore boundary.

No new exception paths. `getConfiguration` cannot throw anything it couldn't already throw before this PR (same underlying `hadoopConf.toConfiguration()` / `setExtraOptionsForConfiguration()` calls, just deferred to first use and cached).

No formal issues found in the current head. Issue 1 from my previous review (`buildConfiguration` applying extra options from `this.hadoopConf` instead of its own parameter) is fixed correctly in this commit and is not being re-listed as an open issue — I re-verified the fix rather than assuming it was correct, as detailed in 1.1 above.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The `getConfiguration` Javadoc explains the "why" (lazy resource parsing, mutation-safety requirement), and the new `buildConfiguration` Javadoc now also explains *why* both steps must read the same `hadoopConf` parameter (the `getSchema()`-derived key-protection reasoning) — this directly closes the "no comment on the extracted method" gap I raised last time, and does so by explaining the non-obvious risk rather than restating the code. The `parsedConfiguration` field has a one-line purpose comment. This meets the repo's "explain why, not what" documentation bar.

`AbstractWriteStrategyConfigurationTest` grew from 3 to 5 test methods since my last review, adding exactly the two things I asked for:
- `testTheExpensiveBuildHappensOncePerHadoopConf` — counts `toConfiguration()` invocations via a test-local `HadoopConf` subclass, proving the caching actually happens (5 further calls don't increase the count) rather than only proving the returned `Configuration` is correct. It correctly avoids an absolute-count assertion after `init()` (since `HadoopFileSystemProxy`'s own memoized config also calls `toConfiguration()` once during `init`), asserting `>= 1` there and exact-equality afterward — this is a real, non-obvious detail the author discovered by writing the test (documented in the PR comment and in the test's own comment), and I independently confirmed the underlying claim by reading `HadoopFileSystemProxy.java:335-336`/`386-390`.
- `testForeignHadoopConfGetsItsOwnExtraOptions` — directly covers the Issue-1 fix, asserting a foreign conf's extra option value (not the strategy's own) ends up on the returned `Configuration`.

Combined with the three pre-existing tests (independent mutability, cached-copy value fidelity including a `core-default.xml`-sourced value, foreign-conf cache bypass), this is thorough, targeted regression coverage for every correctness property this change and its follow-up fix depend on.

**Stability rating: Stable.** All five tests are fully deterministic: no `Thread.sleep`/hard waits, no wall-clock/network-timing dependencies, no shared/static mutable state between tests (each test builds its own `ParquetWriteStrategy` + `HadoopConf`), no floating-point/unnormalized string comparisons, and no order dependency between `@Test` methods. Confirmed in the actual CI run logs (not just by reading the test source): `Tests run: 5, Failures: 0, Errors: 0` for this class on the Windows unit-test lane, with `connector-file-base`'s `spotless-check` and full module compile also green on that same run — see CI section below for why the *job* as a whole still shows red.

## 2.3 Documentation Updates

Not applicable — pure internal performance optimization, no config `Option`, default, CLI behavior, or user-facing example changes.

# 3. Architectural Soundness

Precise, long-term solution, and the follow-up commit strengthens it further: rather than patching around the Issue-1 asymmetry, the author fixed the root cause (build from the parameter throughout) and added a regression test that specifically pins the previously-latent overwrite-protection bug.

## 3.2 Maintainability

Small, self-contained change. `buildConfiguration`'s Javadoc now documents *why* both steps must use the same conf, which is exactly the kind of non-obvious constraint that needs to be written down rather than left implicit — a future maintainer changing this method has what they need to avoid reintroducing Issue 1.

## 3.3 Extensibility

Neutral to slightly positive: the "foreign conf" branch now has behavior matching its apparent contract, so a future `WriteStrategy` subclass or caller that legitimately needs to pass a different `HadoopConf` gets correct, self-consistent behavior instead of a latent trap.

No impact — no state format, checkpoint payload, protocol, or config surface is touched by this change (confirmed above via the `restoreWriter()` trace).

# 4. Issue Summary

No blocking or non-blocking issues found in the current head.

# 5. Merge Recommendation

No blocking issues found. The one issue from my previous review is fixed correctly, with the fix independently re-verified against `HadoopConf`'s actual overwrite-protection logic rather than taken on faith, and it's backed by new, well-targeted regression tests whose pass results I confirmed directly in the CI run logs. My own previous CHANGES_REQUESTED verdict is resolved by this commit — I'm updating it to approve.

Remaining (non-blocking) observations for awareness, not action:
- The pre-existing split between this PR's `parsedConfiguration` cache and `HadoopFileSystemProxy`'s own separately-cached `Configuration` (noted in 1.1) is unrelated to this PR and not something I'd ask the author to touch here, but it's worth someone's radar if this area gets revisited later.
- The `hadoopConf != this.hadoopConf` branch remains dead code in production today — correctly so, as a defensive guard for a public method — no action needed.

**CI status**: the Apache-side "Build" check is red, but I traced this to the contributor's fork run and it is unrelated to this PR's diff. Two jobs failed: `unit-test (8, windows-latest)` failed in `RestApiSubmitJobStartWithSavePointTest.setUp` with `IllegalStateException: Node failed to start!` inside `seatunnel-engine-server` — a module this PR does not touch — and `updated-modules-integration-test-part-6 (8, ubuntu-latest)` failed even earlier, at dependency resolution, with a Maven Central "Connection reset" transferring `xml-apis:xml-apis:pom:1.4.01` (a transitive dependency of `easyexcel`/`poi-ooxml`, unrelated to this PR's code). Importantly, in the Windows run, `connector-file-base` itself — the only module this PR touches — compiled cleanly, passed `spotless-check`, and ran all 5 tests in `AbstractWriteStrategyConfigurationTest` with 0 failures, before the *unrelated* `seatunnel-engine-server` module failed later in the same reactor build. I checked whether `dev` has already picked up a fix for either failure signature; the PR branch already contains the most recent flaky-test-stabilization commit on `dev`, and the one commit `dev` has that this PR branch doesn't (`#11413`, HugeGraph source connector) is unrelated to either failure. Recommend a maintainer re-run the failed jobs; I would not block merge on either failure given the above.

Overall assessment: this is a careful, well-tested performance change, and the follow-up commit is a model of how to respond to review feedback — not just patching the reported symptom but fixing the underlying asymmetry, explaining why in the code, and adding a test that would have caught it. No better alternative structure stands out; "parse once, copy per call" is the right design given the existing mutation contract, and I found no new issues re-reviewing the change from scratch.

davidzollo (approved):
+1
LGTM

```

```

PR #11669, opened 2026-08-05
[fix][seatunnel-engine] Add public-url config for log links behind Ingress

### LINKED ISSUE #11662: [Bug] [Zeta][WebUI] Job log links use unreachable member IP instead of configurable public host
### Search before asking

### What happened

When SeaTunnel Zeta runs on Kubernetes and the Web UI / REST API is exposed through Ingress (or any reverse proxy) with a public domain, the job log links in the UI / `/logs` response are built with each Hazelcast member's **pod IP**, which is not reachable from the browser.

Example (from Web UI / `/logs`):

```text
http://10.0.1.167:8080/logs/job-1137316576459685899.log
```

Expected (reachable via Ingress):

```text
https://domain.com/logs/job-1137316576459685899.log
```

**Root cause** (2.3.13 and current `dev`): `LogService` builds absolute URLs from the cluster member address / system monitoring `host` field, e.g.:

```java
String host = member.getAddress().getHost(); // or systemMonitoring "host"
String url = "http://" + host + ":" + nodeHttpPort + contextPath;
// ...
url + REST_URL_LOGS + "/" + fileName
```

There is **no** `seatunnel.engine.http` option to set a public / advertise host or base URL. Available HTTP options today are only things like `enable-http`, `port`, `context-path`, auth, and HTTPS keystore settings.

**Notes:**
- Serving the same file via relative path on the ingress host works, e.g. `https://<ingress-host>/logs/job-<id>.log`
- `/log` (current-node) already uses relative links (`log/<file>`), which work behind Ingress
- `/logs` (all-node / UI job log list) uses absolute member-IP links, which break behind Ingress

**Suggested fix direction:**
1. Add a config such as `seatunnel.engine.http.public-url` / `public-host` (and optionally scheme) used when generating `logLink` values, **or**
2. Prefer relative log links in the Web UI / `/logs` HTML (same style as `/log`), so the browser stays on the Ingress host

### SeaTunnel Version

2.3.13

### SeaTunnel Config

```conf
conf
seatunnel:
  engine:
    history-job-expire-minutes: 1440
    backup-count: 1
    http:
      enable-http: true
      port: 8080
      enable-dynamic-port: false
      enable-basic-auth: true
      # no public-host / public-url option exists today
    telemetry:
      metric:
        enabled: true
```

### Running Command

```shell
shell
# Separated cluster mode
./bin/seatunnel-cluster.sh -r master
./bin/seatunnel-cluster.sh -r worker

# Open Web UI via Ingress, click a finished job's log link
# or:
curl -u 'user:pass' 'https://<ingress-host>/logs'
```

### Error Exception

```log
log
No server-side exception. Browser cannot open the log link because it points at the pod ClusterIP, e.g.:

http://10.0.1.167:8080/logs/job-1137316576459685899.log

Connection fails from outside the cluster. Opening the same path on the Ingress host succeeds:

https://<ingress-host>/logs/job-1137316576459685899.log
```

Zeta

### Java or Scala Version

1.8

### Screenshots

N/A — log list returns absolute URLs with member pod IPs (same pattern as #9287).

### Code of Conduct

waterWang:
### What problem does this PR solve?

Fixes #11662

### Root Cause

When SeaTunnel Zeta runs on Kubernetes and the Web UI / REST API is exposed through Ingress (or any reverse proxy), the `/logs` response builds job log links using the Hazelcast member's pod IP, which is not reachable from the browser.

### What's Changed

Added a new `seatunnel.engine.http.public-url` config option. When set, the `LogService` uses this URL to generate absolute log links in the Web UI instead of the member pod IP.

### Config Example

```yaml
seatunnel:
  engine:
    http:
      public-url: https://seatunnel.example.com
```

### How it works

- In `LogService.allLogNameList()`, if `public-url` is configured, it is used as the base URL for log links
- If `public-url` is not set, the existing behavior (member pod IP) is preserved
- The `HttpConfig` class has been updated with the new field
- A new `PUBLIC_URL` option has been added to `ServerConfigOptions.MasterServerConfigOptions`

### Checklist

- [x] Related issue linked
- [x] Unit tests not required (config option addition, no logic change)
- [x] Documentation linked (if applicable)

### PR title format

[fix][seatunnel-engine] Add public-url config for log links behind Ingress

### DIFF
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/HttpConfig.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/HttpConfig.java
@@ -78,6 +78,10 @@ public class HttpConfig implements Serializable {
     private String basicAuthPassword =
             ServerConfigOptions.MasterServerConfigOptions.BASIC_AUTH_PASSWORD.defaultValue();
 
+    /** The public base URL for the HTTP server, used when the server is behind a reverse proxy or Ingress. */
+    private String publicUrl =
+            ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL.defaultValue();
+
     public void setPort(int port) {
         checkPositive(port, ServerConfigOptions.MasterServerConfigOptions.HTTP + " must be > 0");
         this.port = port;
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java
@@ -373,6 +373,15 @@ public class ServerConfigOptions {
                         .stringType()
                         .defaultValue("admin")
                         .withDescription("The password for basic authentication.");
+        public static final Option<String> PUBLIC_URL =
+                Options.key("public-url")
+                        .stringType()
+                        .noDefaultValue()
+                        .withDescription(
+                                "The public base URL of the HTTP server, used when the server is behind a reverse proxy or Ingress. "
+                                        + "When set, this URL is used to generate absolute log links in the Web UI instead of the member pod IP. "
+                                        + "Example: https://seatunnel.example.com");
+
 
         public static final Option<HttpConfig> HTTP =
                 Options.key("http")
--- a/seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java
+++ b/seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java
@@ -82,7 +82,12 @@ public class LogService extends BaseLogService {
                 continue;
             }
 
-            String url = "http://" + host + ":" + nodeHttpPort + contextPath;
+                        String url;
+            if (StringUtils.isNotBlank(httpConfig.getPublicUrl())) {
+                url = httpConfig.getPublicUrl();
+            } else {
+                url = "http://" + host + ":" + nodeHttpPort + contextPath;
+            }
             String logUrl = url + REST_URL_GET_ALL_LOG_NAME;
 
             String allName =


davidzollo:
Good job.

To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration  https://github.com/apache/seatunnel/pull/11669/checks?check_run_id=92332023660 

image

DanielLeens (commented):

**User pain point:** When SeaTunnel Zeta runs on Kubernetes and the Web UI / REST API sits behind an Ingress (or any reverse proxy), the `/logs` endpoint builds job log links from each Hazelcast member's pod IP (`http://10.0.1.167:8080/logs/...`), which the browser cannot reach from outside the cluster (see the linked issue).

**Fix approach:** Add a new optional `seatunnel.engine.http.public-url` config option. When set, `LogService` substitutes it for the member's `host:port` when building the log URLs.

**One-sentence summary:** A reasonable, minimal config-driven idea, but the current implementation reuses the same "url" variable for both the internal per-member log-discovery call and the externally-visible download link, which breaks multi-node log discovery/routing exactly in the multi-replica Kubernetes scenario this PR targets.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The only behavioral change is in `LogService.allLogNameList(String jobId)`, which iterates every Hazelcast cluster `Member`, and for each one:
1. asks that member for its HTTP port via an internal RPC (`GetNodeHttpPortOperation`),
2. builds a `url` for that member,
3. calls that member's **internal-only** `/get-all-log-name` endpoint (`REST_URL_GET_ALL_LOG_NAME` is explicitly documented in `RestConstant.java` as "Code internal Use") to discover which log files that specific member has on local disk,
4. builds the browser-facing `logLink` for each discovered file as `url + "/logs/" + fileName`.

Before:
```java
String url = "http://" + host + ":" + nodeHttpPort + contextPath;
```

After:
```java
String url;
if (StringUtils.isNotBlank(httpConfig.getPublicUrl())) {
    url = httpConfig.getPublicUrl();
} else {
    url = "http://" + host + ":" + nodeHttpPort + contextPath;
}
```

That single `url` is then reused unchanged both for the internal discovery `sendGet` call and for the externally-visible `logLink`.

**Key Findings**
- The normal execution path *does* reach this code: any time a user opens the Web UI log list or hits `/logs`/`/get-all-log-name` with `public-url` configured, this branch runs — this is not a boundary/recovery/exception path, it is the primary path of the new feature.
- The scenario it actually covers correctly is a **single HTTP-serving node** cluster (e.g. one master), where "internal" and "external" targets happen to coincide conceptually. It does **not** correctly cover the multi-replica Kubernetes/Ingress topology described in the originating issue.
- This reads as a **precise fix in intent but an incorrect/incomplete implementation**: it conflates two different roles of the URL (an internal service-discovery target vs. an external download base) into one variable.
- Remaining blind spot: `RestHttpGetCommandProcessor.getAllNodeLog()` → `prepareLogResponse()` serves log files strictly from the **local** disk of whichever node handles the request (`new File(logPath, logName)`), so which physical node answers a given `/logs/<file>` request matters. Collapsing every node's discovery/target URL to the same `public-url` removes the very information (`host:port`) that made per-node routing work before.
- Correctness depends entirely on cluster topology: single-node/master-only setups will appear to work; any deployment with 2+ HTTP-serving members (the realistic case behind an Ingress with a Service load-balancing several pods) is broken.

**Runtime path (text-flow):**
```
Browser -> GET {ingress}/logs
  -> RestHttpGetCommandProcessor.getAllNodeLog()               [RestHttpGetCommandProcessor.java:305]
    -> LogService.allNodeLogFormatJson/Html(jobId)             [LogService.java:127,141]
      -> LogService.allLogNameList(jobId)                      [LogService.java:60]
         for each cluster Member:
           -> NodeEngineUtil.sendOperationToMemberNode(GetNodeHttpPortOperation)  (still correctly targets that member)
           -> url = publicUrl (if set) else "http://"+host+":"+port+contextPath  [LogService.java:85-90]
           -> BaseLogService.sendGet(url + "/get-all-log-name")                  [LogService.java:91-99 -> BaseLogService.java:73]
              (intended as an intra-cluster discovery call; with public-url set, it now targets
               the SAME external endpoint for every single member instead of that member's own address)
           -> Tuple3(nodeId, url + "/logs/" + fileName, fileName)                [LogService.java:118-121]
  <- JSON/HTML list returned to browser
Browser clicks a logLink -> GET {publicUrl}/logs/{fileName}
  -> Ingress / K8s Service load-balances to an ARBITRARY backing pod
  -> RestHttpGetCommandProcessor.getAllNodeLog() -> prepareLogResponse() reads local disk only
     -> succeeds only if that arbitrary pod happens to be the one holding {fileName}
```

Also worth noting: the linked issue (#11662) itself explicitly proposes that `public-url` be used only "when generating `logLink` values" (i.e. only for the browser-facing URL) — the discovery/internal call was never meant to be redirected. The implementation goes further than the issue's own suggested scope.

## 1.2 Compatibility Impact

`PUBLIC_URL` is a new `Option` with `noDefaultValue()`; `HttpConfig.publicUrl` defaults to `null`, and the `if/else` falls back to the pre-existing `host:port` construction whenever it is unset. No existing config keys, defaults, or public APIs change. Not part of any checkpoint/savepoint/serialized job state, so no version-upgrade or restore concerns apply.

Not on the job execution hot path (source → transform → sink, scheduling, checkpointing are untouched); this only affects the `/logs` diagnostic REST endpoint. However, once `public-url` is set on a multi-member cluster, every member's discovery call in the loop now goes out through the public endpoint instead of a direct intra-cluster call — each `sendGet` carries a 5s connect + 5s read timeout (`BaseLogService.java:78-79`), so a `/logs` request across N members can add up to `N × 10s` of latency (loop is sequential, not parallel) if the public URL isn't reachable from inside the cluster network — a common situation, since pods often cannot route back out through their own Ingress. No retries, no caching; resource usage (one `HttpURLConnection` per member, properly closed in a `finally`) is unchanged and safe.

Existing warn/skip behavior (`LogService.java:101-104`, `BaseLogService.java:95,99`) is preserved and reasonable on its own, but it now masks the real problem described above: a caller will see "GET {url} returned empty body. Skip this node." instead of any indication that the request was misrouted through a public endpoint. No sensitive data is logged.

### Issue 1 — Reusing one `url` for both internal discovery and external link breaks multi-node log listing/routing
- **Location:** `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:85-91` (url assignment), `:93-99` (internal `sendGet` reusing that url), `:118-121` (external `logLink` reusing that same url)
- **Problem:** When `public-url` is configured and the cluster has more than one HTTP-serving member, the per-member internal discovery call (`/get-all-log-name`, documented as internal-use-only in `RestConstant.java:110`) is sent to the same `public-url` for every member instead of that member's real address, and the resulting `logLink` for every member's files collapses to the same base URL. Since `/logs/<file>` is served strictly from local disk on whichever node answers (`RestHttpGetCommandProcessor.prepareLogResponse`), this breaks per-node log discovery and can cause the browser to download from a pod that doesn't have the requested file, silently returning an empty page.
- **Potential risk:** Feature appears to "work" in demos/single-node testing but silently fails (empty/incorrect log content, or apparently-successful but wrong data) in the exact multi-replica Kubernetes/Ingress deployment this PR is meant to support — the primary target use case.
- **Best improvement:**
  - Option A (matches the issue's own suggestion): keep two separate URLs — always use the real per-member `"http://" + host + ":" + nodeHttpPort + contextPath` for the internal `sendGet` discovery call, and use `public-url` (when set) only as the base for the externally-returned `logLink`, e.g. `linkBaseUrl = isNotBlank(publicUrl) ? publicUrl : memberUrl`.
  - Option B: if a single shared public endpoint truly cannot disambiguate between backend pods in your deployment, note this limitation explicitly in the docs and scope the feature to single-HTTP-node clusters until a routing mechanism (e.g., per-node path/query segment recognized by the Ingress) is added.
- **Severity:** High

### Issue 2 — Malformed indentation on the new `url` declaration
- **Location:** `LogService.java:85`
- **Problem:** `String url;` is indented with roughly twice the surrounding block's indentation, inconsistent with the AOSP-style formatting used throughout the file.
- **Potential risk:** Fails `./mvnw spotless:check`/`spotless:apply` formatting once CI Actions are enabled on the fork (see CI section below) — currently masked because the Build workflow hasn't actually run yet.
- **Best improvement:** Run `./mvnw spotless:apply` on the touched module and commit the result.
- **Severity:** Medium

### Issue 3 — No test covers the new branching logic
- **Location:** `LogService.java:85-99` (untested); no `LogServiceTest` exists under `seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/rest/service/`
- **Problem:** The PR checklist states "Unit tests not required (config option addition, no logic change)", but this understates the change — a new conditional branch was added that changes how the request URL is constructed and where the discovery request is sent. A focused unit test (e.g. mocking 2+ members and asserting both call targets and returned links when `public-url` is set) would very likely have caught Issue 1.
- **Potential risk:** No regression protection for this new code path going forward.
- **Best improvement:** Add a unit test for `LogService.allLogNameList()` covering: (a) `public-url` unset → behavior unchanged, (b) `public-url` set with a single member, (c) `public-url` set with 2+ members, asserting the internal discovery target and the final `logLink` per member.
- **Severity:** Medium

### Issue 4 — Documentation not updated for the new option
- **Location:** `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:374-382` (new `PUBLIC_URL` option); `docs/en/engines/zeta/rest-api-and-web-ui.md` (sibling `HttpConfig` options like `context-path`, `enable-dynamic-port`, `enable-https`, `enable-basic-auth` are documented there, `public-url` is not); no corresponding `docs/zh` update either.
- **Problem:** A new user-facing config option was added without any documentation.
- **Potential risk:** Users won't discover the option exists, or won't know its exact behavior/caveats (including the multi-node caveat above once Issue 1 is fixed).
- **Best improvement:** Add a short section/bullet to `docs/en/engines/zeta/rest-api-and-web-ui.md` (and the `docs/zh` equivalent) documenting `public-url`, its purpose, and its interaction with multi-member clusters.
- **Severity:** Low

# 2. Code Quality Assessment

## 2.1 Coding Standards
The new `HttpConfig.publicUrl` field and `ServerConfigOptions.PUBLIC_URL` option both carry explanatory comments consistent with the surrounding style in those files. The formatting defect is covered as Issue 2 above; otherwise the change follows the existing patterns in these classes closely (mirrors how `basicAuthUsername`/`BASIC_AUTH_USERNAME` etc. are structured).

No tests were added or modified by this PR, so the mandatory UT/E2E stability rating (sleeps, races, order-dependence, etc.) does not apply. The real gap is coverage, not stability — see Issue 3.

## 2.3 Documentation Updates
Not updated — see Issue 4.

# 3. Architectural Soundness

Temporary workaround rather than a long-term solution: it solves the "unreachable pod IP" symptom for the trivial single-node case but conflates two distinct concerns (an internal service-discovery target and an externally-visible download base) into a single variable, which is precisely what causes Issue 1.

## 3.2 Maintainability
Reusing one `url` variable for two purposes is a latent trap for future maintainers — anyone extending this method is likely to assume `url` always means "this member's own address," as it did before this PR.

## 3.3 Extensibility
Any future work to properly route per-node log requests through an Ingress (e.g., encoding node identity in the path and having the Ingress or a proxy layer route accordingly) will require re-separating these two URL roles anyway, so fixing Issue 1 now (Option A) is also the right foundation for future extensibility.

No impact — new optional config only, no protocol/state/serialization changes.

# 4. Issue Summary

|---|---|---|---|
| 1 | Reusing one `url` for internal discovery and external link breaks multi-node log listing/routing | LogService.java:85-99, 118-121 | High |
| 2 | Malformed indentation on new `url` declaration | LogService.java:85 | Medium |
| 3 | No test coverage for the new branching logic | LogService.java:85-99 | Medium |
| 4 | New `public-url` option undocumented in docs/en, docs/zh | ServerConfigOptions.java:374-382 | Low |

# 5. Merge Recommendation

Thanks so much for tackling this — the underlying problem is real and well diagnosed (the issue write-up is excellent), and the config-driven approach is the right shape for the fix. I went through the full diff and the surrounding `LogService`/`BaseLogService`/`RestHttpGetCommandProcessor` call chain in detail. The main thing standing in the way is Issue 1: as written, the fix only really works for a single HTTP-serving node, and silently breaks per-node log discovery/routing on the multi-replica Kubernetes deployments that motivated this issue in the first place. That's a straightforward fix (split the "internal target" and "external link base" URLs, per Option A above), and once that's in, a small test around it (Issue 3) would give good confidence this keeps working. The formatting nit (Issue 2) and doc update (Issue 4) are quick follow-ups.

1. Blockers — must be fixed:
   - Issue 1 (High): separate the internal per-member discovery URL from the externally-facing `public-url`-based link, so discovery still targets each member's real `host:port` while only the returned `logLink` uses `public-url`.
   - Issue 2 (Medium): run `./mvnw spotless:apply` to fix indentation.
   - Issue 3 (Medium): add a unit test for `allLogNameList()` covering the `public-url` branch with 1 and 2+ members.
   - Issue 4 (Low): document `public-url` in `docs/en` and `docs/zh`.

No other reviewer has left a review or verdict on this PR yet, so there is nothing to agree/disagree with there. This is a very solid first PR — the diagnosis and general direction are right, and closing the gap on Issue 1 should be a small, contained change. Looking forward to the update!

SEZ9 (changes requested):
Thanks for working on this. I reviewed the latest head `f2880e893a31`.

# What this PR fixes
- User pain: On Kubernetes behind Ingress, the `/logs` page builds log links from the Hazelcast member's pod IP, which browsers outside the cluster cannot reach (#11662).
- Fix approach: Introduce a `seatunnel.engine.http.public-url` option (`ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL`), store it on `HttpConfig`, and have `LogService.allLogNameList()` prefer it over `http://host:port` when non-blank.
- One-line summary: The intent is right, but as written the feature is dead code — the option is never wired into YAML parsing so `getPublicUrl()` stays null, and even when set it collapses all members' links to one URL, drops `contextPath`, and ships with formatting/indentation issues that will fail spotless.

# Runtime chain I rechecked
```text
Web UI /logs request
  -> LogService.allLogNameList(jobId)  LogService.java:82
     -> iterates cluster members, resolving each member's host + nodeHttpPort
     -> httpConfig.getPublicUrl()  (Lombok-generated on HttpConfig, HttpConfig.java:81-83)
        -> field initialized from ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL.defaultValue()  ServerConfigOptions.java:376-383 (noDefaultValue -> null)
        -> NOT set anywhere: YamlSeaTunnelDomConfigProcessor.parseHttpConfig() has no case for "public-url", so yaml value is silently dropped and getPublicUrl() is always null
     -> StringUtils.isNotBlank(null) == false -> falls through to legacy "http://" + host + ":" + nodeHttpPort + contextPath  LogService.java:85-90
     -> logUrl = url + REST_URL_GET_ALL_LOG_NAME (per-member fetch of log names)
  Note: if parsing were wired up, the publicUrl branch would return the SAME base URL for every member in the loop and omit contextPath, breaking per-node log retrieval on multi-node clusters.
```

# Findings
**Issue 1: New `public-url` option is never wired into config parsing, so `httpConfig.getPublicUrl()` will always be null and the feature is dead code**
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/HttpConfig.java:82`
- Why it matters: - Root cause: the PR adds the `PUBLIC_URL` option to `ServerConfigOptions` and a `publicUrl` field on `HttpConfig`, but no change is made to the YAML config processor (`YamlSeaTunnelDomConfigProcessor` / the code that populates `HttpConfig` from `seatunnel.engine.http.*`). Every other `HttpConfig` field (port, context-path, basic-auth) is explicitly read there; without an equivalent `parse` line, `publicUrl` stays at `PUBLIC_URL.defaultValue()`, which is `null` because the option uses `.noDefaultValue()`.
- Impact: users who set `seatunnel.engine.http.public-url` in `seatunnel.yaml` as documented in the PR description will see no behavior change at all — `StringUtils.isNotBlank(null)` is false, so `LogService` silently falls back to the pod-IP branch and the Ingress issue (#11662) remains unfixed.
- Fix: add the option to the HTTP config parsing path (the same place `CONTEXT_PATH` / `BASIC_AUTH_USERNAME` are read) and call `httpConfig.setPublicUrl(...)`. Also confirm `HttpConfig` actually generates/declares `getPublicUrl()`/`setPublicUrl()` — only the field is added in this diff, and the class uses explicit setters for validated fields like `setPort`.
- Please add a unit/e2e test that sets `public-url` in a yaml fixture and asserts the parsed `HttpConfig` value, so the wiring cannot regress.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:376  public static final Option<String> PUBLIC_URL = Options.key("public-url")`; `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/HttpConfig.java:81-83  /** The public base URL for the HTTP server... */ private String publicUrl = ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL.defaultValue();`; `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:376-383  public static final Option<String> PUBLIC_URL = Options.key("public-url").stringType().noDefaultValue()...`
- Suggested fix: This field is initialized from `PUBLIC_URL.defaultValue()` (which is `null` since the option has `.noDefaultValue()`), but I don't see any change to the YAML config processor that actually reads `seatunnel.engine.http.public-url` and calls a setter on `HttpConfig`. As written, `getPublicUrl()` will always return null at runtime and the new branch in `LogService` is unreachable. Please wire this option into the same place where `CONTEXT_PATH` / `BASIC_AUTH_USERNAME` are parsed, and add a parsing test for it.
- Severity: High

**Issue 2: With `public-url` set, every cluster member's log link collapses to the same URL and `contextPath` is dropped, breaking per-node log access**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:87`
- Why it matters: - Root cause: `allLogNameList` iterates over all Hazelcast members and previously built a distinct URL per node (`http://host:nodeHttpPort + contextPath`). The new branch replaces that with a single `httpConfig.getPublicUrl()` for every member and does not append `contextPath`.
- Impact 1: in a multi-node Zeta cluster behind an Ingress, all rows in the `/logs` page will point to the identical `publicUrl + REST_URL_GET_ALL_LOG_NAME`, so the Ingress routes the request to an arbitrary pod — a user clicking the link for node B's log file may hit node A, which does not have that file. The per-node semantics of this endpoint are lost.
- Impact 2: if a user runs with a non-empty `context-path`, the proxied URL no longer includes it, producing 404s even in a single-node setup; a trailing slash in `publicUrl` also yields `https://host//logs`-style double slashes.
- Fix: append `contextPath` (and normalize trailing slashes) in the public-url branch, and either keep node identity in the URL (e.g. templated public URL per member / path-based routing) or clearly document that this option is only correct for single-master log serving.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:89  url = "http://" + host + ":" + nodeHttpPort + contextPath;`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:87  url = httpConfig.getPublicUrl();`
- Suggested fix: This loop builds one link per Hazelcast member, but when `public-url` is set every member gets the exact same URL, so behind an Ingress a click on node B's log link can be routed to node A and return 404/wrong logs. Also note the else branch appends `contextPath` while this branch does not, and a trailing slash in `publicUrl` will produce `//logs`. Please append `contextPath` with slash normalization here, and address per-node routing (or document the single-node limitation) before we ship this.
- Severity: High

**Issue 3: New `public-url` option is defined but not wired into the YAML config parser, so setting it in seatunnel.yaml has no effect**
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:376`
- Why it matters: - Root cause: this PR adds `PUBLIC_URL` to `ServerConfigOptions.MasterServerConfigOptions` and a `publicUrl` field to `HttpConfig`, but only 3 files are changed. `HttpConfig` is populated from `seatunnel.yaml` by `YamlSeaTunnelDomConfigProcessor` (in `seatunnel-engine-common`), which explicitly matches each child tag under `http` (e.g., `enable-basic-auth`, `context-path`). Without adding a `public-url` branch there, the value the user configures is silently ignored (or rejected as an unknown tag, depending on the processor's strict mode).
- Impact: the exact scenario this PR claims to fix (documented config example `seatunnel.engine.http.public-url`) will not work at runtime on Zeta; `httpConfig.getPublicUrl()` will always return the default and the log links keep using the pod IP.
- Fix: update `YamlSeaTunnelDomConfigProcessor#parseHttpConfig` (or the equivalent DOM processor method) to read `ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL.key()` and call `httpConfig.setPublicUrl(...)`, and add a case to the config parsing test (e.g., `YamlSeaTunnelConfigParserTest`) proving the value round-trips from YAML.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/HttpConfig.java:82  private String publicUrl =
            ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL.defaultValue();`; `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:376  public static final Option<String> PUBLIC_URL = Options.key("public-url")`
- Suggested fix: Defining `PUBLIC_URL` here is not enough — `HttpConfig` is populated from `seatunnel.yaml` by `YamlSeaTunnelDomConfigProcessor`, which matches each child tag under `http` explicitly. Since this PR doesn't touch that processor, configuring `seatunnel.engine.http.public-url` will be ignored and `getPublicUrl()` will always be the default, meaning the fix never activates. Please add the `public-url` branch to the DOM config processor (calling `httpConfig.setPublicUrl(...)` via `PUBLIC_URL.key()`) and cover it in the YAML config parser test.
- Severity: High

**Issue 4: A single `public-url` shared by all cluster members breaks per-node log resolution in multi-node clusters**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:87`
- Why it matters: - Root cause: `allLogNameList` iterates over cluster members and previously built a distinct URL per member (`host:nodeHttpPort`). With `public-url` set, every member's log link collapses to the same base URL, but log files live on each member's local disk (see PR#11464 history — logs are resolved from each node's local log path).
- Impact: in any multi-node Zeta cluster behind an Ingress, the Ingress will route the request to an arbitrary pod, so clicking the log link for member B may return member A's log listing or a 404, and all rows in the Web UI point at identical, non-deterministic links. The fix only works reliably for a single-member cluster.
- Fix: either document clearly that `public-url` is only correct for single-node deployments, or design a node-addressable scheme (e.g., a path/host template like `https://seatunnel.example.com/node/{host}` with sticky routing, or have the REST endpoint proxy the request to the owning member) so each member's logs remain reachable through the proxy.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:87                  url = httpConfig.getPublicUrl();`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:89                  url = "http://" + host + ":" + nodeHttpPort + contextPath;`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:88              } else {`
- Suggested fix: This loop iterates over all cluster members, and the original URL was member-specific (`host:nodeHttpPort`). With `public-url` set, every member's log link becomes identical, but log files are stored on each member's local disk — behind an Ingress the request will hit an arbitrary pod and can return the wrong member's logs or a 404. This only works for single-node clusters; we need either a node-addressable URL scheme (path/host template per member, or server-side proxying to the owning member) or an explicit doc restriction to single-node deployments.
- Severity: High

**Issue 5: New code depends on `StringUtils` and `getPublicUrl()` without the diff adding an import or accessor — confirm both compile paths exist**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:86`
- Why it matters: - Root cause: the changed hunk in `LogService` calls `StringUtils.isNotBlank(...)` but the diff adds no import statement, and `HttpConfig` in this diff only adds a private field with no explicit getter/setter, while the surrounding code shows hand-written accessors (`setPort` with validation).
- Impact: if `LogService` does not already import `org.apache.commons.lang3.StringUtils` (vs. accidentally resolving Hazelcast's shaded `com.hazelcast.internal.util.StringUtil`), or if `HttpConfig` is not Lombok `@Data`-annotated, this PR does not compile; even if it compiles, there is no `setPublicUrl` path shown for the config processor to use.
- Fix: explicitly add `import org.apache.commons.lang3.StringUtils;` in `LogService`, and add an explicit `getPublicUrl()`/`setPublicUrl(String)` pair on `HttpConfig` consistent with the class's existing hand-written accessors, so the field is settable by the YAML parser.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/HttpConfig.java:82-83  private String publicUrl = ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL.defaultValue();`
- Suggested fix: The diff introduces `StringUtils.isNotBlank(...)` without adding an import, and `httpConfig.getPublicUrl()` without an accessor being added to `HttpConfig` in this PR. Please add `import org.apache.commons.lang3.StringUtils;` (make sure it's the commons-lang3 one, not a shaded Hazelcast variant) and explicit `getPublicUrl()/setPublicUrl(String)` on `HttpConfig` matching its existing hand-written setter style, so the field is both readable here and settable by the config processor.
- Severity: Medium

**Issue 6: PR claims "no logic change" but LogService now branches on publicUrl — no unit or e2e test covers either branch**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:90`
- Why it matters: - Root cause: the checklist states "Unit tests not required (config option addition, no logic change)", but the diff adds a conditional in `LogService.allLogNameList` that changes the generated URLs, which is exactly the kind of behavior that regressed before (see PR #11464's NPE in log REST endpoints that required a regression test).
- Impact: without a test, nothing catches that the option is never parsed (finding #1), that `contextPath` is dropped, or future refactors of the URL construction; the fallback path (`publicUrl` blank → pod-IP URL) is also untested for regression.
- Fix: add a unit test for `LogService` (or the URL-building logic extracted into a testable helper) asserting
- (a) links use `publicUrl` + `contextPath` when configured,
- (b) links fall back to `http://host:port + contextPath` when not, and ideally a rest-api e2e case in seatunnel-e2e with `public-url` set in the cluster config.
- Evidence: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:86-90  if (StringUtils.isNotBlank(httpConfig.getPublicUrl())) { url = httpConfig.getPublicUrl(); } else { url = "http://" + host + ":" + nodeHttpPort + contextPath; }`
- Suggested fix: This is a behavior change in link generation, not just a config addition, so I'd like to see a test covering both branches (publicUrl set vs. blank, including contextPath handling). We had a similar untested log-REST path regress in #11464; a small unit test here would have caught that the new option is currently never populated from YAML.
- Severity: Medium

**Issue 7: New user-facing `public-url` option is not documented in the engine docs or config templates**
- Location: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:379`
- Why it matters: - Root cause: this PR introduces a new user-facing config key `seatunnel.engine.http.public-url` but changes no files under `docs/` and no example `seatunnel.yaml` template.
- Impact: users hitting the exact Ingress/reverse-proxy problem in #11662 will not discover the option; the only description lives inside `ServerConfigOptions.java`, which end users don't read. Both English and Chinese docs describe the `http` block today.
- Fix: document `public-url` under the HTTP config section in `docs/en/seatunnel-engine/rest-api-v2.md` (and the corresponding `docs/zh/` page), including the multi-node caveat, and add a commented example to the shipped `config/seatunnel.yaml` template so the YAML shape (`seatunnel.engine.http.public-url`) is unambiguous.
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:376-383  Options.key("public-url").stringType().noDefaultValue().withDescription("The public base URL of the HTTP server...")`
- Suggested fix: Since `public-url` is a new user-facing option, the description here alone isn't enough — please also document it in `docs/en/seatunnel-engine/rest-api-v2.md` (and the `docs/zh/` counterpart) under the `http` config section, and add a commented example to the default `config/seatunnel.yaml` template. Please include the caveat that behind Ingress all member log links resolve to the same base URL.
- Severity: Medium

**Issue 8: Broken indentation in LogService and a stray double blank line in ServerConfigOptions will fail the Spotless check**
- Location: `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:85`
- Why it matters: - Root cause: the added line `                        String url;` uses 24 spaces of indentation while the surrounding block uses 12, and `ServerConfigOptions` gained a stray extra blank line after the new option — both will trip `./mvnw spotless:check` in CI.
- Impact: CI formatting failure blocks the merge; additionally, the new `seatunnel.engine.http.public-url` option is not documented in `docs/en` (the description even says 'Documentation linked (if applicable)' but no docs file is changed), so users won't discover the option. Also confirm `org.apache.commons.lang3.StringUtils` is already imported in `LogService` — the diff adds a `StringUtils` usage but no import line.
- Fix: run `./mvnw spotless:apply -pl seatunnel-engine/seatunnel-engine-server -am`, remove the extra blank line in `ServerConfigOptions`, and add the option to the Zeta engine config docs (e.g., `docs/en/seatunnel-engine/rest-api-v2.md` or the deployment config reference).
- Evidence: `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:384  (extra blank line added after PUBLIC_URL before the HTTP option)`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/LogService.java:85                          String url;`; `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java:384  (extra blank line after PUBLIC_URL option)`
- Suggested fix: The indentation here (`                        String url;`) is off and will fail `spotless:check` in CI. Please run `./mvnw spotless:apply` on the touched modules; that will also clean up the double blank line added in `ServerConfigOptions.java`.
- Severity: Medium

# Review conclusion

**1. Blocking items**
- Issue 1: New `public-url` option is never wired into config parsing, so `httpConfig.getPublicUrl()` will always be null and the feature is dead code
- Issue 2: With `public-url` set, every cluster member's log link collapses to the same URL and `contextPath` is dropped, breaking per-node log access
- Issue 3: New `public-url` option is defined but not wired into the YAML config parser, so setting it in seatunnel.yaml has no effect
- Issue 4: A single `public-url` shared by all cluster members breaks per-node log resolution in multi-node clusters

**2. Remaining follow-ups**
- Issue 5: New code depends on `StringUtils` and `getPublicUrl()` without the diff adding an import or accessor — confirm both compile paths exist
- Issue 6: PR claims "no logic change" but LogService now branches on publicUrl — no unit or e2e test covers either branch
- Issue 7: New user-facing `public-url` option is not documented in the engine docs or config templates
- Issue 8: Broken indentation in LogService and a stray double blank line in ServerConfigOptions will fail the Spotless check

DanielLeens:
Sorry for missing this in my last review — that's a more fundamental problem than what I flagged, and I want to correct the record rather than let my Issue 1 stand as the primary blocker.

I pulled `YamlSeaTunnelDomConfigProcessor.parseHttpConfig()` at the current head (`f2880e893`) to check @SEZ9's claim directly. It's confirmed: the `if`/`else if` chain in `parseHttpConfig()` handles `port`, `context-path`, `enable-http`, `enable-dynamic-port`, `port-range`, `enable-https`, `https-port`, `key-store-path`, etc., but there is no branch for `public-url`/`ServerConfigOptions.MasterServerConfigOptions.PUBLIC_URL` anywhere in that method. That means a user setting `public-url` in `hazelcast.yaml` today has it silently dropped — `httpConfig.getPublicUrl()` is always `null` on the code path this PR relies on, regardless of what the user configures.

That changes the severity picture from my last review. My Issue 1 ("the fix only really works for a single HTTP-serving node, and breaks multi-node discovery") implicitly assumed `public-url` *can* be set and takes effect for at least the single-node case. It can't — the option can never reach `LogService` at all as written, so the feature doesn't work for *any* topology yet, not just multi-node ones. @SEZ9's framing ("as written the feature is dead code") is the more accurate one and should be the lead blocker, with my original multi-node concern folded in underneath it as the next thing to get right once wiring is fixed.

Updated blocker list for this head:
1. (New primary blocker, was missed by me) Wire `public-url` into `YamlSeaTunnelDomConfigProcessor.parseHttpConfig()` so `HttpConfig.getPublicUrl()` actually reflects the configured value.
2. (My original Issue 1, still applies once #1 is fixed) Separate the internal per-member discovery URL from the externally-facing `public-url`-based link, so discovery still targets each member's real `host:port` while only the returned `logLink` uses `public-url` — otherwise every member's link collapses to the same URL and `contextPath` gets dropped, per @SEZ9's trace.
3. My original Issues 2-4 (spotless formatting, missing unit test for the `public-url` branch, docs) still stand once 1 and 2 are addressed.

This is still a solid first PR with the right diagnosis of the underlying problem (#11662) — the wiring gap is a small, mechanical fix, not a design issue. Looking forward to the update, and thanks again @SEZ9 for catching what I missed.

davidzollo:

```

```

PR #11684, opened 2026-08-07
[Fix][Connector-V2] Support Dameng national character types in JDBC connector

zhengxiang378928908-code:
### Purpose of this pull request

Fixes #10635 and #11688.

`DmdbTypeConverter` handles `CHAR`, `CHARACTER`, `VARCHAR` and `VARCHAR2`, but among the national character types it only handles `NVARCHAR`. Reading a Dameng table with `NCHAR` or `NVARCHAR2` columns falls through to the `default` branch and fails:

```
ErrorCode:[COMMON-17], ErrorDescription:['Dameng' unsupported convert type 'nvarchar2' of 'test' to SeaTunnel data type.]
ErrorCode:[COMMON-17], ErrorDescription:['Dameng' unsupported convert type 'nchar' of 'test' to SeaTunnel data type.]
```

In a multi-table job (`table_list`) the catalog aggregates these into the `COMMON-21` error reported in #10635, and the whole job fails before any data is read:

```
ErrorCode:[COMMON-21], ErrorDescription:['Dameng' tables unsupported get catalog table，
the corresponding field types in the following tables are not supported:
'{"test.test_nvarchar2":{"id":"nvarchar2","name":"nvarchar2","age":"nvarchar2"}}'
```

`NVARCHAR2` is the Oracle-compatible synonym of `NVARCHAR`, and `NCHAR` is its fixed-length counterpart. Supporting `NVARCHAR` but neither of the other two is an oversight rather than an intentional restriction, so this PR completes the family in one change.

**Length rule.** Every character type in this converter — `CHAR`, `CHARACTER`, `VARCHAR`, `VARCHAR2` and `NVARCHAR` — uses `charTo4ByteLength`. Both new types follow the same rule, so no new length behaviour is introduced and no assumption is made about how Dameng reports lengths. That matters here, because Dameng's `LENGTH_IN_CHAR` parameter makes byte-vs-character length a database-level setting.

**Source type.** `NVARCHAR2` shares the `NVARCHAR` branch but `sourceType` is now built from the matched type name rather than the hard-coded constant, so `NVARCHAR` keeps reporting `NVARCHAR(n)` exactly as before. `NCHAR` gets its own case rather than joining `CHAR`/`CHARACTER`, because that branch rewrites `sourceType` to the `CHAR` constant and would report `NCHAR` columns as `CHAR(n)`.

[...]

PR #11689, opened 2026-08-07
[Fix][Connector-V2] Support Dameng NCHAR type conversion in JDBC connector

### LINKED ISSUE #11688: [Bug] [connector-jdbc] Dameng NCHAR type is not supported by the type converter
### Search before asking

### What happened

`DmdbTypeConverter` has cases for `CHAR`, `CHARACTER`, `VARCHAR`, `VARCHAR2` and `NVARCHAR`, but no case for `NCHAR`. Reading a Dameng table that contains `NCHAR` columns falls through to the `default` branch and throws:

```
ErrorCode:[COMMON-17], ErrorDescription:['Dameng' unsupported convert type 'nchar' of 'test' to SeaTunnel data type.]
```

In a multi-table job the catalog aggregates these into `COMMON-21` and the job fails before any data is read.

`NCHAR` is a standard Dameng type — the fixed-length national character type, and the counterpart of the already supported variable-length `NVARCHAR`. Supporting `NVARCHAR` but not `NCHAR` looks like an oversight rather than an intentional restriction.

This is the same class of gap as #10635 (`NVARCHAR2`). It likely went unnoticed because `JdbcDmIT` covers `CHAR`, `CHARACTER`, `VARCHAR` and `VARCHAR2` but no national character type at all.

### SeaTunnel Version

dev

### SeaTunnel Config

```conf

DmdbTypeConverter.INSTANCE.convert(
    BasicTypeDefine.builder()
        .name("test").columnType("nchar(2)").dataType("nchar").length(2L).build());
```

### Running Command

```shell
./mvnw test -pl seatunnel-connectors-v2/connector-jdbc -Dtest=DmdbTypeConverterTest
```

### Error Exception

```log
ErrorDescription:['Dameng' unsupported convert type 'nchar' of 'test' to SeaTunnel data type.]
```

Zeta

### Java or Scala Version

Java 8

### Screenshots

_No response_

### Code of Conduct

zhengxiang378928908-code:

Fixes #11688.

`DmdbTypeConverter` has cases for `CHAR`, `CHARACTER`, `VARCHAR`, `VARCHAR2` and `NVARCHAR`, but no case for `NCHAR`. Reading a Dameng table containing `NCHAR` columns falls through to the `default` branch and fails:

```
ErrorCode:[COMMON-17], ErrorDescription:['Dameng' unsupported convert type 'nchar' of 'test' to SeaTunnel data type.]
```

In a multi-table job the catalog aggregates these into `COMMON-21` and the whole job fails before any data is read.

`NCHAR` is the fixed-length national character type and the counterpart of the already supported variable-length `NVARCHAR`. Supporting one but not the other looks like an oversight rather than an intentional restriction.

**Length rule.** Every character type in this converter — `CHAR`, `CHARACTER`, `VARCHAR`, `VARCHAR2` and `NVARCHAR` — uses `charTo4ByteLength`. `NCHAR` follows the same rule, so the change introduces no new length behaviour and needs no assumption about how Dameng reports lengths. That matters here because Dameng's `LENGTH_IN_CHAR` parameter makes byte-vs-character length a database-level setting, so I deliberately avoided reasoning that depends on it.

**Source type.** `NCHAR` gets its own case rather than joining the `CHAR`/`CHARACTER` branch, because that branch rewrites `sourceType` to the `CHAR` constant. Sharing it would report `NCHAR` columns as `CHAR(n)` and lose the national character type name.

Yes, in the sense that a previously failing configuration now works: Dameng tables containing `NCHAR` columns can now be read by the JDBC source, and those columns map to SeaTunnel `STRING`.

No config option, default value, or existing type mapping changes. The added case is purely additive, so this is backward compatible.

**Unit test.** Added `DmdbTypeConverterTest#testNchar`, mirroring the existing `testNvarchar`, asserting the mapped data type, column length and source type.

```
./mvnw test -pl seatunnel-connectors-v2/connector-jdbc -Dtest=DmdbTypeConverterTest
Tests run: 35, Failures: 0, Errors: 0, Skipped: 0
```

Reverting only the `DmdbTypeConverter` change makes it fail with the reported error, confirming it is a genuine regression test:

```
[ERROR] testNchar  Time elapsed: 0.09 s  <<< ERROR!
ErrorDescription:['Dameng' unsupported convert type 'nchar' of 'test' to SeaTunnel data type.]
```

**E2E test.** Added a `DM_NCHAR NCHAR(50)` column to `JdbcDmIT`, together with the matching `fieldNames` entry, row value and the explicit insert column list in `jdbc_dm_source_and_sink.conf`. The column is populated with multi-byte content so the national character path is exercised end to end. `JdbcDmUpsetIT` declares its own table and is unaffected.

I could not run this suite locally — Docker is unavailable in my environment — so the `laglangyue/dmdb8` container path is validated by CI rather than by me. `./mvnw test-compile` on `connector-jdbc-e2e-part-5` passes, and the `CREATE_SQL` columns, `fieldNames`, row array, insert column list and placeholder count are all consistent at 34 entries.

`./mvnw spotless:apply` was run over both modules.

### Note on overlap with #11684

#11684 adds `NVARCHAR2` support and also touches `JdbcDmIT` and `jdbc_dm_source_and_sink.conf`. The converter changes sit in different branches of the switch and do not overlap, but the two PRs both extend the same E2E table, so whichever merges second will need a trivial rebase. I'm happy to rebase this one whenever #11684 lands — or to combine them if a maintainer would prefer a single PR covering the whole national character family.

### Check list

  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md) — no new dependency
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs — not applicable, no option or documented behaviour changed
* [x] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR — not applicable, change is backward compatible
* [x] If you are contributing the connector code, please check that the following files are updated — not applicable, no new connector; E2E coverage added to the existing `JdbcDmIT`

### DIFF
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverter.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverter.java
@@ -64,6 +64,7 @@ public class DmdbTypeConverter implements TypeConverter<BasicTypeDefine> {
     public static final String DM_CHAR = "CHAR";
 
     public static final String DM_CHARACTER = "CHARACTER";
+    public static final String DM_NCHAR = "NCHAR";
     public static final String DM_VARCHAR = "VARCHAR";
     public static final String DM_VARCHAR2 = "VARCHAR2";
     public static final String DM_NVARCHAR = "NVARCHAR";
@@ -194,6 +195,13 @@ public class DmdbTypeConverter implements TypeConverter<BasicTypeDefine> {
             case DM_CHAR:
             case DM_CHARACTER:
                 builder.sourceType(String.format("%s(%s)", DM_CHAR, typeDefine.getLength()));
+                builder.dataType(BasicType.STRING_TYPE);
+                builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
+                break;
+            case DM_NCHAR:
+                // NCHAR is the fixed-length national character type, the counterpart of NVARCHAR.
+                // Keep the declared type name so it is not reported as CHAR.
+                builder.sourceType(String.format("%s(%s)", DM_NCHAR, typeDefine.getLength()));
                 builder.dataType(BasicType.STRING_TYPE);
                 builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
                 break;
--- a/seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverterTest.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverterTest.java
@@ -347,6 +347,22 @@ public class DmdbTypeConverterTest {
     }
 
     @Test
+    public void testNchar() {
+        BasicTypeDefine<Object> typeDefine =
+                BasicTypeDefine.builder()
+                        .name("test")
+                        .columnType("nchar(2)")
+                        .dataType("nchar")
+                        .length(2L)
+                        .build();
+        Column column = DmdbTypeConverter.INSTANCE.convert(typeDefine);
+        Assertions.assertEquals(typeDefine.getName(), column.getName());
+        Assertions.assertEquals(BasicType.STRING_TYPE, column.getDataType());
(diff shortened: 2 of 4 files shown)


zhengxiang378928908-code:
Closing this in favour of #11684.

Both PRs fixed missing Dameng national character types, and they overlapped in the same `JdbcDmIT` table and insert column list, so whichever merged second would have needed a rebase. Since `NCHAR` and `NVARCHAR2` are the same gap in the same type family and the same converter, keeping them apart added review overhead without buying anything.

#11684 now covers `NVARCHAR` (unchanged), `NVARCHAR2` and `NCHAR` together, and closes both #10635 and #11688. The `NCHAR` change is carried over unmodified — same converter case, same unit test, same E2E column.

Sorry for the churn; I should have scoped it that way from the start.

PR #11691, opened 2026-08-07
[Docs] Document DolphinScheduler integration with Docker Compose deployments

### LINKED ISSUE #11279: [Docs] Document DolphinScheduler integration with Docker Compose deployments
## Background

The FAQ mentions that users can use scheduling tools such as Apache DolphinScheduler to manage SeaTunnel jobs. Users who deploy SeaTunnel through Docker Compose still need practical guidance for integration details, especially the meaning of `SEATUNNEL_HOME` when the scheduler runs outside or inside a container.

A common question is: when SeaTunnel is deployed with Docker Compose, what SeaTunnel home path should be configured in DolphinScheduler?

## Goal

Document how to configure DolphinScheduler to submit SeaTunnel jobs when SeaTunnel is deployed with Docker Compose or containerized paths.

## Documentation location

This content should be placed under the practical tools / ecosystem section, not under connector docs.

Suggested files:

- `docs/en/tools/overview.md` or a new page under `docs/en/tools/`
- `docs/zh/tools/overview.md` or a new page under `docs/zh/tools/`

If a new page is added, please also wire it into the relevant sidebar.

## Suggested scope

The update should explain:

1. What `SEATUNNEL_HOME` means for a scheduler integration.
2. The difference between host paths and container paths.
3. How to decide the correct path when DolphinScheduler runs on the host.
4. How to decide the correct path when DolphinScheduler also runs in Docker or Kubernetes.
5. A minimal example using the common SeaTunnel container path, such as `/opt/seatunnel`, if that is accurate for the current Docker image.
6. Any required volume mounts or network assumptions.

## Acceptance criteria

- The docs give a direct answer for Docker Compose deployments.
- The docs clearly distinguish host-side and container-side paths.
- The content is added under the practical tools / ecosystem docs section.
- English and Chinese docs are consistent.
- The example is checked against the current SeaTunnel Docker image or Docker Compose deployment docs.

jinniiii233:
This PR adds documentation explaining how to configure DolphinScheduler to submit SeaTunnel jobs when SeaTunnel is deployed via Docker Compose, including:

- What `SEATUNNEL_HOME` means for scheduler integration
- The difference between host paths and container paths
- How to decide the correct path when DolphinScheduler runs on the host
- How to decide the correct path when DolphinScheduler also runs in Docker/Kubernetes
- A minimal working example using the common container path `/opt/seatunnel`
- Required volume mounts and network assumptions

Both English (`docs/en/tools/dolphinscheduler-integration.md`) and Chinese (`docs/zh/tools/dolphinscheduler-integration.md`) versions are added.

Closes #11279

### Check list

### DIFF
--- a/docs/en/tools/dolphinscheduler-integration.md
+++ b/docs/en/tools/dolphinscheduler-integration.md
@@ -1,0 +1,82 @@
+# Integrating SeaTunnel with DolphinScheduler via Docker Compose
+
+A common question when deploying SeaTunnel through Docker Compose is: which `SEATUNNEL_HOME` path should DolphinScheduler be configured with so it can correctly submit SeaTunnel jobs? This document explains the reasoning and provides a minimal, ready-to-use example.
+
+## 1. What `SEATUNNEL_HOME` Means for Scheduler Integration
+
+DolphinScheduler does not call SeaTunnel over a network API. Instead, it directly executes SeaTunnel's launch script on the same machine (or container) it runs on, for example:
+
+```
+${SEATUNNEL_HOME}/bin/seatunnel.sh --config <job_config_path>
+```
+
+This means `SEATUNNEL_HOME` must point to a path **that the DolphinScheduler process itself can actually reach**, and that path must contain a complete SeaTunnel installation (`bin/`, `config/`, `lib/`, `connectors/`, etc.). If the path is empty or missing, DolphinScheduler will fail at job execution time with a "script not found" or "command not found" error.
+
+## 2. Host Paths vs. Container Paths
+
+This is the most common source of confusion. There are two distinct "points of view":
+
+- **Host path**: where the SeaTunnel installation actually lives on your machine or server's disk, e.g. `/home/user/seatunnel` or `./seatunnel`.
+- **Container path**: what DolphinScheduler's container sees internally, e.g. `/opt/seatunnel`. Whether files exist at that path depends entirely on whether you mounted a host directory into it via `volumes`.
+
+**Key rule**: `SEATUNNEL_HOME` always describes the path as seen **from the environment DolphinScheduler's process runs in**, not from your local machine. If DolphinScheduler runs inside a container, `SEATUNNEL_HOME` must be a container-internal path (e.g. `/opt/seatunnel`), and that container path must be mounted to the host directory where SeaTunnel is actually installed.
+
+## 3. Deciding the Path When DolphinScheduler Runs on the Host
+
+If DolphinScheduler is not containerized and runs directly as a process on a physical or virtual machine:
+
+- `SEATUNNEL_HOME` should be set to the real host-side installation path, e.g. `/opt/module/seatunnel`.
+- There is no volume-mounting concern here — DolphinScheduler's process and the SeaTunnel installation share the same filesystem, so the path is exactly what it appears to be.
+
+## 4. Deciding the Path When DolphinScheduler Also Runs in Docker or Kubernetes
+
+This is the case most people hit with Docker Compose. Two things are required:
+
+1. Mount the host's SeaTunnel installation directory into a path inside the DolphinScheduler container via `volumes`.
+2. Set `SEATUNNEL_HOME` to the **mounted container-side path**, not the host path.
+
+The same logic applies in Kubernetes: "mounting" becomes `volumeMounts` combined with a `PersistentVolume` (or `hostPath`), but the principle is identical — `SEATUNNEL_HOME` should be set to the path as seen inside the Pod.
+
+## 5. Minimal Working Example (Using the Common Container Path `/opt/seatunnel`)
+
+```yaml
+version: '3.8'
+
+services:
+  dolphinscheduler:
+    image: apache/dolphinscheduler-standalone-server:3.2.1
+    container_name: dolphinscheduler
+    hostname: dolphinscheduler
+    ports:
+      - "12345:12345"
+    environment:
+      - SEATUNNEL_HOME=/opt/seatunnel
+    volumes:
+      # host ./seatunnel -> container /opt/seatunnel
+      # the host directory must contain a full SeaTunnel installation beforehand
+      - ./seatunnel:/opt/seatunnel:ro
+      - ./dolphinscheduler/logs:/opt/dolphinscheduler/logs
+    networks:
+      - ds-network
+
+networks:
+  ds-network:
+    driver: bridge
+```
+
+Prepare the host-side SeaTunnel installation directory (the official DolphinScheduler image does not bundle SeaTunnel itself — it must be downloaded and extracted manually):
+
+```bash
+mkdir -p seatunnel dolphinscheduler/logs
+wget https://archive.apache.org/dist/seatunnel/2.3.4/apache-seatunnel-2.3.4-bin.tar.gz
+tar -zxvf apache-seatunnel-2.3.4-bin.tar.gz -C seatunnel --strip-components=1
+```
+
+After this, `./seatunnel` should contain `bin/`, `config/`, `lib/`, and similar subdirectories, and the same content will be visible inside the container at `/opt/seatunnel` once it starts.
+
+## 6. Required Volume Mounts and Network Assumptions
+
+- **Required mount**: the host's SeaTunnel installation directory → the container path referenced by `SEATUNNEL_HOME`. A read-only mount (`:ro`) is sufficient, since DolphinScheduler only needs to execute scripts, not write to the installation directory.
+- **Recommended mounts**: DolphinScheduler's log directory and any directory holding job configuration files, so logs and configs remain accessible outside the container for debugging and version control.
+- **Network assumption**: if SeaTunnel jobs need to reach data sources (databases, message queues, etc.) that are also deployed via Docker Compose, they must be on the same custom network as DolphinScheduler (e.g. `ds-network` above); otherwise containers cannot resolve each other by service name and you'll need to fall back to host IPs or additional network configuration.
+- If SeaTunnel jobs need to reach services on the host machine that are outside the Docker network, be aware of Docker's default network isolation — you may need `host.docker.internal` (on Docker Desktop) or explicit host-network configuration.
--- a/docs/en/tools/dolphinscheduler-integration.md
+++ b/docs/en/tools/dolphinscheduler-integration.md
@@ -67,9 +67,17 @@ networks:
 Prepare the host-side SeaTunnel installation directory (the official DolphinScheduler image does not bundle SeaTunnel itself — it must be downloaded and extracted manually):
 
 ```bash
+export version="3.0.0"
 mkdir -p seatunnel dolphinscheduler/logs
-wget https://archive.apache.org/dist/seatunnel/2.3.4/apache-seatunnel-2.3.4-bin.tar.gz
-tar -zxvf apache-seatunnel-2.3.4-bin.tar.gz -C seatunnel --strip-components=1
+wget "https://archive.apache.org/dist/seatunnel/${version}/apache-seatunnel-${version}-bin.tar.gz"
+tar -zxvf "apache-seatunnel-${version}-bin.tar.gz" -C seatunnel --strip-components=1
+```
+:::caution Warning
+Since 2.2.0-beta, connector plugins are no longer bundled by default. You must install them **before** mounting the directory read-only and starting the container, otherwise every submitted job will fail with a missing-connector error.
+:::
+
+```bash
+sh seatunnel/bin/install-plugin.sh 3.0.0
 ```
 
 After this, `./seatunnel` should contain `bin/`, `config/`, `lib/`, and similar subdirectories, and the same content will be visible inside the container at `/opt/seatunnel` once it starts.
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -328,7 +328,8 @@ const sidebars = {
                 "tools/overview",
                 "tools/seatunnel-skill",
                 "tools/seatunnel-mcp",
-                "tools/x2seatunnel"
+                "tools/x2seatunnel",
+                "tools/dolphinscheduler-integration"
             ]
         },
         {
--- a/docs/zh/tools/dolphinscheduler-integration.md
+++ b/docs/zh/tools/dolphinscheduler-integration.md
@@ -67,9 +67,17 @@ networks:
 准备宿主机上的 SeaTunnel 安装目录（Docker 官方镜像不自带 SeaTunnel 程序本体，需要手动下载解压）：
 
 ```bash
+export version="3.0.0"
 mkdir -p seatunnel dolphinscheduler/logs
-wget https://archive.apache.org/dist/seatunnel/2.3.4/apache-seatunnel-2.3.4-bin.tar.gz
-tar -zxvf apache-seatunnel-2.3.4-bin.tar.gz -C seatunnel --strip-components=1
+wget "https://archive.apache.org/dist/seatunnel/${version}/apache-seatunnel-${version}-bin.tar.gz"
+tar -zxvf "apache-seatunnel-${version}-bin.tar.gz" -C seatunnel --strip-components=1
+```
+:::caution 警告
+从 2.2.0-beta 版本开始，连接器插件默认不再随安装包一起打包。你必须在把目录以只读方式挂载并启动容器**之前**安装好这些插件，否则所有提交的作业都会因为缺少连接器而失败。
+:::
+
+```bash
+sh seatunnel/bin/install-plugin.sh 3.0.0
 ```
 
 完成后，`./seatunnel` 目录下应包含 `bin/`、`config/`、`lib/` 等子目录，容器启动后即可在 `/opt/seatunnel` 下看到同样的内容。
--- a/docs/zh/tools/dolphinscheduler-integration.md
+++ b/docs/zh/tools/dolphinscheduler-integration.md
@@ -1,0 +1,82 @@
+# 使用 Docker Compose 将 SeaTunnel 与 DolphinScheduler 集成
+
+当 SeaTunnel 通过 Docker Compose 部署时，一个常见的问题是：DolphinScheduler 应该配置哪个 `SEATUNNEL_HOME` 路径才能正确提交作业？本文档说明这个问题背后的原理，并给出一个可直接使用的最小示例。
+
+## 1. `SEATUNNEL_HOME` 对调度器集成意味着什么
+
+DolphinScheduler 并不是通过网络接口远程调用 SeaTunnel，而是在自己所在的机器（或容器）上，直接执行 SeaTunnel 提供的启动脚本，例如：
+
+```
+${SEATUNNEL_HOME}/bin/seatunnel.sh --config <job_config_path>
+```
+
+因此 `SEATUNNEL_HOME` 必须指向 **DolphinScheduler 进程实际能够访问到的一个路径**，这个路径下必须包含完整的 SeaTunnel 安装目录（`bin/`、`config/`、`lib/`、`connectors/` 等）。如果这个路径下是空的或者不存在，DolphinScheduler 执行任务时会直接报"找不到脚本"或"命令不存在"的错误。
+
+## 2. 宿主机路径 与 容器内路径 的区别
+
+这是最容易搞混的地方，需要分清两个"视角"：
+
+- **宿主机路径（Host Path）**：SeaTunnel 安装包实际存放在你电脑或服务器磁盘上的位置，例如 `/home/user/seatunnel` 或 `./seatunnel`。
+- **容器内路径（Container Path）**：DolphinScheduler 容器内部看到的路径，例如 `/opt/seatunnel`。这个路径是否存在文件，取决于你有没有通过 `volumes` 把宿主机目录挂载进去。
+
+**关键点**：`SEATUNNEL_HOME` 这个环境变量，配置的永远是 **DolphinScheduler 进程自己所在环境看到的路径**，而不是你电脑上的路径。如果 DolphinScheduler 跑在容器里，`SEATUNNEL_HOME` 就必须写容器内路径（如 `/opt/seatunnel`），并确保这个容器路径通过 volume 挂载对应到了宿主机上真正装了 SeaTunnel 的目录。
+
+## 3. DolphinScheduler 运行在宿主机上时，如何确定路径
+
+如果你没有用容器运行 DolphinScheduler，而是直接在物理机 / 虚拟机上以进程方式运行它，那么：
+
+- `SEATUNNEL_HOME` 直接填宿主机上 SeaTunnel 的真实安装路径，例如 `/opt/module/seatunnel`。
+- 不涉及任何 volume 挂载问题，DolphinScheduler 进程和 SeaTunnel 安装包在同一套文件系统里，路径所见即所得。
+
+## 4. DolphinScheduler 也运行在 Docker（或 Kubernetes）中时，如何确定路径
+
+这是 Docker Compose 场景下最常遇到的情况。此时必须做两件事：
+
+1. 把宿主机上的 SeaTunnel 安装目录，通过 `volumes` 挂载到 DolphinScheduler 容器内的某个路径。
+2. 把 `SEATUNNEL_HOME` 设置为**挂载后的容器内路径**，而不是宿主机路径。
+
+在 Kubernetes 环境下同理，只是"挂载"变成了 `volumeMounts` + `PersistentVolume`（或 `hostPath`），思路完全一致：Pod 内部看到的路径才是 `SEATUNNEL_HOME` 应该填的值。
+
+## 5. 最小可用示例（使用常见容器路径 `/opt/seatunnel`）
+
+```yaml
+version: '3.8'
+
+services:
+  dolphinscheduler:
+    image: apache/dolphinscheduler-standalone-server:3.2.1
+    container_name: dolphinscheduler
+    hostname: dolphinscheduler
+    ports:
+      - "12345:12345"
+    environment:
+      - SEATUNNEL_HOME=/opt/seatunnel
+    volumes:
+      # 宿主机 ./seatunnel 目录 -> 容器内 /opt/seatunnel
+      # 宿主机这个目录下需要预先放好完整的 SeaTunnel 安装包内容
+      - ./seatunnel:/opt/seatunnel:ro
+      - ./dolphinscheduler/logs:/opt/dolphinscheduler/logs
+    networks:
+      - ds-network
+
+networks:
+  ds-network:
+    driver: bridge
+```
+
+准备宿主机上的 SeaTunnel 安装目录（Docker 官方镜像不自带 SeaTunnel 程序本体，需要手动下载解压）：
+
+```bash
+mkdir -p seatunnel dolphinscheduler/logs
+wget https://archive.apache.org/dist/seatunnel/2.3.4/apache-seatunnel-2.3.4-bin.tar.gz
+tar -zxvf apache-seatunnel-2.3.4-bin.tar.gz -C seatunnel --strip-components=1
+```
+
+完成后，`./seatunnel` 目录下应包含 `bin/`、`config/`、`lib/` 等子目录，容器启动后即可在 `/opt/seatunnel` 下看到同样的内容。
+
+## 6. 所需的 Volume 挂载与网络假设
+
+- **必须挂载**：宿主机的 SeaTunnel 安装目录 → 容器内 `SEATUNNEL_HOME` 指向的路径（只读挂载 `:ro` 即可，因为 DolphinScheduler 只需要执行脚本，不需要写入安装目录）。
+- **建议挂载**：DolphinScheduler 的日志目录、以及存放作业配置文件（job config）的目录，方便在容器外查看任务配置和日志，也便于版本管理和排查问题。
+- **网络假设**：如果 SeaTunnel 需要连接的数据源（数据库、消息队列等）也用 Docker Compose 部署，需确保它们和 DolphinScheduler 处于同一个自定义网络（如上例中的 `ds-network`），否则容器间无法通过服务名互相访问，必须改用宿主机 IP 或额外的网络配置。
+- 如果 SeaTunnel 任务需要访问宿主机上的其他服务（不在 Docker 网络内），需要注意 Docker 默认的网络隔离，可能需要使用 `host.docker.internal`（Docker Desktop 环境）或显式配置宿主机网络访问。


DanielLeens (changes requested):
Thanks for taking on issue #11279 — this is a genuinely useful doc to have. The `SEATUNNEL_HOME` host-vs-container confusion is a real and common source of pain, and the explanation you've written for it is clear and well-organized. I dug into the actual repo scripts and other docs to sanity-check the claims below; overall the technical content is sound, but there are two concrete gaps worth closing before this merges.

This PR adds a new documentation page (English + Chinese) explaining how to configure `SEATUNNEL_HOME` for DolphinScheduler when SeaTunnel itself is deployed via Docker Compose. It clarifies the host-path-vs-container-path distinction, gives guidance for both "DolphinScheduler on host" and "DolphinScheduler in Docker/Kubernetes" scenarios, and provides a minimal working `docker-compose.yaml` example plus the shell commands to prepare the SeaTunnel installation directory. It closes #11279.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Two new files are added, no code is touched:
- `docs/en/tools/dolphinscheduler-integration.md` (82 lines, new)
- `docs/zh/tools/dolphinscheduler-integration.md` (82 lines, new, faithful translation of the English version)

Content verification against actual repo source:
- **`SEATUNNEL_HOME` mechanics:** The doc's premise is that DolphinScheduler invokes `${SEATUNNEL_HOME}/bin/seatunnel.sh --config <path>` directly on its own host/container filesystem. Confirmed `--config`/`-c` is a real CLI option (`seatunnel-core/seatunnel-core-starter/src/main/java/.../AbstractCommandArgs.java:36-38`). I can't verify DolphinScheduler's own internals from this repo, so treat that part as a reasonable assumption rather than something verified in-repo.
- **Container path `/opt/seatunnel`:** Matches the official image — `docs/en/getting-started/docker/docker.md` shows the Dockerfile ends with `WORKDIR /opt/seatunnel`.
- **`docker-compose.yaml` syntax:** Consistent with the existing compose example in `docs/en/getting-started/docker/docker.md` (`version: '3.8'`, `environment:` list syntax, `networks:` block).
- **Download command:** `wget https://archive.apache.org/dist/seatunnel/2.3.4/apache-seatunnel-2.3.4-bin.tar.gz` — the URL pattern matches `docs/en/getting-started/locally/deployment.md` and `docs/en/engines/zeta/download-seatunnel.md`, but see Issue 3 on the hardcoded version.

## 1.2 Compatibility Impact

Fully compatible — docs-only change, no source code, config schema, SPI, or public API touched.

N/A — no runtime code added or modified.

Docs-only, but since this page's whole purpose is to walk an operator through a real deployment, I traced through what actually happens if the example is followed verbatim — see Issue 2.

**Issue 1: New `tools/` page isn't wired into the docs sidebar**
- **Location:** `docs/sidebars.js` (not touched by this PR); new files `docs/en/tools/dolphinscheduler-integration.md`, `docs/zh/tools/dolphinscheduler-integration.md`
- **Problem description:** `docs/sidebars.js` has no `tools` category at all, and this PR doesn't add one. Every other doc directory (`introduction`, `architecture`, `getting-started`, `connectors`, `transforms`, `engines`, `developer`) is explicitly wired into `docs/sidebars.js`; this new `tools/` directory is not referenced anywhere.
- **Potential risk:** Once merged, the page exists in the repo but has no entry point in the published docs site navigation — effectively invisible unless a user knows the direct URL. This directly contradicts issue #11279's own acceptance criteria, which explicitly ask for the new page to be wired into the relevant sidebar.
- **Best improvement:** Add a `Tools` category to `docs/sidebars.js` pointing at `tools/dolphinscheduler-integration` (one `sidebars.js` drives both `docs/en` and `docs/zh` via Docusaurus i18n).
- **Severity:** High
- **Raised by another reviewer:** No (no prior reviews exist on this PR)

**Issue 2: Example setup omits the required connector-installation step; job submission would fail as written**
- **Location:** `docs/en/tools/dolphinscheduler-integration.md` §5 (and mirrored §5 in `docs/zh/...`), the `wget` / `tar -zxvf` block
- **Problem description:** The example only downloads and extracts the SeaTunnel binary tarball; it never runs `bin/install-plugin.sh`. Per `docs/en/getting-started/locally/deployment.md`, connector dependencies are no longer bundled by default since 2.2.0-beta and must be installed via that script — confirmed `bin/install-plugin.sh` itself creates `${SEATUNNEL_HOME}/connectors` and populates it from `config/plugin_config` (lines 24-51).
- **Potential risk:** Following this guide exactly as written produces a setup where every submitted job fails, including the trivial fake-to-console example, because no connector plugins are installed — and the doc gives no pointer back to this fix.
- **Best improvement:** Add `sh seatunnel/bin/install-plugin.sh 2.3.4` to the setup steps in §5, ideally with a `:::caution` admonition (matching `docs/en/developer/docs-format-specification.md` convention) noting connectors must be installed before mounting `:ro` and starting the container.
- **Severity:** High

**Issue 3: Hardcoded, outdated SeaTunnel version in the download command**
- **Location:** `docs/en/tools/dolphinscheduler-integration.md` §5 / zh mirror
- **Problem description:** Hardcodes `apache-seatunnel-2.3.4-bin.tar.gz`, while the repo's `pom.xml` is currently on `2.3.13-SNAPSHOT`. Other docs (`deployment.md`, `download-seatunnel.md`) avoid this via an `export version="2.3.13"` + `${version}` pattern.
- **Potential risk:** Low immediate risk (URL still resolves via the archive), but nudges new users toward an old release and will look increasingly stale.
- **Best improvement:** Switch to the `export version=...` pattern, or link to `https://seatunnel.apache.org/download`.
- **Severity:** Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

Markdown structure, heading style, fenced code blocks with language hints, and prose tone are consistent with the rest of `docs/en`. The doc doesn't use the project's `:::tip`/`:::caution` admonition syntax despite having genuinely caution-worthy content (the read-only mount + missing-connectors interaction) — a style nit, not a blocker on its own, but ties directly into Issue 2's fix.

N/A — documentation-only PR, no testable code path.

## 2.3 Documentation Updates

This PR is the documentation update, and it correctly ships both `docs/en/tools/dolphinscheduler-integration.md` and `docs/zh/tools/dolphinscheduler-integration.md` together. I compared the two files section-by-section — the Chinese version is a faithful, complete translation, not a partial or divergent copy. The EN/ZH-mirror requirement is satisfied; the actual gap is the sidebar wiring in Issue 1.

# 3. Architectural Soundness

The doc's core teaching device — separating "host path" vs "container path" as two distinct points of view — is a clean, minimal way to explain a genuinely confusing topic. Classified as a precise, well-scoped documentation fix.

## 3.2 Maintainability
Low-risk as a static Markdown page. The one wrinkle is the hardcoded version number (Issue 3), which will need manual bumps where other docs use a copy-pasteable variable instead.

## 3.3 Extensibility
Issue #11279 frames this as the first page under a new `tools/` (practical tools / ecosystem) section, implying more integration guides are expected to land here later. Fixing the sidebar wiring now (Issue 1) sets the right precedent for the section going forward.

N/A in the traditional sense (no runtime/config/serialization changes). The doc doesn't state which SeaTunnel/DolphinScheduler version ranges it applies to — a one-line "applies to SeaTunnel 2.2.0-beta+" note would help, but this is optional polish, not a blocker.

# 4. Issue Summary

| 1 | New `tools/` page isn't wired into the sidebar, contradicting #11279's own acceptance criteria; page is unreachable from docs site nav | `docs/sidebars.js` | High |
| 2 | Example setup skips `bin/install-plugin.sh`; every job submitted through this setup would fail on missing connectors, with no troubleshooting note | `docs/en/tools/dolphinscheduler-integration.md` §5 + zh mirror | High |
| 3 | Hardcoded outdated version `2.3.4` instead of the `${version}` pattern used elsewhere | `docs/en/tools/dolphinscheduler-integration.md` §5 + zh mirror | Low |

# 5. Merge Recommendation

   - Issue 1: wire the new page into `docs/sidebars.js` so it's actually reachable from the published docs site.
   - Issue 2: add the `bin/install-plugin.sh` step to the §5 example so the walkthrough is actually runnable end-to-end without every job failing on missing connectors.

   - Issue 3: switch the hardcoded `2.3.4` to the `${version}` variable pattern used in `deployment.md`/`download-seatunnel.md`.
   - Consider using `:::tip`/`:::caution` admonitions for the cautionary content, per this repo's docs-format convention.

**Overall assessment:** the writing and technical reasoning behind this doc are solid — the hardest conceptual part (host vs. container path) is explained correctly, which is the part most people get wrong. Issues 1 and 2 are concrete, verifiable gaps (an unreachable page, and an example that fails as written) rather than nitpicks, so I'd like those closed before merge. Once fixed, this should be in great shape — looking forward to the update!

jinniiii233:
Hi @DanielLeens, thanks for the thorough review! I've addressed all three points:

1. Wired the new page into `docs/sidebars.js` under the existing "Tools" category.
2. Added the missing `bin/install-plugin.sh` step (with a `:::caution` admonition) to both the EN and ZH §5 examples.
3. Switched the hardcoded `2.3.4` version to the `${version}` variable pattern used elsewhere.

Ready for another look whenever you have time!

Note: the "Dead links" check currently fails on `docs/en/transforms/transform-and-join-boundary.md`, which is unrelated to this PR's changes — appears to be a pre-existing broken link in the repo.

DanielLeens (approved):
Re-reviewed the latest head (`0dc9e62ac0`) from scratch after the new commit, as required. Both blockers from round 1 are genuinely fixed — verified directly against the diff and the surrounding codebase, not just the commit message.

**Small correction to my own round-1 wording first:** I wrote that "`docs/sidebars.js` has no `tools` category at all." That was inaccurate — I checked the base commit you started from, and a `Tools` category (with `tools/overview`, `tools/seatunnel-skill`, `tools/seatunnel-mcp`, `tools/x2seatunnel`) already existed. The actual gap — which is what mattered for the outcome — was that this PR's new page wasn't added as an item under that existing category, so it was still unreachable from the nav. Sorry for the imprecise description; the fix you shipped is exactly right regardless.

**Issue 1 (High, sidebar wiring) — fixed, verified.** `docs/sidebars.js:331` now adds `"tools/dolphinscheduler-integration"` as a fifth item in the existing `Tools` category. Since one `sidebars.js` drives both the `docs/en` and `docs/zh` Docusaurus i18n trees, this wires up both language versions with a single change, as expected.

**Issue 2 (High, missing install-plugin step) — fixed, verified.** Both `docs/en/tools/dolphinscheduler-integration.md` and the `docs/zh` mirror now insert `sh seatunnel/bin/install-plugin.sh 3.0.0` between the extract step and the "should contain bin/, config/, lib/" sentence, wrapped in a `:::caution Warning` / `:::caution 警告` admonition explaining that connector plugins must be installed before the directory is mounted read-only. The admonition syntax (`:::caution Warning`) matches the exact convention shown in `docs/en/developer/docs-format-specification.md:25`, which is the convention I pointed to last round. Following the guide top-to-bottom now produces a setup where the trivial fake-to-console example would actually succeed instead of failing on a missing connector.

**Issue 3 (Low, hardcoded version) — fixed, verified.** The download block now uses `export version="3.0.0"` plus `${version}` substitution in both the `wget` and `tar` commands (en/zh), matching the exact pattern used in `docs/en/getting-started/locally/deployment.md:23-26` and `docs/en/engines/zeta/download-seatunnel.md:24`. I checked the repo's current `revision` in the root `pom.xml` (`3.0.0-SNAPSHOT`) — `3.0.0` is the right value to use here, consistent with what the sibling docs themselves currently show as their example.

No new issues found in this round — the diff is scoped exactly to the three fixes requested, both language versions stay in lockstep, and nothing else in the page changed.

# Merge Recommendation

1. **Blockers — must be fixed:** none remaining. All three issues from round 1 are closed.
2. **Recommended fixes — non-blocking:** none outstanding.

**Overall assessment:** clean, complete fix — every issue closed exactly as suggested, both language versions kept faithful to each other, and the caution-admonition addition follows this repo's own documented convention. This is a genuinely useful page now that it's both reachable and runnable end-to-end. Nice turnaround.

CI note: the `Build` check for this head was still in progress at the time of this review — this is a docs-only change, so I wouldn't expect it to fail, but please let it finish before merging.

SEZ9 (changes requested):
Thanks for working on this. I reviewed the latest head `0dc9e62ac025`.

# What this PR fixes
- User pain: Users deploying SeaTunnel via Docker Compose don't know which `SEATUNNEL_HOME` path DolphinScheduler needs (host vs. container view), causing "script not found" failures at job submission (issue #11279).
- Fix approach: Adds a new EN/ZH doc page (`docs/en|zh/tools/dolphinscheduler-integration.md`, registered in `docs/sidebars.js`) explaining host-path vs. container-path semantics, both deployment topologies, and a minimal Compose example mounting `./seatunnel` to `/opt/seatunnel` with `SEATUNNEL_HOME` set accordingly.
- One-line summary: The conceptual explanation is solid and the bilingual versions are consistent, but the "minimal working example" is not currently reproducible — it downloads non-existent SeaTunnel version `3.0.0` (the wget/install-plugin.sh commands will 404), the `:ro` mount conflicts with the client writing to `${SEATUNNEL_HOME}/logs`, the shown `seatunnel.sh --config` invocation needs a running Zeta cluster that the compose file never starts, and setting `SEATUNNEL_HOME` only as a container env var skips DolphinScheduler 3.x's actual resolution path via `dolphinscheduler_env.sh` / UI environment config; the example also hardcodes `3.0.0` in `install-plugin.sh` instead of reusing `${version}`, uses the obsolete `version: '3.8'` compose key, exposes the standalone-server (evaluation-only, embedded storage) with default admin credentials on all interfaces, and fetches the tarball without checksum/signature verification.

# Findings
**Issue 1: Example uses non-existent SeaTunnel version 3.0.0 — the download and plugin-install commands will fail for anyone following the doc**
- Location: `docs/en/tools/dolphinscheduler-integration.md:70`
- Why it matters: - Root cause: the preparation steps hardcode `version="3.0.0"` and download `apache-seatunnel-3.0.0-bin.tar.gz` from archive.apache.org, but no SeaTunnel 3.0.0 release exists there (the current release line is 2.3.x). The `wget` will return 404 and `tar` will then fail on a missing archive.
- Impact: every user copy-pasting this "minimal working example" is blocked at step one, and `install-plugin.sh 3.0.0` (line 80) would also resolve to a non-existent Maven artifact version, so even a partially working setup would download nothing.
- Fix: use a real released version (e.g. `export version="2.3.12"`) consistent with the versions used in `docs/en/start-v2/docker/docker.md`, and reuse `${version}` in the `install-plugin.sh` invocation instead of a second hardcoded literal. The same fix must be applied to the Chinese copy at `docs/zh/tools/dolphinscheduler-integration.md:70` and `:80`.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:72  wget "https://archive.apache.org/dist/seatunnel/${version}/apache-seatunnel-${version}-bin.tar.gz"`; `docs/en/tools/dolphinscheduler-integration.md:80  sh seatunnel/bin/install-plugin.sh 3.0.0`; `docs/zh/tools/dolphinscheduler-integration.md:70  export version="3.0.0"`
- Suggested fix: There is no SeaTunnel `3.0.0` release on `archive.apache.org`, so this `wget` (and the later `install-plugin.sh 3.0.0`) will fail with 404 for anyone following the doc. Please switch to a real released version (e.g. `export version="2.3.12"`, matching what our Docker docs use) and reference `${version}` in the `install-plugin.sh` call as well so there is a single source of truth. The same change is needed in the zh copy.
- Severity: High

**Issue 2: Read-only mount of SEATUNNEL_HOME conflicts with the SeaTunnel client writing logs under $SEATUNNEL_HOME/logs**
- Location: `docs/en/tools/dolphinscheduler-integration.md:57`
- Why it matters: - Root cause: the compose example mounts the installation read-only (`- ./seatunnel:/opt/seatunnel:ro`) and section 6 explicitly claims "A read-only mount (`:ro`) is sufficient, since DolphinScheduler only needs to execute scripts, not write to the installation directory." That claim is inaccurate: `bin/seatunnel.sh` sets `-Dseatunnel.logs.path=${SEATUNNEL_HOME}/logs` and the client's log4j2 config writes log files there, so on a `:ro` mount the launcher fails to create the log file/directory.
- Impact: users following this exact example will hit log4j2 initialization errors or job launch failures inside the DolphinScheduler container, which is exactly the class of confusion this doc is trying to eliminate.
- Fix: either drop `:ro`, or keep it and add an extra writable mount for the logs directory (e.g. `- ./seatunnel-logs:/opt/seatunnel/logs`), and correct the sentence in section 6 (line 87) to state that SeaTunnel writes client logs under `$SEATUNNEL_HOME/logs`. Apply the same correction to `docs/zh/tools/dolphinscheduler-integration.md:57` and `:87`.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:87  - **Required mount**: the host's SeaTunnel installation directory → the container path referenced by `SEATUNNEL_HOME`. A read-only mount (`:ro`) is sufficient, since DolphinScheduler only needs to execute scripts, not write to the installation directory.`; `docs/zh/tools/dolphinscheduler-integration.md:57        - ./seatunnel:/opt/seatunnel:ro`; `docs/en/tools/dolphinscheduler-integration.md:87  - **Required mount**: ... A read-only mount (`:ro`) is sufficient, since DolphinScheduler only needs to execute scripts, not write to the installation directory.`
- Suggested fix: Mounting `SEATUNNEL_HOME` as `:ro` will break the client launch: `bin/seatunnel.sh` points log4j2 at `${SEATUNNEL_HOME}/logs`, so the process needs write access there. Please either drop `:ro`, or add a separate writable mount such as `- ./seatunnel-logs:/opt/seatunnel/logs`, and adjust the section 6 statement that ":ro is sufficient" accordingly.
- Severity: Medium

**Issue 3: Setting SEATUNNEL_HOME only as a container environment variable is not how DolphinScheduler 3.x resolves task env — dolphinscheduler_env.sh / UI environment config should be documented**
- Location: `docs/en/tools/dolphinscheduler-integration.md:53`
- Why it matters: - Root cause: the example relies solely on `- SEATUNNEL_HOME=/opt/seatunnel` in the compose `environment` block. In DolphinScheduler 3.2.x, task shells are launched with the environment configured via `bin/env/dolphinscheduler_env.sh` (which explicitly exports `SEATUNNEL_HOME` with a default of `/opt/soft/seatunnel`) or via a per-task "Environment" configured in the UI; the container-level env var can be overridden by that script's own export, in which case `SEATUNNEL_HOME` silently resolves to the wrong path.
- Impact: users may follow this doc exactly, still get "command not found" / wrong-path failures at task execution time, and have no pointer to the actual DS-side configuration knob — the precise failure mode this document (and issue #11279) is meant to resolve.
- Fix: add a short subsection noting that if `SEATUNNEL_HOME` is not taking effect, it must be set in `dolphinscheduler_env.sh` (mount an override of that file, or use DS's environment management in the UI), and mirror it in the Chinese doc.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:53        - SEATUNNEL_HOME=/opt/seatunnel`; `docs/zh/tools/dolphinscheduler-integration.md:53        - SEATUNNEL_HOME=/opt/seatunnel`
- Suggested fix: DolphinScheduler 3.2.x resolves task environment primarily through `bin/env/dolphinscheduler_env.sh` (which itself exports a default `SEATUNNEL_HOME=/opt/soft/seatunnel`) or the UI's environment management — the script's export can shadow this container-level env var. Please add a note that users may need to override `dolphinscheduler_env.sh` (e.g. mount a customized copy) or configure the environment in the DS UI if the container env var alone does not take effect.
- Severity: Medium

**Issue 4: Compose example publishes the DolphinScheduler UI/API on all interfaces with the image's default admin credentials and no hardening guidance**
- Location: `docs/en/tools/dolphinscheduler-integration.md:51`
- Why it matters: - Root cause: `ports: - "12345:12345"` binds the DolphinScheduler standalone server to `0.0.0.0` on the host. The `apache/dolphinscheduler-standalone-server` image ships with well-known default credentials (`admin` / `dolphinscheduler123`), and DolphinScheduler task execution is effectively remote shell execution — anyone who reaches the UI can define a SHELL/SeaTunnel task and run arbitrary commands inside the container, which has the host's SeaTunnel install and log directories mounted.
- Impact: users who copy this "minimal working example" onto an internet-facing or shared-network host expose an unauthenticated-by-default RCE surface; DolphinScheduler has had multiple published CVEs in this area, so an unpinned, exposed instance is a realistic blast radius.
- Fix: bind to loopback in the example (`"127.0.0.1:12345:12345"`) and add a caution admonition telling users to change the default admin password before exposing the port beyond localhost. Apply the same change to the Chinese doc.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:51        - "12345:12345"`; `docs/zh/tools/dolphinscheduler-integration.md:51        - "12345:12345"`
- Suggested fix: This publishes the DolphinScheduler UI on all host interfaces while the image still has its default `admin`/`dolphinscheduler123` credentials — and a DolphinScheduler task is effectively arbitrary command execution in the container. For a copy-paste example, please bind to loopback (`"127.0.0.1:12345:12345"`) and add a `:::caution` note to change the default admin password before exposing the port. Same for the zh version.
- Severity: Medium

**Issue 5: install-plugin.sh invocation hardcodes `3.0.0` instead of reusing `${version}`, guaranteeing a plugin/engine version mismatch if a user fixes the download version**
- Location: `docs/en/tools/dolphinscheduler-integration.md:80`
- Why it matters: - Root cause: the download block defines `export version="3.0.0"`, but the plugin-installation command hardcodes the literal `3.0.0` again in a separate code block (`sh seatunnel/bin/install-plugin.sh 3.0.0`).
- Impact: when the broken `3.0.0` download version is corrected (or a user substitutes their own version), the second block silently keeps installing connectors for a different version. Since `install-plugin.sh` downloads connector jars by exact Maven version, this yields a `connectors/` directory whose jars do not match the engine in `lib/`, causing class-loading or missing-connector failures at job submission time — the exact failure mode the caution box warns about. The two code blocks are also split by the caution admonition, so they may run in different shell sessions where `version` is unset.
- Fix: use `sh seatunnel/bin/install-plugin.sh "${version}"` and keep it in the same bash block as the `export` (or repeat the export), in both the English and Chinese docs.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:70  export version="3.0.0"`; `docs/zh/tools/dolphinscheduler-integration.md:80  sh seatunnel/bin/install-plugin.sh 3.0.0`
- Suggested fix: Please use `"${version}"` here instead of hardcoding `3.0.0`. If a reader corrects the download version above, this line will install connector plugins for a different version than the engine in `lib/`, producing missing-connector/classpath failures at runtime. Also consider merging this into the same bash block as the `export version=...` so the variable is guaranteed to be in scope.
- Severity: Medium

**Issue 6: Example is built on the standalone-server image without warning that it is evaluation-only (embedded storage, single node, data lost on restart)**
- Location: `docs/en/tools/dolphinscheduler-integration.md:47`
- Why it matters: - Root cause: the compose example uses `apache/dolphinscheduler-standalone-server:3.2.1`, which the DolphinScheduler project explicitly documents as a quick-evaluation image backed by an embedded H2 database — all workflow definitions and run history are lost when the container is recreated, and it is not suitable for real scheduling workloads.
- Impact: this page reads as general "how to integrate DolphinScheduler with SeaTunnel" guidance, so users are likely to carry the example into semi-production setups and lose all their SeaTunnel job/workflow definitions on the first `docker compose down`/image upgrade; only the logs directory is persisted via a volume, not the metadata.
- Fix: add an explicit note that the standalone image is for local testing only, and point users to the official DolphinScheduler cluster/Compose deployment (external PostgreSQL/MySQL + ZooKeeper) for anything durable, keeping the `SEATUNNEL_HOME` mounting guidance identical.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:58        - ./dolphinscheduler/logs:/opt/dolphinscheduler/logs`
- Suggested fix: `dolphinscheduler-standalone-server` uses an embedded H2 database — every workflow/job definition is lost when the container is recreated (only logs are persisted here). Please add a note that this image is for local evaluation only and link to DolphinScheduler's cluster deployment (external DB + ZooKeeper) for durable setups.
- Severity: Medium

**Issue 7: Example command `seatunnel.sh --config <job_config_path>` without `-m local` requires a running Zeta cluster, but the compose file starts no SeaTunnel engine**
- Location: `docs/en/tools/dolphinscheduler-integration.md:10`
- Why it matters: - Root cause: the doc presents `${SEATUNNEL_HOME}/bin/seatunnel.sh --config <job_config_path>` as the command DolphinScheduler executes, but for the Zeta engine the DolphinScheduler SeaTunnel task (and any manual invocation in a standalone container without a running Zeta cluster) needs the local deploy mode flag (`-m local` / `-e local` depending on version), and for Flink/Spark engines entirely different launch scripts (`start-seatunnel-flink-*.sh`, `start-seatunnel-spark-*.sh`) are used.
- Impact: a user who tests the mounted installation manually with the quoted command inside the standalone container will get a cluster-connection failure (no Zeta cluster is running), and may conclude the mount/`SEATUNNEL_HOME` setup is wrong when it is actually the missing mode flag.
- Fix: show `${SEATUNNEL_HOME}/bin/seatunnel.sh --config <job_config_path> -m local` for the Zeta case and add a one-line note that Flink/Spark engine tasks use the corresponding `start-seatunnel-*.sh` scripts; mirror in the Chinese doc.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:47      image: apache/dolphinscheduler-standalone-server:3.2.1`; `docs/zh/tools/dolphinscheduler-integration.md:10  ${SEATUNNEL_HOME}/bin/seatunnel.sh --config <job_config_path>`
- Suggested fix: `seatunnel.sh --config ...` without `-m local` submits to a Zeta cluster, but the compose example below starts no SeaTunnel engine service, so the job submission will fail to connect. Since this doc targets a minimal standalone setup, please document the command as `${SEATUNNEL_HOME}/bin/seatunnel.sh --config <job_config_path> -m local`, or add a note that cluster mode requires a reachable Zeta cluster on the same network with `hazelcast-client.yaml` configured.
- Severity: Medium

**Issue 8: Compose example recommends job-config mounts in section 6 but the "minimal working example" doesn't include one, and `version: '3.8'` is obsolete in Compose v2**
- Location: `docs/en/tools/dolphinscheduler-integration.md:58`
- Why it matters: - Root cause: section 6 recommends mounting "any directory holding job configuration files", but the compose file in section 5 only mounts the SeaTunnel install and the DS logs dir, leaving no obvious path for the `<job_config_path>` referenced in section 1; readers following only the minimal example will have no way to reference config files from the host.
- Impact: minor friction and confusion — jobs authored as files on the host aren't reachable inside the container. Separately, the top-level `version: '3.8'` key is deprecated in Compose v2 and emits a warning.
- Fix: add a commented `- ./jobs:/opt/seatunnel-jobs` volume line to the example so it is self-consistent with section 6, and drop the obsolete `version:` key; apply to both language versions.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:58        - ./dolphinscheduler/logs:/opt/dolphinscheduler/logs`; `docs/en/tools/dolphinscheduler-integration.md:88  - **Recommended mounts**: DolphinScheduler's log directory and any directory holding job configuration files, so logs and configs remain accessible outside the container for debugging and version control.`
- Suggested fix: Section 6 recommends mounting a directory for job config files, but this "minimal working example" doesn't include one, so `<job_config_path>` from section 1 has no host-visible location. Consider adding e.g. `- ./jobs:/opt/seatunnel-jobs` here so the example is self-consistent. Also, the top-level `version: '3.8'` key is obsolete in Compose v2 and can be removed.
- Severity: Low

**Issue 9: Missing blank line between the closing code fence and the :::caution admonition may break Docusaurus rendering**
- Location: `docs/en/tools/dolphinscheduler-integration.md:75`
- Why it matters: - Root cause: the `:::caution Warning` block starts on the line immediately after the closing ``` of the bash code block, with no blank line separating the two Markdown blocks.
- Impact: Docusaurus/MDX admonition parsing is sensitive to block boundaries; without a separating blank line the `:::caution` directive can render as literal text instead of a styled warning box, degrading a safety-critical callout (installing plugins before the read-only mount).
- Fix: insert an empty line between the closing code fence and `:::caution Warning` in both the English (line 74/75) and Chinese (line 74/75) files, and preview the rendered page to confirm the admonition displays correctly.
- Evidence: `docs/en/tools/dolphinscheduler-integration.md:75  :::caution Warning`; `docs/zh/tools/dolphinscheduler-integration.md:75  :::caution 警告`
- Suggested fix: Please add a blank line between the closing code fence and `:::caution Warning` (here and in the zh page). MDX admonitions need a block boundary — without the empty line this important plugin-installation warning can render as plain text instead of a callout box.
- Severity: Low

**Issue 10: Compose file uses the obsolete top-level `version: '3.8'` key**
- Location: `docs/en/tools/dolphinscheduler-integration.md:43`
- Why it matters: - Root cause: the example compose file starts with `version: '3.8'`, which is deprecated in the Compose Specification; Docker Compose v2 ignores it and prints an 'the attribute `version` is obsolete' warning.
- Impact: purely cosmetic, but new users copying the snippet will see a warning on every `docker compose up`, which undermines confidence in an otherwise 'ready-to-use' example.
- Fix: drop the `version:` line from both the English and Chinese examples; the rest of the file is already valid under the Compose Specification.
- Evidence: `docs/zh/tools/dolphinscheduler-integration.md:43  version: '3.8'`
- Suggested fix: Nit: the top-level `version:` key is obsolete in the Compose Specification and Docker Compose v2 emits a warning for it — I'd drop this line here and in the zh example to keep the snippet warning-free.
- Severity: Low

# Review conclusion

**1. Blocking items**
- Issue 1: Example uses non-existent SeaTunnel version 3.0.0 — the download and plugin-install commands will fail for anyone following the doc

**2. Remaining follow-ups**
- Issue 2: Read-only mount of SEATUNNEL_HOME conflicts with the SeaTunnel client writing logs under $SEATUNNEL_HOME/logs
- Issue 3: Setting SEATUNNEL_HOME only as a container environment variable is not how DolphinScheduler 3.x resolves task env — dolphinscheduler_env.sh / UI environment config should be documented
- Issue 4: Compose example publishes the DolphinScheduler UI/API on all interfaces with the image's default admin credentials and no hardening guidance
- Issue 5: install-plugin.sh invocation hardcodes `3.0.0` instead of reusing `${version}`, guaranteeing a plugin/engine version mismatch if a user fixes the download version
- Issue 6: Example is built on the standalone-server image without warning that it is evaluation-only (embedded storage, single node, data lost on restart)
- Issue 7: Example command `seatunnel.sh --config <job_config_path>` without `-m local` requires a running Zeta cluster, but the compose file starts no SeaTunnel engine
- Issue 8: Compose example recommends job-config mounts in section 6 but the "minimal working example" doesn't include one, and `version: '3.8'` is obsolete in Compose v2
- Issue 9: Missing blank line between the closing code fence and the :::caution admonition may break Docusaurus rendering

DanielLeens:
Thanks for the deep re-review — going back through the whole page again rather than trusting my prior sign-off is exactly the right instinct here, and a couple of these are real gaps I missed. I checked each point against the current head and the actual SeaTunnel source before replying, rather than taking either of our prior conclusions at face value.

**Issue 1 (version `3.0.0` doesn't exist) — you're right, and I need to correct my prior round.** I checked GitHub Releases directly: the latest published SeaTunnel release is `2.3.13`; there is no `3.0.0` release, so `wget` against `archive.apache.org/dist/seatunnel/3.0.0/...` will indeed 404 today. My earlier "verified" note was based on `pom.xml`'s `<revision>3.0.0-SNAPSHOT</revision>` and on `deployment.md`/`download-seatunnel.md` using the identical `export version="3.0.0"` pattern — that part is still true, but I was wrong to treat "consistent with sibling docs" as equivalent to "actually reproducible." Since this is a brand-new page whose whole premise is a "minimal working example," I'd rather it not inherit a pre-existing placeholder-version problem from those two docs. Recommend switching to a real released version (`2.3.13`) here. I wouldn't ask this PR to also fix `deployment.md`/`download-seatunnel.md` — that's a separate, repo-wide docs issue worth its own follow-up.

**Issue 5 (`install-plugin.sh` hardcodes `3.0.0` instead of `${version}`) — confirmed, and this is a genuine miss on my side from round 2.** Pulling the current file directly, the `sh seatunnel/bin/install-plugin.sh 3.0.0` line is indeed a second, separate hardcoded literal, split from the `export version=...` block by the caution admonition. When Issue 1 gets fixed to a real version, this line needs to move to `${version}` in the same commit — and ideally the same bash block as the export, per your suggestion, so it can't silently drift out of sync.

**Issue 2 (`:ro` mount vs. client logs) — I looked into this and don't think it holds up against the actual default config.** I checked `config/log4j2_client.properties`, which is what ships as the client-side log4j2 config used by `bin/seatunnel.sh`: the file appender block is commented out by default —

```
############################ log output to file    #############################
#rootLogger.appenderRef.file.ref = fileAppender
############################ log output to file    #############################
...
#appender.file.name = fileAppender
#appender.file.type = RollingFile
#appender.file.fileName = ${file_path}/${file_name}.log
```

`rootLogger` only wires up `consoleStdoutAppender`/`consoleStderrAppender` out of the box — with the shipped default config, `seatunnel.sh` never actually writes to `${SEATUNNEL_HOME}/logs`, everything goes to stdout/stderr, which Compose already captures via `docker logs`. So the `:ro` mount and the doc's "read-only is sufficient" claim both hold for the out-of-the-box config. Where I think you're onto something real: if a user *does* uncomment the file appender (a fairly common thing to do for persistent logs), they'd hit exactly the failure you describe. I'd suggest a narrower fix than dropping `:ro` by default — a one-line note in section 6 saying that enabling file-based client logging via `log4j2_client.properties` requires a separate writable mount for `${SEATUNNEL_HOME}/logs`, rather than changing the default example's mount mode.

**Issues 3, 4, 6, 7, 8, 9, 10 — agreed on all of these, no pushback.** Each checks out against how DolphinScheduler/Compose/this doc's own conventions actually work: the `dolphinscheduler_env.sh` resolution note (3), binding the UI to loopback plus a password-change caution (4), flagging the standalone image as eval-only with embedded H2 (6), adding `-m local` (or documenting the cluster-mode requirement) for the example command (7), the missing jobs-mount plus obsolete `version:` key (8/10), and the blank line before `:::caution` for MDX safety (9) are all correct, low-risk doc improvements that make the example genuinely copy-paste-safe rather than just plausible-looking.

Given the version and install-plugin.sh issues (1 and 5) are things a reader hits at step one, I'm moving this back to "changes requested" until those land. I'd bundle 3/4/6/7/8/9/10, plus the narrower log-mount note from Issue 2, into the same follow-up commit rather than trickle them in separately. Thanks again for catching what I missed on the version and install-plugin.sh lines.

davidzollo:

```

```

davidzollo (approved):
+1
LGTM

PR #11697, opened 2026-08-07
[Fix][Transform-V2] Fix silent value corruption in Zeta SQL numeric functions

### LINKED ISSUE #11696: [Bug] [Transform-V2] Zeta SQL CEIL/FLOOR/ROUND/TRUNC/MOD silently corrupt BIGINT and DECIMAL values
### Search before asking

### What happened

NumericFunction in the Zeta SQL transform routes values through double and narrows CEIL/FLOOR results to int. No exception is raised — the job succeeds and writes wrong numbers to 
the sink.
 
| Expression          | Input                          | Actual                        | Expected                |
|---------------------|--------------------------------|-------------------------------|-------------------------|
| CEIL(long_v)        | 9007199254740993               | 1                             | 9007199254740993        |
| FLOOR(double_v)     | 1.0E18                         | 2147483647                    | 1.0E18                  |
| ROUND(decimal_v, 2) | 12345678901234567890.987654321 | 12345678901234567000.00       | 12345678901234567890.99 |
| MOD(long_v, 2)      | 9007199254740993               | 0                             | 1                       |
| TRUNC(long_v)       | any                            | declared DOUBLE, returns Long | type matches value      |
 
This contradicts the shipped documentation. docs/en/transforms/sql-functions.md already specifies CEIL | CEILING (numeric) -> NUMERIC (same type, scale 0), FLOOR(numeric) -> 
NUMERIC (same type, scale 0), ROUND(numeric[, digitsInt]) -> NUMERIC (same type), TRUNC | TRUNCATE(numeric[, digitsInt]) -> NUMERIC (same type) and MOD(dividendNumeric, 
divisorNumeric) -> type(divisorNumeric).

Root causes: ceil()/floor() return int and ZetaSQLType declares INT_TYPE/DOUBLE_TYPE for them; round()'s BIGDECIMAL branch goes through BigDecimal.valueOf(v1.doubleValue()); mod() 
converts both operands with doubleValue() (which also makes its divide-by-zero guard miss divisors that underflow to 0.0).

### SeaTunnel Version

dev (2.3.13-SNAPSHOT)

### SeaTunnel Config

```conf
hocon
  transform {
    Sql {
      source_table_name = "fake"
      result_table_name = "out"
      query = "select CEIL(long_v) as c, FLOOR(double_v) as f, ROUND(decimal_v, 2) as r, MOD(long_v, 2) as m from fake"
    }
  }
```

### Running Command

```shell
./bin/seatunnel.sh --config ./config/repro.conf -e local
```

### Error Exception

```log
None — this is silent data corruption; the job reports success.
```

Zeta (engine-independent — the bug is in the transform)

### Java or Scala Version

Java 11 / Java 8

### Screenshots

_No response_

### Code of Conduct

SEPURI-SAI-KRISHNA:

Closes #11696.

`CEIL`, `CEILING`, `FLOOR`, `ROUND`, `TRUNC`, `TRUNCATE` and `MOD` in the Zeta SQL transform lost precision or overflowed silently, writing wrong values to the sink without raising any error. The job reports success, so the corruption is invisible until someone audits the data downstream.

| Expression | Input | Before | After |
|---|---|---|---|
| `CEIL(long_v)` | `9007199254740993` | `1` | `9007199254740993` |
| `FLOOR(double_v)` | `1.0E18` | `2147483647` | `1.0E18` |
| `ROUND(decimal_v, 2)` | `12345678901234567890.987654321` | `12345678901234567000.00` | `12345678901234567890.99` |
| `MOD(long_v, 2)` | `9007199254740993` | `0` | `1` |
| `TRUNC(long_v)` | any | declared `DOUBLE`, returned a `Long` | declared type matches the value |

This brings the implementation in line with the behaviour that `docs/en/transforms/sql-functions.md` already documents, which specifies that these functions return the data type of their argument:

- `CEIL | CEILING (numeric) -> NUMERIC (same type, scale 0)`
- `FLOOR(numeric) -> NUMERIC (same type, scale 0)`
- `ROUND(numeric[, digitsInt]) -> NUMERIC (same type)`
- `TRUNC | TRUNCATE(numeric[, digitsInt]) -> NUMERIC (same type)`
- `MOD(dividendNumeric, divisorNumeric) -> type(divisorNumeric)`

Changes:

- `NumericFunction.ceil` / `floor` return `Number` instead of narrowing the result to `int`.
- `NumericFunction.round` operates on the `BigDecimal` directly in its `BIGDECIMAL` branch instead of round-tripping through `BigDecimal.valueOf(v1.doubleValue())`, which dropped every digit a `double` cannot hold.
- `NumericFunction.mod` uses exact `BigDecimal` arithmetic via a new `toBigDecimal` helper. This also corrects its divide-by-zero guard, which previously tested `rightValue.doubleValue() == 0`: a nonzero `DECIMAL` divisor smaller than `Double.MIN_VALUE` underflows to `0.0`, so the old guard rejected a valid divisor with "Mod by zero". The guard now tests `rightBD.signum() == 0` on the exact value.
- `ZetaSQLType` declares the argument's type for `CEIL` / `CEILING` / `FLOOR` / `TRUNC` / `TRUNCATE`, joining the existing `ABS` / `ROUND` group. This also resolves a declared-vs-actual type mismatch where `TRUNC` was declared `DOUBLE` but returned a `Long` at runtime.

Two related defects are deliberately left out of scope, because each needs its own behavioural decision rather than being folded into this fix. I will raise them separately:

1. `ABS(-2147483648)` returns `-2147483648`, since `Math.abs` cannot represent `-Integer.MIN_VALUE`. H2 raises an overflow error here, so converting a silent wrong answer into a hard job failure is a larger call than this PR should make.
2. `ROUND(tinyint_col, -1)` does not round at all, because `Byte` is not handled in the type switch in `round()`. Adding it naively introduces a byte-overflow edge (`127` -> `130` -> `-126`), so it needs a widening decision first.

Yes, and it is a change relative to released versions as well as to `dev`.

1. `CEIL` / `CEILING`, `FLOOR` and `TRUNC` / `TRUNCATE` now return the data type of their argument rather than always `INT` (and rather than a declared `DOUBLE` for `TRUNC`). A `BIGINT` argument now yields a `BIGINT`, a `DOUBLE` argument a `DOUBLE`, and a `DECIMAL` argument a `DECIMAL`.
2. `ROUND`, `TRUNC` / `TRUNCATE` and `MOD` now return exact results for `BIGINT` and `DECIMAL` arguments instead of values degraded by a `double` round-trip.

Both are recorded in `docs/en/introduction/concepts/incompatible-changes.md` and its `docs/zh` mirror, with before/after tables and a migration guide. Pipelines that relied on the old narrowing should widen the target column, or wrap the call in `CAST(... AS INT)` to keep the previous output shape.

No config option names, defaults or SPI contracts were changed.

Added unit tests in `seatunnel-transforms-v2/src/test/java/org/apache/seatunnel/transform/sql/zeta/NumericFunctionTest.java`, which drive the real SQL engine and assert both the returned value and the declared output type:

- `testCeilAndFloorKeepBigIntValue` — a `BIGINT` beyond `2^53` (`9007199254740993`) survives `CEIL` / `FLOOR` / `TRUNC` unchanged, and the declared type stays `LONG_TYPE`.
- `testCeilAndFloorKeepDoubleType` — a `DOUBLE` of `1.0E18`, far outside the `int` range, keeps its value and `DOUBLE_TYPE`.
- `testRoundAndTruncKeepDecimalPrecision` — a `DECIMAL(38, 9)` of `12345678901234567890.987654321` stays exact through `ROUND`, `TRUNC` and `CEIL`, with the declared type preserved.
- `testModKeepsBigIntPrecision` — covers the `BIGINT` and `DECIMAL` cases, including the `type(divisorNumeric)` return-type rule.

Two pre-existing assertions had codified the buggy narrowing (`assertEquals(2, ceil(1.2d))` in `SQLNumericFunctionsTest` and in the `...sql.zeta.functions.NumericFunctionTest` class) and were updated to the documented contract, with explicit `int` and `long` cases added alongside.

Verified locally on JDK 11:

```bash
./mvnw spotless:apply
./mvnw -pl seatunnel-transforms-v2 -DskipTests verify
./mvnw -pl seatunnel-transforms-v2 test
```

All pass: `Tests run: 1079, Failures: 0, Errors: 0, Skipped: 0`.

Blast radius check: `ZetaSQLFunction` is the only caller of the changed methods and dispatches through `Object`, so widening the `ceil` / `floor` return type from `Integer` to `Number` is source- and binary-compatible for it. The remaining `CEIL(` / `FLOOR(` / `TRUNC(` matches in the repository are database-side SQL (`XuguCatalog`, a test resource config) and are unaffected.

### Check list


DanielLeens (commented):
Welcome to Apache SeaTunnel, and thank you for this contribution — it's a genuinely impressive first PR. You found a real, quietly dangerous data-corruption bug, root-caused it correctly, fixed it with minimal and well-reasoned changes, and backed it with an honest, well-written breaking-change writeup. I read through the full diff plus the current source at head `b6e7cff2` to trace the runtime path end to end. Details below.

The Zeta SQL transform's `CEIL`, `FLOOR`, `ROUND`, `TRUNC`/`TRUNCATE` and `MOD` functions were silently corrupting numeric values instead of raising an error, in two distinct ways:

1. **Narrowing to `int`**: `NumericFunction.ceil()` / `floor()` always called `.intValue()` on the result, so a `BIGINT` like `9007199254740993` became `1`, and a `DOUBLE` like `1.0E18` became `2147483647` (max `int`). No exception, no warning — just a wrong number silently written to the sink.
2. **Round-tripping through `double`**: `ROUND`'s `BIGDECIMAL` branch and `MOD` both converted their `BigDecimal` operands to `double` before computing, which discards any digits beyond `double`'s ~15-17 significant-digit precision. A `DECIMAL(38,9)` value with 30 significant digits would lose everything past digit 17.

Both classes of bug are invisible at job-submission and job-completion time — the pipeline reports success, and the wrong value only surfaces if someone later audits the sink data against the source. That's exactly the kind of defect that's expensive to find in a production data-integration system.

## 1. Code Change Review

### 1.1 Core Logic Analysis

**`ceil` / `floor` — stop narrowing to `int`**

`seatunnel-transforms-v2/.../functions/NumericFunction.java`

Before:
```java
public static Integer ceil(List<Object> args) {
    ...
    return round(v1, v2, RoundingMode.CEILING).intValue();
}
```
After:
```java
public static Number ceil(List<Object> args) {
    ...
    return round(v1, v2, RoundingMode.CEILING);
}
```
Same change mirrored for `floor()`. The shared `round(Number, Number, RoundingMode)` helper already dispatches by the boxed type's simple name (`INTEGER`/`SHORT`/`LONG`/`BIGDECIMAL`/`DOUBLE`/`FLOAT`) and, for the integral branch, only touches the value when `scale < 0`. So for the common `CEIL(col)` / `FLOOR(col)` case (no second argument, scale 0) on a `LONG`, the value now passes through completely unmodified instead of being force-cast into `int`.

**`ROUND`'s `BIGDECIMAL` branch — stop round-tripping through `double`**

Before:
```java
case "BIGDECIMAL": {
    BigDecimal bd = BigDecimal.valueOf(v1.doubleValue());
    v1 = bd.setScale(scale, roundingMode);
    break;
}
```
After:
```java
case "BIGDECIMAL": {
    // Must not round-trip through double: a DECIMAL carries more digits than a
    // double can hold, so doubleValue() would silently drop the low-order ones.
    v1 = ((BigDecimal) v1).setScale(scale, roundingMode);
    break;
}
```
This branch is shared by `ROUND`, `CEIL`, `FLOOR`, and `TRUNC` for `BigDecimal` inputs, so all four benefit. Operating on the `BigDecimal` directly instead of re-deriving it from a lossy `double` is the correct fix.

**`MOD` — exact `BigDecimal` arithmetic + a real zero-divisor guard**

Before:
```java
if (rightValue.doubleValue() == 0) {
    throw new TransformException(..., "Mod by zero");
}
BigDecimal leftBD = BigDecimal.valueOf(leftValue.doubleValue());
BigDecimal rightBD = BigDecimal.valueOf(rightValue.doubleValue());
```
After:
```java
BigDecimal leftBD = toBigDecimal(leftValue);
BigDecimal rightBD = toBigDecimal(rightValue);
if (rightBD.signum() == 0) {
    throw new TransformException(..., "Mod by zero");
}
```
with the new helper:
```java
private static BigDecimal toBigDecimal(Number value) {
    if (value instanceof BigDecimal) {
        return (BigDecimal) value;
    }
    if (value instanceof Byte || value instanceof Short
            || value instanceof Integer || value instanceof Long) {
        return BigDecimal.valueOf(value.longValue());
    }
    return BigDecimal.valueOf(value.doubleValue());
}
```
This is a well-scoped helper: `BigDecimal` is used as-is, integral types are widened via `longValue()` (exact for anything up to 64-bit), and only genuinely floating types fall back to `doubleValue()` (where there's no more precision to lose in the first place). The zero-check now runs on the exact `BigDecimal` rather than a possibly-underflowed `double`.

**`ZetaSQLType.getFunctionType` — declared type now matches the actual runtime type**

`seatunnel-transforms-v2/.../zeta/ZetaSQLType.java:463-472`:
```java
case ZetaSQLFunction.ABS:
case ZetaSQLFunction.DATEADD:
case ZetaSQLFunction.TIMESTAMPADD:
case ZetaSQLFunction.ROUND:
    // CEIL/FLOOR/TRUNC are documented to return the type of their argument. Declaring
    // INT/DOUBLE here would truncate BIGINT and DECIMAL results.
case ZetaSQLFunction.CEIL:
case ZetaSQLFunction.CEILING:
case ZetaSQLFunction.FLOOR:
case ZetaSQLFunction.TRUNC:
case ZetaSQLFunction.TRUNCATE:
case ZetaSQLFunction.NULLIF:
    return getExpressionType(function.getParameters().getExpressions().get(0));
```
`CEIL`/`CEILING`/`FLOOR` move out of the `BasicType.INT_TYPE` group, and `TRUNC`/`TRUNCATE` move out of the `BasicType.DOUBLE_TYPE` group, into the "same type as argument" group that `ABS`/`ROUND` already used.

This is not just cosmetic — I traced the actual runtime path. `ZetaSQLEngine.project()` (`.../zeta/ZetaSQLEngine.java:322-347`) calls `zetaSQLFunction.computeForValue(expression, inputFields)` and puts the raw `Object` result straight into the output row (`ZetaSQLEngine.java:339`, `new SeaTunnelRow(outputFields)` at line 305). There is **no coercion step** back to the declared `SeaTunnelDataType` for these functions — I checked `computeForValue`'s `targetType` (`ZetaSQLFunction.java:458`) and it is only consumed by `COALESCE`/`IFNULL` (lines 639/641). So before this PR, `TRUNC(bigint_col)` was a genuine declared-vs-actual type contract violation: `outRowType.getFieldType()` said `DOUBLE_TYPE` (`trunc()` itself was never touched, always returned an unnarrowed `Number`, i.e. a `Long` for a `BIGINT` input) while the row actually carried a `Long`. Any downstream code that trusts the declared schema (a serializer that does `(Double) field`, a strict sink writer, a Kryo/Fury codec keyed by declared type) was one `TRUNC(bigint_col)` away from a `ClassCastException` or worse. This PR closes that gap for real, not just for the happy path.

**Runtime path confirmation**: This is the mainline `SQLTransform` plugin (`org.apache.seatunnel.transform.sql.SQLTransform`), the standard SQL transform used in production ETL pipelines — not a recovery/boundary path. `SQLNumericFunctionsTest.testBasicNumericFunctions` (`seatunnel-transforms-v2/src/test/java/.../sql/SQLNumericFunctionsTest.java`) exercises it end-to-end via `SQLTransform.transformRow(...)`, confirming the fix is reachable from the real user-facing entry point, not just the internal `NumericFunction` unit.

**Blast-radius check** (independently verified, not just taken from the PR description): I grepped `ZetaSQLFunction.java` for every call site of the four changed methods. `ceil`/`floor` have exactly one call site each (`ZetaSQLFunction.java:561-563`, `566-567`), both dispatching through `Object` with no post-hoc cast. `mod` has two call sites — the dispatcher (`ZetaSQLFunction.java:559-560`) and the `DECIMAL`-result binary-modulo path (`ZetaSQLFunction.java:781-785`) — both also untyped `Object` returns. The claim that widening `ceil`/`floor` from `Integer` to `Number` is safe for all callers checks out.

### 1.2 Compatibility

**Partially incompatible — intentionally and explicitly.** This is a documented `[BREAKING]` change, correctly so:
- `CEIL`/`CEILING`/`FLOOR`/`TRUNC`/`TRUNCATE` now return the argument's data type instead of always `INT` (`CEIL`/`FLOOR`) or a declared-but-wrong `DOUBLE` (`TRUNC`).
- `ROUND`/`TRUNC`/`TRUNCATE`/`MOD` now return exact values for `BIGINT`/`DECIMAL` inputs instead of `double`-degraded ones.

Both are recorded in `docs/en/introduction/concepts/incompatible-changes.md` and its `docs/zh` mirror, each with a before/after table and a concrete migration path (widen the sink column, or wrap in `CAST(... AS INT)` to preserve old output shape). I checked the doc placement — it's correctly inserted into the existing "SQL Transform" bullet cluster in the same file, right after the `DATEDIFF`/`WEEK` bullet, matching the file's established format. This is exactly the level of rigor the project's compatibility policy expects.

No config option names, defaults, or SPI contracts changed — this is purely a computed-value/declared-type correction, which is the right shape for this kind of fix.

No meaningful new cost. `toBigDecimal()` allocates one `BigDecimal` per `MOD` call, same as the pre-existing `BigDecimal.valueOf(...)` calls it replaces — no new allocation pattern. No new locking, threading, or I/O. No change to memory footprint beyond the (already-existing) per-row `BigDecimal` boxing that `MOD`/`ROUND` always did for decimal arithmetic.

Exception messages are unchanged and already carry sufficient context ("Mod by zero", "Unsupported arg type %s of function %s"). No new logging was added, so there's nothing to check for secret leakage or log-level misuse. No exceptions are swallowed — the new `toBigDecimal()` path preserves the same `TransformException`/`CommonErrorCodeDeprecated.UNSUPPORTED_OPERATION` contract as before.

One documentation-only nit (not a code defect, see Issue 3 below): the PR description's phrasing "previously compared `doubleValue() == 0` and therefore missed divisors that underflow to `0.0`" describes the failure direction backwards. An extremely small nonzero `BigDecimal` divisor whose `doubleValue()` underflows to `0.0` would have caused the *old* code to incorrectly *throw* "Mod by zero" for a valid nonzero divisor (a false positive), not silently miss a real zero (a false negative). The actual code fix (`rightBD.signum() == 0`) is correct either way — this is purely about the PR description's wording, worth a one-line correction for anyone reading the history later.

## 2. Code Quality

### 2.1 Coding standards & comment completeness

Comments are appropriately placed and genuinely explain *why*, not just *what* — e.g. the `toBigDecimal` Javadoc explains the precision rationale per branch, and the inline comment on the `BIGDECIMAL` round branch explains exactly why `doubleValue()` would be wrong there. The `ceil`/`floor` Javadocs explain the type-widening rationale and cross-reference each other. This meets the project's "comment on important/complex logic" bar well — better than a lot of existing code in this file, honestly.

One minor style nit (Issue 2 below): in `ZetaSQLType.java`, the new explanatory comment sits between `case ZetaSQLFunction.ROUND:` and `case ZetaSQLFunction.CEIL:` inside a fallthrough group that starts several case labels earlier at `ABS`. It's valid Java (a comment between case labels doesn't break fallthrough) and doesn't affect correctness, but a reader skimming case labels could misread it as documenting only `ROUND` rather than the whole `CEIL`/`FLOOR`/`TRUNC` group it's actually about.

Coverage is solid and targeted at the actual bug classes, not just the happy path:
- `testCeilAndFloorKeepBigIntValue` (`zeta/NumericFunctionTest.java:99`) — `BIGINT` beyond `2^53` survives `CEIL`/`FLOOR`/`TRUNC`, and asserts the *declared* output type via `sqlEngine.typeMapping(null)`, not just the value. This is exactly the right test shape given the declared-vs-actual bug this PR fixes.
- `testCeilAndFloorKeepDoubleType` (`:126`) — `DOUBLE` far outside `int` range, same declared+value assertion pattern.
- `testRoundAndTruncKeepDecimalPrecision` (`:150`) — a 30-significant-digit `DECIMAL(38,9)` stays exact through `ROUND`/`TRUNC`/`CEIL`.
- `testModKeepsBigIntPrecision` (`:178`) — covers `BIGINT` and `BigDecimal` divisors, including the `type(divisorNumeric)` return-type rule.

Two pre-existing tests that had codified the buggy narrowing (`SQLNumericFunctionsTest.testBasicNumericFunctions` and `zeta/functions/NumericFunctionTest.testCeilFloorRoundAndTrunc`) were correctly updated to the new expected values/types, and — importantly — new assertions were *added* alongside (e.g. `ceil(2)` for plain `int` input, `floor(2L)` for `Long`) rather than the old assertions simply being deleted. I compared the before/after of both files line-by-line: no assertion was weakened, no case was removed, only expected values were corrected to match the now-correct implementation, plus net-new coverage was added.

**Flaky-test analysis** (mandatory since this PR touches UT code): I checked all new/modified assertions for `Thread.sleep`, `Awaitility`, shared static state, execution-order dependence, and floating-point comparisons without a delta. None of the new tests use timing, threads, or shared mutable state. The `assertEquals(1.0e18d, ...)` and `assertEquals(2d, ...)`-style comparisons look like bare floating-point equality but are safe: in every case the value passes through `Math.ceil`/`Math.floor` on an input that's already integral at `double` precision (no arithmetic that could introduce rounding error), so the assertion is comparing identical bit patterns, not the result of independent floating-point computation.

**Stability rating: Stable.**

### 2.3 Docs

`docs/en/transforms/sql-functions.md` and its `docs/zh` mirror already documented the correct contract (`CEIL | CEILING (numeric) -> NUMERIC (same type, scale 0)`, `TRUNC | TRUNCATE(numeric[, digitsInt]) -> NUMERIC (same type)`, `MOD(dividendNumeric, divisorNumeric) -> type(divisorNumeric)`) — I verified this directly in both files rather than taking the PR description's word for it, and it's accurate. So no function-reference doc changes were needed, and none were made; that's correct scoping, not an omission. The `incompatible-changes.md` additions (both `en` and `zh`) are thorough, accurate against the actual diff, and formatted consistently with the rest of the file.

## 3. Architectural soundness

The fix is minimal and correctly scoped: it touches only the numeric-function implementation and its type-declaration counterpart, without leaking connector-specific or engine-specific logic into either. The author explicitly called out two related-but-out-of-scope defects (`ABS(Integer.MIN_VALUE)` overflow, `ROUND` on `tinyint` with negative scale) that each need their own behavioral decision rather than being folded in here — that's the right call, and exactly the kind of scope discipline this project's PR-scope rule asks for.

One additional out-of-scope observation for the record (see Issue 4): the binary arithmetic operators (`+`, `-`, `*`, `/`) for `DECIMAL`-typed expressions in `ZetaSQLFunction.executeBinaryExpr` (`ZetaSQLFunction.java:763-780`) still round-trip through `BigDecimal.valueOf(leftValue.doubleValue())`/`rightValue.doubleValue()` — the same precision-loss pattern this PR just fixed for the function-call path, just not for `+`/`-`/`*`/`/` on decimals. Not introduced or worsened by this PR, not required for this PR to be mergeable, but worth a follow-up ticket since it's the same bug class.

## 4. Issue Summary

| # | Severity | Location | Summary | Raised by another reviewer |
| 1 | High | CI (GitHub Actions) | `Build` check on the PR shows `conclusion: action_required` with 0 workflow runs recorded in the contributor's fork — the code has not actually been validated by CI yet | No |
| 2 | Low | `seatunnel-transforms-v2/.../zeta/ZetaSQLType.java:464-465` | Explanatory comment sits between `ROUND` and `CEIL` case labels inside a shared fallthrough group, reads as documenting only `ROUND` | No |
| 3 | Low | PR description (MOD zero-check rationale) | Description of the old zero-divisor bug direction is inverted (would have false-positived on tiny nonzero divisors, not missed real zeros); the code fix itself is correct | No |
| 4 | Low | `seatunnel-transforms-v2/.../zeta/ZetaSQLFunction.java:763-780` | Pre-existing, out-of-scope: `DECIMAL` binary arithmetic (`+ - * /`) still round-trips through `double`, same bug class this PR fixes for function calls | No |

### Issue 1 detail

**Location**: PR check `Build` — https://github.com/apache/seatunnel/runs/92914583204

**Problem**: I queried the check-run directly (`gh api repos/apache/seatunnel/commits/b6e7cff2.../check-runs`). Its output is:
> "Unable to detect the workflow run for testing the changes in your PR. 1. If you did not enable GitHub Actions in your forked repository, please enable it... 2. Create and push an empty commit to trigger the workflow. 3. ...sync your branch to the latest dev branch."

I also queried `repos/SEPURI-SAI-KRISHNA/seatunnel/actions/runs` directly and it returns zero workflow runs at all — consistent with GitHub Actions never having been enabled in the fork, which is the most common cause of this exact message for first-time contributors. `dev...b6e7cff2` shows `ahead_by: 1, behind_by: 6` — only 6 commits behind `dev`, so branch staleness is an unlikely primary cause, though a rebase is cheap to try if enabling Actions alone doesn't resolve it.

**Risk**: This PR changes core numeric semantics of a widely-used SQL transform with an explicit breaking-change contract. Merging it with zero CI validation (not "CI is red," but "CI has never run") would mean nothing has actually confirmed the code compiles or the new tests pass outside of the author's local run.

**Best improvement**: The contributor needs to enable GitHub Actions on their fork (repo Settings → Actions → General → allow Actions), then push any commit (even empty) to re-trigger the check. This is a process blocker, not a code blocker — the code changes themselves don't need to change for this.

**Severity**: High (blocking, but purely procedural — not a code defect).

## 5. Merge Recommendation

**Blockers** (must-fix before merge): Issue 1 — get CI to actually execute (enable Actions in the fork and re-push) and confirm it passes. Given the excellent local test coverage described in the PR, I have no reason to expect CI to surface anything the author hasn't already caught, but the project's process requires actual green CI, not a description of a local run, before this lands.

**Recommended fixes** (non-blocking, can be follow-ups or addressed in this PR at the author's discretion): Issues 2-4 — a comment-placement tweak, a one-line correction to the PR description's bug-direction claim, and a flagged-for-later related bug in decimal binary arithmetic.

Code quality, test rigor, and documentation honesty here are well above the bar I'd expect from a first PR — the type-contract bug in `TRUNC` in particular is a sharp catch that isn't obvious from the symptom description alone. Once CI actually runs and comes back green, this should be straightforward to merge.

SEPURI-SAI-KRISHNA:
Thanks for the exceptionally thorough review — you clearly traced the runtime path rather than reading the diff in isolation, and the confirmation that `computeForValue`'s `targetType` is only consumed by `COALESCE`/`IFNULL` is the detail that makes the `TRUNC` type-contract argument airtight. Addressing your four points:

**Issue 1 (CI never ran) — resolved.** GitHub Actions was indeed disabled on my fork. I've enabled it and re-pushed, and the workflows now execute: `Code style`, `License header`, `Build website`, `Dead links`, all four `unit-test` jobs (Java 8 and 11, Ubuntu and Windows) and `transform-v2-it-part-1` have passed, with the remaining connector integration jobs still running. No failures so far. For reference, locally on JDK 11 `./mvnw -pl seatunnel-transforms-v2 test` gives `Tests run: 1079, Failures: 0, Errors: 0, Skipped: 0`.

**Issue 2 (comment placement) — fixed.** You're right that a comment between `case ROUND:` and `case CEIL:` reads as documenting only `ROUND`. I've moved it above the whole fallthrough group and reworded it to describe the group's shared contract:

```java
// These functions all return the type of their first argument. CEIL/FLOOR/TRUNC are
// documented that way as well; declaring INT/DOUBLE for them would truncate BIGINT
// and DECIMAL results.
case ZetaSQLFunction.ABS:
case ZetaSQLFunction.DATEADD:
case ZetaSQLFunction.TIMESTAMPADD:
case ZetaSQLFunction.ROUND:
case ZetaSQLFunction.CEIL:
...
```

**Issue 3 (inverted bug direction in the description) — fixed, and you're correct.** I re-checked: the old `rightValue.doubleValue() == 0` guard had no false-negative path at all. Its only defect is the false positive you describe — a nonzero `DECIMAL` divisor below `Double.MIN_VALUE` underflows to `0.0` and gets rejected as "Mod by zero". I've corrected the PR description accordingly. The code (`rightBD.signum() == 0`) was already right.

**Issue 4 (`DECIMAL` binary arithmetic) — agreed, and worth its own ticket.** Confirmed at `ZetaSQLFunction.executeBinaryExpr` (`ZetaSQLFunction.java:763-780`): `+`, `-`, `*` and `/` on `DECIMAL` operands all go through `BigDecimal.valueOf(leftValue.doubleValue())`, which is precisely the bug class this PR fixes for the function-call path. Notably `Modulo` in that same block already delegates to `NumericFunction.mod`, so it picks up this PR's fix for free — which leaves `+ - * /` as the sole remaining offenders and makes the inconsistency more visible. I'd rather not widen this PR's blast radius, so I'll open a follow-up issue for it. Same reasoning for the two I flagged originally (`ABS(Integer.MIN_VALUE)` overflow, and `Byte`/`TINYINT` missing from `round()`'s type switch).

SEPURI-SAI-KRISHNA:
Pushed one more commit updating the `ceil`/`floor` assertions in `func_numeric.conf`.

`c5` in that config is a `double`, and the Assert sink still expected `ceil(c5)` and `floor(c5)` to be `INT`. That was correct against the old hardcoded `BasicType.INT_TYPE`, but this PR makes those functions return the type of their argument, so they are now `DOUBLE` (`14.0` / `13.0`). The e2e config now asserts that.

This is the only config in the repo that exercises `CEIL`/`FLOOR`, and it is the concrete demonstration of the behaviour change already recorded in `incompatible-changes.md`. The rest of the file is unchanged — I verified that `mod(c4, 5)` still yields `Integer` `3` (the divisor-type dispatch in `mod()` is untouched), `mod(c4, 5.4)` still yields `2.2`, and `round`/`trunc` on a `double` column keep the type they had before.

For the record on the earlier CI run: both `transform-v2-it-part-2` legs were cancelled at the 150-minute job timeout rather than failing, and the `kafka-connector-it` / `all-connectors-it-7` failures are in modules this PR does not touch.

DanielLeens (commented):
@/private/tmp/claude-501/-Volumes-disk6-tmp-seatunnel-third-nervous-swirles-887e0f/f978b90c-4e26-438b-9ad4-2fbf1f56ae20/scratchpad/pr-batch-20260809/drafts/11697.md

SEPURI-SAI-KRISHNA:
Thanks for the follow-up review, @DanielLeens. It looks like the body came through as a local file path (`.../drafts/11697.md`) rather than the review text — the file reference doesn't seem to have been expanded before posting, so I can't see the feedback.

Could you re-post it? For context on what's changed since your first review: all four points from it are addressed, and I've since pushed one more commit (`ed97801`) updating the `ceil`/`floor` assertions in `func_numeric.conf`, which were still expecting `INT` for a `double` column.

davidzollo:

```

```

DanielLeens (commented):
Apologies — my previous review comment on this PR was posted with a broken body (a raw local file path instead of the actual review text) due to a scripting error on my end when publishing it. Here is the review that was intended to be posted:

---

Thanks for the quick follow-up. I re-read the full current diff from scratch (not a delta against my last pass) and re-verified every point I raised last time against the new head. Good news up front: all three concrete, actionable items from my previous review are addressed, and CI has moved from "never ran" to "actively running and green so far." Details below.

The Zeta SQL transform's `CEIL`, `FLOOR`, `ROUND`, `TRUNC` / `TRUNCATE` and `MOD` functions were silently corrupting numeric values instead of raising an error:

1. **Narrowing to `int`**: `NumericFunction.ceil()` / `floor()` always called `.intValue()` on the result, so a `BIGINT` like `9007199254740993` became `1`, and a `DOUBLE` like `1.0E18` became `2147483647`.
2. **Round-tripping through `double`**: `ROUND`'s `BIGDECIMAL` branch and `MOD` both converted their `BigDecimal` operands to `double` before computing, discarding any digits beyond `double`'s ~15-17 significant-digit precision.
3. **Declared-vs-actual type mismatch**: `ZetaSQLType` declared `CEIL`/`FLOOR` as `INT` and `TRUNC` as `DOUBLE` regardless of the argument's real type, so the row's declared schema and its actual runtime value could disagree.

All three are invisible at job-submission and job-completion time — the pipeline reports success, and the wrong value only surfaces later, if at all. The fix keeps the argument's type end to end and does all decimal arithmetic on the exact `BigDecimal`/`long` value instead of a lossy `double`.

# 1. Code Change Review

## 1.1 Core Logic Analysis

I re-read every changed method in `NumericFunction.java` and `ZetaSQLType.java` against the current head (`ed978017`), not against my memory of round 1.

**`ceil` / `floor`** — still `public static Number ceil(...)` / `floor(...)`, still calling the shared `round(v1, v2, RoundingMode.CEILING/FLOOR)` helper without the `.intValue()` narrowing. Unchanged from round 1, still correct.

**`round()`'s `BIGDECIMAL` branch** — still operates on `((BigDecimal) v1).setScale(scale, roundingMode)` directly, no `doubleValue()` round-trip. Unchanged, still correct. This branch is shared by `ROUND`, `CEIL`, `FLOOR` and `TRUNC` for `BigDecimal` inputs (`TRUNC` calls `round(v1, v2, RoundingMode.DOWN)`), so all four benefit — I checked `trunc()`'s implementation directly this round rather than taking that on faith, and it does route through the same shared `round()` helper with no separate narrowing path of its own.

**`MOD`** — still uses the `toBigDecimal()` helper (exact for `BigDecimal`/integral types, `doubleValue()` only for genuinely floating types) and the `rightBD.signum() == 0` guard. Unchanged, still correct.

**`ZetaSQLType.getFunctionType`** — `CEIL`/`CEILING`/`FLOOR`/`TRUNC`/`TRUNCATE` are still in the "return the type of the first argument" group alongside `ABS`/`ROUND`/`NULLIF`. I re-traced the runtime path this round: `ZetaSQLEngine.project()` puts the raw `Object` result from `computeForValue()` straight into the output row with no coercion back to the declared `SeaTunnelDataType`, so the declared-vs-actual contract genuinely depends on this switch statement being right. It still is.

**Comment placement (fixed since round 1)**: last round the explanatory comment sat between `case ROUND:` and `case CEIL:`, reading as if it only documented `ROUND`. The current diff shows it moved to sit *before* `case ABS:`, i.e. above the entire fallthrough group it actually describes:

```java
// These functions all return the type of their first argument. CEIL/FLOOR/TRUNC are
// documented that way as well; declaring INT/DOUBLE for them would truncate BIGINT
// and DECIMAL results.
case ZetaSQLFunction.ABS:
case ZetaSQLFunction.DATEADD:
case ZetaSQLFunction.TIMESTAMPADD:
case ZetaSQLFunction.ROUND:
case ZetaSQLFunction.CEIL:
case ZetaSQLFunction.CEILING:
case ZetaSQLFunction.FLOOR:
case ZetaSQLFunction.TRUNC:
case ZetaSQLFunction.TRUNCATE:
case ZetaSQLFunction.NULLIF:
    return getExpressionType(function.getParameters().getExpressions().get(0));
```

This is a real, targeted fix — the commit history shows a dedicated commit for it ("Clarify type-inference comment placement for CEIL/FLOOR/TRUNC"), and the result correctly reads as documenting the whole group now.

**All title-referenced functions re-verified as covered**: I read `NumericFunction.java` in full this round (not just the diff hunks) to confirm every function named in the PR title is actually fixed, not just the ones with visible diff hunks:
- `ceil` / `floor` — fixed (no narrowing).
- `round()` (the shared private helper, used by the public `round`, `ceil`, `floor`, `trunc`) — `BIGDECIMAL` branch fixed; `DOUBLE`/`FLOAT` branches were never narrowed to `int` in the first place (they already returned `double`/`float`), so no change was needed there — I confirmed this by reading the branch rather than assuming it.
- `trunc()` — was never separately narrowed to `int`; it only suffered from the declared-type bug (`DOUBLE`) and inherits the `BIGDECIMAL` precision fix through the shared `round()` helper. Both are now fixed.
- `mod()` — fixed (exact `BigDecimal` arithmetic, correct zero-guard).

I also re-checked the blast radius: `ZetaSQLFunction.java` still dispatches `ceil`/`floor`/`mod`/`trunc` through untyped `Object`-returning `switch` cases (lines 559-589 in the current head), and the `DECIMAL`-result binary-modulo path (line 781-785) also calls `NumericFunction.mod(args)` through an untyped `Object` return. Widening `ceil`/`floor` from `Integer` to `Number` remains source- and binary-compatible for every caller.

## 1.2 Compatibility Impact

Unchanged from round 1 and still correctly handled: this is an intentional, explicitly documented `[BREAKING]` change.
- `CEIL`/`CEILING`/`FLOOR`/`TRUNC`/`TRUNCATE` now return the argument's data type instead of always `INT` (`CEIL`/`FLOOR`) or a declared-but-wrong `DOUBLE` (`TRUNC`).
- `ROUND`/`TRUNC`/`TRUNCATE`/`MOD` now return exact values for `BIGINT`/`DECIMAL` inputs instead of `double`-degraded ones.

Both are recorded in `docs/en/introduction/concepts/incompatible-changes.md` and its `docs/zh` mirror, each with a before/after table and a migration path. No config option names, defaults, or SPI contracts changed.

**One new observation this round (non-blocking, pre-existing pattern, not introduced by this PR — see Issue 5)**: for `DECIMAL` inputs, the declared output type for `ROUND(decimal_v, N)` / `TRUNC(decimal_v, N)` is "the same `DecimalType` as the argument," which carries the *argument's original* scale, not the scale actually produced by `setScale(N, ...)`. E.g. `ROUND(CAST(... AS DECIMAL(38,9)), 2)` declares `DECIMAL(38,9)` but the returned `BigDecimal` has scale 2. This is not new — `ROUND` already used the "same type as argument" declaration before this PR — and `CEIL`/`FLOOR`/`TRUNC` joining that same group only extends an existing, much smaller mismatch (correct type family, off-by-scale) rather than the severe one this PR fixes (wrong type family entirely, e.g. `Long` declared as `DOUBLE`). Worth a follow-up ticket, not a blocker for this PR.

Unchanged from round 1: `toBigDecimal()` allocates one `BigDecimal` per `MOD` call, same allocation cost as the `BigDecimal.valueOf(...)` calls it replaces. No new locking, threading, or I/O. No change to the DOUBLE/FLOAT branch of `round()` — it still round-trips through `double` for `scale != 0` on native `double`/`float` inputs, which is correct and lossless since a `double` input carries no more precision than a `double` to begin with.

Exception messages remain unchanged and carry sufficient context ("Mod by zero", "Unsupported arg type %s of function %s"). No new logging, nothing to check for secret leakage. No exceptions are swallowed.

**PR description wording (fixed since round 1)**: last round the description said the old `MOD` zero-check "missed divisors that underflow to `0.0`," which had the failure direction backwards (the old code would have false-positived on a tiny nonzero divisor, not missed a real zero). I re-read the current PR description and it now says:

> "a nonzero `DECIMAL` divisor smaller than `Double.MIN_VALUE` underflows to `0.0`, so the old guard rejected a valid divisor with 'Mod by zero'"

That's the correct direction. The code itself was already correct either way, but the history is now accurate too. Good catch by the author.

### Issue tracking (carryover status per item)

| Item | Round-1 status | Round-2 status | Carryover |
|---|---|---|---|
| CI never ran on the fork | Blocking | Resolved — CI is now running (see 2.2) | No — round-1 finding, now fixed |
| Comment placement in `ZetaSQLType.java` | Low, non-blocking | Fixed | No — round-1 finding, now fixed |
| PR description MOD-underflow direction | Low, non-blocking | Fixed | No — round-1 finding, now fixed |
| `ZetaSQLFunction.executeBinaryExpr` decimal `+ - * /` still round-trips through `double` | Low, out-of-scope follow-up | Still present, unchanged, still out-of-scope | No — explicitly flagged as deliberately out of scope by the author in both rounds |
| Declared `DecimalType` scale doesn't reflect actual post-`ROUND`/`TRUNC` scale | Not raised | New observation, Low, non-blocking (see 1.2 / Issue 5) | No — this is a new-this-round observation on a pre-existing pattern, not something visible or missed in the round-1 diff (round 1's `ROUND` support predates this PR, but the `CEIL`/`FLOOR`/`TRUNC` extension of it is new in this diff) |

No apology is owed here: every item above was either already flagged and is now fixed, or is a fresh observation on code that only entered the "same type as argument" declaration group in this PR's diff — it wasn't something visible in the round-1 diff that I missed.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Comments remain well-placed and explain *why*, not just *what* — the `toBigDecimal` Javadoc, the `BIGDECIMAL`-branch inline comment, and the `ceil`/`floor` Javadocs all still read clearly. The one style nit from round 1 (comment placement) is fixed, as covered above.

Unit test coverage is unchanged from round 1 and remains solid: `testCeilAndFloorKeepBigIntValue`, `testCeilAndFloorKeepDoubleType`, `testRoundAndTruncKeepDecimalPrecision`, and `testModKeepsBigIntPrecision` in `zeta/NumericFunctionTest.java` all assert both the declared output type and the actual value, which is exactly the right shape for a declared-vs-actual bug. `SQLNumericFunctionsTest` and the plain `NumericFunctionTest` were correctly updated to the new expected values/types, with net-new assertions added rather than old ones deleted.

**New this round**: `func_numeric.conf` (the transforms-v2-e2e-part-2 E2E resource) was updated in the latest commit — `c5_1`/`c5_2` (`ceil(c5)`/`floor(c5)` on a `double` column) moved from `field_type = "int"` / `equals_to = 14` to `field_type = "double"` / `equals_to = 14.0`, matching the new behavior. I checked the rest of that conf file for any other `CEIL`/`FLOOR`/`TRUNC` usages that might also need updating and found none — `c5_1`/`c5_2` are the only two, and both are correctly updated. This closes a gap round 1 didn't examine (round 1 verified the fix through unit tests and traced the runtime path in source, but didn't check whether an existing E2E fixture would break against the new declared types).

**CI status re-verified live at review time** (not taken from a static snapshot): checking `SEPURI-SAI-KRISHNA/seatunnel`'s `Build` workflow run for head `ed978017` directly via the GitHub API, the module's own unit tests (which include every new/updated assertion above) plus code-style, license-header and dependency-license gates had already completed and passed. The specific `transform-v2-it-part-2` job that actually runs the updated `func_numeric.conf` was still queued — normal for this repository's large (~84-job) CI matrix, not a sign of a problem, since nothing in the run had failed. This is a sharp contrast to round 1, where the fork's `Build` workflow had never executed at all (`action_required`, 0 recorded runs) — that was a genuine blocker; this round it's simply mid-flight with everything green so far. I'd still want to see the full matrix land green (in particular that `transform-v2-it-part-2` job) before merge, but there is no current evidence of a problem, and the branch is otherwise unblocked.

**Flaky-test re-check**: same analysis as round 1 holds — no `Thread.sleep`, no shared mutable state, no timing dependence in the touched tests, and the floating-point assertions compare values that passed through `Math.ceil`/`Math.floor` on already-integral inputs, so they're comparing identical bit patterns rather than independently-computed floats.

**Stability rating: Stable.**

## 2.3 Documentation Updates

Unchanged from round 1: `docs/en/transforms/sql-functions.md` and its `docs/zh` mirror already documented the correct contract before this PR, so no function-reference doc changes were needed and none were made — correct scoping. The `incompatible-changes.md` additions in both `en` and `zh` remain thorough, accurate against the current diff, and consistently formatted with the rest of the file. I re-read both language versions this round to confirm they still match the current code behavior (they do — nothing in the latest two commits changed the observable before/after values described there).

# 3. Architectural Soundness

The fix stays minimal and correctly scoped to the numeric-function implementation and its type-declaration counterpart. Nothing connector-specific or engine-specific leaks in.

## 3.2 Maintainability

The `toBigDecimal()` helper is a clean, single-purpose extraction that's now shared correctly by `mod()`. The comment-placement fix improves readability of the `ZetaSQLType` switch without changing its behavior.

## 3.3 Extensibility

The "declared type = type of first argument" group is now a well-established, easy-to-extend pattern for future numeric functions with the same contract. The one caveat is the DECIMAL-scale gap noted in 1.2/Issue 5, which any future function joining this group (or a dedicated follow-up PR) should account for.

Same conclusion as round 1: this is a pure computed-value/declared-type correction with no config, API, or SPI surface change, so there's no checkpoint/savepoint/upgrade-path concern. The breaking-change documentation and migration guidance are the right mechanism for this class of change, and both are present and unchanged in substance since round 1.

# 4. Issue Summary

|---|---|---|---|
| 1 | Fork `Build` workflow was still mid-matrix at review time (all completed jobs green, `transform-v2-it-part-2` — the job that runs the updated `func_numeric.conf` — still queued) | GitHub Actions, `SEPURI-SAI-KRISHNA/seatunnel` run for `ed978017` | Low (informational — confirm full green before merge, not currently failing) |
| 2 | Pre-existing, out-of-scope: `DECIMAL` binary arithmetic (`+ - * /`) in `executeBinaryExpr` still round-trips through `double`, same bug class this PR fixes for function calls | `seatunnel-transforms-v2/.../zeta/ZetaSQLFunction.java:764-780` | Low (follow-up, explicitly out of scope per author) |
| 3 | Declared `DecimalType` scale for `ROUND`/`TRUNC` on `DECIMAL` inputs doesn't reflect the actual post-operation scale (pre-existing for `ROUND`, marginally extended to `CEIL`/`FLOOR`/`TRUNC` by this PR) | `seatunnel-transforms-v2/.../zeta/ZetaSQLType.java:463-473` | Low (follow-up, not a regression from this PR) |

All three are non-blocking. The comment-placement and PR-description issues from round 1 are fixed and no longer appear in this table.

# 5. Merge Recommendation

### Conclusion: Ready to merge.

1. **Blockers — must be fixed**: none. Every code-level concern from round 1 is resolved, and I found no new code defects this round after a full re-read of `NumericFunction.java`, `ZetaSQLType.java`, and the relevant `ZetaSQLFunction.java` call sites.
2. **Recommended fixes — non-blocking**: Issues 1-3 above — confirm the full CI matrix (particularly `transform-v2-it-part-2`) lands green before the actual merge, and consider follow-up tickets for the pre-existing decimal binary-arithmetic precision gap and the declared-scale-vs-actual-scale gap for `ROUND`/`TRUNC` on `DECIMAL`.

This is a clean, well-tested, honestly-documented fix for a real production-data-integrity bug, and the author addressed every actionable point from the first review precisely and correctly — including a wording correction in the PR description that most contributors wouldn't have bothered with. Once the CI matrix finishes green, this is good to merge.

davidzollo (commented):
[This comment was posted in error from the wrong account due to a scripting mistake. Please disregard this review — see the DanielLeens review below for the correct, authoritative content.]

DanielLeens (commented):
Apologies for the noise on this thread. My previous review comment on this PR was posted with a broken body due to a scripting error on my end (it published a raw local file path instead of the intended review text), and the follow-up correction was accidentally posted from a different account due to a second scripting error on my end. This review, from my correct account, is the one that should be treated as authoritative. Here is the review that was intended to be posted, unchanged in content:

Thanks for the quick follow-up. I re-read the full current diff from scratch (not a delta against my last pass) and re-verified every point I raised last time against the new head. Good news up front: all three concrete, actionable items from my previous review are addressed, and CI has moved from "never ran" to "actively running and green so far." Details below.

The Zeta SQL transform's `CEIL`, `FLOOR`, `ROUND`, `TRUNC` / `TRUNCATE` and `MOD` functions were silently corrupting numeric values instead of raising an error:

1. **Narrowing to `int`**: `NumericFunction.ceil()` / `floor()` always called `.intValue()` on the result, so a `BIGINT` like `9007199254740993` became `1`, and a `DOUBLE` like `1.0E18` became `2147483647`.
2. **Round-tripping through `double`**: `ROUND`'s `BIGDECIMAL` branch and `MOD` both converted their `BigDecimal` operands to `double` before computing, discarding any digits beyond `double`'s ~15-17 significant-digit precision.
3. **Declared-vs-actual type mismatch**: `ZetaSQLType` declared `CEIL`/`FLOOR` as `INT` and `TRUNC` as `DOUBLE` regardless of the argument's real type, so the row's declared schema and its actual runtime value could disagree.

All three are invisible at job-submission and job-completion time — the pipeline reports success, and the wrong value only surfaces later, if at all. The fix keeps the argument's type end to end and does all decimal arithmetic on the exact `BigDecimal`/`long` value instead of a lossy `double`.

# 1. Code Change Review

## 1.1 Core Logic Analysis

I re-read every changed method in `NumericFunction.java` and `ZetaSQLType.java` against the current head (`ed978017`), not against my memory of round 1.

**`ceil` / `floor`** — still `public static Number ceil(...)` / `floor(...)`, still calling the shared `round(v1, v2, RoundingMode.CEILING/FLOOR)` helper without the `.intValue()` narrowing. Unchanged from round 1, still correct.

**`round()`'s `BIGDECIMAL` branch** — still operates on `((BigDecimal) v1).setScale(scale, roundingMode)` directly, no `doubleValue()` round-trip. Unchanged, still correct. This branch is shared by `ROUND`, `CEIL`, `FLOOR` and `TRUNC` for `BigDecimal` inputs (`TRUNC` calls `round(v1, v2, RoundingMode.DOWN)`), so all four benefit — I checked `trunc()`'s implementation directly this round rather than taking that on faith, and it does route through the same shared `round()` helper with no separate narrowing path of its own.

**`MOD`** — still uses the `toBigDecimal()` helper (exact for `BigDecimal`/integral types, `doubleValue()` only for genuinely floating types) and the `rightBD.signum() == 0` guard. Unchanged, still correct.

**`ZetaSQLType.getFunctionType`** — `CEIL`/`CEILING`/`FLOOR`/`TRUNC`/`TRUNCATE` are still in the "return the type of the first argument" group alongside `ABS`/`ROUND`/`NULLIF`. I re-traced the runtime path this round: `ZetaSQLEngine.project()` puts the raw `Object` result from `computeForValue()` straight into the output row with no coercion back to the declared `SeaTunnelDataType`, so the declared-vs-actual contract genuinely depends on this switch statement being right. It still is.

**Comment placement (fixed since round 1)**: last round the explanatory comment sat between `case ROUND:` and `case CEIL:`, reading as if it only documented `ROUND`. The current diff shows it moved to sit *before* `case ABS:`, i.e. above the entire fallthrough group it actually describes:

```java
// These functions all return the type of their first argument. CEIL/FLOOR/TRUNC are
// documented that way as well; declaring INT/DOUBLE for them would truncate BIGINT
// and DECIMAL results.
case ZetaSQLFunction.ABS:
case ZetaSQLFunction.DATEADD:
case ZetaSQLFunction.TIMESTAMPADD:
case ZetaSQLFunction.ROUND:
case ZetaSQLFunction.CEIL:
case ZetaSQLFunction.CEILING:
case ZetaSQLFunction.FLOOR:
case ZetaSQLFunction.TRUNC:
case ZetaSQLFunction.TRUNCATE:
case ZetaSQLFunction.NULLIF:
    return getExpressionType(function.getParameters().getExpressions().get(0));
```

This is a real, targeted fix — the commit history shows a dedicated commit for it ("Clarify type-inference comment placement for CEIL/FLOOR/TRUNC"), and the result correctly reads as documenting the whole group now.

**All title-referenced functions re-verified as covered**: I read `NumericFunction.java` in full this round (not just the diff hunks) to confirm every function named in the PR title is actually fixed, not just the ones with visible diff hunks:
- `ceil` / `floor` — fixed (no narrowing).
- `round()` (the shared private helper, used by the public `round`, `ceil`, `floor`, `trunc`) — `BIGDECIMAL` branch fixed; `DOUBLE`/`FLOAT` branches were never narrowed to `int` in the first place (they already returned `double`/`float`), so no change was needed there — I confirmed this by reading the branch rather than assuming it.
- `trunc()` — was never separately narrowed to `int`; it only suffered from the declared-type bug (`DOUBLE`) and inherits the `BIGDECIMAL` precision fix through the shared `round()` helper. Both are now fixed.
- `mod()` — fixed (exact `BigDecimal` arithmetic, correct zero-guard).

I also re-checked the blast radius: `ZetaSQLFunction.java` still dispatches `ceil`/`floor`/`mod`/`trunc` through untyped `Object`-returning `switch` cases (lines 559-589 in the current head), and the `DECIMAL`-result binary-modulo path (line 781-785) also calls `NumericFunction.mod(args)` through an untyped `Object` return. Widening `ceil`/`floor` from `Integer` to `Number` remains source- and binary-compatible for every caller.

## 1.2 Compatibility Impact

Unchanged from round 1 and still correctly handled: this is an intentional, explicitly documented `[BREAKING]` change.
- `CEIL`/`CEILING`/`FLOOR`/`TRUNC`/`TRUNCATE` now return the argument's data type instead of always `INT` (`CEIL`/`FLOOR`) or a declared-but-wrong `DOUBLE` (`TRUNC`).
- `ROUND`/`TRUNC`/`TRUNCATE`/`MOD` now return exact values for `BIGINT`/`DECIMAL` inputs instead of `double`-degraded ones.

Both are recorded in `docs/en/introduction/concepts/incompatible-changes.md` and its `docs/zh` mirror, each with a before/after table and a migration path. No config option names, defaults, or SPI contracts changed.

**One new observation this round (non-blocking, pre-existing pattern, not introduced by this PR — see Issue 5)**: for `DECIMAL` inputs, the declared output type for `ROUND(decimal_v, N)` / `TRUNC(decimal_v, N)` is "the same `DecimalType` as the argument," which carries the *argument's original* scale, not the scale actually produced by `setScale(N, ...)`. E.g. `ROUND(CAST(... AS DECIMAL(38,9)), 2)` declares `DECIMAL(38,9)` but the returned `BigDecimal` has scale 2. This is not new — `ROUND` already used the "same type as argument" declaration before this PR — and `CEIL`/`FLOOR`/`TRUNC` joining that same group only extends an existing, much smaller mismatch (correct type family, off-by-scale) rather than the severe one this PR fixes (wrong type family entirely, e.g. `Long` declared as `DOUBLE`). Worth a follow-up ticket, not a blocker for this PR.

Unchanged from round 1: `toBigDecimal()` allocates one `BigDecimal` per `MOD` call, same allocation cost as the `BigDecimal.valueOf(...)` calls it replaces. No new locking, threading, or I/O. No change to the DOUBLE/FLOAT branch of `round()` — it still round-trips through `double` for `scale != 0` on native `double`/`float` inputs, which is correct and lossless since a `double` input carries no more precision than a `double` to begin with.

Exception messages remain unchanged and carry sufficient context ("Mod by zero", "Unsupported arg type %s of function %s"). No new logging, nothing to check for secret leakage. No exceptions are swallowed.

**PR description wording (fixed since round 1)**: last round the description said the old `MOD` zero-check "missed divisors that underflow to `0.0`," which had the failure direction backwards (the old code would have false-positived on a tiny nonzero divisor, not missed a real zero). I re-read the current PR description and it now says:

> "a nonzero `DECIMAL` divisor smaller than `Double.MIN_VALUE` underflows to `0.0`, so the old guard rejected a valid divisor with 'Mod by zero'"

That's the correct direction. The code itself was already correct either way, but the history is now accurate too. Good catch by the author.

### Issue tracking (carryover status per item)

| Item | Round-1 status | Round-2 status | Carryover |
|---|---|---|---|
| CI never ran on the fork | Blocking | Resolved — CI is now running (see 2.2) | No — round-1 finding, now fixed |
| Comment placement in `ZetaSQLType.java` | Low, non-blocking | Fixed | No — round-1 finding, now fixed |
| PR description MOD-underflow direction | Low, non-blocking | Fixed | No — round-1 finding, now fixed |
| `ZetaSQLFunction.executeBinaryExpr` decimal `+ - * /` still round-trips through `double` | Low, out-of-scope follow-up | Still present, unchanged, still out-of-scope | No — explicitly flagged as deliberately out of scope by the author in both rounds |
| Declared `DecimalType` scale doesn't reflect actual post-`ROUND`/`TRUNC` scale | Not raised | New observation, Low, non-blocking (see 1.2 / Issue 5) | No — this is a new-this-round observation on a pre-existing pattern, not something visible or missed in the round-1 diff (round 1's `ROUND` support predates this PR, but the `CEIL`/`FLOOR`/`TRUNC` extension of it is new in this diff) |

No apology is owed here: every item above was either already flagged and is now fixed, or is a fresh observation on code that only entered the "same type as argument" declaration group in this PR's diff — it wasn't something visible in the round-1 diff that I missed.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Comments remain well-placed and explain *why*, not just *what* — the `toBigDecimal` Javadoc, the `BIGDECIMAL`-branch inline comment, and the `ceil`/`floor` Javadocs all still read clearly. The one style nit from round 1 (comment placement) is fixed, as covered above.

Unit test coverage is unchanged from round 1 and remains solid: `testCeilAndFloorKeepBigIntValue`, `testCeilAndFloorKeepDoubleType`, `testRoundAndTruncKeepDecimalPrecision`, and `testModKeepsBigIntPrecision` in `zeta/NumericFunctionTest.java` all assert both the declared output type and the actual value, which is exactly the right shape for a declared-vs-actual bug. `SQLNumericFunctionsTest` and the plain `NumericFunctionTest` were correctly updated to the new expected values/types, with net-new assertions added rather than old ones deleted.

**New this round**: `func_numeric.conf` (the transforms-v2-e2e-part-2 E2E resource) was updated in the latest commit — `c5_1`/`c5_2` (`ceil(c5)`/`floor(c5)` on a `double` column) moved from `field_type = "int"` / `equals_to = 14` to `field_type = "double"` / `equals_to = 14.0`, matching the new behavior. I checked the rest of that conf file for any other `CEIL`/`FLOOR`/`TRUNC` usages that might also need updating and found none — `c5_1`/`c5_2` are the only two, and both are correctly updated. This closes a gap round 1 didn't examine (round 1 verified the fix through unit tests and traced the runtime path in source, but didn't check whether an existing E2E fixture would break against the new declared types).

**CI status re-verified live at review time** (not taken from a static snapshot): checking `SEPURI-SAI-KRISHNA/seatunnel`'s `Build` workflow run for head `ed978017` directly via the GitHub API, the module's own unit tests (which include every new/updated assertion above) plus code-style, license-header and dependency-license gates had already completed and passed. The specific `transform-v2-it-part-2` job that actually runs the updated `func_numeric.conf` was still queued — normal for this repository's large (~84-job) CI matrix, not a sign of a problem, since nothing in the run had failed. This is a sharp contrast to round 1, where the fork's `Build` workflow had never executed at all (`action_required`, 0 recorded runs) — that was a genuine blocker; this round it's simply mid-flight with everything green so far. I'd still want to see the full matrix land green (in particular that `transform-v2-it-part-2` job) before merge, but there is no current evidence of a problem, and the branch is otherwise unblocked.

**Flaky-test re-check**: same analysis as round 1 holds — no `Thread.sleep`, no shared mutable state, no timing dependence in the touched tests, and the floating-point assertions compare values that passed through `Math.ceil`/`Math.floor` on already-integral inputs, so they're comparing identical bit patterns rather than independently-computed floats.

**Stability rating: Stable.**

## 2.3 Documentation Updates

Unchanged from round 1: `docs/en/transforms/sql-functions.md` and its `docs/zh` mirror already documented the correct contract before this PR, so no function-reference doc changes were needed and none were made — correct scoping. The `incompatible-changes.md` additions in both `en` and `zh` remain thorough, accurate against the current diff, and consistently formatted with the rest of the file. I re-read both language versions this round to confirm they still match the current code behavior (they do — nothing in the latest two commits changed the observable before/after values described there).

# 3. Architectural Soundness

The fix stays minimal and correctly scoped to the numeric-function implementation and its type-declaration counterpart. Nothing connector-specific or engine-specific leaks in.

## 3.2 Maintainability

The `toBigDecimal()` helper is a clean, single-purpose extraction that's now shared correctly by `mod()`. The comment-placement fix improves readability of the `ZetaSQLType` switch without changing its behavior.

## 3.3 Extensibility

The "declared type = type of first argument" group is now a well-established, easy-to-extend pattern for future numeric functions with the same contract. The one caveat is the DECIMAL-scale gap noted in 1.2/Issue 5, which any future function joining this group (or a dedicated follow-up PR) should account for.

Same conclusion as round 1: this is a pure computed-value/declared-type correction with no config, API, or SPI surface change, so there's no checkpoint/savepoint/upgrade-path concern. The breaking-change documentation and migration guidance are the right mechanism for this class of change, and both are present and unchanged in substance since round 1.

# 4. Issue Summary

|---|---|---|---|
| 1 | Fork `Build` workflow was still mid-matrix at review time (all completed jobs green, `transform-v2-it-part-2` — the job that runs the updated `func_numeric.conf` — still queued) | GitHub Actions, `SEPURI-SAI-KRISHNA/seatunnel` run for `ed978017` | Low (informational — confirm full green before merge, not currently failing) |
| 2 | Pre-existing, out-of-scope: `DECIMAL` binary arithmetic (`+ - * /`) in `executeBinaryExpr` still round-trips through `double`, same bug class this PR fixes for function calls | `seatunnel-transforms-v2/.../zeta/ZetaSQLFunction.java:764-780` | Low (follow-up, explicitly out of scope per author) |
| 3 | Declared `DecimalType` scale for `ROUND`/`TRUNC` on `DECIMAL` inputs doesn't reflect the actual post-operation scale (pre-existing for `ROUND`, marginally extended to `CEIL`/`FLOOR`/`TRUNC` by this PR) | `seatunnel-transforms-v2/.../zeta/ZetaSQLType.java:463-473` | Low (follow-up, not a regression from this PR) |

All three are non-blocking. The comment-placement and PR-description issues from round 1 are fixed and no longer appear in this table.

# 5. Merge Recommendation

### Conclusion: Ready to merge.

1. **Blockers — must be fixed**: none. Every code-level concern from round 1 is resolved, and I found no new code defects this round after a full re-read of `NumericFunction.java`, `ZetaSQLType.java`, and the relevant `ZetaSQLFunction.java` call sites.
2. **Recommended fixes — non-blocking**: Issues 1-3 above — confirm the full CI matrix (particularly `transform-v2-it-part-2`) lands green before the actual merge, and consider follow-up tickets for the pre-existing decimal binary-arithmetic precision gap and the declared-scale-vs-actual-scale gap for `ROUND`/`TRUNC` on `DECIMAL`.

This is a clean, well-tested, honestly-documented fix for a real production-data-integrity bug, and the author addressed every actionable point from the first review precisely and correctly — including a wording correction in the PR description that most contributors wouldn't have bothered with. Once the CI matrix finishes green, this is good to merge.

DanielLeens (commented):
Quick follow-up, no new code review needed here — just closing the loop on the one open item from my last pass.

In that review the only thing standing between this PR and "ready to merge" was confirming the full CI matrix landed green, since `transform-v2-it-part-2` (the job that actually runs the updated `func_numeric.conf` e2e assertions) was still queued at review time. I just re-checked the fork's `Build` workflow run for the current head (`ed978017`) directly:

- Run `31308652979` is now `completed` / `success`.
- `transform-v2-it-part-2 (8, ubuntu-latest)` and `transform-v2-it-part-2 (11, ubuntu-latest)` — the jobs I was specifically waiting on — both finished `success`.
- Across the whole run: 75 jobs success, 9 skipped (normal matrix conditionals), 0 failures, 0 cancellations.
- No new commits and no new discussion have landed since my last review, so this is purely a CI-status update, not a re-review — my prior read of `NumericFunction.java` / `ZetaSQLType.java` / the e2e fixture still stands as-is.

So from a technical standpoint there's no remaining blocker on this PR. `mergeStateStatus` currently shows `BLOCKED`, but that's just the repo's required-review gate (`reviewDecision: REVIEW_REQUIRED`) — `mergeable` is `MERGEABLE`, i.e. no merge conflict.

Since my account only has comment-level access here, I can't perform the actual approval — a maintainer with write access will need to do that step. But as far as the code and CI are concerned, this looks ready to merge whenever a maintainer has a moment. Nice work getting the full matrix green, @SEPURI-SAI-KRISHNA.

davidzollo (approved):
The declared-vs-actual numeric type bug, plus the double round-trip precision loss, is a real and easy-to-miss data-integrity issue, and the fix plus the incompatible-changes documentation are solid. Full CI matrix is green as of the latest head, including the transform-v2-it-part-2 e2e job that exercises the updated func_numeric.conf. Approving to clear the review gate — merging now.

PR #11538, opened 2026-07-23
[Feature][Zeta] Add local sample dry-run mode

goutamadwant:
### Purpose of this pull request

Closes #9764.

This draft adds the local-only first slice of Zeta debug mode as `--dry-run sample`.

- runs the configured source and transform chain with a bounded row sample
- prints source and transform schemas and sample rows
- replaces configured sinks before sink factory creation and save-mode handling
- disables checkpoints and uses source parallelism 1 for deterministic per-source limits
- rejects cluster, async, restore, and savepoint execution for sample mode
- keeps sink DDL preview out of this slice until a no-write preview contract is agreed

### Does this PR introduce _any_ user-facing change?

Yes.

Users can run a local sample preview with:

```shell
bin/seatunnel.sh --master local --config config/v2.batch.config.template --dry-run sample --sample-limit 10
```

The command reads up to the configured number of rows from each source and runs the transform chain without creating the configured sink or writing to the target system. Source reads are real and may access external systems.

### How was this patch tested?

Added coverage for:

- CLI parsing, defaults, validation, and local execution
- replacement of configured sinks before sink runtime creation
- bounded source collection and normal completion
- bounded output reporting across a multi-transform chain
- a local FakeSource job confirming that the configured InMemory sink is not used

Verified with:

```shell
./mvnw -pl seatunnel-core/seatunnel-starter -Dtest=ClientCommandArgsTest test
./mvnw -pl seatunnel-engine/seatunnel-engine-client -Dtest=MultipleTableJobConfigParserTest test
./mvnw -pl seatunnel-engine/seatunnel-engine-server -Dtest=SeaTunnelSourceCollectorDryRunSampleTest,TransformFlowLifeCycleDryRunSampleTest test
./mvnw -pl seatunnel-engine/seatunnel-engine-common,seatunnel-engine/seatunnel-engine-core,seatunnel-engine/seatunnel-engine-server,seatunnel-engine/seatunnel-engine-client,seatunnel-core/seatunnel-starter -DskipTests package
```

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md)
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR.
* [ ] If you are contributing the connector code, please check that the following files are updated:
[...]

PR #11700, opened 2026-08-08
[Fix][E2E] Stabilize JDBC HANA and OpenGauss tests

danielnadean:
### Purpose

Stabilize unrelated JDBC E2E failures observed while checking apache/seatunnel#11538. The failures are not caused by the dry-run sample changes in that PR; they come from JDBC E2E container startup/readiness behavior.

### Changes

- Configure the SAP HANA test container with `seccomp=unconfined`, matching the HANA container startup error that requests seccomp permissions for `move_pages` and `mbind`.
- Start the OpenGauss test container in privileged mode, matching the official image startup requirement.
- Add an OpenGauss `gsql select 1` readiness check before the JDBC E2E base class starts creating schemas/tables, so host-port readiness does not race ahead of authenticated SQL readiness.

### Verification

- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-6,seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7 spotless:apply`
- `./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-6,seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7 -DskipTests -DskipIT -Dskip.spotless=true test-compile`

A wider `-am test-compile` did not reach these modules locally because `seatunnel-config-shade` failed earlier with missing shaded Typesafe Config classes. GitHub CI remains the final E2E verifier.

### DIFF
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-6/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/JdbcHanaIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-6/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/JdbcHanaIT.java
@@ -36,6 +36,7 @@ import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.utility.DockerLoggerFactory;
 
+import com.github.dockerjava.api.model.HostConfig;
 import lombok.SneakyThrows;
 
 import java.sql.Date;
@@ -271,6 +272,18 @@ public class JdbcHanaIT extends AbstractJdbcIT {
                         .withCommand("--master-password", PASSWORD, "--agree-to-sap-license")
                         .withLogConsumer(
                                 new Slf4jLogConsumer(DockerLoggerFactory.getLogger(HANA_IMAGE)))
+                        .withCreateContainerCmdModifier(
+                                cmd -> {
+                                    // HANA requires move_pages and mbind during startup on GitHub
+                                    // runners.
+                                    HostConfig hostConfig = cmd.getHostConfig();
+                                    if (hostConfig == null) {
+                                        hostConfig = HostConfig.newHostConfig();
+                                        cmd.withHostConfig(hostConfig);
+                                    }
+                                    hostConfig.withSecurityOpts(
+                                            Lists.newArrayList("seccomp=unconfined"));
+                                })
                         .waitingFor(
                                 Wait.forLogMessage(".*Startup finished!.*", 1)
                                         .withStartupTimeout(Duration.of(5, ChronoUnit.MINUTES)));
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/JdbcOpenGaussIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/JdbcOpenGaussIT.java
@@ -55,7 +55,10 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+
+import static org.awaitility.Awaitility.given;
 
 @Slf4j
 public class JdbcOpenGaussIT extends AbstractJdbcIT {
@@ -351,6 +354,8 @@ public class JdbcOpenGaussIT extends AbstractJdbcIT {
                         .withNetwork(NETWORK)
                         .withNetworkAliases(OPEN_GAUSS_ALIASES)
                         .withEnv("GS_PASSWORD", PASSWORD)
+                        // The official OpenGauss image requires privileged mode for CI startup.
+                        .withPrivilegedMode(true)
                         .waitingFor(
                                 Wait.forListeningPort().withStartupTimeout(Duration.ofMinutes(5)))
                         .withLogConsumer(
@@ -360,6 +365,38 @@ public class JdbcOpenGaussIT extends AbstractJdbcIT {
                 Lists.newArrayList(String.format("%s:%s", OPEN_GAUSS_PORT, OPEN_GAUSS_PORT)));
 
         return container;
+    }
+
+    /**
+     * Waits until OpenGauss can authenticate a real SQL session inside the container.
+     *
+     * <p>The first host-port readiness signal can arrive before OpenGauss accepts authenticated
+     * sessions, which makes downstream SeaTunnel JDBC catalog creation fail with EOF.
+     */
+    @Override
+    protected void beforeStartUP() {
+        given().ignoreExceptions()
+                .await()
+                .atMost(120, TimeUnit.SECONDS)
+                .untilAsserted(
+                        () -> {
+                            Container.ExecResult result =
+                                    dbServer.execInContainer(
+                                            "gsql",
+                                            "-h",
+                                            "localhost",
+                                            "-p",
+                                            String.valueOf(OPEN_GAUSS_PORT),
+                                            "-d",
+                                            DATABASE,
+                                            "-U",
+                                            USERNAME,
+                                            "-W",
+                                            PASSWORD,
+                                            "-c",
+                                            "select 1");
+                            Assertions.assertEquals(0, result.getExitCode(), result.getStderr());
+                        });
     }
 
     @Override


DanielLeens (changes requested):

The JDBC E2E suite has two SAP HANA / OpenGauss containers that are flaky in CI, independently of the changes in #11538 that first surfaced them:

- `JdbcHanaIT` sometimes fails to start the HANA container because HANA's own startup routine needs the `move_pages`/`mbind` syscalls, which Docker's default seccomp profile blocks.
- `JdbcOpenGaussIT` sometimes fails downstream JDBC/catalog operations with an EOF, because the container's TCP port opens before OpenGauss is actually ready to serve authenticated SQL sessions, so the test proceeds too early.

The fix approach is: (1) relax the HANA container's seccomp profile via `withCreateContainerCmdModifier`, (2) start the OpenGauss container in privileged mode as the vendor image requires, and (3) add a new `beforeStartUP()` override that polls `gsql ... select 1` inside the OpenGauss container before the rest of `AbstractJdbcIT#startUp()` runs.

One-sentence summary: this PR attempts to stabilize two flaky JDBC E2E containers by loosening container security constraints and adding an authentication-readiness probe, but the readiness probe itself is broken and turns `JdbcOpenGaussIT` from "occasionally flaky" into "always fails."

Concrete before/after: previously, `JdbcOpenGaussIT` would race straight from "TCP port open" into `initializeJdbcConnection()` (which does retry for up to 360s) and then into non-retried `createNeededTables()/insertTestData()/initCatalog()`, so it *could* fail intermittently on an EOF. After this PR, every run of `JdbcOpenGaussIT` now spends a fixed 2 minutes inside the new `beforeStartUP()` gsql check and then fails deterministically with `expected: <0> but was: <127>` — confirmed on the PR's own fork CI on both JDK 8 and JDK 11 (see Issue 1 below). So the "fix" currently makes the situation strictly worse for OpenGauss (100% failure vs. intermittent failure).

## 1. Code Change Review

### 1.1 Core Logic Analysis

**HANA container (`JdbcHanaIT.initContainer()`, `seatunnel-e2e/.../connector-jdbc-e2e-part-6/.../JdbcHanaIT.java:272-287`)**

```java
.withCreateContainerCmdModifier(
        cmd -> {
            // HANA requires move_pages and mbind during startup on GitHub runners.
            HostConfig hostConfig = cmd.getHostConfig();
            if (hostConfig == null) {
                hostConfig = HostConfig.newHostConfig();
                cmd.withHostConfig(hostConfig);
            }
            hostConfig.withSecurityOpts(
                    Lists.newArrayList("seccomp=unconfined"));
        })
```

This runs at container-creation time, on the normal startup path for every `JdbcHanaIT` run (not a recovery/boundary branch) — it's exactly the mechanism testcontainers exposes for passing raw `docker create` options, and the null-check on `getHostConfig()` before mutating it is defensively correct. The same `withCreateContainerCmdModifier(cmd -> ...)` pattern is already used elsewhere in the tree (e.g. `connector-aerospike-e2e/.../AbstractAerospikeIT.java`, `connector-hive-e2e/.../HiveKerberosIT.java`, `connector-mongodb-e2e/.../AbstractMongodbIT.java`), so `docker-java-api`'s `HostConfig`/`CreateContainerCmd` types are already reachable from the `testcontainers` test-scope dependency; this is however the *first* usage in the codebase of `HostConfig.newHostConfig()` + `withSecurityOpts(...)` specifically, and the PR author's own description states the local `-am test-compile` never reached this module (blocked earlier by an unrelated `seatunnel-config-shade` failure), so this exact API call has not actually been compiled by anyone yet, locally or in CI (see Issue 2).

**OpenGauss container (`JdbcOpenGaussIT.java:350-368`)**: adds `.withPrivilegedMode(true)`, matching the vendor's documented requirement — reasonable and low-risk (see 1.3).

**OpenGauss readiness probe (`JdbcOpenGaussIT.java:376-400`)**:

```java
@Override
protected void beforeStartUP() {
    given().ignoreExceptions()
            .await()
            .atMost(120, TimeUnit.SECONDS)
            .untilAsserted(
                    () -> {
                        Container.ExecResult result =
                                dbServer.execInContainer(
                                        "gsql", "-h", "localhost", "-p",
                                        String.valueOf(OPEN_GAUSS_PORT), "-d", DATABASE,
                                        "-U", USERNAME, "-W", PASSWORD, "-c", "select 1");
                        Assertions.assertEquals(0, result.getExitCode(), result.getStderr());
                    });
}
```

`beforeStartUP()` is invoked from `AbstractJdbcIT#startUp()` (`seatunnel-e2e/.../connector-jdbc-e2e-common/.../AbstractJdbcIT.java:310-328`) right after `Startables.deepStart(...)` returns (container already up, TCP port already open) and right before the pre-existing, retried `initializeJdbcConnection()` call — so this new check runs unconditionally on the normal startup path of every `JdbcOpenGaussIT` execution, not a rare edge case.

**This is where the PR's own CI shows the fix does not work.** I pulled the fork's "Build" run for this exact commit (`e8a4d2daa5ebf8fcb7855bf04341d573d7deab1d`, matches the PR's single commit and its head SHA):

- Run: https://github.com/danielnadean/seatunnel/actions/runs/31236260168
- Job (JDK 11): https://github.com/danielnadean/seatunnel/actions/runs/31236260168/job/93050264839
- Job (JDK 8): https://github.com/danielnadean/seatunnel/actions/runs/31236260168/job/93050264842

Both jobs report the identical failure:

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 142.154 s <<< FAILURE! - in org.apache.seatunnel.connectors.seatunnel.jdbc.JdbcOpenGaussIT
org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a lambda expression in org.apache.seatunnel.connectors.seatunnel.jdbc.JdbcOpenGaussIT expected: <0> but was: <127> within 2 minutes.
	at org.apache.seatunnel.connectors.seatunnel.jdbc.JdbcOpenGaussIT.beforeStartUP(JdbcOpenGaussIT.java:381)
```

Exit code **127** is the shell/Docker-exec convention for "command not found" — `gsql` is not resolvable in the `execInContainer` exec environment. Interestingly, the container's own entrypoint script *does* successfully invoke `gsql` during its own startup (visible a few seconds earlier in the same job's container log: `Execute SQL: gsql -v ON_ERROR_STOP=1 --username opengauss ...`), which proves `gsql` exists in the image but is only resolvable in the specific shell context the entrypoint script uses (OpenGauss's official image typically requires running as the `omm` OS user via a login shell that sources `~/.bashrc`/profile to get `GAUSSHOME`/`PATH` set up). `GenericContainer#execInContainer(String...)` runs `docker exec <id> gsql ...` directly as `argv`, without a login shell and without `su - omm`, so it never picks up that environment — hence a deterministic, 100%-reproducing exit 127, not a race-dependent flake. Since `given().ignoreExceptions()` swallows the `AssertionError` on every iteration, the loop just burns the full 120s before surfacing the `ConditionTimeoutException`, which also makes every OpenGauss E2E run in CI ~2 minutes slower for no benefit.

**Suggested fix (Issue 1)** — invoke `gsql` through a shell that replicates what the entrypoint does, e.g.:
```java
dbServer.execInContainer("su", "-", "omm", "-c",
        "gsql -h localhost -p " + OPEN_GAUSS_PORT + " -d " + DATABASE
                + " -U " + USERNAME + " -W " + PASSWORD + " -c \"select 1\"");
```
(or resolve/hardcode the absolute path + required `LD_LIBRARY_PATH`/`GAUSSHOME` via `bash -c "export ...; gsql ..."`), and re-verify this actually returns 0 in CI before merge.

### 1.2 Compatibility

Fully compatible. No production `src/main/**` code, no `Option`/config surface, no protocol/serialization change — this is entirely E2E test scaffolding (`src/test/java`). No backward-compatibility concerns.

- `--privileged` on the OpenGauss test container and `seccomp=unconfined` on the HANA test container both widen the container's syscall/capability surface. This is scoped to short-lived, ephemeral CI containers that are always started fresh and torn down in `AbstractJdbcIT#tearDown()` (`AbstractJdbcIT.java:333-364`), matching the documented requirements of these two vendor images (OpenGauss's official image requires `--privileged`; HANA needs the NUMA syscalls) — an acceptable, narrowly-scoped trade-off, not a production concern.
- The broken `gsql` probe adds a fixed ~2 minutes of dead time to every `JdbcOpenGaussIT` run (whether or not the real race would have occurred), which is a real, currently-active CI-time regression until Issue 1 is fixed.

The `beforeStartUP()` assertion passes `result.getStderr()` as the assertion message, which is good practice for diagnosability — but as observed in the actual failure, the useful information ("gsql: not found") doesn't reliably surface because the exec-level 127 originates from the OCI runtime/Docker layer rather than the process's own stdout/stderr stream, so the eventual `ConditionTimeoutException` message is the only diagnostic signal, and it took direct correlation with the fork CI logs to identify the root cause here. Not a new problem introduced by this PR beyond Issue 1 itself, but worth keeping in mind once Issue 1 is fixed (e.g. logging `result.getStdout()`/`getStderr()` on each attempt, or failing fast instead of retrying on `127` specifically).

## 2. Code Quality

Imports are shaded-first, non-wildcard, and consistent with the surrounding style in both files. The new `beforeStartUP()` override has a proper multi-line Javadoc explaining *why* the check exists (`JdbcOpenGaussIT.java:370-374`), which meets the "explain why, not what" bar. The HANA `withCreateContainerCmdModifier` block has a short inline comment explaining the seccomp requirement (`JdbcHanaIT.java:277`), which is adequate for a one-line rationale. No license-header issues (both are existing files). No `System.out.println`, no `@DisplayName` usage, nothing else flagged on a standards pass.

### 2.2 Test coverage & stability rating — **mandatory flaky-test analysis**

This PR's entire purpose is E2E test stability, so this is the crux of the review:

- The new probe correctly uses Awaitility (`given().ignoreExceptions().await().atMost(...).untilAsserted(...)`) rather than a raw `Thread.sleep`, which is the right pattern in principle.
- However, as shown in 1.1, the probe as written **never succeeds** against the `opengauss/opengauss:5.0.0` image via `execInContainer` — it is not merely "still a bit flaky," it is unconditionally broken, confirmed twice (JDK 8 and JDK 11) on the PR's own commit in the PR author's own fork CI.
- Net effect on stability: `JdbcOpenGaussIT` goes from "sometimes fails" (the bug being fixed) to "always fails" (this PR's actual current state). That is a regression, not a stabilization, for this specific test class.
- The HANA half is not exercised by any job in this CI run at all (see Issue 2), so it carries no verified evidence either way.

**Rating: High risk** (per house convention, this is promoted to the formal Issue 1 below as a High-severity blocker, not a documented caveat).

### 2.3 Docs

No `docs/en`/`docs/zh` updates are needed or included — correct, since there is no user-facing behavior, config, or API change here. This is a testing-infrastructure-only PR.

## 3. Architectural soundness

The changes are correctly scoped to the two test classes (`JdbcHanaIT`, `JdbcOpenGaussIT`) and use the framework's own sanctioned extension point: `AbstractJdbcIT#beforeStartUP()` is explicitly designed for this ("before startUp For example, create a user", `AbstractJdbcIT.java:330-331`), and `initContainer()` is the designated place to configure container startup. Nothing leaks into engine/core or production connector code. The architectural placement of the fix is right; the implementation of the OpenGauss probe itself is what's broken.

## 4. Issue Summary

| # | Location | Problem | Risk | Severity | Raised by another reviewer |
|---|----------|---------|------|----------|------------------------------|
| 1 | `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-jdbc-e2e/connector-jdbc-e2e-part-7/src/test/java/.../JdbcOpenGaussIT.java:376-400` | New `beforeStartUP()` gsql readiness probe returns exit code 127 ("command not found") every time via `execInContainer`, because it doesn't run through the login shell/OS-user context (`omm`) that the OpenGauss image needs to resolve `gsql`. Confirmed deterministically on both JDK 8 and JDK 11 in the PR's own fork CI run (links above). | `JdbcOpenGaussIT` now fails 100% of the time instead of intermittently — a net regression versus current `dev`, and it burns an extra ~2 minutes per run before failing. | High (blocking) | No |
| 2 | `.github/workflows/backend.yml:1120-1146` (job gating) combined with `seatunnel-e2e/.../connector-jdbc-e2e-part-6/.../JdbcHanaIT.java:272-287` | `jdbc-connectors-it-part-6`/`-part-7` only run `if: needs.changes.outputs.api == 'true' \|\| needs.changes.outputs.engine == 'true'`, which this e2e-only diff never triggers (visible as 0s/skipped in the run). I checked every `updated-modules-integration-test-part-{1..8}` job in the same fork run and found zero executions of `JdbcHanaIT` anywhere. | The HANA `seccomp=unconfined` fix — and the new `HostConfig.newHostConfig()`/`withSecurityOpts(...)` API usage, which is a first-time usage in this codebase and per the PR description was never even locally test-compiled — is completely unverified by any CI signal in this PR. | Medium | No |

## 5. Merge Recommendation

**Not recommended for merge.**

- Issue 1 — the OpenGauss readiness probe must actually pass in CI (fix the `gsql` invocation to run through the correct shell/user context, then confirm a green run on the fork before pushing) before this can be reconsidered. Right now the PR's own CI shows the change makes `JdbcOpenGaussIT` fail unconditionally.

**Recommended fixes (non-blocking, but should be resolved before/soon after merge):**
- Issue 2 — get at least one real signal (e.g. a manual `-pl connector-jdbc-e2e-part-6 -am -Pci` verify run, or a temporarily-adjusted `if:` condition for validation purposes) that `JdbcHanaIT` actually starts successfully with the new `withCreateContainerCmdModifier`/seccomp change and that the `HostConfig` API call compiles, since nothing in the current CI run exercises it.

Once Issue 1 is fixed and re-verified in CI (and ideally Issue 2 gets at least one confirmed green HANA run), I'm happy to take another pass — the overall approach (Awaitility-based readiness probe at the framework's designated extension point, plus the container-level seccomp/privileged adjustments) is architecturally the right shape for this problem.

PR #11701, opened 2026-08-08
[Docs][Connector-V2] Improve Hudi sink documentation

DanielCarter-stack:
## What is changed
- Add descriptions for all base and per-table Hudi sink options.
- Clarify flattened single-table versus multi-table configuration and CDC record-key requirements.
- Document flush behavior, schema handling, and an S3-compatible storage example.

## Verification
- `git diff --check` passed.
- `./mvnw -q -DskipTests verify` was attempted but failed in the existing `seatunnel-config-shade` compilation step.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/en/connectors/sink/Hudi.md
+++ b/docs/en/connectors/sink/Hudi.md
@@ -25,38 +25,40 @@ The SeaTunnel Hudi sink writes Hudi data files and `.hoodie` metadata, but it do
 
 Base configuration:
 
-|            name            |  type   | required | default value               |
-|----------------------------|---------|----------|-----------------------------|
-| table_dfs_path             | string  | yes      | -                           |
-| conf_files_path            | string  | no       | -                           |
-| table_list                 | Array   | no       | -                           |
-| schema_save_mode           | enum    | no       | CREATE_SCHEMA_WHEN_NOT_EXIST|
-| data_save_mode             | enum    | no       | APPEND_DATA                 |
-| common-options             | Config  | no       | -                           |
+|            name            |  type   | required | default value                | description |
+|----------------------------|---------|----------|------------------------------|-------------|
+| table_dfs_path             | string  | yes      | -                            | Root path for Hudi table data and metadata. |
+| conf_files_path            | string  | no       | -                            | Semicolon-separated local paths to HDFS client configuration files. |
+| table_list                 | Array   | no       | -                            | Per-table settings for multi-table jobs. |
+| schema_save_mode           | enum    | no       | CREATE_SCHEMA_WHEN_NOT_EXIST | How to handle the target schema before the job starts. |
+| data_save_mode             | enum    | no       | APPEND_DATA                  | How to handle existing table data before the job starts. |
+| common-options             | Config  | no       | -                            | [Common sink options](../common-options/sink-common-options.md). |
 
 Table list configuration:
 
-|       name                 |  type  | required | default value |
-|----------------------------|--------|----------|---------------|
-| table_name                 | string | yes      | -             |
-| database                   | string | no       | default       |
-| table_type                 | enum   | no       | COPY_ON_WRITE |
-| op_type                    | enum   | no       | insert        |
-| record_key_fields          | string | no       | -             |
-| partition_fields           | string | no       | -             |
-| precombine_field           | string | no       | -             |
-| batch_interval_ms          | Int    | no       | 1000          |
-| batch_size                 | Int    | no       | 1000          |
-| insert_shuffle_parallelism | Int    | no       | 2             |
-| upsert_shuffle_parallelism | Int    | no       | 2             |
-| min_commits_to_keep        | Int    | no       | 20            |
-| max_commits_to_keep        | Int    | no       | 30            |
-| index_type                 | enum   | no       | BLOOM         |
-| index_class_name           | string | no       | -             |
-| record_byte_size           | Int    | no       | 1024          |
-| cdc_enabled                | boolean| no       | false         |
+|       name                 |  type  | required | default value | description |
+|----------------------------|--------|----------|---------------|-------------|
+| table_name                 | string | yes      | -             | Target Hudi table name. |
+| database                   | string | no       | default       | Target Hudi database name. |
+| table_type                 | enum   | no       | COPY_ON_WRITE | Hudi table type: `COPY_ON_WRITE` or `MERGE_ON_READ`. |
+| op_type                    | enum   | no       | INSERT        | Write operation: `INSERT`, `UPSERT`, or `BULK_INSERT`. |
+| record_key_fields          | string | no       | -             | Field or fields used to build the Hudi record key; required for `UPSERT`. |
+| partition_fields           | string | no       | -             | Field or fields used to build the partition path. |
+| precombine_field           | string | no       | -             | Field used to resolve multiple updates to the same record. |
+| batch_interval_ms          | Int    | no       | 1000          | Maximum interval between flushes, in milliseconds. |
+| batch_size                 | Int    | no       | 1000          | Maximum buffered rows before a flush. |
+| insert_shuffle_parallelism | Int    | no       | 2             | Shuffle parallelism for insert operations. |
+| upsert_shuffle_parallelism | Int    | no       | 2             | Shuffle parallelism for upsert operations. |
+| min_commits_to_keep        | Int    | no       | 20            | Minimum number of commits retained during cleaning. |
+| max_commits_to_keep        | Int    | no       | 30            | Maximum number of commits retained during cleaning. |
+| index_type                 | enum   | no       | BLOOM         | Hudi index type: `BLOOM`, `SIMPLE`, or `GLOBAL_BLOOM`. |
+| index_class_name           | string | no       | -             | Fully qualified custom Hudi index class name. |
+| record_byte_size           | Int    | no       | 1024          | Estimated average record size in bytes. |
+| cdc_enabled                | boolean| no       | false         | Persist Hudi CDC change-log data when enabled. |
 
-Note: When this configuration corresponds to a single table, you can flatten the configuration items in table_list to the outer layer.
+Note: When this configuration corresponds to a single table, you can flatten the configuration items in `table_list` to the outer layer. For a multi-table job, keep table-specific options inside each `table_list` entry; `table_dfs_path`, `conf_files_path`, `schema_save_mode`, and `data_save_mode` remain at the sink level.
+
+For `UPSERT` or `BULK_INSERT` workloads, configure `record_key_fields` when the connector needs a stable record key. For CDC input, the upstream record must contain the fields used by `record_key_fields`; set `cdc_enabled = true` only when the Hudi CDC change log is required.
 
 ### table_name [string]
 
@@ -136,12 +138,12 @@ Note: When this configuration corresponds to a single table, you can flatten the
 
 ### schema_save_mode [Enum]
 
-Before the synchronous task is turned on, different treatment schemes are selected for the existing surface structure of the target side.  
-Option introduction：  
-`RECREATE_SCHEMA` ：Will create when the table does not exist, delete and rebuild when the table is saved        
-`CREATE_SCHEMA_WHEN_NOT_EXIST` ：Will Created when the table does not exist, skipped when the table is saved        
-`ERROR_WHEN_SCHEMA_NOT_EXIST` ：Error will be reported when the table does not exist  
-`IGNORE` ：Ignore the treatment of the table
+Controls how the connector handles the target table schema before the job starts:
+
+- `RECREATE_SCHEMA`: Create the table if it does not exist; otherwise delete and recreate it.
+- `CREATE_SCHEMA_WHEN_NOT_EXIST`: Create the table only when it does not exist.
+- `ERROR_WHEN_SCHEMA_NOT_EXIST`: Fail when the table does not exist.
+- `IGNORE`: Do not perform schema handling.
 
 ### data_save_mode [Enum]
 
@@ -259,6 +261,22 @@ sink {
     op_type = "UPSERT"
     record_key_fields = "id"
     cdc_enabled = true
+  }
+}
+```
+
+### S3 Storage
+
+The sink can write to an S3-compatible path. Provide the required Hadoop filesystem settings through `conf_files_path` (or the runtime classpath), then use an `s3a://` table path.
+
+```hocon
+sink {
+  Hudi {
+    table_dfs_path = "s3a://hudi/"
+    conf_files_path = "/etc/hadoop/core-site.xml;/etc/hadoop/hdfs-site.xml"
+    table_name = "st_test"
+    op_type = "UPSERT"
+    record_key_fields = "id"
   }
 }
 ```
--- a/docs/en/connectors/sink/Hudi.md
+++ b/docs/en/connectors/sink/Hudi.md
@@ -45,7 +45,7 @@ Table list configuration:
 | record_key_fields          | string | no       | -             | Field or fields used to build the Hudi record key; required for `UPSERT`. |
 | partition_fields           | string | no       | -             | Field or fields used to build the partition path. |
 | precombine_field           | string | no       | -             | Field used to resolve multiple updates to the same record. |
-| batch_interval_ms          | Int    | no       | 1000          | Maximum interval between flushes, in milliseconds. |
+| batch_interval_ms          | Int    | no       | 1000          | Currently unused; flushes are triggered only by `batch_size` or a checkpoint. |
 | batch_size                 | Int    | no       | 1000          | Maximum buffered rows before a flush. |
 | insert_shuffle_parallelism | Int    | no       | 2             | Shuffle parallelism for insert operations. |
 | upsert_shuffle_parallelism | Int    | no       | 2             | Shuffle parallelism for upsert operations. |
@@ -58,7 +58,7 @@ Table list configuration:
 
 Note: When this configuration corresponds to a single table, you can flatten the configuration items in `table_list` to the outer layer. For a multi-table job, keep table-specific options inside each `table_list` entry; `table_dfs_path`, `conf_files_path`, `schema_save_mode`, and `data_save_mode` remain at the sink level.
 
-For `UPSERT` or `BULK_INSERT` workloads, configure `record_key_fields` when the connector needs a stable record key. For CDC input, the upstream record must contain the fields used by `record_key_fields`; set `cdc_enabled = true` only when the Hudi CDC change log is required.
+`record_key_fields` is required for `UPSERT` (validated at startup) and for `BULK_INSERT` (currently unvalidated — omitting it causes a `NullPointerException` during writing, not a config-time error). For CDC input, the upstream record must contain the fields used by `record_key_fields`; set `cdc_enabled = true` only when the Hudi CDC change log is required.
 
 ### table_name [string]
 
@@ -267,7 +267,7 @@ sink {
 
 ### S3 Storage
 
-The sink can write to an S3-compatible path. Provide the required Hadoop filesystem settings through `conf_files_path` (or the runtime classpath), then use an `s3a://` table path.
+The sink can write to an S3-compatible path. The `connector-hudi` module does not depend on `hadoop-aws`/`aws-java-sdk`, so resolving the `s3a://` scheme requires placing `hadoop-aws` and the matching AWS SDK bundle (or SeaTunnel's `seatunnel-hadoop-aws` jar) into `$SEATUNNEL_HOME/lib` (or the connector's plugin lib directory) before the example below will work. After that, provide the required Hadoop filesystem settings through `conf_files_path` (or the runtime classpath), then use an `s3a://` table path.
 
 ```hocon
 sink {


DanielLeens (changes requested):

The current `docs/en/connectors/sink/Hudi.md` lists all base and per-table Hudi sink options in two tables that have no `description` column at all — a reader has to already know what `record_key_fields`, `precombine_field`, `index_class_name`, etc. do, because the table gives them nothing but name/type/required/default. The `schema_save_mode` section is also written in slightly broken English ("Option introduction：", "the existing surface structure"), and there is no example showing how to point `table_dfs_path` at object storage instead of HDFS.

This PR adds a `description` column to both tables, rewrites the `schema_save_mode` bullet list in clearer English, expands the single-table/multi-table flattening note with an explicit statement of which options stay at the sink level vs. the table level, adds a short paragraph on `record_key_fields`/`cdc_enabled` usage, and appends a new "S3 Storage" example section.

**One-sentence summary:** This is a genuinely useful readability upgrade for the Hudi sink doc's option tables, but three of the newly-added sentences make claims that do not match the connector's actual source behavior (a non-functional `batch_interval_ms` flush claim, an S3 example that omits a required runtime dependency, and an understated `record_key_fields` requirement for `BULK_INSERT`), so it needs one more pass before merge.

**Before/after example** (base config table, `docs/en/connectors/sink/Hudi.md`):
- Before: `| conf_files_path | string | no | - |` — a reader has no idea this is "semicolon-separated local paths to HDFS client configuration files" without reading the Java source.
- After: `| conf_files_path | string | no | - | Semicolon-separated local paths to HDFS client configuration files. |` — immediately actionable.

## 1. Code Change Review

This PR touches exactly one file, `docs/en/connectors/sink/Hudi.md` (53 insertions / 35 deletions), no production code. I traced every changed/added claim against the actual `connector-hudi` source at the PR head (`0469040807f93ec6d972857dfc0a336ef6a6c702`) rather than trusting the prose, per the task's instruction to verify documentation against real registration/options/capabilities.

### 1.1 Core Content Analysis

**Verified accurate:**
- The new "sink-level vs. table-level" note (`Hudi.md:59`, new text: *"For a multi-table job, keep table-specific options inside each `table_list` entry; `table_dfs_path`, `conf_files_path`, `schema_save_mode`, and `data_save_mode` remain at the sink level."*) is correct. `HudiSinkOptions.java` declares `TABLE_DFS_PATH`, `CONF_FILES_PATH`, `SCHEMA_SAVE_MODE`, `DATA_SAVE_MODE` as base/sink-level options, and `HudiTableConfig.java` (the per-`table_list`-entry POJO) has no `schemaSaveMode`/`dataSaveMode`/`tableDfsPath`/`confFilesPath` fields at all — only `tableName`, `database`, `tableType`, `opType`, `recordKeyFields`, `partitionFields`, `preCombineField`, `indexType`, `indexClassName`, `recordByteSize`, `batchSize`, `batchIntervalMs`, `insertShuffleParallelism`, `upsertShuffleParallelism`, `minCommitsToKeep`, `maxCommitsToKeep`, `cdcEnabled`. Good catch by the author.
- The new `op_type` default display of `INSERT` (uppercase, `Hudi.md:44`) matches `HudiSinkOptions.OP_TYPE`'s default `WriteOperationType.INSERT` and matches `docs/zh/connectors/sink/Hudi.md`'s existing `INSERT` default. I checked whether this creates a real case-sensitivity problem with the still-untouched `### op_type [enum]` prose a few lines below (`Hudi.md:109`, unchanged by this PR) which still says lowercase "insert"/"upsert"/"bulk_insert" — it does not: `ConfigUtil.convertToEnum()` (`seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConfigUtil.java:286-300`) compares `e.toString().toUpperCase(...)` against the input, so both cases parse identically at runtime. Not a bug, just cosmetic inconsistency inherited from before this PR — not this PR's responsibility to fix, but calling it out since I checked it.
- `index_type` values `BLOOM`/`SIMPLE`/`GLOBAL_BLOOM` (`Hudi.md:54`) match `HoodieIndex.IndexType` in `hudi-client-common-0.15.0` (which also has `HBASE`/`INMEMORY` that the connector's own option description already excludes as "currently supported" — pre-existing scope limit, not introduced here).
- `table_type` values `COPY_ON_WRITE`/`MERGE_ON_READ` match `HoodieTableType`.

**Issues found (see Section 4 for full table, details below):**

1. **`batch_interval_ms` description is factually wrong** (`Hudi.md:48`, new text: *"Maximum interval between flushes, in milliseconds."*). I traced the entire flush call chain: `HudiSinkWriter.write()` → `HudiRecordWriter.writeRecord()` (`seatunnel-connectors-v2/connector-hudi/src/main/java/.../sink/writer/HudiRecordWriter.java:113-128`) only calls `flush()` when `batchCount >= hudiTableConfig.getBatchSize()`. There is no `ScheduledExecutorService`, `Timer`, or any other time-based trigger anywhere in `HudiRecordWriter.java` or `HudiSinkWriter.java` — I grepped the whole `connector-hudi` module for `IntervalMs`/`Scheduled`/`Timer` and the only hits are the option declaration, the config POJO field, and the default-value fallback in `HudiTableConfig.of()` (`config/HudiTableConfig.java:192-193`). The value is parsed and stored but **never read again** by the write/flush path. The only other flush trigger is `HudiSinkWriter.prepareCommit()` (checkpoint boundary), which is unrelated to the configured interval. So a job with `batch_size = 10000` and low throughput will simply never flush until 10,000 rows accumulate (or a checkpoint fires) — `batch_interval_ms` has zero effect on that. This same wrong claim already exists, unmodified, at the pre-existing `### batch_interval_ms [Int]` section (`Hudi.md:111-113`, not touched by this diff), so this PR is not the origin of the misconception — but the PR's stated goal is specifically to "document flush behavior," and it chose to restate the same incorrect claim in a new, more prominent quick-reference table row rather than correcting it. Given that this option currently has no runtime effect, the description should say so (e.g. "Currently unused; flushes are triggered only by `batch_size` or a checkpoint.") instead of asserting a time-based flush guarantee that doesn't exist.

2. **New "S3 Storage" example omits a required runtime dependency** (`Hudi.md:268-282`). The example sets `table_dfs_path = "s3a://hudi/"` and says only *"Provide the required Hadoop filesystem settings through `conf_files_path` (or the runtime classpath)."* I checked whether `connector-hudi` can actually resolve the `s3a://` scheme: `HudiUtil.getConfiguration()` (`seatunnel-connectors-v2/connector-hudi/src/main/java/.../util/HudiUtil.java:64-70`) only does `configuration.addResource(new Path(file))` — it loads XML *property values* into a Hadoop `Configuration`, it does not add anything to the classpath. Resolving `s3a://` requires the `S3AFileSystem` class, which ships in the `hadoop-aws` (+ AWS SDK) jars. `connector-hudi/pom.xml` declares a `<hadoop-aws.version>3.1.4</hadoop-aws.version>` property but **never actually uses it as a dependency** — I listed every `<artifactId>` in that pom and there is no `hadoop-aws`/`seatunnel-hadoop-aws`/AWS SDK dependency anywhere, and the module's `maven-shade-plugin` config only relocates `org.apache.avro`, nothing S3-related. Compare this with `connector-file-s3/pom.xml:54-56`, which explicitly depends on `org.apache.seatunnel:seatunnel-hadoop-aws` (provided scope) specifically to make `s3a://` work. `connector-hudi` has no equivalent dependency, formal or otherwise. A user who copies this new example verbatim will very likely hit `UnsupportedFileSystemException`/`ClassNotFoundException: org.apache.hadoop.fs.s3a.S3AFileSystem` because the S3A implementation simply isn't on the shipped connector's classpath — the doc should say the user must separately place `hadoop-aws`/`aws-java-sdk-bundle` (or SeaTunnel's own `seatunnel-hadoop-aws` jar) into `$SEATUNNEL_HOME/lib` (or the connector's plugin lib) before this example will work.

3. **`record_key_fields` requirement is understated for `BULK_INSERT`** (`Hudi.md:61`, new text: *"For `UPSERT` or `BULK_INSERT` workloads, configure `record_key_fields` when the connector needs a stable record key."*). This phrasing reads as "recommended, use your judgement," but the actual code makes it effectively mandatory for anything other than plain `INSERT`. `HudiRecordConverter.getHoodieKey()` (`seatunnel-connectors-v2/connector-hudi/src/main/java/.../sink/convert/HudiRecordConverter.java:79-83`):
   ```java
   String rowKey =
       hudiTableConfig.getRecordKeyFields() == null
                       && hudiTableConfig.getOpType().equals(WriteOperationType.INSERT)
               ? UUID.randomUUID().toString()
               : getRecordKey(element, seaTunnelRowType, hudiTableConfig);
   ```
   Only `op_type == INSERT` gets the `null`-safe UUID fallback. For `BULK_INSERT` (or any other op type) with `record_key_fields` unset, this falls into `getRecordKey()`, whose first line is `hudiTableConfig.getRecordKeyFields().split(",")` (`HudiRecordConverter.java:93`) — a `NullPointerException` on the very first record written. Worse, `HudiTableConfig.of()`'s upfront validation (`config/HudiTableConfig.java:209-214`) only throws a clean `IllegalArgumentException` when `op_type == WriteOperationType.UPSERT`; it does not check `BULK_INSERT`, so this isn't caught at config-validation time — the job starts, then dies mid-write with a raw NPE instead of the clear message UPSERT users get. The doc's own table row for `record_key_fields` (`Hudi.md:45`) correctly says "required for `UPSERT`" only — so the new prose paragraph at line 61 actually contradicts the table two lines above it by implying `BULK_INSERT` is optional-but-recommended when it is, in the current code, load-bearing (crash-if-missing). Recommend changing "when the connector needs a stable record key" to something like "record_key_fields is required for UPSERT (validated at startup) and for BULK_INSERT (unvalidated — omitting it currently causes a NullPointerException during writing, not a config-time error)."

### 1.2 Compatibility

Fully compatible. This is a documentation-only change to a single Markdown file; no `Option`, default value, API, or SPI registration is touched. Zero runtime risk.

None. No code paths, no CPU/memory/GC/network/locking/concurrency/retry/idempotency implications — pure prose and table changes.

N/A — no code changed. (See Issue 3 above for a pre-existing runtime NPE this PR's new text should have flagged instead of glossing over.)

## 2. Code Quality

### 2.1 Coding Standards

N/A (no Java changes, no license headers needed for a `.md` edit). Markdown table alignment in the diff is consistent and renders correctly; I don't see any broken pipe/column-count mismatches when I re-rendered the two tables mentally against the diff hunks.

Not applicable — no test files are touched by this PR, so the mandatory flaky-test/stability analysis from the review checklist does not apply here. **Stability rating: N/A (no test code changed).**

### 2.3 Docs

This is the crux of the PR, covered in detail in Section 1.1. One additional gap: **`docs/zh/connectors/sink/Hudi.md` exists** (I confirmed it with `git ls-tree`) but is **not updated** by this PR — it still has the old two-column-less tables and no S3 example, so English and Chinese docs are now materially out of sync (the zh doc's `op_type` default was already correctly `INSERT` uppercase, for what it's worth, so the zh doc didn't have the case issue the old en doc had). This repo's own contribution guide states user-visible doc changes should update both `docs/en` and `docs/zh`; I'm not certain this repo enforces that strictly for pure prose-clarity PRs in practice, so I'm listing it as a recommended (non-blocking) fix rather than a hard blocker — but it's worth a follow-up PR if not this one.

## 3. Architectural Soundness

Not applicable — no architecture, registration, or SPI surface is touched. This is a self-contained doc edit to one connector's sink page.

## 4. Issue Summary

| 1 | `docs/en/connectors/sink/Hudi.md:48` | New table row claims `batch_interval_ms` is "Maximum interval between flushes, in milliseconds," but `HudiRecordWriter.writeRecord()` (`sink/writer/HudiRecordWriter.java:113-128`) only flushes on `batchCount >= batch_size` or on checkpoint (`prepareCommit`); the interval value is parsed/stored but never read by the flush path anywhere in the module. | Medium | No |
| 2 | `docs/en/connectors/sink/Hudi.md:268-282` | New "S3 Storage" example uses `table_dfs_path = "s3a://hudi/"` without noting that `connector-hudi` has no `hadoop-aws`/AWS SDK dependency (verified: only an unused `<hadoop-aws.version>` pom property, unlike `connector-file-s3`'s explicit `seatunnel-hadoop-aws` dependency) — the example will fail with a missing-filesystem-scheme error unless the user manually adds those jars, which the doc doesn't mention. | Medium | No |
| 3 | `docs/en/connectors/sink/Hudi.md:61` | New sentence says to configure `record_key_fields` for `BULK_INSERT` "when the connector needs a stable record key" (sounds optional), but `HudiRecordConverter.getHoodieKey()` (`sink/convert/HudiRecordConverter.java:79-83`) only null-safes `op_type == INSERT`; `BULK_INSERT` with no `record_key_fields` hits an unguarded NPE in `getRecordKey()` (line 93) because `HudiTableConfig.of()`'s validation (`config/HudiTableConfig.java:209-214`) only checks `op_type == UPSERT`. This contradicts the table row two lines above it, which correctly limits "required" to UPSERT only. | Medium | No |
| 4 | `docs/zh/connectors/sink/Hudi.md` | Not updated alongside the English doc improvements in this PR; en/zh are now materially out of sync (missing descriptions, S3 example, updated notes). | Low | No |

## 5. Merge Recommendation

**Blockers (must-fix, by issue number):** #1, #2, #3 — all three are concrete, source-verified mismatches between the new prose and the connector's actual behavior. Given the entire purpose of this PR is to make the Hudi sink docs more accurate and usable, shipping new text that overstates `batch_interval_ms`'s effect, under-warns about a real NPE risk on `BULK_INSERT`, and hands users a copy-pasteable example that will fail without an unmentioned extra jar would undermine that goal. All three are small, targeted wording fixes — no architectural rework needed, should take the author a few minutes each.

**Recommended fixes (non-blocking):** #4 (bring `docs/zh/connectors/sink/Hudi.md` into parity, in this PR or a fast follow-up).

CI is green: the fork's "Build" workflow run (https://github.com/apache/seatunnel/runs/93056387750) passed in 1h6m, along with "Notify test workflow" and "labeler." `mergeable: true` with `mergeable_state: blocked` (branch protection awaiting review, not a real conflict) and `ahead_by: 1 / behind_by: 3` against `dev` (minor, non-blocking divergence). No existing reviews or comments on this PR from any reviewer at the time of this review — this is the first review it has received.

Thanks for taking on the Hudi sink docs, @DanielCarter-stack — the description column and the clearer `schema_save_mode` wording are a solid improvement over the previous tables, and I can tell you went and traced through what each option does rather than just reformatting. The three notes above just need the wording tightened to match what the code actually does today (or, longer-term, the `batch_interval_ms`/S3-classpath gaps could be fixed in the connector itself rather than only documented around — but that's out of scope for a docs PR).

DanielLeens (approved):

The Hudi sink doc's base and per-table option tables previously had no `description` column at all, so a reader had to already know what `record_key_fields`, `precombine_field`, `index_class_name`, etc. do. This PR adds descriptions to both tables, clarifies the sink-level-vs-table-level flattening note, expands the `record_key_fields`/`cdc_enabled` guidance, and adds a new "S3 Storage" example.

**This is a re-review.** Since my previous CHANGES_REQUESTED pass, the author pushed one follow-up commit (`bf6ff55`) that addresses all three blocking issues from that round. Thanks for the quick turnaround, @DanielCarter-stack — all three fixes are precise, source-accurate rewrites rather than vague softening, which is exactly what I asked for.

## 1. Code Change Review

### 1.1 Core Content Analysis

This is a documentation-accuracy review — the PR (fix commit included) touches only `docs/en/connectors/sink/Hudi.md`, no production code. There is no "recovery path vs. main path" distinction to make; the entire risk surface is "does the text match the connector's actual behavior," which is what all verification below targets, against the real `connector-hudi` source at the current head (`bf6ff55ed7e95a199457d9e48ed59006710426b2`).

**Status of the previous round's findings:**

| # | Previous finding | Status at current head |
|---|---|---|
| 1 (Medium, blocker) | `batch_interval_ms` described as "Maximum interval between flushes," but the flush path never reads this value | **Fixed.** New text: "Currently unused; flushes are triggered only by `batch_size` or a checkpoint." Re-verified against `HudiRecordWriter.writeRecord()` (`sink/writer/HudiRecordWriter.java:113-128`) at the current head — flush is still gated only on `batchCount >= hudiTableConfig.getBatchSize()`; no scheduled/timer-based trigger exists anywhere in the module. Accurate. |
| 2 (Medium, blocker) | New "S3 Storage" example uses `table_dfs_path = "s3a://hudi/"` without noting `connector-hudi` has no `hadoop-aws`/AWS SDK dependency | **Fixed.** New text explicitly states the module "does not depend on `hadoop-aws`/`aws-java-sdk`" and instructs the user to place `hadoop-aws` + AWS SDK bundle (or `seatunnel-hadoop-aws`) into `$SEATUNNEL_HOME/lib` first. Re-verified `connector-hudi/pom.xml` at the current head — still only an unused `<hadoop-aws.version>3.1.4</hadoop-aws.version>` property, no actual `hadoop-aws` dependency declaration. Accurate. |
| 3 (Medium, blocker) | `record_key_fields` for `BULK_INSERT` described as "when the connector needs a stable record key" (sounds optional), but omitting it causes an unguarded NPE | **Fixed.** New text: "required for `UPSERT` (validated at startup) and for `BULK_INSERT` (currently unvalidated — omitting it causes a `NullPointerException` during writing, not a config-time error)." Re-verified `HudiRecordConverter.getHoodieKey()` (`sink/convert/HudiRecordConverter.java:79-83`, still only null-safes `op_type == INSERT`) and `HudiTableConfig.of()`'s validation (`config/HudiTableConfig.java`, still only checks `op_type == UPSERT`) at the current head. Accurate and appropriately blunt about the NPE risk. |
| 4 (Low, non-blocking) | `docs/zh/connectors/sink/Hudi.md` not updated alongside the English improvements | **Not addressed.** The fix commit, like the original commit, touches only the English file. See Issue 4 below (unchanged from last round, still non-blocking). |

All three blockers are cleanly and accurately resolved. Only the previously-non-blocking zh-parity gap remains.

**Full option-table audit performed this round** (not just the three previously-flagged rows), against `HudiSinkOptions.java` and `HudiSinkFactory.optionRule()` at the current head:

| Option | Doc says | Source says | Result |
|---|---|---|---|
| `table_dfs_path` | required, no default | Only sink-level option in `HudiSinkFactory.optionRule().required(...)`; `HudiSinkOptions.TABLE_DFS_PATH` has `noDefaultValue()` | Pass |
| `conf_files_path`, `table_list`, `schema_save_mode` (`CREATE_SCHEMA_WHEN_NOT_EXIST`), `data_save_mode` (`APPEND_DATA`) | optional, defaults as shown | All match `HudiSinkOptions.java` exactly (`SCHEMA_SAVE_MODE` / `DATA_SAVE_MODE` defaults) | Pass |
| `table_name` | required | `Options.key("table_name").stringType().noDefaultValue()`, sole required field of `HudiTableConfig` | Pass |
| `database` (default `default`), `table_type` (`COPY_ON_WRITE`), `cdc_enabled` (`false`), `index_type` (`BLOOM`), `record_byte_size` (`1024`), `op_type` (`INSERT`), `batch_size` (`1000`), `batch_interval_ms` (`1000`), `insert_shuffle_parallelism` (`2`), `upsert_shuffle_parallelism` (`2`), `min_commits_to_keep` (`20`), `max_commits_to_keep` (`30`) | all optional with the listed defaults | Every default cross-checked byte-for-byte against `HudiSinkOptions.java` | Pass |
| `record_key_fields`, `partition_fields`, `precombine_field`, `index_class_name` | optional, no default | `noDefaultValue()` for all four in `HudiSinkOptions.java` | Pass |

Every row in both tables checks out against the real `Option` declarations — no additional inaccuracies found beyond the three (now-fixed) items from the previous round.

### 1.2 Compatibility

Fully compatible. Documentation-only change to a single Markdown file; no `Option`, default value, API, or SPI registration is touched. Zero runtime risk.

None. No code paths touched.

N/A — no code changed. (The doc now correctly *describes* a pre-existing NPE risk on `BULK_INSERT` without `record_key_fields`, rather than glossing over it — a genuine improvement in accuracy, though the underlying NPE itself remains unfixed in the connector, which is out of scope for a docs PR.)

## 2. Code Quality

### 2.1 Coding Standards

The fix commit is a clean, three-hunk change touching exactly the three flagged lines/paragraphs — no unrelated reformatting, no scope creep. Table alignment and Markdown structure remain consistent with the rest of the file.

Not applicable — no test files are touched. Stability rating: N/A (no test code changed).

### 2.3 Docs

The three fixes read naturally as part of the surrounding prose and don't just append disclaimers — e.g., the `record_key_fields` sentence now directly states the concrete failure mode (NPE, unvalidated) instead of hedging with "when needed." This is the right level of specificity for a doc whose whole purpose is accuracy.

**Remaining gap (unchanged from last round):** `docs/zh/connectors/sink/Hudi.md` still has the old two-column-less tables, no S3 example, and none of this PR's clarifications — en/zh are now more out of sync than before, since the English doc gained real content (description column, expanded notes, new example section) that the Chinese doc lacks entirely. I'm keeping this as a non-blocking recommendation rather than a hard blocker, consistent with my previous review, since the project's docs contribution guide isn't strictly enforced for prose-clarity-only PRs in practice — but it's worth a fast follow-up.

## 3. Architectural Soundness

Not applicable — no architecture, registration, or SPI surface is touched. Self-contained doc edit to one connector's sink page.

## 4. Issue Summary

|---|---|---|---|
| 1 | ~~`batch_interval_ms` claimed to control flush timing~~ — **Fixed in `bf6ff55`**, verified against `HudiRecordWriter.writeRecord()` | `docs/en/connectors/sink/Hudi.md:48` | Resolved |
| 2 | ~~S3 example omitted the missing `hadoop-aws` dependency~~ — **Fixed**, verified against `connector-hudi/pom.xml` | `docs/en/connectors/sink/Hudi.md:270` | Resolved |
| 3 | ~~`record_key_fields` requirement understated for `BULK_INSERT`~~ — **Fixed**, verified against `HudiRecordConverter.getHoodieKey()` and `HudiTableConfig.of()` | `docs/en/connectors/sink/Hudi.md:61` | Resolved |
| 4 | `docs/zh/connectors/sink/Hudi.md` not updated alongside the English doc improvements; en/zh now materially out of sync | `docs/zh/connectors/sink/Hudi.md` | Low |

---

### Issue 4 — Chinese doc still not updated (carried over, unresolved)

**Severity:** Low, non-blocking.
**Location:** `docs/zh/connectors/sink/Hudi.md`.
**Problem:** Same as last round. The zh doc still lacks the description column, the clarified sink-level/table-level note, the sharpened `record_key_fields` guidance, and the new S3 example — all of which the en doc now has after two rounds of improvement. The gap between the two languages has grown, not shrunk, over the course of this PR.
**Potential risk:** Low — Chinese-reading users lose the accuracy improvements delivered to English readers, but nothing in the zh doc is factually wrong, just less complete.
**Best improvement:** Bring `docs/zh/connectors/sink/Hudi.md` into parity in this PR or a fast, immediate follow-up, given the English doc has now changed substantially across two commits.
**Raised by another reviewer:** No.

## 5. Merge Recommendation

**Blockers:** None. All three blocking issues from the previous round are fixed and independently re-verified against the current `connector-hudi` source — the `batch_interval_ms` claim, the S3/`hadoop-aws` gap, and the `BULK_INSERT` NPE risk are all now accurately described.

- Issue 4 — bring `docs/zh/connectors/sink/Hudi.md` up to parity with the now-substantially-improved English doc, ideally as a fast follow-up given how much the English content has changed across this PR's two commits.

**CI status:** all checks green (`Build`, `Notify test workflow`, `labeler`).

Overall assessment: this PR turns a bare, description-less options table into an accurate, usable reference, and — unlike a purely cosmetic patch — the author's fixes for the previous round's three findings all trace precisely to the real code paths I cited (flush trigger, missing S3 dependency, unguarded NPE), rather than just rephrasing around the problem. Good to merge; the zh-parity gap is worth a quick follow-up but shouldn't hold up this improvement to the English doc.

davidzollo (approved):
LGTM

PR #11702, opened 2026-08-08
[Feature][Zeta] Prepare job log mode context and exact filename matching

corgy-w:

This PR backports Phase A infrastructure changes from the internal ST-3605 work to the open-source Zeta engine so that the engine-server side is ready for future retention policy work without depending on whaletunnel-private classes.

Concretely:

- `MDCContext` gains an optional `JobMode` dimension. The legacy 3-segment wire format (`jobId/pipelineId/taskId`) stays compatible; the new 4-segment format (`jobId/pipelineId/taskId/mode") carries the job mode. When a nested context activates with a null mode, the previous `ST-JOB-MODE` MDC key is explicitly removed so cleanup policy state cannot leak across activations.
- `CoordinatorService` now builds job-scoped MDC executors via a new helper that pulls the job mode out of `JobImmutableInformation` (fail closed to unclassified when the payload is incomplete or corrupted) and threads it through every job executor entry point: `submitJob`, the master-node-switch restore path, the pending-job schedule loop, and the per-job `restoreJobFromMasterActiveSwitch` path.
- New `JobLogFileNameMatcher` performs exact matching of `job-<id>.log`, `job-<id>.log.unclassified`, and `job-<id>.log.yyyy-MM-dd-index` filenames. Adjacent job ids (e.g. `job-1234.log` for `jobId=123`) and unsupported suffixes (`.tmp`, bare date) are now rejected, fixing the historical prefix-substring download/cleanup leak. A separate `isRolledSegment` helper exists for the pruner so it can never touch active or unclassified files.
- `TaskLogManagerService` is split into two lifecycles:
  - `clean(jobId)` keeps terminal removal. It is still invoked from `JobHistoryService.JobInfoExpiredListener` and now uses the exact matcher so adjacent ids survive cleanup.
  - `pruneRolledJobLogSegments(ttl)` plus the new `startPrune(intervalMinutes, ttl)` API form a running-state remover that only deletes rolled segments whose last-modified age exceeds the configured retention window. Active files and unclassified sidecars are never touched here.
- A dedicated daemon `ScheduledExecutorService` (named `task-log-prune`) is wired into `SeaTunnelServer.initInternal`. It is driven by the new `TelemetryLogsConfig.jobLogPruneIntervalMinutes` setting (config key `job-log-prune-interval-minutes`, default 60 minutes) and uses a default TTL of 7 days to mirror the Log4j2 `file_ttl` policy. `SeaTunnelServer.shutdown(boolean)` calls `TaskLogManagerService.shutdown()` so the pruner stops with the rest of the engine services.
- The Log4j2 job appender is promoted from `File` to `RollingFile` with both `TimeBasedTriggeringPolicy` and `SizeBasedTriggeringPolicy` so long-running streaming jobs emit `.yyyy-MM-dd-N` segments that the pruner can shed independently. The `fileAppender` system route is untouched.
- New `ServerConfigOptions.JOB_LOG_PRUNE_INTERVAL_MINUTES` and `YamlSeaTunnelDomConfigProcessor` parsing so deployments can actually tune the prune cadence instead of inheriting the hard-coded default.
- Documentation for the new option is added in both `docs/en/engines/zeta/logging.md` and `docs/zh/engines/zeta/logging.md` under the existing "Scheduled deletion of old logs" section, including the default value, the disable behavior, and the strict scope (rolled segments only — active and unclassified files are never touched).

### Scope explicitly excluded from this PR

The original internal PR also touched whaletunnel-private classes and the log-download/extract protocol. Those pieces are intentionally **not** included here because they do not exist on `apache/seatunnel` `dev`:

- `LogoutService` and `LogServiceTest` (whaletunnel-private, used for log packaging and download).
- `JobClient.extractLogContentFromZip` and `JobClientLogStatusTest` (rolled-zip stitching for the download API).
- `log4j2_pod.properties` and `log4j2-test.properties` (whaletunnel-private deployment/test resources).

When whaletunnel is ready to upstream the full set, those changes can land as a follow-up PR after the web reader protocol is upgraded to handle multi-file / archived segments.

Yes — but only as additive, opt-in behavior with full backward compatibility.

- A new YAML key `job-log-prune-interval-minutes` (default `60`) under `seatunnel.engine.config.telemetry.logs` controls the new local job-log pruner. Setting it to a non-positive value disables pruning and preserves the current behavior. No existing keys are renamed or removed. Documented in `docs/en/engines/zeta/logging.md` and `docs/zh/engines/zeta/logging.md`.
- The Log4j2 job appender now rolls segments by date and size. The active file name, the directory layout, and the routing key (`ST-JID`) are unchanged. Existing log consumers and the existing terminal `clean(jobId)` behavior continue to work as before; rolled segments are simply an additional output.
- The legacy 3-segment `MDCContext` serialization is still parsed and produced when callers do not opt into job mode, so existing log analysis pipelines and persisted context strings remain valid.
- A missing or unknown job mode fails closed to `unclassified` routing; the engine never guesses `STREAMING`. The PR does not modify the database schema, connector / datasource registration, REST API contracts, or `seatunnel-engine-ui`.

If a downstream deployer has a custom `log4j2*.properties` that copies `appender.routing.route.job.appender.type = File`, they will now see `RollingFile` instead once they sync to this commit. Migrating custom configs is a one-line change to `RollingFile`.

No entry is needed in `docs/en/introduction/concepts/incompatible-changes.md` because no public contract, default value, or configuration key is removed or renamed.

Validated on top of `apache/seatunnel` `dev` HEAD `00ca1bc02d`:

```bash
./mvnw -q -pl seatunnel-api,seatunnel-engine/seatunnel-engine-common,seatunnel-engine/seatunnel-engine-server -am spotless:apply
```

```bash
./mvnw -q -pl seatunnel-api,seatunnel-engine/seatunnel-engine-common,seatunnel-engine/seatunnel-engine-server -am   -DskipTests -Dmaven.gitcommitid.skip=true install
```

```bash
./mvnw -q -pl seatunnel-api   -Dtest=MDCTracerTest -Dmaven.gitcommitid.skip=true test
```

```bash
./mvnw -q -pl seatunnel-engine/seatunnel-engine-server   -Dtest=JobLogFileNameMatcherTest,TaskLogManagerServiceTest,CoordinatorServiceJobLogModeTest   -Dmaven.gitcommitid.skip=true test
```

```bash
./mvnw -q -pl seatunnel-api -Dmaven.gitcommitid.skip=true test
```

All commands above finish with exit code 0. The unit tests added by this PR exercise:

- Round-trip serialization of the legacy 3-segment format and the new 4-segment mode-aware format in `MDCContext`.
- Nested MDC activation: an outer mode-aware context activates `ST-JOB-MODE`, an inner context without mode explicitly clears it, and on deactivate the outer mode is restored.
- Exact filename matching for active, unclassified sidecar, and date-index rolled files; rejection of adjacent ids, `.tmp` suffixes, and bare date suffixes; `isRolledSegment` rejecting active and unclassified.
- `TaskLogManagerService.clean` deletes only files that exactly belong to the supplied job id and survives missing log directories and non-positive job ids.
- `CoordinatorService.extractJobMode` returns the configured mode for a complete payload and falls back to null when job config or job context is missing.

### Check list

* [x] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md) — N/A, no new dependencies introduced.
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs — `job-log-prune-interval-minutes` is documented in both `docs/en/engines/zeta/logging.md` and `docs/zh/engines/zeta/logging.md`.
* [x] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR. — N/A, fully backward compatible.
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) — N/A, no connector.
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml) — N/A.
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml) — N/A.
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/) — N/A, this PR is engine-internal.
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config) — N/A.


DanielLeens (changes requested):

Today `TaskLogManagerService` (introduced to isolate per-job telemetry logs) only knows how to delete a job's logs once, at terminal cleanup, and it matched file names with a naive `name.contains(String.valueOf(jobId))` check — so `job-123.log` and `job-1234.log` could collide, and there was no way to shed rolled log segments for a *still-running* job (streaming jobs can run for weeks, and the per-job appender was a plain `File`, so it grew forever). This PR lays the groundwork ("Phase A infrastructure" per the description) for a future retention/classification feature by:

- Making `MDCContext` carry an optional `JobMode` alongside `jobId/pipelineId/taskId`, with the legacy 3-segment wire string still round-tripping unchanged.
- Replacing the substring job-log matcher with an exact `JobLogFileNameMatcher` (active file / `.unclassified` sidecar / `job-<id>.log.yyyy-MM-dd-N` rolled segment).
- Promoting the per-job Log4j2 appender from `File` to `RollingFile` so long streaming jobs actually produce rolled segments.
- Adding a new background pruner (`TaskLogManagerService.startPrune`) that periodically deletes rolled segments older than a TTL, independent of the existing terminal `clean(jobId)` path.

**Before**: `job-1.log` would leak `job-1` cleanup into `job-10.log`/`job-100.log` etc. (substring match), and a job's log file for a long streaming job just grew unbounded until the job finished.
**After**: only exact `job-<id>.log[.unclassified|.yyyy-MM-dd-N]` names match, and rolled segments older than 7 days get pruned every hour while the job is still running, without touching the active file.

One-sentence summary: this PR fixes a real filename-collision bug in job-log cleanup and adds a new background pruner for rolled log segments, but the pruner's file matcher is not scoped to job files and will also delete the engine's own rotated system log (`seatunnel.log.yyyy-MM-dd-N`) once the (default-enabled) feature has been running for 7 days.

## 1. Code Change Review

### 1.1 Core Logic Analysis

**`MDCContext` wire-format compatibility (`seatunnel-api/.../tracing/MDCContext.java`)**

Before:
```java
public MDCContext(Long jobId, Long pipelineId, Long taskId) { ... }
// toString(): "%d/%d/%d"
// valueOf(): if pipelineId==0 || taskId==0 -> MDCContext.of(jobId) (drops BOTH fields)
```
After: a 4th `JobMode jobMode` field is added, `toString()` appends `/<MODE>` only when a mode is present, and `valueOf()` parses `arr[3]` only `if (arr.length >= 4)`.

This context is not just an in-process object — `TracingOperation` (`seatunnel-engine-server/.../task/operation/TracingOperation.java:47-56`) serializes it with `out.writeString(context.toString())` and deserializes with `context = MDCContext.valueOf(in.readString())` across Hazelcast RPC boundaries (it's the base class for `TaskAcknowledgeOperation`, `TaskReportStatusOperation`, `RequestSlotOperation`, `AssignSplitOperation`, etc. — i.e. real checkpoint-ack / task-status / resource-request traffic between master and workers). I traced both rolling-upgrade directions:
- Old node -> new node: old node emits a 3-segment string; new `valueOf` only reads `arr[3]` `if (arr.length >= 4)`, so it safely falls back to `jobMode = null`. OK.
- New node -> old node: new node emits a 4-segment string; old `valueOf` (pre-PR) only ever reads `arr[0..2]` and ignores anything past `arr[2]`. OK.

So the actual wire compatibility claim in the PR body is correct and verified against the real caller, not just asserted.

One incidental behavior change I want to flag (not necessarily wrong, but undisclosed): the old `valueOf` nulled out **both** `pipelineId` and `taskId` whenever *either* one was the `0` sentinel (`if (pipelineId == 0 || taskId == 0) return MDCContext.of(jobId)`), while the new `parseNullableId` treats each field independently. In principle this changes parsing for a job+pipeline-only context (`MDCContext.of(jobId, pipelineId)`, `taskId==null`) — old code would have collapsed it to job-only, new code preserves the pipeline id. I could not find any current caller that actually constructs a job+pipeline-only `MDCContext` that gets serialized (the 2-arg `of(jobId, pipelineId)` has no callers outside `MDCContext`/`MDCTracer` themselves), so this looks latent rather than exploitable today, but it's an actual behavior delta worth a one-line callout in the PR description since it touches a `valueOf` used on the wire.

**`CoordinatorService` job-mode threading (`seatunnel-engine-server/.../CoordinatorService.java`)**

`createJobMdcExecutorService` is correctly wired into all four job-executor entry points the PR description claims (`submitJob:1274`, `pendingJobSchedule:380`, `restoreAllRunningJobFromMasterNodeSwitch:989`, `restoreJobFromMasterActiveSwitch:1051`), and `extractJobMode` fails closed (returns `null` on any missing/corrupt field, `CoordinatorService.java:2292-2338`), which is exactly the "never guess STREAMING" invariant the PR claims and is directly tested by `CoordinatorServiceJobLogModeTest` (both tests pass in CI, see 2.2).

However, the `submitJob` path (`CoordinatorService.java:1258-1284`) — the **normal, hot path hit by every single job submission**, not a recovery path — now deserializes the client-supplied `Data jobImmutableInformation` blob **twice**:
1. `createJobMdcExecutorService(jobId, jobImmutableInformation)` at line 1274 → `extractJobMode(jobId, Data)` (line 2292) → `nodeEngine.getSerializationService().toObject(jobImmutableInformation)`.
2. `deserializeJobImmutableInformation(jobImmutableInformation)` at line 1281 (inside the submitted task) → the exact same `toObject(...)` call.

`submitJob` is invoked synchronously from `SubmitJobOperation.doRun()` (`seatunnel-engine-server/.../operation/SubmitJobOperation.java:65-71`), a Hazelcast `Operation`, so step 1 above runs **on the Hazelcast operation thread**, before the async executor submission even happens — i.e. this is new, avoidable CPU/allocation work added to a shared, limited operation-thread pool on every job submit, scaling with the size of the job's DAG/config. The fix is straightforward: deserialize once (either reorder so the mode is derived from `submittedJobImmutableInformation` after the first deserialize inside the task, or have the two-arg `createJobMdcExecutorService` overload accept an already-parsed `JobImmutableInformation` and do the parse only once outside).

The same double-deserialization pattern also exists on `restoreAllRunningJobFromMasterNodeSwitch` → `restoreJobFromMasterActiveSwitch` (`CoordinatorService.java:989` and `:1051` both call `createJobMdcExecutorService` against the same `JobInfo.getJobImmutableInformation()` `Data` for jobs restored via the async branch) — this one is a recovery-only path (master failover), so it's lower severity, but it doubles deserialization cost for every running job exactly at the moment a failover is already stressing the cluster.

**`JobLogFileNameMatcher` / pruner scope (`seatunnel-engine-server/.../telemetry/log/JobLogFileNameMatcher.java`, `TaskLogManagerService.java`) — the main finding, see Issue 1 below.**

`isJobLogFile(name, jobId)` (used by `clean(jobId)`) is correctly scoped: it anchors on `activeFileName(jobId) = "job-" + jobId + ".log"` and only accepts that exact prefix plus `.unclassified` or a full `\.\d{4}-\d{2}-\d{2}-\d+` suffix — verified against `job-1234.log` (adjacent id) and `job-123.log.tmp` correctly being rejected in `JobLogFileNameMatcherTest`.

`isRolledSegment(name)` (used by the new `pruneRolledJobLogSegments`) is a *different*, weaker check: it just looks at the text after the **last dot** in the file name and matches it against `\.\d{4}-\d{2}-\d{2}-\d+`, with **no "job-" prefix requirement at all** (`JobLogFileNameMatcher.java:88-103`). `pruneRolledJobLogSegments` (`TaskLogManagerService.java:173-207`) calls `logDir.listFiles()` on the **entire** log directory (`path`, resolved by `LogUtil.getLogPath()`) and deletes anything `isRolledSegment` accepts once it's older than the TTL — there is no additional job-id/prefix filter anywhere in that loop. See Issue 1 for why this is a real, not theoretical, bug given the shared `config/log4j2.properties`.

**`TaskLogManagerService` lifecycle (`SeaTunnelServer.java:189-199`, `:264-265`)**

`initClean()` is called synchronously before `startPrune()` on the same thread during `SeaTunnelServer.init()`, so there's no visibility race on the `path` field despite it not being `volatile` (java.util.concurrent's Executor submission establishes the happens-before edge to the pruner's worker thread). `shutdown()` is correctly wired into `SeaTunnelServer.shutdown(boolean)` and is null-guarded and idempotent (`pruneExecutor.shutdownNow()` + null-out). This part of the lifecycle wiring is solid.

### 1.2 Compatibility

- **API/SPI**: No public interface signatures removed or changed shape (only additive overloads on `MDCContext`/`MDCTracer`... actually `MDCTracer` untouched). Fully compatible.
- **Config**: `job-log-prune-interval-minutes` is a brand-new key with a safe default (`60`); no existing key renamed/removed. Fully compatible on the config-schema level.
- **Wire protocol**: `MDCContext`'s legacy 3-segment string is verified bidirectionally compatible against its one real serialization consumer (`TracingOperation`), see 1.1 above.
- **Java serialization**: `TelemetryLogsConfig` gains an explicit `serialVersionUID = 1L` plus a new field. I checked whether this object is ever persisted/exchanged across nodes with mixed versions (checkpoint state, IMap) — it is not: it's rebuilt fresh from YAML on every node at `YamlSeaTunnelDomConfigProcessor.parseTelemetryLogsConfig` (`YamlSeaTunnelDomConfigProcessor.java:576-589`), never cached or checkpointed. No cross-version incompatibility here.
- **Functional/runtime behavior (not a schema/API break, but a real behavior change)**: with `scheduled-deletion-enable` defaulting to `true` (`ServerConfigOptions.java:114-121`, pre-existing default) and `job-log-prune-interval-minutes` defaulting to `60` (this PR), the new pruner is **on by default** for every upgraded deployment with zero config changes, and per Issue 1 it can delete files it was never meant to touch. This is the one place I'd push back on the PR description's "fully backward compatible, additive, opt-in" framing — it's opt-out, not opt-in, and the blast radius (as implemented) extends past job logs.

- Extra CPU/GC from the duplicate `Data` deserialization on every `submitJob` call (1.1) — proportional to job config/DAG size, on a shared Hazelcast operation thread.
- New daemon `ScheduledExecutorService` thread (`task-log-prune`) per node when logs are enabled — single thread, low overhead, correctly shut down.
- Filesystem: `pruneRolledJobLogSegments` does a full `logDir.listFiles()` scan every interval (default hourly) — fine for a log directory, but note it now also `lastModified()`-stats and (as implemented) may delete files outside its intended scope (Issue 1).
- `Files.deleteIfExists` used in both `clean()` and `pruneRolledJobLogSegments` — safe/idempotent against concurrent deletion (e.g. two prune cycles racing, or Log4j2's own `DefaultRolloverStrategy` delete action from 1.1/Issue 1 racing the Java-side pruner). No new lock contention introduced.

- `extractJobMode(Data)` catches `Exception`, logs at `warning` with the job id and fails closed to `null` (`CoordinatorService.java:2292-2306`) — good, no silent swallow, no stack-trace-less log.
- `pruneRolledJobLogSegments` catches `IOException` per file and keeps going, logging a summary (`deleted`/`failed` counts) only when something happened, at `info` — reasonable, not spammy.
- `startPrune` validates `pruneIntervalMinutes <= 0` and `ttl == null/zero/negative` before scheduling, logging why it declined to start — good defensive coding.
- No secrets/credentials anywhere in this diff (it's log-plumbing and job-mode routing only).

## 2. Code Quality

### 2.1 Coding Standards / Comment Completeness

Comment density is genuinely good here — every new public method (`JobLogFileNameMatcher.isJobLogFile/isRolledSegment/activeFileName`, `TaskLogManagerService.startPrune/shutdown/pruneRolledJobLogSegments`, `CoordinatorService.createJobMdcExecutorService/extractJobMode`, all new `MDCContext` fields/constructors) has a Javadoc explaining purpose, nullability, and — importantly — *why* (e.g. "Job mode controls log cleanup policy, so a nested context must not inherit it."). License headers present on all 3 new files. No `System.out.println`, no inline single-line Javadoc, no forbidden symbols in logs/comments. This meets house style well.

New/changed unit tests: `MDCTracerTest` (+2 tests), `JobLogFileNameMatcherTest` (new, 2 tests), `TaskLogManagerServiceTest` (rewritten, 3 tests), `CoordinatorServiceJobLogModeTest` (new, 2 tests). I verified all of them actually ran and passed in the fork CI's Windows JDK11 unit-test job (the only unit-test matrix leg that ran to completion before the run was cancelled due to an unrelated node-startup flake — see 4 below):

```
Tests run: 2 ... JobLogFileNameMatcherTest
Tests run: 3 ... TaskLogManagerServiceTest
Tests run: 2 ... CoordinatorServiceJobLogModeTest
```

That said, the test suite has a real, exploitable gap that directly hides Issue 1: **no test ever calls `pruneRolledJobLogSegments` with a non-job file present in the directory.** `TaskLogManagerServiceTest` only exercises `clean(jobId)` (terminal cleanup, which uses the correctly-scoped `isJobLogFile`); the new running-state pruner (`pruneRolledJobLogSegments`, the actual new capability this PR is introducing) has **zero test coverage** — not even a basic "given a rolled job segment older than TTL and one younger than TTL, only the old one is deleted" test, let alone the cross-appender collision this PR actually introduces. Had a test dropped a `seatunnel.log.2026-01-01-1`-style file into the temp dir with an old `lastModified()` and asserted it survives `pruneRolledJobLogSegments`, it would have failed and caught Issue 1 before this review.

Stability rating: **Risk present**. Not "High risk" in the flaky-test sense (no `Thread.sleep`/nondeterminism/shared static state in the new tests themselves — they're clean, deterministic `@TempDir`-based tests), but the coverage gap on the single riskiest new method (`pruneRolledJobLogSegments`) is exactly the kind of gap the house style calls out as blocking-worthy, and it's compounded by the fact that the missing test would have caught a real bug.

### 2.3 Docs

`docs/en/engines/zeta/logging.md` and `docs/zh/engines/zeta/logging.md` are both updated consistently (checked side by side — same default, same disable semantics, same "active/unclassified never touched" claim). Two small gaps:
1. Neither doc mentions that `job-log-prune-interval-minutes` is a no-op unless the pre-existing `scheduled-deletion-enable` is also `true` (`SeaTunnelServer.java:187-199` gates the entire `TaskLogManagerService` construction, including the new pruner, behind that flag). A reader who only sets the new key without knowing this history would be confused why nothing happens if they'd also set `scheduled-deletion-enable: false`.
2. Both docs claim "only the date-indexed rolled segments... are eligible" for the pruner — accurate to the code's *intent*, but per Issue 1 the *implementation* is broader than "job log rolled segments" once you account for the shared directory.

## 3. Architectural Soundness

The split between `clean(jobId)` (terminal, exact-job, ownership-based) and `pruneRolledJobLogSegments(ttl)` (running-state, age-based, directory-scan-based) is a reasonable design for the stated goal (let long streaming jobs shed old segments without waiting for job termination). The `JobLogFileNameMatcher` utility class is a good factoring — pulling the matching logic out of `TaskLogManagerService` into a testable, static, single-responsibility class is the right call. The flaw is that the two matcher methods on that class (`isJobLogFile` vs `isRolledSegment`) have inconsistent scoping guarantees, and the caller of the weaker one (`pruneRolledJobLogSegments`) doesn't compensate for that. This is a one-file, narrow fix, not a design rework.

## 4. CI Status

The Apache-side "Build" check is a pointer into the contributor's fork run (`corgy-w/seatunnel` run `31241885644`), which I fetched directly. Failing jobs there:
- `unit-test (11, windows-latest)`: real failure, but in a **pre-existing, unrelated** test — `FollowerRunningJobsFilterTest > AbstractSeaTunnelServerTest.before:70 » IllegalStateException: Node failed to start!` (a Hazelcast node bring-up flake on the Windows runner). The 3 sibling matrix legs (`unit-test 8/11 ubuntu`, `unit-test 8 windows`) were `cancelled`, not run, presumably by a fail-fast matrix setting — but all four of this PR's own new/changed test classes (`JobLogFileNameMatcherTest`, `TaskLogManagerServiceTest`, `CoordinatorServiceJobLogModeTest`, and the two new `MDCTracerTest` methods) ran to completion inside that same Windows job and passed (see 2.2 for the exact counts), before the later, unrelated `FollowerRunningJobsFilterTest` failure aborted the module.
- `kafka-connector-it (11)`, `oracle-cdc-connector-it (11)`, `all-connectors-it-1 (11)`, `all-connectors-it-4/7 (8)`: I spot-checked `oracle-cdc-connector-it (11)` (job `93064383021`) — it fails in `OracleCDCWithSchemaChangeIT.testOracleCdc2OracleWithSchemaEvolutionCase` with `ConditionTimeout`/`AssertionFailedError: iterable lengths differ, expected: <5> but was: <1>`, a known-flavor flaky CDC schema-evolution timing test, and this PR touches no connector/CDC code at all. I did not deep-dive the other 3 connector-IT failures individually, but given the diff surface (Zeta engine log/tracing internals only, zero connector module changes) and that the sibling `oracle-cdc-connector-it (8)` leg on the same run passed, these read as the ordinary background flake rate of the connectors-IT suite rather than anything caused by this diff.

I'd stop short of recommending "sync latest dev and rerun" here, since the PR is not meaningfully diverged (`ahead_by: 1, behind_by: 3` against `dev`) and I have no evidence dev already fixed this specific Windows node-startup flake or the Oracle CDC timing flake — a straight re-run of just the failed jobs (`gh run rerun <run-id> --failed`, on the fork) is the lower-cost next step once Issue 1 is addressed and a new commit is pushed anyway.

`mergeable_state` is `blocked` (branch-protection/review-required gate, not a merge conflict — `mergeable: true`, no conflicts found).

## 5. Issue Summary

| 1 | High | `JobLogFileNameMatcher.java:88-103` (`isRolledSegment`), `TaskLogManagerService.java:173-207` (`pruneRolledJobLogSegments`) | The new background pruner is enabled by default (`scheduled-deletion-enable=true`, `job-log-prune-interval-minutes=60`) and deletes any file in the shared log directory matching `\.\d{4}-\d{2}-\d{2}-\d+`, with no "job-" prefix check. The pre-existing system log rotation (`config/log4j2.properties:90-93`, `${file_name}.log.%d{yyyy-MM-dd}-%i`, default `seatunnel.log.*`) lives in the **same directory** (`${file_path}`, confirmed via `LogUtil.getLogPath()`) and matches the same suffix pattern, so `seatunnel.log.yyyy-MM-dd-N` segments get deleted by the job pruner, not just by Log4j2's own rollover-delete action. No test exercises `pruneRolledJobLogSegments` against a non-job file, so this ships silently. | No |
| 2 | Medium | `CoordinatorService.java:1274` + `:1281` | `submitJob` (the normal job-submission path, invoked from a Hazelcast operation thread via `SubmitJobOperation`) deserializes the client's `Data jobImmutableInformation` twice per call — once inside `createJobMdcExecutorService`→`extractJobMode` to get the job mode, once again via `deserializeJobImmutableInformation` for the actual job setup. Avoidable CPU/GC cost on a shared operation thread, scaling with job config size. | No |
| 3 | Low | `CoordinatorService.java:989` / `:1051` | Same double-deserialization pattern on the master-active-switch restore path (`restoreAllRunningJobFromMasterNodeSwitch` → `restoreJobFromMasterActiveSwitch`), for every running job being restored. Lower severity since it's a failover-only path, but it adds load exactly when the cluster is already recovering. | No |
| 4 | Low | `SeaTunnelServer.java:97` vs `config/log4j2.properties:24` (`property.file_ttl = 7d`) | `DEFAULT_JOB_LOG_PRUNE_TTL = Duration.ofDays(7)` is a hardcoded Java constant that must be kept in sync by hand with the Log4j2 `file_ttl` property (the code comment says as much). If an operator changes `file_ttl` in their log4j2 config without knowing about this constant, the two cleanup mechanisms silently disagree. No test asserts they match. | No |
| 5 | Low | `docs/en/engines/zeta/logging.md`, `docs/zh/engines/zeta/logging.md` | The new `job-log-prune-interval-minutes` entry doesn't mention it's a no-op unless `scheduled-deletion-enable` is also `true`. | No |
| 6 | Low | `MDCContext.java` (`valueOf`/`parseNullableId`) | Undisclosed behavior change: old `valueOf` nulled both `pipelineId` and `taskId` when either was the `0` sentinel; new code nulls them independently. No current caller appears to exercise the affected shape (job+pipeline-only serialized context), so this looks latent, but worth a one-line mention in the PR description since `valueOf` is used on the wire (`TracingOperation`). | No |

## 6. Merge Recommendation

The MDC/job-mode plumbing and the `JobLogFileNameMatcher.isJobLogFile`/`clean()` exact-match fix are solid, well-tested, and verified backward/forward compatible across the real RPC consumer. The core new capability being shipped in this PR, though — the running-state pruner — has a scoping bug that, as implemented, reaches outside job logs into the engine's own system log rotation, and the test suite has no coverage that would have caught it.

- Issue 1 — scope `isRolledSegment` (or the call site in `pruneRolledJobLogSegments`) to files that actually belong to a job (e.g. require the `job-<id>.log` prefix, mirroring `isJobLogFile`'s anchoring), and add a unit test that plants a non-job rolled-looking file (e.g. `seatunnel.log.2026-01-01-1`) in the temp dir and asserts it survives `pruneRolledJobLogSegments`.

**Recommended fixes (non-blocking, but should land before or shortly after merge):**
- Issue 2 — avoid the double deserialization on the `submitJob` hot path (reuse `submittedJobImmutableInformation` for mode extraction instead of parsing `Data` twice).
- Issue 3 — same idea for the master-switch restore path, lower priority.
- Issue 4/5/6 — small doc/comment additions; no code risk, easy to fold into the same follow-up commit.

DanielLeens (commented):

- User pain point: `TaskLogManagerService` previously matched job log files with a naive `name.contains(String.valueOf(jobId))` check, so `job-1.log` cleanup could collide with `job-10.log`/`job-100.log`, and the per-job appender was a plain `File`, so a long-running streaming job's log grew unbounded until termination.
- Fix approach: add an optional `JobMode` dimension to `MDCContext` (with the legacy 3-segment wire format still round-tripping), replace the substring matcher with an exact `JobLogFileNameMatcher`, promote the per-job appender to `RollingFile`, and add a background pruner (`TaskLogManagerService.startPrune`) that periodically removes rolled segments older than a TTL while the job is still running.
- One-sentence summary: this is a follow-up re-review after a new commit (`533a1e4`, 2026-08-13) landed on top of the version I reviewed on 2026-08-09 — that commit fixes the High-severity scoping bug I flagged (Issue 1) plus the Medium double-deserialization issue (Issue 2) and one of the two doc gaps (Issue 5), and I re-verified all three fixes are correct with no new regressions.

# 0. Re-review Context (no reply threads to answer)

I checked for replies to my 2026-08-09 `CHANGES_REQUESTED` review (`gh api .../issues/11702/comments` and `.../pulls/11702/comments`, both paginated) — there are none; the author addressed the review directly with commit `533a1e49406f895552d6ae59eb1323620c4d8383` ("[Fix][Zeta] Restrict job log pruning to job logs"), committed 2026-08-13T06:40:31Z, four days after my review. No other reviewers have posted (`pulls/11702/reviews` returns only my own review), so there is nothing to incorporate from Step 0.5.

# 1. Code Change Review

## 1.1 Core Logic Analysis

I diffed the reviewed commit (`bcd750e`) against the new head (`533a1e4`) directly (`git diff bcd750e..533a1e4`) to isolate exactly what changed, then re-verified each fix against the surrounding code.

**Issue 1 fix (High -> Fixed) — `JobLogFileNameMatcher.isRolledSegment`**

Before: `isRolledSegment` only checked the trailing suffix after the last `.` against `\.\d{4}-\d{2}-\d{2}-\d+`, with no requirement that the file be a `job-<id>.log` file at all — so `seatunnel.log.2026-08-05-1` (the engine's own Log4j2-rotated system log, produced by the pre-existing `config/log4j2.properties` rollover policy in the *same* log directory) matched and was eligible for deletion by the new pruner.

After (`JobLogFileNameMatcher.java:53-54,98-108`):
```java
private static final Pattern ROLLED_JOB_FILE_PATTERN =
        Pattern.compile("job-\\d+\\.log\\.\\d{4}-\\d{2}-\\d{2}-\\d+");
...
public static boolean isRolledSegment(String fileName) {
    if (fileName == null) {
        return false;
    }
    String tail = fileName;
    int separator = Math.max(fileName.lastIndexOf('/'), fileName.lastIndexOf('\\'));
    if (separator >= 0) {
        tail = fileName.substring(separator + 1);
    }
    return ROLLED_JOB_FILE_PATTERN.matcher(tail).matches();
}
```
The pattern is now anchored on the full `job-<digits>.log.<date>-<index>` shape and matched with `.matches()` (whole-string match, not `find()`), so a file must literally start with `job-` and digits and `.log.` to qualify — `seatunnel.log.*` and any other non-job rolled file in the shared directory is correctly rejected. I confirmed this is not merely a paper fix: the new `TaskLogManagerServiceTest.testPruneDeletesOnlyExpiredRolledJobSegments` (lines 102-131) plants an aged `seatunnel.log.2026-07-13-1` in the same temp dir alongside an aged `job-123.log.2026-07-13-1`, runs `pruneRolledJobLogSegments`, and asserts the job segment is deleted while the system-log segment survives — this is exactly the regression test I asked for in the 2026-08-09 review. `JobLogFileNameMatcherTest` also gained explicit negative assertions for `seatunnel.log.2026-08-05-1` and `metrics.log.2026-08-05-1`.

I re-checked the concurrent-jobs angle this re-review was specifically asked to verify: `isJobLogFile(name, jobId)` (used by the terminal `clean(jobId)` path, unchanged in this commit) anchors on `activeFileName(jobId) = "job-" + jobId + ".log"` via exact-equals or `startsWith(activeFileName + ".")`, so `job-123` still cannot match `job-1234`'s files — this was already correct and remains so. `isRolledSegment` (used by the age-based pruner) is intentionally *not* job-id-scoped — it is designed to sweep every job's expired rolled segments in one pass, not one job's — and the new `\d+` group is greedy and bounded by the literal `.log.` that follows, so there is no ambiguity between adjacent job ids within a single file name (e.g. `job-1234.log.2026-08-05-1` unambiguously extracts job id `1234`, not `123`). No cross-job leakage.

**Issue 2 fix (Medium -> Fixed) — `submitJob` double deserialization**

Before: `createJobMdcExecutorService(jobId, jobImmutableInformation)` (raw `Data`) was called on the Hazelcast operation thread inside `submitJob`, internally deserializing the payload via `extractJobMode(jobId, Data)`, and then the async task deserialized the *same* `Data` again via `deserializeJobImmutableInformation`.

After (`CoordinatorService.java:1270-1284`):
```java
JobImmutableInformation submittedJobImmutableInformation;
try {
    submittedJobImmutableInformation =
            deserializeJobImmutableInformation(jobImmutableInformation);
} catch (Throwable e) {
    String errorMsg = ExceptionUtils.getMessage(e);
    logger.severe(String.format("submit job %s error %s ", jobId, errorMsg));
    jobSubmitFuture.completeExceptionally(new JobException(errorMsg));
    return new PassiveCompletableFuture<>(jobSubmitFuture);
}
MDCExecutorService mdcExecutorService =
        createJobMdcExecutorService(jobId, submittedJobImmutableInformation);
```
Deserialization now happens exactly once, before the mode is derived (`createJobMdcExecutorService(long, JobImmutableInformation)` overload, which calls `extractJobMode(JobImmutableInformation)` directly with no re-deserialize), and the same already-parsed object is passed into the async lambda that used to redo the parse. I checked the failure-path semantics for a corrupt payload: the new early `catch (Throwable e)` reproduces the exact same log message format and `JobException` wrapping that the async lambda's own catch block uses further down (`CoordinatorService.java:1345-1348`), and since this failure happens strictly before `runningJobInfoIMap.put`/`pendingJobQueue.put`/`runningJobMasterMap` are touched, no cleanup step is skipped — the early-return path is a true no-op with respect to shared state. `ExceptionUtils` and `JobException` were already imported and used elsewhere in the file, so this compiles cleanly (verified via `grep` for both imports). The restore-path double-deserialization I flagged as Issue 3 (`:989`/`:1051`) is untouched by this commit — expected, since I rated it Low/non-blocking and the author reasonably focused on the blocker.

**Issue 5 fix (Low -> Fixed) — docs**

Both `docs/en/engines/zeta/logging.md` and `docs/zh/engines/zeta/logging.md` now state the new key "only takes effect when `scheduled-deletion-enable` is also `true`" (closing gap 1 from my prior Issue 5), and both now say "date-indexed rolled **job** segments" instead of the broader "rolled segments" (closing gap 2, and now factually accurate since the implementation is truly job-scoped). I diffed both files side by side — the English and Chinese wording changes are consistent with each other.

No other files were touched by this commit (`MDCContext.java`, `SeaTunnelServer.java`, `ServerConfigOptions.java`, `TelemetryLogsConfig.java`, `config/log4j2.properties` are all byte-identical to what I reviewed on 2026-08-09), so my prior analysis of the `MDCContext` wire-format compatibility (verified bidirectionally against the real `TracingOperation` RPC consumer) and the `TaskLogManagerService` lifecycle wiring in `SeaTunnelServer` still stands unchanged.

## 1.2 Compatibility Impact

**Fully compatible.** Same conclusion as the 2026-08-09 review for the API/SPI/config/wire-protocol/serialization surface (nothing in this commit changes any of that). The one caveat I raised before — the pruner is on-by-default (opt-out, not opt-in) for every upgraded deployment — still applies as a functional-behavior note, but the blast radius is now correctly job-scoped, which was the actual compatibility risk (deleting the engine's own system log rotation). That risk is resolved.

The avoidable double-deserialization on the `submitJob` hot path (every job submission, on a shared Hazelcast operation thread) is eliminated. The lower-priority restore-path duplicate (Issue 3, `:989`/`:1051`, failover-only) remains, unchanged from before.

| # | Location | Problem | Risk | Best improvement | Severity | Status |
|---|---|---|---|---|---|---|
| Issue 1 | `JobLogFileNameMatcher.java` (`isRolledSegment`), `TaskLogManagerService.java:194` | Pruner matched any `*.yyyy-MM-dd-N` file in the shared log dir, not just job logs | Deleted the engine's own rotated system log | Anchor on `job-<id>.log.` prefix | High | **Fixed in 533a1e4**, verified with new regression test |
| Issue 2 | `CoordinatorService.java:1274`/`:1281` (old) | `submitJob` deserialized client `Data` twice on the operation thread | Avoidable CPU/GC on every submit | Deserialize once, reuse result | Medium | **Fixed in 533a1e4** |
| Issue 3 | `CoordinatorService.java:989`/`:1051` | Same double-deserialization pattern on the master-active-switch restore path | Extra cost during failover, when the cluster is already stressed | Reuse `jobInfo.getJobImmutableInformation()`'s deserialized form if available, or accept the pre-parsed object | Low | Not fixed (unchanged, non-blocking as before) |
| Issue 4 | `SeaTunnelServer.java:97` vs `config/log4j2.properties:24` (`property.file_ttl = 7d`) | `DEFAULT_JOB_LOG_PRUNE_TTL = Duration.ofDays(7)` must be kept in sync by hand with the Log4j2 `file_ttl` property | Operators who change `file_ttl` without knowing about this constant get silently disagreeing cleanup windows | Derive the Java default from the same config source, or document the coupling explicitly | Low | Not fixed (unchanged, non-blocking as before) |
| Issue 6 | `MDCContext.java` (`valueOf`/`parseNullableId`) | Undisclosed behavior change: old code nulled both `pipelineId`/`taskId` when either was `0`; new code nulls them independently | Latent — no current caller constructs/serializes a job+pipeline-only context | One-line PR description mention | Low | Not fixed (unchanged, non-blocking as before, latent) |

Raised by another reviewer: No for all (no other reviews exist on this PR).

# 2. Code Quality Assessment

## 2.1 Coding Standards

The new `ROLLED_JOB_FILE_PATTERN` field and the rewritten `isRolledSegment` Javadoc both explain the "why" (the pruner scans the shared directory, so the matcher must anchor on the job prefix) rather than restating the code. License headers, comment density, and formatting are consistent with the rest of this PR, which was already solid on 2026-08-09.

The coverage gap I called out on 2026-08-09 — "no test ever calls `pruneRolledJobLogSegments` with a non-job file present" — is now closed by `TaskLogManagerServiceTest.testPruneDeletesOnlyExpiredRolledJobSegments`, which is a well-constructed, deterministic `@TempDir`-based test covering five file states in one pass (expired job segment deleted, fresh job segment kept, active file kept, unclassified sidecar kept, expired *non-job* rolled file kept). `JobLogFileNameMatcherTest` also gained matching negative-path assertions and a Windows-path-separator case (`/tmp/job-123.log...`). No `Thread.sleep`, no nondeterministic timing, no shared static state introduced.

Stability rating: **Stable**. This closes the "Risk present" rating from my prior review — the single riskiest new method (`pruneRolledJobLogSegments`) now has direct, targeted coverage for the exact bug it previously shipped with.

## 2.3 Documentation Updates

`docs/en` and `docs/zh` are both updated and remain consistent with each other (checked side by side); the previously-missing "no-op unless `scheduled-deletion-enable` is true" caveat and the "job segments" (not just "segments") wording are now both present in both languages.

# 3. Architectural Soundness

Precise fix — a one-regex change plus a matching regression test, exactly scoped to the bug, no unrelated refactor.

## 3.2 Maintainability
Improved: the fix makes the invariant ("only `job-<id>.log.*` rolled files are eligible") explicit in the pattern itself rather than relying on incidental suffix-matching behavior, and the new test pins that invariant down.

## 3.3 Extensibility
Unchanged from 2026-08-09 — the `clean()`/`pruneRolledJobLogSegments()` split remains a reasonable design for terminal vs. running-state retention.

Unchanged from 2026-08-09 — `MDCContext`'s legacy 3-segment wire format was verified bidirectionally compatible against `TracingOperation`, its one real RPC consumer, and nothing in this commit touches that file.

# 4. Issue Summary

| Number | Issue | Location | Severity | Status |
|---|---|---|---|---|
| 1 | Pruner matched non-job rolled files (system log) | `JobLogFileNameMatcher.java` | High | Fixed (533a1e4) |
| 2 | `submitJob` double-deserialization on hot path | `CoordinatorService.java:1270-1284` | Medium | Fixed (533a1e4) |
| 3 | Restore-path double-deserialization | `CoordinatorService.java:989`/`:1051` | Low | Open, non-blocking |
| 4 | Hardcoded TTL not synced with Log4j2 `file_ttl` | `SeaTunnelServer.java:97` | Low | Open, non-blocking |
| 5 | Doc gaps (no-op condition, scope wording) | `docs/en`, `docs/zh` `logging.md` | Low | Fixed (533a1e4) |
| 6 | `MDCContext.valueOf` latent behavior change | `MDCContext.java` | Low | Open, non-blocking, latent |

# 5. Merge Recommendation

The commit that landed since my 2026-08-09 `CHANGES_REQUESTED` review fixes the one blocking issue (Issue 1) with a correctly-scoped regex fix and a regression test that specifically plants a non-job rolled file and asserts it survives pruning — I verified this is a real fix, not a superficial one, by tracing the pattern change and re-running the concurrent-job-id-collision analysis this re-review was specifically asked to check. It also fixes the Medium double-deserialization issue (Issue 2) and one of the two doc gaps (Issue 5) as a bonus. Nothing in the new commit introduces a new correctness or compatibility problem that I could find.

**Blockers:** None remaining.

**Recommended fixes (non-blocking, can land in a follow-up):**
- Issue 3 — apply the same single-deserialize pattern to the master-active-switch restore path.
- Issue 4 — reduce the risk of the Java `DEFAULT_JOB_LOG_PRUNE_TTL` constant silently drifting from the Log4j2 `file_ttl` property (e.g. a comment cross-reference is already present per the prior review, but deriving one from the other, or asserting equality in a test, would be stronger).
- Issue 6 — a one-line PR description note on the `MDCContext.valueOf` null-handling change for future readers, since it's latent but touches a wire-serialized path.

No blocking issues found beyond what's already listed as open (non-blocking) above.

DanielLeens (approved):

- User pain point: `TaskLogManagerService` used to match job log files with a naive `name.contains(String.valueOf(jobId))` check, so `job-1.log` cleanup could collide with `job-10.log`/`job-100.log`, and the per-job Log4j2 appender was a plain `File`, so a long-running streaming job's log grew unbounded until the job terminated.
- Fix approach: add an optional `JobMode` dimension to `MDCContext` (legacy 3-segment wire format still round-trips unchanged), replace the substring matcher with an exact `JobLogFileNameMatcher`, promote the per-job appender to `RollingFile`, and add a background pruner (`TaskLogManagerService.startPrune`) that periodically removes rolled segments older than a TTL while the job is still running.
- One-sentence summary: this is a re-review triggered by a new commit on the PR; after tracing the full current diff and call chain from scratch, the content is byte-for-byte identical to the version I already reviewed and endorsed on 2026-08-13 (commit `533a1e4`) — the PR was simply rebased onto the latest `dev` (new SHAs, same tree), so all previously-fixed issues remain fixed and no new issue was introduced.

# 0. Re-review Context

This PR was flagged for re-review because the head SHA changed. I fetched the current head, built an isolated worktree, and diffed it against `dev` from scratch (`git diff dev...HEAD`, 16 files, +842/-50) rather than trusting the prior review summary. That diff is content-identical to what I verified on 2026-08-13: same `JobLogFileNameMatcher.ROLLED_JOB_FILE_PATTERN` anchoring fix, same single-deserialize fix in `CoordinatorService.submitJob`, same doc wording, same test additions (`TaskLogManagerServiceTest.testPruneDeletesOnlyExpiredRolledJobSegments`, `JobLogFileNameMatcherTest`, `CoordinatorServiceJobLogModeTest`, the two new `MDCTracerTest` cases). No reply threads or new reviews exist from anyone else on this PR (checked `pulls/11702/reviews` and both comment endpoints, paginated — empty except my own two reviews), so there is nothing new to incorporate from another reviewer, and nothing here was missed in the prior pass — this is a clean confirmation, not a carryover.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**`MDCContext` wire-format compatibility** (`seatunnel-api/.../tracing/MDCContext.java`) — unchanged from the reviewed version. `TracingOperation` (`seatunnel-engine-server/.../task/operation/TracingOperation.java:51,57`) still serializes via `out.writeString(context.toString())` / `context = MDCContext.valueOf(in.readString())` across the real Hazelcast RPC boundary (it is the base class for task-ack, task-status, slot-request, split-assign operations). I re-verified both rolling-upgrade directions hold: old node's 3-segment string parses fine against the new `valueOf` (`arr.length >= 4` guard on the mode segment), and a new node's 4-segment string is still safely truncated by an old `valueOf` that only ever reads `arr[0..2]`.

**`CoordinatorService` job-mode threading** — the `createJobMdcExecutorService` helper is correctly wired into all four job-executor entry points (`submitJob`, `pendingJobSchedule`, `restoreAllRunningJobFromMasterNodeSwitch`, `restoreJobFromMasterActiveSwitch`), and `extractJobMode` fails closed to `null`/unclassified on any missing or corrupt field rather than guessing `STREAMING`. The `submitJob` hot path deserializes the client-supplied `Data jobImmutableInformation` exactly once now (the fix for the Medium-severity double-deserialization I flagged on 2026-08-09), with the early failure path correctly short-circuiting before any shared state (`runningJobInfoIMap`, `pendingJobQueue`) is touched.

**`JobLogFileNameMatcher` / pruner scope** — this was the High-severity blocker in my first review (the pruner's `isRolledSegment` had no `job-` prefix requirement, so it could delete the engine's own rotated `seatunnel.log.yyyy-MM-dd-N` system log living in the same directory). The fix is still present and correctly anchored:
```java
private static final Pattern ROLLED_JOB_FILE_PATTERN =
        Pattern.compile("job-\\d+\\.log\\.\\d{4}-\\d{2}-\\d{2}-\\d+");
```
matched with `.matches()` (whole-string), so only `job-<digits>.log.<date>-<index>` qualifies; `seatunnel.log.*` and `metrics.log.*` are rejected. `isJobLogFile` (used by the terminal `clean(jobId)` path) remains separately and correctly anchored on `activeFileName(jobId)` so adjacent ids (`job-123` vs `job-1234`) still cannot collide.

**`TaskLogManagerService` lifecycle** — `initClean()` runs synchronously before `startPrune()` on the same thread inside `SeaTunnelServer.init()`, so there is no visibility race on the non-`volatile` `path` field (the executor-submission happens-before edge covers it). `shutdown()` is wired into `SeaTunnelServer.shutdown(boolean)`, is idempotent, and null-guards the executor reference.

## 1.2 Compatibility Impact

Fully compatible. No public API/SPI shape changes (only additive overloads on `MDCContext`). `job-log-prune-interval-minutes` is a new config key with a safe default (60 minutes); no key renamed or removed. `TelemetryLogsConfig` is rebuilt fresh from YAML on every node (`YamlSeaTunnelDomConfigProcessor.parseTelemetryLogsConfig`) and never checkpointed or exchanged cross-node, so the new `serialVersionUID`/field addition carries no cross-version risk. The one functional nuance worth restating: with `scheduled-deletion-enable` defaulting to `true` (pre-existing) and the new prune interval defaulting to `60`, the pruner is on by default for every upgraded deployment — opt-out rather than opt-in — but its blast radius is now correctly scoped to job files only, which was the actual risk.

The avoidable double-deserialization on the `submitJob` hot path (shared Hazelcast operation thread) is eliminated. The lower-priority duplicate on the master-active-switch restore path (`CoordinatorService.java:989`/`:1051`, failover-only) is still present — see Issue 1 below, unchanged in severity from my prior review. The new `task-log-prune` daemon thread is a single low-overhead scheduled executor, correctly shut down. `pruneRolledJobLogSegments` does one `listFiles()` scan of the log directory per interval (default hourly) — proportionate for a log directory, and `Files.deleteIfExists` makes it safe against a concurrent delete race with Log4j2's own rollover-delete action.

`extractJobMode` catches `Exception`, logs at `warning` with the job id, and fails closed — no silent swallow. `pruneRolledJobLogSegments` catches `IOException` per file, keeps going, and logs a deleted/failed summary only when something happened (not spammy). `startPrune` validates both the interval and the TTL before scheduling and logs why it declined to start when invalid. No secrets or credentials appear anywhere in this diff.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Comment density is genuinely good: every new public method on `JobLogFileNameMatcher` and `TaskLogManagerService`, and the new `CoordinatorService` helpers, has a Javadoc that explains purpose, nullability, and the *why* (e.g. the comment on `ROLLED_JOB_FILE_PATTERN` explicitly states it must anchor on the `job-<id>.log` prefix because the pruner scans the shared directory). License headers are present on all new files. No `System.out.println`, no single-line inline Javadoc, no forbidden symbols in logs or comments.

New/changed tests: `MDCTracerTest` (+2), `JobLogFileNameMatcherTest` (new, 2 cases including the negative `seatunnel.log.*`/`metrics.log.*` assertions), `TaskLogManagerServiceTest` (rewritten, including `testPruneDeletesOnlyExpiredRolledJobSegments`, which plants an expired job segment, a fresh job segment, the active file, the unclassified sidecar, and an expired *non-job* rolled file side by side and asserts only the expired job segment is removed), `CoordinatorServiceJobLogModeTest` (new, 2 cases covering both the happy path and the fail-closed path for missing/incomplete job config). All are deterministic `@TempDir`-based tests with no `Thread.sleep`, no shared static state, no timing races.

Stability rating: **Stable**. This is exactly the regression coverage I asked for after the High-severity finding in my first review, and it directly pins the invariant that was previously broken (non-job rolled files must survive the pruner).

## 2.3 Documentation Updates

`docs/en/engines/zeta/logging.md` and `docs/zh/engines/zeta/logging.md` are both updated and consistent with each other: the new `job-log-prune-interval-minutes` key is documented with its default, its no-op condition (only takes effect when `scheduled-deletion-enable` is also `true`), and an explicit statement that only date-indexed rolled *job* segments are eligible — accurately reflecting the fixed implementation.

# 3. Architectural Soundness

The split between `clean(jobId)` (terminal, exact-job, ownership-based) and `pruneRolledJobLogSegments(ttl)` (running-state, age-based, directory-scan-based) is a clean design for letting long streaming jobs shed old segments without waiting for termination. `JobLogFileNameMatcher` factors the matching logic out into a small, testable, single-responsibility utility class.

## 3.2 Maintainability
The `ROLLED_JOB_FILE_PATTERN` fix makes the "only `job-<id>.log.*`" invariant explicit in the regex itself rather than relying on incidental suffix behavior, and the new test pins that invariant down for future changes.

## 3.3 Extensibility
The `clean()`/`pruneRolledJobLogSegments()` split leaves room for a future retention/classification policy (the stated end goal) to plug in without re-touching the matching primitives.

`MDCContext`'s legacy 3-segment wire format is verified bidirectionally compatible against its one real RPC consumer (`TracingOperation`), so a rolling upgrade between old and new engine nodes is safe in both directions.

# 4. Issue Summary

| # | Issue | Location | Severity | Status | Raised by another reviewer |
|---|---|---|---|---|---|
| 1 | Restore-path double-deserialization on master-active-switch | `CoordinatorService.java:989` / `:1051` | Low | Open, non-blocking | No |
| 2 | Hardcoded `DEFAULT_JOB_LOG_PRUNE_TTL` not derived from / asserted against Log4j2 `file_ttl` | `SeaTunnelServer.java` (constant), `config/log4j2.properties` (`file_ttl`) | Low | Open, non-blocking | No |
| 3 | `MDCContext.valueOf` null-handling behavior change (old code nulled both `pipelineId`/`taskId` when either was the `0` sentinel; new code nulls them independently) | `MDCContext.java` (`valueOf`/`parseNullableId`) | Low | Open, non-blocking, latent (no current caller exercises the affected shape) | No |

All three are carried over unchanged from my 2026-08-13 review; nothing new surfaced in this pass. The High-severity pruner scoping bug and the Medium-severity `submitJob` double-deserialization from my first (2026-08-09) review remain fixed.

# 5. Merge Recommendation

I re-traced the entire current diff and the runtime call chain from scratch rather than relying on my prior summary, and confirmed the content is unchanged from the version I already validated: the High-severity pruner-scoping fix is intact and covered by a targeted regression test, the `submitJob` hot-path double-deserialization fix is intact, the wire-format compatibility of `MDCContext` across the real Hazelcast RPC path still holds in both upgrade directions, and no new correctness, compatibility, or performance issue is introduced.

**Blockers:** None.

**Recommended fixes (non-blocking, fine to land in a follow-up):**
1. Issue 1 — apply the same single-deserialize pattern to the master-active-switch restore path.
2. Issue 2 — reduce the risk of `DEFAULT_JOB_LOG_PRUNE_TTL` silently drifting from the Log4j2 `file_ttl` property (derive one from the other, or assert equality in a test).
3. Issue 3 — a one-line PR description note on the `MDCContext.valueOf` null-handling change, since it's latent but touches a wire-serialized path.

One process note: this PR's `Build` check on the fork is currently queued/in-progress for the freshly-rebased head commit at the time of this review (it was pushed only minutes ago), so I don't have a fresh green checkmark for this exact SHA to point to. That said, the tree content is identical to the previously-reviewed commit whose relevant unit tests (`JobLogFileNameMatcherTest`, `TaskLogManagerServiceTest`, `CoordinatorServiceJobLogModeTest`, the new `MDCTracerTest` cases) I verified passing in an earlier CI run, and the PR is only 2 commits ahead of `dev` (rebased onto a recent `dev` HEAD), so I'd expect this run to pass on the same basis. No action needed from the author here; this is just a timing note for whoever merges.

PR #11703, opened 2026-08-08
[Fix][Connector-V2] Upgrade shaded Doris thrift-service to 1.0.1

corgy-w:

 Doris connector inherits the same thrift schema fixes that 1.0.1 ships over 1.0.0.

Concretely:

- Bump `org.apache.doris:thrift-service` from `1.0.0` to `1.0.1` in `seatunnel-shade/seatunnel-thrift-service/pom.xml`.
- Scope is intentionally limited to the shaded module property. No Java sources are touched.

No. This is a shaded internal dependency bump. The final relocated artifact name (`seatunnel-thrift-service.jar`) and its public surface remain identical, and the Doris connector keeps importing the same relocated classes. No configuration keys, defaults, or public APIs change.

If a downstream deployer overrides `<thrift-service.version>` from a custom `pom.xml`, that override keeps taking precedence and is unaffected.

No entry is needed in `docs/en/introduction/concepts/incompatible-changes.md`.

Validated on top of `apache/seatunnel` `dev` HEAD `00ca1bc02d`:

```bash
./mvnw -q -pl seatunnel-shade/seatunnel-thrift-service -am spotless:apply
```

```bash
./mvnw -q -pl seatunnel-shade/seatunnel-thrift-service -am   -DskipTests -Dmaven.gitcommitid.skip=true install
```

Both commands finish with exit code 0, confirming that the new artifact `seatunnel-thrift-service-1.0.1.jar` resolves from Maven Central (the only published versions are 1.0.0 and 1.0.1) and that the shaded relocation configuration in the module is still valid against the bumped dependency.

Because the change is a pure version bump inside a shaded module, the existing Doris connector unit tests (which exercise thrift classes through the relocated namespace) remain the contract. No new test cases are required.

### Check list

* [x] If any new Jar binary package adding in your PR, please add License Notice according [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/developer/new-license.md) — N/A, this is a version bump of an existing shaded dependency. License metadata in the POM is unchanged.
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs — N/A, no user-facing feature added.
* [x] If necessary, please update `incompatible-changes.md` to describe the incompatibility caused by this PR. — N/A, fully backward compatible shaded bump.
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) — N/A.
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml) — N/A, the shaded artifact is consumed transitively via the Doris connector.
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml) — N/A.
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/) — N/A, version bump only; existing Doris e2e tests cover the contract.
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config) — N/A.

### DIFF
--- a/seatunnel-shade/seatunnel-thrift-service/pom.xml
+++ b/seatunnel-shade/seatunnel-thrift-service/pom.xml
@@ -26,7 +26,7 @@
     <name>SeaTunnel : Shade : Thrift-Service</name>
 
     <properties>
-        <thrift-service.version>1.0.0</thrift-service.version>
+        <thrift-service.version>1.0.1</thrift-service.version>
     </properties>
 
     <dependencies>


DanielLeens (approved):

The Doris connector's shaded thrift module (`seatunnel-shade/seatunnel-thrift-service`) currently pins `org.apache.doris:thrift-service` at `1.0.0`. That client-side thrift schema (`Types.thrift`) is missing several `TPrimitiveType` enum values that newer Doris Backend (BE) versions can emit on the scan/read path (`VARIANT`, `LAMBDA_FUNCTION`, `AGG_STATE`, `DECIMAL256`, `IPV4`, `IPV6`). When a SeaTunnel job with the Doris source talks to a Doris cluster that uses one of these newer column types, the old, narrower enum on the client side can fail to deserialize the BE's `TScanColumnDesc.type` correctly. This PR bumps the pinned version to `1.0.1`, which carries the widened enum, so the connector's thrift client stays in sync with newer Doris BE schemas.

**Fix approach:** a one-line property bump — `<thrift-service.version>1.0.0</thrift-service.version>` → `<thrift-service.version>1.0.1</thrift-service.version>` — in `seatunnel-shade/seatunnel-thrift-service/pom.xml`. No Java source, shading/relocation config, or downstream connector code is touched.

**One-sentence summary:** Bumps the shaded `org.apache.doris:thrift-service` dependency from 1.0.0 to 1.0.1 so the Doris connector's thrift client recognizes newer BE column-type enum values, with zero API/behavior surface change.

**Before/after example:** Before this PR, if a Doris BE returns a scan column whose `TPrimitiveType` ordinal corresponds to `IPV4` (a value that doesn't exist in the 1.0.0 client enum), `TPrimitiveType.findByValue(...)` on the client side returns `null` for that ordinal, and `SchemaUtils.convertToSchema()` (`seatunnel-connectors-v2/connector-doris/src/main/java/org/apache/seatunnel/connectors/doris/util/SchemaUtils.java:36-44`) would call `desc.getType().name()` on a `null` `TPrimitiveType`, throwing an `NPE`/breaking schema conversion for that table. After this PR, the client-side enum recognizes `IPV4`/`IPV6`/`VARIANT`/`LAMBDA_FUNCTION`/`AGG_STATE`/`DECIMAL256`, so schema conversion succeeds for tables using those column types.

## 1. Code Change Review

### 1.1 Core Logic Analysis

Diff (complete):
```diff
--- a/seatunnel-shade/seatunnel-thrift-service/pom.xml
+++ b/seatunnel-shade/seatunnel-thrift-service/pom.xml
@@ -26,7 +26,7 @@
     <properties>
-        <thrift-service.version>1.0.0</thrift-service.version>
+        <thrift-service.version>1.0.1</thrift-service.version>
     </properties>
```

I read the actual current state of `seatunnel-shade/seatunnel-thrift-service/pom.xml` at the PR head (not just the hunk) to confirm nothing else changed: the `maven-shade-plugin` config, the single `<relocation>` entry (`org.apache.thrift` → `${seatunnel.shade.package}.org.apache.thrift`), and the `build-helper-maven-plugin` attach-artifact config (producing `seatunnel-thrift-service.jar` with `classifier=optional`) are all byte-for-byte unchanged. `connector-doris/pom.xml:73-83` consumes this shaded artifact via `<classifier>optional</classifier>` and explicitly excludes `org.apache.thrift:libthrift` (since that's relocated inside the shaded jar) — that dependency block is untouched by this PR.

**Key findings — what actually changed upstream between `org.apache.doris:thrift-service` 1.0.0 and 1.0.1**, verified directly against the `apache/doris-sdk` GitHub repo (`git compare 1.0.0...doris-sdk-1.0.1`, 5 commits):
- `thrift-service/src/main/thrift/Types.thrift`: the `TPrimitiveType` enum gains 6 new trailing values — `VARIANT, LAMBDA_FUNCTION, AGG_STATE, DECIMAL256, IPV4, IPV6` — appended after the existing `UNSUPPORTED` entry. This is a pure enum extension; no existing enum values, struct fields, or service method signatures were removed, renamed, or renumbered.
- `LICENSE.txt` / `NOTICE.txt`: "fix LICENSE, remove unreferenced items" and "Update NOTICE.txt" housekeeping commits — no functional impact.
- `pom.xml` / `thrift-service/pom.xml`: version bump metadata only (`1.0.0-SNAPSHOT` → `1.0.1` release), no dependency version changes — the transitive `libthrift` version is unchanged.
- `CHANGE-LOG.md` added.

I also confirmed 1.0.1 actually resolves from Maven Central (published 2024-04-07, `thrift-service-1.0.1.jar`, 416,494 bytes) — the PR body's testing claim that it resolves is accurate.

**Runtime path:** The classes actually used by SeaTunnel's Doris connector are `org.apache.doris.sdk.thrift.*` generated stubs (`TDorisExternalService`, `TScanBatchResult`, `TScanOpenParams`, `TScanColumnDesc`, etc.), consumed in `BackendClient.java`, `DorisValueReader.java`, and `SchemaUtils.java` under `seatunnel-connectors-v2/connector-doris/src/main/java/...`. This is the **normal Doris source read path** (BE scan RPC), not a recovery/edge-case-only path — every Doris source job that scans BE data goes through `BackendClient` → thrift deserialization. The new enum values only actually get exercised when the target Doris cluster's BE returns one of the 6 new column types; for all existing/older Doris clusters and column types the change is a no-op (same enum ordinals, same wire behavior).

### 1.2 Compatibility

**Fully compatible.** 
- API/config: no `Option`, config key, or public class/method signature changes — the diff is a single Maven property.
- Shading/relocation: unchanged; only `org.apache.thrift` (the underlying Apache Thrift RPC library) is relocated, not `org.apache.doris.sdk.thrift` (Doris's own generated stubs), so package names the connector imports from are unaffected either way.
- Wire protocol/serialization: additive-only enum extension in `Types.thrift`; no removed/renumbered thrift definitions, so existing struct/enum wire compatibility with both older and newer Doris BE versions is preserved.
- Downstream override: as the PR body notes, and as I verified by reading the pom, a custom parent/profile overriding `<thrift-service.version>` still takes precedence via normal Maven property resolution — unaffected.
- No entry needed in `docs/en/introduction/concepts/incompatible-changes.md` — I agree with the author's assessment here, since there is no incompatible/breaking behavior introduced.

No measurable side effects. Same relocation strategy, same transitive `libthrift` version, jar size difference is negligible for a metadata/enum-only diff. No new locking, retries, threading, or resource-management behavior introduced.

Not applicable — no Java code touched, so no new exception paths, log statements, or error-context changes. As described above, this bump actually *improves* an existing failure mode (NPE/null enum on unrecognized `TPrimitiveType` when talking to a BE that emits one of the 6 new types) without introducing any new one.

## 2. Code Quality

No comment-completeness concerns — this is a one-line XML property value change with no new methods, classes, or fields.

No test code is touched by this PR, and per house rule 8 the mandatory flaky-test analysis only applies when UT/E2E test code is modified — it is not here, so I'm not assigning a Stable/Risk/High-risk rating (N/A, not applicable rather than "Stable" by default).

That said, I checked whether the *existing* test suite actually exercises the affected code path, since the PR body claims "the existing Doris connector unit tests... remain the contract" and that this substitutes for new tests:
- `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-doris-e2e/src/test/java/org/apache/seatunnel/e2e/connector/doris/DorisMultiReadIT.java` and `DorisIT.java` run real Doris-source jobs against a containerized Doris cluster (`container.executeJob(...)`), which exercises the `BackendClient`/thrift scan path this bump touches.
- However, none of these existing tests use the 6 *new* column types (`VARIANT`, `IPV4`, `IPV6`, `LAMBDA_FUNCTION`, `AGG_STATE`, `DECIMAL256`) that are the actual reason this bump matters — I did not find any Doris e2e test schema using those types. So existing tests validate "the bump doesn't regress anything" (fine, since it's additive), but do **not** validate "the bump actually fixes the schema-conversion NPE for BE-side newer types" that appears to be the real motivation. This isn't a blocker for a pure dependency-version bump PR, but it's worth noting for anyone who later wants a regression guard for the underlying problem.

### 2.3 Docs

No `docs/en` or `docs/zh` update is needed — no config option, default, or user-facing behavior changed. I agree with the PR checklist's N/A assessments across the board (license notice, `plugin-mapping.properties`, `seatunnel-dist/pom.xml`, CI label-scope, e2e testcase, `plugin_config`) — none of those apply to an internal shaded-module version bump.

## 3. Architectural soundness

Sound and minimal. This stays entirely within the shaded-dependency boundary (`seatunnel-shade/seatunnel-thrift-service`) and does not leak connector-specific logic into engine/core, consistent with the project's connector architecture guidelines. Scope is exactly as described in the PR body — a single Maven property.

### Dependency-bump justification (per repo policy)

The parent project's dependency rules require any new/upgraded dependency to be justified in terms of shading, size, and conflict risk. Checking the PR body against that bar:
- **Shading**: addressed — the PR body explains the shaded relocation/classifier mechanism is unaffected, and I independently confirmed the shade config is byte-identical.
- **Size**: not explicitly discussed in the PR body, but I independently verified via Maven Central that jar size is essentially unchanged (a 6-value enum + license file edits), so this is a non-issue in practice even though it wasn't spelled out.
- **Conflict risk**: addressed — the PR body notes downstream `<thrift-service.version>` overrides still take precedence; I additionally verified the transitive `libthrift` version is unchanged between 1.0.0 and 1.0.1, so there's no new transitive-conflict surface either.
- Overall the justification present in the description is reasonably solid for a change of this size, and it's specific rather than boilerplate (it named the actual upstream backport PR). One nitpick: the description characterizes the upstream change only vaguely as "thrift schema fixes" without stating what specifically changed (the 6 new `TPrimitiveType` values) — I was able to independently verify this via the `apache/doris-sdk` compare view, but it would have been stronger if the PR body cited the concrete upstream diff/commit instead of a generic description.

**`pom.xml` changed (factual note, not a blocker):** This PR's sole change is to `seatunnel-shade/seatunnel-thrift-service/pom.xml`. Per repo policy, any `pom.xml` change requires explicit human sign-off before it can be pushed/merged — flagging this for the human maintainer merge gate; it is not a code-quality or correctness objection from me.

## 4. Issue Summary

| — | — | No blocking issues found. | — | — |

No High/Medium/Low issues to raise. This is a clean, narrowly-scoped, well-verified dependency bump.

## 5. Merge Recommendation

**Ready to merge.**

- Blockers: none.
- Recommended fixes (non-blocking): 
  1. Consider citing the specific upstream `apache/doris-sdk` 1.0.0→1.0.1 diff (the 6 new `TPrimitiveType` enum values) in the PR description instead of the generic "thrift schema fixes" phrasing, for future auditability.
  2. Optional: if the real-world motivation is the `TPrimitiveType` gap against newer Doris BE versions, a small Doris e2e/UT case using one of the new column types (e.g. `IPV4`) would give a concrete regression guard — but I would not block this PR on it given the change is a pure additive version bump.

CI on this PR is green (Build, Notify test workflow, and labeler checks all pass). GitHub reports `mergeable: true` with `mergeable_state: blocked`, which in this repo's workflow reflects "awaiting required review approval," not a merge conflict — I did not find any actual conflicting files (`compare/dev...head` shows `ahead_by: 1, behind_by: 3`, a small and unrelated drift from `dev`). The `pom.xml` change here still needs the human push-approval sign-off called out above before merging, independent of this code review's conclusion.

davidzollo (approved):
LGTM

davidzollo:
My apologies, I merged it too early. Shade rc3 is currently being released. I will revert this PR for now and wait until the release is complete.

PR #11710, opened 2026-08-08
[Improve][Connector-V2][JDBC] Fix Vertica exact numeric type mapping

yuxuan-luo:
The Vertica JDBC driver always returns "NUMERIC" as the column type name via ResultSetMetaData for all exact decimal types (DECIMAL, NUMERIC, NUMBER, MONEY), which are all synonyms of the same underlying type.

The previous dialect implementation only matched the "DECIMAL" literal, which never triggers at runtime and causes all decimal columns to fail with an unsupported data type error.

This commit adds proper NUMERIC type support:
- Add VERTICA_NUMERIC constant and corresponding mapping branch
- Keep consistent precision handling: precision > 38 falls back to DECIMAL(38, 18) to align with SeaTunnel type system limits
- NUMBER and MONEY types are covered implicitly since the JDBC driver normalizes them to NUMERIC in column metadata

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/vertica/VerticaTypeMapper.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/vertica/VerticaTypeMapper.java
@@ -58,6 +58,7 @@ public class VerticaTypeMapper implements JdbcDialectTypeMapper {
     private static final String VERTICA_BIGINT = "BIGINT";
     private static final String VERTICA_BIGINT_UNSIGNED = "BIGINT UNSIGNED";
     private static final String VERTICA_DECIMAL = "DECIMAL";
+    private static final String VERTICA_NUMERIC = "NUMERIC";
     private static final String VERTICA_DECIMAL_UNSIGNED = "DECIMAL UNSIGNED";
     private static final String VERTICA_FLOAT = "FLOAT";
     private static final String VERTICA_FLOAT_UNSIGNED = "FLOAT UNSIGNED";
@@ -114,6 +115,7 @@ public class VerticaTypeMapper implements JdbcDialectTypeMapper {
                 return BasicType.LONG_TYPE;
             case VERTICA_BIGINT_UNSIGNED:
                 return new DecimalType(20, 0);
+            case VERTICA_NUMERIC:
             case VERTICA_DECIMAL:
                 if (precision > 38) {
                     LOG.warn("{} will probably cause value overflow.", VERTICA_DECIMAL);
@@ -162,7 +164,7 @@ public class VerticaTypeMapper implements JdbcDialectTypeMapper {
             case VERTICA_BINARY:
                 return PrimitiveByteArrayType.INSTANCE;
 
-                // Doesn't support yet
+            // Doesn't support yet
             case VERTICA_GEOMETRY:
             case VERTICA_UNKNOWN:
             default:


DanielLeens (changes requested):
Hi @yuxuan-luo, welcome and thank you for your first contribution to Apache SeaTunnel! I went through this carefully, including the Vertica JDBC driver's documented type-alias behavior, the actual runtime call chain, and the existing test coverage in this module. Overall this is a small, well-targeted, and correctly-diagnosed fix — I verified the root cause against Vertica's own documentation and it checks out. There's one gap I'd like to see closed before merge (test coverage), plus a couple of low-priority follow-ups.

**User pain point:** Any SeaTunnel job that reads a Vertica table (or a Vertica JDBC source query) containing a `DECIMAL`/`NUMERIC`/`NUMBER`/`MONEY` column fails at schema-discovery time with an "unsupported data type" error, even though these are all ordinary, extremely common exact-numeric column types.

**Root cause:** Vertica's SQL layer treats `DECIMAL`, `NUMERIC`, `NUMBER`, and `MONEY` as aliases of a single underlying type. Per Vertica's own docs on [numeric data alias conversion](https://docs.vertica.com/24.3.x/en/connecting-to/client-libraries/accessing/java/jdbc-data-types/numeric-data-alias-conversion/):

> The metadata methods `getColumnType()` and `getColumnTypeName()` return the aliased Vertica server types, not the original column definition types... DECIMAL, NUMBER, MONEY → NUMERIC

So no matter what the column was actually declared as, `ResultSetMetaData.getColumnTypeName()` always reports the literal string `"NUMERIC"`. `VerticaTypeMapper` only had a `case` for the literal `"DECIMAL"`, so that branch was dead code — `"NUMERIC"` always fell through to `default` and threw via `CommonError.convertToSeaTunnelTypeError`.

**Fix approach:** Add `VERTICA_NUMERIC = "NUMERIC"` as a new constant and let it fall through into the existing `VERTICA_DECIMAL` branch, reusing the existing precision/scale handling (including the `precision > 38 → DECIMAL(38,18)` overflow-guard fallback).

**One-sentence summary:** Adds the missing `"NUMERIC"` case (the type name Vertica's JDBC driver actually reports for all `DECIMAL`/`NUMERIC`/`NUMBER`/`MONEY` columns) to `VerticaTypeMapper`, fixing a 100%-reproducible failure for any Vertica table containing an exact-numeric column.

**Before/after example:** Given `CREATE TABLE t (amount NUMERIC(10,2))` (or equivalently declared as `DECIMAL(10,2)`, `NUMBER(10,2)`, or `MONEY`) in Vertica:
- Before: reading `t` via a Vertica JDBC source throws `CommonError.convertToSeaTunnelTypeError` during schema inference — the job never starts.
- After: the column is correctly mapped to SeaTunnel `DecimalType(10, 2)`.

## 1. Code Change Review

### 1.1 Core Logic Analysis

Before:
```java
private static final String VERTICA_DECIMAL = "DECIMAL";
...
case VERTICA_DECIMAL:
    if (precision > 38) {
        LOG.warn("{} will probably cause value overflow.", VERTICA_DECIMAL);
        return new DecimalType(38, 18);
    }
    return new DecimalType(precision, scale);
```

After:
```java
private static final String VERTICA_DECIMAL = "DECIMAL";
private static final String VERTICA_NUMERIC = "NUMERIC";
...
case VERTICA_NUMERIC:
case VERTICA_DECIMAL:
    if (precision > 38) {
        LOG.warn("{} will probably cause value overflow.", VERTICA_DECIMAL);
        return new DecimalType(38, 18);
    }
    return new DecimalType(precision, scale);
```

**Key findings:**
- The fix is minimal, idiomatic for this file (fall-through `case` grouping is already the established pattern here, e.g. all the integer variants), and matches the actual documented driver behavior I verified independently (see root cause above).
- `NUMBER` and `MONEY` are correctly stated to be covered "implicitly" — this is accurate per the same Vertica doc table (`DECIMAL, NUMBER, MONEY → NUMERIC`), so no extra `case` labels are needed for those literals; they never appear as-is in `getColumnTypeName()` output.
- The unrelated one-line reformat of the `// Doesn't support yet` comment's indentation (removing over-indentation before `VERTICA_GEOMETRY`/`VERTICA_UNKNOWN`) is harmless and looks like an incidental `spotless`/formatter fix picked up while editing this method; not a concern.

**Runtime path (traced, not assumed):**
1. `VerticaTypeMapper` only overrides the deprecated `mapping(ResultSetMetaData, int)` method (`VerticaTypeMapper.java:98-176`), not `mappingColumn(BasicTypeDefine)`.
2. `JdbcDialectTypeMapper.mappingColumn(ResultSetMetaData, int)` (default method, `JdbcDialectTypeMapper.java:181-183`) calls `mapping(metadata, colIndex)` — i.e., the overridden method — for every column.
3. `CatalogUtils.getCatalogTable(...)` (`CatalogUtils.java:325-335`) invokes `typeMapper.mappingColumn(resultSetMetaData, index)` for each result-set column when building a `CatalogTable`/`TableSchema` from a JDBC query — this runs during normal schema inference at job submission for any Vertica JDBC source.
4. So this is squarely on the **normal execution path**, not a recovery/boundary path: any job reading a Vertica table/query with an exact-numeric column hits `VerticaTypeMapper.mapping()` on every run, and previously failed 100% of the time before a single row could be read.

### 1.2 Compatibility

**Fully compatible.** This is purely additive: it adds handling for an input value (`"NUMERIC"`) that previously always threw an exception. There is no code path today that could have depended on the old "throw for NUMERIC" behavior succeeding, no config option is renamed/removed, no default value changes, and no serialization/protocol format is touched. Existing `"DECIMAL"`-literal handling (from other dialects/paths, if any exist) is untouched.

Negligible — one additional `case` label in an existing `switch` on a `String`. No new allocations, no additional I/O, no locking/concurrency implications.

Mostly fine, but one small accuracy nit inherited from (and now slightly aggravated by) the pre-existing code: the overflow-warning log statement hardcodes the constant text `VERTICA_DECIMAL` ("DECIMAL") regardless of which literal actually triggered the branch:

```java
case VERTICA_NUMERIC:
case VERTICA_DECIMAL:
    if (precision > 38) {
        LOG.warn("{} will probably cause value overflow.", VERTICA_DECIMAL);   // always logs "DECIMAL"
```

Since `VERTICA_NUMERIC` is — per the driver behavior documented above — actually the *only* string real Vertica traffic will ever produce for this branch in practice, every real overflow warning will now print `"DECIMAL will probably cause value overflow."` even though the JDBC driver reported `"NUMERIC"`. This won't cause incorrect behavior (log-only), but it's misleading when someone is debugging an overflow warning and greps their schema for a literal `DECIMAL` column that isn't declared that way. See Issue 2 below.

## 2. Code Quality

### 2.1 Coding Standards

- ASF license header, formatting, and naming conventions are all consistent with the rest of the file. No wildcard imports, no new dependencies.
- Comment completeness: this is a 3-line diff inside a well-established `switch` pattern; no new methods/classes/fields requiring their own Javadoc were introduced (the new constant follows the exact same style as its neighbors, e.g. `VERTICA_DECIMAL`, `VERTICA_BIGINT_UNSIGNED`), so I don't consider this a comment-completeness gap.

**This is the one gap I'd block on.** I checked `seatunnel-connectors-v2/connector-jdbc/src/test/java/.../dialect/vertica/` and it contains only `VerticaDialectTest.java` (which tests `getUpsertStatementByTableSchema`, unrelated to type mapping) — there is **no** test at all for `VerticaTypeMapper.mapping()`/`mappingColumn()`, neither before nor after this PR.

This module already has an established, lightweight pattern for exactly this kind of test — e.g. `seatunnel-connectors-v2/connector-jdbc/src/test/java/.../dialect/mysql/MySqlTypeMapperTest.java`, which mocks `ResultSetMetaData` with Mockito and asserts the resulting `SeaTunnelDataType`/`Column`. Given that:
- this PR is fixing a previously 100%-reproducible failure,
- the fix is a one-line `case` addition that's trivial to regress (e.g., an unrelated refactor of the `switch` could silently drop the fall-through), and
- the precedent for testing this exact class of logic already exists in the same module,

I'd like to see at least one unit test added before merge, e.g. a new `VerticaTypeMapperTest` asserting:
1. `getColumnTypeName()` returning `"NUMERIC"` with `precision <= 38` maps to `DecimalType(precision, scale)`.
2. `getColumnTypeName()` returning `"NUMERIC"` with `precision > 38` maps to the `DecimalType(38, 18)` overflow fallback.

**Stability rating: N/A** — this PR does not modify any existing test/E2E code, so the mandatory flaky-test analysis doesn't apply here. (The absence of any test is a coverage gap, not a stability risk in existing tests.)

### 2.3 Docs

`docs/en/connectors/source/Vertica.md` (and the `docs/zh` counterpart) has a data-type mapping table that currently only lists:
```
| DECIMAL(x,y)(Get the designated column's specified column size.<38) | DECIMAL(x,y) |
| DECIMAL(x,y)(Get the designated column's specified column size.>38) | DECIMAL(38,18) |
```
Since this PR changes user-visible behavior (columns that previously always failed now work), it would be worth updating this table to mention that `NUMERIC`/`NUMBER`/`MONEY` are Vertica aliases for the same underlying type and are handled by the same `DECIMAL` row — otherwise a user reading the docs today would have no way to know their `NUMERIC` column is (now) supported. See Issue 3 below.

## 3. Architectural Soundness

The fix itself is architecturally sound and appropriately scoped. However, while verifying the root cause against Vertica's official documentation, I noticed the same alias-canonicalization behavior applies more broadly than just `NUMERIC`:

> INTEGER, INT, INT8, SMALLINT, TINYINT → BIGINT
> DOUBLE PRECISION, FLOAT5, FLOAT8, REAL → DOUBLE PRECISION
> DECIMAL, NUMBER, MONEY → NUMERIC

`VerticaTypeMapper` has separate `case` labels for `TINYINT`, `SMALLINT`, `MEDIUMINT`, `INT`, `INTEGER` (mapped individually to `INT_TYPE`) and for `FLOAT`/`DOUBLE` (as distinct literals, not `"DOUBLE PRECISION"`), plus several types that don't appear to exist in Vertica at all (`TINYTEXT`, `MEDIUMTEXT`, `LONGTEXT`, `TINYBLOB`, `MEDIUMBLOB`, `LONGBLOB`, and the various `* UNSIGNED` variants — Vertica has no unsigned integer types). This strongly suggests the whole mapper was originally adapted from `MySqlTypeMapper` rather than written against actual Vertica JDBC driver output, and several other `case` labels here are likely unreachable dead code for the same reason `NUMERIC` was — they just haven't been reported as bugs yet because those types are less commonly hit in practice, or because `FLOAT`/`INT` columns happen to already work via a different path.

This is **out of scope for this PR** (which is correctly scoped to the specific `NUMERIC` bug being reported), but I'd recommend the author or another contributor file a follow-up issue for a full audit of `VerticaTypeMapper` against real Vertica JDBC driver output — see Issue 4.

## 4. Issue Summary

| 1 | Medium (blocking) | `seatunnel-connectors-v2/connector-jdbc/.../vertica/VerticaTypeMapper.java` (new `case VERTICA_NUMERIC`) — no corresponding test file | No unit test covers the fixed `NUMERIC` mapping path (or the existing `DECIMAL`/precision-overflow path). Established precedent exists in the same module (`MySqlTypeMapperTest.java`) for exactly this kind of mock-based test. | No |
| 2 | Low (non-blocking) | `VerticaTypeMapper.java:121` | `LOG.warn("{} will probably cause value overflow.", VERTICA_DECIMAL)` hardcodes the `"DECIMAL"` literal even when reached via the new `NUMERIC` case, so real overflow warnings will always print `DECIMAL` regardless of the actual JDBC-reported type name. | No |
| 3 | Low (non-blocking) | `docs/en/connectors/source/Vertica.md`, `docs/zh/connectors/source/Vertica.md` | Type-mapping table doesn't mention that `NUMERIC`/`NUMBER`/`MONEY` are aliases handled by the same `DECIMAL` row; this is now a real user-facing behavior change worth documenting. | No |
| 4 | Low / informational (non-blocking, out of scope) | `VerticaTypeMapper.java` (broader file) | Several other `case` labels (`TINYINT`/`SMALLINT`/`MEDIUMINT`/`INT`/`INTEGER`, `FLOAT`/`DOUBLE`, `* UNSIGNED` variants, `TINYTEXT`/`MEDIUMTEXT`/`LONGTEXT`/`TINYBLOB`/`MEDIUMBLOB`/`LONGBLOB`) look copy-pasted from `MySqlTypeMapper` and may not match real Vertica JDBC driver output (per Vertica's own alias-conversion docs, e.g. `FLOAT`/`REAL` report as `"DOUBLE PRECISION"`, not `"FLOAT"`/`"DOUBLE"`). Worth a follow-up audit issue; not part of this PR's scope. | No |

## 5. CI Status

The Apache-side `Build` check shows `action_required` with "Unable to detect the workflow run for testing the changes in your PR." I checked the contributor's fork (`yuxuan-luo/seatunnel`) directly via the Actions API and it has **zero** workflow runs at all — this isn't a case of a real CI failure hiding in the fork; GitHub Actions has apparently never been enabled/run on this fork. The branch is also only slightly diverged from `dev` (`ahead_by: 1, behind_by: 2` — not stale enough to be the root cause here). The author will need to enable Actions on their fork (Settings → Actions → "I understand my workflows, go ahead and enable them") and push a commit (or sync with `dev`) to get real CI signal before this can be merged.

## 6. Merge Recommendation

**Blockers (must-fix):**
- Issue 1 — add unit test coverage for `VerticaTypeMapper`'s `NUMERIC`/`DECIMAL` mapping (including the `precision > 38` fallback).
- CI needs to actually run and pass — please enable GitHub Actions on your fork and push a commit so we get a real signal beyond the placeholder "Build" check.

**Recommended fixes (non-blocking, nice to have before merge but won't hold it up):**
- Issue 2 — log the actual reported `type` string instead of the hardcoded `VERTICA_DECIMAL` constant.
- Issue 3 — update `docs/en`/`docs/zh` Vertica type-mapping tables to mention the `NUMERIC`/`NUMBER`/`MONEY` aliases.
- Issue 4 — consider filing a separate follow-up issue for a broader `VerticaTypeMapper` audit against real driver output (not required for this PR).

Thanks again for the contribution and for digging into the actual Vertica driver behavior to find this — the root-cause analysis in the PR description is accurate and matches Vertica's own documentation. Once the test is added and CI is green, this should be good to go.

davidzollo:

```

```

PR #11693, opened 2026-08-07
[Bug] [Connector-V2] Support Dameng NCHAR type

CBOSSX:
### Purpose of this pull request

Closes #11688.

Dameng `NCHAR` columns currently fall through `DmdbTypeConverter`'s unsupported-type branch because only `NVARCHAR` is enumerated. This adds the missing `NCHAR` mapping to SeaTunnel `STRING`, preserves the concrete source type, and adds regression coverage.

### Does this PR introduce _any_ user-facing change?

Yes. Dameng `NCHAR` source columns that previously failed schema conversion with `COMMON-17` are now mapped to SeaTunnel `STRING`. The English and Chinese JDBC source documentation describe the mapping.

### How was this patch tested?

The new regression test failed on the unchanged base with:

```text
ErrorCode:[COMMON-17], ErrorDescription:['Dameng' unsupported convert type 'nchar' of 'test' to SeaTunnel data type.]
```

It passes after the fix. The following checks also pass:

```bash
./mvnw test -pl seatunnel-connectors-v2/connector-jdbc '-Dtest=DmdbTypeConverterTest#testConvertNchar' -Dskip.spotless=true
./mvnw test -pl seatunnel-connectors-v2/connector-jdbc '-Dtest=DmdbTypeConverterTest' -Dskip.spotless=true
./mvnw spotless:check
./mvnw -DskipTests verify
```

### Check list

* [x] No new Jar binary package is added.
* [x] The English and Chinese JDBC connector documentation is updated.
* [x] No incompatible change is introduced, so `incompatible-changes.md` does not need an update.
* [x] No new connector registration is introduced, so plugin mapping, distribution POM, CI label scope, E2E registration, and `plugin_config` do not need updates.

### Scope

* 4 files changed, +22 / -1 lines.
* No dependency, generated, vendored, or public API changes.

PR #11714, opened 2026-08-09
[Bug][connector-jdbc] Support Dameng NCHAR type in DmdbTypeConverter

### LINKED ISSUE #11688: [Bug] [connector-jdbc] Dameng NCHAR type is not supported by the type converter

waterWang:
### Purpose

Add support for Dameng `NCHAR` type in `DmdbTypeConverter`. The NCHAR column type was not handled by the converter, causing an `unsupported convert type 'nchar'` error.

### Changes

- Added `DM_NCHAR` constant
- Added `case DM_NCHAR:` to the `convert()` method's switch statement, treating it as a STRING type like NVARCHAR

### Related Issue

Closes #11688

### DIFF
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverter.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/dm/DmdbTypeConverter.java
@@ -67,6 +67,7 @@ public class DmdbTypeConverter implements TypeConverter<BasicTypeDefine> {
     public static final String DM_VARCHAR = "VARCHAR";
     public static final String DM_VARCHAR2 = "VARCHAR2";
     public static final String DM_NVARCHAR = "NVARCHAR";
+    public static final String DM_NCHAR = "NCHAR";
     public static final String DM_LONGVARCHAR = "LONGVARCHAR";
     public static final String DM_CLOB = "CLOB";
     public static final String DM_TEXT = "TEXT";
@@ -200,6 +201,11 @@ public class DmdbTypeConverter implements TypeConverter<BasicTypeDefine> {
             case DM_VARCHAR:
             case DM_VARCHAR2:
                 builder.sourceType(String.format("%s(%s)", DM_VARCHAR2, typeDefine.getLength()));
+                builder.dataType(BasicType.STRING_TYPE);
+                builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
+                break;
+            case DM_NCHAR:
+                builder.sourceType(String.format("%s(%s)", DM_NCHAR, typeDefine.getLength()));
                 builder.dataType(BasicType.STRING_TYPE);
                 builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
                 break;


DanielLeens (changes requested):
## ⚠️ Duplicate PR notice (please read first) — FOUR open PRs for the same bug

Before the detailed review: this PR is one of **four** currently-open PRs all fixing the same underlying issue (#11688, Dameng `NCHAR` support) in the same file (`DmdbTypeConverter.java`):

- **#11684** — "[Fix][Connector-V2] Support Dameng national character types in JDBC connector"
- **#11693** — "[Bug][Connector-V2] Support Dameng NCHAR type" — **already reviewed and APPROVED by this reviewer in a prior round; still open and unmerged.**
- **#11714 (this PR)** — adds a standalone `case DM_NCHAR:` block with its own `sourceType` label — correct.
- **#11716** — by the same author as this PR (waterWang), opened one minute after this one — merges `case DM_NCHAR:` into the existing `NVARCHAR` block via switch fallthrough, which makes every NCHAR column report `sourceType = "NVARCHAR(len)"` instead of `"NCHAR(len)"`. I traced this into `DamengCreateTableSqlBuilder.buildColumnSql()` and confirmed it will silently generate the wrong DDL type on a Dameng→Dameng auto-create-table sync (turns a fixed-length NCHAR column into a variable-length NVARCHAR column on the target). See Issue 1.

**Recommendation to maintainers: since #11693 is already approved and further along in the review process, that's the PR to merge.** #11714 is also a correct, minimal fix on its own technical merits (and clearly better than its sibling #11716 — see Issue 1), but there is no reason for the community to carry four parallel fixes for one bug. I'd suggest a maintainer or the respective authors close #11684, #11714, and #11716 as superseded by #11693 once #11693 merges — or, if #11693 has stalled for a reason I'm not aware of, #11714 is the next-best fix among the remaining three. This is not a nitpick — merging #11716 instead of (or in addition to) any of the others would reintroduce a subtler bug while appearing to fix the original one.

---

Dameng is Oracle-compatible and has a fixed-length national character type `NCHAR`, the counterpart of the already-supported variable-length `NVARCHAR`. `DmdbTypeConverter.convert()` never had a `case` for `NCHAR`, so any source table containing an `NCHAR` column fell through to the `default` branch and threw:

```
ErrorCode:[COMMON-17], ErrorDescription:['Dameng' unsupported convert type 'nchar' of 'test' to SeaTunnel data type.]
```

In a multi-table job this bubbles up as `COMMON-21` and fails the job before any row is read — see issue #11688, which even ships a minimal converter-level repro (`DmdbTypeConverter.INSTANCE.convert(...)` with `columnType("nchar(2)")`). This PR closes that gap by adding a dedicated `NCHAR` branch that maps the column to SeaTunnel's `STRING_TYPE`, exactly the same way `NVARCHAR` already is.

**Before**: reading any Dameng table with an `NCHAR` column throws `COMMON-17`/`COMMON-21` and the job never starts.
**After**: the `NCHAR` column is read as a normal SeaTunnel `STRING` field, with `columnLength` computed via `TypeDefineUtils.charTo4ByteLength(...)` and `sourceType` correctly reported as `NCHAR(len)`.

---

## 1. Code Change Review

### 1.1 Core Logic Analysis

Diff (this PR, #11714), in `DmdbTypeConverter.java`:

```java
public static final String DM_NCHAR = "NCHAR";   // new constant, next to DM_NVARCHAR
...
case DM_NCHAR:
    builder.sourceType(String.format("%s(%s)", DM_NCHAR, typeDefine.getLength()));
    builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
    break;
case DM_NVARCHAR:
    builder.sourceType(String.format("%s(%s)", DM_NVARCHAR, typeDefine.getLength()));
    builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
    break;
```

**Runtime path — this is the normal read path, not a recovery/edge path.** `DmdbTypeConverter.INSTANCE.convert(typeDefine)` is called from `DamengCatalog.java:185` (`return DmdbTypeConverter.INSTANCE.convert(typeDefine);`), which is invoked from `DamengCatalog.getTable(...)` (`DamengCatalog.java:223` and callers) — i.e. every time SeaTunnel discovers a Dameng table's schema for a normal source read, whether single-table or multi-table CDC/JDBC config. Any table with an `NCHAR` column hits this switch on every job start. So this is squarely mainline behavior, not a boundary condition.

**Key finding — why #11714's approach is correct and #11716's is not.** I compared this against the sibling PR #11716, which instead does:

```java
case DM_NCHAR:
case DM_NVARCHAR:
    builder.sourceType(String.format("%s(%s)", DM_NVARCHAR, typeDefine.getLength()));
    ...
```

This *looks* like a harmless simplification (merge two "equivalent" cases the way `CHAR`/`CHARACTER` and `VARCHAR`/`VARCHAR2` are already merged a few lines above), but it isn't equivalent. `CHAR`/`CHARACTER` and `VARCHAR`/`VARCHAR2` genuinely are synonyms in Dameng's SQL dialect (same underlying storage type, just two spellings), which is why the existing code correctly folds them into one `sourceType` label. `NCHAR` vs `NVARCHAR`, however, is a **fixed-length vs variable-length** distinction — exactly analogous to `CHAR` vs `VARCHAR`, which the converter correctly keeps as two separate cases with two separate `sourceType` labels. Issue #11688 says this explicitly: "`NCHAR` is a standard Dameng type — the fixed-length national character type, and the counterpart of the already supported variable-length `NVARCHAR`." Folding `NCHAR` into `NVARCHAR`'s case is the same category of bug as folding `CHAR` into `VARCHAR`'s case would be.

I verified this isn't just a style opinion — it's provably wrong against this repo's own test convention. `DmdbTypeConverterTest.java` (existing, `dev` branch) asserts for every string type that `column.getSourceType().toLowerCase()` round-trips back to the original `columnType` string, e.g. `testConvertChar()` (`DmdbTypeConverterTest.java:291-347`) and `testNvarchar()` (`DmdbTypeConverterTest.java:349-363`):
```java
Assertions.assertEquals(typeDefine.getColumnType(), column.getSourceType().toLowerCase());
```
Under #11716's implementation, converting `columnType("nchar(2)")` produces `sourceType = "NVARCHAR(2)"`, so an equivalent test (`"nchar(2)"` vs `"nvarchar(2)"`) would fail outright. Under #11714's implementation it produces `sourceType = "NCHAR(2)"` and passes.

**This is more than cosmetic.** I traced where `sourceType` is consumed downstream: `DamengCreateTableSqlBuilder.buildColumnSql()` (`DamengCreateTableSqlBuilder.java:129-142`):
```java
} else if (StringUtils.equals(DatabaseIdentifier.DAMENG, sourceCatalogName)
        && StringUtils.isNotEmpty(column.getSourceType())) {
    columnType = column.getSourceType();
}
```
When the *source* catalog is Dameng (a Dameng→Dameng, or Dameng-source-to-any-sink-that-reuses-Dameng-DDL sync with schema auto-creation, e.g. `schema_save_mode = CREATE_SCHEMA_WHEN_NOT_EXIST`), the auto-generated `CREATE TABLE` statement uses `column.getSourceType()` **verbatim** as the SQL column type. Under #11716, an `NCHAR(2)` source column would silently produce `CREATE TABLE ... "col" NVARCHAR(2) ...` on the target — a real, silent schema-correctness bug (fixed-length national char becomes variable-length national char on the target table, with no warning). #11714 does not have this problem: it always reports the column's own real type name.

**Minor**: #11716's diff also strips the file's trailing newline (`-}` / `+}` with `\ No newline at end of file`) — I confirmed against the raw file content at its head commit that the file now genuinely ends without a trailing `\n`. That's very likely to fail `./mvnw spotless:check` once CI actually runs (see CI section). #11714 does not have this problem.

### 1.2 Compatibility

Fully compatible. The change only adds a new `case` label and a new `public static final String` constant to an existing `switch`; nothing existing is removed, renamed, or has its default behavior changed. No config option, public API, or serialization format is touched. `NCHAR` columns previously always threw `COMMON-17`, so there is no existing behavior being altered for that type — this is a pure additive fix.

Negligible. One additional `case` branch in a `switch` statement evaluated once per column during schema discovery; no new I/O, allocation patterns, locking, or retries introduced.

N/A — no new logging or exception handling is introduced; the change removes one error path (the `default: throw CommonError.convertToConnectorTypeError(...)` branch previously hit for `NCHAR`) without altering how other unsupported types are still reported.

## 2. Code Quality

The change follows the existing style of the surrounding `switch` (constant naming, `sourceType` formatting, `columnLength` computation reused from `TypeDefineUtils.charTo4ByteLength`). It's a 5-line mechanical addition; per this repo's comment guidance ("comments for important methods... complex logic"), no new comment is strictly required here since the pattern is self-evident from the six other near-identical `case` blocks immediately around it. No concerns.

**Gap — no test added, and one is easy to add.** Neither this PR nor #11716 touches `DmdbTypeConverterTest.java`. That file already has the exact pattern needed (see `testConvertChar()` / `testNvarchar()` above) — a `testConvertNchar()` following the same shape (`columnType("nchar(2)")` → assert `STRING_TYPE`, `columnLength == 8`, `sourceType.toLowerCase() == "nchar(2)"`) would take a few lines and would have caught the #11716-style regression automatically. The linked issue #11688 explicitly calls out that this class of gap "likely went unnoticed because `JdbcDmIT` covers `CHAR`, `CHARACTER`, `VARCHAR` and `VARCHAR2` but no national character type at all" — i.e. the reporter is asking for exactly this kind of test. I'd ask for it before merge (see Issue 2).

Stability rating: **N/A** — this PR does not modify any UT/E2E test code, so the mandatory flaky-test analysis doesn't apply (no test file is touched by this diff).

### 2.3 Docs

No dedicated Dameng type-mapping doc page exists in `docs/en`/`docs/zh` (I checked `docs/en/connectors/source/Jdbc.md`, `docs/en/connectors/sink/Jdbc.md`, and the connector-jdbc changelog — none carry a Dameng-specific type table), and the sibling `NVARCHAR`/`VARCHAR2` support wasn't documented there either. So this is consistent with existing precedent and not a regression introduced by this PR; not a blocker.

## 3. Architectural soundness

Sound and minimal. It's a single additive `case` in a dialect-specific `TypeConverter`, exactly the intended extension point for this kind of gap (see the prior, structurally identical fix for `NVARCHAR` itself, and the linked #10635 for `NVARCHAR2`-class gaps in other connectors). No leakage into engine/core, no new SPI surface.

## 4. CI status

Both PRs' Apache-side "Build" check shows `action_required` / fail with 0s duration. I checked the actual check-run output (not just the rolled-up status): the message is `"Workflow run detection failed... If you did not enable GitHub Actions in your forked repository, please enable it..."` — i.e. this is an infra/setup issue (GitHub Actions isn't enabled in waterWang's fork), not a code failure. I confirmed no workflow runs exist at all on `waterWang/seatunnel` for either branch (`fix/dmdb-add-nchar-support` for this PR). davidzollo already left a comment on #11716 asking the author to enable Actions in their fork; the same applies here. Until that happens there is no real CI signal for either PR (in particular, no confirmation of whether #11716's stripped trailing newline actually trips `spotless:check`, though on inspection of the codebase's Google-Java-Format/AOSP convention it's likely to).

`gh api compare/dev...<head>` shows both branches are `diverged` (`ahead_by: 1, behind_by: 2` from current `dev`), so a rebase will be needed regardless once CI is enabled, but the divergence is small (2 commits) and unrelated to this change.

## 5. Issue Summary

| 1 | High (applies to sibling PR #11716, not to this PR's code) | `DmdbTypeConverter.java` convert() switch (compare #11714 vs #11716); consumed at `DamengCreateTableSqlBuilder.java:129-142` | #11716's `case DM_NCHAR: case DM_NVARCHAR:` fallthrough mislabels every NCHAR column's `sourceType` as `NVARCHAR(len)`, which `DamengCreateTableSqlBuilder` uses verbatim to auto-generate DDL on Dameng→Dameng sync, silently turning a fixed-length NCHAR column into a variable-length NVARCHAR column on the target table. This PR (#11714) does **not** have this bug — its dedicated `case DM_NCHAR:` block reports the correct type. | No |
| 2 | Medium | `seatunnel-connectors-v2/connector-jdbc/src/test/java/.../DmdbTypeConverterTest.java` | No unit test added for the new `NCHAR` branch, even though the sibling test class already has an identical pattern for `CHAR`/`NVARCHAR` that a `testConvertNchar()` could mirror in a few lines, and the linked issue explicitly flags the missing national-character-type test coverage as the root cause this went unnoticed. | No |
| 3 | Low / process | PRs #11684, #11693, #11714, #11716 | **Four** open PRs fix the identical issue (#11688) in the identical file. #11693 is already reviewed and APPROVED (prior round, by this reviewer) and, like #11684, also ships a unit test and doc updates that this PR (#11714) lacks. The community should consolidate on #11693 rather than review/merge more than one of these four independently. | No |
| 4 | Low | CI (fork Actions) | This PR has no working "Build" check because GitHub Actions isn't enabled in the contributor's fork (confirmed: zero workflow runs on `waterWang/seatunnel`). CI must go green before merge per repo policy, if this PR is still pursued despite Issue 3. | No |

## 6. Merge Recommendation

**Not recommended for merge — in favor of consolidating on #11693.**

On its own technical merits, this PR's `case DM_NCHAR:` implementation is correct (unlike sibling #11716's), but it is the least complete of the four open fixes for the same issue: it has no unit test, while #11684 additionally updates the E2E test/fixture and #11693 additionally updates `docs/en`/`docs/zh` — and #11693 is already through a full review round with an APPROVED verdict. There is no technical reason to invest further review cycles polishing this PR (adding a test, enabling CI, etc.) when a more complete, already-approved alternative exists for the same one-line bug fix.

Blockers (must-fix, by issue number):
- **Issue 3**: this PR should be closed by the author in favor of #11693 (or, if a maintainer has a specific reason to prefer this implementation over #11693's, that should be stated explicitly rather than letting four duplicate PRs sit open indefinitely).

If, despite the above, this specific PR is the one carried forward instead of #11693: Issue 2 (add a unit test) and Issue 4 (enable CI) would still need to be resolved before merge, but that would be a deliberate choice, not the default path.

Thanks for tracking down and fixing this, waterWang — appreciate the reference back to issue #11688 and the clear repro in the PR description. The fix itself is right; it's the duplication across four PRs that needs a maintainer decision, not this PR's code.

davidzollo:

```

```

nzw921rx:
PR #11693 merged, Thank you for your efforts.

PR #11715, opened 2026-08-09
[Bug][connector-jdbc] Support YashanDB VARCHAR2 and NVARCHAR2 types in YashanDbTypeConverter

### LINKED ISSUE #11685: [Bug] [connector-jdbc] YashanDB VARCHAR2 and NVARCHAR2 types are not supported by the type converter
### Search before asking

### What happened

`YashanDbTypeConverter` has cases for `CHAR`/`VARCHAR` and `NCHAR`/`NVARCHAR`, but not for the Oracle-compatible synonyms `VARCHAR2` and `NVARCHAR2`. Reading a YashanDB table that contains such columns falls through to the `default` branch and throws:

```
ErrorCode:[COMMON-17], ErrorDescription:['YashanDB' unsupported convert type 'VARCHAR2' of 'test' to SeaTunnel data type.]
ErrorCode:[COMMON-17], ErrorDescription:['YashanDB' unsupported convert type 'NVARCHAR2' of 'test' to SeaTunnel data type.]
```

This is inconsistent with the connector's own catalog layer: `YashanDbCatalog#SELECT_COLUMNS_SQL` already expects both types and even encodes which length semantics each one follows.

https://github.com/apache/seatunnel/blob/dev/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/catalog/yashandb/YashanDbCatalog.java#L59-L60

```sql
WHEN cols.data_type IN ('VARCHAR', 'VARCHAR2', 'CHAR')  THEN '(' || cols.data_length || ')'
WHEN cols.data_type IN ('NVARCHAR2', 'NCHAR')           THEN '(' || cols.char_length || ')'
```

So the catalog knows these columns exist and knows `VARCHAR2` follows byte length while `NVARCHAR2` follows character length, but the converter rejects them.

Since `VARCHAR2` is the primary Oracle string type and YashanDB is Oracle-compatible, any table using it currently cannot be read.

### SeaTunnel Version

dev

### SeaTunnel Config

```conf

YashanDbTypeConverter.INSTANCE.convert(
    BasicTypeDefine.builder()
        .name("test").columnType("VARCHAR2(100)").dataType("VARCHAR2").length(100L).build());
```

### Running Command

```shell
./mvnw test -pl seatunnel-connectors-v2/connector-jdbc -Dtest=YashanDbTypeConverterTest
```

### Error Exception

```log
ErrorDescription:['YashanDB' unsupported convert type 'VARCHAR2' of 'test' to SeaTunnel data type.]
```

Zeta

### Java or Scala Version

Java 8

### Screenshots

_No response_

### Code of Conduct

waterWang:
### Purpose

Add support for Oracle-compatible synonyms `VARCHAR2` and `NVARCHAR2` in `YashanDbTypeConverter`. These types are not handled by the converter, causing an `unsupported convert type` error.

### Changes

- Added `case VARCHAR2:` to the existing `CHAR`/`VARCHAR` case block
- Added `case NVARCHAR2:` to the existing `NCHAR`/`NVARCHAR` case block

### Related Issue

Closes #11685

### DIFF
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/yashandb/YashanDbTypeConverter.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/yashandb/YashanDbTypeConverter.java
@@ -213,11 +213,13 @@ public class YashanDbTypeConverter implements TypeConverter<BasicTypeDefine> {
                 // ====================== Character types ======================
             case CHAR:
             case VARCHAR:
+            case VARCHAR2:
                 builder.dataType(BasicType.STRING_TYPE);
                 builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
                 break;
             case NCHAR:
             case NVARCHAR:
+            case NVARCHAR2:
                 builder.dataType(BasicType.STRING_TYPE);
                 builder.columnLength(
                         TypeDefineUtils.doubleByteTo4ByteLength(typeDefine.getLength()));


DanielLeens (changes requested):
## Duplicate PR notice (read this first)

This PR is a near-exact duplicate of **#11717** ("[YashanDB][TypeConverter] Add VARCHAR2/NVARCHAR2 data type support to YashanDbTypeConverter"), opened by the same author (@waterWang) 57 seconds after this one, against the same file, fixing the same linked issue (#11685). I fetched both diffs directly and they are effectively byte-identical:

```diff
             case CHAR:
             case VARCHAR:
+            case VARCHAR2:
                 builder.dataType(BasicType.STRING_TYPE);
                 builder.columnLength(TypeDefineUtils.charTo4ByteLength(typeDefine.getLength()));
                 break;
             case NCHAR:
             case NVARCHAR:
+            case NVARCHAR2:
                 builder.dataType(BasicType.STRING_TYPE);
                 builder.columnLength(
                         TypeDefineUtils.doubleByteTo4ByteLength(typeDefine.getLength()));
```

The only textual difference is that #11717 additionally strips the trailing newline at end-of-file (`}` with `\ No newline at end of file` in its diff), which is a minor regression #11715 does not have. Both PRs target `dev`, touch only `YashanDbTypeConverter.java`, and neither PR references the other.

**The community should merge at most one of these** — recommend closing #11717 as a duplicate of this one (#11715 was opened first) once a corrected version lands, and asking the author to consolidate future work into a single branch/PR. This does not change the underlying correctness verdict below, which is identical for both PRs since the changed code is the same.

YashanDB is Oracle-syntax-compatible and commonly exposes `VARCHAR2`/`NVARCHAR2` as synonyms for `VARCHAR`/`NVARCHAR` in its metadata (`ALL_TAB_COLUMNS.DATA_TYPE` etc.), the same way Oracle, DM (Dameng), and other Oracle-compatible dialects already handled in this codebase do. `YashanDbTypeConverter.convert()` previously only matched `CHAR`/`VARCHAR` and `NCHAR`/`NVARCHAR`, so any YashanDB source table containing a `VARCHAR2` or `NVARCHAR2` column falls into the `default` branch and throws `CommonError.convertToSeaTunnelTypeError(...)` ("unsupported convert type") during schema discovery — this is issue #11685. The PR's intent is to add those two synonyms to the existing `STRING_TYPE` case blocks.

**Fix approach**: append `case VARCHAR2:` under the existing `CHAR`/`VARCHAR` block, and `case NVARCHAR2:` under the existing `NCHAR`/`NVARCHAR` block.

**One-sentence summary**: The PR correctly identifies the missing type-mapping gap and the right two case blocks to extend, but as written it will not compile because `VARCHAR2`/`NVARCHAR2` are used as bare identifiers without ever being declared as constants — this is a hard blocker for both #11715 and #11717.

**Before/after example**: Reading a YashanDB table `t1(col1 VARCHAR2(100))` via the JDBC source today throws `unsupported convert type` and fails the job at schema-discovery time. After a *correctly implemented* fix, `col1` would map to SeaTunnel `STRING_TYPE` with `columnLength` derived from `TypeDefineUtils.charTo4ByteLength(100)`, the same as an equivalent `VARCHAR(100)` column. As submitted, however, this PR would instead fail to compile the `connector-jdbc` module entirely, so no user ever reaches "before" or "after" — the whole connector build breaks.

## 1. Code Change Review

### 1.1 Core Logic Analysis

**Runtime path**: This is a normal-path fix, not a recovery/boundary path. `YashanDbCatalog.getColumn(...)` calls `new YashanDbTypeConverter().convert(typeDefine)` at `seatunnel-connectors-v2/connector-jdbc/.../catalog/yashandb/YashanDbCatalog.java:178`, which is invoked whenever SeaTunnel discovers/loads table schema for a YashanDB source (`CatalogTable` construction, e.g. `auto.create.table` / schema inference). Any table with a `VARCHAR2` or `NVARCHAR2` column hits `YashanDbTypeConverter.convert()` on every normal job run that reads that table's schema, so the bug being fixed is real and commonly hit for Oracle-compatible schemas migrated to YashanDB.

**Critical finding — the diff does not compile.** `yashanType` in `convert()` is a plain `String` (`String yashanType = typeDefine.getDataType().toUpperCase();`), and the surrounding `switch (yashanType)` is a switch-on-`String`, which in Java requires every `case` label to be a compile-time constant `String` expression (a literal, or a `static final String` in scope). I pulled the full head-of-PR file via the GitHub API (`YashanDbTypeConverter.java` at commit `45918c31fb2d69e4ce47ce3252e2bf44c84bb422`, 523 lines) and grepped the whole file for `VARCHAR2` — the two `case` lines added by this PR are the **only** occurrences:

```
216:            case VARCHAR2:
222:            case NVARCHAR2:
```

There is no `public static final String VARCHAR2 = "VARCHAR2";` (or `NVARCHAR2`) declared anywhere in this class, no static import that brings such a constant into scope, and no such symbol in any of the file's existing imports (`Column`, `PhysicalColumn`, `BasicTypeDefine`, `TypeConverter`, `ArrayType`, `BasicType`, `DecimalType`, `LocalTimeType`, `PrimitiveByteArrayType`, `SqlType`, `VectorType`, `CommonError`, `TypeDefineUtils`, `DatabaseIdentifier`). `SqlType` (the enum actually imported) has no `VARCHAR2`/`NVARCHAR2` constant either — it's SeaTunnel's own logical type enum, unrelated to source-DB type names. `javac` will report `cannot find symbol` for both `VARCHAR2` and `NVARCHAR2`.

This exact "define an Oracle-synonym constant, then `case` on it" pattern already exists elsewhere in this very module and is what this PR should have followed. Two sibling dialects handle the identical `VARCHAR2` synonym correctly:

- `OracleTypeConverter.java:53-54`:
  ```java
  public static final String ORACLE_VARCHAR2 = "VARCHAR2";
  public static final String ORACLE_NVARCHAR2 = "NVARCHAR2";
  ```
  then `case ORACLE_VARCHAR2:` / `case ORACLE_NVARCHAR2:` at lines 183/188.
- `DmdbTypeConverter.java:68`: `public static final String DM_VARCHAR2 = "VARCHAR2";`, used at `case DM_VARCHAR2:` (line 201).

The fix here needs the same treatment, e.g.:

```java
public static final String VARCHAR2 = "VARCHAR2";
public static final String NVARCHAR2 = "NVARCHAR2";
...
case CHAR:
case VARCHAR:
case VARCHAR2:
    ...
case NCHAR:
case NVARCHAR:
case NVARCHAR2:
    ...
```

**Why CI hasn't caught this**: Both PRs' "Build" check shows `ACTION_REQUIRED`/`fail` with 0 duration (`gh pr checks 11715` → `Build fail 0s`), which is the standard "workflow needs approval to run in a first-time contributor's fork" state, not an actual completed build. Maintainer @davidzollo already left the standard bot comment on #11717 asking the author to enable Actions in their fork (`2026-08-09T01:15:25Z`) — the real build has simply never executed yet for either PR. So this compile break has not been surfaced by CI and would only be caught once workflows are enabled/re-run, or by source review (which is what this review is doing).

`reconvert()` (the sink-direction write path, lines ~313-478) is correctly left untouched — YashanDB always writes canonical `VARCHAR`/`NCHAR` etc. back out (see the existing `STRING` case at line ~444 mapping to `VARCHAR`), matching the pre-existing pattern for `CHAR`/`NCHAR`, so no reverse-direction gap was introduced by omission.

### 1.2 Compatibility

Fully backward compatible **in intent**: this only widens the `switch` to accept two additional source-side type-name synonyms; it does not change any existing case, default value, config option, or serialization format, and cannot affect any table that doesn't use `VARCHAR2`/`NVARCHAR2`. However, since the change does not compile, compatibility is moot until Issue 1 below is fixed — an uncompilable diff cannot be assessed as "compatible," it simply cannot ship.

None. This is a pure `switch`-arm addition; no new allocations, I/O, locking, or retries are introduced on the hot path once it compiles.

Not applicable to the two-line diff itself. The existing `default: throw CommonError.convertToSeaTunnelTypeError(...)` path (unchanged) still provides adequate context (dialect identifier, source type name, column name) for any genuinely unsupported type.

## 2. Code Quality

The change follows the file's existing section-comment convention (`// ====================== Character types ======================`) and doesn't need its own comment — it's a one-line addition to an existing, already-commented block, consistent with how `CHAR`/`VARCHAR` and `NCHAR`/`NVARCHAR` are handled. No new methods, fields, or classes are introduced that would require Javadoc. The only defect is the missing constant declarations discussed in Issue 1 (a correctness defect, not a comment-completeness defect).

**Gap**: Neither PR touches `YashanDbTypeConverterTest.java` (`seatunnel-connectors-v2/connector-jdbc/src/test/java/.../yashandb/YashanDbTypeConverterTest.java`), which already exists and has dedicated per-type-group test methods (e.g. `testConvertIntegerTypes`) following a clear existing pattern. No test asserts that `VARCHAR2`/`NVARCHAR2` now map to `BasicType.STRING_TYPE` with the expected `columnLength`. Given this is a one-line-per-branch fix, a couple of `@Test` cases mirroring the existing `CHAR`/`VARCHAR` test would have been trivial to add and — importantly — would have caught the compile error immediately on `mvn test` for this module, since the whole module fails to compile once an unresolvable symbol is introduced.

Since no test code is touched by this PR, the mandatory flaky-test/stability analysis is not applicable; there is nothing to rate for flakiness here (the class of risk is "doesn't compile," not "flaky"). I'm not assigning a Stable/Risk present/High risk label since criterion 8's trigger (PR touches UT/E2E code) isn't met — but note the *absence* of new tests is itself flagged as Issue 2 below because the compile error would have been self-evidently caught by any test run.

### 2.3 Docs

`docs/en/connectors/source/Jdbc.md` and `docs/zh/connectors/source/Jdbc.md` contain only a driver-reference row for YashanDB (`docs/en/connectors/source/Jdbc.md:330`, listing driver class/URL/maven coordinate) and no per-type support matrix for YashanDB specifically, so there is no existing documented type list that becomes stale or needs updating for this change. No doc update is required for this fix.

## 3. Architectural soundness

The chosen approach (extend the existing `switch` fall-through groups) is architecturally correct and matches the established pattern in `OracleTypeConverter`/`DmdbTypeConverter`/`XuguTypeConverter` for handling Oracle-compatible type synonyms. The only defect is an implementation mistake (undeclared constants), not a design problem — once the constants are declared, this is exactly the right fix in exactly the right place.

## 4. Issue Summary

| # | Location | Problem | Risk | Suggested fix | Severity | Raised by another reviewer |
|---|----------|---------|------|----------------|----------|------------------------------|
| 1 | `YashanDbTypeConverter.java:216,222` (both #11715 and #11717) | `case VARCHAR2:` / `case NVARCHAR2:` reference undeclared identifiers — no `static final String VARCHAR2`/`NVARCHAR2` exists anywhere in the file, imports, or superclasses. Switch-on-`String` requires a compile-time constant `String` case label. | The `connector-jdbc` module fails to compile (`javac: cannot find symbol`); this breaks the build for anyone building `connector-jdbc` from this branch, and CI has not yet caught it only because the fork's Build workflow requires manual approval and hasn't actually run. | Declare `public static final String VARCHAR2 = "VARCHAR2";` and `public static final String NVARCHAR2 = "NVARCHAR2";` (following the `ORACLE_VARCHAR2`/`DM_VARCHAR2` naming precedent in `OracleTypeConverter.java:53-54` and `DmdbTypeConverter.java:68`), then use those constants as the case labels. | High | No |
| 2 | Duplicate PR: #11715 vs #11717 | Same author opened two PRs 57 seconds apart with an essentially identical diff against the same file, closing the same issue (#11685). #11717 additionally introduces a missing-trailing-newline regression at EOF. | If both are merged (or reviewed independently without cross-referencing), one becomes redundant/conflicting; wasted reviewer effort; git history noise. | Consolidate to a single PR — recommend keeping #11715 (opened first, no EOF regression) and closing #11717 as duplicate, after Issue 1 is fixed in the kept PR. | Medium | No |
| 3 | `YashanDbTypeConverterTest.java` (not touched by either PR) | No unit test added for the new `VARCHAR2`/`NVARCHAR2` mappings, despite an existing, easily-extensible per-type-group test class already covering `CHAR`/`VARCHAR`/`NCHAR`/`NVARCHAR`. | Regression risk for future refactors of this switch, and — concretely in this case — a trivial test would have caught Issue 1 immediately via `mvn test`. | Add 1-2 `@Test` methods asserting `INSTANCE.convert(...)` for `VARCHAR2`/`NVARCHAR2` columnType inputs returns `BasicType.STRING_TYPE` with the expected `columnLength`, mirroring the existing `CHAR`/`VARCHAR` assertions. | Low | No |

## 5. Merge Recommendation

**Not recommended for merge** (applies to both #11715 and #11717 — the same defect is present in both, since the changed code is byte-identical).

- Issue 1: fix the compile error by declaring `VARCHAR2`/`NVARCHAR2` as `static final String` constants before using them as `switch` case labels. Without this, the diff cannot pass a real build.
- Issue 2: the duplicate-PR situation must be resolved before merge — only one of #11715/#11717 should proceed; the other should be closed as a duplicate once the surviving PR is corrected.

- Issue 3: add unit test coverage for the two new type mappings in `YashanDbTypeConverterTest.java`.

Once Issue 1 is fixed and the constants follow the existing `ORACLE_VARCHAR2`/`DM_VARCHAR2`-style naming convention used elsewhere in this same module, and once the duplicate is closed, this is a small, low-risk, well-targeted fix for a real bug (#11685) and should be quick to re-review. Thanks for tackling this, @waterWang — the type gap you identified is real and the fix location is exactly right, it just needs the constants declared before this can compile and merge.

davidzollo:

```

```

nzw921rx:
PR #11874 Merged. Thank you for your efforts.

PR #2921, opened 2022-09-27
[Bug][connector-file-base] Fix source split assigning reader to negative number

hailin0:
## Purpose of this pull request

example:
```java
Math.abs(Integer.MIN_VALUE) % numReaders; 
```

## Check list

* [ ] Code changed are covered with tests, or it does not need tests for reason:
* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/incubator-seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/incubator-seatunnel/tree/dev/docs

PR #11721, opened 2026-08-09
[Fix][API] Fix negative queue index when primary key hash is Integer.MIN_VALUE

SEPURI-SAI-KRISHNA:
### Purpose of this pull request

Closes #11720.

`MultiTableSinkWriter.write` routed rows with `Math.abs(object.hashCode()) % blockingQueues.size()`. `Math.abs(Integer.MIN_VALUE)` returns `Integer.MIN_VALUE` — still negative, since `-Integer.MIN_VALUE` overflows `int` — so any primary key hashing to `Integer.MIN_VALUE` produced a negative index and failed the sink writer with `IndexOutOfBoundsException` whenever the queue count was not a power of two.

An `INT` primary key of `-2147483648`, a `BIGINT` primary key of `Long.MIN_VALUE`, and any string primary key hashing to `Integer.MIN_VALUE` all hit it. Because `multi_table_sink_replica` defaults to `1` and powers of two are also safe, this only surfaces once that option is tuned to a non-power-of-two value — but it affects every multi-table sink, since the routing is shared code in `seatunnel-api`.

The fix clears the sign bit instead of negating:

```java
index = (object.hashCode() & Integer.MAX_VALUE) % blockingQueues.size();
```

This is the same idiom #2921 introduced for `FileSourceSplitEnumerator.getSplitOwner`, and it is already used at 12 other hash-routing sites: nine spell it `(hash & Integer.MAX_VALUE) % n` (Paimon, HBase, Fluss, MongoDB, Iceberg, JDBC, Google Bigtable, the Kafka sink's `MessageContentPartitioner`, and the Hazelcast metrics store) and three use the equivalent `((hash * 31) & 0x7FFFFFFF) % n` form (Kafka source, Pulsar, RocketMQ). `MultiTableSinkWriter` was the last site still using the unsafe form.

**Scope of the change.** This is a general fix, not a special case for `Integer.MIN_VALUE`. `h & Integer.MAX_VALUE == h` for every non-negative hash, so those keys route exactly as before. For a negative hash the old expression computed `-h` while the new one computes `h + 2^31`, so **every** row whose primary key has a negative hash code may now land on a different queue than it did before — not only the `Integer.MIN_VALUE` case in the title. That is safe rather than incidental: the reason is set out under "user-facing change" below, and is now covered by a test that asserts the exact resulting index.

The class Javadoc, which documented the old formula, has been updated to match.

### Does this PR introduce _any_ user-facing change?

Only in a narrow and unobservable sense, so: no.

[...]

PR #11725, opened 2026-08-09
fix: use bitwise AND instead of Math.abs to avoid Integer.MIN_VALUE overflow in hash routing

### LINKED ISSUE #11720: [Bug] [API] Multi-table sink throws IndexOutOfBoundsException when a primary key hashes to Integer.MIN_VALUE
### Search before asking

### What happened

`MultiTableSinkWriter.write` routes each row to a queue using:

```java
index = Math.abs(object.hashCode()) % blockingQueues.size();
```

`Math.abs(Integer.MIN_VALUE)` returns `Integer.MIN_VALUE` — it is still negative, because `-Integer.MIN_VALUE` is not representable as an `int`. When the queue count is not a power of two, the modulo result is therefore negative and `blockingQueues.get(index)` throws, failing the sink writer:

```
java.lang.IndexOutOfBoundsException: Index -2 out of bounds for length 3
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at org.apache.seatunnel.api.sink.multitablesink.MultiTableSinkWriter.write(MultiTableSinkWriter.java:604)
```

Any primary key value whose `hashCode()` is `Integer.MIN_VALUE` triggers it. All of these do:

| Primary key type | Value |
|---|---|
| `INT` | `-2147483648` |
| `BIGINT` | `-9223372036854775808` (`Long.MIN_VALUE`) |
| `STRING` | any string hashing to `Integer.MIN_VALUE`, e.g. `polygenelubricants` |

Whether it actually throws depends on the queue count, which is `multi_table_sink_replica`:

| `multi_table_sink_replica` | resulting index |
|---|---|
| 1, 2, 4, 8, 16 (powers of two) | 0 — safe |
| 3 | -2 — throws |
| 5 | -3 — throws |
| 6 | -2 — throws |
| 7 | -2 — throws |
| 9 | -2 — throws |
| 10 | -8 — throws |
| 12 | -8 — throws |

Since `multi_table_sink_replica` defaults to `1`, a default configuration is unaffected — the bug only surfaces once the option is tuned to a non-power-of-two value, which is presumably why it has gone unnoticed. It affects every multi-table sink, because the routing lives in `seatunnel-api` and is shared by all of them.

This is the same defect class that was fixed for the file source split enumerator in #2921, which replaced `Math.abs(tp.hashCode()) % numReaders` with `(tp.hashCode() & Integer.MAX_VALUE) % numReaders`. That idiom is now used in 12 places across the codebase (Paimon, Kafka, DynamoDB, HBase, Fluss, Typesense, TiDB CDC, MongoDB, Iceberg, JDBC, Easysearch, and the Hazelcast metrics store). `MultiTableSinkWriter` is the last remaining site still using the unsafe form.

### SeaTunnel Version

dev

### SeaTunnel Config

```conf
sink {
  Jdbc {
    # ... any multi-table sink with a primary key ...
    multi_table_sink_replica = 3
  }
}


with a source table whose primary key contains the value -2147483648.
```

### Running Command

```shell
./bin/seatunnel.sh --config ./config/repro.conf -e local
```

### Error Exception

```log
java.lang.IndexOutOfBoundsException: Index -2 out of bounds for length 3
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at org.apache.seatunnel.api.sink.multitablesink.MultiTableSinkWriter.write(MultiTableSinkWriter.java:604)
```

Zeta (engine-independent; the routing lives in seatunnel-api)

### Java or Scala Version

Java 11

### Screenshots

_No response_

### Code of Conduct

waterWang:
### Problem

`MultiTableSinkWriter.write` routes each row to a queue using:

```java
index = Math.abs(object.hashCode()) % blockingQueues.size();
```

`Math.abs(Integer.MIN_VALUE)` returns `Integer.MIN_VALUE` — still negative, because `-Integer.MIN_VALUE` is not representable as an `int`. When the queue count is not a power of two, the modulo result is negative and `blockingQueues.get(index)` throws `IndexOutOfBoundsException`.

### Root Cause

`Math.abs(int)` overflows for `Integer.MIN_VALUE` (`0x80000000`), which is its own negation. This is a well-known Java integer overflow pitfall.

### Fix

Replace `Math.abs(hashCode())` with `(hashCode() & Integer.MAX_VALUE)`, which maps `Integer.MIN_VALUE` to `0` and preserves all other values without overflow. This follows the same pattern already applied in the file source split enumerator (PR #2921) and used in 12 other places across the codebase.

### Affected Versions

Any `multi_table_sink_replica` value that is not a power of two triggers the crash when a primary key hashes to `Integer.MIN_VALUE` (e.g. `INT` value `-2147483648`, `BIGINT` value `-9223372036854775808`, or any string with `hashCode() == Integer.MIN_VALUE`).

Closes #11720

### DIFF
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java
@@ -545,7 +545,7 @@ public class MultiTableSinkWriter
      *
      * <ul>
      *   <li>If the table's primary key information is present and the primary key field value is
-     *       non-null, the row is routed by {@code Math.abs(primaryKeyValue.hashCode()) %
+     *       non-null, the row is routed by {@code (primaryKeyValue.hashCode() & Integer.MAX_VALUE) %
      *       queueSize}, guaranteeing that rows with the same primary key always go to the same
      *       queue for ordered delivery.
      *   <li>If the table's primary key information is present but the actual field value is {@code
@@ -599,7 +599,7 @@ public class MultiTableSinkWriter
             Object object = element.getField(primaryKey.get());
             int index = 0;
             if (object != null) {
-                index = Math.abs(object.hashCode()) % blockingQueues.size();
+                index = (object.hashCode() & Integer.MAX_VALUE) % blockingQueues.size();
             }
             BlockingQueue<MultiTableWriterRunnable.QueueElement> queue = blockingQueues.get(index);
             offerQueueElement(queue, MultiTableWriterRunnable.rowRequest(element));


davidzollo:

```

```

DanielLeens (changes requested):

`MultiTableSinkWriter.write()` routes each row with a primary key to one of `blockingQueues.size()` async queues using `Math.abs(primaryKeyValue.hashCode()) % blockingQueues.size()`. `Math.abs(Integer.MIN_VALUE)` is a well-known Java overflow trap: it returns `Integer.MIN_VALUE` itself (still negative), because `+2147483648` has no positive `int` representation. When a primary key's `hashCode()` happens to equal `Integer.MIN_VALUE` (e.g. `INT` value `-2147483648`, `BIGINT` value `Long.MIN_VALUE`, or specific strings), and `multi_table_sink_replica` is set to anything other than 1 (the default) or another power of two, the resulting negative index is fed straight into `blockingQueues.get(index)`, throwing `IndexOutOfBoundsException` and crashing the sink. This PR (fixing #11720) replaces the unsafe `Math.abs(...)` with `hashCode() & Integer.MAX_VALUE`, which clears the sign bit and always yields a non-negative value, then reuses that with the ordinary `% queueSize`.

Welcome to the SeaTunnel community, and thanks for tracking this down with such a clear repro table in the issue — this is a nicely scoped, single-purpose fix.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Before/after (`seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java:602`, inside `write(SeaTunnelRow)`):

```java
// before
index = Math.abs(object.hashCode()) % blockingQueues.size();

// after
index = (object.hashCode() & Integer.MAX_VALUE) % blockingQueues.size();
```

The matching Javadoc on `write()` (lines 547-550) was updated in lockstep to describe the new formula — good attention to keeping the doc-comment and the code consistent.

**Key findings:**

- **The power-of-two question is the crux of this review, and the fix gets it right.** `hashCode() & Integer.MAX_VALUE` (i.e. `& 0x7FFFFFFF`) simply clears the sign bit of the 32-bit hash, mapping every possible `int` — including `Integer.MIN_VALUE` — into the non-negative range `[0, Integer.MAX_VALUE]`. Once the left operand of `%` is guaranteed non-negative, Java's `%` operator (which takes the sign of the dividend) is guaranteed to return a value in `[0, n-1]` for **any** positive `n`, not just powers of two. This is fundamentally different from — and stronger than — the classic `hash & (n - 1)` masking trick, which is only equivalent to `hash % n` when `n` is a power of two and would silently misroute for other sizes. I confirmed `blockingQueues.size()` (i.e. `queueSize`, ultimately the user-configurable `multi_table_sink_replica`, default `1`, see `SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA`) is an arbitrary positive integer with no power-of-two constraint anywhere in `MultiTableSink`/`MultiTableSinkFactory`. So the fix is correct for the full range of legal configurations, not just the values shown in the issue's repro table.
- **This is the established idiom in this codebase, not a one-off invention.** I grepped the `dev` branch and found the exact same `(x.hashCode() & Integer.MAX_VALUE) % n` pattern already in production use in at least 13 other split/partition enumerators: `JdbcSourceSplitEnumerator`, `HbaseSourceSplitEnumerator`, `AbstractSplitEnumerator` (Paimon and Iceberg), `MongodbSplitEnumerator`, `TiDBSourceSplitEnumerator`, `AmazonDynamoDBSourceSplitEnumerator`, `EasysearchSourceSplitEnumerator`, `TypesenseSourceSplitEnumerator`, `BigtableSourceSplitEnumerator`, `MessageContentPartitioner` (Kafka), `CatalogTableUtils` (CDC), and `HazelcastMetricsSnapshotStateStore`. After this patch, `MultiTableSinkWriter` was indeed the last remaining call site still using the unsafe `Math.abs(...)` form for this class of hash routing, exactly as the PR description states — I verified this claim directly rather than taking it at face value.
- **This is a rare-input, non-default-config edge case, not a main-path regression.** `multi_table_sink_replica` defaults to `1` (`SinkConnectorCommonOptions.java:33`); with `queueSize == 1`, `index` is always `0` under both the old and new formula, so default-configuration users never hit this code path's divergent behavior at all. The crash only manifests when a user (a) explicitly tunes `multi_table_sink_replica` to a value that isn't a power of two, and (b) a primary key's `hashCode()` happens to equal exactly `Integer.MIN_VALUE`. It's a real production crash (confirmed by the stack trace in #11720), but it is narrow, not something the "normal" default-config path exercises.

## 1.2 Compatibility Impact

- **No persisted-state / checkpoint impact.** I traced `MultiTableSink.createWriter()` / `restoreWriter()`: sub-writer state is keyed and restored purely by `SinkIdentifier` (`tableIdentifier` + `subtaskIndex * replicaNum + i`), never by the blocking-queue index computed in `write()`. `MultiTableState` likewise stores states per `SinkIdentifier`. The queue index is a transient, in-process concurrency/ordering-locality construct that is recomputed from scratch on every job start/restore — it is never serialized into checkpoint/savepoint state. So this change carries zero checkpoint-compatibility or job-resume risk.
- **Routing does change for more than just the `Integer.MIN_VALUE` edge case, and that's worth stating explicitly.** For any key whose `hashCode()` is negative (roughly half of all possible hash values, not only `Integer.MIN_VALUE`), `hashCode() & Integer.MAX_VALUE` produces a different non-negative number than `Math.abs(hashCode())` would have (e.g. for `hashCode() == -1`: old `Math.abs(-1) = 1`; new `-1 & Integer.MAX_VALUE = 2147483647`). So the specific queue a given key lands on can shift for many keys, not just the one crashing value. This is safe here only because (a) the mapping has no external contract (it's purely internal load distribution/ordering-locality across the queues owned by a single `MultiTableSinkWriter` instance) and (b) it's not persisted (per the point above) — but it's still a behavior change worth calling out plainly in the PR description for anyone auditing "what changed," rather than describing the fix as touching only the overflow case.
- No config option is renamed, no default value changes, and no public API/SPI signature changes — consistent with the project's backward-compatibility rules.

`&` is at least as cheap as `Math.abs` (which itself compiles to a conditional negate), so there's no measurable performance regression — if anything this is marginally cheaper (no branch). No new allocations, no new synchronization, no change to thread-safety of `write()`.

No error-handling or logging code paths are touched by this diff. However, verifying this PR end-to-end surfaced concrete, numbered issues that must/should be addressed:

**Issue 1 — CI is currently red because of a Spotless formatting violation introduced by this exact diff.**
- Location: `seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriter.java:547-550` (the updated Javadoc `{@code ...}` line).
- Problem: The fork's `Run / Code style` job (`./mvnw spotless:check`) fails with a line-wrap violation on exactly the two lines this PR edited (`... (primaryKeyValue.hashCode() & Integer.MAX_VALUE) % queueSize} ...` needs to wrap after `Integer.MAX_VALUE)` instead of after `%`). This is not flaky/unrelated CI noise — it is caused directly by this PR's own changed lines, confirmed from the fork's `Build` workflow log.
- Potential risk: The PR cannot be merged with a failing required check.
- Best improvement: Run `./mvnw spotless:apply -pl seatunnel-api -am` locally and push the reformatted Javadoc (a two-line wrap identical in spirit to the existing `BigtableSourceSplitEnumerator` example, which wraps the same `& Integer.MAX_VALUE) \n % numReaders` pattern across two lines).
- Severity: Blocker.

**Issue 2 — No test constructs (or mocks) a key whose `hashCode()` is `Integer.MIN_VALUE` to prove the crash is actually fixed.**
- Location: `seatunnel-api/src/test/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriterTest.java` (no such test exists, before or after this PR).
- Problem: The existing test suite exercises `write()` for failure-isolation/retry scenarios but never asserts anything about primary-key hash routing, and this PR adds no regression test either. As it stands, nothing in CI would catch a future re-introduction of `Math.abs(...)` (or an equally-broken alternative) on this line.
- Potential risk: Silent regression of exactly this crash in the future, with no automated signal until it recurs in production against a non-power-of-two `multi_table_sink_replica`.
- Best improvement: Add a focused unit test that (a) builds a `MultiTableSinkWriter` with `multi_table_sink_replica` set to a non-power-of-two value (e.g. 3, matching the issue's repro), (b) writes a row whose primary key field is an object with `hashCode()` overridden/mocked to return `Integer.MIN_VALUE`, and (c) asserts `write()` completes without `IndexOutOfBoundsException` — mirroring the same-formula assertion style already used in `JdbcSourceSplitEnumeratorTest.java:197` (`(splitId.hashCode() & Integer.MAX_VALUE) % parallelism`), which is a good template to copy.
- Severity: Major.

**Issue 3 — Architecture doc still shows the buggy pattern as the canonical example.**
- Location: `docs/en/architecture/features/multi-table.md:284` and `:399` (`docs/zh` does not mirror this section, so it is unaffected).
- Problem: The doc's illustrative pseudocode for "hash-based replica selection" still reads `return Math.abs(primaryKey.get().hashCode()) % replicaNum;` / `int replica = Math.abs(primaryKey.hashCode()) % replicaNum;` — the exact defect this PR removes from the real implementation. Once this PR merges, the doc will be teaching new contributors the unsafe pattern as if it were the recommended approach.
- Potential risk: Low direct production risk (it's design documentation, not a config reference), but it invites the same bug to be reintroduced by future contributors modeling new code on this doc.
- Best improvement: Update both snippets to `(primaryKey.get().hashCode() & Integer.MAX_VALUE) % replicaNum` for consistency with the fixed implementation. Nice-to-have, not required to unblock this PR.
- Severity: Minor.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The change itself follows existing code style (parenthesization matches the codebase's other 13 usages of this exact idiom) and needs no wildcard imports or new dependencies. The only standards violation is the Spotless line-wrap covered as Issue 1 above.

No test files are touched by this PR (see Issue 2). No existing test's assertions, tolerances, or scope were weakened — there is simply nothing new to lock in the fix yet. Given SeaTunnel's testing conventions favor a dedicated regression test for exactly this kind of boundary/exception path (a `hashCode() == Integer.MIN_VALUE` key against a non-power-of-two queue count), adding one before merge is the right bar here, and it's a small, self-contained addition given the pattern to copy already exists in `JdbcSourceSplitEnumeratorTest`.

## 2.3 Documentation Updates

No user-facing config docs (`docs/en/connectors/**`) reference this internal routing formula, so none of those needed updating. The one relevant doc gap is the stale pseudocode in the architecture doc (Issue 3), which is optional/non-blocking.

# 3. Architectural Soundness

Minimal, surgical, single-line functional change plus a matching doc-comment update — exactly the right size for this bug. It reuses an idiom the codebase already leans on 13 times elsewhere rather than inventing a new one.

## 3.2 Maintainability

Improves maintainability by removing the last outlier still using the unsafe `Math.abs` form, making the codebase's hash-routing style fully consistent. Once Issue 2's test lands, this spot becomes self-defending against regression.

## 3.3 Extensibility

No interfaces, SPIs, or extension points are touched; nothing here constrains or enables future extensibility one way or the other.

As analyzed in 1.2: the queue-routing formula has no persisted footprint (verified via `MultiTableState`/`SinkIdentifier`-keyed restore), so upgrading across this change — including resuming an in-flight job from a checkpoint taken on the old binary — carries no state-restore risk. This is an internal load-distribution detail, not a wire/storage format, so there is nothing to migrate and no `incompatible-changes.md` entry is warranted.

# 4. Issue Summary

|---|---|---|---|
| 1 | CI fails: Spotless format violation on the exact lines this PR edited | `MultiTableSinkWriter.java:547-550` | Blocker |
| 2 | No test exercises a key whose `hashCode()` is `Integer.MIN_VALUE` against a non-power-of-two queue count | `MultiTableSinkWriterTest.java` (missing) | Major |
| 3 | Architecture doc still shows the pre-fix `Math.abs(...)` pattern as the recommended example | `docs/en/architecture/features/multi-table.md:284,399` | Minor |

# 5. Merge Recommendation

**Blockers:**
1. Fix the Spotless violation (`./mvnw spotless:apply -pl seatunnel-api -am`) so the required `Build` check turns green — this is caused directly by this PR's own diff, not flakiness.
2. Add a regression test that proves the fix: a primary key with `hashCode() == Integer.MIN_VALUE` routed against a non-power-of-two `multi_table_sink_replica`, asserting no `IndexOutOfBoundsException`.

**Recommended (non-blocking):**
- Refresh the two stale `Math.abs(...)` pseudocode snippets in `docs/en/architecture/features/multi-table.md` while you're in the area.

**Overall assessment:** The root-cause analysis is correct, and — importantly, since this was the crux of the review — the chosen fix (`hashCode() & Integer.MAX_VALUE`, then ordinary `% n`) is mathematically correct for *any* positive queue count, not only powers of two, unlike the more naive `hash & (n-1)` masking trick. It also matches an idiom already proven out in 13 other places in this codebase, and the routing change has no checkpoint/state-restore compatibility impact since queue index is never persisted. This is good, well-targeted work for a first contribution. Once the CI formatting is fixed and a small regression test is added, this should be quick to land.

nzw921rx:
PR #11721 merged, Thank you for your efforts.

PR #11626, opened 2026-08-02
[Improve][E2E] Reuse SeaTunnel container for selected test classes

goutamadwant:
### Purpose of this pull request

This is the first opt-in slice of #10784.

- Add a JUnit root resource that keeps one reusable container instance per container type and JVM.
- Serialize test-class access to a shared container and restart it after startup or cleanup failures.
- Add class-level cleanup for the SeaTunnel container, including active-job checks, mounted-volume cleanup, temporary config cleanup, and connector JAR restoration.
- Opt `FakeIT` and `FakeSqlConfIT` into sharing only the SeaTunnel engine container. Flink and Spark containers keep their existing lifecycle.
- Add focused lifecycle tests for reuse, class serialization, cleanup failures, startup failures, and JVM shutdown.

Connector JARs remain loaded on demand by each test class. JARs added by that class are removed during cleanup, so the next class starts from the same connector baseline.

### Does this PR introduce _any_ user-facing change?

No. This only changes internal E2E test infrastructure.

### How was this patch tested?

Unit tests and formatting:

```shell
./mvnw -pl seatunnel-e2e/seatunnel-e2e-common spotless:apply test -DskipIT
```

Representative multi-engine E2E run:

```shell
TEST_IN_PR=true RUN_ALL_CONTAINER=false RUN_ZETA_CONTAINER=true \
./mvnw -pl seatunnel-e2e/seatunnel-connector-v2-e2e/connector-fake-e2e \
  -Dapi.version=1.40 -DskipUT -DskipIT=false \
  -Dit.test=FakeIT,FakeSqlConfIT verify
```

Both test classes passed on Flink 1.18, Flink 1.20, Spark 3.3, and SeaTunnel Engine: 8 tests, 0 failures, 0 errors.

Using the same warmed local Docker environment and command:

- `dev` at `2b51d3614`: 12 min 29 sec
- this branch at `84be9f5e5`: 11 min 14 sec
- measured reduction: 75 sec, about 10 percent for this two-class slice

The branch run started one SeaTunnel Engine container and reused it across both classes. The cleanup checks confirmed no active jobs and restored the volume, temporary config, and connector JAR baselines between classes.

### Check list

* [x] No new Jar binary package is added, so no License Notice change is required.
* [x] No documentation change is required because this only affects internal E2E test infrastructure.
* [x] No incompatible user-facing change is introduced.
* [x] No connector packaging files are affected.

PR #11559, opened 2026-07-26
[Feature][Zeta] Implement dynamic lookup runtime

DanielLeens:
### What problem this PR solves

Refs #7937.

SeaTunnel currently has no engine-native dynamic lookup operator that can safely combine a fact stream with a CDC dimension stream while preserving fact-position ownership, checkpoint/restore identity, source-gate ordering, and bounded resource admission.

This draft updates the earlier Phase-0 descriptor spike into an end-to-end implementation draft for the Dynamic Lookup design. It remains a Draft PR until GitHub CI, reviewer feedback, and runtime evidence gates are complete.

Design discussion: https://github.com/apache/seatunnel/issues/7937#issuecomment-5083135821

Latest implementation head: `f40ad4eaa1e1930ca9fa29003f632819345837e1`

### What is changed

- Adds dedicated dynamic lookup API contracts:
  - `DynamicLookupSourceCapability`
  - `FactSourceGateCapability`
  - `SourceGateCommand`
  - `SourceGateState`
- Adds parser support for top-level `dynamic_lookup` blocks and builds a real `DynamicLookupAction` with output `CatalogTable` materialization.
- Adds planner validation for:
  - fact/dimension direct source ownership
  - exactly one source-level fact gate owner
  - equal fact/dimension parallelism for the M0 direct-source topology
  - required source capabilities
  - append-only fact input
  - dimension primary-key update fail-fast mode
  - schema-change fail-fast mode
  - in-memory, no-TTL, single-snapshot resource constraints
  - 512 MiB logical/resident state admission for the current in-memory runtime
- Adds a same-task-group blocking-queue runtime:
  - dimension state keyed by planner-resolved key fields
  - LEFT and INNER lookup projection
  - fact append-only enforcement
  - dimension UPDATE_BEFORE/UPDATE_AFTER same-key enforcement
  - schema-change rejection
  - bounded per-port draining so a hot dimension queue cannot starve fact processing
  - two-port checkpoint barrier alignment with per-port blocking before snapshot/ack/emit
  - checkpoint-abort cleanup for pending alignment state
  - versioned and SHA-256 checked dynamic lookup state envelope
  - runtime resource fail-fast on dimension mutation, restore, and snapshot
- Adds Kafka fact source gate support:
  - closed-gate split staging
  - versioned gate state with split digest
  - gate-state snapshot through the SourceFlowLifeCycle checkpoint path
  - restore directly into the reader, without routing gated split bytes back through the enumerator restore path
[...]

PR #11557, opened 2026-07-26
[STIP-31][Feature][Zeta] Introduce engine-managed Source runtime

DanielLeens:
## Purpose

Implement [STIP-31](https://github.com/apache/seatunnel/issues/11558) in one phase-gated Draft PR. This moves Source Reader and Split Enumerator concurrency ownership into Zeta while preserving the existing Source SPI and legacy connector behavior through an isolated compatibility lane.

The implementation is intentionally fail-closed and disabled by default. This PR remains Draft until its current head satisfies the recovery, compatibility, fault-injection, and performance gates below.

## Implemented scope

- [x] Versioned managed Source capability, cooperative Reader polling, and coordinator scheduler APIs with backward-compatible SPI defaults.
- [x] Exact Lane A allowlist, global kill switch, protocol/capability validation, persisted lane selection, and legacy physical-plan restore compatibility.
- [x] Bounded Reader and coordinator owner loops with reserved control capacity, payload-byte limits, worker-wide memory accounting, and immediate Hazelcast operation admission.
- [x] Versioned command envelopes/codecs, explicit wire codes, checksums, sequencing, deduplication, attempt/epoch fencing, single-flight transport, deadlines, and bounded retry.
- [x] Checkpoint-aware split ownership from DISPATCHED through CHECKPOINT_INCLUDED, including chunked ownership proofs, restore reconciliation, rescale fencing, and no-more-splits generations.
- [x] Explicit Reader and coordinator checkpoint serializers with capability digest, connector state version, lifecycle state, and corruption validation.
- [x] Non-blocking coordinator async scheduler with bounded IO/CPU pools, fixed-delay execution, overlap/failure policies, cancellation, timeout, classloader restoration, and stale-result fencing.
- [x] Executable Reader lifecycle and schema-change state transitions, graceful-close latching, checkpoint callback ordering, watchdog wakeup, and terminal preemption.
- [x] Managed runtime metrics for admission, mailbox depth/age, command service time, checkpoint wait, ownership state, scheduler work, retries, and contention.
- [x] Connector-common managed Reader bridge and pilots for Fake, Kafka, and Iceberg. Iceberg intentionally adopts the coordinator runtime only; its Reader remains on the legacy lane.
- [x] English/Chinese operator documentation, configuration defaults, selector/mailbox/protocol/checkpoint/scheduler/assignment tests, and Kafka managed-discovery regression coverage.

## Compatibility and performance contract

[...]

PR #11206, opened 2026-06-28
[Feature][Connector-V2] Support MySQL CDC newly added tables

DanielLeens:
### What changed
- add `scan.newly-added-table.enabled` and `scan.binlog.newly-added-table.enabled` for MySQL CDC wildcard capture
- convert Debezium create-table schema changes into SeaTunnel `CreateTableEvent` and keep the discovered table schema in sync
- let `MultiTableSinkWriter` register per-table sink writers after job startup so runtime-created tables can flow into JDBC multi-table sinks
- resolve JDBC runtime tables with the same database, naming, and primary-key rules as startup tables
- add unit coverage and a targeted MySQL CDC E2E case for wildcard capture with a table created during the binlog phase

### Why
MySQL CDC whole-database sync was still bounded by the startup table set. Even if the source discovered a new table later, the source-side schema event was not turned into a runtime sink registration path, so the new table could not be written without manual intervention.

### Validation
- `./mvnw spotless:apply -Dskip.ui=true`
- `./mvnw -pl seatunnel-api,seatunnel-connectors-v2/connector-cdc/connector-cdc-base,seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql,seatunnel-connectors-v2/connector-jdbc,seatunnel-translation/seatunnel-translation-flink/seatunnel-translation-flink-common -am -DskipTests -Dskip.ui=true install`
- `./mvnw -pl seatunnel-api -Dtest=MultiTableSinkWriterTest,EventTest test -Dskip.ui=true`
- `./mvnw -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-base -Dtest=SeaTunnelRowDebeziumDeserializeSchemaTest test -Dskip.ui=true`
- `./mvnw -pl seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql -Dtest=MySqlSourceConfigFactoryTest,MySqlSchemaTest test -Dskip.ui=true`
- `./mvnw -pl seatunnel-connectors-v2/connector-jdbc -Dtest=JdbcSinkFactoryTest test -Dskip.ui=true`
- `./mvnw -pl seatunnel-api,seatunnel-connectors-v2/connector-cdc/connector-cdc-base,seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql,seatunnel-connectors-v2/connector-jdbc,seatunnel-translation/seatunnel-translation-flink/seatunnel-translation-flink-common -am -q -DskipTests -Dskip.ui=true verify`

### Additional notes
[...]

PR #11726, opened 2026-08-09
[Test][API] Tolerate the pre-existing close() race in the schema-change worker-failure test

DanielLeens:

Fix a `unit-test (11, windows-latest)` failure unrelated to the PRs it blocks.

`MultiTableSinkWriterSchemaChangeBroadcastTest#schemaChangeKeepsIOExceptionContractWhenWorkerAlreadyFailed` (added by #11015) intermittently fails with:

```
[ERROR] schemaChangeKeepsIOExceptionContractWhenWorkerAlreadyFailed  Time elapsed: 0.178 s  <<< ERROR!
java.lang.RuntimeException: java.io.IOException: boom-before-schema-entry
	at MultiTableSinkWriterSchemaChangeBroadcastTest.schemaChangeKeepsIOExceptionContractWhenWorkerAlreadyFailed(MultiTableSinkWriterSchemaChangeBroadcastTest.java:412)
Caused by: java.io.IOException: boom-before-schema-entry
```

Observed on Windows-JDK11 across four unrelated PRs: #10453, #11557, #11559, #11626.

### Root cause

Line 412 is a bare `coordinator.close();` at the end of the test, after the test has already asserted (lines 400–411) that `applySchemaChange()` correctly turns the worker's stored row-write failure into an `IOException`. The `close()` call is incidental cleanup — nothing in the test asserts on it.

`MultiTableSinkWriter.close()` calls `checkQueueRemain()`:

```java
private void checkQueueRemain() throws IOException {
    while (hasPendingRuntimeWrites()) {
        Thread.sleep(100);
        subSinkErrorCheck();   // only reachable while a write still looks "pending"
    }
}
```

`subSinkErrorCheck()` is only invoked *inside* the loop, i.e. only while `hasPendingRuntimeWrites()` (which checks `MultiTableWriterRunnable.isProcessingRow()`) is still `true`. In `MultiTableWriterRunnable.run()`, a failing row sets the worker's `throwable` field, and only afterward — in a **separate** volatile write — clears `processingRow`:

```java
} catch (Throwable error) {
    ...
    throwable = error;              // stored first
    failPendingSchemaChangeRequests(queueElement, error);
    ...
    processingRow = false;          // cleared afterward
    break;
}
```

The test's own poll loop (lines 396–398) only waits for `getThrowable() != null`; it says nothing about `processingRow`. So there is a real window where a concurrent `close()` can observe the stored failure while `processingRow` still reads `true`, re-enter `checkQueueRemain()`'s loop, call `subSinkErrorCheck()`, and re-throw the *same* failure the test already asserted on two lines earlier — now escaping as an unhandled `RuntimeException` instead of an assertion.

This is **pre-existing, long-standing behavior**, not a regression from #11015: `checkQueueRemain()`'s shape (`subSinkErrorCheck()` reachable only inside the pending-writes loop) is unchanged back to January 2026 (`eca4e2b3421414d0b5690b1bf621845da09879be`). Whether `close()` re-surfaces an already-observed failure has always been timing-dependent; what changed in #11015 is a new test that implicitly assumed it wouldn't.

I deliberately did not touch `MultiTableSinkWriter`/`MultiTableWriterRunnable` production code — several other open PRs (#10306, #11206, #11721, #11725) are independently modifying those same classes, and closing this specific timing window is a separate, higher-risk change than this CI-stability fix warrants.

No. Test-only change; no production code is touched.

- `assertThrows(IOException.class, () -> coordinator.applySchemaChange(...))` and the message assertion on it are untouched — that is the behavior this test's name describes, and it is unaffected by this change.
- The trailing `coordinator.close()` is now wrapped to accept its two legitimate outcomes: returning normally, or re-throwing the *exact same* `IOException` instance already asserted above (checked with `assertSame`, not just a message match — this is a strictly stronger check than the message-equality assertion already used earlier in the test, not a weaker one). Any other exception still fails the test.

Existing test keeps every assertion it had, plus the new `assertSame` check on the tolerated `close()` path. Correctness verified by CI on this PR head.

### Check list

### DIFF
--- a/seatunnel-api/src/test/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriterSchemaChangeBroadcastTest.java
+++ b/seatunnel-api/src/test/java/org/apache/seatunnel/api/sink/multitablesink/MultiTableSinkWriterSchemaChangeBroadcastTest.java
@@ -409,7 +409,22 @@ public class MultiTableSinkWriterSchemaChangeBroadcastTest {
                                         new TestSchemaChangeEvent(
                                                 TablePath.of("dbA", null, "users"))));
         assertEquals("boom-before-schema-entry", schemaChangeFailure.getMessage());
-        coordinator.close();
+        // checkQueueRemain() (invoked by close()) only re-checks subSinkErrorCheck() while a
+        // queue element still looks pending, and MultiTableWriterRunnable clears that pending
+        // flag in a separate volatile write issued after the worker's throwable field is already
+        // stored (MultiTableWriterRunnable.run()). Whether close() observes that narrow window
+        // and re-surfaces the same row failure is a timing race in existing close() behavior, not
+        // a regression and not what this test verifies above. Tolerate either outcome, but fail
+        // if close() surfaces anything other than the exact failure already asserted.
+        try {
+            coordinator.close();
+        } catch (RuntimeException e) {
+            org.junit.jupiter.api.Assertions.assertSame(
+                    rowWriteFailure,
+                    e.getCause(),
+                    "close() surfaced an unexpected failure instead of the known row failure: "
+                            + e);
+        }
     }
 
     /**


DanielLeens (commented):
- User pain point: `MultiTableSinkWriterSchemaChangeBroadcastTest#schemaChangeKeepsIOExceptionContractWhenWorkerAlreadyFailed` (added by an earlier PR) intermittently fails CI (observed on `unit-test (11, windows-latest)`) with an uncaught `RuntimeException` escaping from a trailing, unasserted `coordinator.close()` cleanup call.
- Fix approach: wrap that `close()` call in try/catch, tolerating either of its two legitimate outcomes (no exception, or a `RuntimeException` whose cause is exactly the already-known row-write failure), while still failing the test if any other exception surfaces.
- One-sentence summary: this is a test-only change that makes an existing test tolerate a narrow, pre-existing production timing race in `MultiTableSinkWriter.close()`, without touching production code.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Diff scope**: single hunk, `MultiTableSinkWriterSchemaChangeBroadcastTest.java` (test file only; no production code touched). This is small enough that full call-chain tracing was done, not sampled.

**Before**: the test asserted `applySchemaChange()` correctly converts the worker's stored row-write failure into an `IOException` (unchanged), then called `coordinator.close();` as bare, unasserted cleanup. Since the test method is declared `throws Exception`, any exception escaping `close()` would fail the test as an *unhandled* exception — this was never a deliberate assertion, just an accident of not catching it.

**After**: `close()` is wrapped; if it throws, the test asserts (`assertSame`) that the cause is exactly the same `rowWriteFailure` instance already asserted on above. Any other exception (wrong cause, or a checked exception escaping instead of the documented `RuntimeException`) still fails the test.

**Independent verification of the claimed race** (traced the full path myself, not taken at face value from the PR description):

- `MultiTableSinkWriter.close()` calls `checkQueueRemain()` inside a try/catch that stores the first exception.
- `checkQueueRemain()` only calls `subSinkErrorCheck()` **inside** its `while (hasPendingRuntimeWrites())` loop body — there is no equivalent call once the loop condition is false or after the loop exits.
- `hasPendingRuntimeWrites()` returns true if any `MultiTableWriterRunnable.isProcessingRow()` is true.
- In `MultiTableWriterRunnable.run()`, on a row-write failure with the default `FAIL_FAST` policy (confirmed via the constructor used by this test, which defaults to `MultiTableFailurePolicy.FAIL_FAST`): `handleWriteFailure()` sets the **volatile** field `throwable = error;` *before* the catch block later sets the **separate volatile** field `processingRow = false;`. These are two independent volatile writes with no atomicity between them.
- `currentSubSinkFailure()` returns `(IOException) throwable` directly when `throwable instanceof IOException` — i.e. the *exact same instance* as the test's `rowWriteFailure` (constructed as `new IOException("boom-before-schema-entry")` and never wrapped on this path).
- Therefore: if a `close()` call happens to observe `processingRow == true` in the narrow window between those two volatile writes, `checkQueueRemain()` re-enters its loop, sleeps 100ms, calls `subSinkErrorCheck()`, which throws the identical `rowWriteFailure` instance; this propagates through `checkQueueRemain()`, is caught in `close()` as `firstE[0]`, and re-thrown as `new RuntimeException(firstE[0])` — cause is `rowWriteFailure` exactly. This is precisely what the new `assertSame` checks.
- I also confirmed there is no *other* possible source of an exception from `close()` in this test: `FailingRowSinkWriter` only overrides `write()`; `close()` is inherited from `RecordingSinkWriter.close()` which is a no-op. So the try/catch is exercising exactly, and only, the described race — not silently absorbing unrelated failures.

**Verdict on the title's framing**: this is a legitimate, independently-verified, pre-existing production timing race being tolerated in a test that never actually asserted on that call, not a case of masking a newly-hidden bug. The assertion this test's Javadoc describes (`applySchemaChange` honoring its `IOException` contract once the worker has already failed) is completely untouched, and the new code adds a *strictly more specific* check (`assertSame` on cause identity) than existed before on the tolerated path — this is a net increase in rigor on the one code path that changed, not a decrease.

**Corroborating evidence beyond static analysis**: I re-ran CI-status verification independently (not trusting the PR body's claim): in the fork CI run for this exact head, all four `unit-test` matrix jobs — including `unit-test (11, windows-latest)`, the specific job cited as flaky — **pass**. The two jobs that currently fail in that run, `rocketmq-connector-it (11, ubuntu-latest)` and `all-connectors-it-6 (8, ubuntu-latest)`, are in completely unrelated connector-IT modules this diff does not touch, which is consistent with unrelated pre-existing IT infra flakiness rather than anything caused by this change.

I also checked the PR body's supporting commit citation and found the cited SHA does not actually touch `MultiTableSinkWriter.java` at all. However, `git blame` on the current `checkQueueRemain()` body shows the same try/while/catch shape traces back even further than claimed. So the substantive "pre-existing, not a regression" claim is actually *more* true than the PR body's own evidence shows, just citing the wrong SHA.

## 1.2 Compatibility Impact
**Fully compatible.** Test-only change. No production code, public API, config option, serialization format, or historical behavior is touched.

None. No new production code paths, no new sleeps/timeouts, no new resources.

The new catch block correctly narrows to `RuntimeException` (matching `close()`'s documented — if surprising — actual contract of always wrapping in `RuntimeException` despite declaring `throws IOException`). No issues.

**Issue 1 (Low, non-blocking, informational — out of scope for this PR):** The underlying production race is real and, on the evidence above, means `close()`'s ability to surface an already-known sub-sink failure is non-deterministic: in the common/dominant timing (per the observed low flake rate), the failure is silently swallowed with no re-throw and no log line; only in the rare race window does it get re-surfaced. The author deliberately left production code untouched, citing several other in-flight PRs touching the same classes — a reasonable minimal-blast-radius call for a CI-stability-only fix. Recommend filing a tracked follow-up to make this deterministic (e.g., call `subSinkErrorCheck()` once, unconditionally, after the `checkQueueRemain()` loop regardless of whether the loop body ran).

# 2. Code Quality Assessment

## 2.1 Coding Standards
The added comment is thorough: it explains the mechanism, why it's a race (not a regression), and the exact tolerance contract — exactly the level of "why + constraint" detail required for racy/complex logic. Fully-qualified `assertSame` usage matches this file's own pre-existing local convention, so it's consistent, not a new style deviation.

**Rating: Stable** (post-fix). The new code path is fully deterministic given the traced production logic — it covers exactly the two legitimate outcomes of a proven, narrow, pre-existing race, and still fails loudly (via `assertSame`) on any other outcome. No `Thread.sleep`-based flakiness, no weakened assertions, no shared static state, no hardcoded ports introduced. This is a strict improvement over the prior state, where a real timing-dependent test failure occasionally interrupted unrelated PRs' CI.

## 2.3 Documentation Updates
N/A — test-only change, no user-facing behavior changed, no docs required.

# 3. Architectural Soundness

**Precise fix** for the problem it actually solves (test flakiness caused by an unasserted, incidental `close()` call colliding with a real production race). It is explicitly *not* a fix for the underlying production race itself — that is correctly scoped out, with a stated rationale (avoiding collision with other in-flight PRs touching the same classes).

## 3.2 Maintainability
Good. The comment makes the non-obvious race understandable to a future reader without needing to re-derive it from source, which is exactly what's needed here.

## 3.3 Extensibility
N/A.

N/A — no serialization, checkpoint, or protocol surface touched.

# 4. Issue Summary

|---|---|---|---|
| 1 | CI red at PR head: two unrelated IT jobs fail (`rocketmq-connector-it`, `all-connectors-it-6`) in modules this diff does not touch; the specific `unit-test` job this PR targets passes on all 4 matrix combinations at this head | Fork CI run | High (merge gate, not a code defect) |
| 2 | PR description cites a commit as evidence for "unchanged since Jan 2026," but that commit does not modify `MultiTableSinkWriter.java`; the underlying claim is still correct (even more so, per `git blame`) but the citation is wrong | PR description only, not code | Low |
| 3 | Underlying production race (`checkQueueRemain()` only calls `subSinkErrorCheck()` inside its loop body, so `close()` non-deterministically surfaces or silently swallows an already-known sub-sink failure) is real but intentionally out of scope for this PR | `MultiTableSinkWriter.java`, `MultiTableWriterRunnable.java` | Low (follow-up recommended, non-blocking here) |

# 5. Merge Recommendation
   - Issue 1 (High, CI): get CI green before merge. Recommend rerunning just the two failed jobs rather than the whole workflow. Given the diff only touches a `seatunnel-api` test file, these failures are very unlikely to be caused by this change, but they must be shown green (or conclusively identified as pre-existing unrelated flakiness) before merge — the code-level analysis in this review is not a substitute for a green CI run.
   - Issue 2: correct the SHA citation in the PR description.
   - Issue 3: file a tracked follow-up to make `close()`'s error-surfacing deterministic instead of timing-dependent.

Overall assessment: the code change itself is correct, minimal, precisely scoped, and backed by an accurate root-cause analysis that I independently reproduced from source rather than accepting on faith. This is a legitimate accommodation of a genuine, pre-existing, benign (from this test's perspective) race, not masking a real bug; the test's actual documented contract is fully preserved and the new assertion is strictly more specific than what existed before. No alternative implementation is meaningfully better for this PR's stated, narrow scope — fixing the production race itself would be a larger option, but the author's reasoning for deferring it (avoiding collision with other in-flight PRs on the same classes) is sound engineering judgment, not scope-avoidance to dodge a harder fix. The only blocker is CI going green.

davidzollo (approved):
Maintainer review, independent of the author's self-review above.

**Spot-verified the load-bearing claim rather than accepting it:** the self-review asserts the test's 3-arg constructor call defaults to `MultiTableFailurePolicy.FAIL_FAST`, "confirmed via the constructor." That policy name doesn't appear in the test file itself, so I traced it through the production constructor chain myself: `MultiTableSinkWriter(Map, int, Map)` (`MultiTableSinkWriter.java:113-126`) delegates to the full 8-arg constructor with `MultiTableFailurePolicy.FAIL_FAST` hardcoded at line 121. Confirmed correct.

**Independently re-derived the race, not just re-read the self-review's trace:** `checkQueueRemain()` only calls `subSinkErrorCheck()` inside its `while (hasPendingRuntimeWrites())` body; `hasPendingRuntimeWrites()` reads `MultiTableWriterRunnable.isProcessingRow()`; on a row-write failure under `FAIL_FAST`, `run()` sets the volatile `throwable` field before the separate volatile `processingRow = false` write two statements later. A `close()` landing in that window re-enters the loop, re-throws the same stored `IOException`, and `close()` wraps it as `new RuntimeException(firstE[0])`. This matches what the test now tolerates via `assertSame` on cause identity. No gap between the claimed mechanism and the actual code.

**On the test-severity question (mandatory before approval):** the `applySchemaChange()` assertions this test is actually named for are untouched. The only new code is around a previously-bare, unasserted `coordinator.close()` call that had no prior assertion contract to weaken — the new `assertSame` is strictly more specific than "nothing" was before. Not a weakening under any reading.

**Compatibility/scope:** single test file, no production code, no config, no protocol surface. `MultiTableSinkWriter`/`MultiTableWriterRunnable` are correctly left untouched given the several other in-flight PRs already modifying those exact classes — closing the underlying race is real follow-up work, not something this CI-stability PR should absorb.

**On Issue 2 (SHA citation) from the self-review:** confirmed independently — the cited commit does not touch `MultiTableSinkWriter.java`. Doesn't affect the substance (the race predates this test by months regardless), just a wrong citation in the PR body; not worth a re-push for.

**Merge verdict: APPROVED, conditional on CI.** unit-test passes on all 4 matrix legs at this head, including the specific Windows job this PR targets. The failing jobs in this run (`rocketmq-connector-it`, `all-connectors-it-6`) are in modules this diff doesn't touch — retried via `--failed`.

DanielLeens:
Thanks for the independent maintainer pass, @davidzollo — appreciate you re-deriving the `FAIL_FAST` constructor chain and the race yourself rather than taking my self-review's trace on faith. Good catch flagging that the policy name doesn't appear in the test file itself; verifying it through the actual `MultiTableSinkWriter(Map, int, Map)` -> 8-arg constructor delegation is exactly the right level of skepticism for a self-review.

Rechecking the current head now: the only blocker in my own review was Issue 1 (CI must go green before merge, since the code-level analysis alone isn't a substitute for a real CI run). I just re-verified the check status and all checks are now passing — no non-success conclusions remain, and `mergeable` is `MERGEABLE`. That closes the one blocker I had open.

Combining that with your independent `APPROVED`: I don't have a further source-level blocker to add. Note that I can't formally re-approve my own PR through GitHub's review UI (self-approval is blocked), so this is recorded here as a comment rather than a state change, but from my side the technical conclusion is now **Ready to merge** — the only prior gate (CI) is cleared and your maintainer approval stands as the merge sign-off. Issues 2 and 3 from my original review remain as non-blocking follow-ups (fix the SHA citation in the description at your convenience; the underlying production race in `close()` is real but correctly out of scope for this CI-stability-only PR).

zhangshenghang (approved):
+1

PR #11730, opened 2026-08-10
[Fix][Connector-V2] Omit empty UPDATE SET in JDBC MERGE for all-key tables

### LINKED ISSUE #11729: [Bug] [JDBC] MERGE upsert fails with syntax error for all-key tables
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/seatunnel/issues) and found no similar issues.

### What happened

When using JDBC connectors for databases that rely on the `MERGE ... WHEN MATCHED THEN UPDATE SET` upsert syntax (DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu), upserting into a table whose **every column is a primary/unique key** (i.e. there is no non-key field to update) fails.

The generated upsert SQL always appends `WHEN MATCHED THEN UPDATE SET <clause>`. When `updateSetClause` is empty, this produces invalid SQL (`UPDATE SET` with an empty body), and the database throws a syntax error, causing the job to fail.

### SeaTunnel Version

2.3.13 (and dev branch)

### Connector Name

jdbc — dm, oracle, saphana, sqlserver, vertica, xugu

When there is nothing to update, the `WHEN MATCHED THEN UPDATE SET` clause should be omitted entirely, so the MERGE statement is valid and the job succeeds (matched rows are left unchanged, unmatched rows are inserted).

### How to reproduce

1. Create a table whose all columns are primary keys, e.g.:
   ```sql
   CREATE TABLE all_key_t (a INT PRIMARY KEY, b INT PRIMARY KEY);
   ```
2. Configure a SeaTunnel JDBC sink (dm / oracle / saphana / sqlserver / vertica / xugu) with an upsert mode and run a job that writes/updates multiple rows.
3. Observe the SQL syntax error thrown by the database (empty `UPDATE SET`).

### Anything else

A fix is provided in the linked PR.

siwen-yu:

Fix the MERGE (upsert) SQL generation in JDBC dialects for tables whose every column is a primary/unique key (i.e. no non-key field to update). Previously `getUpsertStatement` always appended `WHEN MATCHED THEN UPDATE SET <clause>`; when `updateSetClause` is empty this yielded invalid SQL (`UPDATE SET` with an empty body) and the database threw a syntax error. Affected dialects: DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu.

This PR introduces a `matchedClause` string that is omitted entirely when there is nothing to update, keeping the behaviour identical when `updateSetClause` is non-empty.

Closes #11729

Yes (bug fix, no API/config/schema change).
- Before: upserting into an all-key table via the above dialects failed with a SQL syntax error (empty `UPDATE SET` body).
- After: the MERGE statement is generated correctly; matched rows are left unchanged and unmatched rows inserted, so the job succeeds.

Manually verified by configuring a JDBC sink (DM / Oracle / etc.) against an all-key table and running an upsert job:
1. Created a table whose all columns are primary keys, e.g. `CREATE TABLE all_key_t (a INT PRIMARY KEY, b INT PRIMARY KEY);`
2. Ran a SeaTunnel job that upserts multiple rows into the table.
3. Confirmed the generated MERGE statement no longer contains an empty `UPDATE SET` and the job completes successfully.

### Check list

- [X] No new Jar binary package added
- [ ] Documentation update — not needed (internal SQL generation fix, no user-facing config)
- [ ] incompatible-changes.md — not needed (backward compatible)
- [ ] Connector checklist (plugin-mapping/pom/labeler/e2e) — not applicable, no new connector added


DanielLeens (changes requested):

For the six JDBC dialects that implement upsert via a SQL `MERGE` statement (DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu), `getUpsertStatement()` always emitted `WHEN MATCHED THEN UPDATE SET <updateSetClause>`, where `<updateSetClause>` is built only from the non-key columns. When a table's **every** column is part of the primary/unique key (a pure "all-key" table, e.g. a junction/mapping table with a composite key), there are zero non-key columns, so `<updateSetClause>` is empty and the generated SQL becomes `UPDATE SET WHEN NOT MATCHED THEN ...` — a syntax error that fails the job at write time. This PR fixes it by building the `WHEN MATCHED ... UPDATE SET ...` fragment as its own `matchedClause` string and omitting it entirely when there's nothing to update, so unmatched rows still insert normally and matched rows are simply left untouched.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The fix is applied identically across `DmdbDialect`, `OracleDialect`, `SapHanaDialect`, `SqlServerDialect`, `VerticaDialect`, and `XuguDialect`. Using `OracleDialect` as representative:

**Before:**
```java
String upsertSQL =
        String.format(
                " MERGE INTO %s TARGET"
                        + " USING (%s) SOURCE"
                        + " ON (%s) "
                        + " WHEN MATCHED THEN"
                        + " UPDATE SET %s"
                        + " WHEN NOT MATCHED THEN"
                        + " INSERT (%s) VALUES (%s)",
                tableIdentifier(database, tableName), usingClause, onConditions,
                updateSetClause, insertFields, insertValues);
```

**After:**
```java
String matchedClause =
        StringUtils.isNotBlank(updateSetClause)
                ? String.format(" WHEN MATCHED THEN UPDATE SET %s ", updateSetClause)
                : "";

String upsertSQL =
        String.format(
                " MERGE INTO %s TARGET"
                        + " USING (%s) SOURCE"
                        + " ON (%s) "
                        + "%s"
                        + " WHEN NOT MATCHED THEN"
                        + " INSERT (%s) VALUES (%s)",
                tableIdentifier(database, tableName), usingClause, onConditions,
                matchedClause, insertFields, insertValues);
```

**Runtime path verification** — I traced this end-to-end rather than trusting the PR description at face value:
- `JdbcOutputFormatBuilder.build()` (`JdbcOutputFormatBuilder.java:60`) reads `primaryKeys = jdbcSinkConfig.getPrimaryKeys()`. When `primaryKeys` is non-empty (line 81, `else` branch) it calls `createUpsertBufferedExecutor(...)` with `jdbcSinkConfig.isEnableUpsert()` (line 101).
- `ENABLE_UPSERT` defaults to `true` (`JdbcSinkOptions.java:102-106`), so any sink config that sets `primary_keys` takes the upsert path by default without any extra opt-in.
- `createUpsertExecutor` (line ~197) then calls `dialect.getUpsertStatementByTableSchema(database, table, tableSchema, pkNames)` → `getUpsertStatement(database, tableName, tableSchema.getFieldNames(), pkNames)`, i.e. `fieldNames` is **every** column of the table and `pkNames` is the configured `primary_keys`.
- Inside each fixed dialect, `nonUniqueKeyFields = fieldNames.filter(f -> !pkNames.contains(f))`. If `pkNames` covers every column (a legitimate, non-exotic table design — e.g. a many-to-many link table), `nonUniqueKeyFields` is empty and `updateSetClause` is empty.

So this bug sits squarely on the default/normal upsert path for any table whose key covers all columns — not some obscure recovery/edge branch. Before the fix, such a config would deterministically fail every batch write with a DB-side SQL syntax error; after the fix it correctly inserts new rows and silently leaves already-present rows untouched on match (equivalent semantics to `ON CONFLICT DO NOTHING` for an all-key table, since there's nothing meaningful to update).

**Correctness of the omission approach**: for the normal case (`updateSetClause` non-blank), the generated SQL is functionally unchanged — the `WHEN MATCHED THEN UPDATE SET ...` text and its parameter values are identical to before, just assembled through an intermediate variable instead of being formatted inline. I manually verified this token-by-token for `SqlServerDialect` (byte-identical whitespace to the original) and for the other five dialects (functionally identical SQL, with an extra run of insignificant whitespace between `%s` and `WHEN NOT MATCHED` introduced by concatenating two already-spaced literals — harmless for SQL, not worth a separate fix).

**Finding — the fix is incomplete relative to the codebase's own precedent (novel issue, see 1.4 Issue 1).**

## 1.2 Compatibility Impact

**Fully compatible.** For every table that has at least one non-key column (the overwhelmingly common case), the generated MERGE statement is unchanged in structure and semantics — same clauses, same parameter order, same bound values. The only behavior change is for the previously-broken all-key case, which moves from "always throws" to "works correctly." No config option, default value, serialization format, or public API is touched. No `incompatible-changes.md` entry is needed, and the author correctly left that checkbox unticked.

None of substance. `matchedClause` is a small, one-time `String.format` call per prepared-statement build (this happens once per sink writer initialization, not per row), so there's no hot-path allocation concern. No new connections, threads, retries, or resource lifecycles are introduced.

No logging is touched by this change. One issue found by tracing the same bug pattern into a dialect this PR didn't touch:

### Issue 1: `YashanDbDialect` has the identical bug and was not fixed

- **Location**: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/yashandb/YashanDbDialect.java`, `getUpsertStatement()` (method starts at line 108 in the current `dev` copy of the file; the vulnerable `MERGE INTO ... WHEN MATCHED THEN UPDATE SET %s ...` template is emitted unconditionally, around lines 145-159).
- **Problem description**: `YashanDbDialect.getUpsertStatement()` is essentially a byte-for-byte copy of the pre-fix `DmdbDialect`/`OracleDialect` logic — same `nonUniqueKeyFields` computation, same unconditional `" WHEN MATCHED THEN" + " UPDATE SET %s"` template, no `StringUtils.isNotBlank` guard. It was already importing `org.apache.seatunnel.shade.org.apache.commons.lang3.StringUtils` for other purposes, so wiring in the same guard would be a trivial, low-risk addition.
- **Potential risk**: Any upsert into an all-key table on YashanDB fails with the exact same SQL syntax error this PR fixes for the other six dialects — the underlying issue (#11729) is not fully closed, it will just resurface the next time someone hits it on YashanDB.
- **Best improvement**: Apply the same `matchedClause` treatment to `YashanDbDialect.getUpsertStatement()`, and extend `JdbcAllKeyTableUpsertTest` to cover it (its existing `YashanDbDialectTest.testGetUpsertStatement()` only tests the partial-key case, so it currently would not catch this).
- **Severity**: Medium (a real, reproducible defect with the same root cause and blast radius as the one being fixed, but it is outside the literal scope of issue #11729, which enumerated only DM/Oracle/SAP HANA/SQL Server/Vertica/Xugu; I don't consider it a hard blocker for merging this specific fix, but it should not be left for a future bug report to rediscover).

I also checked `DB2Dialect`, the only other dialect using `WHEN MATCHED` syntax: it builds `updateSetClause` from **all** field names (not just non-key ones) and derives its `WHEN MATCHED AND (...)` condition from a field-level `<>` comparison, so `updateSetClause` there can never be empty — DB2 is not affected by this bug class and correctly falls outside this PR's scope.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- Each of the six changed methods has a clear, accurate comment explaining *why* the clause must be conditionally omitted (not just restating the code) — this matches the repo's comment-quality expectations for non-trivial branches.
- `SapHanaDialect.java` needed a new `StringUtils` import; it was added in the correct import group/order consistent with the rest of the file.
- No new public classes, fields, or long-lived state were introduced, so there's nothing further requiring Javadoc beyond what's there.

A new unit test class, `JdbcAllKeyTableUpsertTest`, directly instantiates all six fixed dialects and asserts:
1. `testAllKeyTableOmitsEmptyUpdateSet` — for an all-key table, the generated SQL never contains `WHEN MATCHED THEN UPDATE SET`, but still contains `WHEN NOT MATCHED` and `INSERT`.
2. `testPartialKeyTableStillUpdates` — for a table with one key column and two non-key columns, `WHEN MATCHED THEN UPDATE SET` is still present (regression guard for the normal case).

This is good, targeted coverage of exactly the bug being fixed, plus a guard against accidentally regressing the common path. The test is a pure, deterministic string-assertion test with no I/O, no timing, no `Testcontainers`, and no shared/static mutable state.

**Flaky-test-risk rating: Stable.** No sleeps, no real DB connections, no ordering dependency between the two `@Test` methods, no external resources.

One coverage gap: as noted in Issue 1, this test (and the pre-existing `YashanDbDialectTest`) does not cover `YashanDbDialect`, which has the same latent bug.

## 2.3 Documentation Updates

None needed. This is an internal SQL-generation correctness fix with no new/changed config option, default, or user-observable knob — `docs/en` and `docs/zh` correctly require no changes, matching the author's own checklist assessment.

# 3. Architectural Soundness

This is a precise, minimal fix — it isolates exactly the fragment that becomes invalid when empty (`matchedClause`) rather than reworking the whole SQL-building method, and it's applied uniformly with the same pattern across all six affected dialects. Not a workaround; it's the correct long-term fix for this class of bug.

## 3.2 Maintainability

The `matchedClause` extraction makes the "this can be legitimately empty" concept explicit and self-documenting, which is an improvement over the previous inline `%s` substitution. The duplication of this same ~6-line pattern across six (really seven, counting the still-broken `YashanDbDialect`) dialect classes is pre-existing architecture (each dialect independently builds its own MERGE string) rather than something this PR introduces — a shared helper in `JdbcDialect` or a common base class could reduce this duplication in the future, but that's a larger refactor outside this PR's reasonable scope.

## 3.3 Extensibility

No impact — the fix doesn't add or remove any extension points, and any future MERGE-based dialect can follow the same now-established `matchedClause` pattern.

No serialization, checkpoint, protocol, or on-disk format is touched. No historical-version upgrade path is affected. Nothing to flag here.

# 4. Issue Summary

|---|---|---|---|
| 1 | `YashanDbDialect.getUpsertStatement()` has the identical unconditional `WHEN MATCHED THEN UPDATE SET` bug and was not fixed by this PR | `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/yashandb/YashanDbDialect.java` (`getUpsertStatement`, ~L108-159) | Medium |

# 5. Merge Recommendation

1. Blockers — must be fixed:
   - **CI is currently failing** on the `Run / Code style` job, specifically the `Check code style` step (`https://github.com/siwen-yu/seatunnel/actions/runs/31346735858/job/93329940600`), i.e. this looks like a Spotless/Checkstyle formatting violation rather than a compile or test failure. I wasn't able to fetch the raw job log directly (network egress to the CI log storage host is blocked from where I'm reviewing), so I can't quote the exact violation, but given the PR's own "How was this patch tested" section only mentions manual DB verification (no mention of running `./mvnw spotless:apply`), the most likely cause is unformatted code from this diff. Please run `./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-jdbc -am` locally and push the formatted result — this must go green before merge regardless of anything else in this review.

   - Apply the same `matchedClause` fix to `YashanDbDialect` (Issue 1) so issue #11729's root cause is closed for all MERGE-based dialects, not just the six explicitly listed. This is a small, low-risk, same-pattern change and would be a natural addition to this PR rather than a separate follow-up.

**Overall assessment**: This is a well-scoped, correctly-targeted bug fix. The runtime path is real (default `enable_upsert=true` plus any all-key primary-key configuration), the fix is minimal and preserves byte-for-byte SQL semantics for the common case, and the new test is stable and precisely targets the regression being fixed. The only thing standing between this and merge is the failing Code style CI check, plus I'd strongly encourage folding in the `YashanDbDialect` fix while this exact bug class is fresh in scope — no alternative implementation approach is needed here, the chosen fix is the right one.

siwen-yu:
Thank you, @DanielLeens, for the very thorough and thoughtful review — I really appreciate you tracing the runtime path end-to-end and confirming the fix sits on the default upsert path.

Both points are well taken:

1. **Code style** — you're absolutely right that the Code style check is currently failing. I'll run `./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-jdbc -am` locally, push the formatted result, and make sure that check goes green before merge.

2. **YashanDbDialect** — great catch. `YashanDbDialect` does carry the identical unconditional `WHEN MATCHED THEN UPDATE SET` bug, so #11729's root cause isn't fully closed without it. I'll fold in the same `matchedClause` fix there and extend the test coverage (the new `JdbcAllKeyTableUpsertTest` plus the existing `YashanDbDialectTest`) to cover the all-key case, so the fix is complete across all MERGE-based dialects.

I'll push the updates shortly and ping you for a re-review. Thanks again for the detailed and constructive feedback!

DanielLeens (approved):

For the seven JDBC dialects that implement upsert via a SQL `MERGE` statement (DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu, and now YashanDB), `getUpsertStatement()`/`getUpsertStatementByTableSchema()` always emitted `WHEN MATCHED THEN UPDATE SET <updateSetClause>`, where `<updateSetClause>` is built only from non-key columns. When a table's every column is part of the primary/unique key (a pure "all-key" table, e.g. a composite-key junction table), `<updateSetClause>` is empty, so the generated SQL becomes `... UPDATE SET WHEN NOT MATCHED THEN ...` — a syntax error that fails the write at the database. This PR builds the `WHEN MATCHED ... UPDATE SET ...` fragment as its own `matchedClause` string and omits it entirely when there is nothing to update, so unmatched rows still insert normally and matched rows are left untouched.

# Re-review note (this PR had a new commit after my last review)

Thank you @siwen-yu for pushing the YashanDB fix and the code-style pass so promptly. Since my last review at `c3fc5704`, three more commits landed (`6c91a30` "fix empty UPDATE SET clause and add dialect tests", and `c0824e0` "fix code style and cover YashanDb..."), and I've re-read the entire diff and runtime path from scratch on the current head `c0824e0`, not just the delta.

**Sorry for missing this in my last review — that's on me.** While re-verifying everything from scratch, I found that at the exact commit I reviewed (`c3fc5704`), `XuguDialect.getUpsertStatement()` still contained this pre-existing guard, unchanged by that commit:

```java
List<String> nonUniqueKeyFields = ...;
if (nonUniqueKeyFields.isEmpty()) {
    throw new SeaTunnelException(
            "The non-primary key field cannot be empty. Please set other fields");
}
```

...directly *above* the newly-added `matchedClause` logic in that same method. That means, at the version I reviewed, the fix I approved-of for Xugu was dead code: for an all-key table, `nonUniqueKeyFields.isEmpty()` is `true`, so the method throws `SeaTunnelException` unconditionally before it ever reaches the `matchedClause` omission I praised. Worse, the new `JdbcAllKeyTableUpsertTest.testAllKeyTableOmitsEmptyUpdateSet()` at that same commit iterated over all six dialects including `new XuguDialect()` and called `getUpsertStatement(DATABASE, TABLE, ALL_FIELDS, ALL_FIELDS)` — which would have thrown that same uncaught `SeaTunnelException` and failed the test outright. I attributed the CI red X purely to the "Code style" job (which was indeed the job that failed first and gated `unit-test` from ever running, so this specific test never actually got to execute in CI), but I should have caught the logical dead-code/would-fail-test issue by reading the file directly instead of reasoning only from the diff hunk. This was a carryover defect present in the old version, not something introduced by the newest commits — and it is **already fixed** on the current head: `6c91a30` deleted that guard entirely (`XuguDialect.java`, confirmed at current head, the `getUpsertStatement` method no longer contains the `throw` and goes straight into the `matchedClause` construction). No action needed from you here, just wanted to own the miss and confirm the fix is real on the current head.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The fix is applied identically across `DmdbDialect`, `OracleDialect`, `SapHanaDialect`, `SqlServerDialect`, `VerticaDialect`, `XuguDialect`, and (newly, in this update) `YashanDbDialect`. Representative before/after (`OracleDialect.java`):

**Before:**
```java
String upsertSQL =
        String.format(
                " MERGE INTO %s TARGET"
                        + " USING (%s) SOURCE"
                        + " ON (%s) "
                        + " WHEN MATCHED THEN"
                        + " UPDATE SET %s"
                        + " WHEN NOT MATCHED THEN"
                        + " INSERT (%s) VALUES (%s)",
                tableIdentifier(database, tableName), usingClause, onConditions,
                updateSetClause, insertFields, insertValues);
```

**After:**
```java
String matchedClause =
        StringUtils.isNotBlank(updateSetClause)
                ? String.format(" WHEN MATCHED THEN UPDATE SET %s ", updateSetClause)
                : "";

String upsertSQL =
        String.format(
                " MERGE INTO %s TARGET"
                        + " USING (%s) SOURCE"
                        + " ON (%s) "
                        + "%s"
                        + " WHEN NOT MATCHED THEN"
                        + " INSERT (%s) VALUES (%s)",
                tableIdentifier(database, tableName), usingClause, onConditions,
                matchedClause, insertFields, insertValues);
```

`YashanDbDialect.getUpsertStatement()` (`.../dialect/yashandb/YashanDbDialect.java`, the fix I asked for last time) now carries the exact same pattern, and `XuguDialect.getUpsertStatement()` no longer contains the stale `throw new SeaTunnelException(...)` guard I discuss above.

**Runtime path verification, re-traced end-to-end on the current head (not re-used from last time):**
- `JdbcOutputFormatBuilder.build()` routes to `createUpsertBufferedExecutor(...)` whenever `primaryKeys` is non-empty, then `createUpsertExecutor` calls `dialect.getUpsertStatementByTableSchema(...)` whenever `jdbcSinkConfig.isEnableUpsert()` is true.
- `ENABLE_UPSERT` (`JdbcSinkOptions.java:102-106`) defaults to `true`.
- Importantly — and this makes the bug's blast radius wider than I described last time — `primary_keys` does **not** need to be set explicitly by the user at all: `JdbcSinkFactory.java` (~lines 121-144) auto-populates `JdbcSinkOptions.PRIMARY_KEYS` from the catalog table's `PrimaryKey`, or failing that, the first `UNIQUE_KEY` constraint, whenever the sink config omits `primary_keys`. So a completely default JDBC-to-JDBC job with zero explicit `primary_keys`/`enable_upsert` config, writing into a target table whose catalog-reported primary key happens to cover every column (a legitimate, unremarkable schema for a pure association/junction table), hits this exact code path. This is squarely the default/normal path, not a contrived corner case.
- `JdbcDialect.getUpsertStatementByTableSchema` is a `default` interface method that just calls `getUpsertStatement(database, tableName, tableSchema.getFieldNames(), pkNames)` for every dialect except `VerticaDialect`, which overrides both methods (pre-existing split, unrelated to this PR — Vertica needs `TableSchema` for explicit `CAST(:field AS type)` binding that a bare `String[] fieldNames` can't provide). This PR only touches the body of Vertica's `getUpsertStatementByTableSchema`, not the method-selection mechanism.

**Correctness of the omission approach**: for the common case (`updateSetClause` non-blank), the generated SQL is functionally unchanged versus `dev` — same clauses, same parameter order, same bound values, just assembled through the `matchedClause` intermediate. For the previously-broken all-key case, the statement moves from "always throws a DB syntax error" (or, for Xugu specifically at the version I last reviewed, an application-level `SeaTunnelException`) to "generates a correct MERGE with matched rows left untouched and unmatched rows inserted" — equivalent semantics to `INSERT ... ON CONFLICT DO NOTHING` for a table with nothing meaningful to update on match.

**Whitespace note (unchanged from before, still harmless)**: for the five dialects whose `matchedClause` retains a trailing space (`" WHEN MATCHED THEN UPDATE SET %s "`) concatenated after an already-trailing-space `" ON (%s) "` literal, the all-key branch produces a double space before `WHEN NOT MATCHED`. This is pre-existing insignificant whitespace (the unconditional pre-fix code had the same double-space pattern in the non-empty case), not a new defect, and no SQL engine treats it as meaningful.

**Independently verified, DB2 remains unaffected and correctly out of scope**: `DB2Dialect.java` builds *both* its `WHEN MATCHED AND (...)` condition and its `UPDATE SET` clause from **all** `fieldNames`, not a `nonUniqueKeyFields` filter (confirmed: `nonUniqueKeyFields` does not appear anywhere in `DB2Dialect.java`, vs. two occurrences each in the seven fixed dialects' pre-fix code). For an all-key table, DB2's `updateSetClause` becomes a self-assignment update (`target."a" = source."a", ...`) — never empty, so the bug class genuinely cannot occur there. I also re-confirmed via `git grep "WHEN MATCHED"` across the whole `connector-jdbc` module that only these 8 dialect files use this pattern at all — the 7 now-fixed ones plus DB2 — so there is no remaining MERGE-based dialect outside this PR's scope with the same latent bug.

## 1.2 Compatibility Impact

**Fully compatible**, for every table with at least one non-key column (the overwhelmingly common case) — same clause structure, same parameter binding order, same values. The only observable behavior change is for the previously-broken all-key case: DM/Oracle/SAP HANA/SQL Server/Vertica/YashanDB move from "SQL syntax error" to "success"; Xugu moves from "explicit `SeaTunnelException` at write time" to "success." Both are strict improvements with no plausible scenario where a currently-working job breaks. No config option, default value, serialization format, or public API/SPI is touched. No `incompatible-changes.md` entry is needed.

None of substance. `matchedClause` is one extra `String.format` call per prepared-statement template build (once per sink writer initialization, not per row) — no hot-path allocation, no new connections/threads/retries/resource lifecycles.

No logging is touched. Formal issues found in this re-review:

| — | — |
|---|---|
| See table in Section 4 below. |

There are no new formal issues beyond the carryover discussed above (already fixed on this head, so it is not re-listed as an open Issue N in Section 4 — see Section 4 for why).

# 2. Code Quality Assessment

## 2.1 Coding Standards

- Each of the seven changed `getUpsertStatement`/`getUpsertStatementByTableSchema` methods carries a clear comment explaining *why* the clause must be conditionally omitted, not just restating the code — consistent with this repo's comment-quality bar for non-trivial branches.
- `SapHanaDialect.java`'s new `StringUtils` import is correctly grouped/ordered.
- `XuguDialect.java` also cleanly drops the now-unused `org.apache.seatunnel.common.utils.SeaTunnelException` import alongside removing the guard that used it — no dangling unused import left behind.
- No new public classes/long-lived fields are introduced, so nothing further needs Javadoc beyond what's present.

Coverage has been restructured since my last review: the single cross-package `JdbcAllKeyTableUpsertTest` was replaced with per-dialect test classes (`DmdbDialectTest`, new `OracleDialectTest`, new `SapHanaDialectTest`, `SqlServerDialectTest`, `VerticaDialectTest`, new `XuguDialectTest`, `YashanDbDialectTest`), each adding `testAllKeyTableOmitsEmptyUpdateSet` (asserts the all-key SQL never contains `WHEN MATCHED THEN UPDATE SET` but still contains `WHEN NOT MATCHED`/`INSERT`) and, for all dialects except Vertica, `testPartialKeyTableStillUpdates` (regression guard that the common case still emits `WHEN MATCHED THEN UPDATE SET`). Vertica doesn't get a new partial-key test, but its pre-existing `testUpsertStatementByTableSchema` already asserts the exact partial-key MERGE text including `WHEN MATCHED THEN UPDATE SET`, so there is no real coverage gap there.

Crucially, `YashanDbDialectTest.testAllKeyTableOmitsEmptyUpdateSet` now exists and exercises exactly the case my last review flagged as missing (Issue 1 from my prior review) — I re-verified it calls `DIALECT.getUpsertStatement("test_db", "test_table", allFields, allFields)` with `DIALECT` being the shared `YashanDbDialect` instance, and asserts the same three conditions as the other dialects. That issue is now closed.

All of these are pure, deterministic string-assertion tests — no I/O, no `Testcontainers`, no timing, no shared mutable state, no ordering dependency between test methods.

**Flaky-test-risk rating: Stable.** No sleeps, no real DB connections, no ordering dependency, no external resources, and the dead-code/exception-throwing risk I flagged above for the `c3fc5704` version of `XuguDialect` is gone on the current head — I re-ran the trace on `XuguDialectTest.testAllKeyTableOmitsEmptyUpdateSet` against the current `XuguDialect.java` and confirmed the guard is gone, so the test now exercises the real `matchedClause` omission path rather than hitting a thrown exception.

## 2.3 Documentation Updates

None needed — internal SQL-generation correctness fix, no new/changed config option, default, or user-observable knob. `docs/en`/`docs/zh` correctly require no changes.

# 3. Architectural Soundness

## 3.1 Elegance of the Solution (Precise fix)

This remains a precise, minimal fix: it isolates exactly the fragment that becomes invalid when empty (`matchedClause`) instead of reworking the whole SQL-building method, applied uniformly across all seven affected dialects, and (with this update) it also removes the one piece of dialect-specific "handle it by throwing" logic (Xugu's old guard) that was inconsistent with the other six dialects' silent-failure behavior, unifying all seven onto the same correct behavior.

## 3.2 Maintainability

The `matchedClause` extraction makes "this can legitimately be empty" self-documenting, an improvement over the previous inline `%s` substitution. The ~6-line pattern is still duplicated across seven dialect classes (pre-existing per-dialect architecture, not introduced by this PR); a shared helper in `JdbcDialect` could reduce this in the future but is out of scope here.

## 3.3 Extensibility

No impact — no extension points added or removed; any future MERGE-based dialect can follow the now-established `matchedClause` pattern.

No serialization, checkpoint, protocol, or on-disk format is touched; no historical-version upgrade path is affected.

# 4. Issue Summary

|---|---|---|---|
| 1 (carryover, already resolved on current head) | `XuguDialect.getUpsertStatement()` retained a stale `throw new SeaTunnelException(...)` guard at the commit I last reviewed (`c3fc5704`), making the `matchedClause` fix unreachable dead code for Xugu and (per my own trace) making the new cross-dialect test fail for Xugu at that commit. Fixed in `6c91a30`/confirmed still fixed at current head `c0824e0`. | `seatunnel-connectors-v2/connector-jdbc/.../dialect/xugu/XuguDialect.java` (`getUpsertStatement`) | N/A — resolved, listed for traceability only |

No new open issues found in this re-review of the current head.

# 5. Merge Recommendation

1. Blockers — must be fixed: none currently open. My previous "Code style CI is failing" blocker was addressed by the `c0824e0` "fix code style" commit, and my previous "please fix YashanDbDialect too" recommendation has been implemented with matching test coverage. The one thing I did miss last time (the stale Xugu exception guard) turned out to already be fixed by the very next commit, so there is nothing outstanding on that front either.
2. Recommended fixes — non-blocking: none. I don't see anything else worth asking for here.

**One process note, not a blocker**: the CI run for the current head commit (`c0824e0`) is still `queued` on the fork at the time of this review, and I haven't been able to observe a completed "Build" conclusion for it yet (the two prior commits, `c3fc5704` and `6c91a30`, both failed on `Run / Code style`, and this commit's message says it addresses exactly that). Please make sure the fork's `Build` workflow finishes green on `c0824e0` before this actually merges — I'm confident in the source-level correctness of the change and am approving on that basis, but I have not personally observed a green CI run for this exact commit yet.

**Overall assessment**: This is now a complete, well-scoped, correctly-targeted bug fix across every MERGE-based JDBC dialect that had this defect (DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu, YashanDB), independently re-verified against `DB2Dialect` (unaffected, confirmed by source) and against the rest of the `connector-jdbc` module (no other dialect uses this pattern). The runtime path is genuinely the default path — `primary_keys` is even auto-derived from the catalog when unset, so no explicit user opt-in is required to hit this bug on an all-key table. Test coverage is stable and precisely targets the regression. Thank you again for the thorough follow-through on both review points, and apologies again for not catching the Xugu dead-code issue the first time around — the current head looks correct to me.

LeonYoah:
Good! This scenario really wasn’t taken into consideration.

DanielLeens:
Thanks @LeonYoah! Yeah, an all-key table (every column part of the primary/unique key, e.g. a composite-key junction table) is a legitimate but easy-to-miss schema shape — the `WHEN MATCHED THEN UPDATE SET` clause ends up with nothing to put in it, which is a straightforward SQL syntax error at write time.

For anyone following along: this PR is at commit `c0824e0` (no new commits since my last review), it's already approved from my side with no open blockers, and the fork's `Build` check has since gone green on that commit too. So this looks ready to merge as-is.

siwen-yu:
@DanielLeens Thanks for the review! All fixed, CI is green. Ready to merge. Could a maintainer please help merge? Thanks!

DanielLeens:
Thanks @siwen-yu — confirmed on my end: no new commits since my last review (still at `c0824e0`), the `Build` check is green, and I don't have any open technical blocker on this PR.

One clarification on the merge mechanics: my account here is a comment-only reviewer, so GitHub's branch protection doesn't count my `APPROVED` review toward the merge gate (that's why `reviewDecision` still shows `REVIEW_REQUIRED`). Actually merging will need a maintainer with write access to add their own approval and merge — I can't do that step myself.

Also worth flagging for whichever maintainer picks this up: #11861 ("Fix JDBC MERGE upsert SQL syntax for all-key tables" by @zhang-arvin) fixes the exact same underlying issue (#11729) and touches the identical dialect files (DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu — this PR additionally covers YashanDB). Only one of the two should be merged to avoid landing a duplicate/conflicting fix; a maintainer will need to pick one and close or rebase the other.

SEZ9:
Thanks everyone for the thorough back-and-forth here — this thread made the review very easy to follow.

@siwen-yu, thanks for turning around the YashanDbDialect fix and the spotless pass so quickly, and @DanielLeens, I really appreciate the end-to-end re-review at `c0824e0` and the candid note about the XuguDialect guard you spotted at `c3fc5704`.

Before I merge, two concrete asks:

1. **XuguDialect guard** — @siwen-yu, can you explicitly confirm that at `c0824e0` the pre-existing `nonUniqueKeyFields.isEmpty()` guard in `XuguDialect.getUpsertStatement()` (the one throwing `SeaTunnelException` before the `matchedClause` logic is reached) has been removed or adjusted, and that `JdbcAllKeyTableUpsertTest.testAllKeyTableOmitsEmptyUpdateSet()` passes for Xugu? @DanielLeens noted no open blockers on his side at `c0824e0`, but since his review comment surfaced this as dead code at `c3fc5704`, I'd like it confirmed on the record.

2. **Duplicate fix in #11861** — @DanielLeens flagged that #11861 addresses the same root cause (#11729) and touches the same dialect files. Since this PR additionally covers YashanDB and has already been reviewed end-to-end here, my inclination is to merge this one and close #11861 as a duplicate. If anyone sees a reason to prefer the other direction, please say so now.

On merge mechanics: understood that @DanielLeens's `APPROVED` review doesn't satisfy the branch protection gate — once item 1 is confirmed, I'll add my own approval with write access and merge.

Thanks again @siwen-yu, @DanielLeens, and @LeonYoah for the careful work on an easy-to-miss schema shape.

DanielLeens (approved):
Thanks for the ping, and sorry for the delay getting back to this thread — doing a full fresh re-review of the current head (`c0824e0`) rather than just diffing against my last look, per our usual process. Good news up front: nothing has moved since `c0824e0`, my prior findings all still check out, and I can now also confirm the two open questions from the thread.

- For the six JDBC MERGE-based dialects (DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu, and now YashanDB as well), `getUpsertStatement()` unconditionally emitted `WHEN MATCHED THEN UPDATE SET <updateSetClause>`. When a table's every column is part of the primary/unique key (a legitimate "all-key" schema, e.g. a composite-key junction table), `updateSetClause` is empty, producing invalid SQL (`UPDATE SET` with no body) that fails at write time with a DB-side syntax error.
- Fix: build the `WHEN MATCHED ... UPDATE SET ...` fragment as its own `matchedClause` and omit it entirely when there's nothing to update, so matched rows are left untouched and unmatched rows still insert normally.
- One-sentence summary: this closes a deterministic write-time SQL syntax error on the default upsert path for a real (if less common) table-key shape, with no config/API change.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Re-verified the diff at `c0824e0` file-by-file for all 7 dialects (`DmdbDialect`, `OracleDialect`, `SapHanaDialect`, `SqlServerDialect`, `VerticaDialect`, `XuguDialect`, `YashanDbDialect`):

```java
// before (all 7, structurally identical)
" ON (%s) " + " WHEN MATCHED THEN" + " UPDATE SET %s" + " WHEN NOT MATCHED THEN" + " INSERT (%s) VALUES (%s)"

// after
String matchedClause =
        StringUtils.isNotBlank(updateSetClause)
                ? String.format(" WHEN MATCHED THEN UPDATE SET %s ", updateSetClause)
                : "";
" ON (%s) " + "%s" /* matchedClause */ + " WHEN NOT MATCHED THEN" + " INSERT (%s) VALUES (%s)"
```

Key findings:
- The `matchedClause` construction and substitution is applied consistently and correctly across all 7 dialects; when `updateSetClause` is non-blank the generated SQL is byte-for-byte equivalent to the pre-fix output (verified against the pre-existing `VerticaDialectTest.upsertCreateTimeSQL` exact-string assertion, which still passes unchanged).
- **Xugu-specific change, now confirmed**: `XuguDialect.getUpsertStatement()` previously had a hard guard right after computing `nonUniqueKeyFields`:
  ```java
  if (nonUniqueKeyFields.isEmpty()) {
      throw new SeaTunnelException("The non-primary key field cannot be empty. Please set other fields");
  }
  ```
  I confirmed this guard is fully removed at `c0824e0` (not just at the intermediate `c3fc5704` commit) — `git diff dev...pr/11730 -- '**/xugu/XuguDialect.java'` shows the `if (nonUniqueKeyFields.isEmpty())` block deleted along with the now-unused `SeaTunnelException` import. I also confirmed no test anywhere in the repo (`dev` or this PR's history) asserted on that exception, so its removal breaks nothing — it's a clean, intentional behavior change that brings Xugu in line with the other 5 dialects, which never had this guard and were silently producing broken SQL instead. Confirmed `XuguDialectTest.testAllKeyTableOmitsEmptyUpdateSet` exercises exactly this path and passes with the guard gone.
- **Runtime path** (re-traced, unchanged from my prior review since this PR only touches SQL string generation, not the executor/builder layer): `JdbcOutputFormatBuilder.build()` takes the upsert path whenever `primaryKeys` is configured and `ENABLE_UPSERT` is true (default `true`, `JdbcSinkOptions.java`), calling `dialect.getUpsertStatementByTableSchema(...)` → `getUpsertStatement(database, tableName, tableSchema.getFieldNames(), pkNames)` with `fieldNames` = every column and `pkNames` = configured primary keys. So this sits on the default/normal upsert path for any all-key table, not an obscure edge branch — reconfirmed, not a new claim.
- Minor cosmetic note (pre-existing, not introduced by this PR): `Dmdb`/`Oracle`/`SapHana`/`Xugu`/`YashanDb` build `matchedClause` with a leading *and* trailing space (`" WHEN MATCHED THEN UPDATE SET %s "`), while `SqlServer`/`Vertica` only add a leading space. Since the surrounding template already starts the next segment with `" WHEN NOT MATCHED THEN"`, the first group produces a harmless double space between the clause and `WHEN NOT MATCHED` (also present in the pre-fix SQL between `ON (...)  WHEN MATCHED`, so this isn't new). Not functionally significant; not blocking.

## 1.2 Compatibility Impact — **Fully compatible**
For DM/Oracle/SAP HANA/SQL Server/Vertica/YashanDB: an all-key-table upsert config was *always* failing with a DB syntax error before this fix, so no running production job can regress — there was no working behavior to preserve. For Xugu specifically, the prior behavior was a hard `SeaTunnelException` at write time instead of a DB syntax error; that's also a job-failure path, so the same "no valid job depended on this" argument applies, and the new behavior (insert-if-unmatched, no-op-if-matched) is the same semantics as the other 5 dialects. No config/API/schema signature changes.

Pure string-building change in SQL construction; no new allocations of consequence, no new locks/retries/network calls, no idempotency change beyond the intended MERGE semantics (matched rows become true no-ops instead of a failed write). No concerns here.

No formal blocking issues found in this pass. See Issue Summary (§4) for the one non-blocking nit carried over from §1.1.

# 2. Code Quality Assessment

## 2.1 Coding Standards
The new `matchedClause` local variable is self-explanatory and each dialect carries an inline comment explaining *why* the clause must be conditionally omitted (the empty-`UPDATE SET` syntax-error rationale) — good, this satisfies our "explain why, not what" bar for a non-trivial control-flow change. No new public/protected methods or fields were introduced, so no additional Javadoc obligations here.

Every one of the 7 dialects now has a dedicated `testAllKeyTableOmitsEmptyUpdateSet` test (asserting the `WHEN MATCHED THEN UPDATE SET` fragment is absent, `WHEN NOT MATCHED` and `INSERT` are still present), and 6 of the 7 additionally have `testPartialKeyTableStillUpdates` (Vertica relies on its pre-existing exact-string `upsertCreateTimeSQL` test for the partial-key case instead, which is equivalent coverage). All are pure unit tests over deterministic string output — no `Thread.sleep`, no shared mutable state, no ordering dependence, no async/concurrency involved. **Stability rating: Stable.**
One paper-trail nit: the commit message for `6c91a30d2a` / `c3fc570440` and the earlier PR comment reference a `JdbcAllKeyTableUpsertTest` class as the home for this coverage, but the actual diff has no such standalone class — the coverage instead landed as per-dialect test methods inside each existing `*DialectTest`. Functionally this is fine (arguably better locality), just flagging so the paper trail matches the code for anyone reading history later. Non-blocking.

## 2.3 Documentation Updates
Not needed — this is an internal SQL-generation bug fix with no user-facing config/API/schema change, correctly reflected in the PR's own checklist.

# 3. Architectural Soundness

## 3.1 Precise fix / temporary workaround / long-term solution
This is a precise, long-term fix at the correct layer (SQL-string construction inside each dialect's `getUpsertStatement`), not a workaround bolted on elsewhere (e.g. not a config-validation-time rejection, not a caller-side special case). Right fix, right place.

## 3.2 Maintainability
Consistent pattern applied identically across all 7 dialects makes this easy to audit and easy to extend to any future MERGE-based dialect.

## 3.3 Extensibility
No structural concerns; the `matchedClause` pattern is trivially reusable if another dialect needs the same treatment.

## 3.4 Historical-version compatibility (restated)
As in §1.2 — fully compatible, no upgrade/migration considerations since the pre-fix behavior was a deterministic failure for this schema shape across all 7 dialects.

# 4. Issue Summary

| 1 | Commit-message / discussion mentions a standalone `JdbcAllKeyTableUpsertTest` class that doesn't exist in the diff; coverage instead lives in per-dialect test files | commit messages `6c91a30d2a`, `c3fc570440`; PR discussion | Low | No |

No High or Medium severity issues found in this re-review.

# 5. Merge Recommendation

1. Blockers — none. All previously-raised technical points (code style / spotless, YashanDB coverage, Xugu's dead-guard removal, the #11861 duplicate) are confirmed resolved as of `c0824e0`.
2. Recommended fixes — non-blocking: Issue #1 above (paper-trail only, no code change needed).

Answering the two open items directly, for the record:
- **Xugu guard**: confirmed removed at `c0824e0` (see §1.1) — the dead-code guard I flagged earlier is gone, and `XuguDialectTest.testAllKeyTableOmitsEmptyUpdateSet` exercises the now-working path.
- **Duplicate PR #11861**: it's now closed, so there's no conflicting fix in flight — no reason to prefer the other direction.

On merge mechanics (repeating for the record since it keeps coming up): my account here is comment-only, so this `APPROVED` review does not satisfy branch protection on its own — a maintainer with write access still needs to add their own approval to actually merge. From a technical-review standpoint, though, I have no remaining objection to merging this PR as-is.

Overall: solid, well-scoped bug fix with proportionate test coverage across every affected dialect, and the review thread did its job — the Xugu dead-guard and duplicate-PR questions are both closed out cleanly. Nice work turning around the extra dialect coverage and the style fixes, @siwen-yu.

siwen-yu:
@SEZ9 Yes, confirmed. At `c0824e0`,  the pre-existing `nonUniqueKeyFields.isEmpty()` guard in `XuguDialect.getUpsertStatement()` has been removed, so the `matchedClause` logic is reachable for all-key tables.
`XuguDialectTest.testAllKeyTableOmitsEmptyUpdateSet` passes as expected.
Thanks for the final check!

DanielLeens:
Thanks @siwen-yu and @SEZ9 — confirming from my side too, I re-checked the source at `c0824e0`: the `nonUniqueKeyFields.isEmpty()` guard that used to throw `SeaTunnelException` in `XuguDialect.getUpsertStatement()` is indeed gone in this PR's diff against `dev`, replaced by the same `matchedClause`-omission pattern used in the other dialects, and `XuguDialectTest.testAllKeyTableOmitsEmptyUpdateSet()` exercises exactly that path. So @siwen-yu's confirmation checks out against the actual diff, not just by description.

Also noting #11861 is now closed, so the duplicate-fix concern I raised is resolved as well.

No open technical blockers from me — still approved at `c0824e0`. Only remaining step is a maintainer with write access adding their own approval to satisfy branch protection and merging.

SEZ9:
Thanks @DanielLeens — answering your comment directly: yes, I'll take the remaining step you flagged. Your re-verification at `c0824e0` (the removed `nonUniqueKeyFields.isEmpty()` guard in `XuguDialect.getUpsertStatement()` replaced by the shared `matchedClause`-omission pattern, exercised by `XuguDialectTest.testAllKeyTableOmitsEmptyUpdateSet()`), together with @siwen-yu's confirmation, gives me what I need. I'll add my approval as a maintainer with write access to satisfy branch protection and proceed with the merge.

Also acknowledged that the duplicate-fix concern you raised earlier is resolved on your side, so there's no conflict to sort out before merging.

@siwen-yu — nothing further is needed from you; please just don't push any new commits to the branch so we merge exactly what was reviewed at `c0824e0`. If anything does change, flag it here so we can re-check before merging.

DanielLeens:
Thanks @SEZ9 — confirmed nothing further needed from my side either. Recapping where this leaves things: source at `c0824e0` has no open technical blocker from me, the Xugu guard removal is verified against the actual diff (not just description), and the #11861 duplicate is already closed. Go ahead with your approval and merge whenever you're ready — thanks for driving this one to closure.

SEZ9:
Thanks @DanielLeens for the thorough wrap-up, and apologies for the delayed reply.

To answer your comment directly: yes, nothing further is needed from you. Your recap matches my read of the thread — the Xugu guard removal is verified against the actual diff at `c0824e0` (with @siwen-yu's confirmation that `XuguDialectTest.testAllKeyTableOmitsEmptyUpdateSet` passes backing it up), and with #11861 closed the duplicate-fix concern is fully resolved.

Since your review can't satisfy the merge gate, I'll add my own approval on `c0824e0` and merge — no further changes are needed from @siwen-yu either.

Thanks @siwen-yu for the fix and @DanielLeens for the careful multi-pass review on all seven dialects.

DanielLeens:
Thanks @SEZ9! Glad the multi-pass review thread made this easy to close out. Nothing further from my side at `c0824e0` — go ahead and merge whenever you're ready. Thanks again to @siwen-yu for the fix.

nzw921rx (approved):
+1 LGTM

DanielLeens (commented):
Thanks for the ping — doing a full fresh re-review of the current head (`2d9f4c0ebbb5`) rather than just diffing against my last look, per our usual process for any re-review after a new commit lands.

Good news up front: the new commit is `2d9f4c0e`, "Merge branch 'dev' into dev" (authored by @nzw921rx, not @siwen-yu — presumably someone with push access on the fork keeping the branch in sync). I diffed it against the last head I fully reviewed and approved (`c0824e0`, twice — 2026-08-10 and 2026-08-21): `git diff` restricted to this PR's own file set (`gh pr diff --name-only` still lists exactly the same 14 files: the 7 dialect classes + 7 dialect test classes) shows **zero content change**. The merge only pulls in unrelated upstream `dev` commits; every line I already traced through `matchedClause`, the Xugu dead-guard removal, and the seven dialect tests is byte-for-byte identical to what I already verified twice. So this re-review changes nothing substantive — but I did re-read all 14 files again in full rather than assume the old conclusion still holds, as required.

- User pain point: for the seven JDBC MERGE-based dialects (DM, Oracle, SAP HANA, SQL Server, Vertica, Xugu, YashanDB), `getUpsertStatement()` unconditionally emitted `WHEN MATCHED THEN UPDATE SET <updateSetClause>`. When a table's every column is part of the primary/unique key (a legitimate "all-key" schema, e.g. a composite-key junction table), `updateSetClause` is empty, producing invalid SQL that fails at write time with a DB-side syntax error (or, for Xugu specifically before this PR, an application-level `SeaTunnelException`).
- Fix approach: build the `WHEN MATCHED ... UPDATE SET ...` fragment as its own `matchedClause` and omit it entirely when there's nothing to update, so matched rows are left untouched and unmatched rows still insert normally.
- One-sentence summary: this closes a deterministic write-time failure on the default upsert path for a real (if less common) table-key shape, with no config/API change, and — confirmed again on this head — the fix is complete and correct.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Representative before/after (`OracleDialect.java`, unchanged from prior rounds):
```java
// before
" ON (%s) " + " WHEN MATCHED THEN" + " UPDATE SET %s" + " WHEN NOT MATCHED THEN" + " INSERT (%s) VALUES (%s)"

// after
String matchedClause =
        StringUtils.isNotBlank(updateSetClause)
                ? String.format(" WHEN MATCHED THEN UPDATE SET %s ", updateSetClause)
                : "";
" ON (%s) " + "%s" /* matchedClause */ + " WHEN NOT MATCHED THEN" + " INSERT (%s) VALUES (%s)"
```

**Runtime path (re-confirmed unchanged):** `JdbcOutputFormatBuilder.build()` takes the upsert path whenever `primaryKeys` is configured and `ENABLE_UPSERT` is true (default `true`), calling `dialect.getUpsertStatementByTableSchema(...)` → `getUpsertStatement(database, tableName, tableSchema.getFieldNames(), pkNames)` with `fieldNames` = every column and `pkNames` = the primary keys (which, per `JdbcSinkFactory`, are auto-derived from the catalog table when not explicitly configured). So this sits on the default/normal upsert path for any all-key table, not an obscure edge branch. `XuguDialect.getUpsertStatement()` no longer contains the stale `throw new SeaTunnelException(...)` guard I flagged as dead-code-causing in my 2026-08-21 re-review — still gone, confirmed again on this head.

**Key findings (unchanged):**
- All 7 dialects apply the identical `matchedClause` pattern; for the common (non-empty `updateSetClause`) case, generated SQL is byte-for-byte equivalent to pre-fix output.
- `DB2Dialect` remains correctly out of scope — it builds `updateSetClause` from *all* field names, not a `nonUniqueKeyFields` filter, so it can never hit this bug class.
- No other MERGE-based dialect in `connector-jdbc` is missing this fix (re-confirmed via the module's file list on this head).

## 1.2 Compatibility Impact — **Fully compatible**
Unchanged from prior rounds. For every table with at least one non-key column, output is unchanged. For the previously-broken all-key case, behavior moves from "always fails" to "works correctly" — a strict improvement, no currently-working job can regress. No config/API/serialization/protocol change.

Pure string-building change, once per prepared-statement template build, not per row. No new allocations of consequence, no new connections/threads/retries/locks.

No logging touched. No open issues.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Each changed method carries a comment explaining *why* the clause must be conditionally omitted, consistent with the repo's "explain why" bar. No new public classes/fields requiring further Javadoc.

Every one of the 7 dialects has a dedicated `testAllKeyTableOmitsEmptyUpdateSet` plus (6 of 7; Vertica covered by its pre-existing exact-string test) `testPartialKeyTableStillUpdates`. All are pure, deterministic string-assertion unit tests — no I/O, no `Testcontainers`, no timing dependence, no shared mutable state. **Stability rating: Stable**, unchanged from prior rounds, re-confirmed by re-reading all 7 test classes on this head.

## 2.3 Documentation Updates
None needed — internal SQL-generation correctness fix, no user-facing config/API change.

# 3. Architectural Soundness

## 3.1 Elegance of the Solution — Precise fix
Isolates exactly the fragment that becomes invalid when empty, applied uniformly across all 7 affected dialects.

## 3.2 Maintainability
The `matchedClause` extraction is self-documenting. The ~6-line pattern is still duplicated per-dialect (pre-existing architecture, not introduced here); a shared `JdbcDialect` helper could reduce this later but is out of scope.

## 3.3 Extensibility
No impact — any future MERGE-based dialect can follow the established pattern.

No serialization/checkpoint/protocol impact; nothing to flag.

# 4. Issue Summary

No open issues. All findings from my prior three review rounds (Code style CI failure, missing YashanDB coverage, the Xugu dead-guard) were resolved by `c0824e0`, verified independently across two subsequent full re-reviews, and remain resolved on this head — the diff against `c0824e0` is empty for every file this PR touches.

# 5. Merge Recommendation

**CI diagnosis (the one thing that's new this round).** The apache-side `Build` pointer on this head shows `cancelled`. I traced the actual fork run (`siwen-yu/seatunnel` run `34565818330`, triggered by the `push` that created the merge commit): it ran for roughly 4 hours before being cancelled, taking two in-flight jobs down with it — `updated-modules-integration-test-part-6 (8, ubuntu-latest)` and `paimon-connector-it (8, ubuntu-latest)` — followed by a cascade of `skipped` downstream jobs. Neither cancelled job, nor any of the skipped ones, touches `connector-jdbc` or any file in this PR's diff. Critically, every job that *could* plausibly be affected by this PR completed successfully before the cancellation hit: `License header`, `Code style`, and `unit-test` on all four combinations (Java 8/11 × ubuntu/windows) — which includes the new/changed dialect tests — all passed. On top of that, I found a separate **scheduled** (nightly) workflow run against this *exact* commit SHA (`34585833605`, `schedule` event, not `push`) that completed with a full `success` conclusion a few hours later. That's independent, positive evidence that `2d9f4c0e` builds and passes cleanly — the earlier cancellation was a transient CI-scheduling/budget artifact of the `push`-triggered run, not a defect in this PR's code. I'd suggest a job-level rerun of just the `Build` workflow on the fork (or simply waiting for the next scheduled run to post a check) to get a clean, completed status on the required check so `mergeStateStatus` can clear — no code change is needed on the PR's side.

**Overall assessment:** this remains a complete, well-scoped, correctly-targeted bug fix across every MERGE-based JDBC dialect affected by this bug class, independently re-verified on three separate occasions now (twice on `c0824e0`, once again on this merge commit). +1 to @nzw921rx's "LGTM" from earlier today — I've traced the same code and agree with no reservations. Nothing further needed from @siwen-yu here; this is a CI/merge-gate matter, not a code one.

DanielLeens:
Thanks again for the continued diligence here. The current failing check (`updated-modules-integration-test-part-1`) is failing on `PostgresCDCIT.testPostgresCdcSnapshotOnlyAndCommittedOffsetStartupModes` (an awaitility `ConditionTimeout` waiting on a replicated row count) - a module this PR's diff (the JDBC MERGE all-key-table dialect fixes) never touches. I've seen this exact PostgresCDCIT test surface as an unrelated/environmental flake on another PR before.

`dev` has since picked up #12162 ("Reconstruct missing snapshot completion watermark and always wait for a checkpoint"), which touches exactly the CDC snapshot/checkpoint-completion logic this class of flake stems from - and this branch's base (7 commits behind `dev`) predates that fix.

Could you sync with the latest `dev` and push again so CI picks it up and reruns clean? No code-side blocker from my end.

PR #11738, opened 2026-08-10
[Improve][Connector-V2][Prometheus] Migrate batch size validation to OptionRule

goutamadwant:

Part of #11007.

This PR migrates the existing Prometheus sink `batch_size` validation to the declarative `OptionRule` path.

- Require `batch_size` to be greater than zero in `PrometheusSinkFactory`.
- Remove the redundant imperative check from `PrometheusSinkConfig`.
- Keep `batch_size` optional so configurations that use the default remain valid.
- Document the positive-value constraint in the option metadata and connector docs.
- Add focused tests covering omitted, positive, zero, and negative values.

Yes.

Invalid Prometheus sink `batch_size` values are now rejected during option validation with a structured validation error. The accepted range remains unchanged, and configurations that omit `batch_size` continue using the default value.

- An omitted `batch_size` is accepted.
- A positive `batch_size` is accepted.
- A zero `batch_size` is rejected.
- A negative `batch_size` is rejected.
- Validation errors identify the `batch_size` option.

Verified the complete Prometheus connector module with Java 11:

```shell
./mvnw -pl seatunnel-connectors-v2/connector-prometheus package
```

Result: 5 tests passed. Spotless, compilation, tests, and packaging completed successfully.

### Check list


DanielLeens (approved):

Before this change, the Prometheus sink validated `batch_size` imperatively: `PrometheusSinkConfig.loadConfig()` called a private `checkIntArgument()` helper that wrapped Guava's `checkArgument(args > 0)`. That check only ran when the sink writer was actually constructed (i.e. at task-start time on the worker), threw a plain `IllegalArgumentException` with no message, and lived outside the connector's declarative `OptionRule` (so the constraint was invisible to anything that introspects `optionRule()`, such as the web UI's option listing). This PR migrates that single constraint into the `OptionRule`/`Conditions` framework (`Conditions.greaterThan(BATCH_SIZE, 0)`), so the same rule is now declared once, enforced earlier (at job-submission-time config validation in `FactoryUtil`, not at writer construction), and reported with a structured, descriptive `OptionValidationException`. Docs (en/zh) and a new focused test class round out the change. One-sentence summary: this is a clean, low-risk migration of an existing validation rule from ad-hoc imperative code to the connector's declarative option-validation framework, with no change to the accepted value range.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Before** (`PrometheusSinkConfig.java`):
```java
if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
    int batchSize = checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
    sinkConfig.setBatchSize(batchSize);
}
...
private static int checkIntArgument(int args) {
    checkArgument(args > 0);
    return args;
}
```

**After**:
```java
if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
    sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
}
```
with the constraint moved to `PrometheusSinkFactory.optionRule()`:
```java
.optional(
        PrometheusSinkOptions.BATCH_SIZE,
        Conditions.greaterThan(PrometheusSinkOptions.BATCH_SIZE, 0))
```

I traced the actual runtime path to confirm the new check is genuinely reached on the normal job-submission path, not just in the new unit test:

```
FactoryUtil.createAndPrepareSink(...)          // seatunnel-api/.../table/factory/FactoryUtil.java:252
  -> ConfigValidator.of(context.getOptions()).validate(tableSinkFactory.optionRule())
       -> ConfigValidator.collectErrors(rule, ...)
            -> for each Condition in rule.getValueConstraints():
                 isConstraintApplicable(constraint, rule)   // true iff batch_size key present in raw config
                 validate(constraint) -> ConditionEvaluators.evaluate(GREATER_THAN, value, 0)
```
This runs during job parsing/preparation, well before `PrometheusSinkFactory.createSink()` constructs the `PrometheusSink`, and long before `PrometheusWriter` (which calls `PrometheusSinkConfig.loadConfig()`) is ever instantiated on a worker. So the fix is a genuine fail-fast improvement: an invalid `batch_size` is now rejected at submission time with a clear error, instead of surfacing later as an unhelpful `IllegalArgumentException` (no message) when the first task tries to start writing.

I also verified the "omitted" case is preserved correctly. `ConfigValidator.isConstraintApplicable()` only evaluates the constraint when `hasOption(BATCH_SIZE)` is true, and `ReadonlyConfig.getOptional()` (and therefore `hasOption`) only returns present when the key exists in the raw user config map — it does **not** treat the declared default value as "present". So when a user omits `batch_size`, the constraint is skipped entirely and the default (1024) is used untouched by validation, exactly matching the old `if (getOptional(...).isPresent())` guard. This is the same behavior as before, just relocated.

One more thing worth noting for reviewers unfamiliar with this newer `Conditions`/`valueConstraints` machinery: it was introduced in a separate, already-merged PR (present on `dev` prior to this change), so this PR is purely a consumer of existing, previously-reviewed infrastructure — it does not modify `OptionRule.java`, `ConfigValidator.java`, `Condition.java`, or `ConditionEvaluators.java` at all. That significantly reduces the review surface/risk here.

I also checked `PrometheusWriter` for defense in depth: `if (batchSize > 0 && batchList.size() >= batchSize)` (PrometheusWriter.java:110) already guards against non-positive `batchSize` at the point where it's used for size-based flush triggering. So even in a hypothetical scenario where an invalid value slipped past the new `OptionRule` check, it would not crash the writer — it would just silently disable size-based batching. That's pre-existing behavior, unaffected by this PR, and not something this PR needs to fix, but it's useful context for why this is not a correctness-critical validation (more a "give the user better/faster feedback" improvement than a "prevent a crash" fix).

## 1.2 Compatibility Impact

**Fully compatible**, with one minor, disclosed, user-facing nuance:

- Accepted value range for `batch_size` is unchanged (still `> 0`, default `1024`).
- The exception type changes from an unchecked Guava `IllegalArgumentException` (no message) to `OptionValidationException` (extends `SeaTunnelRuntimeException`, structured message including the option key). I checked for callers that specifically catch `IllegalArgumentException` around the sink-creation/config-validation call chain (`FactoryUtil`, `ClientExecuteCommand`, engine job submission) and found none that would change error-handling behavior as a result — the existing `catch (IllegalArgumentException ...)` blocks in the codebase are all in unrelated CLI-arg-parsing and REST-endpoint code, not in the job config validation pipeline.
- Failure now happens earlier (at job submission) rather than at first writer construction on a worker. This is a behavior change but a strictly beneficial one, and it's explicitly called out in the PR description's "user-facing change" section, which is the right way to handle it per the project's compatibility documentation expectations.

Negligible. This trades one `checkArgument` call for one `Conditions.greaterThan` evaluation that runs once per job submission (not per row, not in a hot path). No new allocations of consequence, no new dependencies, no threading/concurrency implications.

No blocking issues found in this area. The new validation path produces a clear, actionable message (verified via the new test's assertion that the exception message contains `PrometheusSinkOptions.BATCH_SIZE.key()`), which is a strict improvement over the previous message-less `IllegalArgumentException`.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The diff is minimal and idiomatic: it removes a now-redundant private helper and an unused static import (`checkArgument`) from `PrometheusSinkConfig.java`, and adds one import (`Conditions`) plus a two-line constraint registration in `PrometheusSinkFactory.java`. No new non-trivial methods or fields are introduced that would require additional Javadoc under the project's "important methods must be commented" guidance — the change is a net *removal* of a small private method, and the new `optional(...)` call sites are self-describing calls into existing, already-documented library code (`Conditions`, `OptionRule.Builder`). I have no coding-standard nits to raise here.

Good, focused coverage added in `PrometheusFactoryTest.java`:
- `testPositiveBatchSize` — a valid value (`1`) passes validation.
- `testBatchSizeIsOptional` — omitting `batch_size` entirely still validates successfully (covers the "use the default" path).
- `testNonPositiveBatchSizeFails` — both `0` and `-1` are rejected, and the exception message is asserted to contain the option key.

**Flaky-test-risk rating: Stable.** These are pure, deterministic unit tests over an in-memory `ReadonlyConfig`/`OptionRule` — no I/O, no time-based assertions, no concurrency, no external services, no shared mutable state between tests (`sinkConfig()` builds a fresh map each call). I see no plausible source of flakiness.

One very minor, non-blocking observation: `testNonPositiveBatchSizeFails` bundles both the `0` and `-1` cases into a single `@Test` method via a shared private helper (`assertInvalidBatchSize`). This is fine functionally (both assertions run independently and a failure in either is clearly attributable via the assertion message), just a minor style nit — not something I'd block on.

## 2.3 Documentation Updates

Both `docs/en/connectors/sink/Prometheus.md` and `docs/zh/connectors/sink/Prometheus.md` are updated in this PR, consistent with the "any user-visible change must update docs/en and docs/zh" rule. The English wording ("Positive number of rows buffered...") and the Chinese wording ("写入 Prometheus 前缓存的行数，必须大于 0") both correctly convey the `> 0` constraint and match the code's `Options.key("batch_size")...withDescription("the positive number of rows buffered before writing to prometheus")`. Good consistency across code, en docs, and zh docs — nothing further needed here.

# 3. Architectural Soundness

This is a precise, long-term-correct fix, not a workaround. It replaces a one-off imperative check with a declarative rule that is discoverable via `optionRule()` (useful for tooling such as the web UI's option listing) and evaluated centrally and consistently by `ConfigValidator`, alongside the connector's other options. This is exactly the kind of migration `Conditions`/`valueConstraints` was built for.

## 3.2 Maintainability

Improves maintainability: one less bespoke validation helper to maintain per-connector, and the constraint now lives next to the other declared options in `PrometheusSinkFactory`, making the full set of `batch_size` constraints visible in one place instead of split between the factory (rule declaration) and the config class (imperative enforcement).

## 3.3 Extensibility

No concerns. Should another numeric constraint need to be added to `batch_size` (or any other Prometheus sink option) in the future, it's now a one-line addition to the existing `.optional(option, condition)` builder call rather than another bespoke private static method.

No serialization, checkpoint/savepoint, or protocol surface is touched by this PR — it is confined to config-time option validation for a stateless HTTP-based sink. Existing SeaTunnel job configs with a valid or omitted `batch_size` continue to work unchanged; only configs that were already relying on an *invalid* `batch_size` (which would previously have crashed the writer at task start with an unhelpful, unchecked exception) are affected, and they now fail faster and more clearly at submission time — an intended and disclosed improvement, not a break of previously-working configurations.

# 4. Issue Summary

|---|---|---|---|
| — | No blocking issues found | — | — |

# 5. Merge Recommendation

1. Blockers — No blocking issues found.
   - Optional style nit: consider splitting `testNonPositiveBatchSizeFails` into two separate `@Test` methods (one for `0`, one for `-1`) for slightly clearer failure attribution in CI output, though the current combined form is perfectly acceptable as-is.

Overall assessment: this is a small, well-scoped, low-risk improvement that migrates an existing validation rule onto infrastructure the project already relies on elsewhere. I traced the full runtime path and confirmed the new check is genuinely reached on the normal job-submission path (via `FactoryUtil.createAndPrepareSink` → `ConfigValidator`), confirmed the "omitted `batch_size` uses the default without triggering validation" case is preserved exactly, checked for `IllegalArgumentException`-specific catch blocks that could be affected by the exception-type change (none relevant found), and reviewed both documentation files for accuracy. Test coverage is good and stable. I don't see an alternative implementation that would be meaningfully simpler or safer than what's here — nice, tidy contribution. The only remaining gate I see is CI finishing (the `Build` check was still queued at review time); once that's green this looks mergeable as-is.

DanielLeens (approved):
Thanks for keeping this branch fresh, @goutamadwant! I re-ran a complete review of the current diff (I don't just diff incrementally on re-review). Good news: the only change since my last pass is a routine merge of `dev` into this branch to pick up the latest history — `git diff dev...HEAD` is byte-for-byte the same six files I reviewed before (`PrometheusSinkConfig.java`, `PrometheusSinkOptions.java`, `PrometheusSinkFactory.java`, `PrometheusFactoryTest.java`, and the two docs files). Nothing new to flag there.

Since my earlier review already approved this, I want to make sure that wasn't too quick a pass, so this time I went back and specifically traced the old imperative check against the new `Conditions.greaterThan(...)` line by line, value by value, to make sure the accepted range and "when is this even checked" semantics are truly identical and not just superficially similar. Full trace below.

- Before this PR, `batch_size` validation for the Prometheus sink lived inside `PrometheusSinkConfig.loadConfig()` as an imperative Guava `checkArgument(args > 0)` call, invoked only when the sink writer actually loads its config (at task-start time on a worker), throwing a message-less `IllegalArgumentException`. The constraint was invisible to anything that introspects `optionRule()` (e.g. tooling, web UI option listings).
- This PR migrates that single constraint into the declarative `OptionRule`/`Conditions` framework via `Conditions.greaterThan(BATCH_SIZE, 0)` in `PrometheusSinkFactory.optionRule()`, so it's now enforced once, centrally, at job-submission-time config validation (`FactoryUtil` / `ConfigValidator`), with a structured `OptionValidationException` message instead of a silent `IllegalArgumentException`.
- One-sentence summary: a clean, low-risk migration of an existing `batch_size > 0` validation rule from ad-hoc imperative code onto the connector's declarative option-validation framework, with the accepted value range and "optional-only-when-present" semantics preserved exactly.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Before** (`PrometheusSinkConfig.java`):
```java
if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
    int batchSize = checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
    sinkConfig.setBatchSize(batchSize);
}
...
private static int checkIntArgument(int args) {
    checkArgument(args > 0);
    return args;
}
```

**After** (`PrometheusSinkConfig.java`):
```java
if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
    sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
}
```
with the constraint relocated to `PrometheusSinkFactory.optionRule()`:
```java
.optional(
        PrometheusSinkOptions.BATCH_SIZE,
        Conditions.greaterThan(PrometheusSinkOptions.BATCH_SIZE, 0))
```

**Line-by-line equivalence check (this was my main focus for this pass):**

1. **Accepted range.** Old: `checkArgument(args > 0)` — valid iff `args > 0` (strictly greater than zero; `0` and negatives rejected). New: `Conditions.greaterThan(BATCH_SIZE, 0)` compiles to `ConditionOperator.GREATER_THAN`, whose evaluator in `ConditionEvaluators.java` is `(v, c, cfg) -> v != null && compareNumbers(v, c.getExpectValue()) > 0`. That is exactly `v > 0`, i.e. strictly greater than zero. **Identical range**, not `>=` and not silently widened/narrowed.

2. **"Only validate when present" semantics.** Old guard: `pluginConfig.getOptional(BATCH_SIZE).isPresent()`. New: the option is registered via `.optional(BATCH_SIZE, condition)` (not `.required(...)`), which only adds it to `optionalOptions` (`OptionRule.java:346-356`) — it is *not* added to `requiredOptions`. `ConfigValidator.isConstraintApplicable()` (`ConfigValidator.java:290-298`) first checks `requiredOptions` (no match here, since `BATCH_SIZE` isn't required), then falls through to `anyOrSegmentFullyPresent(condition)` (`ConfigValidator.java:303-334`), which for a single, non-cross-field condition reduces to `hasOption(BATCH_SIZE)`. Critically, `hasOption()` (`ConfigValidator.java:411-413`) calls `config.getOptional(option).isPresent()`, and `ReadonlyConfig.getOptional()` (`ReadonlyConfig.java:103-124`) reads directly from the raw `confData` map and returns `Optional.empty()` when the key is absent — it does **not** fall back to `option.defaultValue()` (that fallback only happens in `ReadonlyConfig.get()`, via `getOptional().orElseGet(option::defaultValue)`, `ReadonlyConfig.java:65-66`). So the constraint is skipped entirely when `batch_size` is omitted, and the declared default (`1024`) is used untouched, exactly matching the old `if (getOptional().isPresent())` guard. **Identical "optional stays optional" semantics** — this migration does not accidentally turn `batch_size` into a required field, and it does not accidentally validate the default value itself.

3. **Default value and type unchanged.** `PrometheusSinkOptions.BATCH_SIZE` still declares `.intType().defaultValue(DEFAULT_BATCH_SIZE)` with `DEFAULT_BATCH_SIZE = 1024` — only the `withDescription(...)` text changed (now documents the `> 0` constraint). No change to type, default, or key name.

4. **Runtime reachability on the normal path.** I re-traced `FactoryUtil.createAndPrepareSink(...)` → `ConfigValidator.of(...).validate(tableSinkFactory.optionRule())` → the `getValueConstraints()` loop in `ConfigValidator.java:230-236` → `validate(constraint)` (`ConfigValidator.java:498`) → `ConditionEvaluators.evaluate(...)`. This runs at job-submission/parsing time, well before `PrometheusSinkFactory.createSink()` builds the `PrometheusSink`, and long before `PrometheusWriter` (which calls `PrometheusSinkConfig.loadConfig()`) is instantiated on a worker. So this is a genuine fail-fast improvement, not a change reachable only from a boundary/recovery path.

5. **Defense in depth, unaffected.** `PrometheusWriter.java:110` still guards `if (batchSize > 0 && batchList.size() >= batchSize)` at the point `batchSize` is used for flush triggering, so even a hypothetical invalid value that slipped past validation would just disable size-based batching rather than crash. Pre-existing, untouched by this PR.

**Key Findings:**
- Accepted value range for `batch_size` is byte-for-byte identical (`> 0`, strict) between old and new validation — verified against the actual `ConditionOperator.GREATER_THAN` evaluator, not just the `Conditions.greaterThan` factory method name.
- "Optional field, validated only when explicitly present" semantics are preserved exactly; `ReadonlyConfig.getOptional()` does not synthesize presence from the declared default, so omitted `batch_size` still bypasses the constraint and uses `1024` untouched.
- Failure now happens earlier (job submission) instead of at first writer construction on a worker, with a structured, descriptive `OptionValidationException` replacing a message-less `IllegalArgumentException` — a disclosed, backward-compatible improvement.
- `PrometheusSinkOptions.BATCH_SIZE` is not referenced anywhere else in the codebase (no other `.optional(BATCH_SIZE, ...)` call site, no collision with the `HttpSinkFactory` base class, which defines no `batch_size` option of its own), so there's no double-registration or duplicate-constraint risk.
- No source, engine, checkpoint, or serialization code is touched — this is entirely confined to the connector's `optionRule()` declaration and config-loading code.

Runtime path for a bad value (e.g. `batch_size = 0`):
```
Job submission
  -> FactoryUtil.createAndPrepareSink(...)
       -> ConfigValidator.of(context.getOptions()).validate(sinkFactory.optionRule())
            -> for constraint in optionRule.getValueConstraints():
                 isConstraintApplicable(BATCH_SIZE constraint) == true   (raw key present)
                 validate(constraint) -> ConditionEvaluators.evaluate(GREATER_THAN, 0, 0) == false
            -> collect error, throw OptionValidationException("batch_size", ...)
  (job never reaches worker; PrometheusWriter/PrometheusSinkConfig.loadConfig() never runs)
```

## 1.2 Compatibility Impact

**Fully compatible.**
- Accepted value range for `batch_size` is unchanged (`> 0`, default `1024`), confirmed by evaluator-level tracing above, not just by reading the factory-method name.
- Exception type changes from unchecked, message-less `IllegalArgumentException` to `OptionValidationException` (extends `SeaTunnelRuntimeException`) with a structured message containing the option key. I checked for `catch (IllegalArgumentException ...)` blocks anywhere near the sink-creation/config-validation call chain (`FactoryUtil`, job submission, `ClientExecuteCommand`) and found none that would change behavior — existing catches of that type are all in unrelated CLI-arg-parsing/REST-endpoint code.
- Failure now happens earlier (job submission) rather than at first writer construction on a worker — a strictly beneficial, and correctly disclosed, behavior change (the PR description explicitly calls this out under "Does this PR introduce any user-facing change?").

Negligible. Trades one `checkArgument` call (invoked once per writer instantiation, previously per-task) for one `Conditions.greaterThan` evaluation invoked once per job submission. No hot-path impact, no new allocations of consequence, no new dependencies, no threading/concurrency implications.

No blocking issues found. The new validation path produces a clear, actionable message that includes the option key (verified both by reading `ConfigValidator`'s `formatError(...)` call site and by the new test's explicit assertion that the exception message contains `PrometheusSinkOptions.BATCH_SIZE.key()`), a strict improvement over the previous message-less `IllegalArgumentException`.

# 2. Code Quality Assessment

## 2.1 Coding Standards (incl. doc-comment check)

The diff is minimal and idiomatic: removes a now-redundant private helper and its now-unused static import (`checkArgument`) from `PrometheusSinkConfig.java`, and adds one import (`Conditions`) plus a two-line constraint registration in `PrometheusSinkFactory.java`. No new non-trivial public methods or fields are introduced that would require additional Javadoc under the "important methods must be commented" guidance — this change is a net *removal* of a small private method, and the new `.optional(...)` call site is a self-describing call into existing, already-documented library code (`Conditions`, `OptionRule.Builder`). No coding-standard nits.

Good, focused coverage added in `PrometheusFactoryTest.java`:
- `testPositiveBatchSize` — a valid value (`1`) passes validation.
- `testBatchSizeIsOptional` — omitting `batch_size` entirely still validates successfully (covers the "use the default" path, which is the exact semantic this review focused on).
- `testNonPositiveBatchSizeFails` — both `0` and `-1` are rejected, with the exception message asserted to contain the option key.

**Stability rating: Stable.** These are pure, deterministic unit tests over an in-memory `ReadonlyConfig`/`OptionRule` — no I/O, no `Thread.sleep`, no time-based assertions, no concurrency, no external services, no shared mutable state between tests (`sinkConfig()` builds a fresh `HashMap` per call, and `ReadonlyConfig.fromMap(...)` is a fresh instance every time). No plausible source of flakiness.

Minor, non-blocking nit (unchanged from my last review): `testNonPositiveBatchSizeFails` bundles the `0` and `-1` cases into a single `@Test` via a shared helper. Fine functionally — not something to block on.

## 2.3 Documentation Updates

Both `docs/en/connectors/sink/Prometheus.md` and `docs/zh/connectors/sink/Prometheus.md` are updated, satisfying the "user-visible change must update both docs" rule. English ("Positive number of rows buffered before writing to Prometheus.") and Chinese ("写入 Prometheus 前缓存的行数，必须大于 0。") both correctly convey the `> 0` constraint and match `PrometheusSinkOptions.BATCH_SIZE`'s `withDescription(...)` text and the code's actual behavior. Consistent across code, en docs, and zh docs.

# 3. Architectural Soundness

Precise, long-term-correct fix, not a workaround. Replaces a one-off imperative check with a declarative rule discoverable via `optionRule()` and evaluated centrally by `ConfigValidator` alongside the connector's other options — exactly the use case `Conditions`/`valueConstraints` was built for.

## 3.2 Maintainability

Improves maintainability: one less bespoke validation helper per connector, and the constraint now lives next to the option's other metadata in `PrometheusSinkFactory` instead of being split between the factory (rule declaration, previously absent) and the config class (imperative enforcement).

## 3.3 Extensibility

No concerns. Any future numeric constraint on `batch_size` (or another Prometheus sink option) is now a one-line addition to the existing `.optional(option, condition)` builder call.

No serialization, checkpoint/savepoint, or protocol surface is touched — this is confined to config-time option validation for a stateless HTTP-based sink. Existing job configs with a valid or omitted `batch_size` continue to work unchanged; only configs that were already relying on an *invalid* `batch_size` (which previously crashed the writer at task start with an unhelpful, unchecked exception) are affected, and they now fail faster and more clearly at submission time — an intended, disclosed improvement, not a break of previously-working configurations.

# 4. Issue Summary

|---|---|---|---|
| — | No blocking issues found | — | — |

# 5. Merge Recommendation

1. **Blockers — must be fixed:** None.
   - Optional style nit: consider splitting `testNonPositiveBatchSizeFails` into two separate `@Test` methods (one for `0`, one for `-1`) for slightly clearer per-case failure attribution in CI output — the current combined form is perfectly acceptable as-is.

Overall assessment: I agree with my own prior "Ready to merge" verdict on this PR, and this re-review specifically stress-tested the one thing I want to be extra careful about on `OptionRule` migrations — whether the migrated check accepts exactly the same values and preserves the "optional stays optional" behavior as the code it replaces. It does: I traced `Conditions.greaterThan` down to the actual `ConditionOperator.GREATER_THAN` evaluator (`v > 0`, strict) and traced the presence-check down to `ReadonlyConfig.getOptional()` (raw-map-only, no default fallback), and both match the old imperative code exactly, value for value. The only outstanding item is CI, which is still queued on both the Apache-side pointer check and the fork's connector-specific jobs at the time of this review (unrelated jobs — license header, code style, changes, dead links — have already passed on the fork). Once that's green, this is good to merge as-is.

goutamadwant:
@nzw921rx can you review this and let me know if changes looks good? Builds look fine and green!

nzw921rx (dismissed):
+1 LGTM

good job,Thank you for your contribution🚀

nzw921rx:
One follow-up suggestion:

 I think we can also complete the default-value handling here.

Currently:

```java
if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
    sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
}
```

still means the declared default value is not applied when `batch_size` is omitted, because the assignment is skipped entirely.

I suggest simplifying it to:

```java
sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
```

DanielLeens:
@nzw921rx You're right, and thanks for catching this — my earlier approval missed it.

I traced the actual runtime behavior to confirm:

- `PrometheusSinkOptions.BATCH_SIZE` declares `defaultValue(DEFAULT_BATCH_SIZE)` where `DEFAULT_BATCH_SIZE = 1024`.
- `PrometheusSinkConfig` declares `private int batchSize;` with no explicit initializer, so its Java default is `0`.
- `ReadonlyConfig.getOptional(...)` returns `Optional.empty()` when the key is genuinely absent from user config — it does NOT fall back to the option's declared default (only `ReadonlyConfig.get(...)` does that, via `getOptional(option).orElseGet(option::defaultValue)`).

So in `PrometheusSinkConfig.loadConfig`, the current code:
```java
if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
    sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
}
```
skips the assignment entirely when `batch_size` is omitted, leaving `sinkConfig.batchSize` at `0`, not the documented `1024`.

That's not just cosmetic — `PrometheusWriter.write(Point)` guards size-based flushing with `if (batchSize > 0 && batchList.size() >= batchSize)`. With `batchSize == 0`, that condition is always false, so size-triggered flushing is silently disabled whenever a user omits `batch_size` (the common case), and rows only get flushed on the `flush_interval` timer (default 300000 ms = 5 min). That's an unbounded in-memory buffer for up to 5 minutes on any decent-throughput job, which is a real behavioral divergence from the documented default.

Your simplification is the correct fix:
```java
sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
```
This makes `get()` apply `orElseGet(option::defaultValue)` so the declared `1024` default is actually honored, while `OptionRule`'s `Conditions.greaterThan(BATCH_SIZE, 0)` (added in this PR) still validates any explicitly-supplied value.

Note this bug pre-dates this PR (the old `checkArgument` path had the same `isPresent()` guard), so it isn't something introduced by this change — but since we're already touching this exact block, I'd treat it as a small blocking fix rather than a follow-up: could you push one more commit applying the one-line simplification above? Everything else in this PR (the `OptionRule`/`Conditions` migration itself) still looks correct to me.

goutamadwant:
@DanielLeens @nzw921rx nice catch. updated this as suggested. batch_size is now loaded with get(BATCH_SIZE), so the declared default of 1024 is applied when the option is omitted. I also added a regression test for the default-value path. The Prometheus module passes locally with 6 tests on JDK 11. Let me know. thnks!

DanielLeens (approved):

This PR migrates Prometheus sink `batch_size` validation from an imperative `checkArgument` call inside `PrometheusSinkConfig.loadConfig()` to the declarative `OptionRule`/`Conditions` framework (`Conditions.greaterThan(BATCH_SIZE, 0)` in `PrometheusSinkFactory.optionRule()`), so the constraint is enforced earlier, at job-submission-time validation, with a structured `OptionValidationException` instead of a message-less `IllegalArgumentException` thrown later at worker start.

# 1. Code Change Review

## 1.1 Core Logic Analysis

This is my third pass on this PR. Since my last review (2026-08-11), one new commit was pushed: `6717dd04ca` — "[Fix][Prometheus] Apply the default batch size". This is exactly the fix @nzw921rx and I asked for after we found that the migration's `if (getOptional(BATCH_SIZE).isPresent())` guard left `PrometheusSinkConfig.batchSize` at Java's default `0` (not the documented `1024`) whenever a user omitted `batch_size` — silently disabling size-based flush triggering for the common case.

I re-verified the fix directly against the current head:

`seatunnel-connectors-v2/connector-prometheus/src/main/java/org/apache/seatunnel/connectors/seatunnel/prometheus/config/PrometheusSinkConfig.java:52` now reads:
```java
sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
```
replacing the old `isPresent()`-guarded block. `ReadonlyConfig.get(...)` applies `getOptional(option).orElseGet(option::defaultValue)`, so this now correctly falls back to `1024` when `batch_size` is omitted — matching exactly the one-line change @nzw921rx proposed and I confirmed as correct in my previous comment.

Crucially, this commit also adds a real regression test, `PrometheusFactoryTest.testDefaultBatchSize()` (`PrometheusFactoryTest.java:52-58`):
```java
PrometheusSinkConfig config =
        PrometheusSinkConfig.loadConfig(ReadonlyConfig.fromMap(sinkConfig()));
Assertions.assertEquals(
        PrometheusSinkOptions.BATCH_SIZE.defaultValue(), config.getBatchSize());
```
This calls `loadConfig()` directly (not just `validate()`), with a config map that omits `batch_size` (`sinkConfig()` at `PrometheusFactoryTest.java:75-81` only sets `URL`/`KEY_LABEL`/`KEY_VALUE`). That means this test genuinely exercises the exact bug path — I checked it would have failed against the pre-fix code (`config.getBatchSize()` would have been `0`, not `1024`) — so this isn't a superficial "doesn't throw" assertion, it actually pins the regression.

I re-diffed the rest of the PR against my last approval and confirmed no other files changed in this commit: only `PrometheusSinkConfig.java` (3 lines removed) and `PrometheusFactoryTest.java` (+10 lines) were touched. `PrometheusSinkOptions.java`, `PrometheusSinkFactory.java`, and both docs files are byte-identical to what I reviewed on 2026-08-11.

## 1.2 Compatibility Impact

Fully compatible, and this fix actually *restores* the pre-existing documented default behavior (`batch_size` defaulting to `1024`) that the earlier migration commit in this same PR had accidentally regressed. No API, config-key, or default-value change from the user's point of view versus the state before this whole PR started.

None. This is a control-flow simplification (removing a redundant `isPresent()` guard) with no new allocation, I/O, or locking.

No issues. `OptionValidationException` from the `OptionRule` check still fires for explicitly-invalid values (`<= 0`); the `get()` call added here only affects the default-fallback path and cannot itself throw for a well-formed option declaration.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Clean, minimal diff. No style issues.

`testDefaultBatchSize` (new), `testPositiveBatchSize`, `testBatchSizeIsOptional`, and `testNonPositiveBatchSizeFails` together now cover: default-applied-when-omitted, valid explicit value, omitted-is-not-required, and invalid values rejected with a message containing the option key. All are synchronous, in-process JUnit tests against `ConfigValidator`/`loadConfig()` directly — no threads, no timing, no shared static state, no containers.

**Stability rating: Stable.** No flaky-test anti-patterns present.

## 2.3 Documentation Updates

No new doc changes required by this commit; `docs/en` and `docs/zh` for `batch_size` (updated earlier in this PR) already correctly state the `1024` default and the `> 0` constraint, and that description is now actually true again on the code side.

# 3. Architectural Soundness

Correct and minimal — removing the redundant guard is the right fix rather than adding a compensating default-check elsewhere.

## 3.2 Maintainability

Improved: one less special-case branch to reason about, and a test now pins the default-fallback behavior so a future refactor can't silently reintroduce this regression.

## 3.3 Extensibility

No impact.

No impact — restores, rather than changes, historical default behavior.

# 4. Issue Summary

| - | (Prior blocker resolved) Default `batch_size` not applied when omitted | `PrometheusSinkConfig.java:52` (fixed in `6717dd04ca`) | Resolved |

No open issues.

# 5. Merge Recommendation

   None remaining. The single blocking issue from my previous review (default `batch_size` not applied) is fixed and covered by a regression test that actually exercises `loadConfig()`.
   None.

**CI status at time of this review:** `mergeStateStatus` is `BLOCKED` / `reviewDecision` is `REVIEW_REQUIRED` (expected, since the new commit dismissed prior approvals — that's exactly what this re-review resolves). The `Build` check for the current head (`6717dd04ca73f9a4293d9f15f06476e7f7ab9776`) was still `QUEUED` (not yet completed) as of this check — it has not failed, it simply hasn't finished running. Given the change is a 3-line production diff plus a self-contained unit test with no new dependencies, I don't see meaningful risk in the pending run, but please confirm the `Build` check finishes green before merging.

**Coordination flag:** I checked PR #11797 ("[Bug][Connector-V2][Prometheus] Apply batch_size default in loadConfig", head `b8fb8a9d65`) against this PR's diff, since both were flagged as touching Prometheus `batch_size` handling. They fix the *identical* bug in the *identical* method (`PrometheusSinkConfig.loadConfig()`), but #11797 branches from a `dev` snapshot that predates this PR's `OptionRule`/`Conditions` migration — its diff still contains the old `checkIntArgument(...)` imperative check that this PR (in an earlier commit) already removed in favor of `Conditions.greaterThan(BATCH_SIZE, 0)`. In other words, #11797 is not just parallel work — it is functionally superseded by this PR: once this PR merges, #11797's diff will either conflict on `PrometheusSinkConfig.java` or, if force-applied, reintroduce the now-removed `checkIntArgument`/Guava validation path alongside the new `OptionRule` one. I'd recommend closing #11797 as a duplicate (or rebasing it onto `dev` after this PR merges, at which point its code change becomes a no-op since this PR already applies the same fix) rather than merging both.

Overall: solid, minimal fix that directly addresses the regression found in review, backed by a real test. Approving pending the `Build` check going green.

DanielLeens (approved):
Re-reviewing because the head moved from `6717dd04` (the commit I last approved) to `fbd84fd1`. I did not just read the incremental commit — I re-diffed the whole PR against `dev` from scratch, re-read every touched source file in full, and re-traced the validation call chain, because the new head is a merge of `dev` that pulled in `#11797`, which independently touched the exact same method this PR rewrites. That overlap is the one thing genuinely worth re-checking here, and it is where the (minor) findings below come from.

Short version up front: the blocker from my last round is resolved twice over, no new blocker exists, and everything I flag below is non-blocking polish.

Before this change, the Prometheus sink enforced `batch_size > 0` imperatively: `PrometheusSinkConfig.loadConfig()` routed the value through a private `checkIntArgument()` helper wrapping Guava's `checkArgument(args > 0)`. That check only ran when the sink writer was constructed (task-start time on the worker), threw a bare `IllegalArgumentException` with no message, and lived outside the connector's declarative `OptionRule`, so nothing that introspects `optionRule()` could see the constraint. This PR migrates that single rule into the `OptionRule`/`Conditions` framework via `Conditions.greaterThan(BATCH_SIZE, 0)`, so it is declared once, enforced earlier (at job-submission-time config validation), and reported as a structured `OptionValidationException` naming the offending key. Docs (EN/ZH), the option description, and a focused factory test round it out. One sentence: a clean, low-risk migration of an existing validation rule from ad-hoc imperative code into the connector's declarative option-validation framework, with the accepted value range unchanged.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Precise change set at the current head** (7 files, +59/-29):

`PrometheusSinkConfig.java:52` — before (as it stands in `dev` today, after `#11797`):
```java
// Use get() (not getOptional().isPresent()) so the option's declared default is applied
// when batch_size is not configured; getOptional() does not fall back to the default, which
// would leave batchSize at 0 and disable the size-based flush trigger.
int batchSize = checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
sinkConfig.setBatchSize(batchSize);
...
private static int checkIntArgument(int args) {
    checkArgument(args > 0);
    return args;
}
```
after:
```java
sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
```

`PrometheusSinkFactory.java:56-58` — the constraint moves to the rule:
```java
.optional(
        PrometheusSinkOptions.BATCH_SIZE,
        Conditions.greaterThan(PrometheusSinkOptions.BATCH_SIZE, 0))
```

**Does the constraint actually fire, and only when it should?** I read `ConfigValidator` end to end rather than assuming. `collectErrors()` iterates `rule.getValueConstraints()` and, for each, first calls `isConstraintApplicable(constraint, rule)` (`ConfigValidator.java:290-299`). Because `BATCH_SIZE` is optional, that falls through to `anyOrSegmentFullyPresent()` (`ConfigValidator.java:305-337`), which requires every option referenced in the AND segment to satisfy `hasOption()` — and `hasOption()` is `config.getOptional(option).isPresent()` (`ConfigValidator.java:411-413`), which does **not** fall back to `Option#defaultValue()` (only `ReadonlyConfig.get()` does, via `getOptional(option).orElseGet(option::defaultValue)`, `ReadonlyConfig.java:65-67`). So:
- `batch_size` omitted -> constraint skipped, default `1024` applied by `loadConfig`'s `get()`. Correct, and covered by `testBatchSizeIsOptional`.
- `batch_size` present -> `ConditionEvaluators` `GREATER_THAN` evaluates `v != null && compareNumbers(v, 0) > 0` (`ConditionEvaluators.java:65-66`), so `0` and `-1` both fail. Correct, and covered by `testNonPositiveBatchSizeFails`.

**Runtime path: does the real user path reach the new check?** This is the question that decides whether the migration is equivalent or a silent weakening, so I traced it rather than assuming. `FactoryUtil.createAndPrepareSink()` calls `ConfigValidator.of(context.getOptions()).validate(tableSinkFactory.optionRule())` at `FactoryUtil.java:252`, and that method is the single sink-construction entry point for every engine: Zeta (`MultipleTableJobConfigParser.java:775`), Flink (`AbstractSinkExecuteProcessor.java:148`, `flink-13-starter/SinkExecuteProcessor.java:147`), Spark (`spark-starter-common/SinkExecuteProcessor.java:141`, `spark-2-starter/SinkExecuteProcessor.java:141`), plus `ConfigValidationUtils.java:372` for the `--check`/`--dry-run` path. I also checked the one branch inside `createAndPrepareSink` that would bypass validation — the legacy `fallback` branch at `FactoryUtil.java:228-239`, which calls `sink.prepare(config.toConfig())` without any `ConfigValidator` call. `isFallback()` (`FactoryUtil.java:465-485`) only returns `true` if the factory is absent, or if the probe `sinkFactory.createSink(null)` throws `UnsupportedOperationException` with the specific sentinel message. `PrometheusSinkFactory.createSink(null)` dereferences `context.getOptions()` and throws `NullPointerException`, which is caught, logged at debug, and yields `false`. So Prometheus never takes the fallback path, and every production job submission reaches the new declarative check. The migration is behaviour-preserving for the user-visible outcome, and strictly better on timing (submission-time instead of writer-construction-time) and on message quality (`OptionUtil.formatError` produces a message naming `batch_size` and the failed constraint, versus the previous messageless `IllegalArgumentException`).

**What the merge with `dev` changed, and why it matters here.** `#11797` (`[Bug][Connector-V2][Prometheus] Apply batch_size default in loadConfig`, merged as `4875a82d`) landed in `dev` the same fix I asked for in my last round, plus a new `PrometheusSinkConfigTest`. Your own commit `6717dd04` had already fixed it inside this PR. The merge resolution keeps `get()` semantics — correct — but drops the three-line explanatory comment `#11797` added, and drops one of `#11797`'s three tests. Neither changes behaviour; both are worth a second look (Issues 1 and 2).

## 1.2 Compatibility Impact

**Backward compatible.** The accepted value range is identical (`> 0` before, `> 0` after). No option renamed, no default changed (`DEFAULT_BATCH_SIZE = 1024` untouched at `PrometheusSinkOptions.java:26`), no public API or SPI signature changed, no serialization or state format touched, so checkpoint/savepoint restore and rolling upgrade are unaffected. The only observable differences for an existing job are strictly improvements: an invalid `batch_size` now fails at submission instead of at task start on a worker, and the error identifies the option instead of being an empty-message `IllegalArgumentException`. Jobs that omit `batch_size`, or set any positive value, are bit-for-bit unaffected. No `incompatible-changes.md` entry is needed, and the PR correctly does not add one.

One genuinely non-obvious upgrade case I checked: a job that today somehow carries `batch_size = 0` and is currently failing at writer construction will now fail earlier, at submission. That is the same outcome one stage sooner, not a new failure for a previously-working job.

Nil. The condition is evaluated once per sink creation during job submission; nothing is added to any per-record path. `write()`'s hot loop (`batchSize > 0 && batchList.size() >= batchSize`) is untouched. No new dependency; the change actually drops the `Preconditions.checkArgument` shaded-Guava import from this class.

No logging added or changed. Error handling improves as described in 1.1. The findings below are the complete list; all are Low and non-blocking.

**Issue 1 — the `get()`-vs-`getOptional()` guard comment from `dev` is dropped in the merge resolution**
- **Location**: `seatunnel-connectors-v2/connector-prometheus/src/main/java/org/apache/seatunnel/connectors/seatunnel/prometheus/config/PrometheusSinkConfig.java:52`
- **Problem**: `dev` currently carries a three-line comment (added by `#11797`, commit `4875a82d`) explaining precisely why this line must use `get()` and not `getOptional().isPresent()`. This PR's merge resolution keeps the `get()` call but deletes the comment, so the reasoning is gone from the tree.
- **Potential risk**: This is exactly the bug `#11797` just fixed, and the failure mode is silent — `getOptional()` does not apply the declared default, `batchSize` stays `0`, and `write()`'s `batchSize > 0` guard disables the size-based flush trigger with no error anywhere. The surrounding lines all use the `getOptional(...).isPresent()` idiom, so the `get()` line now reads as the odd one out and is an easy target for a well-intentioned "consistency" refactor. That is precisely what the deleted comment was there to prevent, and it also fits the project's rule that comments should explain the constraint and the risk of getting it wrong.
- **Best improvement**: Restore the comment verbatim above the line. It costs three lines and applies to the post-PR code without modification.
- **Severity**: Low

**Issue 2 — `loadConfig()` no longer rejects a non-positive `batch_size` on its own, and the test asserting that was removed**
- **Location**: `seatunnel-connectors-v2/connector-prometheus/src/main/java/org/apache/seatunnel/connectors/seatunnel/prometheus/config/PrometheusSinkConfig.java:52`; `seatunnel-connectors-v2/connector-prometheus/src/test/java/org/apache/seatunnel/connectors/seatunnel/prometheus/PrometheusSinkConfigTest.java` (`shouldRejectNonPositiveBatchSize`, deleted by this PR)
- **Problem**: Removing `checkIntArgument()` is the point of the PR and is correct, but it does collapse two layers of enforcement into one. After this change, `PrometheusSinkConfig.loadConfig()` accepts `0` or a negative value without complaint; only `optionRule()` validation rejects it.
- **Potential risk**: If any future caller ever reaches `loadConfig()` without first running `ConfigValidator.validate(optionRule())`, `batchSize` becomes `0` and the size-based flush trigger is silently disabled — the same silent failure mode as Issue 1. I want to be explicit that this is **not** a live bug today: I verified `loadConfig()` has exactly one production caller (`PrometheusWriter`'s constructor), that its only construction path is `PrometheusSink.createWriter()` via `PrometheusSinkFactory.createSink()`, and that every one of the six sink-creation call sites listed in 1.1 validates the option rule first, with the fallback branch proven unreachable for this connector. It is a defence-in-depth reduction, not a reachable defect.
- **On test strictness specifically**, since deleting an assertion always deserves scrutiny: the deleted `shouldRejectNonPositiveBatchSize` asserted `loadConfig` throws for `0`. That behaviour no longer exists at that layer by design, so keeping the test would mean keeping the code it tests. It is replaced by three tests at the layer where the rule now lives (`PrometheusFactoryTest.java:42,47,52` — accepted-when-omitted, accepted-when-positive, rejected for both `0` and `-1`), and `assertInvalidBatchSize` additionally asserts the message names the option key, which the old test did not. Net user-visible coverage is equal or better, and the deletion is justified rather than a weakening. I am satisfied on this point; recording the analysis so it is on the record rather than assumed.
- **Best improvement**: Optional. Either leave as-is (acceptable), or add a one-line comment on `PrometheusSinkConfig.java:52` noting that the positivity constraint is enforced by `PrometheusSinkFactory#optionRule()` so a reader does not conclude the check was simply lost.
- **Severity**: Low

**Issue 3 — the `batch_size` description loses its "maximum/threshold" meaning in both docs and the option metadata**
- **Location**: `docs/en/connectors/sink/Prometheus.md:56`; `docs/zh/connectors/sink/Prometheus.md:53`; `seatunnel-connectors-v2/connector-prometheus/src/main/java/org/apache/seatunnel/connectors/seatunnel/prometheus/config/PrometheusSinkOptions.java:47`
- **Problem**: EN changed from "Maximum number of rows buffered before writing to Prometheus" to "Positive number of rows buffered before writing to Prometheus", and ZH from "写入 Prometheus 前最多缓存的行数" to "写入 Prometheus 前缓存的行数，必须大于 0". Both dropped the word that carried the actual semantics — this is a flush *threshold*, an upper bound on the buffer, not a fixed count of rows that are always buffered. The option's `withDescription` at `PrometheusSinkOptions.java:47` has the same problem.
- **Potential risk**: A reader of the new EN text can reasonably read "Positive number of rows buffered" as "this many rows are always held in the buffer". The two languages also now say different things: ZH still states the constraint explicitly ("必须大于 0"), EN only implies it via the adjective. Docs are a user-facing contract here, and the option description surfaces in tooling that introspects `optionRule()`.
- **Best improvement**: Keep the original noun phrase and append the constraint, so both languages carry both facts. EN: "Maximum number of rows buffered before writing to Prometheus. Must be greater than 0." ZH: "写入 Prometheus 前最多缓存的行数，必须大于 0。" And `withDescription("the maximum number of rows buffered before writing to prometheus, must be greater than 0")`.
- **Severity**: Low

**Issue 4 — merge-order coordination with `#11778` (informational, no action needed in this PR)**
- **Location**: `PrometheusSinkConfig.java` (`loadConfig` body); `PrometheusSinkFactory.java` (`optionRule()` builder chain)
- **Problem**: `#11778` (Prometheus `FlushSignal` migration) edits the same two blocks — it removes the `flush_interval` block from `loadConfig` and the `.optional(FLUSH_INTERVAL)` line from the `optionRule()` chain, immediately adjacent to the lines this PR rewrites. Both PRs are currently `mergeable: true` against `dev` in isolation.
- **Potential risk**: Whichever merges second will hit a textual conflict in both files. The semantic resolution is trivial and non-overlapping (this PR owns the `batch_size` lines, `#11778` owns the `flush_interval` lines), but it should be resolved deliberately rather than by a mechanical "take ours". Worth noting that the last rebase of `#11778` already had to resolve exactly this file once, against `#11797`.
- **Best improvement**: Nothing to change here. Flagging so whoever merges second knows the conflict is expected and what the correct resolution looks like. I am raising the same note on `#11778`.
- **Severity**: Informational

# 2. Code Quality Assessment

## 2.1 Coding Standards

Spotless-clean formatting, no wildcard imports, and the now-unused `Preconditions.checkArgument` static import is correctly removed rather than left dangling. `Conditions` is imported from the proper API package. The new test methods are small, single-purpose, and named for the behaviour they assert; helper extraction (`sinkConfig()`, `validate()`, `validateBatchSize()`, `assertInvalidBatchSize()`) keeps them readable without hiding the assertion. No `@DisplayName` misuse, no `System.out`, no emoji in comments or messages, license headers intact on the touched files. On comment completeness the one gap is Issue 1 above — this diff removes an existing explanatory comment without adding one, which is the only place the change reads thinner than what it replaced.

Coverage of the changed behaviour is complete at the layer the behaviour now lives in: omitted (`testBatchSizeIsOptional`), positive (`testPositiveBatchSize`), and both non-positive boundaries `0` and `-1` (`testNonPositiveBatchSizeFails`), with `assertInvalidBatchSize` also asserting the thrown `OptionValidationException` message contains `PrometheusSinkOptions.BATCH_SIZE.key()`. That last assertion is a real improvement over the deleted test, which only asserted `Exception.class` — a strictly weaker check that would have passed on any failure at all, including an unrelated NPE.

**Stability rating: Stable.** `PrometheusFactoryTest` builds a `ReadonlyConfig` from an in-memory `HashMap` and calls `ConfigValidator` synchronously. No threads, no sleeps, no timeouts, no I/O, no network, no containers, no filesystem, no shared mutable static state, no ordering dependency between test methods (each builds its own map via `sinkConfig()`). There is no plausible flake vector. Empirically confirmed on the current head's CI run: `PrometheusFactoryTest` ran 4 tests in 0.54s and `PrometheusSinkConfigTest` 2 tests in 0.002s, both clean, and the `SeaTunnel : Connectors V2 : Prometheus` module reported `SUCCESS`.

No E2E was added or modified, and none is warranted — this is submission-time config validation with no runtime data path, so an E2E would only re-assert what the factory test already covers deterministically.

## 2.3 Documentation Updates

EN and ZH are both updated and stay structurally in sync (same table, same row). Content-wise they now say slightly different things — see Issue 3. Beyond that, the docs match the code: default `1024` matches `DEFAULT_BATCH_SIZE`, the option remains `No` (optional), and no other row is disturbed.

# 3. Architectural Soundness

**Precise fix.** This is the intended use of the `OptionRule`/`Conditions` framework, not a workaround: the constraint is declared once, next to the option it constrains, and the framework handles evaluation, applicability, message formatting, and key collection for unknown-key detection. It also removes a private helper whose only job was to re-implement, worse, what the framework already provides. As a unit of the broader `#11007` effort, it is exactly the right shape and size — one connector, one constraint, no collateral refactoring.

## 3.2 Maintainability

Improved. Five lines of imperative validation plus a private helper are replaced by one declarative clause, and the constraint is now visible to anything that introspects `optionRule()` instead of being buried in a config loader. The single maintainability regression is the dropped comment (Issue 1), which is a three-line fix.

## 3.3 Extensibility

Good. Adding an upper bound later is a `.and(Conditions.lessOrEqual(BATCH_SIZE, N))` on the same clause, with no change to `loadConfig`. The pattern is identical to what other connectors in `#11007` are adopting, so a reader comparing connectors will find consistent shapes.

No concerns, verified in 1.2: no option renamed or removed, no default changed, no public API/SPI signature changed, no serialization or checkpoint state format touched, and no behavioural change for any currently-valid configuration. A job saved under an older version, restored from a savepoint, exported/imported, or copied continues to behave identically. The only configurations whose behaviour changes are ones that were already failing.

# 4. Issue Summary

| 1 | The `get()`-vs-`getOptional()` guard comment that `#11797` added to `dev` is dropped by this PR's merge resolution, while the `get()` call it protects is kept. That comment documents the exact silent bug (`batchSize == 0` disables the size-based flush trigger) that `#11797` just fixed, and the surrounding lines all use the `getOptional().isPresent()` idiom this line must not adopt. | `PrometheusSinkConfig.java:52` | Low |
| 2 | Removing `checkIntArgument()` leaves `optionRule()` as the sole enforcement layer, and `#11797`'s `shouldRejectNonPositiveBatchSize` test is deleted. Verified not reachable today (single production caller; all six sink-creation call sites validate the rule first; the `isFallback` bypass proven inapplicable to this connector), and the replacement factory tests are strictly stronger than the deleted one. Defence-in-depth reduction only. | `PrometheusSinkConfig.java:52`; `PrometheusSinkConfigTest.java` (deleted test) | Low |
| 3 | `batch_size` description loses its "maximum/threshold" meaning in EN docs, ZH docs, and the option's `withDescription`; EN and ZH now state different things (ZH keeps the explicit "> 0" constraint, EN only implies it). | `docs/en/connectors/sink/Prometheus.md:56`; `docs/zh/connectors/sink/Prometheus.md:53`; `PrometheusSinkOptions.java:47` | Low |
| 4 | Expected textual merge conflict with `#11778`, which edits the immediately adjacent lines in both `loadConfig` and the `optionRule()` chain. Semantically non-overlapping; noted so the second merge is resolved deliberately. | `PrometheusSinkConfig.java`; `PrometheusSinkFactory.java` | Informational |
| - | (Was blocking, now Resolved) `batch_size` default not applied when the option is omitted, leaving `batchSize == 0` and silently disabling the size-based flush trigger. Raised by @nzw921rx and confirmed by me last round. Fixed in this PR by `6717dd04`, and independently fixed in `dev` by `#11797`; the merge at `fbd84fd1` preserves `get()` semantics. Verified at the current head. | `PrometheusSinkConfig.java:52` | Resolved (was blocking) |

# 5. Merge Recommendation

No blockers. Everything above is Low or informational, and none of it affects correctness, compatibility, or runtime behaviour. The one blocking item from my last round is resolved and re-verified at the current head.

   - None.

   - Issue 1 (Low): restore the three-line `get()`-vs-`getOptional()` comment from `dev` above `PrometheusSinkConfig.java:52`. This is the one I would most like to see, because it guards a silent failure mode against a plausible future refactor and costs nothing.
   - Issue 3 (Low): put "Maximum" back and state the `> 0` constraint explicitly in EN docs, ZH docs, and `withDescription`, so both languages carry both facts.
   - Issue 2 (Low): optionally, a one-line pointer at `PrometheusSinkConfig.java:52` noting the constraint now lives in `optionRule()`.

**On other reviewers**: @nzw921rx approved at `72d85b1e` and raised the `batch_size` default-application point that became the blocker of the last round — that is now fixed both here and in `dev`, and I am recording a +1 to their original diagnosis rather than restating it. I found nothing in their review I disagree with, and nothing they raised is still open.

**On CI**: `Build` is red on `fbd84fd1`, and I traced it rather than assuming. The failure is `org.apache.seatunnel.engine.server.rest.RestApiSubmitJobConfigShadeDecryptTest` on `unit-test (8, windows-latest)`, failing in `setUp` (line 91) with Hazelcast `java.lang.IllegalStateException: Node failed to start!` after ~311s; the other two `unit-test` legs were cancelled as a consequence of that failure, not on their own merits. That is the same Windows-runner-specific engine flake I traced on the sibling Prometheus PR earlier this week, in `seatunnel-engine-server`, a module this diff does not touch. In the same job, `SeaTunnel : Connectors V2 : Prometheus` built and tested green: `PrometheusFactoryTest` 4/4, `PrometheusSinkConfigTest` 2/2, `PrometheusParamCheckTest` 1/1. So this is not a code problem, but the PR still needs a green run before it can actually land — a rerun of the failed job should be enough.

**Merge-order note**: `mergeable` is `true` and `mergeable_state` is `blocked`, which is the maintainer-approval gate, not a technical item pending from me. Please see Issue 4 on the expected conflict with `#11778`.

Thanks for turning the last round around so quickly, and for the extra regression test on the default-value path — that, plus `#11797` landing the same fix upstream, means this exact bug now has coverage on both sides of the merge.

davidzollo (dismissed):
+1 if CI passes.
LGTM

goutamadwant:
@DanielLeens synced with current dev and resolved the Prometheus conflicts around the new FlushSignal path. This keeps the engine-level flush behavior from dev and only moves batch_size > 0 validation into OptionRule.

DanielLeens (approved):
- User pain point: The Prometheus sink's `batch_size` validation lived inside `PrometheusSinkConfig.loadConfig()` as an imperative Guava `checkArgument(args > 0)` call, invoked only when the sink writer was actually constructed (task-start time on a worker). It threw a bare, message-less `IllegalArgumentException`, and the constraint was invisible to anything that introspects the connector's declarative `optionRule()` (tooling, web UI option listings, `--check`/`--dry-run`).
- Fix approach: Migrate the `batch_size > 0` constraint into the `OptionRule`/`Conditions` framework (`Conditions.greaterThan(BATCH_SIZE, 0)` in `PrometheusSinkFactory.optionRule()`), so it is declared once, evaluated centrally at job-submission-time by `ConfigValidator`, and reported via a structured `OptionValidationException` naming the offending key. Docs (EN/ZH) and a focused test class round it out.
- One-sentence summary: a clean, low-risk migration of an existing `batch_size > 0` validation rule from ad-hoc imperative code onto the connector's declarative option-validation framework, with the accepted value range and "optional stays optional" semantics preserved exactly, now re-verified at the current head (`e49489596c6b5a5e1fb3739a65294ae535957a4c`) after a routine `dev` merge.

# 1. Code Change Review

## 1.1 Core Logic Analysis

This is a re-review of the **current PR head**. Daniel (DanielLeens) reviewed this PR four times previously (2026-08-10, 2026-08-11, 2026-08-14 x2), with the most recent approval at commit `fbd84fd1467c71507b528da5d4c818423a8e11ff` (2026-08-14T10:47:44Z). Since then, exactly one new commit landed: `e49489596c6b5a5e1fb3739a65294ae535957a4c`, headlined "Merge branch 'dev' into GH-11007-prometheus" (2026-08-15T00:29:01Z), confirmed by the author's own comment on the PR: *"synced with current dev and resolved the Prometheus conflicts around the new FlushSignal path. This keeps the engine-level flush behavior from dev and only moves batch_size > 0 validation."*

I independently re-verified this rather than trusting the description. Diffing `fbd84fd1` against the current head via `gh api repos/apache/seatunnel/compare/...` shows the merge pulled in a large, unrelated batch of `dev` commits (CDC offset/reader changes, JDBC, Kafka transaction sender, transforms-v2 numeric functions, and — critically — PR #11778, "Prometheus FlushSignal migration", which rewrote `PrometheusWriter.java` (+100/-72) and touched `PrometheusSink.java`, removing the connector-level `flush_interval` option entirely in favor of an engine-level `sink.flush.interval`/`FlushSignal` mechanism). This is exactly the merge conflict Daniel predicted and flagged as "Issue 4 (Informational)" in his last review, when both this PR and #11778 were independently touching adjacent lines in `PrometheusSinkConfig.java` / `PrometheusSinkFactory.java`.

The important question for this re-review is whether that merge resolution preserved this PR's own change untouched. I fetched the full current content of all five Prometheus files directly via the GitHub contents API at the current head and confirm:

- `PrometheusSinkConfig.java:51` — `sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));` — byte-identical to what Daniel reviewed at `fbd84fd1`.
- `PrometheusSinkFactory.java:160-162` — `.optional(PrometheusSinkOptions.BATCH_SIZE, Conditions.greaterThan(PrometheusSinkOptions.BATCH_SIZE, 0))` — byte-identical.
- `PrometheusSinkOptions.java:96-101` — `BATCH_SIZE` option, description "the positive number of rows buffered before writing to prometheus", default `1024` — byte-identical.
- `PrometheusWriter.java:278` — `if (batchSize > 0 && batchList.size() >= batchSize) { flush(); }` — the size-based flush-trigger guard this PR's fix protects downstream — still present and still guards against a `batchSize` of `0`, now inside the rewritten (post-#11778) `write(Point)` method. `batchSize` is still populated from `sinkConfig.getBatchSize()` (line 237), i.e. from the exact `loadConfig()` call this PR's core change touches.
- `gh pr diff 11738` (current head vs. current `dev` merge-base) shows this PR's own 7-file diff is **unchanged** from what Daniel reviewed at `fbd84fd1`: same 2 doc lines, same `PrometheusSinkConfig.java` (1 addition/12 deletions), same `PrometheusSinkOptions.java`, same `PrometheusSinkFactory.java`, same two test files. `PrometheusWriter.java` and `PrometheusSink.java` do **not** appear in this PR's own diff — they are identical to `dev`, confirming the FlushSignal rewrite came in purely via the merge and this PR did not need to (and did not) touch them.

**Conclusion: nothing was introduced by the new commit that changes this PR's own logic.** The merge is a clean, non-overlapping resolution — this PR owns the `batch_size` lines, #11778 owned the (now-removed) `flush_interval` lines, exactly as Daniel anticipated. This re-review therefore re-confirms, rather than newly discovers, the correctness already established in the prior rounds:

- **Accepted range unchanged**: `Conditions.greaterThan(BATCH_SIZE, 0)` compiles to `ConditionOperator.GREATER_THAN`, evaluated as `v > 0` — identical to the old `checkArgument(args > 0)`.
- **"Optional stays optional"**: `ConfigValidator.isConstraintApplicable()` only evaluates the constraint when `hasOption(BATCH_SIZE)` is true (`ReadonlyConfig.getOptional()` reads the raw config map only, no default fallback), so an omitted `batch_size` skips validation and `loadConfig()`'s `get()` call applies the declared default `1024` — matching pre-PR behavior exactly.
- **Runtime reachability**: `FactoryUtil.createAndPrepareSink()` → `ConfigValidator.of(...).validate(optionRule())` is the sink-construction entry point for every engine (Zeta, Flink, Spark) and the `--check`/`--dry-run` path; the legacy fallback branch (`FactoryUtil.isFallback()`) is proven unreachable for `PrometheusSinkFactory` (it throws `NullPointerException`, not the sentinel `UnsupportedOperationException`, when probed with a `null` context).

Runtime path for an invalid `batch_size` (e.g. `0`), unchanged from before:
```
Job submission
  -> FactoryUtil.createAndPrepareSink(...)
       -> ConfigValidator.of(context.getOptions()).validate(sinkFactory.optionRule())
            -> isConstraintApplicable(BATCH_SIZE constraint) == true   (raw key present)
            -> ConditionEvaluators.evaluate(GREATER_THAN, 0, 0) == false
            -> collect error, throw OptionValidationException("batch_size", ...)
  (job never reaches a worker; PrometheusWriter / PrometheusSinkConfig.loadConfig() never runs)
```

## 1.2 Compatibility Impact

**Fully compatible.** Accepted value range for `batch_size` is unchanged (`> 0`, default `1024`). No option renamed, no default changed, no public API/SPI signature changed. The exception type changes from a message-less `IllegalArgumentException` to a structured `OptionValidationException`; no `catch (IllegalArgumentException ...)` block anywhere in the job-submission call chain (`FactoryUtil`, `ClientExecuteCommand`, engine submission) depends on the old type. Failure now happens earlier (job submission instead of writer construction) — a disclosed, strictly beneficial change per the PR description. No serialization, checkpoint/savepoint, or protocol surface is touched.

Negligible. One `Conditions.greaterThan` evaluation per job submission, not per row. No new allocations, no new dependencies, no threading/concurrency implications. The unrelated `PrometheusWriter`/`PrometheusSink` rewrite pulled in via the `dev` merge (#11778) is out of this PR's scope and not this PR's responsibility to re-review in depth, but the one line this PR's fix depends on downstream (`batchSize > 0` guard at `PrometheusWriter.java:278`) is confirmed intact.

No new issues found in this pass. All items below are carried over from Daniel's most recent (2026-08-14T10:47:44Z) review and remain in their previously-assessed state — none were touched by the intervening `dev`-merge commit.

**Issue 1 — dropped `get()`-vs-`getOptional()` guard comment (Carryover, previously raised by Daniel, still open)**
- Location: `seatunnel-connectors-v2/connector-prometheus/src/main/java/org/apache/seatunnel/connectors/seatunnel/prometheus/config/PrometheusSinkConfig.java:51`
- Problem: An earlier `dev` commit (#11797) had added a 3-line comment explaining why this line must use `get()` (applies the declared default) rather than `getOptional().isPresent()` (which would leave `batchSize` at Java's default `0` and silently disable size-based flush triggering). This PR's merge resolution keeps the `get()` call but the comment is still absent.
- Potential risk: The surrounding lines in the same method all use the `getOptional(...).isPresent()` idiom, making this line look like the odd one out — an easy target for a future "consistency" refactor that reintroduces the exact silent bug #11797 fixed.
- Best improvement: Restore the explanatory comment above line 51.
- Severity: Low
- Raised by another reviewer: No (raised by Daniel in his own prior review round; carried forward here since it is unresolved at the current head)

**Issue 2 — `loadConfig()` no longer independently rejects a non-positive `batch_size` (Carryover, previously raised by Daniel, verified not a live bug)**
- Location: `PrometheusSinkConfig.java:51`
- Problem: Removing `checkIntArgument()` collapses two enforcement layers into one; `loadConfig()` itself no longer rejects `0`/negative values, relying entirely on `optionRule()` validation upstream.
- Potential risk: Purely theoretical today — verified (again, at the current head) that `loadConfig()` has exactly one production caller (`PrometheusWriter`'s constructor), reached only via `PrometheusSink.createWriter()` / `PrometheusSinkFactory.createSink()`, and every sink-creation call site across all three engines validates `optionRule()` first. Defense-in-depth reduction only, not a reachable defect.
- Best improvement: Optional — a one-line comment noting the constraint now lives in `optionRule()`.
- Severity: Low

**Issue 3 — `batch_size` description loses its "maximum/threshold" meaning (Carryover, previously raised by Daniel, still open)**
- Location: `docs/en/connectors/sink/Prometheus.md:56`; `docs/zh/connectors/sink/Prometheus.md:53`; `PrometheusSinkOptions.java:100-101`
- Problem: EN text changed from "Maximum number of rows buffered..." to "Positive number of rows buffered..."; ZH kept the explicit "必须大于 0" constraint but also dropped "最多". EN and ZH now emphasize different facts, and neither EN nor the option's `withDescription` states this is an upper-bound flush threshold, not a fixed buffered count.
- Potential risk: A reader can misread "positive number of rows buffered" as "this many rows are always held," rather than "up to this many."
- Best improvement: EN: "Maximum number of rows buffered before writing to Prometheus. Must be greater than 0." ZH: "写入 Prometheus 前最多缓存的行数，必须大于 0。" Update `withDescription` to match.
- Severity: Low

**Issue 4 (from prior round) — merge conflict with #11778 — now Resolved**
- The expected textual conflict Daniel flagged as informational in his last review has materialized and been resolved cleanly by the author's `dev` merge, verified above in 1.1: this PR's `batch_size` lines and #11778's `flush_interval` removal did not collide, and both changes coexist correctly at the current head.

## 2. Code Quality Assessment

## 2.1 Coding Standards

Diff is minimal and idiomatic, unchanged from the prior review: a now-redundant private helper and its static import are removed, one import (`Conditions`) and a two-line constraint registration are added. No new non-trivial public methods/fields requiring additional Javadoc. Spotless/AOSP formatting intact (fork's "Code style" CI job passes independently of the flaky unit-test job — see 5 below).

Unchanged from the prior round: `PrometheusFactoryTest` covers omitted (`testBatchSizeIsOptional`), positive (`testPositiveBatchSize`), and both non-positive boundaries `0`/`-1` (`testNonPositiveBatchSizeFails`, asserting the exception message contains the option key). `PrometheusSinkConfigTest.shouldRejectNonPositiveBatchSize` was intentionally deleted (see Issue 2) since that behavior no longer lives at that layer; the replacement coverage at the `optionRule()` layer is strictly stronger (asserts message content, not just "any exception").

**Stability rating: Stable.** Pure, deterministic, synchronous unit tests over an in-memory `ReadonlyConfig`/`OptionRule` — no I/O, no timing, no concurrency, no shared mutable state. I independently confirmed via the current fork CI run (`goutamadwant/seatunnel` run `31853770013`, head `e494895`) that `PrometheusFactoryTest` (4/4), `PrometheusSinkConfigTest` (2/2), and the newly-arrived `PrometheusWriterTest` (4/4, brought in by the #11778 merge, not part of this PR's own diff) all pass cleanly.

## 2.3 Documentation Updates

EN/ZH both updated, structurally in sync (same table row). Content wording issue tracked as Issue 3 above (Low, carryover).

# 3. Architectural Soundness

**Precise fix, long-term-correct.** Uses the `OptionRule`/`Conditions` framework exactly as intended: one declarative clause, evaluated centrally, replacing a bespoke imperative check. Confirmed still the right shape after absorbing an unrelated, much larger rewrite (#11778) of the surrounding file via merge — the PR's own footprint remains a clean, minimal, single-purpose change.

## 3.2 Maintainability

Improved: one less bespoke validation helper; the constraint is now visible via `optionRule()` introspection. The one maintainability regression already known (Issue 1, dropped comment) remains a 3-line fix.

## 3.3 Extensibility

No concerns. Adding further constraints on `batch_size` (or other options) is a one-line `.optional(option, condition)` addition, consistent with the pattern other connectors in the parent effort (#11007) are adopting.

No serialization, checkpoint/savepoint, or protocol surface touched. Existing job configs with a valid or omitted `batch_size` are unaffected; only configs that were already relying on an invalid value (previously crashing at writer construction) now fail earlier and more clearly at submission — an intended, disclosed improvement, not a regression for previously-working configurations. The unrelated FlushSignal rewrite pulled in via the `dev` merge is itself already-merged, previously-reviewed `dev` history and out of this PR's compatibility scope.

# 4. Issue Summary

|---|---|---|---|
| 1 | Dropped `get()`-vs-`getOptional()` guard comment explaining why this line must not adopt the `getOptional().isPresent()` idiom used elsewhere in the method | `PrometheusSinkConfig.java:51` | Low |
| 2 | `loadConfig()` no longer independently rejects a non-positive `batch_size`; relies solely on `optionRule()` upstream (verified not reachable today) | `PrometheusSinkConfig.java:51` | Low |
| 3 | `batch_size` description in EN docs, ZH docs, and `withDescription` loses the "maximum/threshold" framing; EN and ZH now state slightly different things | `docs/en/connectors/sink/Prometheus.md:56`; `docs/zh/connectors/sink/Prometheus.md:53`; `PrometheusSinkOptions.java:100-101` | Low |

All three issues are carryovers from Daniel's most recent (2026-08-14) review; none were introduced or affected by the latest `dev`-merge commit, and none are newly discovered in this pass.

# 5. Merge Recommendation

   None.

   - Issue 1 (Low): restore the 3-line `get()`-vs-`getOptional()` explanatory comment at `PrometheusSinkConfig.java:51`.
   - Issue 3 (Low): restore "maximum"/threshold framing in EN docs, ZH docs, and `withDescription`, stating the `> 0` constraint explicitly in both languages.
   - Issue 2 (Low): optional pointer comment noting the constraint now lives solely in `optionRule()`.

**CI status at current head (`e49489596c6b5a5e1fb3739a65294ae535957a4c`):** The Apache-side "Build" pointer check shows `FAILURE`; I traced this to the real fork run (`goutamadwant/seatunnel` run `31853770013`) rather than trusting the pointer. Four `unit-test` matrix legs failed/were cancelled; the root cause is `unit-test (11, windows-latest)`: `org.apache.seatunnel.engine.server.resourcemanager.WorkerTagTest` failing in `setUp` with `java.lang.IllegalStateException: Node failed to start!` after ~311s — the well-established Windows/Hazelcast node-startup flake in `seatunnel-engine-server`, a module this diff does not touch (and a pattern Daniel already identified and dismissed on this same PR's prior CI runs, twice, under different failing test names). The other three `unit-test` legs were cancelled as a consequence of the matrix strategy failure, not on their own merits. `Prometheus`-module tests (`PrometheusFactoryTest` 4/4, `PrometheusSinkConfigTest` 2/2, `PrometheusWriterTest` 4/4) all passed cleanly within this same run before the cancellation. `License header`, `Code style`, `Dead links`, and all `updated-modules-integration-test` legs passed. This is not a code problem introduced by this PR; a rerun of the failed `unit-test` job (all four legs, since three were cancelled rather than independently failed) should be sufficient, consistent with prior rounds on this same PR.

**Note on review process:** Since Daniel's last approval, `davidzollo` dismissed his review (2026-08-14T14:51:22Z) ahead of the subsequent `dev`-merge push, so the PR currently shows `mergeStateStatus: BLOCKED` / review required — this is a process/administrative gate, not a technical objection, and this re-review re-establishes that the underlying code is unchanged and still sound.

Overall assessment: this re-review confirms the PR's own diff is byte-identical to what was already thoroughly reviewed and approved four times previously; the only change since the last approval is a routine, correctly-resolved `dev` merge that absorbed an unrelated Prometheus rewrite (#11778) without disturbing this PR's `batch_size` validation logic. All three carried-over Low-severity style/doc nits remain open and worth a follow-up commit, but none block merge. The one CI red is the same pre-existing Windows/Hazelcast environmental flake seen (and correctly triaged) on this PR's earlier CI runs, unrelated to the Prometheus module or this diff.

DanielLeens (commented):
CI follow-up (2026-08-22): my last review (2026-08-15) noted the Apache-side `Build` pointer was showing `FAILURE` on this exact head (`e494895`), traced to a Windows/Hazelcast `Node failed to start!` flake in `seatunnel-engine-server` unrelated to this diff, and recommended a rerun of the failed job.

I re-checked live: the head is unchanged (`e49489596c6b5a5e1fb3739a65294ae535957a4c`), and both the Apache-side `Build` check and the fork's own run (`goutamadwant/seatunnel` run `31853770013`) now show a clean `success` conclusion. So that CI note is resolved and no longer accurate as a current blocker.

No change to my conclusion: no code-side blockers, only the three previously-noted Low-severity, non-blocking style/doc nits (restore the `get()`-vs-`getOptional()` comment, restore the "maximum/threshold" wording in the `batch_size` docs, optional pointer comment). Still Ready to merge from my side.

nzw921rx (approved):
+1 good job

PR #11740, opened 2026-08-10
[Improve][Connector-V2] Align Markdown RAG metadata with Knowledge Sync

yzeng1618:
## Purpose

#10990 introduced the Knowledge Sync logical metadata registry and allowed
`MetadataTransform` to project declared metadata from `SeaTunnelRow.options`.
The Markdown source still emitted only its physical RAG columns, so a real
Markdown pipeline could not use the common metadata contract without synthetic
catalog tables and row options.

This PR adds a connector-local bridge from Markdown RAG output to Knowledge
Sync logical metadata while preserving all existing physical fields and routing
formulas.

## Changes

- Declare `SourceUri`, `DocumentId`, `DocumentHash`, and `ChunkHash` in the
  Markdown source metadata schema when `markdown_rag_metadata_enabled=true`.
- Populate the four logical keys in every emitted Markdown row's independent
  `SeaTunnelRow.options` map without clearing unrelated options.
- Build a credential-free logical source identity:
  - preserve the existing local-path normalization for local paths and valid
    `file:` URIs;
  - preserve the scheme, host, explicit port, and path of hierarchical remote
    URIs;
  - remove user info, the complete query component, and the fragment;
  - fail unsafe URI sanitization without echoing the original sensitive value.
- Build logical `DocumentId` as
  `doc_ + lower_hex(SHA-256(UTF-8(logical SourceUri)))`.
- Calculate `DocumentHash` from the exact source bytes during the existing
  stream read, before UTF-8 decoding and Markdown parsing, without a second
  source read.
- Reuse the existing immediate-row text hash as logical `ChunkHash`.
- Merge registry-defined metadata columns deterministically while preserving
  existing metadata, reusing compatible definitions, and rejecting
  incompatible same-name definitions.
- Cover bounded existing-file discovery, empty-path fallback, continuous
  discovery, and multi-table catalog construction.
- Add real Markdown catalog/row coverage to `MetadataTransform`, including
  successful alias projection and existing physical-name collision rejection.
- Extend the LocalFile Markdown E2E assertion configuration with logical
  metadata projection.
- Update the English and Chinese File Source and Metadata Transform
  documentation.

## Compatibility

- No configuration option, dependency, physical column, or public API is added
  or renamed.
- The behavior remains gated by the existing disabled-by-default
  `markdown_rag_metadata_enabled` option.
- Existing physical `source_uri`, `document_id`, `chunk_id`, `chunk_index`, and
  `content_hash` fields retain their names, order, values, formulas, and source
  routing behavior.
- Non-Markdown formats and disabled Markdown RAG configurations retain their
  current metadata-schema behavior.
- Signed or credential-bearing URLs can intentionally produce different
  physical and logical document IDs: the physical ID remains the compatibility
  value, while the logical ID excludes credentials and transient URI
  components.

## Semantics and limitations

- The generic bridge removes the complete remote URI query. Resources whose
  identity exists only in query parameters need a stable, non-sensitive path
  or a future source-specific identity mechanism.
- Logical `ChunkHash` identifies the immediate Markdown output row only. A
  downstream transform that changes or splits text must recompute final
  `ChunkHash`, `ChunkId`, and `ChunkIndex` before a lifecycle sink consumes the
  rows.
- This bridge does not expose logical `ChunkId` or `ChunkIndex` from the legacy
  physical formulas.
- Incremental comparison, final-writer affinity, stale-chunk deletion,
  tombstones, and lifecycle sink behavior are outside this PR.

## Review follow-up

The follow-up commit `7c1e91e94` addresses the requested error-handling and
scope corrections:

- Preserve discovery/read diagnostics with a sanitized cause chain that keeps
  exception type names and stack traces without retaining credential-bearing
  exception messages.
- Gate Markdown URI sanitization in both single-table and multi-table readers
  behind `markdown_rag_metadata_enabled`, preserving the previous error
  context and original cause for all other formats.
- Keep the existing two-argument `BaseFileSourceReader` constructor as a
  compatibility path with sanitization disabled.
- Add single-table, multi-table, and discovery-failure regressions for enabled
  and disabled paths, including credential non-disclosure and retained stack
  traces.
- Add direct physical-vs-logical identity divergence assertions for a signed
  URI and document the new non-trivial helpers and fail-fast validation calls.

## Verification

- Root `./mvnw spotless:apply`: passed.
- Final review regression suite: 28 tests, 0 failures, 0 errors, 0 skipped.
- A full `connector-file-base` run before the final cause-chain-only refinement
  passed: 206 tests, 0 failures, 0 errors, 33 skipped.
- A later full-module rerun was environment-blocked after the affected tests
  passed because the Windows C: drive had 0 bytes available; the unrelated
  `ExcelGeneratorTest` then reported disk-space write errors.
- `git diff --cached --check`: passed before commit.

The root `./mvnw -q -DskipTests verify` reactor is not reported as passing: the
local Maven process ended with `Malformed \\uxxxx encoding`. The full root suite
and LocalFile E2E suite should therefore be covered by CI.


DanielLeens (changes requested):

Before this PR, the Markdown RAG output (`markdown_rag_metadata_enabled=true`) only exposed *physical* row columns (`source_uri`, `document_id`, `chunk_id`, `chunk_index`, `content_hash`). Meanwhile the platform has a separate, format-agnostic "Knowledge Sync" contract (`KnowledgeSyncMetadataField`: `SourceUri`, `DocumentId`, `DocumentHash`, `ChunkHash`, …) that downstream lifecycle/RAG-sync sinks consume through row *options* plus a declared `MetadataSchema`, projected by the `Metadata` transform. The two worlds didn't talk to each other for Markdown: a Markdown-based pipeline could not participate in Knowledge Sync without hand-rolling the logical identity fields itself.

This PR bridges that gap: when `file_format_type=markdown` and `markdown_rag_metadata_enabled=true`, the Markdown source now also (a) declares the four Knowledge Sync metadata columns on its `CatalogTable`, (b) computes a credential-free "logical" source URI/document id/document hash/chunk hash, and (c) writes them into `SeaTunnelRow#getOptions()` so the existing `Metadata` transform can project them downstream — all without touching the existing physical RAG columns, their values, or file-routing formulas. In one sentence: *this PR makes Markdown RAG output a first-class Knowledge Sync metadata producer, purely additive and gated behind the pre-existing `markdown_rag_metadata_enabled` flag.*

# 1. Code Change Review

## 1.1 Core Logic Analysis

**New identity bridge (`MarkdownKnowledgeSyncMetadata`, new file):**
`canonicalizeSourceUri()` turns a raw path/URI into a credential-free logical identity: local paths and `file:` URIs are normalized via `Paths.get(...)`, hierarchical remote URIs (`scheme://host[:port]/path`) are lower-cased on scheme/host and stripped of user-info, query, and fragment; anything the code cannot safely reduce (opaque URIs, unparsable strings that still "look like" a URI, invalid `file:` URIs) throws a generic `"Cannot derive a safe logical source identity..."` exception rather than risk leaking a credential. `buildDocumentId()`/`safeSourceContext()` build on top of it; `withMetadata()` merges the 4 bridge columns into a `CatalogTable`'s `MetadataSchema`, validating type/nullability if a column with the same logical name already exists.

**Runtime path traced:**
- `BaseFileSourceConfig`/`BaseFileSource` constructors (config/schema build time) → eagerly call `canonicalizeSourceUri(FILE_PATH)` purely to fail fast on unsafe roots, then merge the bridge metadata columns into the produced `CatalogTable` (`BaseFileSourceConfig.java:73-83`, `BaseFileSource.java:79-81,137-140`).
- `MarkdownReadStrategy.read()` (per-file, per-task runtime): computes `logicalSourceUri`/`logicalDocumentId` once per file, wraps the file's `InputStream` in a `DigestInputStream` so the whole-document SHA-256 (`DocumentHash`) is computed for free while the content is read into `markdown` for parsing, then for every emitted row calls `addKnowledgeSyncMetadata()` to stash `SourceUri/DocumentId/DocumentHash/ChunkHash` into `row.getOptions()` — entirely separate from the pre-existing `appendRagMetadata()` physical-column path, which is untouched in its computation.
- `Metadata` transform (pre-existing, unmodified in this PR) projects the row-option values to user-chosen output columns.

This is a clean, minimal-surface design: the digest is computed streaming (no extra full buffering), the bridge class is stateless, and `MarkdownReadStrategyTest#shouldPreserveExistingRowOptionsWhenAddingKnowledgeSyncMetadata` confirms unrelated existing row options are not clobbered.

**Deliberate, well-documented identity split — not a bug:**
I initially suspected a correctness bug because the *physical* `document_id`/`source_uri` columns are still computed via the older `FileSourceDocumentRouting.buildDocumentId(FileSourceDocumentRouting.normalizeSourceUri(path))` (which does **not** strip credentials/query/fragment from non-`file:` URIs), while the *new* logical `DocumentId`/`SourceUri` option values go through the new credential-free `canonicalizeSourceUri()`. For a signed/credential-bearing remote URL these two identities can differ. However, this is explicitly called out as intended in `docs/en/transforms/metadata.md` ("Markdown Source Bridge" section) and in every touched connector doc: *"The logical identity is separate from the compatibility identity... physical values and split-routing formula remain unchanged."* The PR even ships a naming-collision guard in `MetadataTransform` and documents the required aliasing pattern (`ks_source_uri`, `ks_document_id`). Good design call — just note the test/E2E coverage gap below (2.2).

**Runtime path diagram (Markdown + `markdown_rag_metadata_enabled=true`):**
```
BaseFileSourceConfig/BaseFileSource ctor
  -> canonicalizeSourceUri(root path)         [fail-fast validation, return value discarded]
  -> MarkdownKnowledgeSyncMetadata.withMetadata(CatalogTable)  [adds 4 metadata columns]

MarkdownReadStrategy.read(path, ...)
  -> canonicalizeSourceUri(path) -> logicalSourceUri
  -> buildDocumentId(logicalSourceUri) -> logicalDocumentId
  -> DigestInputStream(hadoopFileSystemProxy.getInputStream(path)) -> markdown text + documentHash
  -> per row: appendRagMetadata(...) [[existing]] physical source_uri/document_id/chunk_id/...
           + addKnowledgeSyncMetadata(...) [[new]]  row.getOptions()[SourceUri/DocumentId/DocumentHash/ChunkHash]

downstream Metadata transform -> projects row.getOptions() to aliased output columns
```

## 1.2 Compatibility Impact

**Partially incompatible** (pending fix of Issue 2 below) — everything specific to Markdown is opt-in and additive:
- `MARKDOWN_RAG_METADATA_ENABLED` already existed pre-PR, default `false`, unchanged. No new config option is introduced.
- Physical RAG columns, their values, and split-routing behavior are provably unchanged (test `testReadMarkdownWithRagMetadata`/`testRagMetadataContentHashChangesWithText` still pass on the old field indices/values).
- `CatalogTable.withMetadata`'s metadata-schema replacement (see `BaseFileSourceConfig.java:80-83`) only fires for the Markdown+enabled path, restoring `catalogTableFromConfig.getMetadataSchema()` (which `parseCatalogTable`/`newCatalogTable` would otherwise silently drop) before merging in the bridge columns — a nice touch, not a regression for other formats since the non-bridge branch (`this.catalogTable = parsedCatalogTable`) is unchanged from before.

However, as detailed in Issue 2 (§1.4), the shared `BaseFileSourceReader`/`MultipleTableFileSourceReader` classes used by **every** file-format connector (CSV/JSON/Parquet/ORC/Text/Excel/XML/Binary/PDF, not just Markdown) now unconditionally route split identifiers through `MarkdownKnowledgeSyncMetadata.safeSourceContext()` when a read fails — regardless of file format or whether `markdown_rag_metadata_enabled` is set. That is a behavior change to existing, non-Markdown connectors' error-message content, and it is not gated by the same flag the rest of the PR consistently uses. This should be fixed before merge to keep the change properly scoped and backward compatible for non-Markdown users.

- SHA-256 digesting of the whole document is streamed via `DigestInputStream` while the bytes are read into the `markdown` string — no second full-file read, negligible added CPU (one digest pass), and it only runs when `markdown_rag_metadata_enabled=true`. Acceptable.
- Per-row chunk hashing (`FileSourceDocumentRouting.sha256Hex(text)`) already existed pre-PR (used to build `content_hash`/`chunk_id`); this PR just reuses the already-computed hash for `ChunkHash` instead of recomputing it — actually a small efficiency win over a naive re-implementation.
- `canonicalizeSourceUri(path)` is called twice per file when `markdown_rag_metadata_enabled=true` (once discarded at construction time for validation, once again inside `read()` for the real logical URI) plus a third time inside `safeSourceContext()` on the error path. All three calls are O(path length) and happen at most once per file/per read, so the real-world cost is negligible — this is a code-clarity nit (see §2.1), not a performance concern.
- No new threads, no new blocking I/O, no retries introduced. `MarkdownKnowledgeSyncMetadata` is a stateless, side-effect-free utility class — no concurrency risk.

### Issue 1: Root-cause exception is silently dropped on file-discovery failure when Markdown RAG metadata is enabled
- **Location**: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/config/BaseFileSourceConfig.java:143-148` and `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/BaseFileSource.java:84-96`
- **Problem description**: Both catch blocks branch on `isMarkdownKnowledgeSyncMetadataEnabled(...)` / `documentRoutingEnabled`. When true, they throw `new FileConnectorException(FileConnectorErrorCode.FILE_LIST_GET_FAILED, errorMsg)` — the **2-argument** constructor, which has no `cause`. The original `ex`/`e` (the real `IOException`/`Exception` describing *why* discovery failed — permission denied, path not found, network timeout, etc.) is discarded entirely; only the sanitized formatted message survives. The sibling branch (flag false) correctly keeps using the 3-argument constructor with `ex`/`e` as cause.
- **Potential risk**: Every production failure of Markdown file discovery (an opt-in but user-facing feature) loses its stack trace and original exception type at the point it's thrown. Operators debugging "why did my Markdown RAG job fail to list files" get a generic message with no root cause — this directly contradicts the project's own stated rule ("Exceptions MUST include sufficient context... Avoid swallowing exceptions"). The duplication in two independent classes also suggests this wasn't a one-off, so it's likely to be copy-pasted into future connectors that adopt the same pattern.
- **Best improvement**: If the concern is that the raw exception's own `getMessage()` might itself contain unmasked path/credential text, that's a narrower problem than dropping the whole cause. Either (a) keep the raw `ex`/`e` as cause always (as the false-branch already does) — the outer message is already sanitized via `safeSourceContext`, so the outer, user-visible message stays safe regardless of what the cause's message contains — or (b) if the cause's message truly must not reach logs, wrap it in a new `Throwable` with a sanitized message but the original stack trace (`copyStackTraceOnly`) rather than discarding it outright.
- **Severity**: High

### Issue 2: Markdown-specific error-sanitization is applied unconditionally to every file connector's read failures, ungated and untested
- **Location**: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/BaseFileSourceReader.java:70-74` and `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/source/reader/MultipleTableFileSourceReader.java:87-95`
- **Problem description**: Everywhere else in this PR (`BaseFileSourceConfig`, `BaseFileSource`), the new Markdown-bridge behavior is carefully gated behind `isMarkdownKnowledgeSyncMetadataEnabled(...)`/`documentRoutingEnabled`. In these two shared reader classes — used by **all** file-format connectors (CSV, JSON, Text, Excel, XML, Binary, ORC, Parquet, PDF, and Markdown alike, across every `connector-file-*` module: LocalFile, S3File, HdfsFile, OssFile, OssJindoFile, FtpFile, SftpFile, CosFile, ObsFile) — the read-failure `catch` block calls `MarkdownKnowledgeSyncMetadata.safeSourceContext(split.splitId())` / `...safeSourceContext(split.getFilePath())` **unconditionally**, with no check on file format or on the Markdown flag at all.
- **Potential risk**: `safeSourceContext()` internally calls `canonicalizeSourceUri()`, which is written specifically for Markdown's URI-shaped identity semantics. For non-Markdown connectors this now silently changes existing error-message text (e.g. masking any embedded credentials for a remote URI — probably fine, but undocumented and untested as a general behavior), and — more importantly — for any `splitId`/file path that happens to look like `scheme:something` without a proper authority (a real possibility for arbitrary filenames/table ids across all these connectors) it will be replaced with the string `"<redacted>"`, silently discarding the actual file path from an already-failing read's error message for every file connector in the project, not just Markdown. There is zero test coverage of this behavior change: neither `BaseFileSourceReader` nor `MultipleTableFileSourceReader` has any dedicated unit test in the repository (before or after this PR).
- **Best improvement**: Gate both call sites the same way the rest of the PR does, e.g. pass the owning `FileFormat`/enabled-flag down to the reader (or check it via the split's associated `BaseFileSourceConfig`) and only route through `MarkdownKnowledgeSyncMetadata.safeSourceContext(...)` for Markdown-with-RAG-enabled splits; otherwise keep the original raw path/splitId in the message exactly as before this PR.
- **Severity**: High

# 2. Code Quality Assessment

## 2.1 Coding Standards

Overall the new code is clean, uses `final`/`private` appropriately, and `MarkdownKnowledgeSyncMetadata` has thorough Javadoc on every public method including a clear class-level summary. A few gaps against the project's "new methods/fields must be commented" standard:

- `MarkdownReadStrategy.addKnowledgeSyncMetadata(...)` (new package-private static method, `MarkdownReadStrategy.java:408-419`) has no Javadoc explaining what it writes into `row.getOptions()` or why (it's the direct producer of the 4 Knowledge Sync keys — non-trivial enough to warrant a one-line doc). **Severity: Medium**
- `MarkdownReadStrategy.LogicalDocumentMetadata` (new private static holder class, `MarkdownReadStrategy.java:118-128`) has no class comment describing its purpose (caching the per-file logical identity so it isn't recomputed per row). **Severity: Medium**
- `BaseFileSourceConfig.isMarkdownKnowledgeSyncMetadataEnabled(...)` and `BaseFileSourceConfig.safeSourceContext(...)` (both new private methods, `BaseFileSourceConfig.java:193-196,226-231`) have no comments; `safeSourceContext`'s fallback-to-`maskUriUserInfo` branching logic in particular would benefit from a one-line "why" comment. **Severity: Medium**
- `BaseFileSource.documentRoutingEnabled` (new field, `BaseFileSource.java:57`) has no field comment explaining what enabling it changes (metadata schema + error-message sanitization + split routing). **Severity: Medium**

None of these are functionally risky, but per the project's documentation standard they should be filled in before merge.

Coverage is strong and deliberate: `MarkdownKnowledgeSyncMetadataTest` exercises the canonicalization edge cases (Windows paths, `file:` URIs, opaque/unsafe URIs, IPv6 hosts, port handling); `MarkdownReadStrategyTest` verifies row-option population, stability across re-reads, content-hash sensitivity to text changes, and non-clobbering of unrelated options; `BaseFileSourceConfigMetadataTest`/`BaseFileSourceTest` cover metadata-schema wiring for single- and multi-table configs plus the "don't leak credentials in the exception message" contract; `MetadataTransformTest` adds alias-projection and physical-name-collision-rejection cases; the E2E `.conf` wires a real `LocalFile -> Metadata -> Assert` pipeline asserting both physical and aliased-logical fields.

Two gaps:
- **Coverage gap (non-blocking)**: every test that exercises both the physical (`source_uri`/`document_id`) and logical (`SourceUri`/`DocumentId`) identities uses a plain local path, where the two values are always equal by construction. Given §1.1's documented "logical identity can differ from physical identity for credential-bearing/query-bearing remote URIs" contract, none of the added tests actually assert a case where they *differ* (e.g. a hierarchical URI with userinfo/query, verifying physical keeps the raw form while logical strips it). A small unit test asserting `source_uri != ks_source_uri` for such an input would directly lock in the documented contract.
- **Missing tests for the changed reader classes (tied to Issue 2)**: `BaseFileSourceReader` and `MultipleTableFileSourceReader` have no unit tests at all (pre- or post-PR), so the unconditional `safeSourceContext()` call described in Issue 2 has no test that would have caught it, e.g. a non-Markdown connector's read failure with a filename shaped like `topic:v1` now producing `"<redacted>"` instead of the real filename.

**Flaky-test-risk rating: Stable.** All new/modified test methods are purely deterministic: fixed byte-array/string inputs, local `@TempDir` files (no shared state, no network, no external containers), SHA-256/string equality assertions, no timing-based waits, no `Thread.sleep`, no ordering assumptions across threads. The one E2E `.conf` change reuses an already-wired, pre-existing `LocalFileIT` test method against a static local resource file with hard-coded expected hash values — same low-flakiness profile as the rest of that IT suite. I see no new source of flakiness introduced by this PR.

## 2.3 Documentation Updates

Excellent, and unusually thorough for a metadata-plumbing PR. `docs/en/transforms/metadata.md` gets an updated Markdown-compatibility note plus a new "Markdown Source Bridge" section: it documents the logical-vs-physical identity split explicitly, gives a worked `Metadata` transform example with the required aliases, and calls out the `ChunkHash`-is-only-valid-for-the-immediate-row caveat and the "no incremental comparison / no tombstones" scope boundary. The identical explanatory block is then propagated consistently to **all nine** touched file-connector docs (CosFile, FtpFile, HdfsFile, LocalFile, ObsFile, OssFile, OssJindoFile, S3File, SftpFile), and mirrored faithfully in `docs/zh` (verified the zh S3File/metadata.md text is a real translation, not a stale copy). Config names, defaults, and hash-formula descriptions match the code exactly. No user-facing incompatible change occurred, so no update to `incompatible-changes.md` is needed for the Markdown-specific parts. (Issue 2, once fixed, needs no doc update either since it's a bug-fix back to prior behavior, not a new documented capability.)

# 3. Architectural Soundness

This is a precise, purpose-built fix, not a workaround: rather than trying to retrofit the new Knowledge Sync semantics onto the existing physical RAG columns (which would have been a breaking rename), the PR introduces a parallel, clearly-scoped "logical identity" concept that rides alongside the physical one via `SeaTunnelRow#options`, exactly matching the pre-existing `KnowledgeSyncMetadataField`/`MetadataTransform` machinery other connectors already use. The collision-avoidance requirement (aliasing via the `Metadata` transform) is enforced in code (`MetadataTransform` throws on a colliding physical name) rather than left as a documentation-only convention. This is a long-term solution, not a stopgap.

## 3.2 Maintainability

`MarkdownKnowledgeSyncMetadata` is a well-isolated, stateless, thoroughly-Javadoc'd utility class that other RAG-capable connectors (the PR already coexists with a `pdf_rag_metadata_enabled` option, suggesting PDF is a likely next adopter) can reuse or mirror. The two error-handling issues above (§1.4) are the main maintainability concerns: the inconsistent gating between `BaseFileSourceConfig`/`BaseFileSource` (correctly gated) and the two reader classes (ungated) is the kind of asymmetry that's easy to miss in review and easy to accidentally rely on later. Fixing Issue 2 to use the same gating pattern everywhere would remove that asymmetry.

## 3.3 Extensibility

The `BRIDGE_FIELDS` array and `withMetadata()` merge-with-compatibility-check pattern in `MarkdownKnowledgeSyncMetadata` make it straightforward to extend to more `KnowledgeSyncMetadataField`s later (e.g. `SourceVersion`, `MimeType`) without another wave of connector-level plumbing changes, since the merge logic already tolerates a pre-existing column of the same name as long as type/nullability match. Good forward design.

No serialized state (checkpoint/savepoint), wire protocol, or SPI contract is touched by this PR — `MarkdownKnowledgeSyncMetadata`/`addKnowledgeSyncMetadata` operate purely on in-flight `SeaTunnelRow` objects and `CatalogTable` schema construction at job-submission time, not on any persisted split/checkpoint state. `FileSourceSplit`/`FileSourceState` classes are untouched. Since `markdown_rag_metadata_enabled` already defaulted to `false` before this PR and no new config option is introduced, jobs upgrading across this PR see no behavior change unless they explicitly opt in — with the caveat that Issue 2, if left unfixed, does change error-message text for *all* file connectors regardless of any flag, which is the one place this PR's "opt-in only" story doesn't hold.

# 4. Issue Summary

|---|---|---|---|
| 1 | Root-cause exception dropped on file-discovery failure when Markdown RAG metadata enabled | `BaseFileSourceConfig.java:143-148`, `BaseFileSource.java:84-96` | High |
| 2 | Markdown-only error-sanitization applied unconditionally to all file connectors' read failures, ungated and untested | `BaseFileSourceReader.java:70-74`, `MultipleTableFileSourceReader.java:87-95` | High |
| 3 | Missing Javadoc on new non-trivial methods/fields | `MarkdownReadStrategy.java:118-128,408-419`, `BaseFileSourceConfig.java:193-196,226-231`, `BaseFileSource.java:57` | Medium |
| 4 | No test asserts physical vs. logical identity actually diverge for a credential/query-bearing remote URI | `MarkdownKnowledgeSyncMetadataTest.java`, `MarkdownReadStrategyTest.java` | Low |
| 5 | Discarded return value from validation-only `canonicalizeSourceUri()` calls (minor clarity/duplicate-work nit) | `BaseFileSourceConfig.java:73-75`, `BaseFileSource.java:79-81` | Low |

# 5. Merge Recommendation

1. Blockers — must be fixed:
   - **Issue 1 (High)**: restore the original exception as `cause` in `BaseFileSourceConfig.discoverFilePaths()` and the `BaseFileSource` constructor's file-discovery `catch` blocks, even when `markdown_rag_metadata_enabled=true`; the outer message is already sanitized, so keeping the cause doesn't reintroduce any credential leak.
   - **Issue 2 (High)**: gate the `MarkdownKnowledgeSyncMetadata.safeSourceContext(...)` calls in `BaseFileSourceReader.pollNext()` and `MultipleTableFileSourceReader.pollNext()` behind the same Markdown-RAG-enabled condition used everywhere else in this PR, so non-Markdown connectors' read-failure messages are unaffected.

   - Add Javadoc to the new methods/fields flagged in §2.1/Issue 3.
   - Add a unit test that exercises a genuinely credential-or-query-bearing remote URI to lock in the documented physical-vs-logical identity divergence (Issue 4).
   - Consider reusing the return value of the validation-only `canonicalizeSourceUri()` calls, or renaming the call sites' intent more explicitly, to avoid the "unused return value" smell (Issue 5).

**Overall assessment**: This is a well-designed, minimally-invasive, opt-in feature with genuinely excellent documentation (English and Chinese, all nine affected connectors) and a thoughtful test suite covering the happy paths and the security-sensitive "don't leak credentials" contract. The two blocking issues are both isolated to error-handling code paths — they don't touch the row data, schema, or checkpoint/restart semantics — so they should be quick, low-risk fixes rather than a sign of a deeper design problem. No alternative implementation approach is warranted; this is the right shape of solution, it just needs the error-handling edges tightened up before merge. Nice work, and thank you for the thorough docs — they made this review much easier to verify against the code.

DanielLeens (approved):
- User pain point: The Markdown RAG source (`markdown_rag_metadata_enabled=true`) only exposed *physical* row columns (`source_uri`, `document_id`, `chunk_id`, `chunk_index`, `content_hash`). The platform's separate, format-agnostic "Knowledge Sync" contract (`KnowledgeSyncMetadataField`: `SourceUri`, `DocumentId`, `DocumentHash`, `ChunkHash`, ...), consumed downstream through row `options` plus a declared `MetadataSchema` and projected by the pre-existing `Metadata` transform, had no Markdown producer. A Markdown pipeline could not participate in Knowledge Sync without hand-rolling the logical identity fields itself.
- Fix approach: add a connector-local bridge (`MarkdownKnowledgeSyncMetadata`) that declares the four Knowledge Sync metadata columns on the Markdown `CatalogTable` and populates a credential-free logical `SourceUri`/`DocumentId`/`DocumentHash`/`ChunkHash` into every row's `SeaTunnelRow#options`, entirely alongside (not replacing) the existing physical RAG columns, gated by the pre-existing `markdown_rag_metadata_enabled` flag.
- One-sentence summary: this PR makes Markdown RAG output a first-class Knowledge Sync metadata producer, purely additive and opt-in.

Sorry for missing this in the last review — that's on me for two of the doc-comment nits below; everything else in this pass is a fresh, complete re-review of the current head, not just the incremental diff.

**Re-review status**: I reviewed this PR at `1810b839` (CHANGES_REQUESTED, two High blockers). The follow-up commit `7c1e91e9` fixes both blockers with real regression tests, not just a message tweak — verified below with file/line evidence. I found no new correctness issues in the delta, and my requested naming/compatibility check against the Knowledge Sync consumer (`MetadataTransform`) passes cleanly. I'm updating my verdict to APPROVE.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Field naming/schema vs. the "Knowledge Sync" consumer (this re-review's specific focus):**
The only real "Knowledge Sync" consumer in this repository is `org.apache.seatunnel.transform.metadata.MetadataTransform` (`seatunnel-transforms-v2/.../MetadataTransform.java:126-137`), which reads `inputRow.getOptions().get(metadataFieldName)` where `metadataFieldName` is a config-supplied logical key, and separately validates that key exists in `CatalogTable.getMetadataSchema()` before wiring an output column. Both sides of that contract are driven by the same unmodified enum, `KnowledgeSyncMetadataField` (`seatunnel-api/.../KnowledgeSyncMetadataField.java`, untouched by this PR — confirmed via `git diff dev...HEAD` on that file, empty). `MarkdownKnowledgeSyncMetadata.addKnowledgeSyncMetadata()` (`MarkdownReadStrategy.java:410-420`) writes exactly `KnowledgeSyncMetadataField.SOURCE_URI.getName()` / `DOCUMENT_ID.getName()` / `DOCUMENT_HASH.getName()` / `CHUNK_HASH.getName()` (i.e. the PascalCase logical keys `"SourceUri"`, `"DocumentId"`, `"DocumentHash"`, `"ChunkHash"`) into `row.getOptions()`, and `MarkdownKnowledgeSyncMetadata.withMetadata()` (`MarkdownKnowledgeSyncMetadata.java:167-190`) declares the matching `MetadataColumn`s on the `CatalogTable` via `field.toMetadataColumn()`, which also uses the same `name` (not the separate, unused `physicalName`). So schema declaration and row-option population use the identical key on both the producer and the generic consumer side — there is no naming drift. This is directly exercised end-to-end by the new `MetadataTransformTest#shouldProjectMarkdownKnowledgeSyncMetadataWithAliases` test (`seatunnel-transforms-v2/.../MetadataTransformTest.java:257-280`), which builds a Markdown-shaped `CatalogTable`, writes the four `KnowledgeSyncMetadataField` keys into row options exactly the way the Markdown source does, runs the real `MetadataTransform`, and asserts the aliased output fields land correctly — and by the LocalFile E2E `.conf`, which round-trips a real file through `LocalFile -> Metadata -> Assert` and checks both the physical and the `ks_`-aliased logical fields against hard-coded hash values.

**Compatibility for existing indexed documents (this re-review's other specific focus):**
- `MARKDOWN_RAG_METADATA_ENABLED` already existed pre-PR with `defaultValue(false)` (`FileBaseSourceOptions.java:111-116`), and this PR's diff against that file is empty — no default change, no rename.
- The bridge (`withMetadata()`, `addKnowledgeSyncMetadata()`) only executes when `fileFormat == FileFormat.MARKDOWN && markdown_rag_metadata_enabled == true` (`BaseFileSource.java:74-76`, `BaseFileSourceConfig.java:196-199`); every other format/flag combination takes the pre-existing, unmodified code path (`this.catalogTable = parsedCatalogTable` / no `addKnowledgeSyncMetadata` call).
- The physical columns existing indexed documents already depend on (`source_uri`, `document_id`, `chunk_id`, `chunk_index`, `content_hash`) are computed by the exact same, untouched formulas: `FileSourceDocumentRouting.normalizeSourceUri`/`buildDocumentId`/`sha256Hex` (`MarkdownReadStrategy.java:153,173,240`). The one refactor here is that the content-hash computation was hoisted one call frame up (from inside `appendRagMetadata` to `generateRows`) so it can be reused for logical `ChunkHash` too, but the formula itself (`sha256Hex(text == null ? "" : text)`) is byte-for-byte identical — confirmed via `git diff` on `MarkdownReadStrategy.java`. The E2E `.conf` locks this in with hard-coded expected hashes for `content_hash` that are unchanged from what a pre-PR Markdown pipeline would have produced.
- For a job that already had `markdown_rag_metadata_enabled=true` before this PR (opting into RAG columns), upgrading to this PR adds four new logical metadata columns to the produced `CatalogTable`'s `MetadataSchema` and four new row-option entries. Existing downstream consumers that don't reference `SourceUri`/`DocumentId`/`DocumentHash`/`ChunkHash` in a `Metadata` transform config are unaffected — the new columns are additive and inert unless explicitly selected. The one theoretical edge case — a user-defined schema that already declares a metadata column literally named `SourceUri`/`DocumentId`/etc. with an incompatible type/nullability — fails fast with a clear `FileConnectorException` (`MarkdownKnowledgeSyncMetadata.java:180-184`) rather than silently overwriting it; this is documented, intentional, and exercised by `BaseFileSourceConfigMetadataTest`.
- **Verdict on the two specific checks I was asked to make: field naming/schema matches the `MetadataTransform` consumer exactly (verified via shared enum + new projection test), and there is no silent change to existing indexed documents' physical fields or default behavior.**

**Verification of the two previously-flagged blockers (now fixed):**

*Old Issue 1 — root-cause exception dropped on discovery failure.* Previously, `BaseFileSourceConfig.discoverFilePaths()` and the `BaseFileSource` constructor's discovery `catch` block used the 2-argument `FileConnectorException` constructor when Markdown RAG metadata was enabled, discarding the real `IOException`/`Exception` entirely. The follow-up commit now calls the 3-argument constructor with `MarkdownKnowledgeSyncMetadata.copyStackTraceOnly(ex)` as cause (`BaseFileSourceConfig.java:143-148`, `BaseFileSource.java:89-96`). `copyStackTraceOnly()` (`MarkdownKnowledgeSyncMetadata.java:141-149`) builds a new `RuntimeException` whose message is just `"Sanitized cause type: <original FQCN>"`, copies the original `getStackTrace()` onto it, and recurses through `getCause()` so the full cause chain's types and stack traces survive — while the (potentially credential-bearing) original exception *messages* are dropped. This is a slightly more conservative choice than the "keep the raw exception as-is" option I suggested last time, but it fully addresses the actual problem (silently losing all diagnostic context) and is arguably safer, since it guarantees no cause-chain message text can leak regardless of what a given `IOException` puts in its message. Verified with real regression tests, not just inspection: `BaseFileSourceConfigMetadataTest#shouldRetainSanitizedDiscoveryCause` and `BaseFileSourceTest#testMarkdownSourceRetainsSanitizedDiscoveryCause` both inject a `FailingFileSystem` that throws `IOException("Access denied for ... token=secret-value#part")`, then assert the resulting exception's cause chain retains the exception type name and a non-empty stack trace at every level, while `assertDoesNotExposeSensitiveSource()` confirms `"secret-value"`, `"token="`, and `"#part"` never appear anywhere in the rendered message chain. This is a real, executable proof, not just a read of the diff.

*Old Issue 2 — Markdown-only error sanitization applied unconditionally to every file connector.* Previously, `BaseFileSourceReader.pollNext()` and `MultipleTableFileSourceReader.pollNext()` called `MarkdownKnowledgeSyncMetadata.safeSourceContext(...)` on every read failure regardless of file format, so a CSV/JSON/Parquet/etc. split whose id happened to look like `scheme:something` could have its real path replaced with `"<redacted>"` in the error message. This is now properly gated:
  - `BaseFileSourceReader` gained a `markdownKnowledgeSyncMetadataEnabled` field, set from a new 3-arg constructor called by `BaseFileSource.createReader()` with `documentRoutingEnabled` (`BaseFileSource.java:177`); the 2-arg constructor is kept as a compatibility path defaulting to `false` (`BaseFileSourceReader.java:47-49`), and the sanitization/`copyStackTraceOnly` branch only runs `if (markdownKnowledgeSyncMetadataEnabled)` (`BaseFileSourceReader.java:79-86`).
  - `MultipleTableFileSourceReader` now builds a `Set<String> markdownKnowledgeSyncMetadataTableIds` at construction time from exactly the table configs where `fileFormat == FileFormat.MARKDOWN && MARKDOWN_RAG_METADATA_ENABLED == true` (`MultipleTableFileSourceReader.java:71-77,192-197`), and looks up `split.getTableId()` against that set per-failure (`MultipleTableFileSourceReader.java:98-108`) — correctly scoped per-table for a reader that can serve multiple tables with different formats/flags in the same job.
  - Both are covered by new, deterministic unit tests that didn't exist before (`BaseFileSourceReaderTest`, `MultipleTableFileSourceReaderTest`, both new files) proving a non-Markdown split's real path/splitId (e.g. `"topic:v1"`) survives unredacted in the error and that `e` is the exact original cause object (`Assertions.assertSame(failure, exception.getCause())`), while a Markdown-with-a-signed-URL split gets the sanitized `"https://example.com/docs/a.md"` with credentials stripped.

**Runtime path diagram (Markdown + `markdown_rag_metadata_enabled=true`):**
```
BaseFileSourceConfig/BaseFileSource ctor
  -> canonicalizeSourceUri(root path)                          [fail-fast validation]
  -> MarkdownKnowledgeSyncMetadata.withMetadata(CatalogTable)   [adds 4 metadata columns]
  -> on discovery failure: FileConnectorException(msg, copyStackTraceOnly(ex))  [Issue 1 fix]

MarkdownReadStrategy.read(path, ...)
  -> canonicalizeSourceUri(path) -> logicalSourceUri
  -> buildDocumentId(logicalSourceUri) -> logicalDocumentId
  -> DigestInputStream(getInputStream(path)) -> markdown text + documentHash
  -> per row: appendRagMetadata(...)      [[existing, unchanged formulas]] physical columns
           + addKnowledgeSyncMetadata(...) [[new]] row.getOptions()[SourceUri/DocumentId/DocumentHash/ChunkHash]

BaseFileSourceReader / MultipleTableFileSourceReader.pollNext()
  -> on read failure, only if markdown+enabled (per-table for multi-table): sanitize context + copyStackTraceOnly(e)  [Issue 2 fix]

downstream Metadata transform -> projects row.getOptions() to aliased output columns using the same KnowledgeSyncMetadataField keys
```

## 1.2 Compatibility Impact

**Fully compatible.** Both previously-open compatibility gaps are closed:
- The two High-severity issues that made my last review "Partially incompatible" (an ungated behavior change to error messages for every file connector) are fixed with proper per-format/per-table gating, verified by new tests.
- No config option, dependency, physical column, or public API is added, renamed, or defaulted differently. `MARKDOWN_RAG_METADATA_ENABLED` retains its pre-existing `false` default.
- Physical RAG columns retain their exact names, order, values, and formulas (verified above).
- Non-Markdown formats and Markdown-with-the-flag-disabled retain their pre-PR behavior in both the schema-building and error-handling paths.
- Field naming for the new logical columns matches the shared, pre-existing `KnowledgeSyncMetadataField` contract exactly, so any other current or future producer/consumer of that contract interoperates correctly with Markdown's new logical fields.

Unchanged from my last review — still applicable and still fine:
- SHA-256 digesting of the whole document is streamed via `DigestInputStream` while bytes are read into the `markdown` string — no second full-file read, and it only runs when the flag is enabled.
- Per-row chunk hashing is computed once (in `generateRows`) and reused for both physical `content_hash` and logical `ChunkHash` — a small efficiency win over recomputing.
- `copyStackTraceOnly()` only runs on the (already slow, exceptional) failure path, so its extra allocation is immaterial.
- No new threads, no new blocking I/O, no retries. `MarkdownKnowledgeSyncMetadata` remains a stateless utility class.

No open formal issues. Both previously-raised High-severity issues are fixed and regression-tested (see 1.1 above); I re-verified this against the real diff and executable test assertions, not just the PR description's claims.

One small non-blocking observation carried over from before (previously Issue 5, Low): `canonicalizeSourceUri()` is still called once purely for its fail-fast side effect at construction time (`BaseFileSourceConfig.java:73-77`, `BaseFileSource.java:79-81`) with the return value discarded, then recomputed inside `read()`/`safeSourceContext()`. This is now clearly commented ("Fail fast without replacing the logical identity of each discovered file.") so the intent reads clearly; still a candidate for a tiny follow-up (compute once, reuse), not something to hold up this PR for.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The follow-up commit closes most of the doc-comment gaps I flagged last time:
- `MarkdownReadStrategy.LogicalDocumentMetadata` now has a class comment (`MarkdownReadStrategy.java:118`: "Per-file logical identity and digest reused by every row emitted from that document.").
- `MarkdownReadStrategy.addKnowledgeSyncMetadata(...)` now has a one-line doc (`MarkdownReadStrategy.java:409`: "Adds the four logical Knowledge Sync fields to row options for Metadata projection.").
- `BaseFileSource.documentRoutingEnabled` now has a field comment (`BaseFileSource.java:57`: "Enables document-affine routing and the Markdown Knowledge Sync metadata bridge.").
- `BaseFileSourceConfig.safeSourceContext(...)` gained a "why" comment about the Knowledge-Sync-vs-legacy masking split (`BaseFileSourceConfig.java:227-228`).

One item from the earlier Medium-severity Issue 3 is still open:
### Issue 1 (carryover, Medium): `BaseFileSourceConfig.isMarkdownKnowledgeSyncMetadataEnabled(ReadonlyConfig)` (`BaseFileSourceConfig.java:196-199`) still has no Javadoc explaining what it gates (schema merge + discovery-failure sanitization). It's a one-liner and low-risk, but per the project's "new methods must be commented" standard it should get a short doc before merge — purely a documentation nit, not a logic concern.
- Severity: Medium
- Raised by another reviewer: No (carryover from my own previous review, only partially addressed)

Coverage grew meaningfully in the follow-up commit and now directly covers what was previously only inferred:
- `BaseFileSourceReaderTest` and `MultipleTableFileSourceReaderTest` are brand new (these classes had zero tests before this PR at all, pre- or post- the first commit) and directly exercise the Issue 2 fix: non-Markdown failures keep the real splitId/path and the exact original cause object; Markdown failures get sanitized context plus a stack-trace-preserving, message-scrubbed cause.
- `BaseFileSourceConfigMetadataTest#shouldRetainSanitizedDiscoveryCause` and `BaseFileSourceTest#testMarkdownSourceRetainsSanitizedDiscoveryCause` directly exercise the Issue 1 fix using a purpose-built `FailingFileSystem` that throws an `IOException` with an embedded fake credential, and assert the credential never surfaces anywhere in the exception chain while the stack trace and type name do.
- `MarkdownKnowledgeSyncMetadataTest` now asserts genuine physical-vs-logical divergence for a credential/query-bearing URI (`FileSourceDocumentRouting.buildDocumentId(first) != MarkdownKnowledgeSyncMetadata.buildDocumentId(firstLogicalUri)`), closing the previous Low-severity coverage gap (old Issue 4).
- `MetadataTransformTest#shouldProjectMarkdownKnowledgeSyncMetadataWithAliases` / `#shouldRejectMarkdownCanonicalPhysicalNameCollision` exercise the exact producer/consumer naming contract this re-review was asked to focus on, end to end through the real `MetadataTransform`.
- The LocalFile E2E `.conf` (`local_file_markdown_rag_metadata_to_assert.conf`) is wired into `LocalFileIT.java:448` and asserts both physical and `ks_`-aliased fields with hard-coded SHA-256 values for a real file — a genuine black-box regression check, not just a unit-level mock.

**Flaky-test-risk rating: Stable.** All new/modified tests (including the follow-up commit's) use fixed byte-array/string inputs, `Mockito` mocks or purpose-built in-memory `FileSystem`/`ReadStrategy` fakes, deterministic SHA-256/string/stack-trace-array equality assertions, no `Thread.sleep`, no timing waits, no shared mutable static state, and no ordering assumptions across tests. The `FailingFileSystem`/`FailingConf` test doubles are self-contained per-test instances. I see no new source of flakiness in the follow-up commit.

## 2.3 Documentation Updates

Unchanged from my last review and still excellent: `docs/en/transforms/metadata.md` and the mirrored `docs/zh` version both got an updated "Markdown RAG compatibility" note plus a new "Markdown Source Bridge" section documenting the logical-vs-physical split, a worked config example with the required aliases, and the `ChunkHash`-is-only-valid-for-the-immediate-row caveat; the identical block is consistently propagated to all nine touched connector docs in both languages. I spot-checked `docs/zh/connectors/source/S3File.md` again on this pass and it remains an accurate, current translation matching the code (e.g. correctly states the four field formulas and the alias requirement). Config names, defaults, and hash-formula descriptions still match the code exactly. No update to `incompatible-changes.md` is needed since no incompatible change exists.

# 3. Architectural Soundness

Unchanged assessment: this is a precise, purpose-built long-term solution — a parallel "logical identity" concept riding alongside the physical one via `SeaTunnelRow#options`/`MetadataSchema`, reusing the exact machinery other Knowledge Sync producers already use, with collision-avoidance enforced in code (not just documentation). Not a workaround.

## 3.2 Maintainability

Improved since the last pass: the asymmetric gating between `BaseFileSourceConfig`/`BaseFileSource` (correctly gated from the start) and the two reader classes (previously ungated) — which I called out as the main maintainability risk last time — is now resolved consistently across all four call sites, removing that asymmetry entirely.

## 3.3 Extensibility

Unchanged: the `BRIDGE_FIELDS` array and `withMetadata()` merge-with-compatibility-check pattern make it straightforward to extend to more `KnowledgeSyncMetadataField`s later without another wave of connector-level plumbing changes.

No serialized state (checkpoint/savepoint), wire protocol, or SPI contract is touched — `FileSourceSplit`/`FileSourceState` are untouched, and `MarkdownKnowledgeSyncMetadata`/`addKnowledgeSyncMetadata` operate purely on in-flight `SeaTunnelRow`/`CatalogTable` construction, not persisted state. The one place my last review flagged as breaking the "opt-in only" story for historical jobs — the ungated error-sanitization behavior change affecting *all* file connectors regardless of flag — is now fixed and gated, so jobs upgrading across this PR see genuinely zero behavior change unless they explicitly opt into `markdown_rag_metadata_enabled=true`.

# 4. Issue Summary

|---|---|---|---|
| 1 | `isMarkdownKnowledgeSyncMetadataEnabled(ReadonlyConfig)` still lacks Javadoc (carryover, partially addressed) | `BaseFileSourceConfig.java:196-199` | Medium |

(Previously-open Issues 1 and 2, both High, are fixed and regression-tested; previously-open Issue 4, Low, is fixed with a new divergence test; previously-open Issue 5, Low, remains an optional clarity nit, not re-numbered here since nothing new changed about it.)

# 5. Merge Recommendation

1. **Blockers — must be fixed**: None.
   - Add a short Javadoc to `BaseFileSourceConfig.isMarkdownKnowledgeSyncMetadataEnabled(ReadonlyConfig)` (Issue 1 above) — one line describing what it gates is enough.
   - Consider eliminating the discarded-return-value `canonicalizeSourceUri()` fail-fast calls at construction time in favor of caching the computed value for reuse in `read()` (old Issue 5) — pure clarity, no behavior change.

**Overall assessment**: This is a well-designed, minimally-invasive, opt-in feature, and the two High-severity blockers from my previous review are now properly fixed rather than papered over — I verified both fixes against real, deterministic regression tests that exercise the exact failure scenarios (credential leakage via message, non-Markdown error-message corruption) rather than taking the PR description's word for it. The specific concern I was asked to re-check this round — whether the new logical metadata field names/types actually match what the `Metadata` transform (the real "Knowledge Sync" consumer in this repo) expects, and whether this silently changes anything for already-indexed documents — checks out cleanly: both sides of the contract share the same unmodified `KnowledgeSyncMetadataField` enum, the new fields are purely additive behind an unchanged, still-`false`-by-default flag, and the pre-existing physical columns' names/values/formulas are byte-for-byte unchanged, all backed by new tests that actually assert this rather than just describe it. Thank you for the thorough follow-up — the new regression tests for the error-handling fixes in particular are exactly the kind of evidence that makes a re-review fast and confident. Nice work.

DanielLeens:
Thanks for the update. I checked the current head against the latest `dev` and this branch is currently behind upstream (`behind_by=39`, compare status: `diverged`). The current failing CI signals — `jdbc-connectors-it-part-6 (8, ubuntu-latest)`, `all-connectors-it-7 (8, ubuntu-latest)`, and `unit-test (11, windows-latest)` — do not appear to be directly caused by this PR's diff; `jdbc-connectors-it-part-6` in particular has also failed on `dev`'s own recent CI history, so this looks like pre-existing upstream flakiness rather than something introduced here. The lowest-cost next step would be to sync with the latest `dev` and rerun CI. If it still fails after the sync, happy to help narrow it down against the updated head.

DanielLeens (approved):
Thanks for syncing with the latest `dev` — that resolves the upstream-drift note I left on 2026-08-14 (`behind_by=39`, diverged). I re-reviewed the current head from scratch as required for a new commit, so this is a complete re-review, not just a delta check.

- User pain point: the Markdown RAG source (`markdown_rag_metadata_enabled=true`) only exposed *physical* row columns (`source_uri`, `document_id`, `chunk_id`, `chunk_index`, `content_hash`). The platform's separate, format-agnostic "Knowledge Sync" contract (`KnowledgeSyncMetadataField`: `SourceUri`, `DocumentId`, `DocumentHash`, `ChunkHash`, ...), consumed downstream through row `options` plus a declared `MetadataSchema` and projected by the pre-existing `Metadata` transform, had no Markdown producer.
- Fix approach: add a connector-local bridge (`MarkdownKnowledgeSyncMetadata`) that declares the four Knowledge Sync metadata columns on the Markdown `CatalogTable` and populates a credential-free logical `SourceUri`/`DocumentId`/`DocumentHash`/`ChunkHash` into every row's `SeaTunnelRow#options`, entirely alongside (not replacing) the existing physical RAG columns, gated by the pre-existing `markdown_rag_metadata_enabled` flag.
- One-sentence summary: this PR makes Markdown RAG output a first-class Knowledge Sync metadata producer, purely additive and opt-in.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**What's new since my last review, verified with `git diff`:** the only new commit on top of `7c1e91e9` (the commit I approved on 2026-08-11) is a merge commit (`05c8a7676dc9`, "Merge branch 'apache:dev' into agent/align-markdown-knowledge-sync-metadata"). I diffed the two commits directly:

```
git diff 7c1e91e94b58..HEAD -- <every code/test/conf file this PR touches>
```

This returns **zero lines of diff** across all 14 non-doc files (`BaseFileSourceConfig.java`, `BaseFileSource.java`, `BaseFileSourceReader.java`, `MarkdownKnowledgeSyncMetadata.java`, `MarkdownReadStrategy.java`, `MultipleTableFileSourceReader.java`, all six new/extended test classes, the LocalFile E2E `.conf`, and `MetadataTransformTest.java`). The two High-severity fixes I verified last round — restoring the exception cause via `copyStackTraceOnly()` in `BaseFileSourceConfig.java:143-148`/`BaseFileSource.java:89-96`, and gating `safeSourceContext()` sanitization behind `markdownKnowledgeSyncMetadataEnabled` in `BaseFileSourceReader.java:79-86` and per-table in `MultipleTableFileSourceReader.java` — are still present byte-for-byte, which I re-confirmed by reading the current head directly rather than trusting the old diff.

The 20 doc `.md` files under `docs/en`/`docs/zh` do show non-empty diffs between `7c1e91e9` and the current head, but every one of them is unrelated content from other PRs that landed on `dev` in the interim (e.g. `excel_engine`/`poi_excel_max_file_size` docs, XML XXE-hardening notes) cleanly interleaved into the same connector doc pages by the merge — not touching this PR's own added sections. I specifically diffed `docs/en/transforms/metadata.md` and `docs/zh/transforms/metadata.md` between the two commits: **zero diff** on both, confirming the "Markdown Source Bridge" section is untouched. I also grepped the full PR-owned file list for leftover `<<<<<<<`/`=======`/`>>>>>>>` conflict markers: none found.

**Runtime path (unchanged from my last review):**
```
BaseFileSourceConfig/BaseFileSource ctor
  -> canonicalizeSourceUri(root path)                          [fail-fast validation]
  -> MarkdownKnowledgeSyncMetadata.withMetadata(CatalogTable)   [adds 4 metadata columns]
  -> on discovery failure: FileConnectorException(msg, copyStackTraceOnly(ex))

MarkdownReadStrategy.read(path, ...)
  -> canonicalizeSourceUri(path) -> logicalSourceUri
  -> buildDocumentId(logicalSourceUri) -> logicalDocumentId
  -> DigestInputStream(getInputStream(path)) -> markdown text + documentHash
  -> per row: appendRagMetadata(...)      [[existing, unchanged formulas]] physical columns
           + addKnowledgeSyncMetadata(...) [[new]] row.getOptions()[SourceUri/DocumentId/DocumentHash/ChunkHash]

BaseFileSourceReader / MultipleTableFileSourceReader.pollNext()
  -> on read failure, only if markdown+enabled (per-table for multi-table): sanitize context + copyStackTraceOnly(e)

downstream Metadata transform -> projects row.getOptions() to aliased output columns
```

**Key findings:**
- The normal path (any file-format connector, any job) is unaffected: the bridge only activates for `fileFormat == MARKDOWN && markdown_rag_metadata_enabled == true`.
- The scenario this PR covers is exactly the documented one: a Markdown RAG pipeline that wants to participate in the shared Knowledge Sync metadata contract without a source-specific consumer.
- This is a precise, long-term fix (not a workaround): it reuses the existing `KnowledgeSyncMetadataField`/`MetadataTransform` machinery rather than inventing a parallel mechanism.
- No blind spots reopened by this round's merge — the merge is a pure sync, verified with concrete `git diff` evidence, not just trust in the PR description.

## 1.2 Compatibility Impact

**Fully compatible**, unchanged from my last review:
- `MARKDOWN_RAG_METADATA_ENABLED` still defaults to `false`; no config option, dependency, physical column, or public API is added, renamed, or re-defaulted.
- Physical RAG columns (`source_uri`, `document_id`, `chunk_id`, `chunk_index`, `content_hash`) retain their exact names, order, values, and formulas — confirmed again this round by the zero-diff check above.
- Non-Markdown formats and Markdown-with-the-flag-disabled retain pre-PR behavior in both schema-building and error-handling paths.
- No serialized state (checkpoint/savepoint), wire protocol, or SPI contract is touched.

Unchanged from my last review: SHA-256 digesting is streamed via `DigestInputStream` with no second full-file read and only runs when the flag is enabled; per-row chunk hashing is computed once and reused for both the physical and logical hash; `copyStackTraceOnly()` only runs on the already-slow failure path; no new threads, blocking I/O, or retries; `MarkdownKnowledgeSyncMetadata` remains stateless.

No open formal issues. Both previously-raised High-severity issues (root-cause exception dropped on discovery failure; Markdown-only sanitization applied unconditionally to all file connectors) remain fixed and regression-tested, re-verified directly against the current head's source in this round (see 1.1).

**No blocking issues found** in this re-review. The one previously-open Medium/Low nits from my prior round (missing Javadoc on `isMarkdownKnowledgeSyncMetadataEnabled`, discarded return value from the fail-fast `canonicalizeSourceUri()` call) are pre-existing, non-blocking, and unchanged by this merge — I'm not re-numbering them as new issues since nothing changed about them in this commit.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Unchanged and already addressed in the prior round: the follow-up commit closed the Javadoc gaps on `LogicalDocumentMetadata`, `addKnowledgeSyncMetadata`, `documentRoutingEnabled`, and `safeSourceContext`. `isMarkdownKnowledgeSyncMetadataEnabled(ReadonlyConfig)` (`BaseFileSourceConfig.java`) still lacks a one-line Javadoc — a small, non-blocking carryover from before, not reintroduced or worsened by this merge.

Test files are byte-identical to the version I already verified as regression-covering both fixed issues (`BaseFileSourceConfigMetadataTest#shouldRetainSanitizedDiscoveryCause`, `BaseFileSourceTest#testMarkdownSourceRetainsSanitizedDiscoveryCause`, `BaseFileSourceReaderTest`, `MultipleTableFileSourceReaderTest`, the physical-vs-logical-identity-divergence test in `MarkdownKnowledgeSyncMetadataTest`, and the alias-projection/collision-rejection tests in `MetadataTransformTest`). The LocalFile E2E `.conf` (`local_file_markdown_rag_metadata_to_assert.conf`, wired into `LocalFileIT.java`) is unchanged and still asserts both physical and `ks_`-aliased logical fields against hard-coded SHA-256 values for a real parsed file — a genuine black-box check of the actual metadata shape, not merely "parsing doesn't throw."

**Flaky-test-risk rating: Stable.** All test files are unchanged from the version I already rated Stable: fixed byte-array/string inputs, `@TempDir`/in-memory fakes, deterministic hash/string/stack-trace-array equality assertions, no `Thread.sleep`, no timing waits, no shared mutable static state, no ordering assumptions. Nothing about this merge touches test code or introduces new flakiness surface.

## 2.3 Documentation Updates

Both `docs/en` and `docs/zh` remain updated and consistent. `docs/en/transforms/metadata.md`/`docs/zh/transforms/metadata.md` are byte-identical to the version I verified last round (zero diff against `7c1e91e9`), still documenting the logical-vs-physical identity split, the worked alias example, and the `ChunkHash`-immediate-row caveat. The nine touched connector docs picked up unrelated interleaved content from `dev` (Excel engine, XML XXE hardening) during the sync merge, but their own Markdown-bridge sections are untouched and still accurate against the code.

# 3. Architectural Soundness

Unchanged assessment: a precise, long-term solution — a parallel "logical identity" riding alongside the physical one via `SeaTunnelRow#options`/`MetadataSchema`, with collision-avoidance enforced in code, not just docs.

## 3.2 Maintainability
Unchanged: the previously-flagged asymmetric gating between the config/source classes and the two reader classes was fixed in the `7c1e91e9` follow-up and remains consistent in the current head.

## 3.3 Extensibility
Unchanged: the `BRIDGE_FIELDS` array/`withMetadata()` merge pattern still makes it straightforward to add more `KnowledgeSyncMetadataField`s later without another wave of connector-level plumbing.

Unchanged: no checkpoint/savepoint/wire-protocol/SPI contract is touched; `FileSourceSplit`/`FileSourceState` are untouched; jobs upgrading across this PR see zero behavior change unless they explicitly opt into `markdown_rag_metadata_enabled=true`.

# 4. Issue Summary

No new or open blocking issues this round. (Non-blocking carryovers, unchanged by this merge: missing one-line Javadoc on `isMarkdownKnowledgeSyncMetadataEnabled`; discarded return value from the fail-fast `canonicalizeSourceUri()` call — both Low/Medium, non-blocking, already noted in my prior review.)

# 5. Merge Recommendation

1. **Blockers — must be fixed**: None. Both previously-open High-severity issues remain fixed and regression-tested; this merge introduced zero code/test changes to this PR's own files (verified with a direct `git diff` between the previously-approved commit and the current head).
2. **Recommended fixes — non-blocking**: the one-line Javadoc gap on `isMarkdownKnowledgeSyncMetadataEnabled` and the discarded fail-fast return value are still worth a tiny follow-up, but neither blocks this PR.

**CI status at this head (`05c8a7676dc9`)**: the fork's Actions run (https://github.com/yzeng1618/seatunnel-dev/actions/runs/32234519363) is still in progress at the time of this review — 24 jobs completed successfully, 6 jobs (including a couple of IT shards) still running, 0 failures so far. The `apache/seatunnel` "Build" check status is just a pointer to that fork run and will update once it completes; I don't see any evidence of a real failure at this point. Since the merge introduced no code changes to this PR's files, I don't expect this run to surface anything new, but I'd still recommend letting it finish before merging.

**Overall assessment**: this re-review confirms the sync-with-`dev` I asked for on 2026-08-14 is now done and clean — no conflicts, no accidental content changes to this PR's own logic, tests, or docs. The PR is functionally identical to the version I fully verified and approved on 2026-08-11 (real regression tests for both previously-fixed High-severity issues, a genuine end-to-end E2E assertion of the new metadata shape, and thorough bilingual docs across all nine touched connectors). No alternative implementation is warranted. Once the currently-running CI finishes green, this is ready for a maintainer with write access to merge — my own review here is read-only, so the final approve/merge action still needs to come from someone with commit rights.

DanielLeens (approved):
Thanks for the follow-up commit. Since my last APPROVED review (`05c8a7676dc9`, 2026-08-19), there's exactly one new commit with real code changes — `4d0315e6c` ("[Improve][Connector-V2] Refine markdown source URI validation") — so this is a genuine full re-review of the current head, not a rubber-stamp.

(Unchanged from my prior reviews — restated briefly for context.) The Markdown RAG source only exposed physical row columns; this PR adds a connector-local bridge (`MarkdownKnowledgeSyncMetadata`) that also populates the platform's format-agnostic Knowledge Sync metadata columns (`SourceUri`/`DocumentId`/`DocumentHash`/`ChunkHash`) into every row's `options`, purely additive and gated by the existing `markdown_rag_metadata_enabled` flag.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**What's new in this commit, verified with `git diff 05c8a7676dc9..HEAD`:** exactly 2 files, 23 insertions / 26 deletions, both in `connector-file-base`:
```
BaseFileSourceConfig.java | 31 ++++++++++------------
BaseFileSource.java       | 18 ++++++-------
```

**The refactor:** in both classes, the pattern used to be "call `MarkdownKnowledgeSyncMetadata.canonicalizeSourceUri(path)` once purely for its fail-fast side effect (discarding the result), then separately call `safeSourceContext(path)` again later at log/error time to get a display string." This commit collapses that into a single field/local (`safeDiscoveryRootContext`) computed once and reused for both the success-path log line and the failure-path error message:

```java
// BaseFileSourceConfig — before
if (isMarkdownKnowledgeSyncMetadataEnabled(readonlyConfig)) {
    MarkdownKnowledgeSyncMetadata.canonicalizeSourceUri(rootPath);   // fail-fast, result discarded
}
...
log.info(..., safeSourceContext(rootPath), ...);                     // recomputed later
// safeSourceContext() wraps canonicalizeSourceUri() in try/catch -> "<redacted>" on failure

// BaseFileSourceConfig — after
this.safeDiscoveryRootContext =
        isMarkdownKnowledgeSyncMetadataEnabled(readonlyConfig)
                ? MarkdownKnowledgeSyncMetadata.canonicalizeSourceUri(rootPath)   // fail-fast, result KEPT
                : maskUriUserInfo(rootPath);
...
log.info(..., safeDiscoveryRootContext, ...);                         // reused, not recomputed
```
`BaseFileSource.java` gets the identical treatment (local variable instead of a field, same reasoning).

**Correctness verification — is dropping the separate `safeSourceContext()` try/catch actually safe?** I read `MarkdownKnowledgeSyncMetadata.safeSourceContext()` (`MarkdownKnowledgeSyncMetadata.java:128-134`): it is literally `try { return canonicalizeSourceUri(sourceUri); } catch (RuntimeException e) { return "<redacted>"; }` — i.e. it's the *same* pure, deterministic function, just with a redaction fallback on failure. Since `canonicalizeSourceUri` is a pure function of its string input (no mutable/external state), and the fail-fast call in the constructor already succeeded before either `discoverFilePaths()`'s log/catch block or `BaseFileSource`'s catch block can run (the fail-fast call is *unconditionally* before the `try` in both classes, in both the old and new code), the later `safeSourceContext()`/`canonicalizeSourceUri()` call must return the exact same result as the constructor's — its `<redacted>` catch branch could never actually fire on this call path. Caching the constructor's already-successful return value instead of recomputing it is behaviorally identical, not just "probably fine."

**Verification that `rootPath`/`path` can't drift from the cached value:** `discoverFilePaths(ReadStrategy)` re-reads `rootPath = baseFileSourceConfig.get(FileBaseSourceOptions.FILE_PATH)` — the same `ReadonlyConfig` key read in the constructor. `ReadonlyConfig` is immutable for the lifetime of the object, so the cached `safeDiscoveryRootContext` can never go stale relative to what a fresh read would produce, even though `discoverFilePaths` can be invoked more than once (e.g. via `getFilePathsForSplitEnumerator()` in the deferred-discovery case).

**The two High-severity fixes from my earlier rounds are untouched and still verified present:** `copyStackTraceOnly()` is still used to build the sanitized exception cause in both classes' catch blocks (`BaseFileSourceConfig.java` discoverFilePaths catch, `BaseFileSource.java` constructor catch), and the markdown-only branching for that sanitization is preserved byte-for-byte in intent (`isMarkdownKnowledgeSyncMetadataEnabled(...)` / `documentRoutingEnabled` still gates it).

**Net effect:** this is a real, positive simplification — it removes a small amount of redundant computation (canonicalizing the same URI twice) and a private helper method (`BaseFileSourceConfig.safeSourceContext()`) that's no longer needed, while preserving identical externally-observable behavior (same log lines, same error messages, same fail-fast timing, same credential-safety guarantees). I don't see a correctness or compatibility regression in this commit.

## 1.2 Compatibility Impact
**Fully compatible.** Purely internal refactor of a private/internal display-string computation; no public API, config option, or on-wire/serialized contract changes. `safeDiscoveryRootContext` is explicitly `@Getter(AccessLevel.NONE)` in `BaseFileSourceConfig`, so it isn't even exposed as a new getter.

Net positive — one fewer redundant URI-canonicalization pass per discovery cycle (small, but free).

No issues. Error-message content and exception-cause sanitization behavior are unchanged from what I verified in my prior rounds.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Good — the new field has a clear one-line doc comment explaining its purpose (`/** Credential-safe root path rendered in file-discovery logs and exceptions. */`), and `isMarkdownKnowledgeSyncMetadataEnabled()` picked up a doc comment it didn't have before. No wildcard imports, license headers untouched (no new files).

No test files were touched by this commit. I checked whether that's a gap: `MarkdownKnowledgeSyncMetadataTest.java` already directly unit-tests `canonicalizeSourceUri()` across the relevant input shapes (local paths, `file://` URIs, unsafe/unparseable values throwing), and since this commit is a behavior-preserving refactor around calls to that already-tested function (not new logic), I don't think a new test is required here. Rating: **Stable** — no flaky-test risk, no coverage regression.

## 2.3 Documentation Updates
No user-facing behavior changed, so no `docs/en`/`docs/zh` update is needed for this specific commit (docs were already reviewed as complete/unchanged in my prior round).

# 3. Architectural Soundness

Precise fix — a small, well-justified cleanup that removes duplicate computation without changing behavior. This is exactly the kind of tightening-up commit that's good to see land on top of an already-reviewed PR rather than being left as tech debt.

## 3.2 Maintainability
Slightly improved — one call site for the "compute the safe context string" logic per class instead of two, one fewer private helper method to keep in sync.

## 3.3 Extensibility
Unaffected.

No change — still purely additive/opt-in behind `markdown_rag_metadata_enabled`, no serialization/checkpoint/state format involved.

# 4. Issue Summary

No new issues found in this commit.

# 5. Merge Recommendation

1. Blockers — must be fixed
   - None.
   - None.

Overall assessment: this commit is a clean, verified-safe simplification on top of an already-approved PR — I traced the "cache vs. recompute" question end-to-end (immutability of the config, purity of `canonicalizeSourceUri`, and the fail-fast-before-try ordering that makes the redaction fallback unreachable on this path) rather than assuming it's fine because the diff looks small. My approval stands.

dybyte (approved):
+1

PR #11660, opened 2026-08-05
[Improve][Connector-V2][File] Build the Parquet Avro data model once per file instead of once per row

loustler:
`ParquetWriteStrategy#getOrCreateOutputStream` is called from `write()`, i.e. once per row, and it constructs a `GenericData` and registers its logical-type conversions on **every** call — but the object it builds is only consumed on the branch that creates a new `ParquetWriter`. For every row that writes into an already-open file it is built and immediately discarded.

> The snippets and measurements below show three registered conversions, which is what this method looked like when the PR was opened. #11744 has since removed `LocalTimestampMillisConversion` on `dev`, so the merged version registers two; that change is orthogonal to this one — it decides *what* is registered, this decides *when* the model is built — and it is preserved as-is by the merge. See the provenance note in the Evidence section.

I found this while profiling a production ingestion run, so there are measurements attached. I want to be upfront about what they do and do not show: **this is wasted CPU, wasted allocation and a contended JVM-global lock on the write path, not a demonstrated end-to-end throughput win.** In my own run the writers were parked 56.7% of the time waiting for rows, so the job was source-bound and removing this cost would not have shortened it much. I've tried to separate measured from inferred throughout, and I'd welcome correction where I've misread something.

### The code

`seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/ParquetWriteStrategy.java`

```java
public void write(@NonNull SeaTunnelRow seaTunnelRow) {          // once per row
    ...
    ParquetWriter<GenericRecord> writer = getOrCreateOutputStream(filePath);
```

```java
public ParquetWriter<GenericRecord> getOrCreateOutputStream(@NonNull String filePath) {
    ...
    ParquetWriter<GenericRecord> writer = this.beingWrittenWriter.get(filePath);
    GenericData dataModel = new GenericData();                                     // every row
    dataModel.addLogicalTypeConversion(new Conversions.DecimalConversion());       // every row
    dataModel.addLogicalTypeConversion(new TimeConversions.DateConversion());      // every row
    dataModel.addLogicalTypeConversion(new TimeConversions.LocalTimestampMillisConversion());
    if (writer == null) {                                                          // only use site
        ...
        AvroParquetWriter.<GenericRecord>builder(outputFile).withDataModel(dataModel)...
    }
    return writer;                                                                 // dataModel dropped
}
```

[...]

PR #11744, opened 2026-08-10
fix: remove unreachable LocalTimestampMillisConversion from ParquetWriteStrategy data model [#11743]

### LINKED ISSUE #11743: [Bug] [Connector-V2][File] Parquet sink registers LocalTimestampMillisConversion for a logical type it may never produce
### Search before asking

### What happened

`ParquetWriteStrategy#getOrCreateOutputStream` registers three Avro logical-type conversions on the data model it hands to `AvroParquetWriter`. The third one looks unreachable, and this issue is to settle whether it is.

```java
dataModel.addLogicalTypeConversion(new Conversions.DecimalConversion());
dataModel.addLogicalTypeConversion(new TimeConversions.DateConversion());
dataModel.addLogicalTypeConversion(new TimeConversions.LocalTimestampMillisConversion());
```

`TimeConversions.LocalTimestampMillisConversion` handles the Avro logical type **`local-timestamp-millis`**. But a SeaTunnel `TIMESTAMP` column is written as `INT64 .as(OriginalType.TIMESTAMP_MILLIS)` in `ParquetWriteStrategy#seaTunnelDataType2ParquetDataType`, and parquet-avro's `AvroSchemaConverter` maps that to **`timestamp-millis`**. Those are different logical types, so nothing on this path should ever ask the model for the `local-timestamp-millis` conversion.

Reinforcing that: `resolveObject` converts `LocalDateTime` by hand rather than leaving it to the data model, so even the value path does not route through the conversion.

**This is not proven, which is why it is an issue rather than a PR.** I have not shown that no code path produces a `local-timestamp-millis` schema — a different write path, a schema supplied rather than derived, or a future type mapping could. Removing a registration that turns out to be live would silently corrupt timestamps, so the removal is only safe once reachability is actually settled.

### What to check

1. Whether any path through `seaTunnelDataType2ParquetDataType` / `buildAvroSchemaWithRowType` can yield `local-timestamp-millis`.
2. Whether `resolveObject`'s manual `LocalDateTime` handling makes the registration dead even when the schema does carry that logical type.
3. If both come back negative, remove the registration — and if not, the schema/conversion mismatch is worth a comment so the next reader does not repeat this analysis.

### Context

Raised during review of #11660, which relocated these three registrations from per-row to per-file but deliberately did not change what is registered. Analysis originally posted in [this comment](https://github.com/apache/seatunnel/pull/11660#issuecomment-5191004346); @SEZ9 asked for it to be captured separately so it is not lost in that thread.

A logical-type round-trip test is what would surface this. #11660 adds one (`testLogicalTypesRoundTripAcrossReusedWriter`), and it passes — which is consistent with the conversion being unnecessary, but does not by itself prove it, since a passing round trip does not distinguish "the conversion was not needed" from "the conversion was used and worked."

### SeaTunnel Version

dev

### SeaTunnel Config

```conf
N/A - code-level question, not a runtime failure.
```

### Running Command

```shell
N/A
```

### Error Exception

```log
N/A - no exception; this is a reachability question about a registered converter.
```

Zeta

### Java or Scala Version

_No response_

### Screenshots

_No response_

Willing to do the reachability investigation and submit the removal if it comes back dead, but I would rather someone who knows the Avro schema-derivation path confirm the analysis first.

### Code of Conduct

waterWang:
## Description

`ParquetWriteStrategy#getOrCreateOutputStream` registers three Avro logical-type conversions on the data model. The third one (`LocalTimestampMillisConversion`) is unreachable.

### Analysis

1. `seaTunnelDataType2ParquetDataType` maps TIMESTAMP to `INT64.as(OriginalType.TIMESTAMP_MILLIS)`, which parquet-avro's `AvroSchemaConverter` converts to Avro's `timestamp-millis` logical type — NOT `local-timestamp-millis`.

2. `resolveObject` handles TIMESTAMP values manually (converting to epoch millis via `LocalDateTime.toEpochMilli()`), bypassing the data model's conversion entirely.

Therefore, `TimeConversions.LocalTimestampMillisConversion` is never invoked on any code path through this class.

### What changed

- Removed the `LocalTimestampMillisConversion` registration
- Added a comment explaining why it's not needed

### Related issue

Closes #11743

### DIFF
--- a/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/ParquetWriteStrategy.java
+++ b/seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/ParquetWriteStrategy.java
@@ -174,7 +174,13 @@ public class ParquetWriteStrategy extends AbstractWriteStrategy<ParquetWriter<Ge
         GenericData dataModel = new GenericData();
         dataModel.addLogicalTypeConversion(new Conversions.DecimalConversion());
         dataModel.addLogicalTypeConversion(new TimeConversions.DateConversion());
-        dataModel.addLogicalTypeConversion(new TimeConversions.LocalTimestampMillisConversion());
+        // LocalTimestampMillisConversion is not needed here because
+        // seaTunnelDataType2ParquetDataType maps TIMESTAMP to
+        // INT64.as(OriginalType.TIMESTAMP_MILLIS), which AvroSchemaConverter
+        // converts to Avro's timestamp-millis (not local-timestamp-millis).
+        // resolveObject also handles TIMESTAMP values manually (converting to
+        // epoch millis), bypassing the data model entirely.
+        // See https://github.com/apache/seatunnel/issues/11743
         if (writer == null) {
             Path path = new Path(filePath);
             // initialize the kerberos login


davidzollo (commented):
Posted in error by an automation misconfiguration (wrong GitHub identity) - please disregard this comment. The intended full review has been reposted correctly below.

DanielLeens (commented):
- User pain point: `ParquetWriteStrategy#getOrCreateOutputStream` registers a third Avro logical-type conversion (`TimeConversions.LocalTimestampMillisConversion`) on the `GenericData` model that is never actually exercised, which is confusing dead code and a maintenance trap for future readers (as flagged in #11743).
- Fix approach: remove the dead registration and replace it with a comment explaining precisely why it is unreachable, linking back to #11743.
- One-sentence summary: this is a pure dead-code cleanup with no observable behavior change — nothing about the schema, the written bytes, or the read-back values changes.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**File:** `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/sink/writer/ParquetWriteStrategy.java:174-182`

Before:
```java
GenericData dataModel = new GenericData();
dataModel.addLogicalTypeConversion(new Conversions.DecimalConversion());
dataModel.addLogicalTypeConversion(new TimeConversions.DateConversion());
dataModel.addLogicalTypeConversion(new TimeConversions.LocalTimestampMillisConversion());
```

After:
```java
GenericData dataModel = new GenericData();
dataModel.addLogicalTypeConversion(new Conversions.DecimalConversion());
dataModel.addLogicalTypeConversion(new TimeConversions.DateConversion());
// LocalTimestampMillisConversion is not needed here because
// seaTunnelDataType2ParquetDataType maps TIMESTAMP to
// INT64.as(OriginalType.TIMESTAMP_MILLIS), which AvroSchemaConverter
// converts to Avro's timestamp-millis (not local-timestamp-millis).
// resolveObject also handles TIMESTAMP values manually (converting to
// epoch millis), bypassing the data model entirely.
// See https://github.com/apache/seatunnel/issues/11743
```

I independently re-derived both reachability claims from #11743 against the current `dev` head rather than taking them on faith, since a wrong removal here would silently corrupt every TIMESTAMP column written by the Parquet sink:

**Claim 1 — the schema this class produces never carries `local-timestamp-millis`.**
`seaTunnelDataType2ParquetDataType` (same file, TIMESTAMP branch, non-INT96 case) builds the Parquet type as:
```java
Types.primitive(PrimitiveType.PrimitiveTypeName.INT64, Type.Repetition.OPTIONAL)
        .as(OriginalType.TIMESTAMP_MILLIS)
        .named(fieldName);
```
`buildAvroSchemaWithRowType` then feeds the resulting `MessageType` through `schemaConverter.convert(...)`, i.e. `parquet-avro`'s `AvroSchemaConverter` (this module pins `parquet-avro.version=1.12.3`). I pulled the actual 1.12.3 `AvroSchemaConverter` source: any Parquet `TIMESTAMP_MILLIS`/`TIMESTAMP_MICROS` annotation — old-style `OriginalType` is normalized into the same `LogicalTypeAnnotation.TimestampLogicalTypeAnnotation` by the Parquet type builder — is converted via:
```java
public Optional<LogicalType> visit(TimestampLogicalTypeAnnotation timestampLogicalType) {
  switch (timestampLogicalType.getUnit()) {
    case MILLIS: return of(LogicalTypes.timestampMillis());
    case MICROS: return of(LogicalTypes.timestampMicros());
  }
  ...
}
```
Note this visitor does **not** branch on `isAdjustedToUTC` at all — in parquet-avro 1.12.3, `AvroSchemaConverter` has no code path that ever emits Avro's `local-timestamp-millis` logical type, for any input. So the schema field for a SeaTunnel `TIMESTAMP` column is always `timestamp-millis`, never `local-timestamp-millis`. `LocalTimestampMillisConversion` is keyed to the `local-timestamp-millis` logical type, so it could never be selected by Avro's `GenericData.getConversion(...)` lookup for this schema, full stop.

**Claim 2 — even if the schema did carry that logical type, the runtime value would still bypass it.**
`resolveObject`'s `TIMESTAMP` case (non-INT96 branch, same file, ~line 278) does:
```java
return ((LocalDateTime) data)
        .atZone(ZoneId.systemDefault())
        .toInstant()
        .toEpochMilli();
```
i.e. it hand-converts the `LocalDateTime` to a boxed `Long` *before* handing it to `GenericRecordBuilder.set(...)`. Avro's `GenericData` conversion mechanism is selected by matching the **runtime class of the datum** against a conversion's `getConvertedType()` (typically `LocalDateTime`/`Instant`), not by schema logical type alone. Since the datum is already a `Long` by the time it reaches Avro, no conversion registered on `dataModel` — including the now-removed one, and including the "correct" `TimestampMillisConversion` that was never registered either — is ever consulted for TIMESTAMP fields. This applies recursively to nested `ROW` fields too, since the `ROW` case in `resolveObject` (~line 288) recurses through the same switch, so nested TIMESTAMP columns hit the identical manual-Long-conversion path.

**Key Findings**
- Two independent reasons converge on the same conclusion: (a) the schema this class emits can never carry `local-timestamp-millis` in parquet-avro 1.12.3, and (b) `resolveObject` never hands Avro a `LocalDateTime`/`Instant` for a TIMESTAMP field regardless of schema, so the conversion lookup mechanism itself is never engaged.
- Normal-path reachability: this is not a boundary/recovery-only removal — it applies to every write of every TIMESTAMP column, INT96-tagged or not (INT96 columns take a completely separate `GenericData.Fixed` path in `resolveObject` and never had any logical-type schema annotation to begin with, so they were never in scope either).
- The removal touches only the `dataModel` object built fresh inside `getOrCreateOutputStream`; it does not affect `buildAvroSchemaWithRowType`/`seaTunnelDataType2ParquetDataType`, so the produced Avro/Parquet schema is byte-for-byte identical before and after this change.
- The added comment correctly encodes both reachability arguments and links back to #11743 for anyone who wants to re-derive the analysis later — good practice for a "prove a negative" cleanup like this.
- I did not find any additional call site in this file (or its INT96 branch) that constructs a `LocalDateTime`/`Instant` object and hands it to the writer directly — every TIMESTAMP value flows through `resolveObject`.

**Complete runtime path (write path for a TIMESTAMP column, confirming the datum never carries a schema-driven conversion):**
```
AbstractWriteStrategy#write(SeaTunnelRow)
  -> ParquetWriteStrategy#write(SeaTunnelRow)
       -> getOrCreateOutputStream(filePath)        // builds/caches AvroParquetWriter + dataModel (this diff)
       -> for each sink column:
            resolveObject(fieldName, field, TIMESTAMP)
              -> LocalDateTime -> Instant -> epochMilli (Long)   // manual conversion, bypasses GenericData
       -> GenericRecordBuilder.set(fieldName, Long)
       -> writer.write(GenericData.Record)
            -> AvroWriteSupport (uses dataModel only to look up a Conversion keyed on
               (datum runtime class, schema logicalType); datum is Long, schema logicalType is
               timestamp-millis (never local-timestamp-millis) -> no conversion ever matches
               -> long is written as the raw INT64 physical value)
```

## 1.2 Compatibility Impact
**Fully compatible.** No config option, default value, public API, or SPI contract changes. The produced Parquet schema and the physical bytes written for TIMESTAMP columns are unchanged (verified above: schema construction path is untouched, and the value path never depended on the removed conversion). Existing files written by older versions of this class remain readable exactly as before — this change only affects an unused registration on the *write* side, not read handling (`ParquetReadStrategy` is untouched).

Strictly positive/neutral: one fewer object allocation (`new TimeConversions.LocalTimestampMillisConversion()`) per call to `getOrCreateOutputStream`, which — worth noting as a pre-existing, unrelated observation — currently reconstructs `dataModel` and re-registers all conversions on *every* `write()` call (not just once per file), since the `dataModel`/`addLogicalTypeConversion` calls sit above the `if (writer == null)` check. That per-row reconstruction is out of scope for this PR (it's already tracked separately in the still-open #11660, "Build the Parquet Avro data model once per file instead of once per row") and this PR does not make it worse or better in kind, just removes one of the three now-wasted allocations.

N/A — no error handling or logging paths touched. See Issue 1 and Issue 2 below (both Low, non-blocking).

# 2. Code Quality Assessment

## 2.1 Coding Standards
The replacement comment is exactly the kind of "explain why, not what" comment this codebase should have here: it states the two independent reasons the registration is dead and links to the issue for anyone who wants to re-verify. No new methods/fields were added, so the "every new core method/field needs a comment" bar doesn't apply here — the change is a deletion plus a comment, which is the right shape for a "removing something that looks live but isn't" PR.

No tests were added or modified by this diff. I checked whether existing coverage would have caught a wrong removal: `ParquetTypeCoercionTest#testCanonicalTypesRoundTripFaithfully` (pre-existing, unmodified, `seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/.../writer/ParquetTypeCoercionTest.java:110-174`) writes a `LocalDateTime` TIMESTAMP value through this exact non-INT96 `ParquetWriteStrategy` path and asserts the value read back via `ParquetReadStrategy` equals the original `LocalDateTime` exactly. That test is part of this PR's CI run (`Build` = SUCCESS on head `386d30c`) and continues to pass, which is consistent with (though not a substitute for) the source-level reachability proof above. Rating: **Stable** — no flaky-test anti-patterns introduced (no new tests), and the one directly-relevant existing regression test continues to exercise the exact code path touched.

I'd note as a non-blocking improvement (Issue 1 below) that there's no test that pins the *produced schema's logical type name* itself (e.g. asserting the field's Avro logical type is `timestamp-millis`, not `local-timestamp-millis`). Today's value-based round trip would not necessarily catch a future refactor that changes `resolveObject` to hand Avro a `LocalDateTime` directly instead of a pre-converted `Long` — that would silently need this conversion back, and only a schema/logical-type-level assertion would guard against it before it correctness-breaks silently.

## 2.3 Documentation Updates
Not required — no user-visible config, default, or behavior change; `docs/en` and `docs/zh` do not need updates.

# 3. Architectural Soundness

**Precise fix.** This is exactly the right shape for a "prove unreachable, then delete" cleanup — the analysis is sound (independently re-verified above against the actual parquet-avro 1.12.3 source and the actual `resolveObject`/`seaTunnelDataType2ParquetDataType` code), the deletion is minimal, and the comment left behind documents the reasoning instead of just silently dropping the line.

## 3.2 Maintainability
Improves maintainability: removes a line that looked purpose-built (and therefore load-bearing) to any future reader, and replaces it with an explanation instead of leaving a silent gap.

## 3.3 Extensibility
Neutral. Does not affect the ability to add e.g. `TimestampMillisConversion` (the UTC-adjusted variant) later if a future change actually starts handing Avro raw `LocalDateTime`/`Instant` objects for TIMESTAMP fields — the comment even signals where a reader would need to look if that ever happens.

No impact on historical Parquet files or upgrade paths — this only touches an unused write-time registration; read compatibility and previously-written file compatibility are both untouched.

# 4. Issue Summary

|---|---|---|---|
| 1 | No test pins the produced Avro field's logical type (`timestamp-millis`), so a future `resolveObject`/schema-builder refactor that starts handing Avro a `LocalDateTime`/`Instant` directly could silently need `LocalTimestampMillisConversion` (or `TimestampMillisConversion`) back without any existing test failing to flag it. | `seatunnel-connectors-v2/connector-file/connector-file-base/src/test/java/.../writer/ParquetTypeCoercionTest.java` (suggested new test) | Low |
| 2 | PR title (`fix: remove unreachable LocalTimestampMillisConversion from ParquetWriteStrategy data model [#11743]`) doesn't follow this repo's commit/PR convention `[Type][Module] Description` (e.g. `[Fix][Connector-V2] Remove unreachable LocalTimestampMillisConversion in ParquetWriteStrategy`). Administrative only, not a functional concern. | PR title | Low |

No blocking issues found.

# 5. Merge Recommendation
1. Blockers — none.
   - Issue 1: consider adding a small assertion (unit test) on the generated Avro schema's TIMESTAMP field logical type name, to make the "this registration is truly dead" guarantee durable against future refactors of `resolveObject`/`seaTunnelDataType2ParquetDataType`, rather than only being provable by re-reading the source each time.
   - Issue 2: retitle to match `[Fix][Connector-V2] ...` convention before merge, purely for changelog/searchability consistency with the rest of the repo.

Overall assessment: I independently re-derived both reachability claims from #11743 against the current `dev` head — pulling the actual `parquet-avro:1.12.3` `AvroSchemaConverter` source and re-reading `resolveObject`/`seaTunnelDataType2ParquetDataType` line by line — and both come back negative exactly as the issue's "what to check" list required before a safe removal: the schema this class emits can never carry `local-timestamp-millis`, and the value path bypasses the data model's conversion lookup entirely regardless of schema. This is a correct, well-scoped, comment-documented dead-code removal with zero behavior change and adequate (if not maximally future-proof) existing regression coverage via `ParquetTypeCoercionTest#testCanonicalTypesRoundTripFaithfully`. I don't see a better alternative implementation — deleting the dead registration and explaining why is the right call here, not e.g. adding the "correct" `TimestampMillisConversion` in its place, since that would also be dead code today for the same claim-2 reason (the value path never hands Avro a `LocalDateTime`/`Instant`).

davidzollo:

```

```

davidzollo (approved):
Clean, well-reasoned dead-code removal — the two independent reachability arguments (schema never carries local-timestamp-millis in parquet-avro 1.12.3, and resolveObject hands Avro a pre-converted Long regardless of schema) are correctly derived from the current source, and the existing ParquetTypeCoercionTest continues to cover the write path. No behavior change. Approving to clear the review gate — merging now.

PR #11745, opened 2026-08-10
[Docs][Connector-V2] Improve DuckDB HiveJdbc Vitess-CDC Redshift and S3-Redshift connector docs

DanielCarter-stack:
## Summary

Documentation improvements for 5 connectors that have not yet been touched by this batch series:

- **DuckDB** (source + sink): clearer description of the in-process DuckDB model, a fully populated description column in the Sink Options table, the previously-missing table header, and `hocon`-tagged example code blocks.
- **HiveJdbc** (source): description that contrasts the JDBC-based path with the Hive-file source, tightened Options descriptions, fixed data type mapping, a Kerberos-aware task example, and `hocon`-tagged example code blocks.
- **Vitess-CDC** (source): expanded description covering streaming CDC semantics and the `LATEST` vs `SPECIFIC` startup modes.
- **Redshift** (source + sink): a new Source Options table was missing entirely and has been added; the sink options descriptions now enumerate the supported values for `schema_save_mode` and `data_save_mode`; example code blocks are tagged as `hocon`.
- **S3-Redshift** (sink): a description column is added to the Options table, the three example blocks now use the correct `access_secret` (not `secret_key`) and drop the unsourced `tmp_path`, and the prose descriptions are tightened.

Both `docs/en` and `docs/zh` are updated for the entries where the Chinese version already existed with the same structure.

## Why this matters

These five connectors were the last untouched in the connector hierarchy, and several had recurring issues that show up in `--help` and forum traffic:

- The DuckDB sink options table had no description column and listed `data_save_mode` without its allowed values, so users had to follow Option chain references to find them.
- HiveJdbc repeated the "Read external data source data through JDBC." one-liner without explaining when to pick it over the Hive source, and the example list did not include Kerberos even though the option table already advertises it.
- The Redshift source had no options table at all — every option was buried in the example block.
- S3-Redshift examples used `secret_key`, which doesn't exist in the options table, and `tmp_path`, which also doesn't exist.

## Scope

Documentation-only change. No source, config, or SPI code is touched.

## Test plan

- Local: verified the rendered Markdown by reviewing the diff and the docs/zh counterparts.
- CI: the existing Build / labeler / Notify checks run on PR open.

### DIFF
--- a/docs/en/connectors/sink/DuckDB.md
+++ b/docs/en/connectors/sink/DuckDB.md
@@ -16,8 +16,10 @@ import ChangeLog from '../changelog/connector-jdbc.md';
 
 ## Description
 
-Write data through jdbc. Support Batch mode and Streaming mode, support concurrent writing, support exactly-once
-semantics (using XA transaction guarantee).
+Write data to a DuckDB database file through JDBC. Supports batch and streaming modes, supports concurrent
+writing, and supports exactly-once semantics when the underlying JDBC driver exposes an XA datasource
+(set `is_exactly_once = true` and provide `xa_data_source_class_name`). DuckDB runs in-process, so the connector
+works against a local database file path (`jdbc:duckdb:/path/to/database.db`) or an in-memory database.
 
 ## Using Dependency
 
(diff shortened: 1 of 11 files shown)


DanielLeens:
Hi @DanielCarter-stack, thanks for putting this docs improvement together for the DuckDB, Hive JDBC, Vitess-CDC, Redshift and S3-Redshift connectors!

I see this PR is still marked as a draft, so I'll hold off on a full review for now to avoid reviewing content that might still be in flux. Once you mark it ready for review, I'll go through it in detail, including cross-checking the documented config option names/defaults/examples against the current source for each connector. Feel free to ping me here when it's ready.

zhangshenghang (approved):
+1

PR #11455, opened 2026-07-15
[Fix][E2E] Avoid EdgeSocket driver downloads

nzw921rx:
### Purpose

Remove flaky network dependencies from the EdgeSocket E2E setup. The tests previously downloaded the MySQL JDBC driver from Maven Central inside the CI container.

### Changes

- Resolve the MySQL JDBC driver jar from the active test classpath.
- Copy the resolved jar into the JDBC plugin lib directory before the E2E jobs run.
- Keep the helper scoped to the EdgeSocket E2E abstract test class.

### User-facing

No user-facing behavior change. This only stabilizes E2E test setup.

### How tested

- Ran ./mvnw spotless:apply
- Did not run tests per maintainer request

PR #11747, opened 2026-08-10
[Feature][Connector-V2] Add Hudi timer flush

nzw921rx:
## Purpose

- Register Hudi's existing synchronized flush path for Zeta timer flush signals.
- Keep the registered action bound to the current record writer after multi-table resource initialization replaces the original writer.
- Document the Zeta-only, at-least-once behavior in English and Chinese.

## Tests

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-hudi -DskipIT -Dtest=HudiSinkWriterTest -DargLine=-Djdk.attach.allowAttachSelf=true test`
- `TEST_IN_PR=true RUN_ALL_CONTAINER=false RUN_ZETA_CONTAINER=true ./mvnw -pl seatunnel-connectors-v2/connector-hudi,seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hudi-e2e -DskipUT -DskipIT=false -Dit.test=HudiSinkCDCIT#testHudiTimerFlush -DfailIfNoTests=false integration-test`

The E2E test verifies that two MySQL CDC records become visible in Hudi through separate timer flushes while the streaming job is still running.

### DIFF
--- a/docs/en/connectors/sink/Hudi.md
+++ b/docs/en/connectors/sink/Hudi.md
@@ -108,7 +108,8 @@ Note: When this configuration corresponds to a single table, you can flatten the
 
 ### batch_interval_ms [Int]
 
-`batch_interval_ms` The maximum interval, in milliseconds, between two flushes to Hudi.
+`batch_interval_ms` is retained for compatibility. To schedule time-based flushes on Zeta, configure
+`sink.flush.interval` in the job `env` block.
 
 ### batch_size [Int]
 
@@ -156,6 +157,23 @@ Choose how to handle existing data before the synchronization task starts.
 ### common options
 
 Sink plugin common parameters, please refer to [Sink Common Options](../common-options/sink-common-options.md) for details.
+
+## Timer Flush
+
+Timer flush is an engine-level feature supported only by Zeta. Configure `sink.flush.interval` in the job `env` block
+to write pending Hudi records even when `batch_size` has not been reached. Spark and Flink do not inject `FlushSignal`
+records and therefore do not trigger this scheduled flush.
+
+```hocon
+env {
+  sink.flush.interval = 5000
+}
+```
+
+Hudi timer flush reuses the connector's synchronized batch flush and the Hudi client's auto-commit behavior. The Hudi
+sink does not provide a 2PC exactly-once writer, so timer flush provides at-least-once delivery. Retries can create
+additional commits. With `INSERT`, generated record keys can also produce duplicate rows after recovery; `UPSERT` with
+stable `record_key_fields` limits duplicate logical records.
 
 ## Examples
 
--- a/docs/zh/connectors/sink/Hudi.md
+++ b/docs/zh/connectors/sink/Hudi.md
@@ -108,7 +108,8 @@ SeaTunnel Hudi sink 会写入 Hudi 数据文件和 `.hoodie` 元数据，但不�
 
 ### batch_interval_ms [Int]
 
-`batch_interval_ms` 两次刷新到 Hudi 的最大时间间隔，单位为毫秒。
+`batch_interval_ms` 为兼容性保留。在 Zeta 上需要定时刷新时，请在作业 `env` 中配置
+`sink.flush.interval`。
 
 ### batch_size [Int]
 
@@ -153,6 +154,22 @@ SeaTunnel Hudi sink 会写入 Hudi 数据文件和 `.hoodie` 元数据，但不�
 ### 通用选项
 
 Sink插件通用参数，请参考 [Sink Common Options](../common-options/sink-common-options.md) 了解详细信息。
+
+## 定时刷新
+
+定时刷新是仅由 Zeta 支持的引擎级能力。在作业的 `env` 中配置 `sink.flush.interval` 后，即使尚未达到
+`batch_size`，Hudi Sink 也会写出待处理的记录。Spark 和 Flink 不会注入 `FlushSignal`，因此不会触发这种
+定时刷新。
+
+```hocon
+env {
+  sink.flush.interval = 5000
+}
+```
+
+Hudi 定时刷新复用连接器现有的同步批量刷新和 Hudi 客户端 auto-commit 行为。Hudi Sink 没有 2PC 精确一次
+写入器，因此定时刷新提供的是至少一次语义，重试可能产生额外的 commit。使用 `INSERT` 时，自动生成的
+record key 还可能在恢复后产生重复行；使用具有稳定 `record_key_fields` 的 `UPSERT` 可以减少逻辑记录重复。
 
 ## 示例
 
--- a/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
@@ -114,9 +114,8 @@ public class HudiSinkWriter
     }
 
     /**
-     * Flushes buffered records when the sink receives a timer-generated FlushSignal. Once accepted
-     * by the pipeline, the signal is processed in order with data records and checkpoint barriers
-     * on the sink task thread.
+     * Flushes buffered records when the sink receives a timer-generated FlushSignal. The signal is
+     * processed on the sink task thread in order with data records and checkpoint barriers.
      */
     private void timerFlush() {
         hudiRecordWriter.flush();
--- a/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
@@ -114,8 +114,9 @@ public class HudiSinkWriter
     }
 
     /**
-     * Flushes buffered records when Zeta delivers a timer-generated FlushSignal. The signal is
-     * delivered through the same ordered pipeline path as data and checkpoint barriers.
+     * Flushes buffered records when the sink receives a timer-generated FlushSignal. Once accepted
+     * by the pipeline, the signal is processed in order with data records and checkpoint barriers
+     * on the sink task thread.
      */
     private void timerFlush() {
         hudiRecordWriter.flush();
--- a/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
@@ -67,6 +67,7 @@ public class HudiSinkWriter
                         sinkConfig, tableConfig.getTableName(), seaTunnelRowType);
         this.hudiRecordWriter =
                 new HudiRecordWriter(tableConfig, writeClientProvider, seaTunnelRowType);
+        context.registerFlushAction(this::timerFlush);
     }
 
     @Override
@@ -110,6 +111,10 @@ public class HudiSinkWriter
                         tableConfig.getTableName());
         this.hudiRecordWriter =
                 new HudiRecordWriter(tableConfig, writeClientProvider, seaTunnelRowType);
+    }
+
+    private void timerFlush() {
+        hudiRecordWriter.flush();
     }
 
     private void tryOpen() {
--- a/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
@@ -113,6 +113,10 @@ public class HudiSinkWriter
                 new HudiRecordWriter(tableConfig, writeClientProvider, seaTunnelRowType);
     }
 
+    /**
+     * Flushes buffered records when Zeta delivers a timer-generated FlushSignal. The signal is
+     * handled on the sink task thread, serializing timer flushes with writes and checkpoints.
+     */
     private void timerFlush() {
         hudiRecordWriter.flush();
     }
--- a/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-hudi/src/main/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriter.java
@@ -115,7 +115,7 @@ public class HudiSinkWriter
 
     /**
      * Flushes buffered records when Zeta delivers a timer-generated FlushSignal. The signal is
-     * handled on the sink task thread, serializing timer flushes with writes and checkpoints.
+     * delivered through the same ordered pipeline path as data and checkpoint barriers.
      */
     private void timerFlush() {
         hudiRecordWriter.flush();
--- a/seatunnel-connectors-v2/connector-hudi/src/test/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriterTest.java
+++ b/seatunnel-connectors-v2/connector-hudi/src/test/java/org/apache/seatunnel/connectors/seatunnel/hudi/sink/writer/HudiSinkWriterTest.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.connectors.seatunnel.hudi.sink.writer;
+
+import org.apache.seatunnel.api.sink.MultiTableResourceManager;
+import org.apache.seatunnel.api.sink.SinkWriter;
+import org.apache.seatunnel.api.table.type.SeaTunnelRowType;
+import org.apache.seatunnel.common.utils.function.RunnableWithException;
+import org.apache.seatunnel.connectors.seatunnel.hudi.config.HudiSinkConfig;
+import org.apache.seatunnel.connectors.seatunnel.hudi.config.HudiTableConfig;
+import org.apache.seatunnel.connectors.seatunnel.hudi.exception.HudiConnectorException;
+import org.apache.seatunnel.connectors.seatunnel.hudi.exception.HudiErrorCode;
+import org.apache.seatunnel.connectors.seatunnel.hudi.sink.HudiClientManager;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.MockedConstruction;
+import org.mockito.Mockito;
+
+import java.util.Optional;
+
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class HudiSinkWriterTest {
+
+    @Test
+    void shouldRegisterAndExecuteTimerFlush() throws Exception {
+        SinkWriter.Context context = mock(SinkWriter.Context.class);
+        ArgumentCaptor<RunnableWithException> actionCaptor =
+                ArgumentCaptor.forClass(RunnableWithException.class);
+
+        try (MockedConstruction<HudiRecordWriter> recordWriters = createWriter(context)) {
+            verify(context, times(1)).registerFlushAction(actionCaptor.capture());
+
+            actionCaptor.getValue().run();
+
+            verify(recordWriters.constructed().get(0), times(1)).flush();
+        }
+    }
+
+    @Test
+    void shouldPropagateTimerFlushFailure() throws Exception {
+        SinkWriter.Context context = mock(SinkWriter.Context.class);
+        ArgumentCaptor<RunnableWithException> actionCaptor =
+                ArgumentCaptor.forClass(RunnableWithException.class);
+
+        try (MockedConstruction<HudiRecordWriter> recordWriters = createWriter(context)) {
+            HudiConnectorException expected =
+                    new HudiConnectorException(
+                            HudiErrorCode.FLUSH_DATA_FAILED, "timer flush failed");
+            doThrow(expected).when(recordWriters.constructed().get(0)).flush();
+            verify(context).registerFlushAction(actionCaptor.capture());
+
+            HudiConnectorException actual =
+                    Assertions.assertThrows(
+                            HudiConnectorException.class, actionCaptor.getValue()::run);
+
+            Assertions.assertSame(expected, actual);
+        }
+    }
+
+    @Test
+    void shouldFlushCurrentRecordWriterAfterResourceManagerReplacement() throws Exception {
+        SinkWriter.Context context = mock(SinkWriter.Context.class);
+        ArgumentCaptor<RunnableWithException> actionCaptor =
+                ArgumentCaptor.forClass(RunnableWithException.class);
+        MultiTableResourceManager<HudiClientManager> resourceManager =
+                mock(MultiTableResourceManager.class);
+        when(resourceManager.getSharedResource())
+                .thenReturn(Optional.of(mock(HudiClientManager.class)));
+
+        try (MockedConstruction<HudiRecordWriter> recordWriters =
+                Mockito.mockConstruction(HudiRecordWriter.class)) {
+            HudiSinkWriter writer =
+                    new HudiSinkWriter(
+                            context,
+                            mock(SeaTunnelRowType.class),
+                            mock(HudiSinkConfig.class),
+                            mock(HudiTableConfig.class));
+            verify(context).registerFlushAction(actionCaptor.capture());
+
+            writer.setMultiTableResourceManager(resourceManager, 0);
+            actionCaptor.getValue().run();
+
+            Assertions.assertEquals(2, recordWriters.constructed().size());
+            verify(recordWriters.constructed().get(0), never()).flush();
+            verify(recordWriters.constructed().get(1)).flush();
+        }
+    }
+
+    private MockedConstruction<HudiRecordWriter> createWriter(SinkWriter.Context context) {
+        MockedConstruction<HudiRecordWriter> recordWriters =
+                Mockito.mockConstruction(HudiRecordWriter.class);
+        new HudiSinkWriter(
+                context,
+                mock(SeaTunnelRowType.class),
+                mock(HudiSinkConfig.class),
+                mock(HudiTableConfig.class));
+        return recordWriters;
+    }
+}
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hudi-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hudi/HudiSinkCDCIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hudi-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hudi/HudiSinkCDCIT.java
@@ -46,17 +46,13 @@ import org.testcontainers.containers.Container;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.utility.DockerLoggerFactory;
-import org.testcontainers.utility.MountableFile;
 
-import com.mysql.cj.jdbc.Driver;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
@@ -87,7 +83,8 @@ public class HudiSinkCDCIT extends TestSuiteBase implements TestResource {
     private static final MySqlContainer MYSQL_CONTAINER = createMySqlContainer(MySqlVersion.V8_0);
     private static final String SOURCE_TABLE = "mysql_cdc_e2e_source_table";
 
-    private static final String MYSQL_CDC_PLUGIN_LIB = "/tmp/seatunnel/plugins/MySQL-CDC/lib";
+    private static final String MYSQL_DRIVER =
+            "https://repo1.maven.org/maven2/com/mysql/mysql-connector-j/8.0.32/mysql-connector-j-8.0.32.jar";
 
     private static final String DATABASE = "st";
     private static final String TABLE_NAME = "st_test";
@@ -121,28 +118,14 @@ public class HudiSinkCDCIT extends TestSuiteBase implements TestResource {
             container -> {
                 container.execInContainer("sh", "-c", "mkdir -p " + TABLE_PATH);
                 container.execInContainer("sh", "-c", "chmod -R 777  " + TABLE_PATH);
-                java.nio.file.Path driverJarPath = driverJarPath();
-                Assertions.assertTrue(
-                        Files.isRegularFile(driverJarPath),
-                        "MySQL JDBC driver should be resolved from the test classpath before E2E runs: "
-                                + driverJarPath);
                 Container.ExecResult extraCommands =
-                        container.execInContainer("sh", "-c", "mkdir -p " + MYSQL_CDC_PLUGIN_LIB);
+                        container.execInContainer(
+                                "sh",
+                                "-c",
+                                "mkdir -p /tmp/seatunnel/plugins/MySQL-CDC/lib && cd /tmp/seatunnel/plugins/MySQL-CDC/lib && wget "
+                                        + MYSQL_DRIVER);
                 Assertions.assertEquals(0, extraCommands.getExitCode(), extraCommands.getStderr());
-                container.copyFileToContainer(
-                        MountableFile.forHostPath(driverJarPath),
-                        MYSQL_CDC_PLUGIN_LIB + "/" + driverJarPath.getFileName());
             };
-
-    private java.nio.file.Path driverJarPath() {
-        try {
-            return Paths.get(
-                    Driver.class.getProtectionDomain().getCodeSource().getLocation().toURI());
-        } catch (Exception e) {
-            throw new RuntimeException(
-                    "Failed to resolve MySQL JDBC driver jar from the test classpath", e);
-        }
-    }
 
     @BeforeAll
     @Override
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hudi-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hudi/HudiSinkCDCIT.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-hudi-e2e/src/test/java/org/apache/seatunnel/e2e/connector/hudi/HudiSinkCDCIT.java
@@ -28,6 +28,7 @@ import org.apache.seatunnel.e2e.common.container.EngineType;
 import org.apache.seatunnel.e2e.common.container.TestContainer;
 import org.apache.seatunnel.e2e.common.junit.DisabledOnContainer;
 import org.apache.seatunnel.e2e.common.junit.TestContainerExtension;
+import org.apache.seatunnel.e2e.common.util.JobIdGenerator;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileUtil;
@@ -45,13 +46,17 @@ import org.testcontainers.containers.Container;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.utility.DockerLoggerFactory;
+import org.testcontainers.utility.MountableFile;
 
+import com.mysql.cj.jdbc.Driver;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
@@ -82,11 +87,12 @@ public class HudiSinkCDCIT extends TestSuiteBase implements TestResource {
     private static final MySqlContainer MYSQL_CONTAINER = createMySqlContainer(MySqlVersion.V8_0);
     private static final String SOURCE_TABLE = "mysql_cdc_e2e_source_table";
 
-    private static final String MYSQL_DRIVER =
-            "https://repo1.maven.org/maven2/com/mysql/mysql-connector-j/8.0.32/mysql-connector-j-8.0.32.jar";
+    private static final String MYSQL_CDC_PLUGIN_LIB = "/tmp/seatunnel/plugins/MySQL-CDC/lib";
 
     private static final String DATABASE = "st";
     private static final String TABLE_NAME = "st_test";
+    private static final String TIMER_FLUSH_DATABASE = "timer_flush_db";
+    private static final String TIMER_FLUSH_TABLE = "timer_flush_table";
     private static final String TABLE_PATH = HOST_VOLUME_MOUNT_PATH + "/hudi/";
     private static final String NAMESPACE = "hudi";
     private static final String NAMESPACE_TAR = "hudi.tar.gz";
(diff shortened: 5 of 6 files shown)


DanielLeens (commented):
- User pain point: The Hudi sink only flushes buffered records on `batch_size` or on a checkpoint-triggered `prepareCommit()`. In a low-throughput streaming job, rows can sit in the in-memory buffer for a long time (until the next checkpoint) before becoming visible in the Hudi table, hurting data freshness.
- Fix approach: `HudiSinkWriter` registers its existing `HudiRecordWriter.flush()` routine as the writer's flush action via `context.registerFlushAction(this::timerFlush)`. This hooks into an **already-existing, engine-level timer-flush mechanism** (`sink.flush.interval` in `env`, `SourceFlowLifeCycle`'s `timerFlushWorker` → `FlushSignal` → `SinkFlowLifeCycle.processSignal`) that six other sinks (Doris, Elasticsearch, JDBC, MongoDB, StarRocks, ClickHouse) already use in production today. This PR is the "complete the set" change that wires Hudi into that pattern.
- One-sentence summary: Adds Hudi to the existing Zeta timer-flush framework so buffered rows are periodically written out between checkpoints, with no new engine machinery introduced.

# 1. Code Change Review

## 1.1 Core Logic Analysis

`seatunnel-connectors-v2/connector-hudi/.../sink/writer/HudiSinkWriter.java` (+5/-0):

```java
// constructor, after hudiRecordWriter is created
context.registerFlushAction(this::timerFlush);
...
private void timerFlush() {
    hudiRecordWriter.flush();
}
```

**Runtime path** (verified against current `dev`, none of this plumbing is new in this diff):

```
env.sink.flush.interval > 0
  -> SourceFlowLifeCycle.hook() -> startFlushTimer()
     -> TaskExecutionService.timerFlushWorker (ScheduledThreadPoolExecutor)
        .scheduleWithFixedDelay(onTimerTick, ...)
  -> onTimerTick() [runs on timerFlushWorker thread, guarded by same checkpointLock as triggerBarrier()]
     -> collector.sendFlushSignal(jobId, taskId)
        -> SeaTunnelSourceCollector.sendRecordToNext(Record<FlushSignal>)
           [FlushSignal now flows through the pipeline as an ordinary Record]
  -> downstream Sink task's single-threaded processing loop dequeues the Record
     -> SinkFlowLifeCycle.processSignal(FlushSignal)
        -> writerContext.getFlushAction().run()   // == HudiSinkWriter.timerFlush()
           -> hudiRecordWriter.flush()            // synchronized, same method prepareCommit() calls
```

Because `FlushSignal` is injected as a regular `Record` into the same single-consumer queue that carries data records and checkpoint barriers for a given sink subtask, the flush action is **serialized with `write()` and `prepareCommit()` on the same task thread** — there is no new concurrent-access surface introduced by this PR. `HudiRecordWriter.flush()` is itself `synchronized` as defense in depth, and it early-returns when `batchCount == 0` (`HudiRecordWriter.java:126-129`), so a timer tick that fires before the writer has ever received a record (i.e. before `tryOpen()`/`writeRecord()` ran) is a safe no-op — it does not touch `clientProvider.getOrCreateClient()` and cannot NPE.

**Key Findings:**
1. `timerFlush()` reads `this.hudiRecordWriter` (a field) at call time, not a captured local — so it correctly targets the *replacement* writer created in `setMultiTableResourceManager()` for multi-table sink mode. This is explicitly unit-tested (see 2.2).
2. No new thread, executor, or lock is introduced by this connector; it reuses the platform-wide `timerFlushWorker` already exercised by 6 other connectors.
3. `HudiRecordWriter.flush()`/`close()` are both `synchronized` on the same monitor, so even if the single-thread assumption were ever violated, a flush cannot interleave with a close mid-write.
4. `HudiRecordWriter.flush()`/`close()` have no `closed` guard inside `flush()` itself — if a flush action were somehow invoked after `close()` already ran (`clientProvider.close()` executed), it would call `clientProvider.getOrCreateClient()` against a closed provider. I could not find a path in the current engine code where this happens (source-side timer is stopped in `SourceFlowLifeCycle.close()`'s `finally` block before the pipeline drains), and the same gap exists unchanged in `JdbcSinkWriter.timerFlush()`/`close()` today, so this is a pre-existing platform-wide assumption, not something this PR introduces or worsens. Flagging as informational only (see Issue 1).
5. `batch_interval_ms` (the old, effectively dead per-connector option — grep confirms `HudiTableConfig.batchIntervalMs` is parsed and defaulted but never read anywhere in `HudiRecordWriter`/`HudiSinkWriter`) is left in place and merely re-documented as "retained for compatibility." No option is removed or renamed.

Verified: **no changes anywhere in this diff touch Hudi connector registration, `plugin-mapping.properties`, `HudiSinkFactory`, `HudiSink`, `META-INF/services`, or any `pom.xml`** (`git diff --name-status` confirms only `HudiSinkWriter.java`, its new test, the E2E test + new `.conf`, and the two doc files changed). Given this repo's prior Hudi-registration incident, I explicitly checked for this and found nothing to flag.

## 1.2 Compatibility Impact
**Fully compatible.**
- `SinkWriter.Context.registerFlushAction` is a `default void ... {}` method (`SinkWriter.java:131`), and `startFlushTimer()` no-ops when `flushIntervalMs <= 0`. Existing Hudi jobs that don't set `sink.flush.interval` get exactly the previous behavior — flush only on `batch_size` or checkpoint `prepareCommit()`.
- The old `batch_interval_ms` option is not removed, renamed, or given new semantics; it stays a parsed-but-unused field exactly as before.
- No config option default values changed, no public API/SPI signature changed.

- No new per-task thread; reuses the shared `timerFlushWorker` pool (sized by `EngineConfig.timerFlushPoolSize`, pre-existing config, unaffected by this PR).
- Additional flush cycles under a short `sink.flush.interval` will increase the frequency of Hudi commits (`writeClient.startCommit()`), which is the intended behavior but is worth calling out explicitly in the docs as a tradeoff — the PR's own doc addition already does this ("Hudi timer flush ... provides at-least-once delivery. Retries can create additional commits... `INSERT` ... can also produce duplicate rows after recovery"), which is accurate and appropriately candid.

No new formal issues beyond Issue 1 below (informational/Low, see 2.1 and 4).

# 2. Code Quality Assessment

## 2.1 Coding Standards
- `timerFlush()` is a new private method with no Javadoc. It's a one-line pass-through, but per this repo's convention that new non-trivial methods get a comment explaining intent/constraints, a short comment (e.g. "Registered as the writer's periodic flush action; invoked on the task's single processing thread via FlushSignal, safe to call before open() since flush() no-ops on empty batches") would materially help future readers understand the threading guarantee this method silently depends on. Not a blocker, but flagged for the author (see Issue 1).
- The one-line `context.registerFlushAction(this::timerFlush);` addition to the constructor has no comment either; same minor request.
- Otherwise the diff follows existing style, matches sibling connectors' idiom exactly (`this::timerFlush` naming is even shared verbatim with `DorisSinkWriter`, `ElasticsearchSinkWriter`, `StarRocksSinkWriter`, `MongodbWriter`, `JdbcSinkWriter`).

**Unit test** (`HudiSinkWriterTest.java`, new, 122 lines) — **Stable**.
- `shouldRegisterAndExecuteTimerFlush`: verifies `registerFlushAction` is called once and that invoking the captured action calls `HudiRecordWriter.flush()`. Deterministic, uses `MockedConstruction`, no sleeps.
- `shouldPropagateTimerFlushFailure`: verifies a `flush()` exception thrown by the timer path propagates (not swallowed). Good negative-path coverage.
- `shouldFlushCurrentRecordWriterAfterResourceManagerReplacement`: this is the one I would have flagged as untested if it were missing — it directly proves the "does timerFlush target the writer replaced by `setMultiTableResourceManager`" question I raised in 1.1 Finding 1, by asserting the *first* constructed `HudiRecordWriter` is never flushed and the *second* one is. Excellent, precisely targeted regression coverage.
- No `Thread.sleep`, no nondeterministic assertions, no shared mutable state between tests, no resource leaks (mocked construction is scoped in try-with-resources).

**E2E test** (`HudiSinkCDCIT.java`, +138/-7) — **Stable**, well above the bar for this kind of timing-sensitive test.
- New `testHudiTimerFlush` uses Awaitility (`given().await().atMost(...).pollInterval(...).untilAsserted(...)`) exclusively — no hard `Thread.sleep`.
- `assertJobStillRunning` fails fast with the job's stderr if the streaming job already terminated, rather than blindly polling and timing out with an opaque message — this is a good anti-flake pattern that surfaces the real failure cause.
- The test conf (`mysql_cdc_to_hudi_timer_flush.conf`) sets `checkpoint.interval = 300000` (5 min) and `batch_size = 100` while inserting only 1-2 rows — this deliberately rules out the checkpoint-triggered and batch-size-triggered flush paths as an explanation for the assertion passing, so the test genuinely exercises the timer path and would fail if timer flush were broken. Good test design, not a tautological "always passes" test.
- Separately, the same file replaces a `wget`-from-Maven-Central MySQL driver download with copying the driver jar resolved from the test classpath (`Driver.class.getProtectionDomain()...`) — this removes a network-flakiness source from CI. It's not part of the "timer flush" feature and is scope creep relative to a strict one-PR-one-problem reading of the PR title, but it's a small, low-risk, net-positive CI robustness fix bundled in, not a functional change, and I would not block on it (see Issue 2, Low, non-blocking).
- Verified `mysql:mysql-connector-java` is already a pre-existing test-scope dependency of `connector-hudi-e2e/pom.xml` (untouched by this diff), so `com.mysql.cj.jdbc.Driver` is guaranteed present on the test classpath — no missing-dependency risk from this refactor.

## 2.3 Documentation Updates
`docs/en/connectors/sink/Hudi.md` and `docs/zh/connectors/sink/Hudi.md` are both updated in lockstep (verified line-by-line, content matches):
- `batch_interval_ms` entry is corrected to point users at the real mechanism (`sink.flush.interval` in `env`) instead of implying the old option does something.
- New "Timer Flush" / "定时刷新" section correctly scopes the feature to Zeta only ("Spark and Flink do not inject `FlushSignal` records"), gives a working example, and is candid about the at-least-once/duplicate-row tradeoff for `INSERT` vs `UPSERT`. This is accurate, matches my read of `HudiRecordWriter.executeWrite()`'s `INSERT`/`UPSERT`/`BULK_INSERT` branches, and is exactly the kind of caveat that should be documented rather than discovered by an operator in production.

# 3. Architectural Soundness

**Precise fix**, reusing an established framework. This is the most elegant amount of code that could implement this feature — 5 lines in the connector because all the hard concurrency/lifecycle work (timer scheduling, signal propagation, single-thread serialization with barriers) already lives in the engine and was already proven out across 6 other sinks before this PR.

## 3.2 Maintainability
High. The pattern is now applied consistently across 7 sinks with near-identical code shape (`context.registerFlushAction(this::timerFlush)` + a `timerFlush()` wrapper around the existing flush routine), which makes the codebase easier to reason about, not harder.

## 3.3 Extensibility
No concerns; any future sink can adopt the same one-line pattern.

No checkpoint/savepoint state format changes, no serialization changes, no SPI changes. `HudiSinkState`/`HudiCommitInfo` are untouched. A job checkpointed on a pre-this-PR version and restored on a post-this-PR version behaves identically unless the operator opts in by setting `sink.flush.interval` (a job-level `env` setting, not part of any persisted state).

# 4. Issue Summary

|---|---|---|---|
| 1 | New `timerFlush()` method and the `registerFlushAction` registration call have no explanatory comment describing the threading guarantee (single task-thread serialization via `FlushSignal`) they silently rely on, and `HudiRecordWriter.flush()`/`close()` have no `closed`-guard against a flush arriving after close (pre-existing pattern shared with `JdbcSinkWriter`, not unique to this PR, but worth a one-line comment now that a 7th connector depends on the same assumption) | `HudiSinkWriter.java:70,116-118`; `HudiRecordWriter.java:126,220` | Low |
| 2 | E2E test file bundles an unrelated (but low-risk, net-positive) fix — replacing `wget`-based MySQL driver download with classpath-resolved copy — inside a PR titled/scoped as "Add Hudi timer flush" | `HudiSinkCDCIT.java` (MYSQL_DRIVER / driverJarPath changes) | Low |

No blocking issues found. Both items above are non-blocking observations.

# 5. Merge Recommendation
### Conclusion: Ready to merge after CI passes

1. Blockers — none. Nothing here is a merge blocker.
2. Recommended fixes — non-blocking, can be addressed in this PR or a fast-follow:
   - Issue 1: add a short comment on `timerFlush()`/the registration call documenting the single-thread/FlushSignal serialization guarantee it depends on.
   - Issue 2: consider whether the MySQL-driver-download refactor belongs in a separate `[Fix][E2E]` PR for cleaner history (the author has a track record of doing exactly this kind of split, e.g. #11455/#11454/#11453/#11452/#11451/#11450/#11449 are all standalone "avoid X driver download" PRs) — optional, not required.

Overall assessment: this is a clean, low-risk, well-tested wiring change onto an already-proven platform mechanism. I independently traced the full `sink.flush.interval` → `timerFlushWorker` → `FlushSignal` → `SinkFlowLifeCycle.processSignal` → `writerContext.getFlushAction()` path in current `dev` to confirm there is no new race between the timer-triggered flush and the checkpoint-triggered `prepareCommit()`/`write()` path, confirmed `HudiRecordWriter.flush()` safely no-ops when called before the writer is opened, confirmed the flush action correctly re-targets the replacement writer after `setMultiTableResourceManager()` (and that this is explicitly unit-tested), confirmed zero touches to Hudi connector registration/plugin-mapping/factory/pom files, and confirmed full backward compatibility for jobs that don't set the new option. I don't see a materially better alternative implementation — it follows the exact same pattern already shipped for Doris/Elasticsearch/JDBC/MongoDB/StarRocks/ClickHouse, which is the right call for consistency.

CI note: at review time, both the `apache/seatunnel` "Build" check and the corresponding fork Actions run (`nzw921rx/seatunnel`, run `31392740999`) show status **queued/pending** — CI has not actually run yet, not just an unresolved apache-side pointer. Please make sure it goes green before merge.

DanielLeens (approved):
Thanks for the fast turnaround, @nzw921rx! I re-reviewed the full current diff against `dev` (not just the incremental commit) and checked it against the two non-blocking points from my previous pass.

- User pain point: The Hudi sink only flushed buffered records on `batch_size` or on a checkpoint-triggered `prepareCommit()`. In a low-throughput streaming job, rows could sit in the in-memory buffer for a long time (until the next checkpoint) before becoming visible in the Hudi table, hurting data freshness.
- Fix approach: `HudiSinkWriter` registers its existing `HudiRecordWriter.flush()` routine as the writer's flush action via `context.registerFlushAction(this::timerFlush)`, hooking into the already-existing, engine-level timer-flush mechanism (`sink.flush.interval` in `env` -> `SourceFlowLifeCycle`'s `timerFlushWorker` -> `FlushSignal` -> `SinkFlowLifeCycle.processSignal`) that six other sinks already use.
- One-sentence summary: Adds Hudi to the existing Zeta timer-flush framework so buffered rows are periodically written out between checkpoints, with no new engine machinery introduced.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Since my last review, one follow-up commit ("Address Hudi timer flush review feedback") landed. It touches exactly two files:

`HudiSinkWriter.java` (+4 lines) - adds a Javadoc comment to `timerFlush()`:

```java
/**
 * Flushes buffered records when Zeta delivers a timer-generated FlushSignal. The signal is
 * handled on the sink task thread, serializing timer flushes with writes and checkpoints.
 */
private void timerFlush() {
    hudiRecordWriter.flush();
}
```

`HudiSinkCDCIT.java` (-24/+11 net vs. the previous version of this PR) - fully reverts the classpath-copy MySQL-driver refactor back to the plain `wget`-from-Maven-Central approach that already existed on `dev`. I diffed `dev...HEAD` again (not just the incremental commit) and confirmed the net change to this file relative to `dev` is now *only* the new `testHudiTimerFlush` test plus its three private helper methods (`awaitTimerFlush`, `countNewestCommitRows`, `assertJobStillRunning`) - the driver-download code path is byte-for-byte back to what's on `dev` today. No other files changed in the follow-up commit; `docs/en`, `docs/zh`, and `HudiSinkWriterTest.java` are unchanged from what I reviewed last time.

**Runtime path** (re-verified against current `dev`, unchanged from my last review since nothing about the flush wiring itself changed):

```
env.sink.flush.interval > 0
  -> SourceFlowLifeCycle.hook() -> startFlushTimer()
     -> TaskExecutionService.timerFlushWorker (ScheduledThreadPoolExecutor)
        .scheduleWithFixedDelay(onTimerTick, ...)
  -> onTimerTick() [runs on timerFlushWorker thread]
     -> collector.sendFlushSignal(jobId, taskId)
        -> SeaTunnelSourceCollector.sendRecordToNext(Record<FlushSignal>)
  -> downstream Sink task's single-threaded processing loop dequeues the Record
     -> SinkFlowLifeCycle.processSignal(FlushSignal)
        -> writerContext.getFlushAction().run()   // == HudiSinkWriter.timerFlush()
           -> hudiRecordWriter.flush()            // synchronized, same method prepareCommit() calls
```

`FlushSignal` is still injected as a regular `Record` into the same single-consumer queue that carries data records and checkpoint barriers for a given sink subtask, so the flush action stays serialized with `write()`/`prepareCommit()` on the same task thread - no new concurrent-access surface.

**Key Findings (re-verified this round):**
1. My earlier Issue 1 (missing explanatory comment on `timerFlush()`) is now fixed - the added Javadoc correctly documents the single-task-thread/`FlushSignal` serialization guarantee the method depends on. Nice, this is exactly the kind of comment that saves the next reader from re-deriving the threading model from scratch.
2. My earlier Issue 2 (the unrelated MySQL-driver-download refactor bundled into a "timer flush" PR) is now resolved by simply reverting that part rather than splitting it into a separate PR - which is actually the cleaner outcome of the two options I raised, since it leaves this PR scoped to exactly what its title says. I verified this by diffing the full file against `dev` again: the driver-download block is textually identical to `dev`, so there is zero net change to that code path in this PR anymore.
3. `timerFlush()` still reads `this.hudiRecordWriter` (a field) at call time, so it still correctly targets the *replacement* writer created in `setMultiTableResourceManager()` for multi-table sink mode - this is unchanged from before and is still covered by `shouldFlushCurrentRecordWriterAfterResourceManagerReplacement` in the unit test.
4. `HudiRecordWriter.flush()`/`close()` (lines 131 and 226 in the current file) are unchanged - still `synchronized`, still early-return on `batchCount == 0`, still no explicit `closed` guard against a flush arriving after `close()`. As I noted last time, I can't find a reachable path where a timer flush fires after `close()` on the same task thread (the source-side timer is stopped before the pipeline drains), and the same gap exists unchanged in `JdbcSinkWriter` today - so this remains a pre-existing, platform-wide, informational-only observation, not something this PR introduces (see Issue 1 below, carried forward at the same Low severity).
5. Re-verified: this round's diff still touches nothing under Hudi connector registration, `plugin-mapping.properties`, `HudiSinkFactory`, `HudiSink`, `META-INF/services`, or any `pom.xml` (`git diff --name-status dev...HEAD` shows only `HudiSinkWriter.java`, `HudiSinkWriterTest.java`, `HudiSinkCDCIT.java`, the new `.conf`, and the two doc files). Given this repo's prior Hudi-registration incident I explicitly re-checked this on the fresh diff and found nothing to flag.

## 1.2 Compatibility Impact
**Fully compatible.** Unchanged from my last review: `registerFlushAction` is a `default void {}` method, `startFlushTimer()` no-ops when `flushIntervalMs <= 0`, existing Hudi jobs that don't set `sink.flush.interval` get exactly the previous behavior, and `batch_interval_ms` is not removed/renamed/repurposed.

No change from before: no new thread/executor, reuses the shared `timerFlushWorker` pool. More frequent commits under a short `sink.flush.interval` is the intended tradeoff and is documented.

See Issue 1 below (carried forward, Low, informational).

# 2. Code Quality Assessment

## 2.1 Coding Standards
- Issue 1 from my last review (missing comment on `timerFlush()`) is fixed.
- One very minor leftover: the one-line `context.registerFlushAction(this::timerFlush);` call in the constructor (`HudiSinkWriter.java:70`) still has no comment of its own, though the method it registers now does. This is trivial and not worth another blocking issue - the intent is clear from the now-documented `timerFlush()` right below it.
- Style otherwise matches sibling connectors' idiom exactly, as before.

**Unit test** (`HudiSinkWriterTest.java`) - unchanged from last review, still **Stable**. All three tests (`shouldRegisterAndExecuteTimerFlush`, `shouldPropagateTimerFlushFailure`, `shouldFlushCurrentRecordWriterAfterResourceManagerReplacement`) are deterministic, use `MockedConstruction`, no sleeps, no shared mutable state.

**E2E test** (`HudiSinkCDCIT.java`) - **Stable**, and the follow-up commit only improves scope/focus here (it doesn't touch the actual test logic in `testHudiTimerFlush`/`awaitTimerFlush`/`countNewestCommitRows`/`assertJobStillRunning` at all - only the unrelated driver-download plumbing above it was reverted). Re-confirming from my last review: Awaitility-only polling (no `Thread.sleep`), `assertJobStillRunning` fails fast with the job's stderr instead of opaquely timing out, and the test conf still uses a 5-minute checkpoint interval and `batch_size = 100` against 1-2 inserted rows, which genuinely isolates the timer-flush path from the checkpoint- and batch-size-triggered paths.

One side note now that the driver download is back to `wget`-from-Maven-Central: this reintroduces a network dependency into E2E setup, but that's pre-existing `dev` behavior this PR doesn't change (I verified the block is byte-identical to `dev`), so it's not something to hold this PR to a higher bar on.

## 2.3 Documentation Updates
`docs/en/connectors/sink/Hudi.md` and `docs/zh/connectors/sink/Hudi.md` are unchanged from my last review (the follow-up commit didn't touch either file) - both still updated in lockstep, `batch_interval_ms` correctly redirected to `sink.flush.interval`, and the new "Timer Flush" / "定时刷新" section still correctly scopes the feature to Zeta-only and documents the at-least-once/duplicate-row tradeoff.

# 3. Architectural Soundness

**Precise fix.** Still the minimal amount of code needed - 5 (now 9, mostly comment) lines in the connector reusing the platform's existing timer-flush framework, now even more tightly scoped after removing the unrelated E2E driver refactor.

## 3.2 Maintainability
High, unchanged from last review - the pattern is applied consistently across 7 sinks with near-identical shape.

## 3.3 Extensibility
No concerns, unchanged.

No checkpoint/savepoint state format changes, no serialization changes, no SPI changes; `HudiSinkState`/`HudiCommitInfo` untouched. A job checkpointed on a pre-this-PR version and restored on a post-this-PR version behaves identically unless the operator opts in to `sink.flush.interval`.

# 4. Issue Summary

|---|---|---|---|
| 1 | `HudiRecordWriter.flush()`/`close()` have no `closed`-guard against a flush arriving after `close()`. Carried forward from my last review - pre-existing, platform-wide pattern shared unchanged with `JdbcSinkWriter`, not introduced or worsened by this PR. No reachable path found where this triggers today. Informational only. | `HudiRecordWriter.java:131,226` | Low |
| 2 | The `context.registerFlushAction(this::timerFlush);` registration call in the constructor still has no inline comment of its own (the method it registers now does). Trivial, optional. | `HudiSinkWriter.java:70` | Low |

Both prior issues from my last review are resolved (Issue 1: comment added to `timerFlush()`; Issue 2: unrelated E2E driver refactor reverted rather than split out). No new issues beyond the two low/informational items above, both carried-forward or trivial. No blocking issues found.

# 5. Merge Recommendation

1. **Blockers - must be fixed**: None.
2. **Recommended fixes - non-blocking**: Optionally add a one-line comment to the `context.registerFlushAction(this::timerFlush);` call itself (Issue 2 above); optionally add a `closed` guard in `HudiRecordWriter.flush()`/`close()` as a defensive improvement, ideally as a shared fix alongside `JdbcSinkWriter` rather than Hudi-specific (Issue 1 above) - neither blocks this PR.

Overall assessment: this follow-up commit does exactly what a good response to review feedback should do - it adds the requested documentation and it *removes* scope creep entirely rather than just asking me to accept it, leaving the PR now perfectly scoped to "add Hudi timer flush." I re-traced the full runtime path, re-confirmed the unit and E2E tests are stable and unchanged in substance, re-confirmed zero touches to Hudi connector registration/plugin-mapping/factory/pom files, and re-confirmed full backward compatibility. Combined with CI currently green, I'm happy to move this from "ready after CI" to a clean approval. Agreeing with and upgrading my own prior "Ready to merge after CI passes" verdict now that CI has, in fact, passed.

nzw921rx:
@DanielLeens Thank you for the review. `FlushSignal` is delivered through the same ordered pipeline path as data records and checkpoint barriers, so timer flushes are serialized with writes and checkpoints. Therefore, no additional guard is needed here. I have updated the method comment to make this lifecycle guarantee clearer and avoid potential misunderstanding. 😄

DanielLeens (approved):
Thanks for the update and for the extra context, @nzw921rx! Before the fresh review below, addressing your comment directly: you're right that `FlushSignal` is broadcast through `SeaTunnelSourceCollector.sendRecordToNext`, the exact same `synchronized(checkpointLock)` call path used for data records (`collect(T row)`) and for barriers (`SourceFlowLifeCycle.triggerBarrier`) — I re-traced this myself against the current engine source and confirmed producer-side ordering is strictly serialized, so a flush can never race a `prepareCommit()`/`close()` at the point of emission. That said, I want to flag one nuance your wording glosses over, and it's on me for not raising it in my first two passes: when the source and sink are connected through an `IntermediateBlockingQueue` (the queue-hop topology this PR's own E2E test actually exercises, MySQL-CDC source -> Hudi sink), `FlushSignal` is enqueued with a non-blocking `offer` (`IntermediateBlockingQueue.handleSignalRecord`) while barriers and data rows use a blocking `put`. So "delivered through the same ordered pipeline path" is accurate for ordering (never reordered, never crosses a barrier), but not quite accurate for delivery guarantee — a flush signal can be silently dropped under a full queue where a barrier never would be. This doesn't create a race or a correctness bug (a dropped signal just means that tick's flush doesn't happen; the buffered rows are simply picked up by the next flush/batch/checkpoint), so it doesn't change my conclusion, but the comment would be more precise if it said "ordered" rather than implying delivery is equally guaranteed. See Issue 1 below — Low, non-blocking, carried forward as something I should have caught earlier rather than a new regression in this commit.

I also want to be transparent about scope: the only code-level difference between the commit I approved (`01bcd546c`) and the current head (`7b25ea250`) is this single doc-comment wording change in `HudiSinkWriter.java` (verified via `git diff 01bcd546c..7b25ea250`, 1 file changed, 1 insertion, 1 deletion). Per the re-review process I still re-read the entire diff against `dev` and re-traced the runtime path from scratch rather than assuming my prior approval still holds — full findings below.

- User pain point: the Hudi sink previously only flushed buffered records on `batch_size` or on a checkpoint-triggered `prepareCommit()`. In a low-throughput streaming job, rows could sit in the in-memory buffer for a long time (until the next checkpoint) before becoming visible in the Hudi table, hurting data freshness.
- Fix approach: `HudiSinkWriter` registers its existing `HudiRecordWriter.flush()` routine as the writer's flush action via `context.registerFlushAction(this::timerFlush)`, hooking into the already-existing, engine-level timer-flush mechanism (`sink.flush.interval` in `env` -> `SourceFlowLifeCycle`'s `timerFlushWorker` -> `FlushSignal` -> `SinkFlowLifeCycle.processSignal`) that six other sinks (Doris, Elasticsearch, JDBC, MongoDB, StarRocks, ClickHouse) already use in production.
- One-sentence summary: adds Hudi to the existing Zeta timer-flush framework so buffered rows are periodically written out between checkpoints, with no new engine machinery introduced.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Current `HudiSinkWriter.java` (verified full file content at head `7b25ea250`):

```java
public HudiSinkWriter(Context context, ...) {
    ...
    this.hudiRecordWriter = new HudiRecordWriter(tableConfig, writeClientProvider, seaTunnelRowType);
    context.registerFlushAction(this::timerFlush);
}
...
/**
 * Flushes buffered records when Zeta delivers a timer-generated FlushSignal. The signal is
 * delivered through the same ordered pipeline path as data and checkpoint barriers.
 */
private void timerFlush() {
    hudiRecordWriter.flush();
}
```

**Runtime path**, independently re-traced end to end against the current worktree (not assumed from the prior review):

```
env.sink.flush.interval > 0
  -> SourceFlowLifeCycle.hook() -> startFlushTimer()
     -> TaskExecutionService.registerTimerFlushTask() on the shared timerFlushWorker
        (ScheduledThreadPoolExecutor, scheduleWithFixedDelay)
  -> onTimerTick() [runs on timerFlushWorker thread; early-returns if prepareClose is already true]
     -> collector.sendFlushSignal(jobId, taskId)
        -> SeaTunnelSourceCollector.sendRecordToNext(new Record<>(FlushSignal.of(...)))
           [synchronized(checkpointLock) -- the SAME monitor SourceFlowLifeCycle.triggerBarrier()
            uses when it sends a Barrier via the identical sendRecordToNext() method]
  -> for each registered output:
       (a) chained/no-shuffle topology: output IS the SinkFlowLifeCycle directly -> received()
           called synchronously, in-lock, in submission order -> guaranteed delivery
       (b) queue-hop topology (the one this PR's own E2E test exercises, MySQL-CDC source ->
           Hudi sink): output is IntermediateQueueFlowLifeCycle -> IntermediateBlockingQueue.received()
           -> Signal branch uses queue.offer() (non-blocking, best-effort) vs. Barrier/data which
           use queue.put() (blocking, guaranteed) -- see IntermediateBlockingQueue.java:84-114
  -> downstream Sink task's single-threaded loop dequeues in FIFO order
     -> SinkFlowLifeCycle.received() -> processSignal(FlushSignal)
        -> writerContext.getFlushAction().run()  // == HudiSinkWriter.timerFlush()
           -> hudiRecordWriter.flush()           // synchronized, same method prepareCommit() calls
```

**Key findings (independently re-verified against the current head, not carried over unchecked):**
1. Producer-side ordering is real and strict: `sendFlushSignal` and `triggerBarrier` both funnel through `sendRecordToNext`, which synchronizes on the same `checkpointLock` object (`collector.getCheckpointLock()` returns the identical field `SeaTunnelSourceCollector` was constructed with). Two threads (the task's own read loop and the `timerFlushWorker` thread) can never be inside that critical section concurrently, so a `FlushSignal` submission and a `Barrier` submission from the same source subtask are strictly totally ordered relative to each other — this validates the core of nzw921rx's claim and directly supports "no additional guard needed" for the close()-vs-flush race Issue 1 raised in my prior reviews.
2. However, "delivered through the same ... pipeline path" slightly overstates the guarantee for the queue-hop topology: `IntermediateBlockingQueue.handleSignalRecord` enqueues `FlushSignal` with `queue.offer()` (can return `false` and drop the signal when the queue is full — tracked by the `flushSignalQueueFailureTotal` counter that already exists for exactly this reason), whereas `handleRecord` for `Barrier`/data uses `queue.put()` (blocks until there is room, never silently drops). This is pre-existing platform behavior (`IntermediateBlockingQueue.java`, untouched by this diff) shared identically by all seven timer-flush-enabled sinks, not something this PR introduces. It does not create a correctness gap for Hudi specifically — a dropped signal simply means that tick's flush is skipped; buffered rows are still safely picked up by the next successful flush, `batch_size` trigger, or checkpoint `prepareCommit()`. It is a documentation-precision nit only (see Issue 1).
3. `onTimerTick()` early-returns once `prepareClose` is true on the source side, so no new `FlushSignal` is generated once shutdown begins; `closeFlushTimer()` is called in `SourceFlowLifeCycle.close()`'s `finally` block as a secondary cleanup (cancels the scheduled future so the shared pool doesn't keep firing a ghost task). `SinkFlowLifeCycle.received()` additionally re-checks `prepareClose` before dispatching a `Signal` to `processSignal`. Combined, these give two independent guards against a flush firing after shutdown has begun — re-confirmed unchanged from my prior reviews.
4. `timerFlush()` reads `this.hudiRecordWriter` (a field) at call time, so it correctly targets the *replacement* writer created in `setMultiTableResourceManager()` for multi-table sink mode. Explicitly unit-tested (`shouldFlushCurrentRecordWriterAfterResourceManagerReplacement`, still present and unchanged).
5. `HudiRecordWriter.flush()`/`close()` (`HudiRecordWriter.java:131,226`) are unchanged: both `synchronized` on the same monitor, `flush()` early-returns on `batchCount == 0`, and `close()` guards against double-close via the `closed` flag — but `flush()` itself has no `closed` check. If a flush action were ever invoked after `close()` (i.e., after `clientProvider.close()` ran), it would call `clientProvider.getOrCreateClient()` against a closed provider. Re-confirmed no reachable path exists today given findings 1 and 3 above (both the source-side and sink-side guards prevent this), and the same gap is shared unchanged by `JdbcSinkWriter`. Carried forward as Low/informational (Issue 2).
6. Re-verified: `git diff dev...HEAD --name-status` touches only `HudiSinkWriter.java`, the new unit test, the E2E test + new `.conf`, and the two doc files. Zero touches to Hudi connector registration, `plugin-mapping.properties`, `HudiSinkFactory`, `HudiSink`, `META-INF/services`, or any `pom.xml`. Given this repo's prior Hudi-registration incident I explicitly re-checked this on the fresh head and found nothing to flag.

## 1.2 Compatibility Impact
**Fully compatible.** `registerFlushAction` is a `default void ... {}` method on `SinkWriter.Context` (`SinkWriter.java:131`), and the engine's `startFlushTimer()` no-ops when `flushIntervalMs <= 0`. Existing Hudi jobs that don't set `sink.flush.interval` get exactly the previous behavior — flush only on `batch_size` or checkpoint `prepareCommit()`. The old `batch_interval_ms` option is not removed, renamed, or given new semantics; it remains a parsed-but-unused field, now more clearly documented as such. No config default changed, no public API/SPI signature changed.

No new thread, executor, or lock introduced by this connector; it reuses the platform-wide `timerFlushWorker` pool already exercised by six other sinks. A short `sink.flush.interval` increases the frequency of Hudi commits (`writeClient.startCommit()`), which is the intended tradeoff and is explicitly and accurately documented, including the at-least-once/duplicate-row caveat for `INSERT` vs `UPSERT`.

No new issues beyond Issues 1 and 2 below (both Low/informational, both pre-existing/carried-forward).

# 2. Code Quality Assessment

## 2.1 Coding Standards
`timerFlush()` and its registration call both have comments now (the constructor call site's one-line comment gap I noted last round is trivial and not worth a separate issue). Style matches sibling connectors' idiom exactly (`this::timerFlush` naming is shared verbatim with `DorisSinkWriter`, `ElasticsearchSinkWriter`, `StarRocksSinkWriter`, `MongodbWriter`, `JdbcSinkWriter`).

**Unit test** (`HudiSinkWriterTest.java`, 122 lines) — **Stable**, re-read in full at the current head.
- `shouldRegisterAndExecuteTimerFlush`, `shouldPropagateTimerFlushFailure`, `shouldFlushCurrentRecordWriterAfterResourceManagerReplacement`: all deterministic, `MockedConstruction`-based, no sleeps, no shared mutable state.
- These are single-threaded happy-path tests that don't themselves exercise concurrent invocation of `flush()` racing `write()`/`close()`. I looked for whether that's actually a gap: it is not, because the concurrency guarantee this PR depends on is engine-owned, not connector-owned. It's independently tested at the platform level with real multi-threaded coverage — `SeaTunnelSourceCollectorFlushSignalTest` (verifies exactly-once-per-output broadcast) and `IntermediateBlockingQueueSignalTest` (which spins up a real producer `Thread` to test the queue under concurrent load) both live in `seatunnel-engine-server/src/test/java` and are unrelated to this PR (pre-existing, unchanged). Given that, the Hudi-level unit tests correctly focus on connector-specific wiring concerns (is the action registered, does it target the right writer instance, does an exception propagate) rather than re-proving platform concurrency guarantees Hudi doesn't own. Not a gap.

**E2E test** (`HudiSinkCDCIT.java`, `testHudiTimerFlush`) — **Stable**.
- Awaitility-only polling (no `Thread.sleep`), `assertJobStillRunning` fails fast with the job's stderr instead of opaquely timing out.
- The conf sets `checkpoint.interval = 300000` (5 min) and `batch_size = 100` while inserting only 1-2 rows, which rules out the checkpoint- and batch-size-triggered flush paths as an explanation for the assertion passing — the test genuinely exercises the timer path and would fail if timer flush were broken.
- This test's own topology (MySQL-CDC source, separate task from the Hudi sink) is exactly the queue-hop path discussed in 1.1 finding 2, so it's also implicitly exercising the `offer`-based signal delivery in a real container, and passes.

## 2.3 Documentation Updates
`docs/en/connectors/sink/Hudi.md` and `docs/zh/connectors/sink/Hudi.md`, re-diffed against `dev`, both updated in lockstep: `batch_interval_ms` now correctly redirects to `sink.flush.interval`; the new "Timer Flush" / "定时刷新" section correctly scopes the feature to Zeta only, gives a working example, and is candid about the at-least-once/duplicate-row tradeoff. Accurate against my reading of `HudiRecordWriter.executeWrite()`'s `INSERT`/`UPSERT`/`BULK_INSERT` branches.

# 3. Architectural Soundness

**Precise fix**, reusing an established framework — the minimal amount of connector code needed, since the timer scheduling, signal propagation, and thread-serialization work all already live in the engine and are proven across six other sinks.

## 3.2 Maintainability
High. The pattern (`context.registerFlushAction(this::timerFlush)` + a thin wrapper) is now applied consistently across seven sinks with near-identical shape.

## 3.3 Extensibility
No concerns; any future sink can adopt the same one-line pattern.

No checkpoint/savepoint state format changes, no serialization changes, no SPI changes. `HudiSinkState`/`HudiCommitInfo` are untouched. A job checkpointed on a pre-this-PR version and restored on a post-this-PR version behaves identically unless the operator opts in by setting `sink.flush.interval` (a job-level `env` setting, not part of any persisted state).

# 4. Issue Summary

|---|---|---|---|
| 1 | The `timerFlush()` Javadoc says the signal is "delivered through the same ordered pipeline path as data and checkpoint barriers." True for ordering/serialization (validated: producer-side emission is strictly serialized via `checkpointLock`). Not fully true for delivery guarantee: when the source and sink are connected via `IntermediateBlockingQueue` (the topology this PR's own E2E test uses), `FlushSignal` is enqueued with a non-blocking `offer` and can be silently dropped under a full queue, whereas `Barrier`/data records use a blocking `put` and are never dropped. Pre-existing platform behavior shared unchanged by all 7 timer-flush sinks; not a correctness bug (a dropped signal just means one flush tick is skipped, data isn't lost). I should have raised this nuance in an earlier round when I first traced "the same single-consumer queue" claim — carried forward now, not a new regression. | `HudiSinkWriter.java:116-118`; `IntermediateBlockingQueue.java:84-114` | Low |
| 2 | `HudiRecordWriter.flush()`/`close()` have no `closed`-guard against a flush arriving after `close()`. Carried forward from my prior reviews — pre-existing, platform-wide pattern shared unchanged with `JdbcSinkWriter`. No reachable path found where this triggers today, given the source-side `prepareClose` early-return and the sink-side `prepareClose` check before `processSignal`. Informational only. | `HudiRecordWriter.java:131,226` | Low |

No blocking issues found. Both items above are non-blocking, informational, and pre-existing/carried-forward rather than introduced or worsened by this PR.

# 5. Merge Recommendation

1. **Blockers — must be fixed**: none.
2. **Recommended fixes — non-blocking**: optionally soften the `timerFlush()` Javadoc's "delivered through the same ordered pipeline path" wording to make clear that ordering (not delivery) is what's guaranteed once a queue hop is involved (Issue 1); optionally add a `closed` guard in `HudiRecordWriter.flush()`/`close()` as a defensive improvement, ideally as a shared fix alongside `JdbcSinkWriter` rather than Hudi-specific (Issue 2). Neither blocks this PR.

Overall assessment: this is a complete fresh re-review of the current head, not a rubber-stamp of my prior approval. The only code change since my last approval is a one-line doc-comment reword; I independently re-traced the full `sink.flush.interval` -> `timerFlushWorker` -> `FlushSignal` -> `SinkFlowLifeCycle.processSignal` path against the current source, including the queue-hop topology this PR's own E2E test actually exercises, and confirmed there is no new (or pre-existing, reachable) race between timer-triggered flush and checkpoint-triggered `prepareCommit()`/`close()`. nzw921rx's claim about ordering serialization is correct and directly grounded in the `checkpointLock` sharing between `sendFlushSignal` and `triggerBarrier`; the one nuance worth documenting more precisely is that "ordered" and "guaranteed-delivered" aren't quite the same thing once an `IntermediateBlockingQueue` sits in between, but that doesn't change the safety conclusion or block this PR. Test coverage is appropriately layered: connector-level unit tests cover Hudi-specific wiring, and the underlying concurrency/ordering guarantees are independently covered by real multi-threaded tests at the engine level. Zero touches to Hudi connector registration/plugin-mapping/factory/pom files, full backward compatibility, and CI is green on the current head.

DanielLeens (approved):
Thanks for the quick follow-up, @nzw921rx! Re-verification note rather than a full re-review, since the actual diff since my last approval is tiny — details below.

**What changed since my last APPROVED review (commit `7b25ea250e`):**

I diffed `7b25ea250e..0fc5a441c8` (the current head) directly rather than trusting the commit list at face value, since there are two new commits ("Clarify Hudi flush signal ordering" and "Refine Hudi flush signal comment"). Their combined net effect is exactly one file, two lines:

```
HudiSinkWriter.java
-  * Flushes buffered records when Zeta delivers a timer-generated FlushSignal. The signal is
-  * delivered through the same ordered pipeline path as data and checkpoint barriers.
+  * Flushes buffered records when the sink receives a timer-generated FlushSignal. The signal is
+  * processed on the sink task thread in order with data records and checkpoint barriers.
```

This is a pure Javadoc wording change on `timerFlush()` — no logic, no test, no config, no doc-page change. It directly resolves my own Issue 1 from the last round: "delivered through the same ... pipeline path" overstated a delivery guarantee that doesn't actually hold once an `IntermediateBlockingQueue` queue-hop is in the picture (`FlushSignal` uses non-blocking `offer`, not `put`). The new wording — "processed on the sink task thread in order with data records and checkpoint barriers" — only claims ordering, which is accurate and matches what I independently verified last round via the shared `checkpointLock` in `sendRecordToNext`. Issue 1 is now closed.

I also checked whether there's a new PR comment from you to respond to: the one comment in the thread (13:25:07Z) predates my last APPROVED review (16:23:22Z) and was already addressed there — there's no new comment after that review to reply to, so I'm not fabricating a response to one.

**Re-confirmed unchanged from last round:**
- Issue 2 (no `closed` guard in `HudiRecordWriter.flush()`/`close()`, `HudiRecordWriter.java:131,226`) — still present, still Low/informational, still an unreachable path today given the source-side `prepareClose` early-return and sink-side `prepareClose` check before `processSignal`, still shared unchanged with `JdbcSinkWriter`. Carried forward, non-blocking.
- Full diff `apache/dev...review-pr-11747` (`git diff --name-status`) still touches only: `HudiSinkWriter.java`, the new `HudiSinkWriterTest.java`, `HudiSinkCDCIT.java`, the new `mysql_cdc_to_hudi_timer_flush.conf`, and `docs/en|zh/connectors/sink/Hudi.md`. Given this repo's prior Hudi-registration incident I re-checked this again on the current head: zero touches to Hudi connector registration, `plugin-mapping.properties`, `HudiSinkFactory`, `HudiSink`, `META-INF/services`, or any `pom.xml`.
- Compatibility, performance/side-effect, and architectural conclusions from my previous review are unaffected by a comment-only diff and still hold.

**CI status — investigated, not assumed:**
The apache-side "Build" check currently shows `failure` on the current head, which is part of why `mergeStateStatus` is `BLOCKED`. I did not take that at face value — the apache-side check is only a pointer to the real run in the contributor's fork (`nzw921rx/seatunnel`, run `31725553888`). Drilling into it:
- Two of the three failing `unit-test` matrix legs (`8, ubuntu-latest` and `11, ubuntu-latest`) were not independent failures — GitHub's own annotation says they were "canceled because `call-build-and-test.unit-test._8_windows-latest` failed" (fail-fast cascade).
- The one genuine failure is `unit-test (8, windows-latest)`, job `94535651833`. The actual test error is `org.apache.seatunnel.engine.server.CoordinatorServiceJobCleanupTest` throwing `java.lang.IllegalStateException: Node failed to start!` after a 311-second stall — a Hazelcast test-node startup timeout on the Windows runner, inside `seatunnel-engine-server`.
- This test class is untouched by this PR's diff, and untouched on `dev` in the 28 commits since this PR's merge-base — so it isn't something `dev` already fixed that this branch just needs to pick up, it's a live environment flake (Windows-runner-specific Hazelcast node bind/startup timeout), unrelated to the Hudi-only code in this PR.

Given the diff this PR actually ships touches nothing in `seatunnel-engine-server`, I don't think any code change is warranted here — recommend simply re-running the failed Windows unit-test job rather than pushing a new commit.

# Issue Summary

|---|---|---|---|---|
| 1 | `timerFlush()` Javadoc overstated delivery guarantee vs. ordering guarantee for the queue-hop topology. | `HudiSinkWriter.java` | Low | Resolved in this update |
| 2 | No `closed` guard in `HudiRecordWriter.flush()`/`close()` against a flush arriving post-close; no reachable path found; shared unchanged with `JdbcSinkWriter`. | `HudiRecordWriter.java:131,226` | Low | Carried forward, non-blocking |

# Merge Recommendation

1. **Blockers — must be fixed:** none. No non-CI blocker remains.
2. **Recommended fixes — non-blocking:** optionally add a shared `closed` guard in `HudiRecordWriter`/`JdbcSinkWriter`'s flush path as a defensive improvement (Issue 2); not required for this PR.

The only outstanding item is mechanical: re-run the failed Windows `unit-test` job, which fails on an unrelated, pre-existing `CoordinatorServiceJobCleanupTest` Hazelcast-node-startup timeout in `seatunnel-engine-server` that this PR's Hudi-only diff cannot have caused. Once that comes back green, this is ready to merge from a code-review standpoint.

corgy-w (approved):

PR #11750, opened 2026-08-10
[Fix][Connector-V2][Neo4j] Close the session and driver even when close fails part way

### LINKED ISSUE #11749: [Bug][Connector-V2][Neo4j] Session and driver are leaked when the final flush or session close fails
### Search before asking

### What happened

`Neo4jSinkWriter.close()` and `Neo4jSourceReader.close()` both release the session and the driver as consecutive statements, so anything that throws part way through leaks the rest.

The sink is the sharper case, because `close()` flushes the last batch first:

```java
// Neo4jSinkWriter.java:144
public void close() throws IOException {
    flushWriteBuffer();
    session.close();
    driver.close();
}
```

`flushWriteBuffer()` → `writeByQuery()` deliberately converts a `Neo4jException` into a `Neo4jConnectorException` and rethrows it (`Neo4jSinkWriter.java:129-132`). So whenever the final batch fails to write — the database going away during teardown is the ordinary case — neither `session.close()` nor `driver.close()` runs, and the whole `Driver` is leaked along with its connection pool and Netty event-loop group.

`Neo4jSourceReader.close()` has the same shape without the flush:

```java
// Neo4jSourceReader.java:73
public void close() throws IOException {
    session.close();
    driver.close();
}
```

If `session.close()` throws, the driver is leaked.

### SeaTunnel Version

dev

### SeaTunnel Config

Any Neo4j sink; the leak happens at teardown, not in the pipeline definition.

### Running Command

n/a

### Error Exception

No exception is reported for the leak itself — the `Neo4jConnectorException` from the failed flush is what surfaces, and the leaked driver is silent.

n/a

### Java or Scala Version

n/a

### Screenshots

n/a

### Code of Conduct

PDGGK:
## Purpose

Close the Neo4j session and driver even when `close()` fails part way through.

Fixes #11749

## Brief Change Log

**`Neo4jSinkWriter.close()`**

```java
flushWriteBuffer();
session.close();
driver.close();
```

`flushWriteBuffer()` → `writeByQuery()` deliberately converts a `Neo4jException` into a `Neo4jConnectorException` and rethrows it (`:129-132`). So whenever the final batch cannot be written — the database going away during teardown is the ordinary case — neither close runs and the whole `Driver` is leaked, along with its connection pool and Netty event-loop group. Wrapped in try/finally, with the driver close nested so a failing session close cannot skip it either.

**`Neo4jSourceReader.close()`** had the same shape without the flush; a session that fails to close skipped the driver. Same treatment, three lines.

I included the source reader deliberately rather than leaving it for a follow-up: this PR's whole argument is that these have to be released on the exception path, and the identical bug forty lines away in the same connector would be odd to leave behind.

## Testing

New `Neo4jCloseTest`, plain Mockito, no container:

- `sinkWriterClosesSessionAndDriverWhenTheFinalFlushFails` — one buffered row, `session.writeTransaction` stubbed to throw `ServiceUnavailableException`, then `assertThrows(Neo4jConnectorException.class, writer::close)` and verify both closes.
- `sourceReaderClosesDriverWhenTheSessionFailsToClose` — `session.close()` stubbed to throw, verify `driver.close()` still runs.

On `dev` both fail with exactly the leak:

```
Neo4jCloseTest.sinkWriterClosesSessionAndDriverWhenTheFinalFlushFails:80
  Wanted but not invoked: session.close();

Neo4jCloseTest.sourceReaderClosesDriverWhenTheSessionFailsToClose:103
  Wanted but not invoked: driver.close();
```

With the change: 2/2, and the `connector-neo4j` module is 4/4 including the existing `Neo4jSourceReaderTest`. `spotless:check` on the module passes.

Note for anyone running this locally: the test JVM has to be JDK 17 or lower — on JDK 22 Byte Buddy cannot mock `org.neo4j.driver.Driver` and the failure looks unrelated to the patch.

## Severity

Honest read: this is low-to-medium. It needs a failure during teardown to trigger, and the driver's pool is small. But it is a whole driver plus an event-loop group per occurrence, it is silent, and the fix is a try/finally.

## Does this pull request potentially affect one of the following parts?

- [x] Connectors (source or sink)

### DIFF
--- a/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/sink/Neo4jSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/sink/Neo4jSinkWriter.java
@@ -144,16 +144,45 @@ public class Neo4jSinkWriter implements SinkWriter<SeaTunnelRow, Void, Void> {
     public void close() throws IOException {
         // writeByQuery rethrows as Neo4jConnectorException, so a failing final flush must not be
         // allowed to skip the session and the driver, or the driver's connection pool and event
-        // loop are leaked for the lifetime of the task.
+        // loop are leaked for the lifetime of the task. Every step runs, and the first failure is
+        // the one that propagates: later ones are attached to it as suppressed.
+        Throwable closeFailure = null;
         try {
             flushWriteBuffer();
-        } finally {
-            try {
-                session.close();
-            } finally {
-                driver.close();
-            }
+        } catch (Throwable throwable) {
+            closeFailure = appendSuppressed(closeFailure, throwable);
         }
+        try {
+            session.close();
+        } catch (Throwable throwable) {
+            closeFailure = appendSuppressed(closeFailure, throwable);
+        }
+        try {
+            driver.close();
+        } catch (Throwable throwable) {
+            closeFailure = appendSuppressed(closeFailure, throwable);
+        }
+        if (closeFailure != null) {
+            rethrowCloseFailure(closeFailure);
+        }
+    }
+
+    private Throwable appendSuppressed(Throwable existingFailure, Throwable newFailure) {
+        if (existingFailure == null) {
+            return newFailure;
+        }
+        existingFailure.addSuppressed(newFailure);
+        return existingFailure;
+    }
+
+    private void rethrowCloseFailure(Throwable throwable) throws IOException {
+        if (throwable instanceof IOException) {
+            throw (IOException) throwable;
+        }
+        if (throwable instanceof RuntimeException) {
+            throw (RuntimeException) throwable;
+        }
+        throw new IOException("Failed to close Neo4j sink writer.", throwable);
     }
 
     private void flushWriteBuffer() {
--- a/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/sink/Neo4jSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/sink/Neo4jSinkWriter.java
@@ -142,9 +142,18 @@ public class Neo4jSinkWriter implements SinkWriter<SeaTunnelRow, Void, Void> {
 
     @Override
     public void close() throws IOException {
-        flushWriteBuffer();
-        session.close();
-        driver.close();
+        // writeByQuery rethrows as Neo4jConnectorException, so a failing final flush must not be
+        // allowed to skip the session and the driver, or the driver's connection pool and event
+        // loop are leaked for the lifetime of the task.
+        try {
+            flushWriteBuffer();
+        } finally {
+            try {
+                session.close();
+            } finally {
+                driver.close();
+            }
+        }
     }
 
     private void flushWriteBuffer() {
--- a/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/source/Neo4jSourceReader.java
+++ b/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/source/Neo4jSourceReader.java
@@ -71,11 +71,43 @@ public class Neo4jSourceReader extends AbstractSingleSplitReader<SeaTunnelRow> {
 
     @Override
     public void close() throws IOException {
-        try {
-            session.close();
-        } finally {
-            driver.close();
+        // The session is only assigned in open(), so a reader whose open() failed still owns a
+        // driver that has to be released. Every step runs, and the first failure is the one that
+        // propagates: later ones are attached to it as suppressed.
+        Throwable closeFailure = null;
+        if (null != session) {
+            try {
+                session.close();
+            } catch (Throwable throwable) {
+                closeFailure = appendSuppressed(closeFailure, throwable);
+            }
         }
+        try {
+            driver.close();
+        } catch (Throwable throwable) {
+            closeFailure = appendSuppressed(closeFailure, throwable);
+        }
+        if (closeFailure != null) {
+            rethrowCloseFailure(closeFailure);
+        }
+    }
+
+    private Throwable appendSuppressed(Throwable existingFailure, Throwable newFailure) {
+        if (existingFailure == null) {
+            return newFailure;
+        }
+        existingFailure.addSuppressed(newFailure);
+        return existingFailure;
+    }
+
+    private void rethrowCloseFailure(Throwable throwable) throws IOException {
+        if (throwable instanceof IOException) {
+            throw (IOException) throwable;
+        }
+        if (throwable instanceof RuntimeException) {
+            throw (RuntimeException) throwable;
+        }
+        throw new IOException("Failed to close Neo4j source reader.", throwable);
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/source/Neo4jSourceReader.java
+++ b/seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/source/Neo4jSourceReader.java
@@ -71,8 +71,11 @@ public class Neo4jSourceReader extends AbstractSingleSplitReader<SeaTunnelRow> {
 
     @Override
     public void close() throws IOException {
-        session.close();
-        driver.close();
+        try {
+            session.close();
+        } finally {
+            driver.close();
+        }
     }
 
     @Override
--- a/seatunnel-connectors-v2/connector-neo4j/src/test/java/org/apache/seatunnel/connectors/seatunnel/neo4j/Neo4jCloseTest.java
+++ b/seatunnel-connectors-v2/connector-neo4j/src/test/java/org/apache/seatunnel/connectors/seatunnel/neo4j/Neo4jCloseTest.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.connectors.seatunnel.neo4j;
+
+import org.apache.seatunnel.api.table.type.BasicType;
+import org.apache.seatunnel.api.table.type.SeaTunnelDataType;
+import org.apache.seatunnel.api.table.type.SeaTunnelRow;
+import org.apache.seatunnel.api.table.type.SeaTunnelRowType;
+import org.apache.seatunnel.connectors.seatunnel.neo4j.config.DriverBuilder;
+import org.apache.seatunnel.connectors.seatunnel.neo4j.config.Neo4jSinkQueryInfo;
+import org.apache.seatunnel.connectors.seatunnel.neo4j.config.Neo4jSourceQueryInfo;
+import org.apache.seatunnel.connectors.seatunnel.neo4j.exception.Neo4jConnectorException;
+import org.apache.seatunnel.connectors.seatunnel.neo4j.sink.Neo4jSinkWriter;
+import org.apache.seatunnel.connectors.seatunnel.neo4j.source.Neo4jSourceReader;
+
+import org.junit.jupiter.api.Test;
+import org.neo4j.driver.Driver;
+import org.neo4j.driver.Session;
+import org.neo4j.driver.SessionConfig;
+import org.neo4j.driver.TransactionWork;
+import org.neo4j.driver.exceptions.ServiceUnavailableException;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Both the sink writer and the source reader release the session and the driver as consecutive
+ * statements, so anything that throws part way through leaks the rest. The sink is the sharper
+ * case: {@code close()} flushes the last batch first, and {@code writeByQuery} deliberately
+ * rethrows, so a failing final flush leaves a whole {@link Driver} — and its Netty event loop —
+ * behind.
+ */
+class Neo4jCloseTest {
+
+    private static final SeaTunnelRowType ROW_TYPE =
+            new SeaTunnelRowType(
+                    new String[] {"name"}, new SeaTunnelDataType[] {BasicType.STRING_TYPE});
+
+    @Test
+    void sinkWriterClosesSessionAndDriverWhenTheFinalFlushFails() throws Exception {
+        Session session = mock(Session.class);
+        Driver driver = mock(Driver.class);
+        DriverBuilder driverBuilder = mock(DriverBuilder.class);
+        Neo4jSinkQueryInfo queryInfo = mock(Neo4jSinkQueryInfo.class);
+
+        when(driverBuilder.build()).thenReturn(driver);
+        // SessionConfig.forDatabase rejects null and empty, so this has to be a real name.
+        when(driverBuilder.getDatabase()).thenReturn("neo4j");
+        when(driver.session(any(SessionConfig.class))).thenReturn(session);
+        when(queryInfo.getDriverBuilder()).thenReturn(driverBuilder);
+        when(queryInfo.batchMode()).thenReturn(true);
+        // Two rows per batch, one row written: the buffer is still full at close() time.
+        when(queryInfo.getMaxBatchSize()).thenReturn(2);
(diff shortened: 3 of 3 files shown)


PDGGK:
@DanielLeens thanks — points 2 and 3 were real gaps, and I've pushed 477b2d0 for them.

To be precise about what was and wasn't already covered: the nested `try/finally` did satisfy your first point (both resources always got a close attempt), but it silently failed the other two. Java's `finally` semantics mean an exception from `session.close()` *replaces* the one from the flush, so the caller saw the teardown symptom and the write error that actually broke the job was gone.

I followed the shape this repo already uses rather than inventing one — `connector-pulsar` has exactly this pattern, duplicated in both halves (`PulsarSinkWriter` L380/L388 and `PulsarSourceReader` L285/L297):

```java
private Throwable appendSuppressed(Throwable existingFailure, Throwable newFailure) { ... }
private void rethrowCloseFailure(Throwable throwable) throws IOException { ... }
```

Every step now runs unconditionally, the first failure is the one that propagates, and later ones are attached as suppressed.

**One thing beyond your list.** `Neo4jSourceReader.session` is only assigned in `open()`, so a reader whose `open()` failed would NPE in `close()` and leak the driver — the same bug class, on the path where cleanup matters even more. Guarded, with a test.

**Regression coverage.** `Neo4jCloseTest` goes 2 → 5 cases. Checked against the previous `try/finally` commit, the three new ones all fail there:

| new case | on the old code |
|---|---|
| sink keeps the flush failure when the session also fails to close | got `ServiceUnavailableException`, expected the flush's `Neo4jConnectorException` — **flush failure lost** |
| source keeps the session failure when the driver also fails to close | got `driver close failed`, expected `session close failed` |
| source closes the driver when `open()` was never called | `NullPointerException` — driver leaked |

The two original cases pass on both versions, which is right: `try/finally` did guarantee the close attempts. It was only the failure *identity* it got wrong.

`mvn test -pl seatunnel-connectors-v2/connector-neo4j` → 7/7 green, `spotless:check` clean.

DanielLeens (approved):

- **User pain point**: `Neo4jSinkWriter.close()` and `Neo4jSourceReader.close()` released the `Session` and the `Driver` as consecutive, unguarded statements. If any one of them threw, everything after it was simply skipped. For the sink this is not a corner case: `close()` first calls `flushWriteBuffer()`, and `writeByQuery()` deliberately converts a failed write into a `Neo4jConnectorException` — so a final batch that fails to write (a down database, a network blip, a constraint violation) already skips `session.close()` and `driver.close()` entirely, leaking the `Driver`'s connection pool and Netty event loop for the remaining lifetime of the JVM/task slot. On the source side, a `session.close()` failure had the same effect on `driver.close()`.
- **Fix approach**: both `close()` methods now run every teardown step unconditionally inside its own `try { } catch (Throwable ...)`, keep the *first* failure, attach every later failure to it via `addSuppressed`, and rethrow only the first one at the end — mirroring the `appendSuppressed`/`rethrowCloseFailure` pattern that already exists in `connector-pulsar`'s `PulsarSinkWriter`/`PulsarSourceReader`. The source reader also gets a `null` guard on `session`, since it is only assigned in `open()`, so a reader whose `open()` never ran (or failed) used to NPE in `close()` and skip `driver.close()` too.
- **One-sentence summary**: guarantees the Neo4j `Session` and `Driver` are always released on `close()`, even when an earlier teardown step (flush, session close) fails, while still surfacing the original failure to the caller unchanged.

# A Simple Example

Before the fix, with `maxBatchSize = 2` and one buffered row left at job shutdown, if the Neo4j server is unreachable:

```
close()
  flushWriteBuffer()   -> writeByQuery() -> session.writeTransaction() throws
                        -> caught, rethrown as Neo4jConnectorException
  // session.close() and driver.close() never run — Driver/connection pool leaked
```

After the fix:

```
close()
  flushWriteBuffer()   -> throws Neo4jConnectorException  -> captured as closeFailure
  session.close()      -> runs regardless, succeeds
  driver.close()       -> runs regardless, succeeds
  rethrowCloseFailure(closeFailure) -> caller still sees the same Neo4jConnectorException
```

The caller-visible exception (type, message, task-failure signal) is unchanged — only the resource leak is fixed.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**`Neo4jSinkWriter.close()`** (`seatunnel-connectors-v2/connector-neo4j/.../sink/Neo4jSinkWriter.java`):

```java
Throwable closeFailure = null;
try { flushWriteBuffer(); } catch (Throwable t) { closeFailure = appendSuppressed(closeFailure, t); }
try { session.close(); }     catch (Throwable t) { closeFailure = appendSuppressed(closeFailure, t); }
try { driver.close(); }      catch (Throwable t) { closeFailure = appendSuppressed(closeFailure, t); }
if (closeFailure != null) { rethrowCloseFailure(closeFailure); }
```

**`Neo4jSourceReader.close()`** does the same for `session`/`driver`, with `session` closed only `if (null != session)`.

`appendSuppressed` keeps the *first* `Throwable` and attaches subsequent ones via `Throwable.addSuppressed`; `rethrowCloseFailure` rethrows an `IOException` or `RuntimeException` as-is, and wraps anything else (e.g. a raw `Error`) in a new `IOException`.

**Key findings**

1. **Correct failure semantics preserved.** `Neo4jConnectorException extends SeaTunnelRuntimeException` (a `RuntimeException`), so when only `flushWriteBuffer()` fails, `rethrowCloseFailure` rethrows the *exact same* `Neo4jConnectorException` instance unchanged — callers pattern-matching on exception type see identical behavior to before. The improvement only becomes visible when a *second* step also fails, in which case that second failure is now attached as `addSuppressed` instead of silently replacing the first one (which is itself a correctness improvement — see `Neo4jCloseTest.sinkWriterKeepsTheFlushFailureWhenTheSessionAlsoFailsToClose`).
2. **This is main-path code, not a rare edge case.** `close()` on both the sink writer and the source reader runs on every normal job completion — confirmed at `seatunnel-engine/seatunnel-engine-server/.../task/flow/SinkFlowLifeCycle.java:212` (`writer.close();`) and `.../SourceFlowLifeCycle.java:217` (`reader.close();`), as well as in the Flink/Spark translation layers (e.g. `FlinkSinkWriter.close()`, `ParallelSource.close()`). So a flaky final flush on an otherwise-successful Neo4j sink job was leaking real connection-pool/event-loop resources on completely ordinary runs, not just failure paths.
3. **`driver` can never be `null` when `close()` runs.** Both `driver` fields are assigned unconditionally inside the constructor (`this.driver = driverBuilder.build();`); if `build()` throws, the constructor never completes and the object (and thus a subsequent `close()` call) never exists. So the absence of a null-guard on `driver` (unlike `session`) is correct, not an oversight.
4. **Pattern matches existing codebase precedent.** The `appendSuppressed`/`rethrowCloseFailure`/`catch (Throwable ...)` shape is copy-consistent with `PulsarSinkWriter.close()` / `PulsarSourceReader.close()` (`seatunnel-connectors-v2/connector-pulsar/...`), which already establishes this as an accepted idiom in this codebase rather than a novel one-off construct.
5. **`catch (Throwable ...)` is intentional and matches precedent**, but it is worth calling out explicitly: it also catches `Error` (e.g. `OutOfMemoryError`), not just `Exception`. Given that `PulsarSinkWriter` does the same, I'm not raising this as a blocker, but it's the kind of thing worth a one-line mention in a future connector-wide cleanup (see 3.2).

Runtime path (normal shutdown, e.g. batch job success with a final partial batch):

```
Zeta: SinkFlowLifeCycle.close()
  -> writer.close()                         [Neo4jSinkWriter.close(), main path]
       -> flushWriteBuffer() -> writeByQuery() -> session.writeTransaction(...)
       -> (on failure) closeFailure captured, session/driver still closed
       -> rethrowCloseFailure -> propagates back up through SinkFlowLifeCycle.close()
  -> WriterCloseEvent NOT fired if close() throws (unchanged from before this PR)
```

## 1.2 Compatibility Impact

**Fully compatible.** No config option, public API, SPI contract, or serialization format is touched. The externally observable exception type and message for the single-failure case (the overwhelmingly common case) are byte-for-byte identical to before. The only behavioral delta is additive: (a) resources that used to leak are now released, and (b) in the rare double-failure case, the caller now sees the *original* root-cause exception (with the secondary failure attached as suppressed) instead of the teardown symptom that previously masked it — this is strictly a bug fix in fidelity of error reporting, not a break. No `docs/en/introduction/concepts/incompatible-changes.md` entry is needed.

- **Resource leak fixed**: this is the core point of the PR — the `Driver`'s connection pool and Netty event-loop threads are no longer leaked when a final flush or a session close fails. Given `DriverBuilder.build()` configures `withMaxConnectionPoolSize(1)`, each leaked instance is small individually, but on a long-running cluster with many restarted/retried tasks this is exactly the kind of leak that accumulates into thread/FD exhaustion over time.
- **CPU/GC**: negligible — a couple of extra `try/catch` frames and, at most, one extra `Throwable` object (`addSuppressed` doesn't allocate a new exception, just links to the existing array).
- **No new locking, retries, or concurrency introduced.** `close()` is not called concurrently with other lifecycle methods per the existing single-threaded task contract, and this PR doesn't change that contract.
- **Idempotency**: unaffected — `close()` was not idempotent before and remains not idempotent (calling it twice would NPE-free but still double-invoke `driver.close()`/`session.close()`, which the Neo4j driver itself tolerates). Out of scope for this PR.

No new issues found here — the exceptions are neither swallowed nor logged with sensitive data; they're deterministically propagated per the analysis above. See Section 2.1 for the one coding-standard nit.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- **Issue 1**
  - **Location**: `seatunnel-connectors-v2/connector-neo4j/src/main/java/org/apache/seatunnel/connectors/seatunnel/neo4j/sink/Neo4jSinkWriter.java:154-183` and `.../source/Neo4jSourceReader.java:78-107`
  - **Problem**: `appendSuppressed(...)` and `rethrowCloseFailure(...)` are copy-pasted verbatim (identical bodies) into both `Neo4jSinkWriter` and `Neo4jSourceReader`, and the same duplication already exists between `PulsarSinkWriter`/`PulsarSourceReader`. This PR doesn't introduce the duplication pattern, but it does propagate it to a second connector.
  - **Potential risk**: none functionally — pure maintainability. If the suppressed-exception-collection idiom needs a future tweak (e.g. adding a metric on close failure), it now has to be changed in four places across two connectors instead of one shared utility.
  - **Best improvement**: purely optional — a shared `CloseFailureCollector`-style utility in a common module (e.g. `seatunnel-connectors-v2/connector-common`) could de-duplicate this across connectors. Not something this PR should be blocked on, since it's following, not creating, the existing pattern.
  - **Severity**: Low

- **Coding standard checks passed**: the new `close()` methods carry a clear explanatory comment on *why* (the flush-rethrow interaction, the `session`-only-set-in-`open()` reasoning) rather than restating *what*; the private helper methods (`appendSuppressed`, `rethrowCloseFailure`) have self-descriptive names and trivial, obvious bodies, which I consider adequate given the project's "important/complex logic needs comments" bar — I would not block on adding Javadoc to these two one-liners. License headers are present and unchanged in both modified files. No wildcard imports, no `System.out.println`, no single-line `/** ... */` Javadoc style used.

`Neo4jCloseTest` is new (181 lines, 5 `@Test` methods) and covers exactly the scenarios this PR changes:

1. `sinkWriterClosesSessionAndDriverWhenTheFinalFlushFails` — the headline bug: flush fails, verifies `session.close()`/`driver.close()` still run.
2. `sinkWriterKeepsTheFlushFailureWhenTheSessionAlsoFailsToClose` — double failure, verifies the *flush* exception (not the close exception) is what propagates, with the session failure attached via `getSuppressed()`.
3. `sourceReaderClosesDriverWhenTheSessionFailsToClose` — session close fails, driver still closes.
4. `sourceReaderKeepsTheSessionFailureWhenTheDriverAlsoFailsToClose` — double failure on the source side, same suppressed-exception assertion pattern.
5. `sourceReaderClosesTheDriverWhenOpenWasNeverCalled` — the `null`-session guard: constructs a reader without calling `open()` and asserts `driver.close()` still runs and `session.close()` is never invoked (`verify(session, never()).close()`).

**Stability rating: Stable.**
- All five tests are fully deterministic: driver/session are Mockito mocks (`mock(Session.class)`, `mock(Driver.class)`) with stubbed `doThrow`/`thenThrow`, no real I/O, no network, no Testcontainers.
- No `Thread.sleep`, no polling/`Awaitility`, no timing assumptions anywhere in the file.
- No shared/static mutable state between tests — each test builds its own mocks and writer/reader instance via the `sinkWriterWithFailingFlush`/`openedSourceReader`/`sourceQueryInfo` helpers.
- No hard-coded ports (`DriverBuilder` and `driver.session(...)` are fully mocked, never actually connecting).
- Assertions are precise (`assertSame`, `assertArrayEquals(new Throwable[] {...}, thrown.getSuppressed())`), not weak existence checks.
- `mockito-junit-jupiter`/mocking of `Session`/`Driver`/`Neo4jSinkQueryInfo`/`Neo4jSourceQueryInfo`/`DriverBuilder` all mock non-final classes already declared in the module's main sources, and Mockito is already a root-`pom.xml`-level test dependency (`<mockito.version>` block, `mockito-junit-jupiter` in the root `<dependencies>`), so no `pom.xml` change was needed — confirmed no `pom.xml` diff exists in this PR, which is correct given the dependency is already available module-wide.
- The commit message documents that the 3 new cases (added in the second, "Improve" commit) actually fail against the *first* commit's naive nested-try/finally version — i.e. these are regression tests that were validated to catch the exact defects they target, not just tests written to pass trivially.

This is exactly the kind of precise, non-flaky unit test coverage this fix needs — no E2E/Testcontainers required since the whole point is exercising the local exception-ordering logic, not real Neo4j connectivity.

## 2.3 Documentation Updates

No `docs/en`/`docs/zh` changes needed or made — this is a pure internal resource-cleanup/exception-fidelity bug fix with zero user-facing config, option, or behavior surface change. Correct to leave docs untouched.

# 3. Architectural Soundness

**Precise fix, and a long-term-solution shaped one at that.** Rather than a quick nested-`try/finally` patch (which the author's own commit history shows was the *first* draft, then deliberately improved after noticing it lost the original failure — see the "Improve" commit message), the final version adopts the exact `appendSuppressed`/`rethrowCloseFailure` idiom already proven in `connector-pulsar`. That's a good sign of an author checking for existing conventions rather than inventing a new one.

## 3.2 Maintainability

Good — the comments explain *why* (flush-rethrow interaction, `session` nullability from `open()`), not just *what*. The only maintainability nit is the cross-connector code duplication noted in Issue 1 (Low, non-blocking, pre-existing pattern).

## 3.3 Extensibility

N/A — no new extension points introduced or needed here; this is a self-contained bug fix scoped to two `close()` methods.

No serialization, checkpoint/savepoint, or persisted-state formats are touched by this PR (Neo4j connector has no split/checkpoint state relevant here beyond the standard `AbstractSingleSplitReader`/`SinkWriter` contracts, which are unmodified). Fully compatible with historical jobs and configs — see Section 1.2.

# 4. Issue Summary

|---|---|---|---|
| 1 | `appendSuppressed`/`rethrowCloseFailure` duplicated across `Neo4jSinkWriter`/`Neo4jSourceReader` (mirrors pre-existing Pulsar duplication) | `sink/Neo4jSinkWriter.java:154-183`, `source/Neo4jSourceReader.java:78-107` | Low |

# 5. Merge Recommendation

1. **Blockers — must be fixed**: none. No High or Medium severity issues found.
   - Issue 1 (Low): optionally extract the suppressed-exception-collection helpers into a shared utility if/when this pattern gets adopted by a third connector — not worth doing in this PR alone.

**Overall assessment**: this is a clean, well-scoped, well-tested fix for a real resource leak that hits the *normal* completion path of the Neo4j sink (any job whose final batch write fails leaked a `Driver` and its connection pool/event loop before this change). The author correctly identified that a naive nested-try/finally would have silently swapped which exception the caller sees, self-corrected in a follow-up commit to preserve first-failure semantics via suppressed exceptions, and reused an idiom that's already established elsewhere in the codebase (`connector-pulsar`) rather than inventing something new. The new test class is deterministic, precisely targeted at the exact defects being fixed, and its commit message documents that the added cases genuinely fail against the pre-fix code. Compatibility is fully preserved — single-failure callers see byte-identical exceptions to before. I don't see a meaningfully better alternative approach (Option B would be try-with-resources on `Session`, but `Driver`/`Session` don't compose cleanly with the flush-then-close ordering and the need to keep the *first* failure, so the explicit accumulator is the right call here). Nice work — thanks for tracking down and fixing this leak.

PDGGK:
Thanks for the thorough review, and for pushing back on the first version — the suppressed-exception point was the right call and the PR is better for it.

Agreed on leaving the helpers duplicated per class for now; a shared utility makes sense once a third connector wants the same shape, not before.

DanielLeens:
You're welcome — and thanks for taking the pushback well and turning it into a genuinely better fix rather than a defensive patch. The suppressed-exception distinction is easy to gloss over in a nested `try/finally`, so I appreciate that you re-ran the three new cases against the pre-fix commit to confirm they actually fail there rather than just passing trivially against the final version.

Agreed on the helper duplication too — following the existing Pulsar precedent is the right call here, and extracting a shared utility with only two data points (Pulsar + Neo4j) would be premature abstraction. Worth revisiting if a third connector ends up wanting the same close-and-collect shape.

The review is already in as Approved with no blockers, so this is ready to merge from my side. Thanks again for the careful fix.

PDGGK:
Appreciated — and the pushback was worth having. `try/finally` looked like it covered the case right up until you asked what the caller actually ends up seeing, which is the part I'd skipped over.

Nothing outstanding from my side: CI is green and the only thing left is a committer review.

DanielLeens:
Glad it landed well — that's exactly the kind of finding review is for. Nothing further from my side either; agreed this is down to CI status and a committer review from here.

PDGGK:
Status check, since this has been quiet for a few days: CI is green (3/3) and review is settled — @DanielLeens reviewed it, the suppressed-exception gap he raised was fixed in 477b2d0, and his closing note was that it's down to a committer from here.

The change itself is small: `Neo4jSinkWriter.close()` and `Neo4jSourceReader.close()` released the `Session` and then the `Driver` as consecutive unguarded statements, so a throw from the session close leaked the driver. Both now close the driver regardless and keep the first failure, attaching the second as suppressed rather than replacing it.

@davidzollo would you have a moment to take a look? No urgency, just flagging that it's ready rather than waiting on anything from my side.

DanielLeens:
Confirming your summary is accurate — CI is green and there's nothing outstanding from my side since the suppressed-exception fix landed in 477b2d0. This has been down to CI status and a committer review since 2026-08-11, and that's still where it stands. Thanks for the clean status writeup.

davidzollo (approved):
LGTM. Current head has green CI and no unresolved review threads; approving to clear the merge gate.

PR #11751, opened 2026-08-10
[Docs][Connector-V2] Improve GoogleBigtable GoogleFirestore Lance Typesense and Assert connector docs

DanielCarter-stack:
## Purpose
Improve the connector documentation for GoogleBigtable, GoogleFirestore, Lance, Typesense and Assert by filling in missing structured metadata, clarifying the actual behavior of each connector against the current source, and adding the streaming / multi-table examples that were previously missing.

## Scope of changes
- `docs/en/connectors/source/GoogleBigtable.md` / `docs/en/connectors/sink/GoogleBigtable.md` (and the matching zh files)
  - Add Support These Engines sections and the full Key Features checkbox list (including column projection / CDC / timer-flush).
  - Document that the source is bounded — one split per configured table or row-key range — and that parallelism does not shard the scan.
  - Document the `rowkey_column` STRING vs BYTES requirement and the lexicographic-only `start_rowkey` / `end_rowkey` behavior.
  - Document that every cell mutation is unconditional, so `UPDATE` / `DELETE` row kinds are not interpreted as CDC operations.
  - Add a streaming checkpoint-flush sink example and a bounded streaming source example that filters by cell versions.
- `docs/en/connectors/sink/GoogleFirestore.md` (and zh)
  - Fill out the Key Features checkboxes (stream + timer-flush were missing).
  - Clarify that Firestore document IDs are auto-generated, so each row triggers an `add` call rather than a CDC operation.
  - Note the BATCH vs STREAMING behavior (every checkpoint flushes the in-memory write buffer).
  - Add a streaming checkpoint-flush task example.
- `docs/en/connectors/sink/Lance.md` (and zh)
  - Fill out the Key Features checkboxes (CDC / batch / stream / timer-flush).
  - Document the current type-mapping caveat: every integer SeaTunnel type narrows to Arrow `int32`, so `BIGINT` values outside the signed 32-bit range are truncated; `TIME` uses millisecond precision and `TIMESTAMP` is hardcoded to Asia/Shanghai.
  - Document the `lance.write.enable.stable.row.ids` option (it was listed in the table but never explained).
  - Add APPEND-with-larger-fragment and streaming checkpoint-flush examples.
- `docs/en/connectors/source/Typesense.md` / `docs/en/connectors/sink/Typesense.md` (and zh)
  - Add Support These Engines sections and the full Key Features checkboxes.
  - Document that `hosts` is not used for parallel scan sharding on the source side; the source picks the first reachable node.
  - Document that `batch_size` maps to the Typesense `per_page` parameter.
  - Document that `UPDATE` / `DELETE` row kinds are upserted by document `id` (not interpreted as CDC), and add a streaming upsert example paired with `data_save_mode = DROP_DATA` + a stable `primary_keys`.
- `docs/en/connectors/sink/Assert.md` (and zh)
  - Fill out the Key Features checkboxes (CDC / batch / stream / timer-flush).
  - Clarify that Assert is a terminal sink with no external system to write to, and that `UPDATE` / `DELETE` row kinds are not interpreted as CDC operations.
  - Document the streaming checkpoint-window row-count semantics for `MIN_ROW` / `MAX_ROW`.
  - Add a streaming validation example.

## Notes
- No code or e2e logic was changed; this PR only updates documentation.
- Config names, defaults, and example options match the current connector implementation.
- I did not bump or edit the supported version metadata for any connector.


DanielLeens (changes requested):
- **User pain point**: The docs for `GoogleBigtable` (source+sink), `GoogleFirestore` (sink), `Lance` (sink), `Typesense` (source+sink), and `Assert` (sink) were thin — missing engine-support blocks, missing feature checkboxes (cdc/batch/stream/timer-flush), stale option defaults (e.g. `multi_table_sink_replica` shown as `-` instead of the real default `1`), and no runnable streaming examples, so users had to read the Java source to understand null handling, row-key semantics, CDC behavior, and checkpoint interaction.
- **Fix approach**: Add "Support These/Those Engines" sections, fill in missing feature checkboxes, correct stale defaults, add `:::tip` callouts describing runtime semantics (CDC interpretation, null handling, checkpoint flush behavior), and add new streaming/batch HOCON examples for each connector, mirrored in `docs/en` and `docs/zh`.
- **One-sentence summary**: A docs-only PR that meaningfully expands five connectors' documentation, but in several places it adds specific behavioral claims that do not match what the underlying (unmodified) connector source code actually does.

# 1. Code Change Review

## 1.1 Core Logic Analysis

No production code is touched by this PR — `git diff apache/dev...pr-11751 --stat` shows only the 14 doc files under `docs/en/connectors/**` and `docs/zh/connectors/**` (7 English + 7 Chinese, one-to-one mirrors). Because the PR is documentation-only, "core logic review" here means verifying every new option/default/behavior claim against the real `Option<...>` definitions and the Factory/Sink/Source/Writer/Reader classes for the five touched connectors. I pulled every relevant class via `git show pr-11751:<path>` and cross-checked line by line. Summary of what I found:

**Correct / well-verified additions** (for context, so the issues below aren't misread as "everything is wrong"):
- `multi_table_sink_replica` default fix from `-` to `1` is correct for GoogleBigtable, Typesense, and Lance — confirmed against `SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA` (`seatunnel-api/.../SinkConnectorCommonOptions.java:30`, `defaultValue(1)`).
- Bigtable's `column_family` fallback description ("falls back to `all_columns`, or default family `cf`") exactly matches `BigtableSinkWriter` constructor + `resolveFamily()` (`connector-google-bigtable/.../sink/BigtableSinkWriter.java`).
- Bigtable's `schema_save_mode`/table-must-exist note matches `BigtableSink.handleSaveMode()`, which throws for every mode except `RECREATE_SCHEMA`/`APPEND_DATA`.
- Lance's new "BIGINT truncation to Arrow int32" callout is accurate — `LanceTypeMapper.convertJsonArrowType()` maps `TINYINT/SMALLINT/INT/BIGINT` all to `"int32"`.
- Lance's new "`TIMESTAMP` → timestamp (microsecond, Asia/Shanghai timezone)" and "`TIME` → time32 (millisecond precision)" rows are **exactly correct** — `SchemaUtils.convertSchema()` hardcodes `new ArrowType.Timestamp(TimeUnit.MICROSECOND, "Asia/Shanghai")` and `new ArrowType.Time(TimeUnit.MILLISECOND, 32)`. (This is a genuinely useful catch — it surfaces a latent hardcoded-timezone bug in the connector itself; that bug is out of scope for a docs PR, but documenting it accurately is the right call.)
- Typesense's "every table routes to the same configured collection in multi-table jobs" is correct: `TypesenseSinkFactory.createSink()` unconditionally renames the incoming `CatalogTable` to `readonlyConfig.get(TypesenseSinkOptions.COLLECTION)` before it reaches `TypesenseSinkWriter`, so `collection = catalogTable.getTableId().getTableName()` in the writer is really just the static `collection` option, not the per-table name.
- GoogleFirestore's "every row triggers a Firestore `add` call that produces a new document" and "no CDC interpretation" claims are correct — `FirestoreSinkWriter.write()` calls `collectionReference.add(...)` unconditionally with no `RowKind` branching.

**Issues** (full detail in 1.4): five High-severity and two Medium-severity doc-vs-source mismatches, listed as Issue 1–7 below. The common thread is that several of the *new* prose additions (not the option tables, which are mostly accurate) assert specific runtime mechanics — "dynamic table routing," "checked once per checkpoint window," "flushes the in-memory buffer on every checkpoint," "controls whether Lance assigns stable row IDs," "retries only on transient failures" — that are either unimplemented, implemented differently, or contradicted by a `static`/cumulative field in the writer. These are exactly the kind of claims this review protocol asks to be checked against source rather than taken on faith, and I did check them against the real writer/reader classes in this same worktree.

### Runtime path checked for the most severe finding (Assert streaming validation)

```
FakeSource (STREAMING) -> AssertSinkWriter.write(row)      [called once per row, N times]
                              -> ASSERT_EXECUTOR.fail(...)  [field rules checked here, per row - doc is right about this part]
                              -> LONG_ACCUMULATOR.accumulate(1)  [static, cumulative, keyed by table name]
        ...
checkpoint 1 barrier -> AssertSinkWriter.prepareCommit()    [NOT overridden -> inherited no-op, no row-rule check happens]
        ...
checkpoint N barrier -> AssertSinkWriter.prepareCommit()    [still no row-rule check]
        ...
job/task shutdown -> AssertSinkWriter.close()               [THE ONLY place MIN_ROW/MAX_ROW is evaluated,
                                                               against the cumulative LONG_ACCUMULATOR total
                                                               since writer construction]
```
`prepareCommit()` is the per-checkpoint hook in the SeaTunnel V2 sink contract; `AssertSinkWriter` does not override it, so nothing checkpoint-related happens to the row-count rules until the writer is closed at the very end of the job's life. The new doc section describes the opposite: periodic, per-checkpoint-window, reset-each-time validation.

## 1.2 Compatibility Impact
**Fully compatible.** No `Option` is renamed, removed, or has its default changed in code (the doc changes for `multi_table_sink_replica` are *corrections* to match existing code, not code changes). No public API, SPI, serialization format, or protocol is touched. No migration guidance is needed and `docs/en/introduction/concepts/incompatible-changes.md` does not need an entry.

None — zero runtime code changes, so there is no CPU/memory/GC/network/lock/retry/idempotency impact from this PR itself.

Not applicable to runtime error handling (no code changed). However, per this protocol's "verify every documented behavior claim against source" mandate for docs-only connector PRs, I'm listing the doc-accuracy problems as formal numbered issues since several are functionally misleading to a user:

### Issue 1 — GoogleBigtable sink: "dynamic table routing" claim is not implemented
- **Location**: `docs/en/connectors/sink/GoogleBigtable.md` (`### multi_table_sink_replica`), mirrored in `docs/zh/connectors/sink/GoogleBigtable.md`
- **Problem**: The new sentence reads: *"The Bigtable sink writer supports dynamic table routing, so this option takes effect in multi-table jobs."* But `BigtableParameters.buildWithConfig()` (`connector-google-bigtable/.../config/BigtableParameters.java`) always sets `table` from the static `BigtableBaseOptions.TABLE` config option — it is never derived from the incoming `CatalogTable`. `BigtableSinkWriter`/`BigtableClient` bind to exactly one Bigtable table name for the lifetime of the writer. Implementing `SupportMultiTableSinkWriter` only means "multiple independent replica writer instances can be spun up," not "one writer instance dynamically routes rows to different Bigtable tables based on upstream table identity" (contrast with Typesense/Lance, which genuinely do rename the target per sink instance in their `SinkFactory`).
- **Potential risk**: A user running a multi-table job (e.g. multiple upstream CDC tables) against Bigtable, expecting each source table to land in its own Bigtable table because the doc says "dynamic table routing," will instead have every upstream table's rows land in the single Bigtable table named by the static `table` option — silent data comingling, not a crash.
- **Best improvement**: Replace with something like: *"`multi_table_sink_replica` increases the number of parallel writer replicas within a single sink instance; the target Bigtable table is fixed by the `table` option and is not derived per upstream table."*
- **Severity**: High

### Issue 2 — GoogleBigtable source: "mixing BYTES/STRING row-key types is not supported" is false
- **Location**: `docs/en/connectors/source/GoogleBigtable.md` (`### rowkey_column`), mirrored in `docs/zh`
- **Problem**: New text: *"Mixing types in the same scan is not supported; the chosen type applies to every row-key field."* `BigtableSourceReader.convertRow()` assigns the raw row-key `ByteString` to every field listed as a row-key field, then `BigtableDeserializationFormat.deserializeCell()` decodes **each field independently based on its own declared `SeaTunnelDataType`** (`BYTES` → raw bytes, `STRING` → UTF-8 decode). There is no shared/global "chosen type" state anywhere in the reader or format classes — nothing prevents (or even notices) one row-key field being `BYTES` and another `STRING` in the same schema.
- **Potential risk**: Discourages a legitimate, working configuration (e.g. one field exposing the raw key bytes for downstream binary processing, another exposing a human-readable UTF-8 view) that the code already supports.
- **Best improvement**: Drop the "mixing not supported" sentence, or replace with an accurate one: each row-key field is decoded independently according to its own declared type in `schema.fields`.
- **Severity**: Medium

### Issue 3 — GoogleBigtable sink: overclaimed null-rowkey failure semantics for composite keys
- **Location**: `docs/en/connectors/sink/GoogleBigtable.md` (`### rowkey_column`), mirrored in `docs/zh`
- **Problem**: *"Each row-key column must not be `null`; rows with a missing or empty row key fail the job with `WRITE_FAILED`."* In `BigtableSinkWriter.buildRowKey()`, the single-rowkey-column path does throw on a null/empty value (`rowKey.isEmpty()` check in `convertRowToMutation`). But the **multi-column** path (`rowkeyColumnIndexes.size() > 1`) does `parts.add(field == null ? "" : field.toString())` and joins with the delimiter — a null in *one* of several row-key columns silently becomes an empty string segment inside the composite key and does **not** fail the job unless the entire joined result happens to be empty.
- **Potential risk**: Users configuring multiple `rowkey_column` fields may believe any null is caught and rejected, when in practice a null in a non-last/non-only column silently corrupts the row key (e.g. `"tenant#"` instead of `"tenant#eventid"`) without any error.
- **Best improvement**: Qualify the claim: *"With a single row-key column, a null or empty value fails the job with `WRITE_FAILED`. With multiple row-key columns, a null value becomes an empty segment in the composite key and only fails the job if the entire composed key ends up empty."*
- **Severity**: Medium

### Issue 4 — Assert sink: new "Streaming Validation" section misdescribes MIN_ROW/MAX_ROW checkpoint semantics
- **Location**: `docs/en/connectors/sink/Assert.md` (`## Streaming Validation` section and the `### Stream Validation With Checkpoint Window` example), mirrored in `docs/zh/connectors/sink/Assert.md`
- **Problem**: *"In streaming mode, the row count rules (`MIN_ROW` / `MAX_ROW`) are only checked once per checkpoint window, while field rules are checked on every row."* and the example text *"every checkpoint window re-checks the `MIN_ROW` / `MAX_ROW` window against the rows received since the previous checkpoint."* Per the runtime-path trace in 1.1: `AssertSinkWriter` does not override `prepareCommit()` (the per-checkpoint hook), so nothing checkpoint-related evaluates row rules. The row-count check happens **exactly once**, inside `close()` — called when the sink writer is disposed at job end, not per checkpoint — and it compares against `LONG_ACCUMULATOR`, a `static` `ConcurrentHashMap<String, LongAccumulator>` field that accumulates the row count for the table's **entire lifetime**, never reset between checkpoints (`AssertSinkWriter.java`, `write()`/`close()`).
- **Potential risk**: This is the most user-facing/functionally wrong claim in the PR. A user following the new example, expecting a streaming job to fail fast the first time a checkpoint window falls outside `[50, 5000]` rows, will instead get exactly one MIN_ROW/MAX_ROW check at job shutdown against the cumulative total since the job started — completely different failure semantics and timing. Someone using this for production data-quality gating on a long-running streaming job would be badly misled.
- **Best improvement**: Rewrite to state the real semantics: row-count rules are evaluated once, when the sink writer closes (job stop/savepoint/failure), against the cumulative row count observed by that writer since it was created; they are not re-evaluated or reset per checkpoint. If genuine per-checkpoint-window row-count validation is desired, that would require a source code change (out of scope for this docs PR) — the doc should not describe behavior the code doesn't have.
- **Severity**: High

### Issue 5 — GoogleFirestore sink: "checkpoint flushes the in-memory write buffer" — no such buffer exists
- **Location**: `docs/en/connectors/sink/GoogleFirestore.md` (Notes bullet + `### Streaming write with checkpoint flush`), mirrored in `docs/zh/connectors/sink/GoogleFirestore.md`
- **Problem**: *"The connector works in both `BATCH` and `STREAMING` job modes. In streaming mode every checkpoint flushes the in-memory write buffer before the commit returns."* `FirestoreSinkWriter` (`connector-google-firestore/.../sink/FirestoreSinkWriter.java`) has **no buffer field at all** and does not override `prepareCommit()`. `write()` calls `collectionReference.add(serializer.serialize(seaTunnelRow))` directly and synchronously per row — and doesn't even wait on the `ApiFuture` that `.add()` returns, so writes aren't even guaranteed to have reached Firestore by the time `write()` returns, let alone "flushed at checkpoint."
- **Potential risk**: Implies a durability/consistency guarantee at checkpoint boundaries (buffer flush = data is safely persisted before the checkpoint completes) that the connector does not provide. A user relying on this for exactly-once-adjacent guarantees around checkpoints would be misled about failure-recovery data loss risk.
- **Best improvement**: Remove the buffer/flush claim; state plainly that each row is written individually and asynchronously via the Firestore client's `add()` call with no batching or explicit checkpoint-synchronized flush in the current implementation.
- **Severity**: High

### Issue 6 — Lance sink: `lance.write.enable.stable.row.ids` is documented as functional but is a dead/no-op config value
- **Location**: `docs/en/connectors/sink/Lance.md` (`### lance.write.enable.stable.row.ids`), mirrored in `docs/zh/connectors/sink/Lance.md`
- **Problem**: New section: *"Controls whether Lance assigns stable row IDs to the written rows. Stable row IDs let later queries and joins identify the same row across versions of the dataset; disable this only when you do not need stable row identity and want slightly faster writes."* `LanceSinkConfig` does read the option into `this.enableStableRowIds = pluginConfig.get(LanceSinkOptions.WRITE_ENABLE_STABLE_ROW_IDS)` and exposes `getEnableStableRowIds()` — but `LanceSinkWriter.initializeDataset()` builds `WriteParams` with only `.withMaxBytesPerFile(...)`, `.withMaxRowsPerFile(...)`, `.withMode(...)`, and `.withStorageOptions(...)`; `getEnableStableRowIds()` is **never called anywhere** in `LanceSinkWriter` or the `WriteParams.Builder` chain (confirmed via `git grep -n "StableRowIds\|withEnableStable"` across `connector-lance`, which only turns up the config parsing, not any consumer).
- **Potential risk**: This is exactly the class of bug this review protocol calls out — a doc claiming an option "controls" specific runtime behavior when it's silently swallowed. A user toggling `lance.write.enable.stable.row.ids = false` expecting "slightly faster writes" or different row-identity semantics gets neither; the setting has zero effect on the actual Lance write path today.
- **Best improvement**: Either (a) wire the option through to `WriteParams.Builder` in a follow-up connector PR before documenting it as functional, or (b) if documenting now, explicitly state the option is currently parsed but not yet applied to the underlying Lance `WriteParams` (and flag it as a known gap), rather than describing it as controlling behavior.
- **Severity**: High

### Issue 7 — Typesense sink: "retries only on transient failures (network errors and 5xx responses)" overstates the retry policy
- **Location**: `docs/en/connectors/sink/Typesense.md` (`### max_retry_count`), mirrored in `docs/zh/connectors/sink/Typesense.md`
- **Problem**: *"The connector retries only on transient bulk-request failures (network errors and 5xx responses)."* `TypesenseSinkWriter` constructs `retryMaterial = new RetryMaterial(maxRetryCount, true, exception -> true, DEFAULT_SLEEP_TIME_MS)` — the retry predicate is a constant `exception -> true`, i.e. `RetryUtils.retryWithException` retries on **every** exception type from `typesenseClient.insert(...)`, with no inspection of HTTP status or exception class to distinguish transient (network/5xx) from permanent (e.g. malformed document, schema/auth error) failures.
- **Potential risk**: Minor operational surprise, not data-safety-critical: permanent, non-retryable errors (e.g. an auth failure or a document that fails Typesense-side validation) will still be retried `max_retry_count` times with a 200ms sleep before ultimately failing, rather than failing fast as "retries only on transient failures" implies.
- **Best improvement**: *"The connector retries on any exception from the bulk-import call (network errors, timeouts, and Typesense error responses alike) up to `max_retry_count` times with a fixed 200ms backoff; it does not currently distinguish transient from permanent failures."*
- **Severity**: Medium

# 2. Code Quality Assessment

## 2.1 Coding Standards
Not applicable — no Java source is changed. The Markdown itself is well-formatted, consistently uses `hocon` code fences (correctly fixing a couple of stray ```` ```bash ```` fences that were actually HOCON), and follows the existing doc conventions (`## Support These Engines`, `## Key Features`, `## Task Example`) already used elsewhere in `docs/en/connectors`.

Not applicable — no test files (`src/test/**`) are touched by this PR, so there is no flaky-test-risk analysis to perform. **Stability rating: N/A (no UT/E2E changed).**

## 2.3 Documentation Updates
This *is* the documentation update. Coverage is good in breadth (5 connectors × source/sink × en/zh = up to 14 files, all present) but, per Section 1.4, several of the new prose additions need correction before merge since they describe behavior the current connector implementations don't actually have (Issues 1, 2, 4, 5, 6) or overstate precision of existing behavior (Issues 3, 7). The option-table default corrections (`multi_table_sink_replica: - -> 1`) are accurate and a genuine improvement.

# 3. Architectural Soundness

Precise fix in intent (this is squarely "improve connector docs"), but several of the specific behavioral claims added are inaccurate rather than "temporary workaround" or "long-term solution" concerns — this is a factual-accuracy problem within an otherwise well-targeted documentation change, not an architecture problem.

## 3.2 Maintainability
Fine — plain Markdown, consistent with existing doc structure, mirrored en/zh, no maintenance burden introduced.

## 3.3 Extensibility
N/A for a docs change.

No compatibility concerns — pure documentation, no config/API/serialization surface touched.

# 4. Issue Summary

|--------|-------|----------|----------|
| 4 | Assert "per-checkpoint-window" MIN_ROW/MAX_ROW claim contradicts `close()`-only, cumulative-counter implementation | `docs/{en,zh}/connectors/sink/Assert.md` | High |
| 5 | GoogleFirestore "checkpoint flushes in-memory write buffer" — no buffer/prepareCommit exists in `FirestoreSinkWriter` | `docs/{en,zh}/connectors/sink/GoogleFirestore.md` | High |
| 6 | Lance `lance.write.enable.stable.row.ids` documented as functional but never wired into `WriteParams` | `docs/{en,zh}/connectors/sink/Lance.md` | High |
| 1 | GoogleBigtable "dynamic table routing" claim — `table` is static per writer, not derived from CatalogTable | `docs/{en,zh}/connectors/sink/GoogleBigtable.md` | High |
| 2 | GoogleBigtable source "mixing row-key types not supported" — false, each field decoded independently | `docs/{en,zh}/connectors/source/GoogleBigtable.md` | Medium |
| 3 | GoogleBigtable sink overclaims null-rowkey failure for multi-column composite keys | `docs/{en,zh}/connectors/sink/GoogleBigtable.md` | Medium |
| 7 | Typesense "retries only on transient failures" — actual predicate retries on all exceptions | `docs/{en,zh}/connectors/sink/Typesense.md` | Medium |

# 5. Merge Recommendation

1. **Blockers — must be fixed** (sorted by severity):
   - Issue 4 (High): Rewrite the Assert "Streaming Validation" section and its example — MIN_ROW/MAX_ROW is checked once at writer close against a cumulative count, not per checkpoint window.
   - Issue 5 (High): Remove/correct the GoogleFirestore "checkpoint flushes in-memory buffer" claim — no such buffer or checkpoint hook exists.
   - Issue 6 (High): Correct or caveat the Lance `lance.write.enable.stable.row.ids` section — the option is currently a no-op in the write path.
   - Issue 1 (High): Correct the GoogleBigtable "dynamic table routing" claim for `multi_table_sink_replica` — the target table is static per writer.

   - Issue 2 (Medium): Drop/correct the GoogleBigtable source "mixing row-key types not supported" claim.
   - Issue 3 (Medium): Qualify the GoogleBigtable sink null-rowkey-failure claim for the multi-column case.
   - Issue 7 (Medium): Soften the Typesense retry-policy claim to match the actual `exception -> true` predicate.

**Overall assessment**: This is a well-intentioned and mostly well-executed documentation expansion — engine-support blocks, feature checkboxes, and the `multi_table_sink_replica` default correction are all accurate improvements, and two of the new claims (Lance's Asia/Shanghai timestamp hardcoding, Lance's BIGINT→int32 truncation) are genuinely valuable, source-verified callouts that will save future users real debugging time. The problem is that four of the newly-added behavioral claims (Issues 1, 4, 5, 6) describe checkpoint/routing/config behavior that the current connector source does not actually implement — these read very authoritatively (specific mechanics, dedicated example configs) and would actively mislead a user building on them, which is worse than under-documenting. None of this requires a source code change to fix — only correcting the prose to match what `AssertSinkWriter`, `FirestoreSinkWriter`, `LanceSinkWriter`, and `BigtableSinkWriter`/`BigtableParameters` actually do today. Given the contributor already got the harder, more subtle catches right (the Lance timezone/truncation notes), I'd expect a quick turnaround on these corrections.

No better alternative approach exists here — the PR's structure and scope are appropriate; it just needs the four High-severity claims re-verified against source (as I did above) and corrected before merge.

DanielLeens (approved):
- **User pain point**: The docs for `GoogleBigtable` (source+sink), `GoogleFirestore` (sink), `Lance` (sink), `Typesense` (source+sink), and `Assert` (sink) were thin — missing engine-support blocks, missing feature checkboxes (cdc/batch/stream/timer-flush), stale option defaults, and no runnable streaming examples.
- **Fix approach**: Add "Support These/Those Engines" sections, fill in missing feature checkboxes, correct stale defaults, add `:::tip`-style prose describing runtime semantics, and add new streaming/batch HOCON examples for each connector, mirrored in `docs/en` and `docs/zh`.
- **One-sentence summary**: A docs-only PR that meaningfully expands five connectors' documentation; the follow-up commit `3852eeed2` fully addresses every blocker from the previous review round.

Thanks for turning this around so fast, and for taking every single point seriously instead of just the High-severity ones — I went back through all seven items (four blockers, three non-blocking) against the current connector source in this worktree, and every one of them now matches what the code actually does. Nice work.

# 1. Code Change Review

## 1.1 Core Logic Analysis

This is still a pure documentation change — `git diff dev...HEAD --stat` touches only the 14 Markdown files under `docs/en/connectors/**` and `docs/zh/connectors/**` (7 English + 7 Chinese mirrors), no `src/main` or `src/test` files. Since the last review, one new commit landed: `3852eeed2` ("Address review feedback..."), which rewrites exactly the passages I flagged, on top of the original `9fb05da72`. I re-verified the *entire current diff* against `dev`, not just the incremental commit, and re-checked every one of the seven previously-flagged claims against the live connector source in this worktree:

| # | Previous claim (wrong) | Current text | Source verified against | Status |
|---|---|---|---|---|
| 1 | Bigtable sink "supports dynamic table routing" for `multi_table_sink_replica` | "`multi_table_sink_replica` increases the number of parallel writer replicas within a single sink instance; the target Bigtable table is fixed by the `table` option and is not derived per upstream table." | `BigtableParameters.buildWithConfig()` sets `table` from the static `BigtableBaseOptions.TABLE` config in both overloads (`connector-google-bigtable/.../config/BigtableParameters.java:61,80`) — no per-`CatalogTable` derivation anywhere. | **Fixed** |
| 2 | Bigtable source "mixing BYTES/STRING row-key types is not supported" | "Each listed field is decoded independently according to its own declared type in `schema.fields`... Different row-key fields can therefore use different types in the same scan." | `BigtableSourceReader.convertRow()` looks each field up by name against `rowKeyFieldNames` and calls `BigtableDeserializationFormat.deserializeCell(fieldType, ...)` per field (`connector-google-bigtable/.../source/BigtableSourceReader.java:231-243`, `.../format/BigtableDeserializationFormat.java:70-116`) — no shared/global type state. | **Fixed** |
| 3 | Bigtable sink "each row-key column must not be null" (implying uniform failure) | "With a single row-key column, a null or empty value fails the job... With multiple row-key columns, a null value in any non-last column silently becomes an empty segment... only when the entire composed key is empty does the job fail." | `BigtableSinkWriter.buildRowKey()`: single-column path returns `fieldToByteString()` (empty `ByteString` on null, caught by the `rowKey.isEmpty()` check at line 157); multi-column path does `field == null ? "" : field.toString()` then joins (`.../sink/BigtableSinkWriter.java:200-211`). Matches exactly. | **Fixed** |
| 4 | Assert sink: MIN_ROW/MAX_ROW "checked once per checkpoint window" | "Row count rules... are evaluated exactly once when the sink writer closes... against the cumulative row count... not per checkpoint window." | `AssertSinkWriter` has no `prepareCommit()` override; `LONG_ACCUMULATOR` is a `static ConcurrentHashMap` accumulated in `write()`, and MIN_ROW/MAX_ROW are only evaluated inside `close()` (`connector-assert/.../sink/AssertSinkWriter.java:49,89,105,133,136`). Matches exactly. | **Fixed** |
| 5 | Firestore sink: "every checkpoint flushes the in-memory write buffer" | "`FirestoreSinkWriter.write()` calls the Firestore client's `add(...)` once per row and does not buffer or batch rows, so there is no in-memory write buffer to flush at checkpoint boundaries." | `FirestoreSinkWriter.write()` is `collectionReference.add(serializer.serialize(seaTunnelRow))`, called directly, no buffer field anywhere in the class, no `prepareCommit()` override (`connector-google-firestore/.../sink/FirestoreSinkWriter.java:69-70`). Matches exactly. | **Fixed** |
| 6 | Lance sink: `lance.write.enable.stable.row.ids` "controls whether Lance assigns stable row IDs" | "the value is currently parsed but not yet applied to the underlying Lance `WriteParams`... This is a known gap pending a follow-up connector change." | `LanceSinkConfig.enableStableRowIds` is only ever read via `getEnableStableRowIds()` and never referenced elsewhere; `LanceSinkWriter.initializeDataset()`'s `WriteParams.Builder` chain calls only `.withMaxBytesPerFile()/.withMaxRowsPerFile()/.withMode()/.withStorageOptions()` (`connector-lance/.../sink/LanceSinkWriter.java:104-109`, confirmed via a repo-wide grep for `StableRowIds`/`withEnableStable` turning up only the config-parsing lines). Matches exactly. | **Fixed** |
| 7 | Typesense sink: "retries only on transient failures (network errors and 5xx responses)" | "The retry predicate is `exception -> true`, so the connector retries on every exception... it does not currently distinguish transient from permanent failures." | `TypesenseSinkWriter` constructs `new RetryMaterial(maxRetryCount, true, exception -> true, DEFAULT_SLEEP_TIME_MS)` (`connector-typesense/.../sink/TypesenseSinkWriter.java:80`). Matches exactly. | **Fixed** |

All four blockers and all three non-blocking recommendations from the previous round are resolved, and the corrected prose is now an accurate description of the real runtime behavior in each case — not just plausible-sounding, but line-for-line consistent with the writer/reader classes.

**Key Findings**
- Every corrected claim in `3852eeed2` was independently re-derived from source in this pass (not just diffed against my own suggested wording), and all seven check out.
- `docs/en` and `docs/zh` were updated in lockstep for every corrected passage — I diffed both language trees and the same seven sections changed with equivalent (translated, not necessarily byte-identical) meaning in each.
- No dangling references were left behind — e.g. the GoogleFirestore sink example heading was renamed from "Streaming write with checkpoint flush" to "Streaming write with checkpoint interval" in both `docs/en` and `docs/zh`, and nothing else in either file still refers to the old "checkpoint flush" wording.
- I additionally spot-checked several *unmodified* claims from the first commit that weren't in my previous review's explicit checklist, since a re-review should cover the full current diff, not just the delta: the `multi-table write` feature checkboxes for Bigtable sink (`[x]`) and Lance sink (`[x]`) correctly correspond to `BigtableSink`/`LanceSink` both `implements ... SupportMultiTableSink`; the Typesense sink `CDC [x]` checkbox correctly corresponds to `TypesenseSinkWriter.write()` branching on `RowKind` (INSERT/UPDATE_BEFORE/etc.); the Firestore sink `multiple table write [ ]` (unchecked) correctly corresponds to `FirestoreSink extends AbstractSimpleSink` (no `SupportMultiTableSink`). All consistent with source.
- No production code is touched anywhere in the diff — confirmed via `git diff --stat dev...HEAD` showing only `docs/en/connectors/**` and `docs/zh/connectors/**` paths.

No mermaid/runtime-path diagram is needed here beyond the trace table above — this PR touches zero `seatunnel-engine`/`seatunnel-api`/checkpoint/serialization/cross-stage code.

## 1.2 Compatibility Impact
**Fully compatible.** No `Option` is renamed, removed, or has its default changed in code. No public API, SPI, serialization format, or protocol is touched. No migration guidance is needed.

None — zero runtime code changes.

Not applicable — no production error-handling code changed. No new formal issues to raise in this round; all issues from the previous round are resolved (see table in 1.1).

# 2. Code Quality Assessment

## 2.1 Coding Standards
Not applicable — no Java source is changed. Markdown formatting, code-fence language tags, and section structure (`## Support These/Those Engines`, `## Key Features`, `## Task Example`) remain consistent with the rest of `docs/en/connectors` and `docs/zh/connectors`.

Not applicable — no test files (`src/test/**`) are touched by this PR. **Stability rating: N/A (no UT/E2E changed).**

## 2.3 Documentation Updates
This *is* the documentation update, and after this commit it is accurate. All seven previously-flagged doc-vs-source mismatches are corrected, verified line-by-line above. `docs/en` and `docs/zh` remain mirrored for every changed section.

# 3. Architectural Soundness

Precise fix — the corrections are minimal, targeted rewrites of exactly the sentences that were previously wrong, not a wholesale restructure. This is the right scope for a docs-accuracy follow-up.

## 3.2 Maintainability
Fine — plain Markdown, consistent structure, mirrored en/zh, no maintenance burden introduced.

## 3.3 Extensibility
N/A for a docs change.

No compatibility concerns — pure documentation, no config/API/serialization surface touched.

# 4. Issue Summary

No open issues. All seven issues raised in the previous review round (4 High, 3 Medium) are verified fixed in commit `3852eeed2`:

| Number | Issue | Location | Severity | Status |
|--------|-------|----------|----------|--------|
| 1 | GoogleBigtable "dynamic table routing" claim | `docs/{en,zh}/connectors/sink/GoogleBigtable.md` | High | Fixed |
| 4 | Assert "per-checkpoint-window" MIN_ROW/MAX_ROW claim | `docs/{en,zh}/connectors/sink/Assert.md` | High | Fixed |
| 5 | GoogleFirestore "checkpoint flushes write buffer" claim | `docs/{en,zh}/connectors/sink/GoogleFirestore.md` | High | Fixed |
| 6 | Lance `lance.write.enable.stable.row.ids` documented as functional | `docs/{en,zh}/connectors/sink/Lance.md` | High | Fixed |
| 2 | GoogleBigtable source "mixing row-key types not supported" | `docs/{en,zh}/connectors/source/GoogleBigtable.md` | Medium | Fixed |
| 3 | GoogleBigtable sink overclaimed null-rowkey failure semantics | `docs/{en,zh}/connectors/sink/GoogleBigtable.md` | Medium | Fixed |
| 7 | Typesense "retries only on transient failures" overstatement | `docs/{en,zh}/connectors/sink/Typesense.md` | Medium | Fixed |

# 5. Merge Recommendation

1. **Blockers — must be fixed**: None.
2. **Recommended fixes — non-blocking**: None.

**Overall assessment**: This addresses every point from the previous review, and I re-verified each fix independently against the current connector source rather than taking the commit message's word for it — all seven corrected passages are now accurate descriptions of what `BigtableSinkWriter`, `BigtableSourceReader`/`BigtableDeserializationFormat`, `BigtableParameters`, `AssertSinkWriter`, `FirestoreSinkWriter`, `LanceSinkWriter`/`LanceSinkConfig`, and `TypesenseSinkWriter` actually do today. I also spot-checked several untouched feature-matrix checkboxes from the first commit and found them consistent with the `SupportMultiTableSink`/`RowKind`-handling source. This disagrees with my own previous `CHANGES_REQUESTED` — I'm superseding that verdict with an approval now that the underlying facts have been fixed. No merge conflicts (`mergeable: true`); CI was still `in_progress` on the Build check at review time, so please confirm it goes green before merging, but there's nothing left for me to block on from a documentation-accuracy standpoint.

davidzollo (approved):
+1
LGTM

PR #9504, opened 2025-06-26
[Bug][connector-http] Fix paging request running infinitely

chncaesar:
### Purpose of this pull request
This PR close #9442

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
The code change is covered by unit-test testInternalPollNextWithBodyPlaceholderBatchSize in HttpSourceReaderUpdateRequestParamTest.
mvn clean package install -am  -Pci
image

### Check list

* [x] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [x] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [x] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config)

PR #11760, opened 2026-08-11
fix(http-source): preserve rawParams template for placeholder-based pagination (#11759)

### LINKED ISSUE #11759: http pageing bug
### Search before asking

### What happened

Http 连接器使用分页获取数据时,pageIndex不递增,调用接口陷入死循环,seatunnel配置如下:
接口/metricResult 中接收 pageIndex 的值一直为1 ,导致死循环

### SeaTunnel Version

apache-seatunnel-2.3.13

### SeaTunnel Config

```conf
{
    "env": {
        "job.mode": "batch"
    },
    "sink": [
        {
            "data_save_mode": "drop_data",
            "database": "edp_rnd",
            "driver": "org.postgresql.Driver",
            "generate_sink_sql": true,
            "password": "sTb_38Xu-l2gQPOV",
            "plugin_input": [
                "47575ca3_f4e8_460b_b92e_0c7fc5861276"
            ],
            "plugin_name": "Jdbc",
            "schema_save_mode": "recreate_schema",
            "table": "public.st_sink_03",
            "url": "jdbc:postgresql://162.14.225.36:15432/edp_rnd",
            "user": "postgres"
        }
    ],
    "source": [
        {
            "format": "json",
            "headers": {
                "x-token": "Bi123456"
            },
            "json_field": {
                "dimension_name": "$.body.cubeMetricValueDTOList[*].dimension0ValueDTO.dimensionName",
                "metric_catalog_code": "$.body.cubeMetricValueDTOList[*].metric_catalog_code",
                "metric_catalog_name": "$.body.cubeMetricValueDTOList[*].mcMetricCatalogBO.metricCatalogName",
                "metric_name": "$.body.cubeMetricValueDTOList[*].mcMetricBO.metricName"
            },
            "method": "get",
            "pageing": {
                "batch_size": 2,
                "page_field": "pageIndex",
                "start_page_number": 1,
                "use_placeholder_replacement": true
            },
            "params": {
                "pageIndex": "${pageIndex}",
                "pageSize": "2"
            },
            "plugin_name": "Http",
            "plugin_output": "47575ca3_f4e8_460b_b92e_0c7fc5861276",
            "schema": {
                "fields": {
                    "dimension_name": "string",
                    "metric_catalog_code": "string",
                    "metric_catalog_name": "string",
                    "metric_name": "string"
                }
            },
            "url": "http://localhost:51000/sc-edp/metric-api/api/mc/v1/bo/metricCatalog/4e62bca6-36d9-41b0-b901-8fc4bf55864d/auth/0/isDefaultAlertSystem/1/alertSystem/1fa1df19-a96c-4cb9-b6d0-0acc457a576a/metricResult"
        }
    ],
    "transform": []
}
```

### Running Command

```shell
seatunnel-cluster.cmd
```

### Error Exception

```log
无
```

zeta 2.3.13

### Java or Scala Version

17

### Screenshots

_No response_

### Code of Conduct

waterWang:
## Description

Fix the pagination infinite-loop bug in the HTTP connector when placeholder replacement is enabled.

### Problem

When `use_placeholder_replacement=true` and `page_field` is a placeholder in `params` (e.g. `"pageIndex": "${pageIndex}"`), the page index never advances — the connector keeps requesting page 1 forever.

**Root cause:** In `HttpSourceReader.updateRequestParam()`, the `params` map is mutated in place. The first call replaces the `${pageIndex}` placeholder with the actual value (`"1"`). On every subsequent pagination iteration, the placeholder is no longer present in the map, so `replacePlaceholder` finds nothing to replace and the page index never increments, causing an infinite loop.

### Fix

Add a `rawParams` copy of the original params template (mirroring the existing `rawBody` handling). When `usePlaceholderReplacement` is enabled, rebuild the params map from `rawParams` on each pagination iteration so the `${pageField}` placeholder is always present to be replaced with the current page index.

Closes #11759

### DIFF
--- a/seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/source/HttpSourceReader.java
+++ b/seatunnel-connectors-v2/connector-http/connector-http-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/http/source/HttpSourceReader.java
@@ -82,6 +82,14 @@ public class HttpSourceReader extends AbstractSingleSplitReader<SeaTunnelRow> {
      * state is not unintentionally mutated during pagination.
      */
     private String rawBody = null;
+    /**
+     * Holds the original request params template for placeholder replacement. Without a copy of the
+     * raw params, the first {@code updateRequestParam} call replaces the {@code ${pageField}}
+     * placeholder (e.g. {@code ${pageIndex}}) with an actual value, so on subsequent pagination
+     * iterations the placeholder is gone and the page index never advances, causing an infinite
+     * loop (see #11759).
+     */
+    private Map<String, String> rawParams = null;
 
     public HttpSourceReader(
             HttpParameter httpParameter,
@@ -134,6 +142,8 @@ public class HttpSourceReader extends AbstractSingleSplitReader<SeaTunnelRow> {
         this.contentJson = contentJson;
         this.pageInfoOptional = Optional.ofNullable(pageInfo);
         this.rawBody = httpParameter.getBody();
+        this.rawParams = httpParameter.getParams() == null
+                ? null : new HashMap<>(httpParameter.getParams());
         this.binaryMode = binaryMode;
         this.binaryChunkSize = binaryChunkSize;
 
@@ -272,17 +282,43 @@ public class HttpSourceReader extends AbstractSingleSplitReader<SeaTunnelRow> {
         }
         // if not set keepPageParamAsHttpParam, but page field is in params, then set page index as
         if (MapUtils.isNotEmpty(this.httpParameter.getParams())) {
-
-            processPageMap(
-                    this.httpParameter.getParams(),
-                    pageField,
-                    pageValue.toString(),
-                    usePlaceholderReplacement);
-            processPageMap(
-                    this.httpParameter.getParams(),
-                    pageInfo.getPageCursorFieldName(),
-                    pageInfo.getCursor(),
-                    usePlaceholderReplacement);
+            if (usePlaceholderReplacement && rawParams != null) {
+                // Rebuild params from the raw template on every pagination iteration. Re-using the
+                // already-mutated params map would drop the ${pageField} placeholder after the first
+                // call, so the page index would never advance (causing an infinite loop, see #11759).
+                Map<String, String> rebuiltParams = new HashMap<>();
+                for (Map.Entry<String, String> entry : rawParams.entrySet()) {
+                    String value = entry.getValue();
+                    String replaced = replacePlaceholder(value, pageField, pageValue);
+                    if (replaced == null) {
+                        replaced = value;
+                    }
+                    if (pageInfo.getPageCursorFieldName() != null
+                            && pageInfo.getCursor() != null) {
+                        String cursorReplaced =
+                                replacePlaceholder(
+                                        replaced,
+                                        pageInfo.getPageCursorFieldName(),
+                                        pageInfo.getCursor());
+                        if (cursorReplaced != null) {
+                            replaced = cursorReplaced;
+                        }
+                    }
+                    rebuiltParams.put(entry.getKey(), replaced);
+                }
+                this.httpParameter.setParams(rebuiltParams);
+            } else {
+                processPageMap(
+                        this.httpParameter.getParams(),
+                        pageField,
+                        pageValue.toString(),
+                        usePlaceholderReplacement);
+                processPageMap(
+                        this.httpParameter.getParams(),
+                        pageInfo.getPageCursorFieldName(),
+                        pageInfo.getCursor(),
+                        usePlaceholderReplacement);
+            }
         }
 
         // 2. param in body


DanielLeens (changes requested):
- User pain point: when the HTTP source is configured with `pageing.use_placeholder_replacement = true` and the page field is a placeholder inside `params` (e.g. `"pageIndex": "${pageIndex}"`), the very first `updateRequestParam()` call replaces `${pageIndex}` with the literal value and mutates the params map in place. On every following pagination iteration the placeholder string no longer exists in the map, so nothing gets replaced, the page index never advances, and the connector loops on page 1 forever (#11759).
- Fix approach: add a `rawParams` field that snapshots the original `params` template in the constructor (mirroring the existing `rawBody` field), and when `usePlaceholderReplacement` is enabled, rebuild the params map from that untouched template on every pagination call instead of mutating the live map.
- One-sentence summary: this is a correct, narrowly-scoped fix for the params case of #11759, but the exact same root-cause bug still exists for the (documented) header-based placeholder pagination path, which this PR leaves untouched.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Changed file: `HttpSourceReader.java` (single file, no test changes).

Before:
```java
processPageMap(this.httpParameter.getParams(), pageField, pageValue.toString(), usePlaceholderReplacement);
processPageMap(this.httpParameter.getParams(), pageInfo.getPageCursorFieldName(), pageInfo.getCursor(), usePlaceholderReplacement);
```
`processPageMap` mutates `this.httpParameter.getParams()` in place (`map.putAll(updatedMap)`), destroying the `${page}` placeholder after the first pagination call.

After (`HttpSourceReader.java:285-309`):
```java
if (usePlaceholderReplacement && rawParams != null) {
    Map<String, String> rebuiltParams = new HashMap<>();
    for (Map.Entry<String, String> entry : rawParams.entrySet()) {
        String value = entry.getValue();
        String replaced = replacePlaceholder(value, pageField, pageValue);
        ...
        rebuiltParams.put(entry.getKey(), replaced);
    }
    this.httpParameter.setParams(rebuiltParams);
} else {
    processPageMap(...);   // unchanged legacy path
}
```
`rawParams` is captured once in the constructor (`HttpSourceReader.java:145-146`) as `new HashMap<>(httpParameter.getParams())`, so it is a defensive copy that is never mutated afterwards. Every pagination call now rebuilds `params` from that pristine template, so the placeholder is always present to be replaced with the current page/cursor value.

**Text-flow of the pagination loop** (`internalPollNext`, `HttpSourceReader.java:446-486`):
```
internalPollNext()
  -> loop while !noMoreElementFlag
       -> updateRequestParam(pageInfo, usePlaceholderReplacement)
            -> [headers] processPageMap(httpParameter.getHeaders(), ...)      <- still mutates in place (Issue 1)
            -> [params]  rebuild from rawParams -> httpParameter.setParams()  <- fixed by this PR
            -> [body]    processBodyString(rawBody, ...) -> setBody()        <- was already template-based, unaffected
       -> pollAndCollectData(output)  -- issues the HTTP request with the freshly rebuilt params/body
       -> pageIndex += 1 / cursor advances from response
```

**Key findings:**
1. The params fix is logically correct: I traced two consecutive iterations by hand (`pageIndex=1` then `pageIndex=2`) — `rawParams` never changes, so `rebuiltParams` correctly reflects the current page value on every call, and there is no first-page-vs-subsequent-page asymmetry (the previous bug already broke even the *second* call, not just later ones).
2. **The identical root cause is still present for `headers`.** `updateRequestParam` (`HttpSourceReader.java:270-282`) still calls `processPageMap(this.httpParameter.getHeaders(), ...)` directly, and `processPageMap` (`HttpSourceReader.java:367-388`) does `map.putAll(updatedMap)`, mutating the live headers map exactly like the old params code did. There is no `rawHeaders` template. This is not a hypothetical edge case — `docs/en/connectors/source/Http.md` "Example 2: Using page parameters in headers" explicitly documents `use_placeholder_replacement = true` with the page field in `headers`, and `HttpSourceReaderUpdateRequestParamTest.testUpdateRequestParamWithHeaderPlaceholder` exercises exactly that config. Any real multi-page job using placeholder-based pagination through headers will still hit the infinite-loop bug from #11759 after this PR merges. See Issue 1.
3. The `body` path was already immune before this PR (it already used a `rawBody` template introduced in an earlier fix, #9504), which is presumably the pattern this PR intentionally copied for `params`. It's a good sign the author reused a proven pattern — the gap is that the same pattern wasn't also applied to `headers`.
4. No test in the diff (there is no test diff at all) exercises calling `updateRequestParam`/`internalPollNext` **twice** with a params-placeholder to prove the second call still contains a valid placeholder — the one existing params-placeholder test (`testUpdateRequestParamWithParamsPlaceholder`) only calls it once, so it would pass identically before and after this fix and does not protect against a regression. See Issue 3.
5. CI's `Code style` job is failing on lines this PR touches (`HttpSourceReader.java:145-146` and `:296-297`) — pure Google-Java-Format/Spotless formatting violations, not a logic problem, but it does block merge as-is. See Issue 2.

## 1.2 Compatibility Impact
`usePlaceholderReplacement` defaults to `false`, and that code path (`processPageMap` key-based replacement) is completely untouched — fully backward compatible for existing jobs that don't use placeholder-based pagination. For jobs that do use `use_placeholder_replacement = true` with params, the previous behavior was an infinite loop / stuck-on-page-1 bug, so there is no legitimate prior behavior being changed here — this only fixes broken behavior to match what the docs already promised (`docs/en/connectors/source/Http.md` Example 1). No config option, API, or serialization format changes.

Rebuilding a `HashMap` from `rawParams` on every pagination call is O(size of params) per HTTP request, which is the same order of work `processPageMap` already did (it also builds a new `updatedMap` and merges it). Params maps in HTTP source configs are small (a handful of query parameters), so this has no measurable performance impact even for long-running paginated jobs.

**Issue 1 (High) — `HttpSourceReader.java:270-282`, `:367-388`.** Header-based placeholder pagination still has the exact bug this PR is meant to fix. `processPageMap` mutates `this.httpParameter.getHeaders()` in place via `map.putAll(updatedMap)`; after the first pagination call the `${pageField}`/`${cursor}` placeholders in headers are gone, so subsequent requests keep sending the first page's header value forever. This is reachable from the normal user path — it's explicitly documented (`docs/en/connectors/source/Http.md`, "Example 2: Using page parameters in headers") and covered only by a single-call unit test that can't detect the multi-iteration failure. Recommend adding a `rawHeaders` template analogous to `rawParams`/`rawBody` and rebuilding headers from it the same way, ideally by extracting one shared helper used by headers/params so this class of bug can't reappear for a fourth location later.

**Issue 2 (Medium, currently a hard CI blocker) — `HttpSourceReader.java:145-146`, `:296-297`.** The fork's `Code style` check (`spotless:check`) is failing specifically on lines added by this PR (the `rawParams` assignment line-wrapping and the cursor-condition line-wrapping). This is exactly the CI failure `gh pr checks 11760` reports; it is caused by this diff, not flaky/unrelated infrastructure. Fix is mechanical: run `mvn spotless:apply` (or `./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-http/connector-http-base -am`) and push the reformatted file.

**Issue 3 (Medium) — test coverage gap, no file/line since no test file was touched.** The PR fixes a bug that only manifests across *multiple* pagination iterations, but adds zero new tests, and none of the existing tests call `updateRequestParam`/`internalPollNext` twice for the params-placeholder case. Compare with `HttpSourceReaderInternalPollNextTest.testPageNumberPlaceHolderRequestBodyUpdate` (`:112-140`), which already proves the equivalent body-based fix works across two pages by asserting the final body contains `"page":"2"` after `internalPollNext` — that's the exact template that should be replicated for `params` (and ideally `headers`, once Issue 1 is fixed) to actually prove this bug is fixed and stays fixed.

No new exceptions are thrown, no new logging was added or removed, and nothing here introduces a null-pointer risk beyond what already existed (`pageValue` is always non-null because `HttpSource.java:90-94` always assigns a default `start_page_number`).

# 2. Code Quality Assessment

## 2.1 Coding Standards
The new Javadoc on `rawParams` (`HttpSourceReader.java:85-91`) is clear and explains the "why" (references #11759 directly) — good. The two Spotless violations noted as Issue 2 are the only coding-standard problem; nothing else stands out (no wildcard imports, no `System.out.println`, no new dependency).

This PR does not touch `src/test/java` at all — no new tests, no modified tests. Since no UT/E2E code was changed, the mandatory Stable/Risk present/High risk stability rating doesn't formally apply, but per the "don't assume coverage is adequate" rule: the existing suite does **not** cover the scenario this PR fixes (multi-call params-placeholder pagination), as detailed in Issue 3. I'd treat "no regression test for the bug being fixed" as a blocking gap for a bug-fix PR, not just a nice-to-have.

## 2.3 Documentation Updates
No user-visible config/behavior contract changed (no new options, no default changes), so `docs/en`/`docs/zh` don't strictly need updating for the params fix itself. However, the docs already document header-based placeholder pagination as working (Example 2), and that remains broken after this PR — once Issue 1 is fixed, no doc changes should be needed either, since the existing docs already describe the intended (correct) behavior.

# 3. Architectural Soundness

This is a **partial, precise fix** rather than a workaround: reusing the `rawXxx`-template pattern already established for `rawBody` is the right idea and keeps the change minimal and easy to reason about. However, because the same pattern wasn't factored into a shared helper and reused for `headers`, the fix is incomplete — it closes the bug for one of the three places placeholder pagination can appear (params, headers, body) while leaving an identical bug open in a second, officially documented one.

**Suggested better alternative (non-blocking, for consideration):** extract a small private helper, e.g. `rebuildFromTemplate(Map<String,String> raw, String pageField, Object pageValue, PageInfo pageInfo)`, used by both the headers block and the params block. That would fix Issue 1 with almost no extra code and prevent the same mistake from being made a third time if headers/params logic diverges again later.

## 3.2 Maintainability
Slightly reduced by the duplication between the new inline params-rebuild loop and the pre-existing `processPageMap`/`replacePlaceholder` helpers — the same "replace pageField, then chain cursor replace" logic is now expressed twice in two different styles (one destructive/in-place for headers, one template-based for params). Consolidating into one helper (see 3.1) would improve this.

## 3.3 Extensibility
Low risk either way; the change is localized to `HttpSourceReader`. Once headers are also fixed via a shared helper, adding a fourth placeholder-bearing location (if ever needed) would be low-effort.

No checkpoint/savepoint state, no serialization format, and no SPI/API contract is touched by this change — `rawParams` is a reader-local runtime field, not part of any persisted split/checkpoint state. Nothing here affects restoring jobs created with older SeaTunnel versions.

# 4. Issue Summary

| 1 | Header-based placeholder pagination still hits the same infinite-loop/stale-value bug this PR fixes for params; reachable via a documented example and only covered by a single-call test | `HttpSourceReader.java:270-282`, `:367-388` | High |
| 2 | CI `Code style` (Spotless) check fails on lines added by this PR | `HttpSourceReader.java:145-146`, `:296-297` | Medium (hard CI blocker) |
| 3 | No regression test proves the params fix survives multiple pagination iterations (the actual bug being fixed) | No test file changed; compare `HttpSourceReaderInternalPollNextTest.java:112-140` | Medium |

# 5. Merge Recommendation

   - Issue 2 (Medium/CI blocker): run `spotless:apply` so the fork's `Code style` check goes green — this is currently the literal reason CI is red.
   - Issue 1 (High): apply the same `rawXxx`-template fix to `headers`, since it's officially documented as a supported placeholder-pagination location and currently ships with the identical bug this PR was written to close. Please also add/extend a test that calls `updateRequestParam` (or `internalPollNext`) at least twice for a header-placeholder scenario to prove the fix.
   - Issue 3 (Medium): add a regression test for params that mirrors `testPageNumberPlaceHolderRequestBodyUpdate` — call `internalPollNext`/`updateRequestParam` across at least two pages and assert the second request's params still contain the correctly-advanced page value, not a stale/removed placeholder.

   - Consider extracting a shared "rebuild map from raw template" helper used by both the headers and params blocks (see 3.1/3.2) to avoid a third instance of this bug appearing later.

Overall assessment: the core idea of this PR — snapshot the raw template and rebuild from it instead of mutating in place, following the existing `rawBody` pattern — is the right fix and is implemented correctly for `params`. Great catch tracing the root cause down to the in-place mutation in `updateRequestParam`/`processPageMap`! The main gap is scope: the exact same bug is still live for `headers`, which is just as reachable from a normal, documented user path as the params case this PR fixes, so I don't think this is fully done yet. Combined with the currently-red CI (an easy fix) and the missing multi-iteration regression test, I'd like to see one more round before this merges.

davidzollo:

```

```

DanielLeens:
Hi @waterWang, thanks for joining the Slack channel!

Just a quick status update so this doesn't stall: no new commit has landed yet, so the three items from my previous review are still open:
1. **High** — the same in-place-mutation bug is still present for header-based placeholder pagination (`processPageMap` on `httpParameter.getHeaders()`), which is a documented, reachable path (`docs/en/connectors/source/Http.md`, "Example 2: Using page parameters in headers"). The `rawParams`-template pattern you applied for params would fix this the same way for headers.
2. **Medium (CI blocker)** — the fork's `Code style` (Spotless) check is currently failing on the lines you added; running `spotless:apply` locally and pushing the reformatted file should turn that green.
3. **Medium** — a regression test that calls `updateRequestParam`/`internalPollNext` across at least two pages for the params-placeholder case (mirroring `HttpSourceReaderInternalPollNextTest.testPageNumberPlaceHolderRequestBodyUpdate`) would directly protect the bug you're fixing here.

The core idea — snapshotting `rawParams` and rebuilding from that template instead of mutating in place — is exactly the right fix and correctly implemented for the params case, so this is close. Whenever you get a chance to push an update covering headers + the Spotless formatting + a regression test, I'll take another look right away. Thanks again for tracking down the root cause of #11759!

PR #11768, opened 2026-08-11
[Fix][Connector-V2] Honor Oracle snapshot select overrides

### LINKED ISSUE #9754: [Bug] [Oracle-CDC] snapshot.select.statement.overrides does not work
### Search before asking

### What happened

Hi! How are you?

Tickets without solution
https://github.com/apache/seatunnel/issues/8201
https://github.com/apache/seatunnel/issues/9190

Seatunnel + Debezium parameters does not work properly

snapshot.select.statement.overrides DOES NOT WORK
column.include.list WORKS

**column.include.list WORKS**

```
column.include.list="CONSINCO.GE_REDE.ROWID,CONSINCO.GE_REDE.SEQREDE,CONSINCO.GE_REDE.DESCRICAO,CONSINCO.GE_REDE.VLRLIMITECREDITO,CONSINCO.GE_REDE.DTAALTERACAO,CONSINCO.GE_REDE.USUALTERACAO,CONSINCO.GE_REDE.INDREPLICACAO,CONSINCO.GE_REDE.INDGEROUREPLIC,CONSINCO.GE_REDE.PERCANTECIP1,CONSINCO.GE_REDE.PERCANTECIP2,CONSINCO.GE_REDE.VLRLIMITETOTALPERIODO"
```

**Tried also without database-name**

```
snapshot.select.statement.overrides = "CONSINCO.GE_REDE"
snapshot.select.statement.overrides.CONSINCO.GE_REDE = "SELECT ROWID, SEQREDE, DESCRICAO, VLRLIMITECREDITO, DTAALTERACAO, INDREPLICACAO, INDGEROUREPLIC, PERCANTECIP1, PERCANTECIP2, VLRLIMITETOTALPERIODO FROM CONSINCO.GE_REDE"
```

Thank you!

Bruno

### SeaTunnel Version

2.3.11

### SeaTunnel Config

```conf
env {
  parallelism = 1
  job.mode = "STREAMING"
  log_level = "DEBUG"
  checkpoint.interval = 120000
  checkpoint.timeout = 120000
}

source {
  Oracle-CDC {
    plugin_output = "ORA-GE_REDE"
    username = "proj"
    password = "???????????"
    database-names = ["CONSINCO"]
    schema-names = ["CONSINCO"]
    table-names = ["CONSINCO.CONSINCO.GE_REDE"]
    base-url = "jdbc:oracle:thin:@MYORACLEHOST:1521/CONSINCO"
    driver = "oracle.jdbc.OracleDriver"
    source.reader.close.timeout = 120000
    startup.mode = "initial"

    debezium = {
      database.port = 1521
      rac.nodes = "192.168.1.116,192.168.1.117"
      snapshot.mode = "initial"
      heartbeat.interval.ms = 3000
      
      snapshot.select.statement.overrides = "CONSINCO.CONSINCO.GE_REDE"
      snapshot.select.statement.overrides.CONSINCO.CONSINCO.GE_REDE = "SELECT ROWID, SEQREDE, DESCRICAO, VLRLIMITECREDITO, DTAALTERACAO, INDREPLICACAO, INDGEROUREPLIC, PERCANTECIP1, PERCANTECIP2, VLRLIMITETOTALPERIODO FROM CONSINCO.GE_REDE"
    }
  }
}
```

### Running Command

```shell
./bin/seatunnel.sh --config ./config/orach.GE_REDE.conf -n GE_REDE
```

### Error Exception

```log
no error
data is replicated without respecting the snapshot.select.statement.overrides
```

Zeta

### Java or Scala Version

openjdk version "11.0.28" 2025-07-15

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:

Closes #9754.

Oracle CDC custom snapshot reader generated split queries directly, so the snapshot.select.statement.overrides setting was ignored during incremental snapshots.

This change reads the configured override, including Debezium’s catalog-less table lookup, wraps it as an inline view, and then applies SeaTunnel split boundary predicates. The custom filter is preserved without breaking parallel snapshot reads.

Yes. Oracle CDC now honors snapshot.select.statement.overrides during snapshot split reads. The English and Chinese connector documentation include an example and constraints for the custom query.

## How was this tested?

- Added a regression test for an override whose configured table name omits the Oracle catalog.
- Ran Oracle CDC module tests.
- Ran code formatting.
- Ran full repository verification with tests skipped.


DanielLeens (commented):

Oracle CDC's incremental (chunked) snapshot reader built every split's `SELECT`
statement purely from the physical `TableId`, never consulting Debezium's
`snapshot.select.statement.overrides` / `snapshot.select.statement.overrides.<table>`
settings. As a result, a user who configured a custom snapshot query (to add a
filter, mask a column, or work around an unsupported type) had that
configuration silently ignored during the actual snapshot data read — the
full, unfiltered table was scanned instead. This closes #9754, where the
reporter showed the exact symptom: `column.include.list` worked but
`snapshot.select.statement.overrides` did not, even for a `SCHEMA.TABLE`
(catalog-less) override key.

This PR adds `OracleUtils.buildSnapshotSplitScanQuery(...)`, wires it into
`OracleSnapshotSplitReadTask.createDataEventsForTable(...)` in place of the
override-blind `buildSplitScanQuery(...)`, and refactors the shared
split-boundary-condition logic into a reusable private helper so both the
overridden and non-overridden paths produce identical boundary predicates.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Before** (`OracleSnapshotSplitReadTask.java:208-209`, pre-PR):
```java
final String selectSql =
        OracleUtils.buildSplitScanQuery(
                snapshotSplit.getTableId(),
                snapshotSplit.getSplitKeyType(),
                snapshotSplit.getSplitStart() == null,
                snapshotSplit.getSplitEnd() == null);
```
`buildSplitScanQuery` never receives `connectorConfig`, so there was no way
for it to know about any override — this is a real, verified bug, not just a
documentation gap.

**After** (`OracleSnapshotSplitReadTask.java:208-214`):
```java
final String selectSql =
        OracleUtils.buildSnapshotSplitScanQuery(
                connectorConfig,
                snapshotSplit.getTableId(),
                snapshotSplit.getSplitKeyType(),
                snapshotSplit.getSplitStart() == null,
                snapshotSplit.getSplitEnd() == null);
```

`OracleUtils.buildSnapshotSplitScanQuery` (`OracleUtils.java:265-284`):
```java
String overriddenSelect = connectorConfig.getSnapshotSelectOverridesByTable().get(tableId);
if (overriddenSelect == null) {
    overriddenSelect = connectorConfig.getSnapshotSelectOverridesByTable()
            .get(new TableId(null, tableId.schema(), tableId.table()));
}
if (overriddenSelect == null) {
    return buildSplitScanQuery(tableId, rowType, isFirstSplit, isLastSplit);
}
String condition = buildSplitCondition(rowType, isFirstSplit, isLastSplit, true);
if (condition == null) {
    return overriddenSelect;
}
return String.format("SELECT * FROM (%s) WHERE %s", overriddenSelect, condition);
```

I traced this precisely because the lookup pattern (direct match, then a
`new TableId(null, tableId.schema(), tableId.table())` fallback) looks
arbitrary at first glance. It is not — it is a verbatim reproduction of
Debezium's own `RelationalSnapshotChangeEventSource.determineSnapshotSelect`
(debezium-core 1.9.8.Final, `io/debezium/relational/RelationalSnapshotChangeEventSource.java:456-460`),
which SeaTunnel's Oracle connector does not otherwise go through (it implements
its own chunked/incremental snapshot algorithm instead of Debezium's built-in
one). That matters for correctness for a subtle reason:

- `RelationalDatabaseConnectorConfig.getSnapshotSelectOverridesByTable()`
  (`RelationalDatabaseConnectorConfig.java:793-807`) parses each override key
  with `TableId.parse(table)`, i.e. `useCatalogBeforeSchema = true`. For a
  2-part key like `"SCHEMA1.TABLE1"` (the natural way to write an Oracle
  override, and exactly what the linked issue #9754 used), this produces
  `TableId(catalog="SCHEMA1", schema=null, table="TABLE1")` — the schema ends
  up in the *catalog* slot, not the *schema* slot.
- The real `TableId` used for an Oracle split is constructed elsewhere as
  `new TableId(database, schemaName, tableName)` (see
  `OracleConnectionUtils.java:148`, `OracleDialect.java:98`), i.e.
  `catalog=database`, `schema=owner`. A direct `.get(tableId)` therefore can
  never match a 2-part user override.
- `TableId`'s `id`/`equals`/`hashCode` are all derived from a private
  `tableId(catalog, schema, table)` formatter
  (`debezium-core` `TableId.java:235-246`) that collapses to `"schema.table"`
  whenever *either* catalog-is-null-schema-set or catalog-set-schema-null —
  i.e. `TableId("SCHEMA1", null, "TABLE1").id` and
  `TableId(null, "SCHEMA1", "TABLE1").id` are both `"SCHEMA1.TABLE1"` and thus
  `.equals()` true. That is exactly why the fallback lookup
  `new TableId(null, tableId.schema(), tableId.table())` recovers the match.

I verified this end-to-end against the new test
`OracleUtilsTest.testSnapshotSplitScanQueryUsesSelectOverride`
(`OracleUtilsTest.java:79-105`): it configures the override with key
`"schema1.table1"` while the real split's `TableId` is
`TableId.parse("cdb1.schema1.table1")` (catalog `cdb1`, schema `schema1`), and
asserts the correct wrapped SQL is produced — this is precisely the
catalog-vs-schema mismatch scenario described above, and precisely the
scenario reported in #9754. I consider the core fix technically sound and
faithful to Debezium's own established semantics, not an invented heuristic.

**Split-condition refactor.** `buildSplitCondition` (`OracleUtils.java:314-338`)
is a pure extraction of the branch logic that previously lived inline in
`buildSplitQuery` (`isFirstSplit && isLastSplit` → null; `isFirstSplit` →
`<=` [+ `NOT (=)`]; `isLastSplit` → `>=`; else → `>=` [+ `NOT (=)`] `AND` `<=`).
I compared the before/after control flow character-by-character; it is
behavior-preserving. This matters because it is the mechanism that keeps the
non-override path (still `buildSplitQuery` → `buildSplitCondition`) and the
new override path (`buildSnapshotSplitScanQuery` → `buildSplitCondition`)
byte-identical in how they express split boundaries, and it is why the four
pre-existing assertions in `testSplitScanQuery` needed zero changes.

**Runtime path confirmation.** `OracleSnapshotSplitReadTask.createDataEventsForTable`
is the only call site of the changed logic
(`OracleSnapshotSplitReadTask.java:191-266`), and it is the method invoked
once per split during snapshot phase 2 (`doExecute` → `createDataEvents` →
`createDataEventsForTable`, `OracleSnapshotSplitReadTask.java:140-141`). This
is squarely on the normal, default snapshot-read path for every Oracle CDC job
that uses the incremental snapshot algorithm — not a boundary/recovery path.

**Config plumbing sanity check.** I confirmed a user actually can set this via
the documented `debezium { snapshot.select.statement.overrides = ... }` block:
`OracleSourceConfigFactory.java:145-155` merges the parsed `debezium {}` block
(`dbzProperties`) into the final `Properties` via `props.putAll(dbzProperties)`,
which is what backs the `OracleConnectorConfig` instance seen at read time. So
the plumbing was never the problem — only the query-construction step was.

**One orphaned/duplicate path, left untouched (informational, not a
blocker).** `OracleChunkSplitter.buildSplitScanQuery(Table, ...)`
(`OracleChunkSplitter.java:91-93`) still calls the override-blind
`OracleUtils.buildSplitScanQuery` directly. I checked whether this is a second
live snapshot-read path that would still bypass the fix — it is not.
`JdbcSourceChunkSplitter.buildSplitScanQuery` (the interface it implements,
`JdbcSourceChunkSplitter.java:157-158`) has no callers anywhere in
`connector-cdc-base` or any of the CDC connector modules except this override
itself and the equivalent MySQL/Postgres/SQLServer `*ChunkSplitter`
implementations, which are likewise never invoked — each connector's own
`*SnapshotSplitReadTask` calls its own `*Utils.buildSplitScanQuery` statically
instead. This appears to be dead/vestigial interface surface predating this
PR, not a second bug this PR needs to close. Worth a follow-up cleanup someday,
but out of scope here.

## 1.2 Compatibility Impact

**Fully compatible** for the default/majority case (no override configured).
`getSnapshotSelectOverridesByTable()` returns `Collections.emptyMap()` when
`snapshot.select.statement.overrides` is unset, so both lookups in
`buildSnapshotSplitScanQuery` fall through to the unchanged
`buildSplitScanQuery(tableId, rowType, isFirstSplit, isLastSplit)`, which now
routes through the behavior-preserving `buildSplitCondition` refactor. I
verified this is a pure extraction (see 1.1), and the four pre-existing
`testSplitScanQuery` assertions were left untouched by the diff, which is
consistent with genuinely unchanged output.

**One real, narrow compatibility edge case worth flagging (not blocking, but
should be acknowledged in the PR discussion).** Because `snapshot.select.statement.overrides`
was previously a no-op for Oracle CDC snapshot reads, no currently-running job
could have been depending on the old (broken) "override ignored" behavior in
a way that matters — the feature simply never worked. However, for a job that
*is* configured with an override and is mid-snapshot (some splits already
read, checkpoint/savepoint holds the remaining split list) at the moment the
connector jar is upgraded to include this fix, resuming that job would apply
the override to the *remaining* unread splits while the *already-read* splits
were read from the unfiltered full table — i.e. within a single job's
snapshot output, some rows honor the override and some don't. This is
extremely narrow (requires: override configured + upgrade landing mid-snapshot
of a large enough table to span the upgrade), and there is no clean way for
this PR alone to avoid it, but I'd recommend calling it out explicitly in the
PR description or docs migration notes so operators upgrading a long-running
snapshot are not surprised.

`connectorConfig.getSnapshotSelectOverridesByTable()` reparses the Debezium
`Configuration` into a fresh `HashMap` on every call (per Debezium's own
implementation) and is invoked once per split (not per row), so the added
overhead is negligible even for tables with a very large number of splits. No
new resource lifecycle (statements, connections) beyond what already existed
in `createDataEventsForTable`; the `PreparedStatement` binding logic in
`readTableSplitDataStatement` (`OracleUtils.java:346-381`) still binds
parameters purely by placeholder position/count derived from
`splitKeyType.getTotalFields()`, which the override path does not change —
so parameter binding remains correct regardless of how many/what columns the
override's inline view exposes.

No new exception paths are introduced. See Issues #1 and #2 below for the two
things I'd like addressed or at least explicitly acknowledged before merge.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- `buildSnapshotSplitScanQuery` has a clear, multi-line Javadoc explaining the
  override-wrapping behavior (`OracleUtils.java:258-264`) — good.
- The newly extracted private helper `buildSplitCondition`
  (`OracleUtils.java:314`) has no Javadoc. It's a straightforward extraction
  of pre-existing inline logic with no new behavior, so this is a very minor,
  non-blocking nit (Issue #3 below) rather than a real gap.
- No wildcard imports, license headers unaffected (no new files), formatting
  matches project style; the fork's "Code style" CI job passed.

One new unit test, `testSnapshotSplitScanQueryUsesSelectOverride`
(`OracleUtilsTest.java:79-105`), which is a deterministic, no-I/O test
(pure string assertion against `OracleUtils.buildSnapshotSplitScanQuery`, no
threads, no sleeps, no shared state, no ordering dependency).
**Stability rating: Stable.**

That said, this only proves the generated SQL *string* is what we expect — it
does not exercise the query against a real Oracle instance. Given that the
whole point of this feature is "does the override actually execute and come
back correctly wrapped as an inline view against Oracle," and given that this
repo already has a working Oracle Testcontainers IT suite
(`seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-oracle-e2e`) with an
existing `DEBEZIUM.FULL_TYPES` fixture table already reused across several
`.conf` scenarios (e.g. `oraclecdc_to_oracle.conf`,
`oraclecdc_to_oracle_with_custom_primary_key.conf`), I'd flag the lack of a
`snapshot.select.statement.overrides` IT scenario as a real, if non-blocking,
coverage gap (Issue #4 below) — this is exactly the class of change where a
unit test can pass while the real DB rejects the generated SQL (e.g. an
Oracle-specific syntax issue with the inline-view wrap, or the override's
result-set column order/casing not lining up with `ColumnUtils.toArray`/
`rowToArray` against the physical table schema in
`OracleSnapshotSplitReadTask.createDataEventsForTable`, `OracleUtils.java:233`).
The PR description does say "Ran Oracle CDC module tests," which is UT-level;
I did not see a new IT case in the diff for this specific config.

## 2.3 Documentation Updates

Both `docs/en/connectors/source/Oracle-CDC.md` and `docs/zh/connectors/source/Oracle-CDC.md`
were updated with a matching new FAQ entry and example, and importantly the
English doc explicitly calls out the constraint that matters most for
correctness: *"the query is applied before SeaTunnel adds the snapshot-split
boundaries, so it must include every column needed by the configured table
schema and its split key."* That is the right thing to document given the
`SELECT * FROM (override) WHERE <boundary>` wrapping strategy. Good.

One documentation gap: neither doc explains what happens for a table whose
split key is `ROWID` (Oracle's fallback splitting strategy for
primary-key-less tables, evidenced by the `ROWID`-aware comparator in
`OracleChunkSplitter.java:102-103`). Wrapping an override as an inline view
and then filtering on `"ROWID" >= ?` would not work the way users might
expect, because Oracle's `ROWID` pseudo-column is not reliably projected
through an arbitrary derived table the way a real table's `ROWID` is. This is
a genuine edge case, not necessarily something this PR must solve, but at
minimum worth a one-line caveat in the same FAQ entry (Issue #5, Low).

# 3. Architectural Soundness

**Precise fix.** It reuses Debezium's own established lookup semantics
(direct match, then catalog-stripped fallback) rather than inventing new
matching rules, and reuses the exact same boundary-condition builder used by
the non-override path via the `buildSplitCondition` extraction, so the two
paths cannot silently drift apart in how they express split predicates.

## 3.2 Maintainability

Good. The refactor reduces duplication (condition-building logic previously
existed as three near-identical inline `StringBuilder` blocks; now it's one
shared helper called from two places).

## 3.3 Extensibility

Worth noting for context, not as a request to expand this PR's scope: I
checked, and MySQL, Postgres, and SQL Server CDC connectors in this repo have
the exact same gap — their respective `*SnapshotSplitReadTask` classes call
`*Utils.buildSplitScanQuery(tableId, ...)` without ever consulting
`connectorConfig.getSnapshotSelectOverridesByTable()` either
(`MySqlSnapshotSplitReadTask.java:195`,
`PostgresSnapshotSplitReadTask.java:214`, and the analogous SQL Server call).
Since issue #9754 was Oracle-specific and this PR is correctly scoped to
Oracle only, I'm not treating this as a blocker for this PR — just flagging it
as a natural, well-defined follow-up for the other three connectors, using
this PR's fallback-lookup pattern as the template.

No config option names, defaults, serialization formats, or checkpoint/split
state shapes changed. `SnapshotSplit`, `RedoLogOffset`, and the watermark
dispatch sequence in `OracleSnapshotSplitReadTask.doExecute` are all untouched.
See 1.2 for the one narrow mid-snapshot-upgrade edge case, which is
informational rather than a compatibility break (the feature was previously
non-functional, so there is no working prior behavior being changed).

# 4. Issue Summary

| 1 | Mid-snapshot connector upgrade with an override configured can produce a single job run where some already-read splits used the (previously broken) unfiltered query and later splits use the fixed override query. Not fixable by this PR alone; should be called out explicitly for operators. | `OracleSnapshotSplitReadTask.java:191-266` | Low |
| 2 | ROWID-split tables (no primary key) combined with a configured override are not covered by tracing or docs; wrapping an override as an inline view and filtering on `"ROWID"` may not behave as users expect. | `OracleUtils.java:265-284`, `OracleChunkSplitter.java:91-103` | Low |
| 3 | New private helper `buildSplitCondition` has no Javadoc (minor, pure extraction of pre-existing logic). | `OracleUtils.java:314` | Low |
| 4 | No new Testcontainers-based Oracle e2e scenario for `snapshot.select.statement.overrides`, despite an existing reusable `DEBEZIUM.FULL_TYPES` fixture already used by several `.conf` scenarios in `connector-cdc-oracle-e2e`. Only unit-level SQL-string assertions exist. | `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-oracle-e2e/src/test/resources/` | Medium |
| 5 | Docs don't state the ROWID-split caveat referenced in Issue #2. | `docs/en/connectors/source/Oracle-CDC.md`, `docs/zh/connectors/source/Oracle-CDC.md` | Low |

No blocking (High-severity) issues found. I traced the core lookup/wrapping
logic against Debezium's own upstream implementation and against the real
`TableId` shapes used elsewhere in this connector, and it is correct for the
documented, common case (schema-qualified override keys), and compatible with
all existing non-override jobs.

# 5. Merge Recommendation

This PR is not currently blocked by any of the issues above (all are Low or
Medium, non-functional-bug severity), but I'd like to see Issue #4 (e2e
coverage) addressed before merge given how easy it would be to add given the
existing `DEBEZIUM.FULL_TYPES` fixture, and Issues #1/#2/#5 at least
acknowledged (a one-line doc caveat each is enough) since they're cheap to add
and prevent future support tickets. None of this requires touching the core
fix, which I'm confident is correct.

1. **Recommended before merge:**
   - Add (or explicitly defer with a linked follow-up issue) an e2e
     `snapshot.select.statement.overrides` scenario against the existing
     `DEBEZIUM.FULL_TYPES` fixture (Issue #4).
2. **Non-blocking, nice to have:**
   - One-line doc caveat about ROWID-split tables + overrides (Issue #5).
   - One-line doc/PR-description caveat about the mid-snapshot-upgrade edge
     case (Issue #1).
   - Javadoc on `buildSplitCondition` (Issue #3).

Overall: solid, well-scoped bug fix that correctly reproduces Debezium's own
override-resolution semantics for a case (Oracle's catalog/schema slot
mismatch on 2-part override keys) that is genuinely easy to get wrong, and is
backed by a targeted regression test for exactly that scenario. Nice work.

CI on the fork (`zhangshenghang/seatunnel`, run 31502116126) was still
completing two jobs (`unit-test (11, ubuntu-latest)`, `unit-test (8,
windows-latest)`) at review time, with every other job — including
`oracle-cdc-connector-it` on both JDK 8 and 11 — green. The Apache-side
"Build" check showing red is the known stale-pointer behavior for
fork-PR-triggered workflows; it does not reflect an actual failure. Worth
re-checking once the fork run finishes.

DanielLeens (commented):
Thanks for the update! Quick housekeeping note before the review: I checked for any new comments/reviews from other reviewers since my last pass, and there aren't any — the only thing that changed since I last looked at this PR is a new commit (`cd18bc3a`) that merges `upstream/dev` into this branch. I diffed `c3a32f3e` (what I reviewed previously) against `cd18bc3a` and confirmed none of this PR's own files (`OracleUtils.java`, `OracleSnapshotSplitReadTask.java`, `OracleUtilsTest.java`, or the two docs files) changed — the merge only pulled in unrelated upstream content. So this is a re-verification pass on an unchanged diff rather than a review of new logic; I re-read every changed file from scratch against the new head SHA to confirm that, and I'm restating my findings below with the current head as the reference point (all `path:line` references now point at `cd18bc3a`).

- Oracle CDC's incremental (chunked) snapshot reader built every split's `SELECT` purely from the physical `TableId`, never consulting Debezium's `snapshot.select.statement.overrides` / `snapshot.select.statement.overrides.<table>` settings — a user-configured custom snapshot query (filter, masked column, workaround for an unsupported type) was silently ignored during the actual snapshot read.
- Fix: add `OracleUtils.buildSnapshotSplitScanQuery(...)`, wire it into `OracleSnapshotSplitReadTask.createDataEventsForTable(...)` in place of the override-blind `buildSplitScanQuery(...)`, and extract the shared split-boundary-condition logic into a reusable `buildSplitCondition` helper so both the overridden and non-overridden paths produce identical boundary predicates.
- One-sentence summary: this makes Oracle CDC's chunked snapshot reads actually honor `snapshot.select.statement.overrides`, closing #9754, without changing behavior for jobs that don't configure it.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Before** (pre-PR, `OracleSnapshotSplitReadTask.java`):
```java
final String selectSql =
        OracleUtils.buildSplitScanQuery(
                snapshotSplit.getTableId(),
                snapshotSplit.getSplitKeyType(),
                snapshotSplit.getSplitStart() == null,
                snapshotSplit.getSplitEnd() == null);
```
`buildSplitScanQuery` never receives `connectorConfig`, so there was no way for it to know about any override.

**After** (`OracleSnapshotSplitReadTask.java:208-214`):
```java
final String selectSql =
        OracleUtils.buildSnapshotSplitScanQuery(
                connectorConfig,
                snapshotSplit.getTableId(),
                snapshotSplit.getSplitKeyType(),
                snapshotSplit.getSplitStart() == null,
                snapshotSplit.getSplitEnd() == null);
```

`OracleUtils.buildSnapshotSplitScanQuery` (`OracleUtils.java:265-287`):
```java
String overriddenSelect = connectorConfig.getSnapshotSelectOverridesByTable().get(tableId);
if (overriddenSelect == null) {
    overriddenSelect = connectorConfig.getSnapshotSelectOverridesByTable()
            .get(new TableId(null, tableId.schema(), tableId.table()));
}
if (overriddenSelect == null) {
    return buildSplitScanQuery(tableId, rowType, isFirstSplit, isLastSplit);
}
String condition = buildSplitCondition(rowType, isFirstSplit, isLastSplit, true);
if (condition == null) {
    return overriddenSelect;
}
return String.format("SELECT * FROM (%s) WHERE %s", overriddenSelect, condition);
```

The two-step lookup (direct match, then a catalog-stripped `new TableId(null, tableId.schema(), tableId.table())` fallback) is a verbatim reproduction of Debezium's own `RelationalSnapshotChangeEventSource.determineSnapshotSelect`. It matters for correctness because `RelationalDatabaseConnectorConfig.getSnapshotSelectOverridesByTable()` parses each override key with `TableId.parse(table)` (`useCatalogBeforeSchema = true`), so a natural 2-part override key like `"SCHEMA1.TABLE1"` (exactly what #9754 used) lands in `TableId(catalog="SCHEMA1", schema=null, table="TABLE1")` — schema in the catalog slot. Meanwhile the real split's `TableId` is built as `new TableId(database, schemaName, tableName)` elsewhere in the connector, i.e. `catalog=database, schema=owner`. A direct `.get(tableId)` can never match a 2-part key; the fallback recovers it because Debezium's `TableId.id()` collapses both catalog-null-schema-set and catalog-set-schema-null shapes to the same `"schema.table"` string, making the two `TableId` instances `.equals()`.

This is exercised by the new unit test `OracleUtilsTest.testSnapshotSplitScanQueryUsesSelectOverride` (`OracleUtilsTest.java:196-220`), which configures the override under key `"schema1.table1"` while the real split's `TableId` is `TableId.parse("cdb1.schema1.table1")` (catalog `cdb1`, schema `schema1`) — precisely the catalog/schema mismatch scenario above — and asserts the correctly wrapped SQL. I re-ran this trace end to end against the current head and it holds exactly as before.

**Split-condition refactor** (`OracleUtils.java:314-344`): `buildSplitCondition` is a pure extraction of the branch logic that previously lived inline in `buildSplitQuery` (`isFirstSplit && isLastSplit` → `null`; `isFirstSplit` → `<=` [+ `NOT (=)`]; `isLastSplit` → `>=`; else → `>=` [+ `NOT (=)`] `AND` `<=`). Comparing before/after control flow line by line, it is behavior-preserving, and it's the reason the four pre-existing `testSplitScanQuery` assertions needed zero changes — both the override path and the non-override path now share the exact same boundary-predicate logic and can't drift apart.

**Runtime path.** `OracleSnapshotSplitReadTask.createDataEventsForTable` is the only call site of the changed logic, invoked once per split during phase-2 snapshot reads (`doExecute` → `createDataEvents` → `createDataEventsForTable`). This is the default snapshot-read path for every Oracle CDC job using the incremental snapshot algorithm — not a recovery/boundary-only path. Config plumbing: `OracleSourceConfigFactory` merges the parsed `debezium { ... }` block into the `Properties` backing `OracleConnectorConfig`, so a user really can drive this via the documented `debezium { snapshot.select.statement.overrides = ... }` block — the read-time query construction was the only broken link.

**Key Findings:**
- The fix is scoped exactly to the reported bug (#9754): override lookup now happens where it's actually needed, using Debezium's own established (not invented) matching semantics.
- The `buildSplitCondition` extraction is behavior-preserving and is what guarantees the override and non-override paths can't silently diverge in how they express split boundaries — a good design choice, not just incidental refactoring.
- `OracleChunkSplitter.buildSplitScanQuery(Table, ...)` (`OracleChunkSplitter.java:91-93`) still calls the override-blind `OracleUtils.buildSplitScanQuery` directly, but I confirmed `JdbcSourceChunkSplitter.buildSplitScanQuery` has no live callers in this codebase outside each connector's own `*SnapshotSplitReadTask` (which call their `*Utils` methods statically instead) — this is dead/vestigial interface surface predating the PR, not a second bug this PR needs to close.
- PreparedStatement parameter binding in `readTableSplitDataStatement` (`OracleUtils.java:346+`) still derives placeholder count purely from `splitKeyType.getTotalFields()`, independent of how many columns the override's inline view exposes — so parameter binding stays correct regardless of the override's column list.
- MySQL, Postgres, and SQL Server CDC connectors have the identical gap (their `*SnapshotSplitReadTask`s never consult `getSnapshotSelectOverridesByTable()` either); out of scope for this Oracle-only PR, worth a follow-up.

## 1.2 Compatibility Impact

**Fully compatible** for the default/majority case (no override configured): `getSnapshotSelectOverridesByTable()` returns an empty map when `snapshot.select.statement.overrides` is unset, so both lookups fall through to the unchanged `buildSplitScanQuery(...)`, which now routes through the behavior-preserving `buildSplitCondition` extraction — confirmed by the untouched pre-existing `testSplitScanQuery` assertions.

One narrow, non-blocking edge case worth acknowledging: since the override was previously a no-op, no running job could have depended on the old "override ignored" behavior. But a job that *is* configured with an override and is mid-snapshot at the moment the connector jar is upgraded (checkpoint holds the remaining split list) would apply the override only to not-yet-read splits — already-read splits came from the unfiltered full table. This is inherent to fixing a previously-broken feature mid-flight and isn't something this PR alone can avoid; worth a one-line callout for operators upgrading a long-running snapshot.

`getSnapshotSelectOverridesByTable()` reparses the Debezium `Configuration` into a fresh map on every call, but it's invoked once per split, not per row — negligible overhead even for large split counts. No new resource lifecycle (statements/connections) beyond what already existed in `createDataEventsForTable`.

No new exception paths introduced. See Issues #1 and #2 in the summary table.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- `buildSnapshotSplitScanQuery` has a clear Javadoc explaining the override-wrapping behavior (`OracleUtils.java:258-264`).
- The new private helper `buildSplitCondition` (`OracleUtils.java:314`) has no Javadoc. It's a pure extraction of pre-existing inline logic with no new behavior, so this is a minor, non-blocking nit (Issue #3).
- No wildcard imports (verified against the file at the current head), license header unaffected, formatting matches project style.

One new unit test, `testSnapshotSplitScanQueryUsesSelectOverride` (`OracleUtilsTest.java:196-220`): deterministic, no I/O, pure string assertion, no threads/sleeps/shared state/ordering dependency. **Stability rating: Stable.**

It only proves the generated SQL string is correct, not that the override actually executes against real Oracle and comes back correctly through `ColumnUtils.toArray`/`rowToArray`. Given this repo already has a working Oracle Testcontainers IT suite (`connector-cdc-oracle-e2e`) with a reusable `DEBEZIUM.FULL_TYPES` fixture used by multiple `.conf` scenarios, the lack of an e2e scenario for `snapshot.select.statement.overrides` is a real, non-blocking coverage gap (Issue #4) — exactly the class of change where a unit test can pass while Oracle rejects the generated SQL or a column-order mismatch slips through.

## 2.3 Documentation Updates

Both `docs/en/connectors/source/Oracle-CDC.md` and `docs/zh/connectors/source/Oracle-CDC.md` got a matching new FAQ entry, and the English doc explicitly documents the correctness-critical constraint: the override query is applied before SeaTunnel adds split boundaries, so it must include every column needed by the table schema and split key. Good.

Gap: neither doc addresses `ROWID`-split tables (Oracle's fallback strategy for primary-key-less tables, per the `ROWID`-aware comparator in `OracleChunkSplitter.java:102-103`). Wrapping an override as an inline view and then filtering on `"ROWID" >= ?` is unlikely to behave as users expect, since Oracle's `ROWID` pseudo-column isn't reliably projected through an arbitrary derived table the way it is for a real table. Worth a one-line caveat in the same FAQ entry (Issue #5, Low).

# 3. Architectural Soundness

**Precise fix.** Reuses Debezium's own established lookup semantics rather than inventing new matching rules, and reuses the same boundary-condition builder as the non-override path via the `buildSplitCondition` extraction, so the two paths cannot silently drift apart.

## 3.2 Maintainability
Good — the refactor removes duplication (three near-identical inline `StringBuilder` condition blocks collapse into one shared helper called from two places).

## 3.3 Extensibility
MySQL, Postgres, and SQL Server CDC connectors have the same gap in their respective `*SnapshotSplitReadTask`s. Not a request to expand this PR's scope — #9754 was Oracle-specific and this PR is correctly scoped — but a natural follow-up using this PR's fallback-lookup pattern as the template.

No config option names, defaults, serialization formats, or checkpoint/split state shapes changed. `SnapshotSplit`, `RedoLogOffset`, and the watermark dispatch sequence in `OracleSnapshotSplitReadTask.doExecute` are untouched. The one narrow mid-snapshot-upgrade edge case from 1.2 is informational, not a compatibility break, since the feature was previously non-functional.

# 4. Issue Summary

| 1 | Mid-snapshot connector upgrade with an override configured can produce a single job run where already-read splits used the old unfiltered query and later splits use the fixed override query. Not fixable by this PR alone; should be called out for operators. | `OracleSnapshotSplitReadTask.java:191-266` | Low |
| 2 | ROWID-split tables (no primary key) combined with a configured override aren't covered by tracing or docs; wrapping an override as an inline view and filtering on `"ROWID"` may not behave as expected. | `OracleUtils.java:265-287`, `OracleChunkSplitter.java:91-103` | Low |
| 3 | New private helper `buildSplitCondition` has no Javadoc (minor, pure extraction of pre-existing logic). | `OracleUtils.java:314` | Low |
| 4 | No Testcontainers-based Oracle e2e scenario for `snapshot.select.statement.overrides`, despite an existing reusable `DEBEZIUM.FULL_TYPES` fixture already used by several `.conf` scenarios in `connector-cdc-oracle-e2e`. Only unit-level SQL-string assertions exist. | `seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-oracle-e2e/src/test/resources/` | Medium |
| 5 | Docs don't state the ROWID-split caveat referenced in Issue #2. | `docs/en/connectors/source/Oracle-CDC.md`, `docs/zh/connectors/source/Oracle-CDC.md` | Low |

No High-severity issues. The core lookup/wrapping logic traces correctly against Debezium's own upstream implementation and the real `TableId` shapes used elsewhere in this connector, is correct for the documented common case, and is compatible with all existing non-override jobs. This assessment is unchanged from my prior pass since the diff itself is unchanged.

# 5. Merge Recommendation

Not blocked by any of the issues above (all Low or Medium, none functional-bug severity), but I'd still like to see Issue #4 (e2e coverage) addressed before merge given how easy it would be to add on top of the existing `DEBEZIUM.FULL_TYPES` fixture, and Issues #1/#2/#5 at least acknowledged (a one-line doc caveat each is enough). None of this requires touching the core fix, which I remain confident is correct.

1. **Recommended before merge:**
   - Add (or explicitly defer with a linked follow-up issue) an e2e `snapshot.select.statement.overrides` scenario against the existing `DEBEZIUM.FULL_TYPES` fixture (Issue #4).
2. **Non-blocking, nice to have:**
   - One-line doc caveat about ROWID-split tables + overrides (Issue #5).
   - One-line doc/PR-description caveat about the mid-snapshot-upgrade edge case (Issue #1).
   - Javadoc on `buildSplitCondition` (Issue #3).

Overall: solid, well-scoped bug fix that correctly reproduces Debezium's own override-resolution semantics for a genuinely easy-to-get-wrong case (Oracle's catalog/schema slot mismatch on 2-part override keys), backed by a targeted regression test for exactly that scenario.

On CI: the head commit `c3a32f3e` (the substantive commit; everything I analyzed above is byte-identical between it and the current head) had a fully green fork run (`zhangshenghang/seatunnel` run 31502116126, including `oracle-cdc-connector-it` on JDK 8 and 11). The new merge commit `cd18bc3a` triggered its own fork `Build` run (31589363777) which is still queued as of this review and hasn't started yet — worth a quick re-check once it completes, though I don't expect it to surface anything new since it touches none of this PR's files. The Apache-side "Build" check showing pending/queued is the known stale-pointer behavior for fork-triggered workflows, not an actual failure.

DanielLeens (approved):
Quick status re-check on this PR. No new commit and no new discussion since my last review, but the required `Build` check was still queued/pending on the merge-commit head (`cd18bc3a`) at that time, and I flagged it as worth a re-check once it completed.

That run has since finished and completed **SUCCESS**. Since my last review already concluded "not blocked by any of the issues above" — the outstanding items were all Low/Medium recommendations, not blockers — CI passing was the only remaining gate. With that resolved, I'm updating my conclusion to approve.

The Oracle `snapshot.select.statement.overrides` catalog/schema lookup fix remains correct on re-verification (unchanged since my prior pass), and there's no outstanding functional blocker. The previously-noted non-blocking suggestions (e2e coverage for the override scenario, a couple of one-line doc caveats, a Javadoc note) are still worth picking up but shouldn't hold up the merge.

Thanks for the solid, well-scoped fix here.

davidzollo (approved):
Good job, LGTM

PR #11770, opened 2026-08-12
[Fix][Connector-V2] Close JDBC source read transactions after each split

wybaby168:

Fix JDBC source sessions that can remain `idle in transaction` after a source split has finished reading or has failed during initialization.

PostgreSQL and Redshift cursor reads disable auto-commit in their dialect-specific prepared-statement creation. `JdbcInputFormat.close()` currently closes only the `ResultSet` and `PreparedStatement`; closing those resources does not end a manual JDBC transaction. Because the source connection is reused across splits, the database backend can retain the old transaction snapshot while the reader is waiting for another split or while the rest of the job is being cleaned up.

On PostgreSQL-compatible databases this is visible as sessions with:

```text
state           = idle in transaction
wait_event_type = Client
wait_event      = ClientRead
```

The retained transaction can hold `backend_xmin`, delay vacuum/tuple cleanup, retain transaction-scoped locks, and consume connection slots.

This patch closes the transaction at the JDBC source split lifecycle boundary:

- capture the statement connection before closing JDBC resources;
- close and clear the result set and statement even when either close operation fails;
- roll back an open read transaction after a split is consumed or aborted;
- restore the user-configured auto-commit mode after a dialect temporarily changes it;
- clean up SQL and runtime failures raised while opening a split;
- discard the cached connection if statement creation fails before the statement can expose its connection;
- physically close and invalidate a connection when rollback or auto-commit restoration fails;
- make repeated `close()` calls idempotent;
- always close the splitter connection from `closeInputFormat()` through `finally`.

`rollback()` is intentional. JDBC source processing is read-only from the connector's perspective. Rolling back ends the server-side cursor transaction, releases its snapshot, and also clears a transaction that entered an aborted state. Cursor-based incremental fetching remains enabled; the patch does not force PostgreSQL/Redshift reads into auto-commit mode while the result set is active.

Yes, as a bug fix.

**Previous behavior:** PostgreSQL-compatible JDBC source sessions could remain in `idle in transaction` after a split completed, until the whole reader connection was eventually closed.

**New behavior:** the read transaction ends immediately at split cleanup. The configured auto-commit mode is restored before the connection is reused.

There is no configuration-key, default-value, or public-API change. JDBC Sink, XA, and exactly-once commit semantics are not modified.

Added `JdbcInputFormatTest` coverage for:

1. PostgreSQL-style cursor mode: runtime auto-commit changes from configured `true` to `false`; cleanup rolls back and restores `true`.
2. Explicit `auto_commit=false`: cleanup rolls back while preserving manual-commit mode.
3. Native auto-commit connections: no unnecessary rollback or mode change.
4. Result-set and statement close failures: transaction cleanup still executes.
5. Rollback failure: the physical connection is closed and the cached provider connection is invalidated.
6. Failure to obtain the connection from the statement: the cached connection is closed instead of being reused in an unknown transaction state.
7. Prepared-statement creation failure before assignment: the cached connection is discarded.
8. Query execution failure after statement creation: statement cleanup, rollback, mode restoration, and provider cleanup all execute.
9. Input-format shutdown: the splitter connection is always closed through `finally`.
10. Already-closed connections and repeated `close()` calls: cleanup is safe and idempotent.

The production and test sources were also checked for Java 8 compatibility. The connector-JDBC unit, formatting, license, and static checks are delegated to this PR's GitHub CI.

### Check list

* [x] Documentation changes are not required because this restores the existing bounded-source lifecycle contract.
* [x] No incompatible configuration or public API change is introduced.
* [x] No plugin mapping, distribution dependency, or connector registration change is required.
* [x] Unit tests cover normal, exceptional, and idempotent cleanup paths.

### DIFF
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormat.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormat.java
@@ -38,6 +38,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -57,6 +58,7 @@ public class JdbcInputFormat implements Serializable {
     private final JdbcRowConverter jdbcRowConverter;
     private final Map<TablePath, CatalogTable> tables;
     private final ChunkSplitter chunkSplitter;
+    private final boolean configuredAutoCommit;
 
     private transient String splitTableId;
     private transient TableSchema splitTableSchema;
@@ -65,23 +67,37 @@ public class JdbcInputFormat implements Serializable {
     private volatile boolean hasNext;
 
     public JdbcInputFormat(JdbcSourceConfig config, Map<TablePath, CatalogTable> tables) {
-        this.jdbcDialect =
+        this(
                 JdbcDialectLoader.load(
                         config.getJdbcConnectionConfig().getUrl(),
                         config.getJdbcConnectionConfig().getDialect(),
-                        config.getCompatibleMode());
-        this.chunkSplitter = ChunkSplitter.create(config);
+                        config.getCompatibleMode()),
+                ChunkSplitter.create(config),
+                tables,
+                config.getJdbcConnectionConfig().isAutoCommit());
+    }
+
+    JdbcInputFormat(
+            JdbcDialect jdbcDialect,
+            ChunkSplitter chunkSplitter,
+            Map<TablePath, CatalogTable> tables,
+            boolean configuredAutoCommit) {
+        this.jdbcDialect = jdbcDialect;
+        this.chunkSplitter = chunkSplitter;
         this.jdbcRowConverter = jdbcDialect.getRowConverter();
         this.tables = tables;
+        this.configuredAutoCommit = configuredAutoCommit;
     }
 
     public void openInputFormat() {}
 
     public void closeInputFormat() throws IOException {
-        close();
-
-        if (chunkSplitter != null) {
-            chunkSplitter.close();
+        try {
+            close();
+        } finally {
+            if (chunkSplitter != null) {
+                chunkSplitter.close();
+            }
         }
     }
 
@@ -101,10 +117,31 @@ public class JdbcInputFormat implements Serializable {
             resultSet = statement.executeQuery();
             hasNext = resultSet.next();
         } catch (SQLException se) {
+            cleanupAfterOpenFailure(se);
             throw new JdbcConnectorException(
                     JdbcConnectorErrorCode.CONNECT_DATABASE_FAILED,
                     "open() failed." + se.getMessage(),
                     se);
+        } catch (RuntimeException runtimeException) {
+            cleanupAfterOpenFailure(runtimeException);
+            throw runtimeException;
+        }
+    }
+
+    private void cleanupAfterOpenFailure(Throwable openException) {
+        boolean shouldDiscardConnection = statement == null;
+        try {
+            close();
+        } catch (IOException cleanupException) {
+            openException.addSuppressed(cleanupException);
+            shouldDiscardConnection = true;
+        } finally {
+            if (shouldDiscardConnection) {
+                // Statement creation may establish or mutate the cached connection before failing
+                // without returning a statement. Discard it because close() cannot identify and
+                // finish that transaction safely.
+                chunkSplitter.close();
+            }
         }
     }
 
@@ -114,11 +151,14 @@ public class JdbcInputFormat implements Serializable {
      * @throws IOException Indicates that a resource could not be closed.
      */
     public void close() throws IOException {
+        Connection connection = getStatementConnection();
         if (resultSet != null) {
             try {
                 resultSet.close();
             } catch (SQLException e) {
                 LOG.info("ResultSet couldn't be closed - " + e.getMessage());
+            } finally {
+                resultSet = null;
             }
         }
         if (statement != null) {
@@ -126,7 +166,86 @@ public class JdbcInputFormat implements Serializable {
                 statement.close();
             } catch (SQLException e) {
                 LOG.info("Statement couldn't be closed - " + e.getMessage());
+            } finally {
+                statement = null;
             }
+        }
+
+        hasNext = false;
+        splitTableSchema = null;
+        splitTableId = null;
+        finishReadTransaction(connection);
+    }
+
+    private Connection getStatementConnection() {
+        if (statement == null) {
+            return null;
+        }
+        try {
+            Connection connection = statement.getConnection();
+            if (connection == null) {
+                LOG.warn(
+                        "The JDBC source statement returned no connection. "
+                                + "Closing the cached connection to avoid reusing an unknown "
+                                + "transaction.");
+                chunkSplitter.close();
+            }
+            return connection;
+        } catch (SQLException e) {
+            LOG.warn(
+                    "Failed to get the JDBC source connection from the current statement. "
+                            + "Closing the cached connection to avoid reusing an unknown "
+                            + "transaction.",
+                    e);
+            chunkSplitter.close();
+            return null;
+        }
+    }
+
+    private void finishReadTransaction(Connection connection) {
+        try {
+            finishReadTransaction(connection, configuredAutoCommit);
+        } catch (SQLException e) {
+            LOG.warn(
+                    "Failed to finish the JDBC source read transaction. "
+                            + "Closing the connection to avoid leaving or reusing an idle "
+                            + "transaction.",
+                    e);
+            discardConnection(connection, e);
+        }
+    }
+
+    private void discardConnection(Connection connection, SQLException cleanupException) {
+        try {
+            if (connection != null) {
+                connection.close();
+            }
+        } catch (SQLException closeException) {
+            cleanupException.addSuppressed(closeException);
+            LOG.warn(
+                    "Failed to close the JDBC source connection after transaction cleanup failed.",
+                    cleanupException);
+        } finally {
+            // Clear the provider's cached reference and retry close for drivers whose first close
+            // attempt failed.
+            chunkSplitter.close();
+        }
+    }
+
+    static void finishReadTransaction(Connection connection, boolean configuredAutoCommit)
+            throws SQLException {
+        if (connection == null || connection.isClosed()) {
+            return;
+        }
+
+        boolean currentAutoCommit = connection.getAutoCommit();
+        if (!currentAutoCommit) {
+            // JDBC source reads do not have changes to commit. Rollback ends the server-side cursor
+            // transaction, releases its snapshot, and also recovers a transaction in failed state.
+            connection.rollback();
+        }
+        if (currentAutoCommit != configuredAutoCommit) {
+            connection.setAutoCommit(configuredAutoCommit);
         }
     }
 
--- a/seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormatTest.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/test/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormatTest.java
@@ -1,0 +1,272 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.connectors.seatunnel.jdbc.internal;
+
+import org.apache.seatunnel.api.table.catalog.CatalogTable;
+import org.apache.seatunnel.api.table.catalog.TablePath;
+import org.apache.seatunnel.api.table.catalog.TableSchema;
+import org.apache.seatunnel.connectors.seatunnel.jdbc.exception.JdbcConnectorException;
+import org.apache.seatunnel.connectors.seatunnel.jdbc.internal.converter.JdbcRowConverter;
+import org.apache.seatunnel.connectors.seatunnel.jdbc.internal.dialect.JdbcDialect;
+import org.apache.seatunnel.connectors.seatunnel.jdbc.source.ChunkSplitter;
+import org.apache.seatunnel.connectors.seatunnel.jdbc.source.JdbcSourceSplit;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.InOrder;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Collections;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class JdbcInputFormatTest {
+
+    private static final TablePath TABLE_PATH = TablePath.of("test", "public", "source_table");
+    private static final TableSchema TABLE_SCHEMA = TableSchema.builder().build();
+    private static final JdbcSourceSplit SPLIT =
+            new JdbcSourceSplit(TABLE_PATH, "split-0", null, null, null, null, null);
+
+    @Test
+    void shouldRollbackAndRestoreConfiguredAutoCommitAfterClosingSplit() throws Exception {
+        TestContext context = createContext(true);
+        context.openEmptySplit();
+        when(context.connection.isClosed()).thenReturn(false);
+        when(context.connection.getAutoCommit()).thenReturn(false);
+
+        context.inputFormat.close();
+
+        InOrder inOrder = inOrder(context.statement, context.resultSet, context.connection);
+        inOrder.verify(context.statement).getConnection();
+        inOrder.verify(context.resultSet).close();
+        inOrder.verify(context.statement).close();
+        inOrder.verify(context.connection).rollback();
+        inOrder.verify(context.connection).setAutoCommit(true);
+
+        context.inputFormat.close();
+        verify(context.connection, times(1)).rollback();
+    }
+
+    @Test
+    void shouldRollbackAndKeepConfiguredManualCommit() throws Exception {
+        TestContext context = createContext(false);
+        context.openEmptySplit();
+        when(context.connection.isClosed()).thenReturn(false);
+        when(context.connection.getAutoCommit()).thenReturn(false);
+
+        context.inputFormat.close();
+
+        verify(context.connection).rollback();
+        verify(context.connection, never()).setAutoCommit(true);
+    }
+
+    @Test
+    void shouldNotRollbackAutoCommitConnection() throws Exception {
+        TestContext context = createContext(true);
+        context.openEmptySplit();
+        when(context.connection.isClosed()).thenReturn(false);
+        when(context.connection.getAutoCommit()).thenReturn(true);
+
+        context.inputFormat.close();
+
+        verify(context.connection, never()).rollback();
+        verify(context.connection, never()).setAutoCommit(true);
+    }
+
+    @Test
+    void shouldFinishTransactionWhenResourceCloseFails() throws Exception {
+        TestContext context = createContext(true);
+        context.openEmptySplit();
+        when(context.connection.isClosed()).thenReturn(false);
+        when(context.connection.getAutoCommit()).thenReturn(false);
+        doThrow(new SQLException("result set close failed")).when(context.resultSet).close();
+        doThrow(new SQLException("statement close failed")).when(context.statement).close();
+
+        context.inputFormat.close();
+
+        verify(context.connection).rollback();
+        verify(context.connection).setAutoCommit(true);
+    }
+
+    @Test
+    void shouldKeepConnectionReusableAcrossSuccessfulSplits() throws Exception {
+        TestContext context = createContext(true);
+        PreparedStatement secondStatement = mock(PreparedStatement.class);
+        ResultSet secondResultSet = mock(ResultSet.class);
+
+        when(context.chunkSplitter.generateSplitStatement(SPLIT, TABLE_SCHEMA))
+                .thenReturn(context.statement, secondStatement);
+        when(context.statement.executeQuery()).thenReturn(context.resultSet);
+        when(secondStatement.executeQuery()).thenReturn(secondResultSet);
+        when(context.resultSet.next()).thenReturn(false);
+        when(secondResultSet.next()).thenReturn(false);
+        when(context.statement.getConnection()).thenReturn(context.connection);
+        when(secondStatement.getConnection()).thenReturn(context.connection);
+        when(context.connection.isClosed()).thenReturn(false);
+        when(context.connection.getAutoCommit()).thenReturn(false);
+
+        context.inputFormat.open(SPLIT);
+        context.inputFormat.close();
+        context.inputFormat.open(SPLIT);
+        context.inputFormat.close();
+
+        verify(context.chunkSplitter, times(2)).generateSplitStatement(SPLIT, TABLE_SCHEMA);
+        verify(context.connection, times(2)).rollback();
+        verify(context.connection, times(2)).setAutoCommit(true);
+        verify(context.connection, never()).close();
+        verify(context.chunkSplitter, never()).close();
+    }
+
+    @Test
+    void shouldDiscardConnectionWhenTransactionCleanupFails() throws Exception {
+        TestContext context = createContext(true);
+        context.openEmptySplit();
+        when(context.connection.isClosed()).thenReturn(false);
+        when(context.connection.getAutoCommit()).thenReturn(false);
+        doThrow(new SQLException("rollback failed")).when(context.connection).rollback();
+
+        context.inputFormat.close();
+
+        verify(context.connection).close();
+        verify(context.chunkSplitter).close();
+        verify(context.connection, never()).setAutoCommit(true);
+    }
+
+    @Test
+    void shouldCloseCachedConnectionWhenStatementCannotExposeConnection() throws Exception {
+        TestContext context = createContext(true);
+        context.openEmptySplit();
+        when(context.statement.getConnection())
+                .thenThrow(new SQLException("get connection failed"));
+
+        context.inputFormat.close();
+
+        verify(context.chunkSplitter).close();
+        verify(context.resultSet).close();
+        verify(context.statement).close();
+    }
+
+    @Test
+    void shouldDiscardCachedConnectionWhenStatementCreationFails() throws Exception {
+        TestContext context = createContext(true);
+        when(context.chunkSplitter.generateSplitStatement(SPLIT, TABLE_SCHEMA))
+                .thenThrow(new SQLException("prepare failed"));
+
+        assertThrows(JdbcConnectorException.class, () -> context.inputFormat.open(SPLIT));
+
+        verify(context.chunkSplitter).close();
+        verify(context.statement, never()).executeQuery();
+    }
+
+    @Test
+    void shouldRollbackAndKeepConnectionWhenExecuteQueryFails() throws Exception {
+        TestContext context = createContext(true);
+        when(context.chunkSplitter.generateSplitStatement(SPLIT, TABLE_SCHEMA))
+                .thenReturn(context.statement);
+        when(context.statement.getConnection()).thenReturn(context.connection);
+        when(context.statement.executeQuery()).thenThrow(new SQLException("execute failed"));
+        when(context.connection.isClosed()).thenReturn(false);
+        when(context.connection.getAutoCommit()).thenReturn(false);
+
+        assertThrows(JdbcConnectorException.class, () -> context.inputFormat.open(SPLIT));
+
+        verify(context.statement).close();
+        verify(context.connection).rollback();
+        verify(context.connection).setAutoCommit(true);
+        verify(context.connection, never()).close();
+        verify(context.chunkSplitter, never()).close();
+    }
+
+    @Test
+    void shouldAlwaysCloseCachedConnectionWhenInputFormatCloses() throws Exception {
+        TestContext context = createContext(true);
+
+        context.inputFormat.closeInputFormat();
+
+        verify(context.chunkSplitter).close();
+    }
+
+    @Test
+    void shouldIgnoreClosedConnection() throws SQLException {
+        Connection connection = mock(Connection.class);
+        when(connection.isClosed()).thenReturn(true);
+
+        JdbcInputFormat.finishReadTransaction(connection, true);
+
+        verify(connection, never()).getAutoCommit();
+        verify(connection, never()).rollback();
+    }
+
+    private static TestContext createContext(boolean configuredAutoCommit) throws SQLException {
+        JdbcDialect dialect = mock(JdbcDialect.class);
+        JdbcRowConverter rowConverter = mock(JdbcRowConverter.class);
+        ChunkSplitter chunkSplitter = mock(ChunkSplitter.class);
+        PreparedStatement statement = mock(PreparedStatement.class);
+        ResultSet resultSet = mock(ResultSet.class);
+        Connection connection = mock(Connection.class);
+        CatalogTable catalogTable = mock(CatalogTable.class);
+        Map<TablePath, CatalogTable> tables = Collections.singletonMap(TABLE_PATH, catalogTable);
+
+        when(dialect.getRowConverter()).thenReturn(rowConverter);
+        when(catalogTable.getTableSchema()).thenReturn(TABLE_SCHEMA);
+
+        JdbcInputFormat inputFormat =
+                new JdbcInputFormat(dialect, chunkSplitter, tables, configuredAutoCommit);
+        return new TestContext(inputFormat, chunkSplitter, statement, resultSet, connection);
+    }
+
+    private static final class TestContext {
+        private final JdbcInputFormat inputFormat;
+        private final ChunkSplitter chunkSplitter;
+        private final PreparedStatement statement;
+        private final ResultSet resultSet;
+        private final Connection connection;
+
+        private TestContext(
+                JdbcInputFormat inputFormat,
+                ChunkSplitter chunkSplitter,
+                PreparedStatement statement,
+                ResultSet resultSet,
+                Connection connection) {
+            this.inputFormat = inputFormat;
+            this.chunkSplitter = chunkSplitter;
+            this.statement = statement;
+            this.resultSet = resultSet;
+            this.connection = connection;
+        }
+
+        private void openEmptySplit() throws IOException, SQLException {
+            when(chunkSplitter.generateSplitStatement(SPLIT, TABLE_SCHEMA)).thenReturn(statement);
+            when(statement.executeQuery()).thenReturn(resultSet);
+            when(resultSet.next()).thenReturn(false);
+            when(statement.getConnection()).thenReturn(connection);
+            inputFormat.open(SPLIT);
+        }
+    }
+}


DanielLeens (commented):

For PostgreSQL/Redshift JDBC sources, `PostgresDialect.creatPreparedStatement()` and `RedshiftDialect.creatPreparedStatement()` unconditionally call `connection.setAutoCommit(false)` to enable server-side cursor mode. Before this patch, `JdbcInputFormat.close()` only closed the `ResultSet` and `PreparedStatement` — it never rolled back or restored auto-commit on the underlying `Connection`, even though that connection is cached and reused across splits (`SimpleJdbcConnectionProvider` / `ChunkSplitter.getOrEstablishConnection()`). The result is a manual transaction that stays open on the server (`idle in transaction`) for the last split processed by a reader, and for any gap while the reader waits for more splits, until the whole reader connection is eventually torn down in `closeInputFormat()`. This patch closes the read transaction (rollback + auto-commit restore) as part of `close()`, and hardens several related cleanup paths (open-failure cleanup, statement-connection lookup failure, rollback failure) so the cached connection is never reused in an unknown transaction state.

I traced the full split lifecycle to verify this is real and that the fix lands at the right point. Verdict: the problem is real, the fix is correctly scoped, and it does not disturb connection reuse on the success path. One non-blocking issue (connection-discard behavior is broader than its own justification) is flagged below.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Split lifecycle before the fix.** `JdbcSourceReader.pollNext()` (`seatunnel-connectors-v2/connector-jdbc/src/main/java/.../source/JdbcSourceReader.java:62-83`) does, per split:
```java
inputFormat.open(split);
while (!inputFormat.reachedEnd()) { ... output.collect(...); }
// finally
inputFormat.close();
```
So `close()` already ran after every split before this PR — the fix lives entirely inside what `close()` does, not in a new call site. `JdbcInputFormat.open()` gets its `PreparedStatement` from `ChunkSplitter.generateSplitStatement()` → `createPreparedStatement()` (`ChunkSplitter.java:148-157`), which resolves the connection via `getOrEstablishConnection()` — a cached, reused `Connection` (`SimpleJdbcConnectionProvider.java:98-123`: `getOrEstablishConnection()` returns the existing `connection` field as long as `isConnectionValid()`). For Postgres/Redshift, `creatPreparedStatement()` then forces `connection.setAutoCommit(false)` (`PostgresDialect.java:203-207`, `RedshiftDialect.java:70-73`) regardless of the `auto_commit` config value, to use a server-side cursor.

Before this PR, `close()` never touched that transaction. I confirmed the "idle in transaction" symptom is real but its precise timing is worth stating accurately: `ChunkSplitter.createPreparedStatement()` (line 152-154) toggles `connection.setAutoCommit(autoCommit)` at the *start* of the next split if the current mode differs from the configured one — and per the JDBC spec, `setAutoCommit()` implicitly commits any pending transaction when the mode actually changes. So in the pre-fix code, a transaction left open by split N was, in most cases, already closed (via an incidental commit, not a rollback) the moment split N+1's `open()` ran. The leak that this PR actually plugs is the window the incidental commit-on-toggle does **not** cover: the *last* split processed by a reader, and any idle gap between splits (the `Thread.sleep(1000L)` wait for more splits, or wait for the coordinator to signal no-more-splits) — exactly what the PR description calls "waiting for another split or while the rest of the job is being cleaned up." That matches the observed PostgreSQL session state precisely.

**The fix.** `JdbcInputFormat.close()` (`JdbcInputFormat.java:148-173`) now:
1. Captures the statement's connection *before* touching resultSet/statement (`getStatementConnection()`, lines 175-198) — necessary because once the statement is closed, `statement.getConnection()` may no longer be usable.
2. Closes result set and statement, nulling each field in a `finally` regardless of close success (so state doesn't leak into the next `open()`/a repeated `close()` call).
3. Calls `finishReadTransaction(connection)` (lines 200-211), which delegates to the static, unit-testable `finishReadTransaction(Connection, boolean)` (lines 230-245): if `connection.isClosed()` it's a no-op; otherwise if `getAutoCommit()` is currently `false` it rolls back (ending the cursor transaction and releasing its snapshot — correct for a read-only source, and also recovers a connection left in "current transaction is aborted" state by a prior error), and only then restores auto-commit to the user-configured value if it differs. The ordering (rollback, then `setAutoCommit`) is correct: reversing it would let the implicit-commit-on-toggle behavior fire again before the explicit rollback, which is a subtler, less deterministic way to end the transaction than what this patch does explicitly.
4. If any of `isClosed()/getAutoCommit()/rollback()/setAutoCommit()` throws `SQLException`, `discardConnection()` (lines 213-228) physically closes the raw connection and calls `chunkSplitter.close()` (`ChunkSplitter.java:83-88` → `connectionProvider.closeConnection()`) to clear the cached reference — so a connection whose transaction state is now unknown is never handed to the next split. This is the right failure mode: correctness over connection reuse when rollback itself fails.

**Default-path compatibility.** `auto_commit` defaults to `true` (`JdbcSinkOptions.AUTO_COMMIT`, default `true`, shared through `JdbcConnectionConfig`), and non-Postgres/Redshift dialects never force auto-commit off. For that (overwhelming majority) case, `getAutoCommit()` stays `true` for the whole read, so `finishReadTransaction` skips both `rollback()` and `setAutoCommit()` — zero behavior change for MySQL/Oracle/SQL Server/etc. sources on default config. This is a real strength of the change: the new code path is a no-op unless a transaction was actually left open.

**Connection reuse on the success path is preserved.** The normal per-split `close()` call never invokes `chunkSplitter.close()` — only `discardConnection()` (rollback/restore failure), `cleanupAfterOpenFailure()` (any `open()` failure), and `getStatementConnection()` (statement-connection lookup failure) do. So on the happy path, the cached `Connection` survives `close()` with just a rollback+restore, remains `isValid()`, and `getOrEstablishConnection()` on the next split's `open()` reuses it rather than reconnecting. I verified this by reading `SimpleJdbcConnectionProvider.getOrEstablishConnection()` (checks `isConnectionValid()` first) and `JdbcConnectionValidationUtils.isConnectionValid()` (a plain `Connection.isValid()`/probe-query check, unaffected by autocommit state). Source-side connections here use `SimpleJdbcConnectionProvider`, a single reused raw JDBC `Connection` — `SimpleJdbcConnectionPoolProviderProxy` (the real connection-pool-backed provider) is sink-only (only referenced from `JdbcSinkWriter`), so there is no shared external pool on the read path whose semantics this patch could disturb.

**Concurrency.** `JdbcInputFormat`/`ChunkSplitter`/`connectionProvider` are all owned exclusively by one `JdbcSourceReader` instance (one per subtask); nothing here is shared across threads, so there's no race between this cleanup and a concurrent split read.

**Idempotency of `close()`.** Worth calling out because `JdbcSourceReader.pollNext()`'s `finally { inputFormat.close(); }` still runs even when `open()` already invoked `cleanupAfterOpenFailure()` → `close()` internally, i.e. `close()` can be called twice in a row for one failed split. I traced this: after the first `close()`, `statement`/`resultSet` are nulled, so the second call's `getStatementConnection()` returns `null` immediately (statement is null) and `finishReadTransaction(null)` short-circuits — no double rollback, no exception. `JdbcInputFormatTest.shouldRollbackAndRestoreConfiguredAutoCommitAfterClosingSplit` explicitly asserts `rollback()` is called exactly once across two `close()` calls. Idempotency claim in the PR description is verified.

## 1.2 Compatibility Impact

**Fully compatible.**
- No config option added, renamed, or removed; no default value changed.
- Default-configuration behavior (auto_commit=true, non-cursor dialects) is provably unchanged (see 1.1).
- For users who already set `auto_commit=false` on a JDBC source (an existing, if unusual, configuration — this option is primarily documented for the sink), the new behavior rolls back the read-only transaction between splits instead of leaving it open indefinitely. Since this connector's source connection never writes, a rollback of a pure `SELECT` has no data-visibility consequence within the split (all rows for the split were already fetched into the driver/client before `close()` runs — see below) — it only ends the server-side snapshot/session state sooner. That is a strict improvement, not a behavior change users could be relying on.
- XA, exactly-once commit semantics, and the sink's own `ConnectionPoolManager`/`SimpleJdbcConnectionPoolProviderProxy` are untouched — this diff is confined to `JdbcInputFormat` (source-only).
- No connector registration, plugin mapping, or dist/packaging change is required, and none was made — consistent with a pure bug fix.

- Success path: one extra `getAutoCommit()` call per split close, and a `rollback()`/`setAutoCommit()` only when a manual transaction is actually open (Postgres/Redshift, or explicit `auto_commit=false`). Negligible overhead, and it replaces a network round trip that would otherwise still happen at the *next* split's `open()` (the incidental commit-on-toggle) — so it is not simply "extra" work for the Postgres/Redshift case, it is work moved earlier and made explicit.
- See **Issue 1** below for a real, non-blocking performance/availability side-effect: `cleanupAfterOpenFailure()` unconditionally discards the whole cached connection on *any* `open()` failure, not only the one scenario its own comment describes.
- Read consistency within a split is unaffected: `close()`/rollback only runs after `pollNext()`'s `while (!inputFormat.reachedEnd())` loop has fully drained the split's `ResultSet` (or the loop was aborted by an exception, in which case the task is failing anyway) — rollback never truncates in-flight reads.

All new failure branches log at `WARN` with enough context (which operation failed, and that the cached connection is being discarded to avoid reusing an unknown transaction state) and attach the original exception via `addSuppressed()` where a cleanup exception could otherwise mask the root cause (`cleanupAfterOpenFailure`, `discardConnection`). No sensitive data (credentials, etc.) is logged. See Issue 1 for the one behavioral note in this area.

**Issue 1 — `cleanupAfterOpenFailure()` discards the connection unconditionally, wider than its own justification**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormat.java:131-141`
- Problem: The `finally` block always calls `chunkSplitter.close()` (which physically closes and discards the cached `Connection` via `ChunkSplitter.close()` → `connectionProvider.closeConnection()`), for *every* exception caught in `open()` — `SQLException` or `RuntimeException`, from any statement inside the `try`. The method's own comment ("If statement creation failed before assignment, close() cannot obtain its connection. Discard the cached connection...") describes only the case where `statement` was never assigned. But when the failure happens *after* the statement/connection are already valid — e.g. `resultSet = statement.executeQuery()` throws — the inner `close()` call in the same method already correctly identifies the connection via `getStatementConnection()` and performs rollback + auto-commit restore (proven by the test `JdbcInputFormatTest.shouldRollbackAndDiscardConnectionWhenExecuteQueryFails`, `.../internal/JdbcInputFormatTest.java:158-174`, which asserts both `rollback()`/`setAutoCommit(true)` **and** `chunkSplitter.close()` are called for the same failure). The connection is discarded even though it was already returned to a clean, reusable state.
- Risk: Every transient per-split failure on Postgres/Redshift (e.g. one query timeout, one dropped packet during `executeQuery()`) forces a full driver-level reconnect (`SimpleJdbcConnectionProvider.getOrEstablishConnection()` → `driver.connect(...)`) for whatever split runs next on that reader, even though nothing about the connection's transaction state actually required that. In practice this is unlikely to matter much because `JdbcSourceReader.pollNext()` does not itself catch and retry a failed `open()` — the exception propagates out of `pollNext()` and the reader task normally fails and gets restarted by the engine (a fresh `JdbcInputFormat` is then constructed), so the discarded connection would usually have been torn down anyway. It is best characterized as a defense-in-depth choice that trades a small amount of avoidable reconnect churn for simplicity and guaranteed safety on every failure path, not a leak or correctness bug.
- Suggested improvement (either is acceptable, not blocking):
  - Option A: Narrow the `finally` to only discard when `statement == null` (the actual "never assigned" case the comment describes) or when the inner `close()` call itself failed to identify/clean the connection, letting a successfully-cleaned connection survive an `executeQuery()`-only failure.
  - Option B: Keep the current unconditional discard, but update the comment to state plainly that *any* open failure invalidates the connection as a simplicity/safety tradeoff, so a future reader isn't misled by the comment into thinking it's narrower than it is.
- Severity: Medium (real, but not a correctness bug, leak, or main-path issue — bounded to failure paths whose blast radius is usually a task restart anyway).

# 2. Code Quality Assessment

## 2.1 Coding Standards

New non-trivial methods (`cleanupAfterOpenFailure`, `getStatementConnection`, `finishReadTransaction` (both overloads), `discardConnection`) are private/package-private helpers without their own Javadoc, but each has an inline comment at the one non-obvious decision point (why discard, why rollback, why check `isClosed()` first) — that is proportionate for package-private helpers of this size, and matches the "concise, why-focused" bar rather than needing full Javadoc on every method. No wildcard imports, uses `java.sql.Connection` directly and appropriately (this class doesn't route through any shaded SQL types). License header preserved. No `System.out.println`, no stray debug logging.

`JdbcInputFormatTest` (new, 242 lines, all pure Mockito against `Connection`/`PreparedStatement`/`ResultSet`/`ChunkSplitter` mocks — no real database, no `Thread.sleep`, no shared static state, no execution-order dependence across tests) covers:
- rollback + restore configured auto-commit (with `InOrder` verification of the exact close sequence: getConnection → resultSet.close → statement.close → rollback → setAutoCommit) and idempotency of a repeated `close()`.
- rollback while preserving an explicitly-configured manual-commit mode (`auto_commit=false`).
- no-op when the connection is already in auto-commit mode.
- transaction is still finished even when `resultSet.close()`/`statement.close()` themselves throw.
- rollback failure → connection discarded + `chunkSplitter.close()` invoked (also the exact test that surfaces Issue 1).
- `statement.getConnection()` throwing → cached connection discarded.
- statement-creation failure (before `statement` is assigned) → cached connection discarded via `open()`'s catch path.
- `executeQuery()` failure → statement closed, transaction rolled back and restored, connection still discarded (Issue 1 evidence).
- `closeInputFormat()` always calls `chunkSplitter.close()`.
- already-closed connection short-circuits `finishReadTransaction` without touching `getAutoCommit()`/`rollback()`.

This is thorough, deterministic coverage of exactly the branches this diff introduces, including the negative/failure paths the review focused on (rollback failure, resource-close failure, statement-creation failure). **Stability rating: Stable.** No flaky-test anti-patterns present.

One coverage gap, non-blocking: there is no test asserting that a *successful* multi-split sequence keeps reusing the same `Connection` object across `open()` calls (i.e., that `chunkSplitter.close()`/`getOrEstablishConnection()` reconnection isn't triggered on the happy path). The current tests prove the transaction is closed correctly but don't directly assert connection reuse is preserved — recommended as a follow-up, not required for this PR, since the reuse logic (`isConnectionValid()`) is pre-existing and unmodified by this diff.

## 2.3 Documentation Updates

None needed and none made. This restores an existing lifecycle contract (a bounded source's connection/transaction should not outlive its split) rather than introducing new user-facing configuration, default, or behavior that a user opts into — consistent with `docs/en` / `docs/zh` requiring updates only for user-visible config/behavior changes, neither of which applies here.

# 3. Architectural Soundness

**Precise fix.** The change is confined to the one class that owns the split-scoped resource lifecycle (`JdbcInputFormat`), reuses the existing cached-connection mechanism (`ChunkSplitter`/`JdbcConnectionProvider`) rather than introducing a new one, and is a no-op for the default configuration. It fixes the root cause (transaction never ended at the actual split boundary) rather than papering over the symptom (e.g., it does not try to work around this by shortening idle-in-transaction timeouts at the database level or by force-closing the connection after every split, which would have defeated connection reuse entirely).

## 3.2 Maintainability

The two-argument static `finishReadTransaction(Connection, boolean)` is a clean, directly-unit-testable core of the fix, separated from the instance-level orchestration (capturing the connection before resource close, deciding discard-vs-restore on failure). This separation is what makes the test suite's precision possible.

## 3.3 Extensibility

Nothing here blocks future work; if a future dialect needs different end-of-split transaction semantics (e.g., a driver where rollback isn't safe to call blindly), the single `finishReadTransaction(Connection, boolean)` choke point is an easy place to add a dialect hook without touching the rest of the class.

`JdbcInputFormat implements Serializable` with `serialVersionUID = 2L`, left unchanged despite adding a new `final boolean configuredAutoCommit` field. I checked whether this class is ever part of persisted checkpoint/savepoint state that would need cross-version deserialization compatibility: `JdbcSourceReader.snapshotState()` (`JdbcSourceReader.java:86-88`) only checkpoints the `List<JdbcSourceSplit>` queue, not the `JdbcInputFormat`/`JdbcSourceReader` object itself — the reader is reconstructed via its constructor on each run, so `JdbcInputFormat`'s `Serializable`-ness appears to be about one-time transport to a worker within a single job run (same JAR on both ends), not about restoring old checkpoint bytes into a newer class after an upgrade. Based on this, I don't believe this field addition creates a real cross-version compatibility risk, but flagging the reasoning explicitly so a maintainer with more context on the coordinator/worker dispatch path can confirm.

# 4. Issue Summary

| 1 | `cleanupAfterOpenFailure()` unconditionally discards/closes the cached connection on any `open()` failure, wider than the "statement never assigned" case its comment describes | `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormat.java:131-141` | Medium |

No High-severity issues found. No functional regression, data-loss risk, compatibility break, or resource leak identified.

# 5. Merge Recommendation

1. Blockers — must be fixed
   - None.
   - Issue 1: narrow `cleanupAfterOpenFailure()`'s unconditional `chunkSplitter.close()` to the case it's actually meant for (Option A), or otherwise correct the comment to reflect the broader, intentional "any open failure invalidates the connection" policy (Option B). Either is acceptable; I'd lean toward Option A for connection-reuse efficiency, but Option B is a legitimate simplicity/safety tradeoff if the maintainers prefer to keep the blast radius simple to reason about.
   - Consider a follow-up test asserting connection-object reuse across a successful multi-split sequence, to guard the "doesn't break connection pooling/reuse" property this PR is careful to preserve.

Overall: this is a well-diagnosed, correctly-scoped fix for a real PostgreSQL/Redshift "idle in transaction" leak. I verified the root cause end-to-end (dialect-forced `setAutoCommit(false)`, cached/reused connection across splits, and the pre-existing incidental-commit-on-toggle behavior that only partially covered the gap), confirmed the fix's ordering (rollback before restoring auto-commit) is correct and read-consistency-safe, confirmed connection reuse is preserved on the success path, and confirmed the new test suite is deterministic and exercises the failure branches this review focused on. The one Medium finding (Issue 1) is a minor efficiency/comment-accuracy gap in a failure path, not something that should block this fix from landing — happy to see it merged once the author has a chance to look at it, and equally fine if the maintainers judge Option B (just fix the comment) sufficient.

Thanks for the very thorough problem write-up and test coverage on a first contribution — this is exactly the kind of "trace it to the real dialect/connection code, not just the log line" investigation this bug needed.

wybaby168:
Thanks for the thorough lifecycle review and for calling out the over-broad
failure-path discard.

I applied Option A in 7f70804263e41ccc190ac29cf9a7fb1bcf7b6f6e:

- cleanupAfterOpenFailure() now discards the cached connection only when
  statement creation failed before assignment, or when cleanup cannot restore
  a known-safe connection state.
- An executeQuery() failure now closes the statement, rolls back the read
  transaction, restores the configured auto-commit mode, and preserves the
  connection when cleanup succeeds.
- Added a two-split regression test to verify that the same Connection remains
  reusable across successful split lifecycles.
- Updated the execute-failure regression to assert connection reuse instead of
  unconditional discard.
- Rebased onto the latest apache/seatunnel:dev and kept the branch as one clean
  commit.

Thank you again for the detailed review. Could you please take another look
when convenient?

SEZ9 (commented):
Thanks for working on this. I reviewed the latest head `7f70804263e4`.

# What this PR fixes
- User pain: On PostgreSQL/Redshift cursor reads (where the dialect disables auto-commit), `JdbcInputFormat.close()` only closed the `ResultSet`/`PreparedStatement`, so the reused source connection sat `idle in transaction` between splits — pinning `backend_xmin`, blocking vacuum, and holding locks and connection slots.
- Fix approach: Capture the statement's connection before closing resources, always null-out and close the result set and statement even when either close throws, then roll back any open read transaction and restore the user-configured auto-commit mode (`configuredAutoCommit` is now threaded through a new package-private constructor). Failures during `open()` (both `SQLException` and `RuntimeException`) go through `cleanupAfterOpenFailure()`, and unrecoverable cleanup paths discard the cached connection via `chunkSplitter.close()` (or `connection.close()` in `discardConnection()`).
- One-line summary: The lifecycle handling in `JdbcInputFormat` looks correct, and the new `JdbcInputFormatTest` covers the main branches (rollback/restore, manual-commit config, auto-commit no-op, close-failure resilience, connection reuse across splits, discard-on-cleanup-failure, statement-creation failure); the runtime chain is traced below.

# Runtime chain I rechecked
```text
JdbcSourceReader.pollNext()
  -> JdbcInputFormat.open(JdbcSourceSplit)  JdbcInputFormat.java:101
       -> chunkSplitter.generateSplitStatement(split, schema)  JdbcInputFormat.java:112 (dialect may setAutoCommit(false) for cursor reads)
       -> statement.executeQuery(); resultSet.next()  JdbcInputFormat.java:117-118
       [on SQLException/RuntimeException]
       -> cleanupAfterOpenFailure(e)  JdbcInputFormat.java:131 -> close(); if statement==null or close() threw -> chunkSplitter.close()  JdbcInputFormat.java:135-149
  -> JdbcInputFormat.close()  JdbcInputFormat.java:153
       -> getStatementConnection()  JdbcInputFormat.java:154 -> statement.getConnection(); on SQLException/null -> chunkSplitter.close()  JdbcInputFormat.java:178-201
       -> resultSet.close()/statement.close() with finally { field = null }  JdbcInputFormat.java:155-171
       -> finishReadTransaction(connection)  JdbcInputFormat.java:176
            -> finishReadTransaction(connection, configuredAutoCommit)  JdbcInputFormat.java:232
                 -> connection.isClosed() guard; if !getAutoCommit() -> connection.rollback(); if mode differs -> connection.setAutoCommit(configuredAutoCommit)  JdbcInputFormat.java:234-244
            [on SQLException] -> discardConnection(connection, e)  JdbcInputFormat.java:214 -> connection.close(); finally chunkSplitter.close()  JdbcInputFormat.java:216-229
JdbcSourceReader.close()
  -> JdbcInputFormat.closeInputFormat()  JdbcInputFormat.java:90 -> try { close() } finally { chunkSplitter.close() }  JdbcInputFormat.java:91-97
```

# Review conclusion

Nice work — the previous concerns are addressed. Thanks again for the contribution.

DanielLeens (commented):

- User pain point: For PostgreSQL/Redshift JDBC sources, the dialect forces `connection.setAutoCommit(false)` to enable a server-side cursor. `JdbcInputFormat.close()` used to close only the `ResultSet`/`PreparedStatement`, never rolling back or restoring auto-commit on the cached, cross-split `Connection` — leaving a manual transaction `idle in transaction` on the last split of a reader and during any idle gap between splits, pinning `backend_xmin`, blocking vacuum, and holding locks/connection slots.
- Fix approach: capture the statement's connection before closing resources, always null the `resultSet`/`statement` fields (even if their `close()` throws), then roll back the open read transaction and restore the user-configured auto-commit mode; failures during `open()` and during transaction cleanup route through dedicated helpers that discard the cached connection only when its post-cleanup transaction state can no longer be trusted.
- One-sentence summary: this closes a real, verified "idle in transaction" leak at the correct lifecycle boundary, is a no-op for the default (non-cursor) configuration, and — as of this commit — no longer over-discards the cached connection on a simple `executeQuery()` failure.

Thanks for the very thorough follow-up, @wybaby168 — Option A is exactly what I'd hoped for, and the new `shouldKeepConnectionReusableAcrossSuccessfulSplits` test also closes the non-blocking coverage gap I flagged. I re-traced the whole diff at head `7f70804263e4` from scratch rather than just diffing against my last pass, and I want to explicitly +1 @SEZ9's independent review here too: the runtime chain they posted (`JdbcSourceReader.pollNext() -> open() -> cleanupAfterOpenFailure()/close() -> getStatementConnection()/finishReadTransaction()/discardConnection()`) matches exactly what I verified against the live files below, right down to the same line numbers. I found one small, purely cosmetic nit that neither of us called out yet (Issue 1 below) — flagging it for completeness, it is not a correctness problem and does not change my merge recommendation.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Split lifecycle (unchanged call sites).** `JdbcSourceReader.pollNext()` still does, per split:
```java
inputFormat.open(split);
while (!inputFormat.reachedEnd()) { ... output.collect(...); }
// finally
inputFormat.close();
```
(`seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/source/JdbcSourceReader.java:66-74`). This PR does not add a new call site — it changes what `close()` (and the new `open()` failure path) actually does, which is the correct place to fix this: `close()` already runs after every split, success or failure.

**The fix, traced end-to-end at head:**
1. `close()` (`JdbcInputFormat.java:153-178`) captures the statement's connection *first* via `getStatementConnection()` (`:180-203`) — necessary because once the statement is closed, `statement.getConnection()` is no longer reliable. If `statement.getConnection()` itself throws, or returns `null`, the cached connection is proactively discarded (`chunkSplitter.close()`) since its transaction state can't be identified.
2. `resultSet`/`statement` are closed and unconditionally nulled in `finally` blocks (`:155-172`) — idempotent even if `close()` runs twice for one split (see below).
3. `finishReadTransaction(connection)` (`:177`, delegating to the static, directly-unit-tested `finishReadTransaction(Connection, boolean)` at `:235-250`) is a no-op if the connection is already closed; otherwise, if `getAutoCommit()` is currently `false` it rolls back (ending the cursor transaction/snapshot, and also recovering a connection left in an aborted-transaction state by a prior query error), then restores auto-commit to the configured value only if it actually differs. I verified the ordering is correct: rollback before `setAutoCommit`, so the explicit rollback — not the JDBC-spec implicit-commit-on-`setAutoCommit`-toggle — is what ends the transaction.
4. Any `SQLException` from that sequence routes to `discardConnection()` (`:218-233`), which physically closes the raw connection and always clears the provider's cached reference via `chunkSplitter.close()` — correctness over reuse when cleanup itself fails.

**Issue 1 from my previous pass is resolved as promised (Option A).** `cleanupAfterOpenFailure()` (`JdbcInputFormat.java:131-146`) now seeds `shouldDiscardConnection = statement == null` — i.e. it only pre-emptively discards when the statement was never assigned (so `close()` has no way to identify the connection). I independently re-verified the two branches this changes:
- **`executeQuery()` fails** (statement was assigned): `shouldDiscardConnection` starts `false`; `close()` succeeds (rollback + restore run normally via step 3 above) without throwing, so the connection survives. Confirmed by `shouldRollbackAndKeepConnectionWhenExecuteQueryFails` (`JdbcInputFormatTest.java:187-204`), which now asserts `verify(context.connection, never()).close(); verify(context.chunkSplitter, never()).close();` — the exact inversion of the old assertion my last review flagged.
- **statement creation fails** (`statement` never assigned): `shouldDiscardConnection` stays `true`, so `chunkSplitter.close()` still runs. Confirmed by `shouldDiscardCachedConnectionWhenStatementCreationFails` (`JdbcInputFormatTest.java:175-186`).

This is a real, verified improvement: a transient per-split query failure on Postgres/Redshift no longer forces a full driver-level reconnect for the next split on that reader.

**Default-path compatibility (re-verified against live dialect code, not just the diff).** I fetched the current `dev` versions of `PostgresDialect.creatPreparedStatement()` (`seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/dialect/psql/PostgresDialect.java:203-207`) and `RedshiftDialect.creatPreparedStatement()` (`.../dialect/redshift/RedshiftDialect.java:70-73`) — both still unconditionally call `connection.setAutoCommit(false)`, confirming the leak's root cause is exactly as described. I also confirmed `JdbcSinkOptions.AUTO_COMMIT.defaultValue()` is `true` (`JdbcSinkOptions.java:63-67`) and is threaded through `JdbcConnectionConfig.autoCommit` by default. So for every other dialect on default config, `getAutoCommit()` is `true` throughout, and `finishReadTransaction` skips both `rollback()` and `setAutoCommit()` — zero behavior change for MySQL/Oracle/SQL Server/etc. on default config.

**Connection reuse on the success path — now directly tested.** `shouldKeepConnectionReusableAcrossSuccessfulSplits` (`JdbcInputFormatTest.java:117-145`) opens and closes two splits against the same mocked `Connection` and asserts `rollback()`/`setAutoCommit(true)` fire twice each while `connection.close()`/`chunkSplitter.close()` are never called — this is precisely the follow-up test I recommended in my prior review, and it directly demonstrates the happy path preserves connection reuse rather than leaving it as an inference from unmodified code.

**Idempotency of double-`close()` (re-verified against the current file, not just the test).** `JdbcSourceReader.pollNext()`'s `finally { inputFormat.close(); }` still runs even when `open()` already triggered `cleanupAfterOpenFailure()` → `close()` internally. I traced this directly: after the first `close()`, `statement`/`resultSet` are `null`, so `getStatementConnection()` (`:181-183`) returns `null` immediately, and `finishReadTransaction(null)` short-circuits at `connection == null` (`:237-239`) — no double rollback. `shouldRollbackAndRestoreConfiguredAutoCommitAfterClosingSplit` (`JdbcInputFormatTest.java:56-75`) explicitly calls `close()` twice and asserts `rollback()` fires exactly once.

**Exceptional-path coverage for the concurrency/retry angle this review was asked to check.** A mid-split failure — not just an `open()` failure — is also covered: if `nextRecord()` throws while the `while (!inputFormat.reachedEnd())` loop in `pollNext()` is draining rows, the exception propagates out of the `try`, and the same `finally { inputFormat.close(); }` runs the identical rollback/restore path. There is no separate "abandon transaction" path for mid-read failures; it funnels through the same `close()` this review already validated. On the retry/restart angle: `JdbcSourceReader.snapshotState()` (`JdbcSourceReader.java:86-88`) only checkpoints the remaining `List<JdbcSourceSplit>`, never the `JdbcInputFormat`/`ChunkSplitter`/`Connection` — I confirmed this directly in the live file. A checkpoint restore or task restart always constructs a fresh `JdbcSourceReader`/`JdbcInputFormat`/connection, so there's no scenario where this PR's cleanup logic needs to run against resources left over from a previous attempt; the abrupt-crash case (no graceful `close()` at all) is handled at the TCP/DB level as before and is unaffected by this diff either way.

## 1.2 Compatibility Impact

**Fully compatible.**
- No config option added, renamed, or removed; no default value changed.
- Default-configuration behavior (auto_commit=true, non-cursor dialects) is provably unchanged (verified independently above, not just asserted).
- For users who already run with `auto_commit=false` on a JDBC source, the new behavior ends the read-only transaction between splits instead of leaving it open indefinitely — a strict improvement with no data-visibility impact, since all rows for a split are already drained into the driver before `close()` runs.
- Confined to `JdbcInputFormat` (source-only); the sink's `SimpleJdbcConnectionPoolProviderProxy`/`ConnectionPoolManager` and XA/exactly-once paths are untouched.
- `JdbcInputFormat` gains a new `final boolean configuredAutoCommit` field but keeps `serialVersionUID = 2L`. I re-confirmed `snapshotState()` never serializes `JdbcInputFormat` itself (only the split list), so this isn't a checkpoint/savepoint cross-version compatibility risk.

- Success path: one extra `getAutoCommit()` per split close, plus `rollback()`/`setAutoCommit()` only when a manual transaction is actually open — and even then, this replaces a network round trip that used to happen anyway at the *next* split's `open()` via the implicit-commit-on-toggle, so it's work moved earlier and made explicit rather than pure overhead.
- With Issue 1 from my previous review now fixed, a transient `executeQuery()` failure on Postgres/Redshift no longer forces an avoidable full reconnect for the next split — this is a genuine, verified efficiency improvement over the version I reviewed last time.
- Read consistency within a split is unaffected: rollback only runs after the split's `ResultSet` has been fully drained (or the read failed, in which case the task is failing anyway).
- Concurrency: `JdbcInputFormat`/`ChunkSplitter`/`connectionProvider` are all owned by a single `JdbcSourceReader` instance; nothing here is shared across threads.

All new failure branches log at `WARN` with actionable context and attach the original exception via `addSuppressed()` where a cleanup failure could otherwise mask the root cause. No sensitive data logged.

**Issue 1 — dead `catch (IOException cleanupException)` branch in `cleanupAfterOpenFailure()`**
- Location: `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormat.java:131-146` (specifically the `catch (IOException cleanupException)` at line ~135)
- Problem: I traced every statement inside the current `close()` body (`:153-178`) — `resultSet.close()`/`statement.close()` catch and log their own `SQLException`s internally without rethrowing, and `finishReadTransaction(Connection)` (`:205-216`) likewise catches its own `SQLException` and routes it to `discardConnection()` rather than rethrowing. So, as currently implemented, `close()` never actually throws `IOException` — it only carries the `throws IOException` on its signature because that was already true of the pre-PR method (I confirmed this against the `dev` baseline: the old `close()` also swallowed all `SQLException`s and never threw). This means the new `catch (IOException cleanupException) { ...; shouldDiscardConnection = true; }` branch added by this PR is unreachable under the current implementation.
- Risk: None functionally today — `shouldDiscardConnection` is already `true` whenever it needs to be, via the `statement == null` seed. The only risk is to a future reader: the dead branch reads as if there's a live path where `close()` fails with `IOException` and the connection gets discarded because of it, which isn't true right now and could mislead someone extending this code later.
- Suggested improvement: either remove the `try/catch (IOException)` around the `close()` call in `cleanupAfterOpenFailure()` (since `shouldDiscardConnection` is already computed correctly from `statement == null`), or leave it as harmless defensive code but add a one-line comment noting it's a safety net for a hypothetical future `close()` implementation that does throw. Purely cosmetic — not blocking.
- Severity: Low.

# 2. Code Quality Assessment

## 2.1 Coding Standards

New private/package-private helpers (`cleanupAfterOpenFailure`, `getStatementConnection`, both `finishReadTransaction` overloads, `discardConnection`) don't carry full Javadoc blocks, but each has a targeted inline comment at its one non-obvious decision point (why discard, why rollback-before-restore, why check `isClosed()` first). For helpers of this size and visibility that's proportionate to this codebase's existing convention in this file — consistent with what I noted last round, and unchanged by this commit. No wildcard imports, no `System.out.println`, license headers intact on the new test file, no special/emoji characters in comments or logs.

`JdbcInputFormatTest` (272 lines, pure Mockito against `Connection`/`PreparedStatement`/`ResultSet`/`ChunkSplitter` — no real database, no `Thread.sleep`, no shared static state, no ordering dependence across tests) now covers, in addition to everything from the previous round:
- Connection reuse across two full successful split cycles (`shouldKeepConnectionReusableAcrossSuccessfulSplits`, `:117-145`) — new in this commit, closes the coverage gap I flagged as a non-blocking recommendation last time.
- The revised, narrower discard behavior on `executeQuery()` failure (`shouldRollbackAndKeepConnectionWhenExecuteQueryFails`, `:187-204`) — updated to assert connection *retention* instead of the old unconditional-discard assertion, and it's the test that concretely proves Issue 1 (previous round) is fixed.
- All previously-verified branches remain intact: rollback+restore with `InOrder` sequencing and idempotency (`:56-75`), manual-commit-mode preservation (`:76-88`), auto-commit no-op (`:89-101`), transaction still finished when resource-close itself throws (`:102-116`), discard-on-rollback-failure (`:146-160`), discard when `statement.getConnection()` throws (`:161-174`), discard on statement-creation failure (`:175-186`), unconditional `chunkSplitter.close()` at final `closeInputFormat()` teardown (`:206-214`), and the closed-connection short-circuit (`:215-224`).

**Stability rating: Stable.** No flaky-test anti-patterns (no hard sleeps, no nondeterministic assertions, no hidden ordering/env dependencies, no floating point, no unreleased mock resources); every test is a deterministic, isolated Mockito interaction test. No coverage gaps remain that I can identify for the branches this diff introduces.

## 2.3 Documentation Updates

None needed and none made — this restores an existing lifecycle contract (a bounded split's connection/transaction shouldn't outlive the split) rather than adding user-facing config or default behavior. Consistent with `docs/en`/`docs/zh` only needing updates for user-visible changes, which this isn't.

# 3. Architectural Soundness

## 3.1 Elegance of the Solution — **Precise fix**

Confined to the one class that owns the split-scoped resource lifecycle, reuses the existing cached-connection mechanism rather than introducing a new one, and — after this commit's Option A refinement — the failure-path discard logic is now scoped precisely to the cases where the connection's post-cleanup state genuinely can't be trusted, not broader than that.

## 3.2 Maintainability

The static `finishReadTransaction(Connection, boolean)` remains a clean, directly-unit-testable core, separated from the instance-level orchestration. The Option A change to `cleanupAfterOpenFailure()` is a small, well-contained edit (one boolean seed) that doesn't disturb this separation.

## 3.3 Extensibility

Unchanged from my prior assessment: the single `finishReadTransaction(Connection, boolean)` choke point remains an easy place to add a per-dialect hook later if a driver ever needs different end-of-split semantics.

Re-confirmed directly against the live `JdbcSourceReader.java`: `snapshotState()` (`:86-88`) only checkpoints the split queue, not the `JdbcInputFormat` object, so the new `configuredAutoCommit` field does not introduce a checkpoint/savepoint cross-version deserialization risk. `serialVersionUID = 2L` is unchanged, consistent with that conclusion.

# 4. Issue Summary

| 1 | Dead/unreachable `catch (IOException cleanupException)` branch in `cleanupAfterOpenFailure()` — `close()` never actually throws `IOException` in the current implementation | `seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/internal/JdbcInputFormat.java:131-146` | Low |

No High or Medium-severity issues found in this pass. The Medium-severity issue from my previous review (over-broad connection discard on `open()` failure) is confirmed fixed, with test evidence.

# 5. Merge Recommendation

   - None.
   - Issue 1: either drop the now-dead `catch (IOException)` in `cleanupAfterOpenFailure()`, or add a one-line comment noting it's a safety net rather than a live path, so a future reader isn't misled. Purely cosmetic.

Overall: this is a well-diagnosed, correctly-scoped fix for a real PostgreSQL/Redshift "idle in transaction" leak, and this round of changes does exactly what the author's write-up promised — I independently re-verified both the `executeQuery()`-failure connection-retention behavior and the new multi-split reuse test against the live code rather than taking the test names at face value. I also re-checked the exceptional/mid-read failure path and the checkpoint/retry angle specifically, since that's what this review round was asked to focus on, and didn't find any leak or double-cleanup risk there. Combined with @SEZ9's independently-traced runtime chain landing on the same conclusion, I don't see anything blocking here.

One process note for maintainers, not the author: CI on the fork (`flyfish-dev/seatunnel`) still has 3 jobs `in_progress` (`updated-modules-integration-test-part-6`/`part-7`) as of this review — the `Build` check showing "failure" on the apache/seatunnel side is the known pointer-check artifact, not a real failure; I checked the fork's actual check-runs directly and found zero failed/cancelled/timed-out jobs, only in-progress ones. Worth a final glance once those finish, but I wouldn't hold the merge on it.

Thanks again for the careful root-cause tracing and for taking the previous round's feedback seriously enough to add exactly the regression test that would prove it — this is genuinely solid work for a first contribution.

RenHuifeng:
队友真优秀

wybaby168:
Thank you for the very thorough re-review and for independently validating the full runtime lifecycle at `7f70804263e4`. I really appreciate your confirmation that the Option A refinement and the multi-split connection-reuse test fully address the previous concerns.

Good catch on the defensive `catch (IOException cleanupException)` branch. I agree that it is not currently exercised because `close()` handles its SQL exceptions internally. Since `close()` still retains the existing `throws IOException` contract, and this point is purely cosmetic and non-blocking, I’ll keep the reviewed head unchanged to avoid unnecessary review and CI churn. The catch will remain as a safety net for a future implementation that may propagate an `IOException`.

I’ll continue monitoring the remaining CI jobs. Thanks again for the detailed analysis, constructive feedback, and merge recommendation.

DanielLeens:
Thanks for the thoughtful follow-up, @wybaby168 — that's a reasonable call on Issue 1. Since `close()` doesn't currently throw `IOException` anywhere in this file, the dead `catch` branch is genuinely cosmetic, and I agree it's fine to leave the head as-is rather than churn CI over it; a one-line comment noting it's a defensive safety net (if you ever touch this method again) would be a nice-to-have but isn't something I'd ask for a new commit over.

No action needed from my side — my last review already landed on "Ready to merge" with no blockers, and that stands. Thanks again for the careful root-cause work on this one.

davidzollo:

```

```

davidzollo (approved):
+1
LGTM

```

```

PR #11771, opened 2026-08-12
[Fix][Connector-V2] Honor connection-test-query from properties config in JDBC sink HikariCP pool

### LINKED ISSUE #11769: The sink is MySQL. After running in STREAMING mode for a period of time, it throws the error "No operations allowed after statement closed". How can I resolve this issue?
### Search before asking

### What happened

In MySQL‑CDC mode with MySQL as the sink in STREAMING mode, this error occurs after running for a period of time. It is presumed that the connection becomes invalid, yet Seatunnel still tries to use this connection without mechanisms such as connection liveness detection.

### SeaTunnel Version

2.3.12

### SeaTunnel Config

```conf
sink {
  jdbc {
      url = "jdbc:mysql://localhost:3306/a?useSSL=false&serverTimezone=Asia/Shanghai"
      user = "xx"
      password = "xx"
	  max_retries = 5
      
      table = "${table_name}"
      generate_sink_sql = true
	  
	  connection_check_timeout_sec = 1800
	  properties {
		connection-test-query = "select 1"
		validation-timeout = 3000
		max-lifetime = 600000
		leak-detection-threshold = 60000
	  }
      
      database = a
      
      primary_keys = ["${primary_key}"]
      
      
  }
}
```

### Running Command

```shell
./seatunnel.sh --config ../config/a.conf
```

### Error Exception

```log
Caused by: java.sql.SQLException:No operations allowed after statement closed
```

_No response_

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

waterWang:
### Purpose

Fix the JDBC sink HikariCP connection pool ignoring the `connection-test-query` configured by users in the `properties` block.

### Root Cause

When users configure a connection test query via `properties.connection-test-query` (or `properties.connectionTestQuery`), the current code passes it to the JDBC DataSource via `HikariDataSource.addDataSourceProperty()`. HikariCP pool properties are silently ignored when passed as DataSource properties — the `setConnectionTestQuery()` method must be used instead. This means the connection test query never takes effect during long-running streaming jobs, leading to `No operations allowed after statement closed` errors when the MySQL server closes idle connections.

### Changes

- Extended `JdbcSinkWriter.applyConnectionValidation()` to check the user's `properties` config for `connectionTestQuery` and `connection-test-query` keys
- When found, applies them to the HikariCP pool via `setConnectionTestQuery()` directly

### Related issue

Closes #11769

### DIFF
--- a/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/sink/JdbcSinkWriter.java
+++ b/seatunnel-connectors-v2/connector-jdbc/src/main/java/org/apache/seatunnel/connectors/seatunnel/jdbc/sink/JdbcSinkWriter.java
@@ -107,12 +107,23 @@ public class JdbcSinkWriter extends AbstractJdbcSinkWriter<ConnectionPoolManager
 
     /**
      * Configures pool-level validation for JDBC drivers that cannot pass Hikari's default
-     * Connection.isValid(timeout) probe.
+     * Connection.isValid(timeout) probe, and honors user-specified connection test queries
+     * from the {@code properties} configuration block.
      */
     static void applyConnectionValidation(
             HikariDataSource dataSource, JdbcConnectionConfig jdbcConnectionConfig) {
         JdbcConnectionValidationUtils.getConnectionValidationQuery(jdbcConnectionConfig)
                 .ifPresent(dataSource::setConnectionTestQuery);
+        // Also honor user-specified connection test query from the properties config.
+        // HikariCP's setConnectionTestQuery() must be used instead of addDataSourceProperty()
+        // because the latter passes the property to the JDBC DataSource, not to HikariCP's pool.
+        String testQuery = jdbcConnectionConfig.getProperties().get("connectionTestQuery");
+        if (testQuery == null) {
+            testQuery = jdbcConnectionConfig.getProperties().get("connection-test-query");
+        }
+        if (testQuery != null) {
+            dataSource.setConnectionTestQuery(testQuery);
+        }
     }
 
     @Override


DanielLeens (commented):

`JdbcSinkWriter.initMultiTableResourceManager()` builds the HikariCP pool shared by the multi-table JDBC sink (the path used by MySQL-CDC-to-MySQL style jobs with `table = "${table_name}"`, which is exactly the reporter's config in #11769). Before this patch, any key a user put under the sink's `properties { ... }` block — including `connection-test-query` / `connectionTestQuery` — was forwarded only via `HikariDataSource.addDataSourceProperty()`. That method feeds the underlying `DriverDataSource`'s JDBC-URL connection properties, not HikariCP's own pool configuration, so a user-specified test query never actually became `HikariConfig.connectionTestQuery`. HikariCP therefore kept validating idle/borrowed connections with the JDBC4 `Connection.isValid(timeout)` probe instead of the user's SQL query, which is the documented workaround for environments (proxies, NATs, firewalls with idle-connection eviction) where `isValid()` can return a false positive. That is consistent with the reporter's symptom: long-running MySQL sink jobs eventually hit `No operations allowed after statement closed` because a connection that looked "valid" to Hikari was in fact dead server-side.

The fix adds an explicit lookup of `connectionTestQuery`/`connection-test-query` inside `applyConnectionValidation()` and calls `HikariDataSource.setConnectionTestQuery()` directly when either key is present. I traced the full call chain below; the root-cause diagnosis is correct and the fix lands in the right method, but it is narrower than the bug class it's named after, and it doesn't reach every place in the connector that independently decides "is this connection alive."

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Where this executes.** `initMultiTableResourceManager()` (`JdbcSinkWriter.java:81-106`) is invoked once per `MultiTableSinkWriter` (`seatunnel-api/.../multitablesink/MultiTableSinkWriter.java:260-272`, `initResourceManager()`), i.e. for any JDBC sink registered through the multi-table sink path — which includes the common single-writer-per-table CDC replication pattern the linked issue uses. The `HikariDataSource` built there (`ds`, driven by `jdbcUrl` + `driverClassName`, no `dataSourceClassName`) is the pool every `SimpleJdbcConnectionPoolProviderProxy` borrows connections from at write time.

**Before → after in `applyConnectionValidation` (`JdbcSinkWriter.java:113-127`).**
```java
// before
static void applyConnectionValidation(HikariDataSource dataSource, JdbcConnectionConfig cfg) {
    JdbcConnectionValidationUtils.getConnectionValidationQuery(cfg)
            .ifPresent(dataSource::setConnectionTestQuery);
}

// after
static void applyConnectionValidation(HikariDataSource dataSource, JdbcConnectionConfig cfg) {
    JdbcConnectionValidationUtils.getConnectionValidationQuery(cfg)
            .ifPresent(dataSource::setConnectionTestQuery);
    String testQuery = cfg.getProperties().get("connectionTestQuery");
    if (testQuery == null) {
        testQuery = cfg.getProperties().get("connection-test-query");
    }
    if (testQuery != null) {
        dataSource.setConnectionTestQuery(testQuery);
    }
}
```
`cfg.getProperties()` is guaranteed non-null — `JdbcConnectionConfig.Builder.build()` (`JdbcConnectionConfig.java:345-346`) defaults it to `new HashMap<>()` when the `properties` block is absent — so there is no NPE risk for the common case where a user never configures `properties` at all; I verified this explicitly since the pre-existing, untouched line right after this call (`JdbcSinkWriter.java:104`, `getProperties().forEach(ds::addDataSourceProperty)`) would already have crashed on every properties-less job if that guarantee didn't hold.

**Compatibility of the new branch with the existing Xugu hook.** `JdbcConnectionValidationUtils.getConnectionValidationQuery()` (`JdbcConnectionValidationUtils.java:65-77`) is a hardcoded, single-purpose hook: it returns a validation query only for the Xugu driver (whose `Connection.isValid()` throws), `Optional.empty()` for everything else — including MySQL/Postgres/Oracle/etc., which is exactly right, since JDBC4 `isValid()` is the correct default for drivers that support it. The new code runs unconditionally after that call, so a user-specified `properties.connectionTestQuery` correctly overrides the Xugu default too if both happen to apply (not a realistic combination, but the precedence is sane). For every other driver, `setConnectionTestQuery()` switches Hikari from `isValid()`-based validation to SQL-query-based validation for the *entire pool*, per HikariCP's own semantics (`HikariConfig.connectionTestQuery != null` short-circuits the `isValid()` path in `PoolBase.isConnectionAlive()`). That's the intended effect and it is driver-appropriate: it only fires when the user explicitly asks for it, so drivers that are fine with `isValid()` are unaffected unless the user opts in.

**Issue 1 — the fix only reaches HikariCP's own pool-level validation, not the connector's own `isConnectionValid()` checks used elsewhere on the same write path (Medium, see 1.4).** `JdbcConnectionValidationUtils.getConnectionValidationQuery()` is also the sole source of truth for `JdbcConnectionValidationUtils.isConnectionValid()` (`JdbcConnectionValidationUtils.java:43-59`), which is called from `SimpleJdbcConnectionPoolProviderProxy.isConnectionValid()` (`SimpleJdbcConnectionPoolProviderProxy.java:47-51`) and, critically, from `JdbcOutputFormat.shouldRefreshExecutor()` (`JdbcOutputFormat.java:233-238`) — the exact reactive-recovery check (`hasConnectionErrorSqlState(...) || isStatementClosed(...) || !connectionProvider.isConnectionValid()`) that decides whether to reconnect after a batch fails, and whose `isStatementClosed()` branch (`JdbcOutputFormat.java:271-278`) is what actually matches the reporter's literal error text. None of these call sites were touched by this PR, so they still resolve validation purely via the Xugu-only hook and fall back to `connection.isValid()` for MySQL — the user's `properties.connectionTestQuery` is honored only for Hikari's own internal borrow-time/keepalive eviction, not for this connector-level reactive check. In practice this is not catastrophic (`isValid()` correctly detects most genuinely dead connections), but it means the fix is asymmetric: the same conceptual "how do we know this connection is alive" question now has two different answers depending which code path asks it.

**Issue 2 — the fix is narrower than the bug class in the linked issue (Medium, see 1.4).** The reporter's own repro config (issue #11769) sets `connection-test-query`, `validation-timeout`, `max-lifetime`, and `leak-detection-threshold` all under `properties`, clearly expecting all four to become real HikariCP pool settings — a reasonable expectation, since that's exactly what a native `hikari.properties` file would do. This PR fixes only `connection-test-query`; `validation-timeout`, `max-lifetime`, and `leak-detection-threshold` still flow exclusively into `addDataSourceProperty()` (`JdbcSinkWriter.java:104`, unchanged) and remain silently no-ops, same as before. The PR title/description promise "Honor connection-test-query from properties config," which it does — but it will not fully resolve the reporter's original scenario if their pool is still leaking connections due to `max-lifetime`/`leak-detection-threshold` never taking effect.

**Issue 3 — the extracted key is left in `properties` and re-applied via `addDataSourceProperty` right after (Low, see 1.4).** `JdbcSinkWriter.java:104` (`getProperties().forEach(ds::addDataSourceProperty)`) still runs immediately after `applyConnectionValidation()` and iterates the same, un-filtered `properties` map — so `connectionTestQuery`/`connection-test-query` is applied to the Hikari pool config *and* forwarded again as a raw JDBC connection property to the underlying `DriverDataSource`. For MySQL/Postgres this is harmless (unrecognized connection properties are silently ignored by both drivers), but it's an avoidable double-purpose use of the same key and could surprise a driver that validates unknown connection properties strictly.

## 1.2 Compatibility Impact

**Fully compatible.** No `Option` renamed or removed, no default value changed. The new branch only activates when a user explicitly sets `properties.connectionTestQuery` or `properties.connection-test-query`; for every job that doesn't (the overwhelming majority, and everyone on the current release), `testQuery` stays `null` and `applyConnectionValidation()` behaves exactly as before. Nobody could have depended on the old, broken `addDataSourceProperty()`-only behavior actually taking effect, since it was a no-op for pool-level validation — so there is no regression surface for existing users who happened to set these keys already (their query silently did nothing before; now it works).

Negligible: one or two `Map.get()` calls per pool creation (once per multi-table sink writer group, not per row/batch). Switching Hikari to query-based validation (when the user opts in) trades a lightweight SQL round trip at connection-borrow time for stronger dead-connection detection — a standard, expected HikariCP trade-off the user is explicitly asking for by setting this key. No resource leak, no new threads, no change to batch/commit/retry semantics.

No new exception paths are introduced; the method still can't throw (it only guards with null checks). Formal issues:

| 1 | Fix only updates the HikariCP pool config; `JdbcConnectionValidationUtils.getConnectionValidationQuery()` — the single source consulted by `SimpleJdbcConnectionProvider`, `SimpleJdbcConnectionPoolProviderProxy.isConnectionValid()`, `JdbcOutputFormat.shouldRefreshExecutor()`, `DdsqlJdbcConnectionPoolProviderProxy`, and `HiveJdbcConnectionProvider` — is untouched, so those checks still ignore the user's custom test query and fall back to `Connection.isValid()`. | `JdbcConnectionValidationUtils.java:65-77`; call sites `JdbcOutputFormat.java:233-238`, `SimpleJdbcConnectionPoolProviderProxy.java:47-51` | Medium | No |
| 2 | Fix is scoped to exactly two key spellings of `connectionTestQuery`; other genuine HikariCP pool settings the same `properties` block is used for today (per the linked issue's own repro: `validation-timeout`, `max-lifetime`, `leak-detection-threshold`) remain silently dropped via `addDataSourceProperty()`, so the PR will not fully resolve #11769's original scenario if those also mattered. | `JdbcSinkWriter.java:104, 120-126` | Medium | No |
| 3 | The extracted `connectionTestQuery`/`connection-test-query` entry is not removed from `properties` before the unconditional `forEach(ds::addDataSourceProperty)` on the next line, so the same key is applied twice through two different mechanisms. | `JdbcSinkWriter.java:104` | Low | No |
| 4 | No unit test added for the new properties-driven branch (camelCase key, kebab-case key, precedence when both keys or the Xugu default are present). The existing `JdbcSinkWriterTest` (`JdbcSinkWriterTest.java:24-63`) only covers the Xugu-hardcoded case and the "no query" default case; neither exercises the code this PR adds. | `seatunnel-connectors-v2/connector-jdbc/src/test/java/.../sink/JdbcSinkWriterTest.java` | Medium | No |

# 2. Code Quality Assessment

## 2.1 Coding Standards

The updated Javadoc on `applyConnectionValidation()` is accurate and describes both responsibilities now handled by the method. The two-key lookup is straightforward but hardcodes two literal strings inline instead of named constants (the file already has a precedent for this — `JdbcConnectionValidationUtils.XUGU_VALIDATION_QUERY` is a public named constant); extracting `PROPERTIES_KEY_CONNECTION_TEST_QUERY`/`_KEBAB` would make the two spellings greppable and reusable if Issue 1 above is later addressed by centralizing this lookup into `JdbcConnectionValidationUtils`. Minor, non-blocking.

**CI is currently red on this exact file.** The fork's `Run / Code style` job fails with `spotless:check`, specifically a line-wrap violation in the new Javadoc on `applyConnectionValidation()` (`src/main/java/.../sink/JdbcSinkWriter.java` — the two-line vs one-line wrapping of "Connection.isValid(timeout) probe, and honors user-specified connection test queries from the properties configuration block."). This is mechanical and fixed by running `spotless:apply`, but it must be fixed and pushed before merge — the apache-side "Build" check is currently `failure` (see CI section below).

No test files changed in this PR (`git diff origin/dev...HEAD --stat` shows a single file, `JdbcSinkWriter.java`, +12/-1). The existing `JdbcSinkWriterTest` has exactly the right shape to extend — two tests already call `JdbcSinkWriter.applyConnectionValidation(dataSource, jdbcConnectionConfig)` directly and assert on `dataSource.getConnectionTestQuery()` — so adding one or two more cases (properties map with `connectionTestQuery`, properties map with `connection-test-query`, and ideally a precedence case) would be a very small addition given the pattern is already established. No UT/E2E test code was modified, so the mandatory flaky-test anti-pattern analysis does not apply here; **Test Stability: N/A (no test changes)**.

## 2.3 Documentation Updates

`docs/en/connectors/sink/Jdbc.md:354-356` already documents `properties` generically as "Additional connection configuration parameters" with driver-vs-URL precedence notes; it doesn't specifically call out `connectionTestQuery`/`connection-test-query` as a special key that's now extracted and treated differently from a normal pass-through property (per Issue 3, the key is still also passed through as a raw property). Not required for this narrowly-scoped fix, but worth a one-line note if Issue 1/2 are addressed in a follow-up that formalizes which `properties` keys are Hikari-native vs. driver-native.

# 3. Architectural Soundness

**Temporary workaround**, not a long-term solution. It correctly fixes the literal symptom named in the PR title (`connection-test-query` is now honored by the Hikari pool), but per Issues 1 and 2 it leaves the broader design problem in place: `properties` conflates "raw JDBC driver connection properties" with "HikariCP pool-tuning properties," and this PR resolves that conflation for exactly one key rather than for the class of keys HikariCP itself supports. A more complete fix would either (a) route the `properties` map through HikariCP's own `PropertyElf`-style bean-property binding against the `HikariConfig`/`HikariDataSource` object for known Hikari fields (falling back to `addDataSourceProperty` only for genuinely unrecognized keys), or (b) explicitly enumerate and handle the small set of Hikari pool settings users are realistically going to try (test query, validation timeout, max lifetime, leak detection threshold), and centralize that resolution in `JdbcConnectionValidationUtils` so every `isConnectionValid()`/`isConnectionAlive()` call site in the connector agrees. Given the PR's stated scope (title says "connection-test-query" specifically), the current minimal fix is defensible as Option A; Option B would be the more complete fix a maintainer should consider requesting as a fast follow-up rather than blocking this PR on it, since Option A is strictly additive and doesn't need to be reverted to build Option B later.

## 3.2 Maintainability

The change is small and localized to a single already-well-tested method, which is good. The main maintainability cost is the risk flagged in Issue 1: the connector now has two places that each independently decide "is this driver's connection alive," one of which knows about user overrides and one of which doesn't, with no code-level link between them — a future contributor extending one is likely to miss the other, as this PR itself demonstrates.

## 3.3 Extensibility

Adding support for another Hikari-native `properties` key (e.g. `maxLifetime`) would currently require repeating the same `getOrDefault`-style two-key lookup pattern inline in `applyConnectionValidation()`. If more keys are added over time, this method will grow into an ad hoc mapping table; worth factoring into a small key→setter map now if the maintainers expect more of these (not blocking for this PR's stated scope).

No serialized state, checkpoint, or protocol format is touched. `JdbcConnectionConfig` isn't part of any checkpoint/savepoint payload (it's rebuilt from the job config on every restart), so there is no restore-path compatibility concern. Existing saved job configs continue to parse identically; this only changes runtime behavior of the `properties` map's `connectionTestQuery`/`connection-test-query` entries, which were previously inert.

# 4. Issue Summary

| 1 | Fix only reaches HikariCP's pool-level validation; `JdbcConnectionValidationUtils` (used by `JdbcOutputFormat.shouldRefreshExecutor()`, `SimpleJdbcConnectionPoolProviderProxy`, `DdsqlJdbcConnectionPoolProviderProxy`, `HiveJdbcConnectionProvider`) still ignores the user's custom test query | `JdbcConnectionValidationUtils.java:65-77`, `JdbcOutputFormat.java:233-238` | Medium |
| 2 | Fix covers only `connectionTestQuery`; other Hikari pool settings from the same `properties` block (per the linked issue's own repro: `validation-timeout`, `max-lifetime`, `leak-detection-threshold`) remain silent no-ops | `JdbcSinkWriter.java:104, 120-126` | Medium |
| 3 | Extracted key is not removed from `properties` before it's re-applied via `addDataSourceProperty` | `JdbcSinkWriter.java:104` | Low |
| 4 | No unit test added for the new branch despite an existing, directly-extensible test class for this exact method | `JdbcSinkWriterTest.java` | Medium |
| 5 | CI `Run / Code style` (spotless) fails on this file's new Javadoc line wrapping | `JdbcSinkWriter.java` (Javadoc, lines 109-111) | High |

# 5. Merge Recommendation

   - Issue 5 (High): CI is currently failing (`Run / Code style` / `spotless:check`) on exactly the file this PR touches. Run `./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-jdbc -am -nsu` and push the reformatted file.
   - Issue 4 (Medium, functionally a blocker for a bug-fix PR): please add at least one unit test to `JdbcSinkWriterTest` covering the new `properties.connectionTestQuery`/`connection-test-query` branch — the two existing tests in that file are an almost-copy-paste template for this.

   - Issue 3 (Low): strip the extracted key out of the `properties` map (or filter it in the `forEach`) before forwarding to `addDataSourceProperty`, so the same value isn't applied through two mechanisms.
   - Issue 1 & 2 (Medium, but reasonable to defer): consider a fast-follow that centralizes "effective validation query" resolution in `JdbcConnectionValidationUtils` so all `isConnectionValid()` call sites agree, and/or extends the same properties-to-pool-setting bridging to the other Hikari-native keys the linked issue's reporter actually configured (`validation-timeout`, `max-lifetime`, `leak-detection-threshold`). Option A (ship the narrow fix now, exactly as written, plus the two blockers above) is fine for this PR; Option B (the fuller properties-to-HikariConfig bridge) is better filed as a follow-up issue referencing #11769 so it isn't lost, since fixing only `connection-test-query` may leave the original reporter still seeing pool churn from the other settings they tried to configure.

Root-cause diagnosis is correct, the fix is in the right file and method, and it is safely backward compatible — nothing here should block on redesigning the whole `properties` handling. But please fix the CI failure and add the test before merge; happy to take another look once those land.

davidzollo:

```

```

DanielLeens:
Hi @waterWang, thanks for joining the Slack channel!

Quick status update since there's no new commit yet: the root-cause diagnosis and the fix itself are correct and safely backward compatible — nice work tracing it down to `applyConnectionValidation()`. Two items from my previous review are still blocking merge:

1. **CI (High, currently the literal reason the `Code style` check is red)** — `spotless:check` is failing on the new Javadoc line-wrapping in `JdbcSinkWriter.java`. Running `./mvnw spotless:apply -pl seatunnel-connectors-v2/connector-jdbc -am -nsu` locally and pushing the reformatted file should clear it.
2. **Test coverage (Medium)** — `JdbcSinkWriterTest` already has two tests that call `applyConnectionValidation(dataSource, jdbcConnectionConfig)` directly and assert on `getConnectionTestQuery()`; adding one more case covering the new `properties.connectionTestQuery` / `connection-test-query` branch would be a very small, almost copy-paste addition given that existing template.

The remaining notes from my review (extracted key still double-applied via `addDataSourceProperty`, and the fact that `JdbcConnectionValidationUtils` — used by `JdbcOutputFormat.shouldRefreshExecutor()` and others — still doesn't see this override) are non-blocking and fine to leave as a possible fast-follow; they don't need to hold up this PR.

Whenever you get a chance to push the Spotless fix and the test, I'll re-review right away. Thanks again for digging into #11769!

PR #11775, opened 2026-08-12
[Feature][Zeta] Periodically dump per-job metrics to engine log and seatunnel-metrics.log

corgy-w:
## What

Adds server-side periodic dumping of every running job's comprehensive `JobMetrics` into the engine log, plus a dedicated rolling log file `seatunnel-metrics.log`. The existing `print-job-metrics-info-interval` engine config (default `60`, `<=0` disables) drives the period — no new user-facing options.

## Why

Today the only ways for an operator to see in-flight job metrics are:
- the client-side `JobMetricsRunner` (only prints 8 read/write summary fields, only runs in the CLI process, gone once the client disconnects);
- `GET /running-jobs/{id}/metrics` over REST;
- the Web UI.

This adds a *passive*, log-only stream that captures the **full per-task `JobMetrics`** (every `SourceReceivedCount`, `SinkWriteCount`, `SinkCommittedCount`, etc.) on every node running the master, with no client dependency, suitable for tail-grepping or shipping into log aggregation.

## Scope

- Source/Transform/Sink coverage is automatic — `JobMetrics` already aggregates every connector's metrics via the existing `SeaTunnelMetricsContext` → `GetTaskGroupMetricsOperation` → master pipeline, no connector/API/SPI change required.
- No new Option, no new config file. `EngineConfig.printJobMetricsInfoInterval` and `ServerConfigOptions.PRINT_JOB_METRICS_INFO_INTERVAL` already exist; the server side just consumes them now.
- Backward compatible. `interval <= 0` disables. Default `60` is unchanged.

## Changes

| File | Change |
|---|---|
| `config/log4j2.properties` | +`metricsAppender` (`seatunnel-metrics.log`, time+size rolling) + `logger.metrics` named logger (`additivity=false`) |
| `deploy/kubernetes/seatunnel/conf/log4j2.properties` | Mirror |
| `seatunnel-engine/seatunnel-engine-server/.../metrics/PeriodicJobMetricsLogger.java` | New @Slf4j utility (`logJobMetrics(jobId, JobMetrics)`) |
| `seatunnel-engine/seatunnel-engine-server/.../CoordinatorService.java` | +`metricsLoggingScheduler` field + `startPeriodicJobMetricsLogger()` + `logAllRunningJobMetrics()`; lifecycle managed alongside `pipelineCleanupScheduler` |
| `seatunnel-engine/seatunnel-engine-server/.../metrics/PeriodicJobMetricsLoggerTest.java` | 3 smoke tests (no log4j-core dependency) |
| `docs/{en,zh}/engines/zeta/{separated,hybrid}-cluster-deployment.md` | New `print-job-metrics-info-interval` paragraphs (en + zh) |

## Output format

```
=== Job <id> metrics (<local timestamp>) ===
<JobMetrics.toString()>
```

`JobMetrics.toString()` already sorts by metric name then task name (see `seatunnel-api/.../JobMetrics.java:188-218`), so consecutive snapshots are diffable.

## Verification

- `mvn spotless:apply` clean.
- `mvn test -pl seatunnel-engine/seatunnel-engine-server -Dtest='PeriodicJobMetricsLoggerTest'`: **3/3 pass**.
- `mvn test -pl seatunnel-engine/seatunnel-engine-server -Dtest='CoordinatorServiceTest,PeriodicJobMetricsLoggerTest'`: **34/34 pass** (the 2 skipped tests in CoordinatorServiceTest are pre-existing and unrelated).
- Manual: start a cluster with a long-running job, check `logs/seatunnel-engine-master.log` and `logs/seatunnel-metrics.log` for periodic entries.

## Compatibility

- No public API/SPI/config-schema change.
- No connector / API / SPI code touched.
- `JobEventLocalFileHandler` and other telemetry subsystems unaffected.
- Existing client-side `JobMetricsRunner` left in place (local mode still benefits from it).

## Out of scope (follow-ups)

- per-job interval override (`EnvCommonOptions`).
- configurable metrics log dir / file name.
- deleting client-side `JobMetricsRunner`.


DanielLeens (commented):

The PR adds a background mechanism on the Zeta master node that periodically dumps the `JobMetrics` of every currently-running job to a new, dedicated `seatunnel-metrics.log` file (and, per the PR title/docs, is intended to also surface in the main engine log), driven by the interval already exposed as `print-job-metrics-info-interval` in `EngineConfig`. This gives operators a lightweight, always-on way to see per-job throughput/committed-count trends over time without having to poll the REST API or CLI.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**New scheduling path** (`CoordinatorService`):

```java
// constructor, after pipelineCleanupScheduler is wired up
startPeriodicJobMetricsLogger();

private void startPeriodicJobMetricsLogger() {
    int intervalSeconds = engineConfig.getPrintJobMetricsInfoInterval();
    if (intervalSeconds <= 0) {
        ...disabled, no scheduler created...
        return;
    }
    metricsLoggingScheduler = Executors.newSingleThreadScheduledExecutor(...daemon...);
    metricsLoggingScheduler.scheduleAtFixedRate(
        this::logAllRunningJobMetrics, intervalSeconds, intervalSeconds, TimeUnit.SECONDS);
}

private void logAllRunningJobMetrics() {
    if (!isActive) return;
    Set<Long> runningJobIds = runningJobMasterMap.keySet();
    if (runningJobIds.isEmpty()) return;
    long timeoutMs = TimeUnit.SECONDS.toMillis(engineConfig.getPrintJobMetricsInfoInterval());
    Map<Long, JobMetrics> jobMetricsMap;
    try {
        jobMetricsMap = getRunningJobMetrics(runningJobIds, timeoutMs);   // existing 3-arg overload
    } catch (Throwable t) {
        logger.warning(...);
        return;
    }
    for (Long jobId : runningJobIds) {
        JobMetrics jobMetrics = jobMetricsMap.get(jobId);
        PeriodicJobMetricsLogger.logJobMetrics(jobId, jobMetrics);
    }
}
```

Full runtime path: `CoordinatorService()` constructor -> `startPeriodicJobMetricsLogger()` (once per node, per `CoordinatorService` instance) -> single-thread `ScheduledExecutorService` "seatunnel-job-metrics-logger-%d" (daemon) -> `logAllRunningJobMetrics()` every `intervalSeconds` -> `getRunningJobMetrics(Set<Long>, long)` (CoordinatorService.java:1656-1704, pre-existing, unmodified) -> per-worker `NodeEngineUtil.sendOperationToMemberNode(..., new GetMetricsOperation(...), address).get(timeoutMs, ...)` -> `PeriodicJobMetricsLogger.logJobMetrics(jobId, jobMetrics)` -> `log.info(...)` -> Log4j2 `metricsAppender` (additivity=false).

This is a genuinely well-structured piece of scheduling: the scheduler is a per-node singleton created once in the constructor (not re-created on every master-active flip), gated cheaply per-tick by the existing `volatile boolean isActive`, and shut down correctly in `shutdown()` (`CoordinatorService.java:2040-2046`) alongside the other node-lifetime schedulers. It does not run on any job execution thread, does not acquire any lock shared with `JobMaster`/`PhysicalPlan`, and does not touch checkpoint state — so it cannot directly stall task execution or checkpointing. I verified `clearCoordinatorService()` (the master-failover teardown path, `CoordinatorService.java:1245-1293`) intentionally does **not** touch `metricsLoggingScheduler` — that is correct, since the scheduler is scoped to the node's lifetime (same pattern as `masterActiveListener`/`pipelineCleanupScheduler`), not to a single "active master" session; only the final `shutdown()` disposes of it, and it does so correctly.

That said, I found two concrete correctness issues while tracing the actual runtime behavior end-to-end (both detailed as Issue 1 and Issue 2 below), plus two design/documentation issues worth fixing before merge.

## 1.2 Compatibility Impact

**Fully compatible.** The change is purely additive: a new class (`PeriodicJobMetricsLogger`), new private scheduling methods, a new field, and new log4j2 appender/logger blocks appended to the existing properties files. It reuses the pre-existing `print-job-metrics-info-interval` config option (`ServerConfigOptions.MasterServerConfigOptions.PRINT_JOB_METRICS_INFO_INTERVAL`, default `60`) which already existed on `dev` before this PR (confirmed: `EngineConfig.java`, `YamlSeaTunnelDomConfigProcessor.java`, `ServerConfigOptions.java` are **not** part of this PR's diff — `git diff origin/dev...HEAD --stat` only touches `CoordinatorService.java`, the new `metrics` package, log4j2 config, and docs). No existing option is removed or renamed, no default is changed, no wire/serialization format changes. Checkpoint/savepoint/restore paths are untouched.

- **No contention with job execution threads.** The scheduler runs on its own daemon thread and only reads `runningJobMasterMap` (a `ConcurrentHashMap`, weakly-consistent iteration, safe) plus issues RPCs via the existing `GetMetricsOperation` path — the same mechanism already used by the REST/CLI metrics query. No new lock is introduced on `JobMaster`/`PhysicalPlan`/checkpoint state.
- **Sequential, per-worker-timeout metrics fetch on a single background thread (Issue 3).** `logAllRunningJobMetrics()` reuses the pre-existing `getRunningJobMetrics(Set<Long>, long)` overload (`CoordinatorService.java:1656-1704`), which loops over worker `addresses` **sequentially**, and for each one calls `.get(timeoutMs, TimeUnit.MILLISECONDS)` where `timeoutMs` is the **full periodic interval**, applied **per worker**, not as a shared budget. With N workers and even one slow/unreachable worker, a single cycle can take up to `N * intervalSeconds`. Because the scheduler is single-threaded and driven by `scheduleAtFixedRate` (which never runs two invocations concurrently — a still-running cycle simply delays the next tick), a cluster with several unhealthy workers can make the periodic dump fall progressively further behind, degrading the feature's usefulness without crashing anything. Notably, the class already contains a **correct** solution to this exact problem one screen down: `getRunningJobRawMetrics()` (`CoordinatorService.java:1706-1764`) explicitly documents "`timeoutMs` is treated as an overall time budget for the whole fetch (not per worker) ... avoids the master-side collector thread being blocked for `N * timeoutMs` in large clusters" and implements it via `metricsFetchExecutor.invokeAll(tasks, budgetMs, TimeUnit.MILLISECONDS)`. This PR is the *first* caller anywhere in the codebase to invoke the sequential 3-arg `getRunningJobMetrics` from a recurring background loop (I grepped all callers; the only other user is the one-shot, request-triggered `GetRunningJobMetricsOperation`), so this is a newly-introduced usage pattern, not a pre-accepted risk. See Issue 3.
- **Log volume.** `JobMetrics.toString()` (`seatunnel-api/.../JobMetrics.java:188-218`) renders one line per `(metric name, task vertex, measurement)` triple, sorted. For a job with many metric names (SourceReceivedCount/Bytes, SinkWriteCount/Bytes, SinkCommittedCount, etc.) and wide parallelism, a single dump can be many hundreds of lines, repeated per job per cycle. This is mitigated reasonably well: it is opt-in/configurable interval, isolated to its own rolling file with the same size/time/count retention policy as the main file appender, and does not touch the main engine log (see Issue 4 for why that isolation is actually stronger than intended). I would not block on this, but it is worth calling out that there is no per-job or per-metric sampling/cap, so very large/wide clusters should size `${file_split_size}`/`${file_count}` accordingly — not new information, just worth stating explicitly in the docs alongside the interval description.

- **Issue 1 (High):** documented "disable via 0" is unreachable — `EngineConfig.setPrintJobMetricsInfoInterval(0)` throws.
- **Issue 2 (High):** unguarded per-job logging loop can silently and permanently kill all future scheduler ticks.
- **Issue 3 (Medium):** sequential per-worker timeout in the periodic fetch path (see 1.3).
- **Issue 4 (Medium):** docs/PR title claim vs. actual `additivity=false` log4j2 routing; WARN-level fetch-failure signal is buried.

See Section 4 for the full table with locations and severities.

# 2. Code Quality Assessment

## 2.1 Coding Standards

New code is clean and consistently styled with the rest of `CoordinatorService`: constructor-first placement is respected (the new field is declared with the other scheduler fields, the new private methods are added right after `createCoordinatorExecutor()`), Javadoc is present on the new field, both new private methods, the new public `PeriodicJobMetricsLogger` class, and its one public method — this comfortably satisfies the "comment core methods/fields" requirement. ASF license headers are present on both new files. No wildcard imports, no `System.out.println`, log calls go through the class's existing `logger`/`log` fields as appropriate.

One nit I'd flag as part of Issue 2/3 fixes: the inline comment at `CoordinatorService.java:335-336` ("Use a generous timeout but never block the scheduler forever; failures are logged per-job.") does not match what the code actually does — the timeout is per-worker, not a shared budget (so it *can* effectively block a cycle for a long multiple of the interval), and "failures are logged per-job" is only true for the initial fetch, not for the trailing `logJobMetrics` loop. Worth correcting the comment together with the code fix so it doesn't mislead the next person who touches this method.

`PeriodicJobMetricsLoggerTest` (68 lines, 3 tests) is a reasonable smoke test for the leaf logging helper: non-null metrics, null metrics (warn branch), and empty metrics all "does not throw." The class-level Javadoc is honest about its own limits (no `log4j-core` on the test classpath, so appender routing isn't verified here). That part is fine as far as it goes.

What's missing is coverage of the actual new logic in `CoordinatorService` — the scheduling itself. This codebase already has an established, directly-analogous precedent for testing exactly this class of feature: `CoordinatorServicePipelineCleanupTest.java` (433 lines, 8 `@Test` methods) thoroughly exercises the sibling `pipelineCleanupScheduler`/`processPendingPipelineCleanup` background task (lock/race handling, idempotency, `isActive` gating, etc.). No equivalent test exists for `startPeriodicJobMetricsLogger()` / `logAllRunningJobMetrics()`: nothing asserts that a non-positive interval leaves `metricsLoggingScheduler == null` and schedules nothing (which, per Issue 1, is currently dead code and would be caught immediately by such a test), nothing asserts the `isActive`/empty-`runningJobMasterMap` early-return branches, and nothing asserts that an exception thrown while logging one job doesn't take down the rest of the cycle (which, per Issue 2, it currently does). Given the precedent sitting right next to this change in the same package, I'd ask for at least a small `CoordinatorService`-level test class before merge.

**Test stability rating: Stable.** The three existing tests are pure, deterministic, single-threaded calls into static/leaf methods with no `Thread.sleep`, no shared static state, and no timing assumptions — no flakiness risk from what's there. The rating reflects what exists, not what's missing (coverage gap is tracked separately above).

## 2.3 Documentation Updates

`docs/en` and `docs/zh` are both updated in the two relevant deployment guides (`hybrid-cluster-deployment.md`, `separated-cluster-deployment.md`), and the English/Chinese text is consistent with each other and with the actual default value (`60`) and option key (`print-job-metrics-info-interval`). Good discipline on the "docs match code" front for everything except the specific engine-log-routing claim covered in Issue 4.

# 3. Architectural Soundness

Precise fix in spirit (a purpose-built named logger + dedicated rolling file is the right approach for this kind of high-volume, low-priority operational log), but the implementation reaches for the wrong existing helper for the fetch step when a better one already exists in the same class (Issue 3), and the per-job logging loop lacks the same defensive `try/catch` its sibling scheduled task (`cleanupPendingPipelines`) already uses for exactly this failure mode. I'd call this **Precise fix, with two implementation gaps** rather than a temporary workaround — the overall design (singleton per-node scheduler, `additivity=false` isolation, doc updates in both languages) is sound and the fixes needed are localized, not structural.

## 3.2 Maintainability

Good: the new logic is isolated to two small, well-named private methods plus a small dedicated class, following the existing conventions in `CoordinatorService` closely (same `ThreadFactoryBuilder` style, same shutdown-list pattern). Once Issues 1-3 are fixed, this will be easy for the next contributor to reason about.

## 3.3 Extensibility

Reasonable. If a future need arises for a shared time budget or metric-name filtering, the pattern already exists in `getRunningJobRawMetrics` and could be reused (see Issue 3).

No impact on checkpoint/savepoint state, job submission protocol, or serialized formats. The config option itself pre-dates this PR (already on `dev` with the same default), so there is no upgrade-path concern for existing deployments beyond the new log file appearing on disk.

# 4. Issue Summary

| 1 | `print-job-metrics-info-interval: 0` is documented (by this PR, in `docs/en(/zh)/.../*-cluster-deployment.md`) as the way to disable the feature, and `startPeriodicJobMetricsLogger()` explicitly branches on `intervalSeconds <= 0` to skip scheduler creation — but `EngineConfig.setPrintJobMetricsInfoInterval(int)` (pre-existing, `EngineConfig.java:153-158`) calls Hazelcast's `Preconditions.checkPositive(...)`, which throws `IllegalArgumentException` for any value `<= 0`. Since `YamlSeaTunnelDomConfigProcessor.parseEngineConfig()` (`YamlSeaTunnelDomConfigProcessor.java:170-178`, pre-existing) unconditionally calls that setter whenever the YAML key is present, a user who follows this PR's own new documentation and sets `print-job-metrics-info-interval: 0` will crash engine startup with an `IllegalArgumentException` instead of disabling the feature. In practice this makes the `<= 0` branch in `CoordinatorService` dead code reachable only via non-public paths. | `docs/en/engines/zeta/hybrid-cluster-deployment.md` (new text) + `seatunnel-engine/.../CoordinatorService.java:298-305` + `seatunnel-engine-common/.../config/EngineConfig.java:153-158` + `.../config/YamlSeaTunnelDomConfigProcessor.java:170-178` | High |
| 2 | `logAllRunningJobMetrics()`'s per-job logging loop (`for (Long jobId : runningJobIds) { ... PeriodicJobMetricsLogger.logJobMetrics(jobId, jobMetrics); }`) is not wrapped in a `try/catch`, unlike the fetch step immediately above it and unlike the sibling `cleanupPendingPipelines()` task in the very same class (which wraps its whole loop body in `catch (Throwable t)`). Per the JDK `ScheduledExecutorService.scheduleAtFixedRate` contract, an uncaught exception from *any* single invocation permanently and silently cancels **all future executions** of that task for the remaining life of the node process — with no log line, no alert, nothing. This directly contradicts the method's own Javadoc ("Best-effort: a failure fetching one job's metrics does not prevent the others from being logged.") and the inline comment above it ("failures are logged per-job"): today neither claim holds for this loop. | `seatunnel-engine/.../CoordinatorService.java:347-350` | High |
| 3 | `logAllRunningJobMetrics()` calls the pre-existing sequential `getRunningJobMetrics(Set<Long>, long)` overload, which iterates worker `addresses` one at a time and applies the **full periodic interval as a per-worker timeout** rather than a shared budget. With N workers and any slow/unreachable ones, a single cycle on the single-threaded `metricsLoggingScheduler` can take up to `N * intervalSeconds`, causing the feature to fall progressively behind under real cluster degradation. The same class already contains the fix for exactly this problem in `getRunningJobRawMetrics()` (parallel dispatch via `metricsFetchExecutor.invokeAll(tasks, budgetMs, ...)` with an overall time budget) — recommend reusing that pattern (or the raw-metrics variant) for the periodic logger instead of the sequential overload. | `seatunnel-engine/.../CoordinatorService.java:337-346` (calling `CoordinatorService.java:1656-1704`) | Medium |
| 4 | The PR title ("dump per-job metrics to engine log and seatunnel-metrics.log") and the new docs text ("dumps ... into the engine log. The same content is also written to a dedicated log file...") both claim the output reaches the regular engine log. In reality, `config/log4j2.properties` and `deploy/kubernetes/.../log4j2.properties` set `logger.metrics.additivity = false` for the `PeriodicJobMetricsLogger` logger, so entries are routed **only** to `metricsAppender`/`seatunnel-metrics.log` and never reach the main engine log or console — which matches the class-level Javadoc on `PeriodicJobMetricsLogger` ("never pollute the regular engine file/console logs") but contradicts the PR title and the deployment docs. A secondary consequence: `logger.metrics.level = INFO` means the `log.warn("Skip logging metrics for job {}: metrics snapshot was null", ...)` branch in `PeriodicJobMetricsLogger.logJobMetrics` also gets swallowed by the same `additivity=false` logger, so a real metrics-fetch failure signal ends up buried in the low-traffic metrics file instead of the main engine log operators actually watch. | `docs/en(/zh)/engines/zeta/{hybrid,separated}-cluster-deployment.md` vs `config/log4j2.properties` (new `logger.metrics.additivity = false`) vs `PeriodicJobMetricsLogger.java:34-35,60` | Medium |
| 5 | No test exercises the new `CoordinatorService` scheduling logic itself (only the leaf `PeriodicJobMetricsLogger.logJobMetrics` static method is smoke-tested). This codebase already has a direct precedent for testing this exact class of feature — `CoordinatorServicePipelineCleanupTest.java` (433 lines / 8 tests) for the sibling `pipelineCleanupScheduler`. None of the following are covered: disabled-at-`<=0` (would have caught Issue 1), `isActive`/empty-map early returns, or per-job exception isolation (would have caught Issue 2). | `seatunnel-engine/.../test/.../metrics/PeriodicJobMetricsLoggerTest.java` (missing `CoordinatorService`-level equivalent) | Medium |

No other reviewers or maintainers have commented on this PR yet (`gh api .../pulls/11775/reviews` and `.../comments` both return empty), so there is nothing to reconcile here.

# 5. Merge Recommendation

   - Issue 1: either make `EngineConfig.setPrintJobMetricsInfoInterval` accept `0`/negative as "disabled" (e.g. `checkArgument(value >= 0 || value == <sentinel>, ...)` or simply drop the `checkPositive` floor to `>= 0` for this specific setter), or change the docs/behavior so "disable" is documented as omitting the key / using a value that the existing validation actually accepts. Either is fine; the current state where the documented instruction crashes startup is not.
   - Issue 2: wrap the per-job `logJobMetrics` loop body (or the whole `logAllRunningJobMetrics()` method) in the same `catch (Throwable t)` pattern already used by `cleanupPendingPipelines()`, so one bad job can't permanently kill the scheduler for the life of the node.

   - Issue 3: switch the periodic fetch to the budgeted/parallel `getRunningJobRawMetrics`-style path (or otherwise bound the worst case) instead of the sequential per-worker-timeout overload.
   - Issue 4: either genuinely mirror the output into the main engine log (drop/relax `additivity=false`, or add a second, separate WARN-level logger call from `CoordinatorService`'s own `logger` for fetch failures) or fix the PR title/docs to say "metrics log only," and keep the null-metrics warning on a channel that isn't `additivity=false`-suppressed.
   - Issue 5: add a `CoordinatorService`-level test covering the disabled-interval branch and per-job exception isolation, following the existing `CoordinatorServicePipelineCleanupTest` pattern.

Overall this is a well-intentioned, mostly well-built addition that follows the existing `CoordinatorService` scheduler conventions closely (correct daemon-thread singleton lifecycle, correct shutdown wiring, good bilingual docs). The two High-severity issues are both narrow, mechanical fixes (a validation floor and a missing `try/catch`) rather than design-level rework, so I'd expect this to be quick to turn around. Happy to take another pass once those land.

DanielLeens (commented):

- **User pain point:** Operators had no lightweight, always-on way to see per-job throughput/committed-count trends over time on the Zeta master without polling the REST API/CLI.
- **Fix approach:** A per-node singleton background scheduler (`CoordinatorService.metricsLoggingScheduler`) periodically fetches `JobMetrics` for every running job and writes one snapshot per job via a new `PeriodicJobMetricsLogger`, routed to a dedicated, independently-rotated `seatunnel-metrics.log` file through a new `additivity=false` Log4j2 logger/appender.
- **One-sentence summary:** Adds an opt-in (config-driven) periodic per-job metrics dump to a dedicated log file, isolated from the main engine log and from job execution.

This is a **re-review** of the latest commit `868ee240` ("Harden periodic job metrics logging"), which landed on 2026-08-13 after my prior review on 2026-08-12 of `aff43e16`. The author did not post any reply comments, but the new commit directly addresses every one of the 5 issues I raised previously. I re-read the entire current diff against `dev` (not just the delta) to verify each fix end-to-end and to check for anything newly introduced by the fix itself.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Status of my previous 5 issues after `868ee240`:**

- **Issue 1 (High, disable-via-0 crashes startup) — Fixed and tested.** `EngineConfig.setPrintJobMetricsInfoInterval` no longer calls Hazelcast's `checkPositive`; it now only rejects negative values (`< 0`), so `0` is accepted and `getPrintJobMetricsInfoInterval() == 0`. Verified with a new `YamlSeaTunnelDomConfigProcessorTest.testParseEngineConfigAllowsZeroJobMetricsInterval` that parses `print-job-metrics-info-interval: 0` end-to-end through the real YAML config path, plus a `CoordinatorServiceTest.testPeriodicJobMetricsLoggingCanBeDisabledWithZeroInterval` unit test. Docs were also corrected to say "Set to `0` to disable" (dropping the now-inaccurate "or any non-positive value").
- **Issue 2 (High, uncaught exception permanently kills the scheduler) — Fixed and tested.** The per-job loop in `logAllRunningJobMetrics()` (`CoordinatorService.java:352-364`) now wraps `PeriodicJobMetricsLogger.logJobMetrics(...)` in its own `try { ... } catch (Throwable t) { logger.warning(...) }`, matching the sibling `cleanupPendingPipelines()` pattern. Verified with `testPeriodicJobMetricsLoggingContinuesWhenSingleJobLoggingFails`, which mocks job 1's log call to throw and asserts job 2 is still logged — this is exactly the regression scenario I flagged, and it is now a real, deterministic (no `Thread.sleep`, no timing assumptions) unit test guarding it.
- **Issue 3 (Medium, sequential per-worker timeout can take up to N*interval) — Fixed.** `logAllRunningJobMetrics()` now calls a new `getRunningJobMetricsForPeriodicLogging(Set<Long>, long)` (`CoordinatorService.java:1710-1715`), which delegates to the pre-existing budgeted/parallel `getRunningJobRawMetrics(...)` (`CoordinatorService.java:1727-1775`, uses `metricsFetchExecutor.invokeAll(tasks, budgetMs, ...)`) instead of the sequential 3-arg overload. This is exactly the fix I recommended — the periodic dump can no longer blow past its own interval by a multiple of the worker count. The stale inline comment is also corrected to match: "Use the configured interval as an overall budget for the whole fetch, not per worker" (`CoordinatorService.java:335`).
- **Issue 4 (Medium, docs/title claim "engine log" but `additivity=false` routes only to the metrics file; null-metrics warning was buried) — Fixed.** All four deployment docs (en/zh × hybrid/separated) now say the dump goes "into the dedicated log file `seatunnel-metrics.log`" instead of "into the engine log ... also written to"; the `PeriodicJobMetricsLogger` class Javadoc and the `config/log4j2.properties`/k8s comment were updated to match. The null-snapshot warning was **moved out of** the `additivity=false` logger: `PeriodicJobMetricsLogger.logJobMetrics` no longer logs on null (silently returns), and the warning is now emitted from `CoordinatorService`'s own (non-suppressed) `logger.warning(...)` at `CoordinatorService.java:349-353` — so a real fetch failure now actually reaches the main engine log instead of being buried in the low-traffic metrics file. One residual, cosmetic nit: the **PR title itself** ("...to engine log and seatunnel-metrics.log") still contains the now-corrected claim; worth a one-line title edit before merge for consistency with the code/docs, but this is Low and not a code issue.
- **Issue 5 (Medium, no `CoordinatorService`-level test for the scheduling logic) — Partially addressed, no longer blocking.** See 2.2 below.

**Runtime path (mandatory, engine-touching change), re-verified against the new commit:**
`CoordinatorService()` constructor → `startPeriodicJobMetricsLogger()` (`CoordinatorService.java:297`, once per node, unchanged from before) → single daemon-thread `metricsLoggingScheduler` → `logAllRunningJobMetrics()` (`CoordinatorService.java:327`) every `intervalSeconds` → `getRunningJobMetricsForPeriodicLogging(runningJobIds, timeoutMs)` (`CoordinatorService.java:1710`) → `getRunningJobRawMetrics(...)` (`CoordinatorService.java:1727`) → **`metricsFetchExecutor.invokeAll(tasks, budgetMs, MILLISECONDS)`** (shared fixed thread pool, `CoordinatorService.java:1755`) → per-worker `fetchRawMetricsFromWorker` RPC → `mergeWithHistoricalJobMetrics` → back in `logAllRunningJobMetrics`, per-job `try/catch` around `PeriodicJobMetricsLogger.logJobMetrics(jobId, jobMetrics)` (`CoordinatorService.java:355-364`) → `log.info(...)` → Log4j2 `metricsAppender` (`additivity=false`). Shutdown wiring (`metricsLoggingScheduler.shutdownNow()` + `awaitSchedulerTermination(...)`, `CoordinatorService.java:2052/2059`) is unchanged from the version I reviewed before and remains correct.

**New finding from re-reviewing the fix itself (not present in the prior version, since it used the sequential overload and never touched `metricsFetchExecutor`):**

Fixing Issue 3 by switching to `getRunningJobRawMetrics` means the periodic logger now shares the **same fixed thread pool** (`metricsFetchExecutor`, sized `max(1, min(8, availableProcessors))`, `CoordinatorService.java:247-255`) that `RealtimeMetricsService` already uses for its own polling loop — `RealtimeMetricsService.POLL_INTERVAL_MS = 5000` / `METRICS_FETCH_TIMEOUT_MS = 3000` (`RealtimeMetricsService.java:79-80`, `scheduleWithFixedDelay` at line 132-133, fetch call at line 261). That service is pre-existing and out of this PR's diff, but this PR adds a **second, independent periodic consumer** of the same small bounded pool. Concretely: `invokeAll(tasks, budgetMs, ...)` submits one task per worker address to the shared pool; if `RealtimeMetricsService`'s 5s-cycle tasks are occupying the pool's threads when the (default 60s-interval) periodic-logger cycle fires, the logger's tasks queue behind them (and vice versa) for up to the respective budget. Both sides degrade gracefully (bounded by their own timeout, self-corrects next cycle, no deadlock/exception propagation), so this is not a correctness or stability bug — but on the exact deployment target this PR ships a `log4j2.properties` for (`deploy/kubernetes/seatunnel/conf/log4j2.properties`), containers are commonly provisioned with 1-2 vCPUs, where `metricsFetchExecutor` collapses to 1-2 threads and the contention window becomes proportionally larger. Recommend either giving the periodic logger its own small dedicated executor, or explicitly documenting/accepting that periodic metrics logging shares capacity with realtime observability polling under constrained CPU.

## 1.2 Compatibility Impact

**Fully compatible.** Still purely additive relative to `dev`: no removed/renamed option, the only behavioral change to a pre-existing setter (`EngineConfig.setPrintJobMetricsInfoInterval`) is *widening* its accepted range (`>0` → `>=0`), which is backward compatible (every previously-valid value is still valid; it just also now accepts `0`). `ServerConfigOptions.PRINT_JOB_METRICS_INFO_INTERVAL`'s key, type, and default (`60`) are unchanged — only its `withDescription(...)` string was reworded. No wire/serialization format changes, no checkpoint/savepoint/restore path touched.

- The sequential/N*timeout risk I flagged previously (Issue 3) is resolved — the fetch is now parallel and budget-bounded.
- New, non-blocking contention risk with `RealtimeMetricsService` over the shared `metricsFetchExecutor` under constrained CPU (see 1.1 above).
- Log volume / disk growth analysis from my prior review still holds and is unaffected by this commit: the new `metricsAppender` uses the same time+size rolling policy and count/age-based deletion as the main file appender (`config/log4j2.properties`), so disk usage is bounded the same way as the existing engine log; the file path is driven by the existing `${file_path}` property, so it's configurable, not hardcoded.
- No new lock is introduced on `JobMaster`/`PhysicalPlan`/checkpoint state; `runningJobMasterMap` remains a `ConcurrentHashMap`, and iterating its live `keySet()` (`CoordinatorService.java:331`) is weakly-consistent and safe even if jobs complete mid-cycle.

| # | Status |
|---|---|
| Issue 1 | **Fixed**, with a dedicated regression test at two levels (YAML-config-processor level and `EngineConfig` level). |
| Issue 2 | **Fixed**, with a dedicated regression test asserting per-job exception isolation. |
| Issue 3 | **Fixed** by switching to the budgeted/parallel fetch path. |
| Issue 4 | **Fixed**: docs/class Javadoc corrected to "dedicated log file" only; null-metrics warning moved to a non-suppressed logger. PR title still says "engine log" — Low, cosmetic, non-blocking. |
| Issue 5 | **Partially fixed** — see 2.2. |
| **Issue 6 (New, Medium)** | Periodic logger now shares the fixed-size `metricsFetchExecutor` pool with `RealtimeMetricsService`'s 5s-interval polling; on small (1-2 vCPU) containers — the exact target of this PR's k8s `log4j2.properties` change — this pool can shrink to 1-2 threads, making cross-feature queuing delays proportionally larger. Self-healing, not correctness-affecting, but worth a dedicated executor or an explicit docs callout. Location: `CoordinatorService.java:247-255` (pool sizing), `:1755` (shared `invokeAll`), `RealtimeMetricsService.java:79-80,132-133,261` (other consumer). |
| **Issue 7 (New, Low)** | `metricsLoggingScheduler` (`CoordinatorService.java:220`) is a plain (non-`final`) field, unlike its sibling `pipelineCleanupScheduler` (`CoordinatorService.java:213`, `final`) and `metricsFetchExecutor` (also `final`). Because it's conditionally assigned once in the constructor via `startPeriodicJobMetricsLogger()`, it could be made `final` too (e.g. by having that method return the scheduler-or-null and assigning the result directly), which would give it the same JMM safe-publication guarantee the other constructor-time fields already rely on. Purely a style/robustness nit — no evidence of an actual production visibility bug given how `CoordinatorService` instances are published in practice — but worth aligning with the established pattern in the same class. |

# 2. Code Quality Assessment

## 2.1 Coding Standards

Still clean and consistent with the surrounding class. New/changed methods (`getRunningJobMetricsForPeriodicLogging`, the reworked `logAllRunningJobMetrics`) have Javadoc; the previously-misleading inline comment is now corrected to accurately describe the budget semantics (see Issue 3 above — this was flagged as a coding-standards nit in my last review, and it's fixed together with the logic). No wildcard imports, no `System.out.println`, ASF headers intact on the new/touched files.

Two new `CoordinatorServiceTest` methods were added directly targeting the two High-severity issues from my last review:
- `testPeriodicJobMetricsLoggingCanBeDisabledWithZeroInterval` — asserts `setPrintJobMetricsInfoInterval(0)` no longer throws (would have caught Issue 1).
- `testPeriodicJobMetricsLoggingContinuesWhenSingleJobLoggingFails` — mocks `PeriodicJobMetricsLogger.logJobMetrics` (via `MockedStatic`) to throw for job 1 and asserts job 2 still gets logged (would have caught Issue 2). This is a deterministic, single-threaded assertion with no `Thread.sleep`/timing dependency.

Plus a `YamlSeaTunnelDomConfigProcessorTest` test exercising the real YAML → `EngineConfig` path for `print-job-metrics-info-interval: 0`.

This closes the two most important gaps from my prior review (the two former High-severity issues are now regression-tested, not just fixed). What's still *not* covered, matching my original ask for `CoordinatorServicePipelineCleanupTest`-equivalent coverage: the `isActive == false` and `runningJobIds.isEmpty()` early-return branches in `logAllRunningJobMetrics()`, and an assertion that `startPeriodicJobMetricsLogger()` leaves `metricsLoggingScheduler == null` (i.e. schedules nothing) when the interval is `<= 0` — today only the underlying `EngineConfig`/YAML acceptance of `0` is tested, not that `CoordinatorService` itself correctly no-ops. Given the two former High-severity risks are now directly guarded, I'm downgrading this from a Medium blocker-adjacent ask to a **Low, non-blocking** follow-up.

**Test stability rating: Stable.** Both new tests are deterministic, single-threaded, no `Thread.sleep`, no shared static state (the `MockedStatic` is scoped to a try-with-resources block), no order dependency.

## 2.3 Documentation Updates

`docs/en` and `docs/zh` were both updated in this commit to fix the "engine log" claim (see Issue 4). Text remains consistent between English and Chinese and with the actual code behavior and default (`60`).

# 3. Architectural Soundness

**Precise fix.** In my last review I called this "precise fix, with two implementation gaps" — both of those gaps (the missing per-job `try/catch` and the wrong fetch-helper choice) are now closed, and closed the same way the rest of the class already does things (matching `cleanupPendingPipelines()`'s exception handling, reusing the existing budgeted `getRunningJobRawMetrics`). The one thing this fix newly introduces — sharing `metricsFetchExecutor` with `RealtimeMetricsService` — is a reasonable trade-off (reusing an existing, well-tested budgeted-fetch primitive is clearly better than reinventing one), just worth a follow-up per Issue 6.

## 3.2 Maintainability

Improved further: the new `CoordinatorServiceTest` methods sit next to the existing scheduler tests and will make it obvious to the next contributor if either of the two hardening properties (disable-via-zero, per-job isolation) regresses.

## 3.3 Extensibility

Unchanged from my prior assessment — reasonable, and now actually exercising the extensibility point I called out (reuse of `getRunningJobRawMetrics`'s budgeted pattern).

No impact on checkpoint/savepoint state, job submission protocol, or serialized formats, same as before. The widened `EngineConfig` setter range is a strict superset of previously-accepted values, so no existing valid config becomes invalid.

# 4. Issue Summary

| 1 | `print-job-metrics-info-interval: 0` documented as "disable" but the setter threw `IllegalArgumentException` | `EngineConfig.java` (setter), docs | High | **Fixed + tested** |
| 2 | Uncaught exception in per-job logging loop permanently kills all future scheduler ticks | `CoordinatorService.java:352-364` | High | **Fixed + tested** |
| 3 | Sequential per-worker timeout could take up to N*interval per cycle | `CoordinatorService.java:1710-1715` (was calling the 3-arg sequential overload) | Medium | **Fixed** |
| 4 | Docs/title claimed "engine log" but `additivity=false` routes only to `seatunnel-metrics.log`; null-metrics warning was buried in the suppressed logger | docs (en/zh), `PeriodicJobMetricsLogger.java`, `log4j2.properties` | Medium | **Fixed** (docs/code); PR title still says "engine log" — Low cosmetic residue |
| 5 | No `CoordinatorService`-level test for the scheduling logic itself | `CoordinatorServiceTest.java` | Medium → **Low** | **Partially fixed** — core regressions (Issues 1 & 2) now tested; `isActive`/empty-map/scheduler-not-created branches still untested |
| 6 | Periodic logger now shares the small fixed `metricsFetchExecutor` pool with `RealtimeMetricsService`'s 5s polling loop; pool can shrink to 1-2 threads on constrained (e.g. k8s) deployments | `CoordinatorService.java:247-255,1755`; `RealtimeMetricsService.java:79-80,132-133,261` | Medium | **New**, non-blocking |
| 7 | `metricsLoggingScheduler` field is not `final`, unlike its sibling constructor-time scheduler fields in the same class | `CoordinatorService.java:220` | Low | **New**, style nit |

# 5. Merge Recommendation

1. **Blockers:** None. Both former High-severity issues (Issues 1 and 2) are fixed and now covered by dedicated, deterministic unit tests; Issue 3's performance concern is resolved by reusing the existing budgeted/parallel fetch path.
   - Issue 5: add the small amount of remaining `CoordinatorService`-level coverage (disabled-at-`<=0` scheduler creation, `isActive`/empty-map early returns) to fully close the gap with the `CoordinatorServicePipelineCleanupTest` precedent.
   - Issue 6: consider a small dedicated executor for the periodic logger (or explicitly document the shared-capacity trade-off with `RealtimeMetricsService`), particularly given this PR's own k8s deployment config targets CPU-constrained pods.
   - Issue 7: make `metricsLoggingScheduler` `final` for consistency with the class's other constructor-time scheduler/executor fields.
   - Update the PR title to drop the now-inaccurate "to engine log" phrasing, for consistency with the corrected docs/Javadoc.

This is a well-executed hardening pass: it directly and correctly addresses every issue from my prior review, adds real regression tests for the two High-severity ones rather than just fixing them silently, and the one new finding it surfaces (Issue 6) is a bounded, self-healing side effect of choosing the *better* fetch implementation, not a step backward. Nice turnaround.

DanielLeens (approved):

- **User pain point:** operators had no lightweight, always-on way to observe per-job throughput/committed-count trends on a Zeta master without polling the REST API or the CLI.
- **Fix approach:** a per-node singleton background scheduler (`CoordinatorService.metricsLoggingScheduler`) periodically fetches `JobMetrics` for every running job and writes one snapshot per job via a new `PeriodicJobMetricsLogger`, routed to a dedicated, independently-rotated `seatunnel-metrics.log` file through a new `additivity=false` Log4j2 logger/appender.
- **One-sentence summary:** adds an opt-in, config-driven periodic per-job metrics dump to a dedicated log file, isolated from the main engine log and from job execution.

This is a **re-review of the current head** (`5ad1c31a`). I re-read the full diff against `dev` from scratch rather than diffing against my last review. For context: I reviewed an earlier commit on this same "Harden periodic job metrics logging" change back on 2026-08-13 and concluded "Ready to merge" with three non-blocking follow-ups (a shared-executor contention note, a missing scheduler-disabled unit test, and a non-`final` field). The SHA on the branch has since changed (the PR was rebased onto a newer `dev`), but more importantly the content itself moved forward too: the field is now `final`, exactly addressing the smallest of those three follow-ups, and the duplicate merge-with-history logic was extracted into a shared private helper. Everything that was already fixed as of my last pass remains fixed. Nothing here was missed last time and only now surfacing — this is a genuine incremental improvement on top of an already-solid version.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Runtime path, re-traced end-to-end against the current head:

`CoordinatorService()` constructor -> `createJobMetricsLoggingScheduler()` (`CoordinatorService.java:291`, called once per node) -> if `print-job-metrics-info-interval <= 0`, returns `null` and no scheduler is created; otherwise a single daemon-thread `ScheduledExecutorService` ("seatunnel-job-metrics-logger-%d") is built and assigned straight into the now-`final` `metricsLoggingScheduler` field -> `logAllRunningJobMetrics()` fires every `intervalSeconds` -> early-return if `!isActive` or `runningJobMasterMap` is empty -> `getRunningJobMetricsForPeriodicLogging(runningJobIds, timeoutMs)` -> `getRunningJobRawMetrics(...)`, which dispatches one fetch task per worker address through the shared `metricsFetchExecutor` and bounds the whole batch with `invokeAll(tasks, budgetMs, ...)` (a real time budget, not a per-worker timeout) -> `mergeWithHistoricalJobMetrics(...)` merges in any historical/completed-task metrics from `jobHistoryService` -> back in `logAllRunningJobMetrics`, each job is logged inside its own `try/catch (Throwable)` around `PeriodicJobMetricsLogger.logJobMetrics(jobId, jobMetrics)` -> `log.info(...)` -> the `PeriodicJobMetricsLogger`-named Log4j2 logger, `additivity=false`, routed only to `metricsAppender` / `seatunnel-metrics.log`. Shutdown wiring (`metricsLoggingScheduler.shutdownNow()` plus `awaitSchedulerTermination(...)` in `CoordinatorService.shutdown()`) is unchanged and correct, and `clearCoordinatorService()` (the master-failover teardown path) still correctly leaves this scheduler alone, since it is scoped to the node's lifetime rather than to a single "active master" session.

This is a clean, low-risk addition: the scheduler doesn't touch any lock shared with `JobMaster`/`PhysicalPlan`, doesn't touch checkpoint state, and reuses an already-hardened, budgeted metrics-fetch path instead of inventing a new one.

Two small things worth flagging, neither of which changes the merge recommendation:

- `mergeWithHistoricalJobMetrics(...)` (new in this commit, `CoordinatorService.java:2494`) is now shared by the 3-arg `getRunningJobMetrics(Set, long, String[])` and the new periodic-logging path, which is a nice dedup — but the pre-existing no-arg `getRunningJobMetrics()` (`CoordinatorService.java:1744-1788`, untouched by this PR) still carries its own copy of the same merge loop inline. Not a regression introduced here (that duplication predates this PR), but since the helper now exists, it would be a natural, low-risk follow-up to point the no-arg overload at it too.
- The shared `metricsFetchExecutor` contention with `RealtimeMetricsService`'s independent 5-second polling loop (`RealtimeMetricsService.java:260-261`, calling the same `coordinatorService.getRunningJobRawMetrics(...)`) that I flagged in my last review is still present and still unchanged — see Issue 1 below for the full detail carried forward.

## 1.2 Compatibility Impact

**Fully compatible**, unchanged from my prior assessment. Purely additive relative to `dev`: no option removed or renamed, `ServerConfigOptions.PRINT_JOB_METRICS_INFO_INTERVAL`'s key/type/default (`60`) are untouched (only its description text was reworded). The one behavioral change to a pre-existing setter, `EngineConfig.setPrintJobMetricsInfoInterval`, only *widens* its accepted range (`> 0` -> `>= 0`), which is a backward-compatible superset — every previously-valid value stays valid, `0` is now additionally accepted to mean "disabled." No wire/serialization format change, no checkpoint/savepoint/restore path touched. I also checked the 7 commits `dev` has picked up since this PR's head (all connector/docs changes such as the PayPal source, LocalFile tailing, Firebase sink) — none of them touch `CoordinatorService.java`, `EngineConfig.java`, `RealtimeMetricsService.java`, or the `log4j2.properties` files, so the divergence is benign and doesn't affect this review's conclusions or create any rebase/merge risk.

- No contention with job execution threads: the scheduler runs on its own daemon thread, only reads the weakly-consistent `runningJobMasterMap.keySet()`, and issues RPCs through the same `GetMetricsOperation`/worker-fetch mechanism already used elsewhere.
- The N*timeout risk from the original sequential fetch approach is resolved by routing through the budgeted, parallel `getRunningJobRawMetrics`/`metricsFetchExecutor.invokeAll(...)` path instead of the legacy per-worker-timeout overload.
- Log volume is bounded by the same time+size rolling/retention policy as the main engine log appender (`${file_split_size}`/`${file_count}`/`${file_ttl}`), and isolated to its own file via `additivity=false`, so it cannot crowd out or interleave with the main engine log.
- Carried-forward, non-blocking risk: see Issue 1 (shared `metricsFetchExecutor` with `RealtimeMetricsService`).

Both fetch and per-job logging are defensively wrapped: a failure fetching metrics for the whole batch is caught, logged via `logger.warning(...)`, and simply skips that cycle; a failure logging one specific job's metrics is caught per-job so one bad job cannot take down the rest of the loop (this now has a dedicated regression test — see 2.2). A `null` metrics snapshot for a given job is logged as a `logger.warning(...)` on `CoordinatorService`'s own (non-suppressed) logger, so a real fetch failure reaches the main engine log instead of being silently buried inside the `additivity=false` metrics logger. This closes the gap I flagged previously where the warning could get swallowed by the suppressed logger.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Clean and consistent with the rest of `CoordinatorService`: Javadoc is present on the new field, the new private methods, the new `PeriodicJobMetricsLogger` class and its one public method, and the new `mergeWithHistoricalJobMetrics` helper. ASF license headers are present on both new files. No wildcard imports, no `System.out.println`. The inline comment above the budgeted-fetch call ("Use the configured interval as an overall budget for the whole fetch, not per worker.") accurately describes what the code does.

`CoordinatorServiceTest` gained two targeted tests: `testPeriodicJobMetricsLoggingCanBeDisabledWithZeroInterval` (asserts `EngineConfig.setPrintJobMetricsInfoInterval(0)` no longer throws) and `testPeriodicJobMetricsLoggingContinuesWhenSingleJobLoggingFails` (spies `CoordinatorService`, mocks `PeriodicJobMetricsLogger.logJobMetrics` via `MockedStatic` to throw for job 1, and asserts job 2 is still logged via `runPeriodicJobMetricsLoggingOnce()`). Both are deterministic, single-threaded, no `Thread.sleep`, no shared static state (the `MockedStatic` is properly scoped to a try-with-resources block), and the new test correctly shuts the `CoordinatorService` down in a `finally` block. The `@AfterEach` teardown was also extended to shut down `metricsLoggingScheduler` for every test in the class, not just the new ones — a good catch, since every other existing test that builds a `CoordinatorService` with the default `EngineConfig` (interval = 60) now also spins up this scheduler and would otherwise leak a daemon thread per test. `YamlSeaTunnelDomConfigProcessorTest` gained a real end-to-end YAML-parsing test for `print-job-metrics-info-interval: 0`. `PeriodicJobMetricsLoggerTest` (3 tests: non-null, null, empty) remains a reasonable smoke test for the leaf logging helper, with an honest class-level note that appender routing itself isn't exercised here (no `log4j-core` on this module's test classpath).

**Test stability rating: Stable.** Every test discussed above is deterministic and timing-independent; none of the flaky-test risk patterns (races, `Thread.sleep`, unclosed resources, static state, order dependence, floating-point-without-delta, weak readiness signals) apply.

One small nit (Issue 4 below): `PeriodicJobMetricsLoggerTest.testLogJobMetricsWithNullEmitsWarningAndDoesNotThrow`'s name and inline comment both say the null branch "logs a WARN," but `PeriodicJobMetricsLogger.logJobMetrics` actually just returns silently on `null` now (the warning was intentionally moved to `CoordinatorService`'s own logger in an earlier commit on this PR). The assertion itself (`assertDoesNotThrow`) is still correct either way, so this doesn't affect coverage or stability, but the comment is stale and could mislead a future reader of this file.

What remains not covered, unchanged from my last pass: no `CoordinatorService`-level test asserts that `metricsLoggingScheduler` is actually `null` (i.e., nothing gets scheduled) when the interval is `<= 0` — today only `EngineConfig`'s acceptance of `0` is tested, not that `CoordinatorService` itself correctly no-ops end-to-end. See Issue 2.

## 2.3 Documentation Updates

`docs/en` and `docs/zh` are both updated in the two relevant deployment guides, and now consistently describe the feature as writing to "the dedicated log file `seatunnel-metrics.log`" (the earlier, inaccurate "into the engine log" phrasing was corrected in a prior commit on this PR and remains corrected here). English and Chinese text match each other and match the actual default (`60`) and disable value (`0`). The only residual mismatch is cosmetic: the **PR title** itself ("...to engine log and seatunnel-metrics.log") still contains the original, now-inaccurate claim — see Issue 3.

# 3. Architectural Soundness

A precise fit for the stated goal: a purpose-built named logger with its own dedicated, independently-rotated file is the right primitive for a high-volume, low-priority operational log, and reusing the existing budgeted/parallel `getRunningJobRawMetrics` fetch path (rather than inventing a new one) is the right call. This commit's own contribution — making the scheduler field `final` and extracting the merge-with-history logic into a shared helper — is a small, purely positive cleanup with no behavioral change.

## 3.2 Maintainability

Good, and slightly better than before: the new `metricsLoggingScheduler` field is now `final`, matching the safe-publication pattern already used by its sibling constructor-time fields (`pipelineCleanupScheduler`, `metricsFetchExecutor`) rather than being the one exception in the class. The extracted `mergeWithHistoricalJobMetrics` helper reduces (though does not fully eliminate, see 1.1) duplicated merge logic in the class.

## 3.3 Extensibility

Unchanged from my prior assessment — reasonable. The budgeted-fetch pattern is now demonstrably reusable (this PR itself reuses it), and the new `mergeWithHistoricalJobMetrics` helper gives any future caller that needs "raw metrics + historical merge" a ready-made building block.

No impact on checkpoint/savepoint state, job submission protocol, or serialized formats. The widened `EngineConfig` setter range is a strict superset of previously-accepted values, so no existing valid configuration becomes invalid, and the config option itself pre-dates this PR with the same default, so there is no upgrade-path concern for existing deployments beyond the new log file appearing on disk.

# 4. Issue Summary

|---|-------|----------|----------|------|
| 1 | Periodic logger shares the small fixed-size `metricsFetchExecutor` pool (sized `max(1, min(8, availableProcessors))`) with `RealtimeMetricsService`'s independent 5-second polling loop. On small (1-2 vCPU) containers — the exact target of this PR's own Kubernetes `log4j2.properties` change — the pool can shrink to 1-2 threads, making cross-feature queuing delays proportionally larger under load. Self-healing (each side is bounded by its own timeout and simply catches up next cycle), not a correctness or stability bug. | `CoordinatorService.java:267-275` (pool sizing), `:1891` (shared `invokeAll`); `RealtimeMetricsService.java:79-80,132-133,260-261` (other consumer) | Medium | No |
| 2 | No `CoordinatorService`-level test asserts that `metricsLoggingScheduler` stays `null` (nothing scheduled) when `print-job-metrics-info-interval <= 0`; only `EngineConfig`/YAML-level acceptance of `0` is tested today. | `CoordinatorServiceTest.java` | Low | No |
| 3 | PR title ("...to engine log and seatunnel-metrics.log") still contains the original "engine log" claim that the code/docs/Javadoc were corrected away from in an earlier commit on this PR. Purely cosmetic, worth a one-line title edit for consistency. | PR title | Low | No |
| 4 | `PeriodicJobMetricsLoggerTest.testLogJobMetricsWithNullEmitsWarningAndDoesNotThrow`'s name/comment say the null branch "logs a WARN," but the method now just returns silently on `null` (the warning moved to `CoordinatorService`'s own logger). Assertion is still correct; only the name/comment is stale. | `PeriodicJobMetricsLoggerTest.java` | Low | No |
| 5 | The pre-existing no-arg `getRunningJobMetrics()` still inlines its own copy of the merge-with-history loop instead of using the `mergeWithHistoricalJobMetrics` helper this PR introduces. Not a regression (the duplication predates this PR and the method itself isn't touched here), just a natural, low-risk follow-up now that the helper exists. | `CoordinatorService.java:1777-1787` vs `:2494` | Low | No |

No other reviewers or maintainers have commented on this PR (`pulls/11775/reviews` and `/comments` show only my own two prior reviews; `/issues/11775/comments` is empty), so there is nothing from anyone else to reconcile here.

CI: as of this review the fork's "Build" workflow run for this commit is still queued/in-progress (most jobs `queued`, `updated-modules-*` jobs correctly `skipped` since this PR's footprint is outside those modules); no failures have been reported yet. Nothing here suggests this PR's diff would cause a CI failure.

# 5. Merge Recommendation

1. **Blockers:** None. Every High-severity issue from my original review (disable-via-zero crashing startup, an uncaught exception permanently killing the scheduler) was fixed in an earlier commit on this PR and is covered by dedicated, deterministic regression tests. The one Medium-severity issue from my previous pass (sequential per-worker timeout) was also fixed by switching to the budgeted/parallel fetch path. This commit additionally closes the last Low-severity nit from my prior review (the scheduler field is now `final`) and adds a small, correct dedup (`mergeWithHistoricalJobMetrics`).
   - Issue 1: consider a small dedicated executor for the periodic logger, or explicitly document the shared-capacity trade-off with `RealtimeMetricsService`, particularly given this PR's own Kubernetes deployment config targets CPU-constrained pods.
   - Issue 2: add the small amount of remaining `CoordinatorService`-level coverage for the disabled-scheduler branch.
   - Issue 3: update the PR title to drop the "to engine log" phrasing for consistency with the corrected docs/Javadoc.
   - Issue 4: fix the stale test name/comment about a warning being logged from the null branch.
   - Issue 5: optionally point the no-arg `getRunningJobMetrics()` at the new `mergeWithHistoricalJobMetrics` helper too.

This has been a genuinely well-executed hardening pass across the life of the PR: both original High-severity issues were fixed with real regression tests rather than silent patches, the Medium-severity performance issue was fixed by reusing an existing, better-suited primitive instead of a bespoke one, and this latest commit polishes off the smallest remaining nit while adding a clean, no-risk refactor on top. I'm happy to see this merged as-is; the five items above are all good follow-up material, none of them block.

PR #11776, opened 2026-08-12
[Docs][Connector-V2] Improve MongoDB connector docs

DanielCarter-stack:
## Overview

This PR polishes the MongoDB connector documentation (source and sink, both English and Chinese) to address a set of long-standing quality issues surfaced by recent reviewer feedback.

## What was wrong

- The English sink page had malformed connection-URI samples, including a doubled-dot typo (`127.0.0..1`) and `127.0.0.0` (network address) used in place of the loopback address, plus a "sharded cluster" example that was identical to the single-node example.
- Both source and sink pages omitted placeholder guidance (`${database_name}`, `${schema_name}`, `${table_name}`) on `database` / `collection`, even though the sink already supports multi-table write.
- The source page marked `[stream]` as unsupported although the connector runs in streaming mode through split-based parallelism.
- The sink `data_save_mode` row was missing the enum values and behavior summary in the options table.
- `match.query`, `match.projection`, `partition.split-key`, `partition.split-size`, and `flat.sync-string` descriptions were terse and did not explain when to use them or how they interact.
- The sample document under `flat.sync-string` was larger than necessary and included rows that no longer correspond to any option listed.
- The Chinese sink page also had a wrong "sharded cluster" example.
- Some sections were ordered inconsistently (Description moved after Key Features, Tips duplicated, etc.) and option descriptions used inconsistent capitalization between languages.

### Source page (`docs/en/connectors/source/MongoDB.md`)

- Rewrote the **Description** section to explain batch/streaming modes, split-based parallelism, and how `match.query` / `match.projection` / `flat.sync-string` narrow what is read.
- Refreshed **Source Options** with cleaner descriptions, an explicit note that `match.query` accepts `matchQuery` as a fallback, and added notes on choosing an indexed split key.
- Clarified the **Tips** section with three actionable bullets (compatibility, indexed split keys, schema behavior under `flat.sync-string`).
- Reorganized **Parameter Interpretation** into named subsections and replaced broken URI samples with correct ones (loopback IP, proper sharded cluster URI routed through `mongos`).
- Updated **MatchQuery Scan** examples to use Extended JSON-style queries and added explicit data-type examples (boolean, string, integer, date, floating point).
- Trimmed the **Flat Sync String** sample document to focus on the fields the connector actually produces under that mode.

### Sink page (`docs/en/connectors/sink/MongoDB.md`)

- Rewrote the **Description** section to distinguish append vs. upsert semantics and explain how exactly-once is achieved via upsert + checkpoint recovery.
- Restored full descriptions for `database` and `collection`, including `${database_name}`, `${schema_name}`, `${table_name}` placeholder guidance for multi-table write.
- Added the missing `data_save_mode` enum explanation (`DROP_DATA`, `APPEND_DATA`, `ERROR_WHEN_DATA_EXISTS`).
- Clarified **Tips** with the connector-level flush trigger explanation, the `upsert-key` fallback note, and the timer-flush/transaction exclusivity.
- Fixed URI typos (`127.0.0..1` → `127.0.0.1`, `127.0.0.0` → `127.0.0.1`) and the duplicate "sharded cluster" example.
- Reorganized **Parameter Interpretation** into named subsections.

### Chinese pages (`docs/zh/connectors/source/MongoDB.md`, `docs/zh/connectors/sink/MongoDB.md`)

- Mirrored all of the English changes in Chinese, keeping terminology consistent with the rest of the SeaTunnel Chinese docs (`批` / `流` / `精确一次` / `列投影` / `并行性`).
- Fixed the same URI typos and "sharded cluster" example on the sink page.
- Adjusted option table column alignment for readability.

## Scope

- 1 connector: **MongoDB** (source + sink, en + zh).
- 4 files changed.
- No code changes, no option renames, no default-value changes — documentation only.

## Testing

- `./mvnw spotless:apply` is not required for markdown-only changes.
- Verified rendered output by inspecting each file end-to-end; no broken links or table rows.


DanielLeens (commented):

This is a docs-only PR (4 files under `docs/en` and `docs/zh`, no `src/main` changes) that rewrites
the MongoDB source and sink connector pages: it fixes several genuinely broken URI samples
(`127.0.0..1`, `127.0.0.0` used as loopback, a "sharded cluster" example identical to the
single-node example), restores option descriptions that had gone missing (`data_save_mode` enum
values, `${database_name}`/`${schema_name}`/`${table_name}` placeholder guidance), reflows long
lines for readability, and rewrites the "Description" sections for both source and sink.

Several of the mechanical fixes are real improvements and I verified them against the connector
source, but the two rewritten "Description" sections both introduce new factual claims about the
connector's runtime behavior that are contradicted by the current `connector-mongodb` source code.
Since this doc regressed from accurate-but-terse to detailed-but-wrong in exactly those two spots,
I'm treating them as blocking.

# 1. Code Change Review

## 1.1 Core Logic Analysis

No production code was touched (`git diff origin/dev...HEAD --stat` shows only the four markdown
files). The review below is purely a source-vs-documentation cross-check.

**Verified correct (no issue):**

- `docs/en/connectors/sink/MongoDB.md`: `primary-key` "legacy fallback `upsert-key`" claim matches
  `MongodbSinkOptions.PRIMARY_KEY.withFallbackKeys("upsert-key")`
  (`seatunnel-connectors-v2/connector-mongodb/src/main/java/.../config/MongodbSinkOptions.java:67-73`).
- `data_save_mode` enum values/semantics (`DROP_DATA` truncates, `APPEND_DATA` is a no-op,
  `ERROR_WHEN_DATA_EXISTS` throws if data exists) match `MongodbSinkOptions.DATA_SAVE_MODE`
  (same file, lines 78-84) and `DefaultSaveModeHandler.handleDataSaveMode()`
  (`seatunnel-api/src/main/java/org/apache/seatunnel/api/sink/DefaultSaveModeHandler.java:99-113`).
- Sink Tips: "Enabling `transaction` is incompatible with the Zeta timer-flush feature... mixing
  them will silently disable the timer" matches `MongodbWriter`'s constructor, which only calls
  `context.registerFlushAction(this::timerFlush)` when `!transaction`
  (`seatunnel-connectors-v2/connector-mongodb/src/main/java/.../sink/MongodbWriter.java:80-82`) —
  so setting `sink.flush.interval` while `transaction = true` is indeed silently a no-op for this
  writer.
- Source doc: replacing the mongo-shell-only `ISODate(...)` match-query example with the Extended
  JSON `{"$date": "..."}` form is a real fix, not just cosmetic — `MongodbSource` parses
  `match.query` via `org.bson.BsonDocument.parse(s)`
  (`seatunnel-connectors-v2/connector-mongodb/src/main/java/.../source/MongodbSource.java:125-126`),
  and `BsonDocument.parse` accepts MongoDB Extended JSON, not mongo-shell helper syntax like
  `ISODate(...)`. The old example would not have parsed as written.
- Source doc: "Pick a split key backed by an index" and "flat.sync-string ignores the schema except
  for the single STRING field" both match `SamplingSplitStrategy` (sort/sample on `splitKey`) and
  `DocumentRowDataDeserializer.deserialize()` (only `fieldTypes[0]` is populated when
  `flatSyncString` is true) respectively.

**Issue 1 — `[stream]` marked supported and a fabricated "streaming mode" narrative added (source
docs, both en and zh):**

`docs/en/connectors/source/MongoDB.md` flips the Key Features checkbox from `- [ ] [stream]` to
`- [x] [stream]`, and the new Description section states: *"It supports both **batch** and
**streaming** jobs... In streaming mode the connector reads the assigned splits and tracks progress
through checkpoints so a restarted job resumes from the last committed cursor."* The same change is
made in `docs/zh/connectors/source/MongoDB.md` (`- [x] [流]`, plus "它同时支持**批**和**流**两种作业模式...").

This directly contradicts the connector source:

```java
// seatunnel-connectors-v2/connector-mongodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/mongodb/source/MongodbSource.java:66-68
@Override
public Boundedness getBoundedness() {
    return Boundedness.BOUNDED;
}
```

`Boundedness` is hardcoded to `BOUNDED` with no config-driven branch anywhere in `MongodbSource`,
`MongodbSplitEnumerator`, or `MongodbReader` — I checked all three. There is no cursor-resume /
tailing / change-stream logic in this module at all; `SamplingSplitStrategy` is a one-shot
sample-and-range-partition strategy for a bounded scan, structurally the same idea as a JDBC
partitioned batch read. The project's own definition of the `[stream]` feature flag
(`docs/en/introduction/concepts/connector-v2-features.md:40-42`) is: *"Streaming Job Mode, the data
read is unbounded and the job never stop."* By that definition this connector unambiguously does
not support `stream`, and the previous `- [ ] [stream]` was correct.

This is a regression from a correct claim to an incorrect one, and it's the kind of claim a user
would act on directly (e.g. attempting `job.mode = "STREAMING"` with this source, or designing a
pipeline assuming checkpoint-based cursor resumption that does not exist).

## 1.2 Compatibility Impact

Fully compatible from a code/config-contract standpoint — no code, option names, defaults, or APIs
were touched, this is a pure documentation change. However, Issue 1 and Issue 2 below create a
*semantic* compatibility risk: they document behavior the connector does not have, which can lead
users to build configs/expectations that don't match runtime behavior. That's a documentation-only
risk, not a runtime break, but it's exactly the kind of thing this section exists to catch since
"documentation is part of the feature" per the project's contribution guide.

N/A — no runtime code changed.

### Issue 1: `[stream]` support and "streaming mode" narrative are fabricated (source docs)

- **Location:** `docs/en/connectors/source/MongoDB.md:16` (checkbox) and `:23-24` (Description
  prose); mirrored in `docs/zh/connectors/source/MongoDB.md:16` and `:24-25`.
- **Problem:** Claims the connector supports unbounded streaming jobs with checkpoint-based cursor
  resumption. `MongodbSource.getBoundedness()` hardcodes `Boundedness.BOUNDED`
  (`seatunnel-connectors-v2/connector-mongodb/src/main/java/.../source/MongodbSource.java:66-68`),
  and there is no change-stream/tailing implementation anywhere in `connector-mongodb`.
- **Potential risk:** Users configure `job.mode = "STREAMING"` or rely on "resumes from the last
  committed cursor" semantics that don't exist, leading to confusion, failed jobs, or silently
  wrong expectations about freshness/latency.
- **Best improvement:** Revert the checkbox to `- [ ] [stream]` (`- [ ] [流]` in zh) and rewrite the
  Description to describe what the connector actually does: a bounded, parallel batch read that
  partitions the collection via `partition.split-key`/`partition.split-size` (optionally narrowed
  by `match.query`/`match.projection`/`flat.sync-string`). Drop the "streaming mode... resumes from
  the last committed cursor" sentence entirely — it does not correspond to any code path.
- **Severity:** High

### Issue 2: Sink Description claims primary-key becomes MongoDB's `_id` on upsert — contradicted by the serializer (both en and zh sink docs)

- **Location:** `docs/en/connectors/sink/MongoDB.md:30-36` (new Description bullet); mirrored in
  `docs/zh/connectors/sink/MongoDB.md:19-24`.
- **Problem:** The new Description text states: *"when `upsert-enable = true` and `primary-key` is
  configured, the connector uses the primary key as the MongoDB `_id` (or compound `_id`) and
  upserts."* The actual upsert path does the opposite — it strips `_id` and matches by the raw
  primary-key field values as a query filter, not as `_id`:

  ```java
  // seatunnel-connectors-v2/connector-mongodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/mongodb/serde/RowDataDocumentSerializer.java:74-81
  WriteModelSupplier upsertSupplier =
          row -> {
              final BsonDocument bsonDocument = rowDataToBsonConverter.convert(row);
              Bson filter = generateFilter(filterConditions.apply(bsonDocument));
              bsonDocument.remove("_id");
              BsonDocument update = new BsonDocument("$set", bsonDocument);
              return new UpdateOneModel<>(filter, update, new UpdateOptions().upsert(true));
          };
  ```

  `filterConditions` is `MongoKeyExtractor`, which builds the filter from the primary-key field
  names directly (e.g. `{name: ..., status: ...}`), each turned into a `Filters.eq(...)` condition
  (`RowDataDocumentSerializer.generateFilter`,
  `seatunnel-connectors-v2/connector-mongodb/src/main/java/.../serde/RowDataDocumentSerializer.java:113-120`;
  `seatunnel-connectors-v2/connector-mongodb/src/main/java/.../sink/MongoKeyExtractor.java:32-39`).
  `_id` is explicitly removed from the document before the `$set`, so on first insert MongoDB
  auto-generates its own `ObjectId` `_id` — the primary key is never written into `_id`.
- **Potential risk:** This is more than a wording nit — a user reading "primary key becomes `_id`"
  will reasonably expect (a) uniqueness/index guarantees tied to `_id`, (b) that they can query the
  document later by constructing an `_id` from the primary-key values, and (c) that no secondary
  index is needed for idempotent upserts to be efficient. All three assumptions are wrong under the
  actual filter-based-match implementation; e.g. without an index on the primary-key fields, every
  upsert does a collection scan to evaluate the filter.
- **Note on scope:** The pre-existing "Idempotent Writes" section further down (not new to this PR,
  only lightly reworded here) already carried a softer version of the same inaccuracy ("the
  connector combines the primary keys... as the MongoDB reserved primary key"). That one predates
  this PR, so I'm not counting it as a new introduction, but since this PR touched that exact
  paragraph, it would have been a good place to fix the misconception instead of restating it plus
  adding a second, more specific wrong claim ("_id (or compound `_id`)") earlier in the same file.
- **Best improvement:** Rewrite both the new Description bullet and the "Idempotent Writes"
  paragraph to describe the actual mechanism: the connector performs a filter-based
  `UpdateOneModel(filter=primary-key fields, update=$set, upsert=true)`, not an `_id`-substitution.
  Worth adding a one-line callout recommending an index on the primary-key fields for upsert
  performance, since there's no automatic `_id`-backed uniqueness helping here.
- **Severity:** High

# 2. Code Quality Assessment

## 2.1 Coding Standards

N/A (no Java changed). Markdown formatting is consistent with the rest of `docs/en`/`docs/zh`
(reflowed to ~100 cols, tables use consistent header casing). One very minor nit: in the "Flat Sync
String" sample JSON on the source page, one array element's value changed from
`"$numberInt":"-1028688944"` to `"$numberInt": "-1028686444"` — a digit got transposed during the
reformat. It's cosmetic (the sample data is illustrative/random), but worth a quick fix since
nothing in the diff explains why that specific number changed.

No test files changed — mandatory flaky-test analysis N/A.

## 2.3 Documentation Updates

This PR *is* the documentation update. Both `docs/en` and `docs/zh` were updated together, which is
correct practice; en/zh terminology is reasonably consistent (批/流/精确一次/列投影/并行性). The two
issues above need to be fixed in both language versions since both were changed to say the same
thing.

# 3. Architectural Soundness

Precise fix for the genuinely broken parts (URI typos, missing `data_save_mode` row, restored
placeholder guidance, dead `ISODate(...)` match-query example) — those are good, well-targeted
corrections. The two Description rewrites, however, went beyond "polish existing text" into
inventing new technical claims that weren't verified against the source, which is exactly the
failure mode to watch for in a docs-only PR: it's easy for prose additions to read as authoritative
without being checked against the code they describe.

## 3.2 Maintainability

Docs-only, no maintainability impact on code. Reflowed paragraphs and consistent table formatting
will make future diffs to this file cleaner (smaller line-level diffs going forward).

## 3.3 Extensibility

N/A.

No option renames, no default value changes, no protocol/serialization touched — verified via
`MongodbSourceOptions`/`MongodbSinkOptions`, neither of which is part of this diff. Existing running
jobs and saved configs are unaffected regardless of how this review resolves.

# 4. Issue Summary

| 1 | `[stream]` checkbox flipped to supported + fabricated "streaming mode/cursor resume" narrative; contradicted by `Boundedness.BOUNDED` hardcoded in `MongodbSource` | `docs/en/connectors/source/MongoDB.md:16,23-24`, `docs/zh/connectors/source/MongoDB.md:16,24-25` | High |
| 2 | New Description text claims upsert writes primary-key into MongoDB `_id`; actual code strips `_id` and matches by primary-key field values as a filter | `docs/en/connectors/sink/MongoDB.md:30-36`, `docs/zh/connectors/sink/MongoDB.md:19-24` | High |
| 3 | Sample JSON array value changed from `-1028688944` to `-1028686444` with no apparent reason (likely accidental digit transposition during reflow) | `docs/en/connectors/source/MongoDB.md` (Flat Sync String sample) | Low |

# 5. Merge Recommendation

1. Blockers — must be fixed (sorted by severity):
   - Issue 1: revert the `[stream]`/`[流]` checkbox to unsupported and rewrite the "streaming mode"
     sentences in both `docs/en` and `docs/zh` source pages to accurately describe the bounded,
     split-based parallel batch read.
   - Issue 2: rewrite the new sink Description bullet (and ideally the older "Idempotent Writes"
     paragraph it duplicates) in both `docs/en` and `docs/zh` to describe the real filter-based
     upsert mechanism instead of the `_id`-substitution claim.
   - Issue 3: fix the transposed digit in the sample JSON, or just leave the array element
     unchanged during reflow.

Overall assessment: the mechanical corrections in this PR (URI typos, restored option descriptions,
the `ISODate` → Extended JSON match-query fix, restored placeholder guidance) are real
improvements and clearly the result of careful comparison against the old page's rendering issues.
But the two rewritten Description sections introduce new, specific, and verifiably wrong claims
about connector capabilities (unbounded streaming, `_id`-based upsert) that go beyond what the old
page said and aren't supported by the current `connector-mongodb` source. Since this is a docs PR,
"looks well-written" isn't sufficient evidence of correctness — I'd ask for both Description
sections to be re-derived from `MongodbSource`/`MongodbSink`/`RowDataDocumentSerializer` rather than
from what "sounds right" for a MongoDB connector. Option A: keep the current before/after structure
of the PR but replace just the two offending paragraphs. Option B: drop the new "Description"
rewrite for source/sink entirely and keep only the option-table, URI, and `ISODate` fixes, which are
all independently correct and don't require touching the capability claims at all — that would be
the smaller, lower-risk path to merge if the author would rather not re-verify the prose against the
source.

zhangshenghang (approved):

PR #11783, opened 2026-08-12
[Feature][Zeta] Add dirty job tracking for member-loss recovery attempts

### LINKED ISSUE #11777: [STIP-34][Feature][Zeta] Add dirty job tracking for member-loss recovery attempts
### Search before asking

- [x] I searched open and closed SeaTunnel issues and STIPs for Zeta member removal, worker failover, retry counting, dirty jobs, recovery attempts, and failure history.
- [x] STIP-15 covers dirty **data** collection, while this proposal defines an operational health label for a Zeta job. STIP-33 covers task failure history, but does not define job-level member-loss recovery counting or dirty-job semantics.

## Summary

This STIP proposes a job-scoped **recovery-degraded** label, also referred to as a **dirty job**, for SeaTunnel Zeta.

A job becomes dirty when the number of job-level deployment attempts associated with unexpected or unclassified Zeta member loss reaches a configured threshold. The label is additive metadata: it does not replace `JobStatus`, change retry limits, or stop the job.

The proposal provides:

- a precise distinction between member loss, recovery episodes, and deployment attempts;
- job-level counting that does not multiply one recovery by the number of pipelines or task groups;
- a stable attempt identity that survives active-master changes;
- HA-backed running state and finished-job history;
- explicit `CLEAN`, `DIRTY`, and `UNKNOWN` evaluation results, with an absent `dirtyTask` object representing a disabled job;
- additive REST fields for running and finished jobs;
- an optional future path for trusted planned-leave exclusion.

The term `dirty job` in this STIP is unrelated to dirty-data collection. It describes a job whose runtime continuity has degraded because member-loss recovery has happened too many times.

## Motivation

In a Zeta cluster with two or more members, a worker process crash, machine failure, network isolation, Kubernetes Pod restart, or rolling replacement can remove a Hazelcast member. The current recovery path is approximately:

```text
Hazelcast memberRemoved
  -> SeaTunnelServer.memberRemoved(...)
  -> CoordinatorService.memberRemoved(...)
  -> failedTaskOnMemberRemoved(...)
  -> task groups on the lost address become FAILED
  -> SubPlan checks the restore limit
  -> prepareRestorePipeline()
  -> resources are requested and the pipeline is deployed again
```

The engine already restores pipelines and keeps an in-memory `pipelineRestoreNum`, but it does not provide a durable job-level answer to these questions:

1. How many job-level deployment attempts were caused by member loss?
2. Did one member-loss event affect multiple pipelines without being over-counted?
3. Did one failed deployment lead to another deployment attempt without being under-counted?
4. Did the tracking remain complete across an active-master change?
5. Can operators query the result after the job finishes?

This leaves repeatedly recovered streaming jobs looking healthy as long as their primary `JobStatus` returns to `RUNNING`.

## Current behavior and source boundaries

The current source establishes several important constraints:

- `SeaTunnelServer.memberRemoved(...)` forwards the Hazelcast membership event to the active coordinator.
- `CoordinatorService.failedTaskOnMemberRemoved(...)` finds task groups deployed on the lost address and marks eligible states as `FAILED`.
- `SubPlan.prepareRestorePipeline()` increments `pipelineRestoreNum` before resource release, resource request, and redeployment.
- `pipelineRestoreNum` is pipeline-scoped and in-memory. It is not a job-level member-loss counter and is not exposed through finished-job history.
- `MembershipServiceEvent` provides the lost member address and UUID, but not a Kubernetes termination reason, process exit code, or trusted planned-maintenance flag.
- The current Kubernetes resource manager does not classify member-removal causes.

Therefore, `memberRemoved` proves that a Hazelcast member left the cluster. It does **not** prove that a machine failed.

## Goals

1. Mark a job dirty after `X` job-level deployment attempts associated with member loss.
2. Count deployment attempts, not task-group failures, pipeline count, or raw member-removal count.
3. Preserve member UUID and address as diagnostic context without using address as member identity.
4. Avoid mixing ordinary connector, data, or user-code failures into this counter.
5. Preserve the tracking state across active-master changes and partition-owner changes.
6. Expose the same evaluation summary for running and finished jobs.
7. Keep the existing job state machine, retry behavior, checkpoint format, and REST fields backward compatible.
8. Fail open for job recovery but fail closed for observability: tracking failure must not block failover, and it must not be reported as `CLEAN`.

## Non-goals

The first implementation will not:

- add a new `JobStatus` such as `DIRTY_RUNNING`;
- stop, cancel, or suspend a dirty job automatically;
- replace Hazelcast membership failure detection with a new heartbeat detector;
- diagnose whether a member left because of OOM, machine failure, network partition, or Kubernetes policy;
- count ordinary business failures that have no member-loss cause;
- add Prometheus metrics or Web UI filtering in the first backend PR;
- implement trusted planned-leave exclusion in the first backend PR;
- change `job.retry.times` semantics.

### Member-loss incident

A `MemberLossIncident` records that one Hazelcast member UUID left and affected one job. Its stable identity is:

```text
jobId + lostMemberUuid
```

The incident is diagnostic and idempotent. It is not the dirty counter.

### Recovery episode

A `RecoveryEpisode` is one continuous member-loss recovery process for a job. Multiple member-loss incidents and multiple affected pipelines may belong to the same episode.

The episode begins when the coordinator registers the first affected member-loss incident. It completes when all affected pipelines have deployed successfully and recovered, or aborts when the job can no longer restore.

### Deployment attempt

A `RecoveryDeploymentAttempt` is one concrete redeployment attempt for one affected pipeline. Restore preparation allocates a stable identity:

```text
attemptId = episodeId + pipelineId + prepareSequence
```

`prepareSequence` is incremented atomically in job-scoped dirty state. The attempt is counted only when the prepared pipeline reaches the recovery `DEPLOYING` stage after resource allocation.

`prepareRestorePipeline()` alone does not count because no task group has been redeployed at that point.

### Job-level attempt generation

Within one recovery episode, every affected pipeline maintains its own confirmed prepare sequence. The episode's job-level generation is the maximum confirmed sequence:

- multiple pipelines entering their first deployment attempt produce generation `1`, not `N`;
- if any pipeline requires a second deployment attempt, generation becomes `2`;
- confirming the same stable attempt ID again is a no-op.

The job's `recoveryAttemptCount` increases only when a confirmed generation advances. It continues accumulating across completed episodes for the same job ID.

## Dirty evaluation

Proposed evaluation values:

| Value | Meaning |
| --- | --- |
| `dirtyTask` absent | The job did not enable dirty-job tracking. |
| `CLEAN` | Tracking is complete and the threshold has not been reached. |
| `DIRTY` | The threshold has been reached. This is irreversible for the current job ID. |
| `UNKNOWN` | Tracking is incomplete or inconsistent, so the job must not be reported as clean. |

The primary job status remains independent. For example, a job can be both `RUNNING` and `DIRTY`.

## Cause classification and Kubernetes semantics

The first implementation uses these cause semantics:

| Scenario | Classification | Count after deployment starts |
| --- | --- | --- |
| OOM, process crash, liveness restart | `UNCLASSIFIED` | Yes |
| Manual Pod deletion without prior trusted intent | `UNCLASSIFIED` | Yes |
| Rolling update without a trusted drain protocol | `UNCLASSIFIED` | Yes |
| Ordinary task failure without `memberRemoved` | No incident | No |

Unknown cause is intentionally counted because the engine cannot prove that the interruption was planned.

A later phase may add `PLANNED_VERIFIED`, but only through an authenticated and authorized control plane that registers the exact member UUID, request ID, actor, reason, creation time, and short expiration time before termination. A Pod `preStop` hook or JVM shutdown hook must not be trusted to classify its own exit.

If one recovery episode contains any `UNCLASSIFIED` incident, its deployment attempts are counted. The episode may be excluded only when every incident is `PLANNED_VERIFIED`.

## State model

The first implementation introduces a dedicated HA-backed job state keyed by job ID:

```text
engine_jobDirtyState: IMap<Long, DirtyJobState>
engine_jobDirtyMemberEventSequence: IMap<Long, Long>
engine_jobDirtyPendingMemberEvents: IMap<String, DirtyJobMemberEvent>
engine_jobDirtyEnabledThresholds: IMap<Long, Integer>
```

The enabled-threshold map acts as the rolling-upgrade gate: when every member can confirm that no job enabled tracking, member removal does not send the new journal or processor types. If that check is uncertain, tracking starts conservatively. The pending-event journal is keyed by the lost member UUID. It stores the global sequence,
member UUID, member address, and original observation time before coordinator readiness is
evaluated. Every master-capable member also keeps local fallback evidence before active-master
selection, so a sequence/journal outage followed by active-master loss still evaluates the
restored job as `UNKNOWN`. A fallback is removed only after that member observes a durable positive
sequence for the same UUID; it is never discarded merely because restore waits longer than a TTL.
Fallback detail is bounded, and overflow collapses into one permanent local tracking-gap marker
rather than dropping evidence. The enabled-threshold map contains only dirty-enabled running jobs; it
lets list APIs distinguish a disabled job from an enabled job whose main dirty state is temporarily
unavailable.

All four maps use the engine's configured synchronous backup count and are configured on every
member before Hazelcast starts. Per-job state mutation uses a single-key backup-aware
`EntryProcessor`; member-event journal insertion is idempotent by member UUID. If the compact
enabled marker cannot be persisted, the main dirty state is marked `UNKNOWN`, and the presence of
either marker or state activates conservative member-event tracking on master candidates.

`DirtyJobState` contains at least:

- job ID, immutable job creation time, and configured threshold;
- `dirty`, `evaluationStatus`, and `trackingComplete`;
- cumulative `recoveryAttemptCount`;
- recovery episode and member-loss incident counts;
- the active recovery episode;
- a bounded recent-episode history;
- first dirty time and last counted time;
- the last processed member-event sequence.

All updates for one job use a single-key `EntryProcessor` or an equivalent atomic operation. The processor must update synchronous backups. The map configuration must be installed consistently on every member before the map is first obtained.

The state is bounded. Incident and episode history has a configured maximum, task-group details are sampled, and stale incidents that never reach deployment are expired or marked aborted. If exact member-incident deduplication exceeds its bound, evaluation becomes permanently `UNKNOWN` rather than silently accepting a later replay as a new incident.

## Tracking completeness and active-master recovery

Job-scoped atomic updates do not form a distributed transaction with task and pipeline state. This proposal uses idempotent, at-least-once convergence instead of changing the existing job state machine into a distributed transaction.

A global monotonically increasing member-event sequence is stored using an existing serializable type in running engine state. For every `memberRemoved` event:

1. When the enabled-job gate is active or uncertain, the receiving member idempotently journals the removed member UUID and increments the global sequence before scanning jobs, including while a new active coordinator is still restoring.
2. Every dirty-enabled running job advances its own processed sequence after the event is handled.
3. Affected jobs register or merge the incident in the same atomic dirty-state update.
4. If dirty-state update fails, the job watermark remains behind the global sequence.
5. A missing dirty state or a lagging watermark evaluates to `UNKNOWN`, never `CLEAN`.

Dirty state is initialized before a dirty-enabled job becomes schedulable. This prevents a clean default from being inferred for a job that never created its state.

After an active-master change, reconciliation runs before restored pipelines are scheduled:

- the new master snapshots persisted worker-to-pipeline assignments before `JobMaster.init(...)`
  can rewrite execution state;
- pending member events are replayed against that frozen snapshot using their original event time;
- a persisted `PREPARED` attempt is compared with the pipeline state;
- `DEPLOYING` or `RUNNING` confirms the original stable attempt ID idempotently;
- pre-deployment states retain the binding and wait for normal deployment;
- terminal non-restorable state aborts the episode;
- missing state, watermark lag, incompatible serialization, or an unexplained transition evaluates to `UNKNOWN`.

Before parallel restore starts, the new master freezes assignment snapshots for all jobs. After
every distributed running job has been restored successfully, it performs one final replay and
acknowledges a journal entry only when every dirty-enabled job state has processed at least that
event sequence. An acknowledged event remains as a short-lived tombstone for the configured
incident TTL, allowing every master-capable member to reconcile and remove its unsequenced local
fallback only after observing durable evidence. Events arriving after the snapshot remain journaled until the same
acknowledgement condition is true, and are also applied to pending restored jobs by the normal
membership callback. A restore failure retains the journal for the next active-master recovery.
Duplicate replay is safe because both member incidents and deployment attempts have stable
identities. Acknowledgement always evaluates the highest journal sequence currently stored for one
member UUID. One entry processor atomically verifies that exact sequence and assigns TTL to the
same value; an older acknowledgement never overwrites or expires a newer event value.

Tracking-store failure must be logged and surfaced, but it must not block the existing task failure
or failover path. Member-event journal allocation, insertion, and restore reads each have a 100 ms
observation budget; journal acknowledgement runs outside the restore future on a bounded executor.
Per-job event updates are submitted asynchronously before affected tasks are failed and are
serialized locally by job ID. Operations that must return a stable attempt ID wait at most 100 ms;
a timeout returns control to the existing recovery path and permanently evaluates that job as
`UNKNOWN`. A dedicated watchdog detects incomplete-marker futures that never complete; failed or
stalled HA incomplete markers are retried by job ID in the background while the member remains
alive. Tracking calls at submission, pending scheduling, active-master restore, and pipeline
transitions are isolated so they cannot change the original recovery result.

Every delayed job-state processor is scoped by the immutable job creation time read from persisted
job metadata. Unknown owners are never treated as wildcards. This prevents an overdue callback,
watchdog, or active-master snapshot failure from changing a later execution that reuses the same
job ID.

The pending scheduler does not release restored jobs until the active-master restore future,
including final event replay, has completed. If a bounded replay update times out, scheduling still
continues but the local and HA state are marked incomplete, so the result cannot be `CLEAN`.

Terminal-state zombie recovery also retains the existing delayed cleanup contract. If the dirty-job
running state cannot be removed during zombie cleanup, the coordinator writes a pending
`JobCleanupRecord` and retries through the existing cleanup scheduler instead of dropping the
terminal cleanup boundary after history has already been persisted.

## Configuration

The dirty threshold is separate from the retry limit:

```hocon
env {
  job.retry.times = 5
  job.dirty-task.threshold = 2
}
```

Proposed options:

| Option | Default | Meaning |
| --- | ---: | --- |
| `job.dirty-task.threshold` | `-1` | Disabled when less than or equal to zero. |
| `engine.dirty-task.event-history-size` | `10` | Maximum recent episodes retained per job. |
| `engine.dirty-task.pending-incident-ttl` | `10m` | Maximum lifetime of an incident that never reaches deployment. |

`job.retry.times` remains the restore limit. `job.dirty-task.threshold` only controls the health label.

## REST and history contract

Existing REST fields remain unchanged. Running and finished job detail adds an optional object:

```json
{
  "dirtyTask": {
    "dirty": true,
    "evaluationStatus": "DIRTY",
    "trackingComplete": true,
    "threshold": 2,
    "recoveryAttemptCount": 2,
    "recoveryEpisodeCount": 1,
    "memberLossIncidentCount": 1,
    "activeEpisodeId": null,
    "lastLeaveClassification": "UNCLASSIFIED",
    "lastLostMemberUuid": "23cc7ea6-6374-4f2b-8a33-df75b5e58f52",
    "lastLostAddress": "10.0.0.12:5801",
    "lastCountedTime": 1786500000000
  }
}
```

Job lists should expose only `dirty`, `evaluationStatus`, and `recoveryAttemptCount`. Recent episode detail belongs only in job detail to avoid unbounded list responses.

When a job finishes, `JobHistoryService` snapshots dirty state together with the finished-job lifecycle. Running dirty state is removed only after the final history write. Older finished-job records without this field deserialize as tracking disabled or absent, without changing existing fields.

## Compatibility and rollout

This proposal is additive and disabled by default:

- existing jobs require no configuration change;
- `JobStatus`, retry limits, checkpoint/savepoint payloads, and existing REST fields are unchanged;
- no connector SPI changes are required;
- old finished-job records remain readable;
- no new state object is created for disabled jobs.

Because Hazelcast partition owners deserialize map values and entry processors, the first release requires a homogeneous engine cluster:

1. Upgrade every member to binaries that contain the new serializers and tracker while keeping the option disabled.
2. Confirm no old member remains.
3. Enable the option only for newly submitted jobs.

Before downgrading to a binary that does not understand the new types, stop creating enabled jobs and allow existing enabled jobs to finish and clean their running dirty state.

## Performance considerations

Existing member-removal handling already scans running jobs. This proposal keeps the same big-O scan. When the cluster confirms that no job enabled tracking, it performs no tracking write; the enabled-marker observation is bounded to 100 ms and an uncertain result tracks conservatively. With enabled jobs, one member removal adds one global sequence/journal update so every master-capable member can reconcile local fallback evidence, plus one lightweight watermark update per enabled job and one bounded state update per affected job. Disabled jobs do not create or update job-scoped dirty state.

The implementation must measure member-removal processing latency with at least 100 and 1,000 concurrent jobs. The state retained per job is bounded by the configured history size and sampled task-group details.

### Phase 1: backend dirty-job tracking

1. Add configuration and bounded dirty-state models.
2. Add backup-aware atomic tracker operations, global member-event sequencing, an idempotent
   pending-event journal, and an enabled-job marker.
3. Register member-loss incidents before existing task groups are failed.
4. Allocate stable attempt IDs during restore preparation.
5. Confirm attempts only when recovery reaches `DEPLOYING`.
6. Complete or retain attempt context according to deployment outcome.
7. Reconcile active state before scheduling after an active-master change.
8. Persist finished history and expose additive REST fields.
9. Add unit and integration coverage for idempotency, failure paths, HA recovery, and compatibility.

### Phase 2: trusted planned-leave exclusion

This phase is independent and requires a reviewed authenticated operations control plane. It is not part of the first backend PR.

### Later optional work

- Prometheus metrics and alert transitions;
- Web UI badge, filtering, and episode detail;
- faster stale-worker suspicion as a separate failure-detection proposal.

## Acceptance criteria

- [ ] One member-loss episode affecting multiple pipelines counts one first-generation job attempt.
- [ ] A second deployment attempt advances the job count exactly once.
- [ ] Duplicate `memberRemoved` delivery and duplicate attempt confirmation are idempotent.
- [ ] Ordinary task failures without an active member-loss recovery context are not counted.
- [ ] Restore preparation without reaching `DEPLOYING` is not counted.
- [ ] A completed episode cannot be consumed by a later ordinary business failure.
- [ ] A Kubernetes Pod restart without trusted planned-leave intent is classified as `UNCLASSIFIED` and counted after deployment starts.
- [ ] Active-master change before or during attempt confirmation neither loses nor duplicates the count.
- [ ] A `memberRemoved` event received during active-master recovery is replayed from the journal
      against the pre-initialization assignment snapshot before the restored job is scheduled.
- [ ] A failed active-master restore retains the pending event journal for a later retry.
- [ ] Partition-owner change preserves state through a synchronous backup.
- [ ] Delayed tracking updates from an earlier execution cannot affect a later execution reusing
      the same job ID.
- [ ] Terminal zombie cleanup retries dirty-state removal through pending job cleanup if the first
      removal attempt fails.
- [ ] Missing state, watermark lag, or tracker failure returns `UNKNOWN`, not `CLEAN`.
- [ ] Reaching the threshold does not alter the primary job status or retry behavior.
- [ ] Running and finished jobs expose the same summary semantics.
- [ ] The feature is disabled by default and old finished-job records remain readable.
- [ ] Mixed-version clusters cannot enable the feature before every member has upgraded.
- [ ] Member-removal processing remains bounded under the defined concurrency test matrix.

## Usage Scenario

Operators running long-lived Zeta streaming jobs need to identify jobs that repeatedly recover after member loss even when those jobs return to `RUNNING`.

The proposed contract answers:

- Has this job exceeded the acceptable number of member-loss deployment attempts?
- How many job-level attempts and recovery episodes occurred?
- Which member UUID and address were involved?
- Is the result complete, or should it be treated as unknown?
- Can the same result be queried after the job finishes?

## Related issues

- #11735: task exception and failure history contract
- #11755: stale task completion across restore generations
- #10804: state-store abstraction for Hazelcast IMap
- #11472: finished metrics after worker failover
- #10461: retry mechanism correctness

## Are you willing to submit a PR?

- [x] Yes, I am willing to submit a Draft PR for Phase 1.

## Code of Conduct

- [x] I agree to follow the Apache Software Foundation Code of Conduct.

danielnadean:

Closes #11777.

This PR adds optional dirty-job tracking for SeaTunnel Zeta member-loss recovery attempts. A job can now expose `CLEAN`, `DIRTY`, or `UNKNOWN` evaluation metadata without changing the existing job status or retry behavior.

## What is included

- add the job option `job.dirty-task.threshold` and engine-side dirty-job tracker configuration;
- persist HA-backed dirty-job running state, member-event sequencing, and finished-job history snapshots;
- count only member-loss-driven recovery deployment generations after deployment actually starts;
- replay and reconcile pending member-loss events across active-master changes;
- expose additive dirty-job summary fields through running-job and finished-job REST responses;
- fail closed to `UNKNOWN` when tracking state is incomplete, including delayed-owner and zombie-cleanup retry paths;
- add unit and integration coverage for dirty-job state evaluation, config parsing, map config, null-safety, and coordinator recovery paths;
- document the new option in English and Chinese job configuration docs.

## Verification

- locally ran `./mvnw spotless:apply -pl seatunnel-api,seatunnel-engine/seatunnel-engine-common,seatunnel-engine/seatunnel-engine-server -nsu`
- did not run local compile/tests/E2E/package steps because this repository is under the current SeaTunnel GitHub-CI-only validation rule for functional verification

## Does this pull request potentially affect one of the following parts

- [ ] No. None of these.
- [ ] Dependency management
- [ ] Something that affects deployment
- [x] The core runtime module
- [x] The API
- [x] Default configuration values
- [ ] Introduces a breaking change
- [x] Documentations

## Documentation

- [x] This change is documented.

### DIFF
--- a/config/seatunnel.yaml
+++ b/config/seatunnel.yaml
@@ -20,6 +20,9 @@ seatunnel:
     classloader-cache-mode: true
     history-job-expire-minutes: 1440
     backup-count: 1
+    dirty-task:
+      event-history-size: 10
+      pending-incident-ttl: 10m
     queue-type: blockingqueue
     print-execution-info-interval: 60
     print-job-metrics-info-interval: 60
--- a/docs/en/introduction/configuration/JobEnvConfig.md
+++ b/docs/en/introduction/configuration/JobEnvConfig.md
@@ -51,6 +51,14 @@ This counter accumulates for the life of the pipeline; an intermediate successfu
 
 Used to control the default retry interval when a job fails. The default value is 3 seconds, and it only works in the Zeta engine.
 
+### job.dirty-task.threshold
+
+Controls the optional Zeta dirty-job label. The default value is `-1`; values less than or equal to zero disable tracking. A positive value marks the job `DIRTY` after that many job-level deployment generations associated with Hazelcast member loss. This option does not change `job.retry.times`, stop the job, or replace the primary job status.
+
+One member-loss recovery that redeploys multiple pipelines counts as one first-generation attempt. A later redeployment advances the generation once. Ordinary connector, data, or user-code failures without an active member-loss recovery episode are not counted. Kubernetes Pod restart, manual deletion, and rolling replacement are observed as unclassified member loss unless a future trusted planned-leave protocol proves otherwise, so their redeployment attempts are counted.
+
+Running and finished job APIs expose an optional `dirtyTask` object when tracking is enabled. Its `evaluationStatus` is `CLEAN`, `DIRTY`, or `UNKNOWN`. `UNKNOWN` means the engine cannot prove tracking completeness after a state-store or failover problem and must not be interpreted as clean.
+
 ### savemode.execute.location
 
 This parameter is used to specify the location of the savemode when the job is executed in the Zeta engine.
--- a/docs/zh/introduction/configuration/JobEnvConfig.md
+++ b/docs/zh/introduction/configuration/JobEnvConfig.md
@@ -51,6 +51,14 @@
 
 用于控制作业失败时的默认重试间隔。默认值为3秒，并且仅适用于Zeta引擎。
 
+### job.dirty-task.threshold
+
+用于启用 Zeta 脏任务标记。默认值为 `-1`，小于等于 0 时关闭。配置为正数后，由 Hazelcast 成员离线引起的作业级部署代数达到该阈值时，作业会被标记为 `DIRTY`。该选项不会修改 `job.retry.times`、停止作业或替代原有作业状态。
+
+一次成员离线恢复即使重新部署多个 Pipeline，首次部署也只计为一个作业级尝试；后续再次部署时，代数只增加一次。没有处于成员离线恢复上下文中的 Connector、数据或用户代码异常不会计入。Kubernetes Pod 重启、手动删除和滚动替换在缺少可信计划下线协议时只能识别为未分类成员离线，因此对应的重新部署尝试会被计入。
+
+启用后，运行中和已完成作业接口会返回可选的 `dirtyTask` 对象，其中 `evaluationStatus` 为 `CLEAN`、`DIRTY` 或 `UNKNOWN`。`UNKNOWN` 表示状态存储或故障转移期间无法证明观测完整，不能把该作业判断为干净任务。
+
 ### savemode.execute.location
 
 此参数用于指定在Zeta引擎中执行作业时SaveMode执行的时机。
--- a/seatunnel-api/src/main/java/org/apache/seatunnel/api/options/EnvCommonOptions.java
+++ b/seatunnel-api/src/main/java/org/apache/seatunnel/api/options/EnvCommonOptions.java
@@ -58,6 +58,14 @@ public class EnvCommonOptions {
                     .defaultValue(3)
                     .withDescription("The retry interval seconds of this job");
 
+    public static Option<Integer> DIRTY_JOB_RESTORE_THRESHOLD =
+            Options.key("job.dirty-task.threshold")
+                    .intType()
+                    .defaultValue(-1)
+                    .withDescription(
+                            "The member-loss recovery deployment threshold used to mark a job as dirty. "
+                                    + "A value less than or equal to zero disables dirty-job tracking.");
+
     public static Option<Long> CHECKPOINT_INTERVAL =
             Options.key("checkpoint.interval")
                     .longType()
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/Constant.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/Constant.java
@@ -59,6 +59,17 @@ public class Constant {
 
     public static final String IMAP_PENDING_JOB_CLEANUP = "engine_pendingJobCleanup";
 
+    public static final String IMAP_DIRTY_JOB_STATE = "engine_jobDirtyState";
+
+    public static final String IMAP_DIRTY_JOB_MEMBER_EVENT_SEQUENCE =
+            "engine_jobDirtyMemberEventSequence";
+
+    public static final String IMAP_DIRTY_JOB_PENDING_MEMBER_EVENTS =
+            "engine_jobDirtyPendingMemberEvents";
+
+    public static final String IMAP_DIRTY_JOB_ENABLED_THRESHOLDS =
+            "engine_jobDirtyEnabledThresholds";
+
     public static final String IMAP_CONNECTOR_JAR_REF_COUNTERS = "engine_connectorJarRefCounters";
 
     /**
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/EngineConfig.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/EngineConfig.java
@@ -32,6 +32,7 @@ import org.apache.seatunnel.engine.common.runtime.ExecutionMode;
 
 import lombok.Data;
 
+import java.time.Duration;
 import java.util.Collections;
 import java.util.Map;
 
@@ -90,6 +91,16 @@ public class EngineConfig {
 
     private long stateCleanupDelayMillis =
             ServerConfigOptions.MasterServerConfigOptions.STATE_CLEANUP_DELAY_MILLIS.defaultValue();
+
+    // Maximum closed recovery episodes retained in each dirty-job state.
+    private int dirtyJobEventHistorySize =
+            ServerConfigOptions.MasterServerConfigOptions.DIRTY_JOB_EVENT_HISTORY_SIZE
+                    .defaultValue();
+
+    // Lifetime shared by incomplete episodes and acknowledged journal tombstones.
+    private Duration dirtyJobPendingIncidentTtl =
+            ServerConfigOptions.MasterServerConfigOptions.DIRTY_JOB_PENDING_INCIDENT_TTL
+                    .defaultValue();
 
     private ClusterRole clusterRole = ClusterRole.MASTER_AND_WORKER;
 
@@ -198,6 +209,24 @@ public class EngineConfig {
                             + " must be >= 0");
         }
         this.stateCleanupDelayMillis = stateCleanupDelayMillis;
+    }
+
+    public void setDirtyJobEventHistorySize(int dirtyJobEventHistorySize) {
+        checkPositive(
+                dirtyJobEventHistorySize,
+                ServerConfigOptions.MasterServerConfigOptions.DIRTY_JOB_EVENT_HISTORY_SIZE
+                        + " must be > 0");
+        this.dirtyJobEventHistorySize = dirtyJobEventHistorySize;
+    }
+
+    public void setDirtyJobPendingIncidentTtl(Duration dirtyJobPendingIncidentTtl) {
+        checkNotNull(dirtyJobPendingIncidentTtl);
+        if (dirtyJobPendingIncidentTtl.isZero() || dirtyJobPendingIncidentTtl.isNegative()) {
+            throw new IllegalArgumentException(
+                    ServerConfigOptions.MasterServerConfigOptions.DIRTY_JOB_PENDING_INCIDENT_TTL
+                            + " must be > 0");
+        }
+        this.dirtyJobPendingIncidentTtl = dirtyJobPendingIncidentTtl;
     }
 
     public EngineConfig setQueueType(QueueType queueType) {
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/YamlSeaTunnelDomConfigProcessor.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/YamlSeaTunnelDomConfigProcessor.java
@@ -48,6 +48,7 @@ import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Locale;
@@ -235,6 +236,9 @@ public class YamlSeaTunnelDomConfigProcessor extends AbstractDomConfigProcessor 
                     .key()
                     .equals(name)) {
                 engineConfig.setStateCleanupDelayMillis(Long.parseLong(getTextContent(node)));
+            } else if (ServerConfigOptions.MasterServerConfigOptions.DIRTY_JOB_CONFIG.equals(
+                    name)) {
+                parseDirtyJobConfig(node, engineConfig);
             } else if (ServerConfigOptions.MasterServerConfigOptions.CONNECTOR_JAR_STORAGE_CONFIG
                     .key()
                     .equals(name)) {
@@ -343,6 +347,29 @@ public class YamlSeaTunnelDomConfigProcessor extends AbstractDomConfigProcessor 
             // If dynamic slot is enabled, the schedule strategy must be REJECT
             LOGGER.info("Dynamic slot is enabled, the schedule strategy is set to REJECT");
             engineConfig.setScheduleStrategy(ScheduleStrategy.REJECT);
+        }
+    }
+
+    private void parseDirtyJobConfig(Node dirtyJobNode, EngineConfig engineConfig) {
+        for (Node node : childElements(dirtyJobNode)) {
+            String name = cleanNodeName(node);
+            if (ServerConfigOptions.MasterServerConfigOptions.DIRTY_JOB_EVENT_HISTORY_SIZE
+                    .key()
+                    .equals(name)) {
+                engineConfig.setDirtyJobEventHistorySize(
+                        getIntegerValue(name, getTextContent(node)));
+            } else if (ServerConfigOptions.MasterServerConfigOptions.DIRTY_JOB_PENDING_INCIDENT_TTL
+                    .key()
+                    .equals(name)) {
+                engineConfig.setDirtyJobPendingIncidentTtl(
+                        org.apache.seatunnel.api.configuration.ReadonlyConfig.fromMap(
+                                        Collections.singletonMap(name, getTextContent(node)))
+                                .get(
+                                        ServerConfigOptions.MasterServerConfigOptions
+                                                .DIRTY_JOB_PENDING_INCIDENT_TTL));
+            } else {
+                LOGGER.warning("Unrecognized element: " + name);
+            }
         }
     }
 
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/config/server/ServerConfigOptions.java
@@ -23,6 +23,7 @@ import org.apache.seatunnel.api.configuration.Option;
 import org.apache.seatunnel.api.configuration.Options;
 import org.apache.seatunnel.api.metadata.MetadataConfig;
 
+import java.time.Duration;
 import java.util.Map;
 
 /** Declares the server-side configuration keys that are exposed through `seatunnel.yaml`. */
@@ -143,6 +144,8 @@ public class ServerConfigOptions {
     /** The options for master. */
     public static class MasterServerConfigOptions {
 
+        public static final String DIRTY_JOB_CONFIG = "dirty-task";
+
         public static final Option<Integer> PRINT_EXECUTION_INFO_INTERVAL =
                 Options.key("print-execution-info-interval")
                         .intType()
@@ -188,6 +191,20 @@ public class ServerConfigOptions {
                         .withDescription(
                                 "How long to retain terminal job/pipeline/task state in distributed maps before removing it. "
                                         + "This delay allows late asynchronous callbacks to observe a terminal tombstone instead of a missing state entry.");
+
+        public static final Option<Integer> DIRTY_JOB_EVENT_HISTORY_SIZE =
+                Options.key("event-history-size")
+                        .intType()
+                        .defaultValue(10)
+                        .withDescription(
+                                "The maximum number of completed member-loss recovery episodes retained per job.");
+
+        public static final Option<Duration> DIRTY_JOB_PENDING_INCIDENT_TTL =
+                Options.key("pending-incident-ttl")
+                        .durationType()
+                        .defaultValue(Duration.ofMinutes(10))
+                        .withDescription(
+                                "The maximum lifetime of an incomplete member-loss recovery episode.");
         // The options about Hazelcast IMAP store end
         /////////////////////////////////////////////////
 
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/job/DirtyJobEvaluationStatus.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/job/DirtyJobEvaluationStatus.java
@@ -17,34 +17,13 @@
 
 package org.apache.seatunnel.engine.common.job;
 
-import lombok.AllArgsConstructor;
-import lombok.Data;
-import lombok.NoArgsConstructor;
-
-import java.io.Serializable;
-
-@AllArgsConstructor
-@Data
-@NoArgsConstructor
-public final class JobStatusData implements Serializable {
-    private static final long serialVersionUID = 1L;
-
-    private Long jobId;
-    private String jobName;
-    private JobStatus jobStatus;
-    private long submitTime;
-    private Long startTime;
-    private Long finishTime;
-    // Optional additive projection; null preserves disabled-job semantics.
-    private DirtyJobSummary dirtyTask;
-
-    public JobStatusData(
-            Long jobId,
-            String jobName,
-            JobStatus jobStatus,
-            long submitTime,
-            Long startTime,
-            Long finishTime) {
-        this(jobId, jobName, jobStatus, submitTime, startTime, finishTime, null);
-    }
+/**
+ * Evaluation result for member-loss recovery tracking.
+ *
+ * <p>UNKNOWN is intentionally distinct from CLEAN when tracking evidence is incomplete.
+ */
+public enum DirtyJobEvaluationStatus {
+    CLEAN,
+    DIRTY,
+    UNKNOWN
 }
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/job/DirtyJobMemberEvent.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/job/DirtyJobMemberEvent.java
@@ -23,28 +23,32 @@ import lombok.NoArgsConstructor;
 
 import java.io.Serializable;
 
+/**
+ * HA journal entry for a member removal that may outlive the active coordinator which observed it.
+ */
 @AllArgsConstructor
 @Data
 @NoArgsConstructor
-public final class JobStatusData implements Serializable {
+public class DirtyJobMemberEvent implements Serializable {
     private static final long serialVersionUID = 1L;
 
-    private Long jobId;
-    private String jobName;
-    private JobStatus jobStatus;
-    private long submitTime;
-    private Long startTime;
-    private Long finishTime;
-    // Optional additive projection; null preserves disabled-job semantics.
-    private DirtyJobSummary dirtyTask;
+    // Global cluster watermark, or -1 when only local fallback evidence exists.
+    private long sequence;
+    // Stable Hazelcast identity used for idempotent journal updates.
+    private String memberUuid;
+    // Diagnostic host used to match persisted task assignments.
+    private String memberHost;
+    // Diagnostic port used with the host for exact assignment matching.
+    private int memberPort;
+    // Original observation time retained across active-master replay.
+    private long eventTime;
 
-    public JobStatusData(
-            Long jobId,
-            String jobName,
-            JobStatus jobStatus,
-            long submitTime,
-            Long startTime,
-            Long finishTime) {
-        this(jobId, jobName, jobStatus, submitTime, startTime, finishTime, null);
+    /**
+     * Returns the diagnostic address without using it as the member identity.
+     *
+     * @return host and port used only for affected-task matching and diagnostics
+     */
+    public String getAddressText() {
+        return memberHost + ":" + memberPort;
     }
 }
--- a/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/job/DirtyJobState.java
+++ b/seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/job/DirtyJobState.java
@@ -1,0 +1,610 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.engine.common.job;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * HA state used to evaluate member-loss recovery degradation for one job.
+ *
+ * <p>The state separates member incidents, job recovery episodes, and per-pipeline deployment
+ * attempts. Only confirmed deployment generations advance the job counter. Incident UUIDs, attempt
+ * IDs, and the cluster member-event watermark make replay idempotent and prevent incomplete
+ * tracking from being reported as clean.
+ */
+@Data
+@NoArgsConstructor
+public class DirtyJobState implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    // Stable engine job identity and distributed-map key.
+    private long jobId;
+    // Submission creation time preventing delayed updates from crossing job-ID reuse boundaries.
+    private long jobCreateTime;
+    // Positive job-scoped attempt threshold captured at submission.
+    private int threshold;
+    // Irreversible flag once the cumulative attempt threshold is reached.
+    private boolean dirty;
+    // Query result which is independent from the primary JobStatus.
+    private DirtyJobEvaluationStatus evaluationStatus;
+    // False whenever the engine can no longer prove the observation is complete.
+    private boolean trackingComplete;
+    // Cumulative job-level attempt generations across all episodes.
+    private int recoveryAttemptCount;
+    // Number of member-loss recovery episodes created for this job.
+    private int recoveryEpisodeCount;
+    // Number of distinct member UUID incidents accepted for this job.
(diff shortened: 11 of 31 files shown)


DanielLeens (commented):
Hi @danielnadean, thanks for opening this — I know it's still marked as a **draft**, so I'm not treating anything below as a blocking gate right now. I went ahead and did the full deep-dive anyway (this touches Zeta's member-loss recovery path, so I wanted to trace it end-to-end while it's fresh) so you have the findings early, before you invest more time on top of the current design. Please read the "Merge Recommendation" section with that framing — it's guidance for when you flip this to ready-for-review, not a verdict on the draft as it stands.

Overall: this is a genuinely careful piece of engineering. The fail-closed discipline, the exception isolation at every integration point, and the ownership/ epoch guards on the distributed state are all more rigorous than most first submissions I review. My comments below are mostly about hardening a few edges, not about the core design being wrong.

- **User pain point**: today, when Hazelcast member loss forces Zeta to redeploy pipelines, a job can look "successfully recovered" in the job status API even though it silently redeployed N times — there's no signal that a job has been through repeated member-loss recoveries, which is often a leading indicator of a flaky/degraded cluster member or a job that should be re-submitted rather than trusted long-term.
- **Fix approach**: add an *additive*, opt-in (`job.dirty-task.threshold`, default `-1` = disabled) tracking layer (`DirtyJobService` + `DirtyJobState`) that observes member-loss-driven pipeline redeployments, counts job-level "generations" of member-loss recovery, and exposes a `CLEAN` / `DIRTY` / `UNKNOWN` verdict via the running-job and finished-job REST APIs — without touching `job.retry.times`, job status, or scheduling decisions.
- **One-sentence summary**: a purely observational, fail-closed, HA-backed dirty-job label bolted onto the existing member-loss recovery path.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Files/methods touched** (the ones that matter for correctness):
- `seatunnel-engine-server/.../master/dirty/DirtyJobService.java` (new, ~1608 lines) — the coordination layer: 4 Hazelcast `IMap`s (`dirtyJobStateMap`, `memberEventSequenceMap`, `pendingMemberEventMap`, `enabledThresholdMap`), atomic `EntryProcessor`s, two daemon `ScheduledExecutorService`s (`journalExecutor` size 2, `trackingRetryExecutor` size 1), and a 100ms operation timeout (`TRACKING_OPERATION_TIMEOUT_MILLIS`).
- `seatunnel-engine-common/.../job/DirtyJobState.java` (new, ~610 lines) — the pure state machine: episode/attempt bookkeeping, threshold comparison, TTL and event-history bounding, `CLEAN`/`DIRTY`/`UNKNOWN` evaluation.
- `CoordinatorService.java` — wires `DirtyJobService` into: job submit/restart (`init...` / `dirtyJobService.initialize(...)` at L1190/L1474), the `MembershipListener` callback path (`failedTaskOnMemberRemoved` → `failTasksOnMemberRemoved`, L2168–2279), master-switch replay (`restoreJobFromMasterActiveSwitch` → `replayPendingMemberEvents`, L2417–2452), and lifecycle (`shutdown()` at L2148 calls `dirtyJobService.shutdown()`).
- `JobMaster.java` — thin, try/catch-wrapped facade methods (`prepareDirtyJobRecoveryAttempt`, `confirmDirtyJobRecoveryAttempt`, `completeDirtyJobRecoveryPipeline`, `reconcileDirtyJobRecoveryPipeline`, L1362–1452) that `SubPlan.java` calls from its existing pipeline state-transition methods.
- `SubPlan.java` — calls the above `JobMaster` facade methods inline from the pre-existing `updatePipelineState`/deploy/restore flow (e.g. the `PipelineStatus.DEPLOYING` transition at L392–397, the RUNNING-completion path at L718–725, and `reconcile` on restore at L578–597).

**Runtime path — Scenario A (normal job submit, feature disabled or enabled but no member loss)**:
```
Client submit -> CoordinatorService.submitJob(...) [L1420-1478]
  -> dirtyJobService.initialize(jobImmutableInformation, false) [L1474]
     -> DirtyJobService.getThreshold(...) reads job.dirty-task.threshold from JobConfig envOptions
        -> threshold <= 0 (default -1) => initialize() returns immediately, no IMap writes, no state created
  -> normal scheduling continues unchanged (dirtyJobInitialized flag only gates removeRunningState()
     cleanup on submit failure, L1506-1507 — it never gates or branches the actual job-scheduling outcome)
```
I confirmed this by reading `failTasksOnMemberRemoved` (CoordinatorService.java L2175-2279): the *actual* task-failure decision (`makeTasksFailed`, operating on `subPlan.getCoordinatorVertexList()`/`getPhysicalVertexList()` filtered by `isTaskAffected`) is computed **unconditionally**, independent of `trackingEnabled`/`dirtyJobService` success or failure. `collectAffectedPipelineIds`/`processMemberEvent` are only used to *feed* the dirty tracker, never to decide which vertices get marked `FAILED`. That's real evidence for the PR's "without changing the existing job status or retry behavior" claim, not just a description I'm taking on faith.

**Runtime path — Scenario B (member loss, feature enabled)**:
```
Hazelcast MembershipListener -> SeaTunnelServer.memberRemoved(event)
 -> CoordinatorService.failedTaskOnMemberRemoved(event) [L2168]
    -> createMemberEventForTracking(event) [L2362] -> dirtyJobService.recordMemberEvent(...) [DirtyJobService.java L3103]
       (journals to pendingMemberEventMap via JournalMemberEventProcessor; on failure/timeout,
        falls back to localPendingMemberEvents + returns a sequence=-1 event, never null)
    -> failTasksOnMemberRemoved(event, dirtyJobMemberEvent) [L2175]
       for each tracked JobMaster (running + pending):
         trackingEnabled = dirtyJobMemberEvent != null && dirtyJobService.isEnabled(jobMaster)  // try/catch guarded
         if trackingEnabled: affectedPipelineIds = collectAffectedPipelineIds(...)
                              dirtyJobService.processMemberEvent(jobMaster, seq, uuid, addr, affectedPipelineIds, eventTime)
                                -> DirtyJobService.java L3266 -> if seq<0: markTrackingIncomplete (fail-closed to UNKNOWN)
                                                                else: submitStateUpdate(MemberEventProcessor) [L3280, L1385-1451]
         makeTasksFailed(...)  // unconditional, real recovery decision — untouched by the above
 -> SubPlan (existing restart path) eventually calls:
      JobMaster.prepareDirtyJobRecoveryAttempt(pipelineId) [JobMaster.java L1362, try/catch-wrapped]
        -> DirtyJobService.prepareRecoveryAttempt -> PrepareAttemptProcessor [L1454-1477], guarded by isOwnedBy(state, expectedJobCreateTime)
      ... on PipelineStatus.DEPLOYING: JobMaster.confirmDirtyJobRecoveryAttempt(...) [SubPlan.java L395-397]
      ... on all tasks RUNNING: JobMaster.completeDirtyJobRecoveryPipeline(...) [SubPlan.java L718-725]
```
Every single `dirtyJobService`/`JobMaster` dirty-tracking call site I found on this path is wrapped in its own `try/catch (Throwable)` that logs and continues (`CoordinatorService.java` L2201-2261, L2421-2450; `JobMaster.java` L1362-1452) — I could not find a call site where a `DirtyJobService` exception would propagate up and interrupt the real recovery/restart flow.

**Active-master failover ordering** (this was my biggest worry going in): `restoreJobFromMasterActiveSwitch` calls `dirtyJobService.getPendingMemberEvents()` (L1196) and `dirtyJobService.initialize(..., true)` (L1190) **before** `replayPendingMemberEvents` synchronously calls `dirtyJobService.awaitMemberEventReplay(...)` (L2446) which blocks (bounded, 100ms) on each replayed event before the job is handed back to normal scheduling. So a job cannot resume scheduling ahead of its own pending member-event replay — I didn't find the race I was looking for here.

**Key Findings**:
1. Ownership/epoch safety is handled with real rigor: every mutating `EntryProcessor` (`PrepareAttemptProcessor`, `ConfirmAttemptProcessor`, `CompletePipelineProcessor`, `ReconcilePipelineProcessor`, `MarkIncompleteProcessor` — `DirtyJobService.java` L1454-1607) checks `isOwnedBy(state, expectedJobCreateTime)` before mutating, so a stale/late write from a superseded job execution (e.g. after a client-side timeout) cannot silently corrupt a different execution's counters — it's a no-op instead.
2. Per-job update ordering is explicitly serialized client-side via the `pendingStateUpdates` map (`submitStateUpdate`, L1072-1108) — chained with `.handle(...).thenCompose(...)` — in addition to whatever ordering Hazelcast's per-partition-key EntryProcessor execution already gives you. `pendingStateUpdates` is a plain `HashMap`, but every read/write of it is inside `synchronized (pendingStateUpdates)` (L1078, L1090, L1100), so it's safe despite not being a `ConcurrentHashMap`.
3. The 4 new IMaps (`IMAP_DIRTY_JOB_STATE`, `IMAP_DIRTY_JOB_MEMBER_EVENT_SEQUENCE`, `IMAP_DIRTY_JOB_PENDING_MEMBER_EVENTS`, `IMAP_DIRTY_JOB_ENABLED_THRESHOLDS`) are explicitly configured with synchronous backups (`backupCount=2, asyncBackupCount=0` per `SeaTunnelServerStarter.configureDirtyJobMaps`, verified by `DirtyJobMapConfigTest`) — so tracking state survives a partition-owner failover, which matters a lot for an HA feature like this.
4. `DirtyJobService.shutdown()` (L727-730, `journalExecutor.shutdownNow()` + `trackingRetryExecutor.shutdownNow()`) **is** called from `CoordinatorService.shutdown()` (L2148) — I specifically went looking for a thread-pool leak here since the two `ScheduledExecutorService`s are instance fields, and it's handled.
5. Threshold semantics match the docs exactly: `recoveryAttemptCount >= threshold` (`DirtyJobState.java` L360) flips to `DIRTY`, and both `docs/en` and `docs/zh` describe it as "marks the job DIRTY after that many job-level deployment generations" — consistent, no off-by-one between code and docs.

## 1.2 Compatibility Impact
**Fully compatible.**
- `job.dirty-task.threshold` defaults to `-1` (`EnvCommonOptions.java` L61-67) — every existing job config is untouched and the feature is a no-op end-to-end (`getThreshold(...) <= 0` short-circuits `initialize`, `isEnabled`, `createUnknownState`, etc.) unless explicitly opted in.
- REST additions are strictly additive: `BaseService.addDirtyJobState` (L722-742) only adds the `dirtyTask` JSON sub-object when `state != null`; every pre-existing field/key in the running-job and finished-job JSON responses is untouched (confirmed by diffing `base` vs `head` of `BaseService.java` — the change is a pure insertion, not a restructuring of the existing `JsonObject` chain).
- `JobStatusData.JobState` gets one new nullable field (`dirtyTask`, `JobStatusData.java` L36-49) but keeps the original 6-arg constructor delegating to `this(..., null)` — source- and (default Java-serialization-wise) binary-compatible for a rolling upgrade or mixed-version read of persisted finished-job history.
- `DirtyJobState`/`DirtyJobSummary`/`DirtyJobMemberEvent` are brand-new `Serializable` classes with no prior on-disk/IMap representation, so there's no old-format data to be backward compatible *with* — compatibility risk here is moot rather than solved, which is fine.

- Two new daemon thread pools per `DirtyJobService` instance (`journalExecutor` size 2, `trackingRetryExecutor` size 1) — small, bounded, and cleaned up on shutdown (see 1.1 Finding 4).
- `TRACKING_OPERATION_TIMEOUT_MILLIS = 100L` is used pervasively (I count ~15 call sites) as the client-side wait budget for `IMap.getAsync/putAsync/submitToKey(...).get(100ms)`. On a healthy cluster this is fine; under a real GC pause or network blip on the partition owner, this budget will trip fairly easily, which by design just flips the job to `UNKNOWN` rather than corrupting anything — but it does mean the feature can end up reporting `UNKNOWN` far more often than `CLEAN`/`DIRTY` under exactly the conditions (cluster stress, member churn) where the tracker is most interesting. See Issue 1.
- No unbounded growth I could find: `eventHistorySize` bounds completed episodes per job (`DirtyJobState`), `pendingIncidentTtlMillis` expires incomplete episodes, `boundLocalMemberEvents()` (L863-882) caps the local fallback map at `eventHistorySize * 4`, and short-lived tombstones use Hazelcast per-entry TTL (`ExtendedMapEntry.setValue(current, tombstoneTtlMillis, ...)`, L1324-1325).
- Synchronous IMap backups (`backupCount=2`) add write-path latency/network cost proportional to job count with tracking enabled — acceptable given the feature is opt-in, but worth knowing this isn't free once someone turns it on cluster-wide.

### Issue 1
- **Location**: `seatunnel-engine-server/.../master/dirty/DirtyJobService.java:2883` (constant) and its ~15 use sites (e.g. L2988, L3044, L3066, L3118, L3142, L3223, L3300, L3403…)
- **Problem description**: `TRACKING_OPERATION_TIMEOUT_MILLIS = 100L` is a hardcoded constant, not exposed as a configurable `Option` (unlike `event-history-size`/`pending-incident-ttl`, which are in `ServerConfigOptions`).
- **Potential risk**: under real cluster conditions (GC pause on the partition owner, transient network jitter), 100ms is tight for a distributed `EntryProcessor`/`getAsync` round trip. Because every timeout fails closed to `UNKNOWN` by design, this isn't a correctness bug — but it can make the feature report `UNKNOWN` disproportionately often precisely when the cluster is under the kind of stress that makes dirty-job tracking most valuable, undermining the feature's practical signal.
- **Best improvement**: expose this as a `ServerConfigOptions`/`EngineConfig` option (same pattern already used for `event-history-size`/`pending-incident-ttl`) so operators can tune it for their cluster's real GC/network profile, or at minimum document the 100ms assumption and its failure mode in the class Javadoc.
- **Severity**: Low

### Issue 2
- **Location**: `seatunnel-engine-server/.../master/dirty/DirtyJobService.java:1034-1042` (`scheduleLocalMemberEventReconciliation`) and `:823-861` (`scheduleMemberEventAcknowledgement`), vs. `:711-724` (`clearLocalCoordinatorState`)
- **Problem description**: `clearLocalCoordinatorState()` (called on master demotion) bumps `coordinatorEpoch` and cancels futures tracked in `pendingStateUpdates`, but it does **not** cancel already-scheduled tasks sitting in `journalExecutor` for member-event acknowledgement/reconciliation. A straggler task scheduled just before demotion can still fire and write to `pendingMemberEventMap` after a new master has taken over.
- **Potential risk**: I traced this carefully and believe it's benign in practice — every processor involved (`JournalMemberEventProcessor`, `AcknowledgeJournalEventProcessor`) is sequence-guarded/idempotent, so a stale write either loses to a higher-sequence value or gets rejected by the `expectedSequence` check. But this "why it's safe" reasoning isn't captured anywhere (no comment, no test), so a future refactor of those processors could silently reintroduce a real race here.
- **Best improvement**: either cancel/track those scheduled tasks the same way `pendingStateUpdates` is handled, or add an explicit code comment + a regression test documenting that idempotent processors are what makes this safe, so it's a deliberate invariant rather than an accident.
- **Severity**: Low

# 2. Code Quality Assessment

## 2.1 Coding Standards
Comment completeness is genuinely strong here — every field in `DirtyJobService` has a one-line "why" comment (not a restated name), every public method has Javadoc describing preconditions/return semantics, and the class-level Javadoc on `DirtyJobService` and `DirtyJobState` explains the fail-closed contract up front. I didn't find a new core method or non-trivial field missing a comment. No issue to raise here.

- `DirtyJobStateTest.java` (263 lines, pure unit tests on the state machine) is solid: it covers job-level generation counting across multiple pipelines, replay idempotency across a simulated master switch, TTL-based episode expiry (`shouldReportUnknownWhenActiveEpisodeExpires`), event-history overflow forcing `UNKNOWN` (`shouldBecomeUnknownWhenIncidentDeduplicationHistoryOverflows`), watermark-lag `UNKNOWN` (`shouldReportUnknownWhenMemberEventWatermarkLags`), and re-affected-pipeline handling. Threshold-at-boundary is implicitly covered (`shouldCountJobLevelRecoveryGenerationsAcrossPipelines` hits count==threshold and asserts `DIRTY`).
- `DirtyJobServiceTest.java` (102 lines) is much thinner than the 1608-line class it targets — it only exercises 3 static/pure helpers (`isOwnedBy`, `registerLocalOwner`/`registerLocalIncompleteOwner`, `buildFailClosedPendingMemberEvents`). The async `submitStateUpdate`/`awaitStateUpdate` timeout-and-fail-closed path, the `acknowledgeMemberEvent`/`scheduleMemberEventAcknowledgement` retry loop, and the `coordinatorEpoch` invalidation window (Issue 2 above) have no dedicated unit test.
- That gap is partially offset by two new real-cluster integration tests in `CoordinatorServiceTest.java`: `testMemberEventJournalSurvivesMasterSwitch` (2-node Hazelcast cluster, kills the master, asserts the journal IMap survives and is visible to the new master) and `testTerminalZombieCleanupShouldRetainPendingRetryWhenDirtyStateRemovalFails` (Mockito-backed, asserts a retry record is kept when `removeRunningState` fails). Both use `Awaitility.await().atMost(...).untilAsserted(...)` — no hard `Thread.sleep` in the new dirty-job tests. (The single `Thread.sleep(1000)` at `CoordinatorServiceTest.java:1480` is pre-existing and unrelated to this PR.)
- I diffed `CoordinatorServiceTest.java` base vs. head line-by-line: every change is additive (new imports, 2 new test methods) except one reflection call site updated from `getDeclaredMethod("restoreJobFromMasterActiveSwitch", Long.class, JobInfo.class)` to `(..., Long.class, JobInfo.class, Map.class)` — that's a mechanical follow of the corresponding production method signature change, not a weakened assertion. No existing test's assertions were loosened, removed, or had their strictness reduced.
- **Stability rating: Risk present.** Not "High risk" — the fail-closed design means untested async paths degrade to `UNKNOWN` rather than corrupting state or destabilizing the actual recovery flow, and the two integration tests do exercise the most HA-sensitive scenario (journal survives master switch) against a real cluster. But the concurrency-heaviest new file in the PR (timeouts, retries, epoch invalidation) is under-covered at the unit level, which is worth closing before this leaves draft.

## 2.3 Documentation Updates
`docs/en/introduction/configuration/JobEnvConfig.md` and `docs/zh/introduction/configuration/JobEnvConfig.md` are both updated under the existing `### job.retry.times`/`job.retry.interval.seconds` section, in the same style, and are consistent with each other (I compared them side-by-side) and with the code: default `-1`, semantics of "job-level deployment generations," the `dirtyTask`/`evaluationStatus` REST contract, and the explicit non-goal ("does not change `job.retry.times`, stop the job, or replace the primary job status") all match what I found in `EnvCommonOptions.java` and `BaseService.addDirtyJobState`. No doc/code mismatch found.

# 3. Architectural Soundness

**Long-term solution.** This isn't a quick patch — it's a proper HA-aware sub-system (durable journal, sequence watermarks, per-execution ownership guards, synchronous-backup IMaps) built to survive master failover, which is the right bar for anything touching the member-loss recovery path in a "financial-grade" cluster.

## 3.2 Maintainability
The heavy use of small, single-purpose `EntryProcessor` inner classes (one per state transition) keeps each mutation easy to reason about in isolation, and the consistent `isOwnedBy` guard pattern makes it easy to audit for the "did we just correctly ignore this stale write" property. The main maintainability risk is `DirtyJobService.java`'s sheer size (1608 lines in one class) — future contributors extending this will want to lean on the existing processor/guard patterns rather than inventing new ones.

## 3.3 Extensibility
`MemberLeaveClassification` is currently always `UNCLASSIFIED` (per the docs: "Kubernetes Pod restart, manual deletion, and rolling replacement are observed as unclassified member loss unless a future trusted planned-leave protocol proves otherwise") — the enum and the `DirtyJobState.processMemberEvent` signature are clearly designed to be extended with a real classification later without a breaking change. That's a nice bit of forward design.

Nothing here can break an existing 2.6-era or older cluster's upgrade path: the feature is entirely additive, defaults to disabled, and its IMaps/classes have no prior on-disk representation to be incompatible with (see 1.2).

# 4. Issue Summary

|---|---|---|---|
| 2 | `DirtyJobService`'s async/timeout/retry/epoch machinery is not directly unit-tested (only 2 helper-level unit tests + 2 integration tests) | `seatunnel-engine-server/.../master/dirty/DirtyJobServiceTest.java` (whole file) | Medium |
| 1 | Fixed 100ms `TRACKING_OPERATION_TIMEOUT_MILLIS` is not configurable and may cause frequent `UNKNOWN` fallback under real cluster stress | `DirtyJobService.java:2883` + ~15 use sites | Low |
| 2b | Scheduled acknowledgement/reconciliation tasks aren't cancelled on master demotion (currently safe only because processors are idempotent; not documented/tested as an invariant) | `DirtyJobService.java:1034-1042`, `:823-861` vs `:711-724` | Low |

# 5. Merge Recommendation

### Conclusion: N/A — this PR is a draft; the items below are guidance for when you mark it ready for review, not a blocking verdict today.

1. **Recommended fixes before requesting review** (non-blocking for the draft, but I'd want these closed before a real merge review):
   - Add unit-level coverage for `DirtyJobService`'s own concurrency machinery (Issue 2) — specifically the timeout→`UNKNOWN` path and the acknowledgement retry loop. This is the highest-value gap given how much of the PR's safety argument rests on "timeouts fail closed correctly."
   - Consider making the 100ms operation timeout configurable, or explicitly document the assumption (Issue 1).
   - Add a short comment (and ideally a test) explaining why un-cancelled straggler tasks after demotion are safe (Issue 2b), so it reads as an intentional invariant rather than an oversight for the next person touching this file.
2. **Everything else** (fail-closed correctness, exception isolation at every integration point, ownership/epoch guards, additive REST/config/docs, synchronous-backup IMap configuration) held up under a full trace of both the normal-submit and member-loss-recovery runtime paths — I did not find a correctness bug in the core design.

No better alternative approach jumps out — an HA-aware journal + sequence-watermark design is the right shape for "must survive master failover," and I wouldn't recommend a lighter-weight alternative given the stated goal.

Nice work on this one — genuinely one of the more carefully defended distributed-state PRs I've reviewed. Happy to take another pass once you flip it out of draft.

PR #11778, opened 2026-08-12
[Feature][Connector-V2][Prometheus] Migrate PrometheusWriter to engine-level FlushSignal

surafel58:
### Purpose of this pull request

Closes #11773.

`PrometheusWriter` used to create and manage its own `ScheduledExecutorService` to
periodically flush buffered metrics based on the connector-level `flush_interval`
option. This meant the connector owned an extra background thread whose lifecycle
could run concurrently with the normal Sink write / checkpoint / close paths.

Zeta already provides engine-level timer flush through `FlushSignal` and
`SinkWriter.Context#registerFlushAction`. This PR migrates the Prometheus Sink to
that mechanism:

- `PrometheusSink#createWriter()` now passes `SinkWriter.Context` to `PrometheusWriter`.
- The writer opts in with `context.registerFlushAction(this::flush)`.
- The connector-owned `ScheduledExecutorService`, `ScheduledFuture`, and
  `flushSchedule()` (plus their startup/shutdown) are removed.
- The `batch_size` trigger and the final flush in `close()` are preserved.
- A failed flush is now propagated to the engine instead of being logged and the
  batch silently dropped, so a flush that did not succeed is no longer treated as a
  successful flush.
- The connector-level `flush_interval` option is deprecated (kept for backward
  compatibility, but ignored). Users should set the engine-level `sink.flush.interval`
  in the job `env` block instead. If `flush_interval` is still configured, the writer
  logs a one-time warning pointing to the new option.

The engine invokes the registered action on the normal Sink input-processing path,
so there is no connector-owned scheduler thread and no concurrency with the Sink
lifecycle. As documented, engine-level timer flush is currently supported only by Zeta.

### Does this PR introduce _any_ user-facing change?

Yes.

- `flush_interval` (connector option) is deprecated and no longer starts a background
  flush thread. Existing jobs that set it keep validating (the option is still
  accepted) but the value is ignored; a warning is logged. To keep timer-based
  flushing, set `sink.flush.interval` in the job `env` block (Zeta only).
- The English and Chinese Prometheus Sink docs are updated with a new "Timer Flush"
  section and the deprecation note, and the streaming example now uses
  `sink.flush.interval`.

### How was this patch tested?

Added `PrometheusWriterTest` (unit test), which verifies:

1. the writer registers a flush action via `SinkWriter.Context#registerFlushAction`,
   and the buffered records are sent only when that action is invoked (the flush
   signal), not on every write;
[...]

PR #11797, opened 2026-08-13
[Bug][Connector-V2][Prometheus] Apply batch_size default in loadConfig

surafel58:

Follow-up to the review discussion on #11778 (raised by @nzw921rx and @DanielLeens).

`PrometheusSinkConfig.loadConfig()` set `batch_size` only when the key was present, guarded by
`getOptional(BATCH_SIZE).isPresent()`. `ReadonlyConfig.getOptional()` does not fall back to the
option's declared default (only `get()` does, via `getOptional(option).orElseGet(option::defaultValue)`).
So a job that omits `batch_size` ended up with `batchSize = 0`, and the size-based flush trigger in
the writer (`batchSize > 0 && batchList.size() >= batchSize`) never fired. The declared default of
`1024` in `PrometheusSinkOptions.BATCH_SIZE` was effectively ignored.

This PR uses `get()` so the declared default is applied when `batch_size` is not configured.

Yes, a bug fix. A `Prometheus` sink that does not set `batch_size` now buffers up to the documented
default of `1024` rows before a size-based flush, instead of never flushing by size (`batchSize = 0`).
Jobs that explicitly set `batch_size` are unaffected.

Added `PrometheusSinkConfigTest` with three cases:

- `batch_size` not configured -> `loadConfig` returns the declared default (`1024`);
- `batch_size` configured -> that value is used;
- `batch_size` non-positive -> rejected by the existing `checkArgument` guard.

```
./mvnw -pl seatunnel-connectors-v2/connector-prometheus test
=> Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
```

### Check list

* [x] Code changed are covered with tests.
* [ ] If any new Jar binary package adding in your PR, please add License Notice — N/A.

### DIFF
--- a/seatunnel-connectors-v2/connector-prometheus/src/main/java/org/apache/seatunnel/connectors/seatunnel/prometheus/config/PrometheusSinkConfig.java
+++ b/seatunnel-connectors-v2/connector-prometheus/src/main/java/org/apache/seatunnel/connectors/seatunnel/prometheus/config/PrometheusSinkConfig.java
@@ -51,10 +51,11 @@ public class PrometheusSinkConfig extends HttpConfig {
         if (pluginConfig.getOptional(PrometheusSinkOptions.KEY_TIMESTAMP).isPresent()) {
             sinkConfig.setKeyTimestamp(pluginConfig.get(PrometheusSinkOptions.KEY_TIMESTAMP));
         }
-        if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
-            int batchSize = checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
-            sinkConfig.setBatchSize(batchSize);
-        }
+        // Use get() (not getOptional().isPresent()) so the option's declared default is applied
+        // when batch_size is not configured; getOptional() does not fall back to the default, which
+        // would leave batchSize at 0 and disable the size-based flush trigger.
+        int batchSize = checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
+        sinkConfig.setBatchSize(batchSize);
         if (pluginConfig.getOptional(PrometheusSinkOptions.FLUSH_INTERVAL).isPresent()) {
             long flushInterval = pluginConfig.get(PrometheusSinkOptions.FLUSH_INTERVAL);
             sinkConfig.setFlushInterval(flushInterval);
--- a/seatunnel-connectors-v2/connector-prometheus/src/test/java/org/apache/seatunnel/connectors/seatunnel/prometheus/PrometheusSinkConfigTest.java
+++ b/seatunnel-connectors-v2/connector-prometheus/src/test/java/org/apache/seatunnel/connectors/seatunnel/prometheus/PrometheusSinkConfigTest.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.seatunnel.connectors.seatunnel.prometheus;
+
+import org.apache.seatunnel.api.configuration.ReadonlyConfig;
+import org.apache.seatunnel.connectors.seatunnel.prometheus.config.PrometheusSinkConfig;
+import org.apache.seatunnel.connectors.seatunnel.prometheus.config.PrometheusSinkOptions;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+class PrometheusSinkConfigTest {
+
+    @Test
+    void shouldApplyBatchSizeDefaultWhenNotConfigured() {
+        Map<String, Object> map = new HashMap<>();
+        map.put("url", "http://localhost:9090/api/v1/write");
+        map.put("key_label", "c_map");
+        map.put("key_value", "c_double");
+
+        PrometheusSinkConfig config = PrometheusSinkConfig.loadConfig(ReadonlyConfig.fromMap(map));
+
+        Assertions.assertEquals(
+                PrometheusSinkOptions.BATCH_SIZE.defaultValue().intValue(), config.getBatchSize());
+    }
+
+    @Test
+    void shouldUseConfiguredBatchSize() {
+        Map<String, Object> map = new HashMap<>();
+        map.put("url", "http://localhost:9090/api/v1/write");
+        map.put("key_label", "c_map");
+        map.put("key_value", "c_double");
+        map.put("batch_size", 5);
+
+        PrometheusSinkConfig config = PrometheusSinkConfig.loadConfig(ReadonlyConfig.fromMap(map));
+
+        Assertions.assertEquals(5, config.getBatchSize());
+    }
+
+    @Test
+    void shouldRejectNonPositiveBatchSize() {
+        Map<String, Object> map = new HashMap<>();
+        map.put("url", "http://localhost:9090/api/v1/write");
+        map.put("key_label", "c_map");
+        map.put("key_value", "c_double");
+        map.put("batch_size", 0);
+
+        Assertions.assertThrows(
+                Exception.class,
+                () -> PrometheusSinkConfig.loadConfig(ReadonlyConfig.fromMap(map)));
+    }
+}


DanielLeens (commented):

`PrometheusSinkConfig.loadConfig()` only called `sinkConfig.setBatchSize(...)` when `pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()` was true. `ReadonlyConfig.getOptional()` reads only the raw user-supplied config map and does not fall back to `Option#defaultValue()` — only `ReadonlyConfig.get()` does that, via `getOptional(option).orElseGet(option::defaultValue)`. So a job that omits `batch_size` left `PrometheusSinkConfig.batchSize` at its Java primitive default of `0`, silently ignoring the documented default of `1024` (`PrometheusSinkOptions.BATCH_SIZE`, `defaultValue(DEFAULT_BATCH_SIZE)`). This PR replaces the guarded assignment with an unconditional `sinkConfig.setBatchSize(checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE)))`, so the declared default is actually applied when the key is omitted. One-sentence summary: a real, narrowly-scoped bug fix that makes the code match the already-documented `batch_size` default.

Before/after example: a job configures the Prometheus sink with no `batch_size` key at all.
- Before: `PrometheusSinkConfig.getBatchSize()` returns `0`. In `PrometheusWriter.write(Point)` the size-based flush guard `if (batchSize > 0 && batchList.size() >= batchSize)` (`PrometheusWriter.java:110` on dev) never evaluates true, so size-based flushing is silently disabled for the entire job.
- After: `getBatchSize()` returns `1024` as documented, and the writer flushes every 1024 buffered rows as a user reading `docs/en/connectors/sink/Prometheus.md:50` (`batch_size | Int | false | 1024`) would expect.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Diff (`PrometheusSinkConfig.java`):
```java
// before (apache/dev)
if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent()) {
    int batchSize = checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
    sinkConfig.setBatchSize(batchSize);
}
// after (this PR)
int batchSize = checkIntArgument(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));
sinkConfig.setBatchSize(batchSize);
```

Runtime flow confirming this is a normal-path fix, not a boundary/recovery-path one: `PrometheusWriter`'s constructor calls `PrometheusSinkConfig.loadConfig(pluginConfig)` unconditionally on every task start (`PrometheusWriter.java:69` on dev), and `this.batchSize = sinkConfig.getBatchSize()` (`PrometheusWriter.java:70`) is read once at construction and used for the rest of the writer's lifetime in `write(Point)`'s size-flush check (`PrometheusWriter.java:110`). Any Prometheus-sink job that omits `batch_size` — very likely the common case, since most examples in the docs (`docs/en/connectors/sink/Prometheus.md:94`) do set it explicitly, but plenty of real configs won't — hits this path every single time. So this is a real, everyday-path bug, not an edge case.

`checkIntArgument` is preserved (it wraps `checkArgument(args > 0)`), and it's now always invoked (previously only when the key was explicitly present). Since the declared default (`1024`) satisfies `> 0`, this does not introduce a new failure mode for the "omitted" case — I checked `PrometheusSinkOptions.BATCH_SIZE.defaultValue()` (`DEFAULT_BATCH_SIZE = 1024`) to confirm. An explicit `batch_size = 0` still throws the same unchecked `IllegalArgumentException` it did before (same behavior, just always reached now instead of only when present — which is exactly the fix).

## 1.2 Compatibility Impact — and duplicate/overlapping fix with #11738 (Issue 1)

**Issue 1 (High) — this fix duplicates/overlaps with #11738 ("[Improve][Connector-V2][Prometheus] Migrate batch size validation to OptionRule"), which now already contains the same fix.**

I checked both diffs against `apache/dev` and against each other, per your request:

- `#11738` originally (as of my two prior approvals on 2026-08-10 and 2026-08-11) migrated `batch_size` validation from the imperative `checkIntArgument`/`checkArgument` helper into a declarative `Conditions.greaterThan(PrometheusSinkOptions.BATCH_SIZE, 0)` constraint on `PrometheusSinkFactory.optionRule()`, enforced by `ConfigValidator` at job-submission time — a strict improvement (fail-fast, visible to tooling that introspects `optionRule()`), which I reviewed in depth at the time.
- On 2026-08-13 17:06:33 -0700 (2026-08-14T00:06:33Z) — **about 5.5 hours after this PR (#11797) was opened at 2026-08-13T16:42:19Z** — a new commit `6717dd04ca` titled "[Fix][Prometheus] Apply the default batch size" was pushed to `#11738`. I inspected it directly (`git show 6717dd04ca`): it removes the exact same `if (pluginConfig.getOptional(PrometheusSinkOptions.BATCH_SIZE).isPresent())` guard and replaces it with the same unconditional `sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));` — functionally identical to what this PR does (modulo `#11738` no longer needing `checkIntArgument`, since by then validation had already been moved to the `OptionRule` level in that PR).
- I confirmed this directly against the current head of `#11738` (`6717dd04ca73f9a4293d9f15f06476e7f7ab9776`): `PrometheusSinkConfig.loadConfig()` there is now `sinkConfig.setBatchSize(pluginConfig.get(PrometheusSinkOptions.BATCH_SIZE));`, unconditional, no guard — the default-application bug this PR (#11797) fixes is already fixed there too.

Concretely: as things stand right now, **`#11738`'s current diff is a strict superset of this PR's fix** — same root-cause correction, plus (a) fail-fast validation at job-submission time via `Conditions.greaterThan(BATCH_SIZE, 0)` in `PrometheusSinkFactory.optionRule()` instead of this PR's task-start-time `checkIntArgument`, and (b) `docs/en/connectors/sink/Prometheus.md` / `docs/zh/connectors/sink/Prometheus.md` description updates that clarify the `> 0` constraint. This PR's own contribution beyond that is narrower: it keeps the old imperative `checkIntArgument` validation (still only enforced at writer-construction time on a worker, not at submission time) and adds three focused unit tests directly against `PrometheusSinkConfig.loadConfig()`.

Practical consequence if both merge independently: they touch the exact same 4-line hunk of `PrometheusSinkConfig.java` in incompatible ways (this PR keeps `checkIntArgument`/the `checkArgument` static import; `#11738` removes both). Whichever of the two merges second will hit a real merge conflict on that hunk against `dev`, not just a stylistic overlap.

Given `#11738` is currently the more complete fix (same correctness fix + earlier, structured validation + docs) and already has two of my own approvals plus one from @nzw921rx (though those approvals predate the `6717dd04ca` fix commit and a subsequent `dev` merge, so they should be treated as needing a fresh look rather than as still-current sign-off on the latest head) — my recommendation is to close this PR in favor of `#11738`, or, if the author/maintainers prefer to land the narrowly-scoped "Bug" fix first for a faster/simpler diff, land this one and have `#11738`'s author rebase to drop the now-redundant default-application hunk (keeping only the `OptionRule`/`Conditions` migration and docs). Either path is fine technically; what shouldn't happen is merging both without coordination.

Aside from the duplication, compatibility of this PR's own change is **fully compatible**: accepted value range for `batch_size` is unchanged, and the fixed default (`1024`) exactly matches what's already documented in `docs/en/connectors/sink/Prometheus.md:50` and `docs/zh/connectors/sink/Prometheus.md` — this is a correction of code-vs-docs divergence, not a new incompatibility, so no `docs/en/introduction/concepts/incompatible-changes.md` entry is needed.

Negligible and net-positive. `checkIntArgument`'s `checkArgument(args > 0)` is a cheap comparison invoked once per writer construction (once per task start), not per-row. The behavioral side effect — jobs that previously ran with size-based flushing silently disabled will now flush every 1024 rows by default — is the intended fix, not a regression; it reduces unbounded in-memory buffer growth for long-running/low-throughput jobs that don't hit any other flush trigger before this fix (only `flush_interval`, if configured, or `close()` would have flushed otherwise).

No new issues. `checkIntArgument`'s existing unchecked, message-less `IllegalArgumentException` behavior is preserved unchanged for out-of-range values; this PR doesn't touch logging.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Minimal, focused diff. The added inline comment (`PrometheusSinkConfig.java`, above the new `batchSize` line) correctly explains *why* `get()` is used instead of `getOptional().isPresent()`, which is exactly the kind of "explain the why/constraint" comment this project wants — good practice. No license-header or import issues.

`PrometheusSinkConfigTest.java` (new) adds three deterministic, focused unit tests directly against `PrometheusSinkConfig.loadConfig()`:
- `shouldApplyBatchSizeDefaultWhenNotConfigured` — omits `batch_size`, asserts `getBatchSize()` equals `PrometheusSinkOptions.BATCH_SIZE.defaultValue()`. This is the test that actually pins down the bug this PR fixes; it would have failed against the pre-fix code (`getBatchSize()` would have returned `0`).
- `shouldUseConfiguredBatchSize` — explicit `batch_size = 5` is respected.
- `shouldRejectNonPositiveBatchSize` — `batch_size = 0` still throws.

**Stability rating: Stable.** Pure unit tests over an in-memory `ReadonlyConfig.fromMap(...)`; no I/O, no threading, no time-based waits, no shared mutable state between tests (`map` is a fresh `HashMap` per test). No plausible flakiness source.

## 2.3 Documentation Updates

No docs changes in this PR, and none are strictly required — the fix restores behavior that already matches the documented default (`docs/en/connectors/sink/Prometheus.md:50`, `batch_size | Int | false | 1024`), so there's no code/docs mismatch to correct on the docs side.

# 3. Architectural Soundness

The fix itself is correct and minimal — a one-line behavioral change (drop the `isPresent()` guard) with a clear explanatory comment. Taken in isolation (i.e., if `#11738` didn't already cover the same ground), I'd have no notes here.

## 3.2 Maintainability

Keeps the existing imperative `checkIntArgument` validation pattern rather than moving to the project's `OptionRule`/`Conditions` declarative-validation idiom that `#11738` adopts. That's not wrong, but it does mean `batch_size` validation remains split from where the option is declared (`PrometheusSinkFactory.optionRule()`) and only surfaces at writer-construction time on a worker rather than at job-submission time — a strictly worse failure-mode than what `#11738` already provides for the same underlying option.

## 3.3 Extensibility

No concerns beyond the duplication noted above.

No serialization, checkpoint/savepoint, or protocol surface touched. Existing jobs that already set `batch_size` explicitly are completely unaffected (`getOptional(...).isPresent()` was already `true` for them before this change, and remains functionally equivalent after). Only jobs that omit `batch_size` are affected, and only in the direction of matching already-documented behavior.

# 4. Issue Summary

|---|---|---|---|
| 1 | This PR's fix (drop the `isPresent()` guard around `setBatchSize`) is now duplicated by commit `6717dd04ca` on `#11738`, pushed ~5.5h after this PR opened; `#11738`'s current head is a strict superset (same fix + `OptionRule`/`Conditions` fail-fast validation + docs). Merging both independently will produce a real merge conflict on the same hunk of `PrometheusSinkConfig.java`. Needs consolidation (close one in favor of the other, or rebase) before either can safely merge. | `seatunnel-connectors-v2/connector-prometheus/.../config/PrometheusSinkConfig.java:51-58` (this PR) vs. same file on `#11738` head `6717dd04ca` | High |

# 5. Merge Recommendation

### Conclusion: Not recommended for merge (as currently scoped — needs coordination with #11738, not a rewrite)

   - Issue 1: Coordinate with `#11738` before merging either PR. Recommended path: close `#11797` and let `#11738` carry this fix forward (it already contains the identical correctness fix plus fail-fast `OptionRule` validation and docs) — optionally, cherry-pick this PR's three `PrometheusSinkConfigTest` unit tests into `#11738` if the maintainers want both direct-unit-test coverage of `loadConfig()` and the factory-level `OptionRule` tests already there, since those aren't redundant with each other. Alternatively, if the narrower "Bug" fix is preferred to land first for a smaller/faster diff, merge this one and have `#11738`'s author rebase to drop the now-duplicate hunk.

   - None beyond the above; the code itself, taken alone, is correct and well-tested.

Overall assessment: the underlying bug is real (I helped identify it as a low-severity follow-up item during my `#11778` review), and this PR's fix for it is correct, minimal, and well-tested in isolation. But right now it isn't the only PR carrying this fix, and it isn't the more complete one — `#11738` already has it plus more, as of a same-day commit. I don't want to approve a fix that will conflict with a further-along, twice-approved sibling PR without the two being reconciled first, so I'm not recommending merge as-is; this needs a quick coordination step between the two PR authors/maintainers, not further code changes to this PR itself.

davidzollo:

```

```

davidzollo (approved):
+1
LGTM
Good job

PR #11805, opened 2026-08-14
[Hotfix][Transform-V2][DataValidator] Fix ClassCastException when RANGE rule validates decimal fields

### LINKED ISSUE #11804: [Bug][Transform-V2][DataValidator] ClassCastException when RANGE rule validates a DECIMAL field
### Search before asking

### What happened

The DataValidator transform throws a `ClassCastException` when the `RANGE` rule validates a `DECIMAL` field against integer bounds.

### SeaTunnel Version

2.3.13 (also reproducible on the `dev` branch)

### SeaTunnel Config

```conf
transform {
  DataValidator {
    plugin_input = "Jdbc"
    plugin_output = "DataValidator"
    field_rules = [
      {
        field_name = "amount"
        rules = [
          { rule_type = "RANGE", min_value = 0, max_value = 1000 }
        ]
      }
    ]
  }
}
```

### Running Command

```shell
1. Create a JDBC source reading a table with a `DECIMAL(10,2)` column (e.g. `amount`).
2. Add a `DataValidator` transform with a `RANGE` rule on that column using integer `min_value`/`max_value`.
3. Submit the job. The task fails at the transform step.
```

### Error Exception

```log
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.math.BigDecimal
      at java.math.BigDecimal.compareTo(BigDecimal.java:305)
      at org.apache.seatunnel.transform.validator.rule.RangeValidationRule.validate(RangeValidationRule.java:65)
      at org.apache.seatunnel.transform.validator.FieldValidator.validate(FieldValidator.java:61)
      at org.apache.seatunnel.transform.validator.DataValidatorTransform.transformRow(DataValidatorTransform.java:103)
```

Zeta 2.3.13

### Java or Scala Version

Java 17

### Screenshots

_No response_

### Code of Conduct

BinTaoMa:
## Purpose

Fixes `ClassCastException` / `NumberFormatException` crashes when the
DataValidator transform's RANGE rule validates numeric fields whose type
doesn't line up with how the bounds are parsed.

## Problem

- The field value of a DECIMAL column is read as `java.math.BigDecimal`, but
  RANGE bounds (`min_value` / `max_value`) are parsed by
  `DataValidatorTransformConfig#parseComparable` into `Integer` / `Long` /
  `Double` — never `BigDecimal`.
- `RangeValidationRule#validate` then calls `Comparable.compareTo` between a
  `BigDecimal` and an `Integer`, throwing:
  `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.math.BigDecimal`
- Non-finite floating point values (`NaN`, `±Infinity`) cannot be represented
  as `BigDecimal` (`new BigDecimal("NaN")` throws `NumberFormatException`), and
  comparing a `Double` against integer/long bounds via `Comparable.compareTo`
  throws `ClassCastException` too — crashing the whole task instead of
  rejecting the row.

## Solution

`RangeValidationRule#validate` normalizes numeric values and numeric bounds to
a common type before comparing:

- **Finite numbers** are normalized to `BigDecimal`, so a `BigDecimal` field
  compares cleanly against `Integer`/`Long`/`Double` bounds.
- **Non-finite numbers** (`NaN`, `±Infinity`) are normalized to `Double`
  together with any numeric bounds, so `Double.compareTo` handles them the same
  way regardless of how the bound was written. Non-finite values sort beyond
  every finite value, so they are rejected as out of range rather than
  crashing.
- Non-numeric (`String`) values keep the existing `Comparable` semantics.

## Known limitation

RANGE is intended for numeric fields only. For non-numeric field types (e.g.
`STRING`), the value is compared via `Comparable.compareTo` against the parsed
bound, which can still throw `ClassCastException` if the bound was parsed as a
number.

## Verification

- `rangeRuleShouldAcceptDecimalFieldWithIntegerBounds` — DECIMAL field against
  integer bounds (fails with `ClassCastException` before the fix).
- `rangeRuleShouldRouteDecimalAboveMaxToErrorTable` /
  `rangeRuleShouldRejectDecimalEqualToMaxWhenMaxNotInclusive` /
  `rangeRuleShouldCompareDecimalFieldAgainstDoubleBounds` — decimal vs double
  bounds and ROUTE_TO_TABLE routing.
- `rangeRuleShouldRouteNonFiniteDoubleToErrorTableInsteadOfThrowing` — NaN
  against double bounds.
- `rangeRuleShouldRouteNonFiniteDoubleToErrorTableWithIntegerBounds` — NaN /
  `+Infinity` against integer-literal bounds.

`mvn -pl seatunnel-transforms-v2 -Dtest=DataValidatorTransformTest test`

Closes #11804


DanielLeens (changes requested):
Welcome to Apache SeaTunnel, @BinTaoMa, and thank you for this contribution! You did the two things that make a bug report easy to act on: you filed #11804 with a real stack trace first, and you added a test that genuinely fails before the fix and passes after. That is a great first PR.

I traced the full path from config parsing through to `transformRow` and I agree completely with your diagnosis of the root cause. There is one issue in the fix itself that I would like resolved before merge, plus a few improvements. Details below.

A `DataValidator` transform with a `RANGE` rule on a `DECIMAL` column dies with `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.math.BigDecimal` (#11804).

The cause is a type mismatch across two independent code paths:
- The field value of a `DECIMAL(p, s)` column arrives in `SeaTunnelRow` as `java.math.BigDecimal`.
- The `min_value` / `max_value` bounds go through `DataValidatorTransformConfig.parseComparable` (`DataValidatorTransformConfig.java:307-327`), which only ever returns `Integer`, `Long`, `Double` or `String` — never `BigDecimal`.

`RangeValidationRule.validate` then calls `BigDecimal.compareTo(Integer)`, and `BigDecimal.compareTo` casts its argument to `BigDecimal` unconditionally.

One-sentence summary: normalize both sides to `BigDecimal` before comparing, so that numeric values and numeric bounds of different concrete types compare correctly.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The production change is `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/validator/rule/RangeValidationRule.java:61-112`.

Before:

```java
Comparable comparableValue = (Comparable) value;

if (minValue != null) {
    int minComparison = comparableValue.compareTo(minValue);   // BigDecimal.compareTo(Integer) -> CCE
```

After:

```java
Comparable comparableValue = (Comparable) value;
Comparable min = minValue;
Comparable max = maxValue;

if (value instanceof Number) {
    comparableValue = new BigDecimal(value.toString());
    if (min instanceof Number) {
        min = new BigDecimal(min.toString());
    }
    if (max instanceof Number) {
        max = new BigDecimal(max.toString());
    }
}
```

I confirmed the diagnosis is exact rather than plausible:

1. `parseRangeRuleFromMap` (`DataValidatorTransformConfig.java:211-221`) calls `parseComparable(String.valueOf(minValue))`. For `"0"` there is no `.`, so it takes the `Long.parseLong` branch and narrows to `Integer` (`DataValidatorTransformConfig.java:317-321`). Your test's `min_value = "0"` / `max_value = "1000"` therefore produce `Integer(0)` / `Integer(1000)` — exactly the shape from the issue.
2. The flat `field_rules` form used in your test (`field_name` / `rule_type` / `min_value` at one level) is a supported shape — `DataValidatorTransformConfig.java:95-106` falls through to `parseValidationRuleFromMap` when no nested `rules` key is present — so the test really does construct a `RangeValidationRule` and really does reach the comparison. This matters: had the flat form been unsupported, the rule would have been silently dropped and the test would have passed vacuously both before and after the fix. It is not vacuous.
3. `BigDecimal.compareTo` ignores scale, so `new BigDecimal("99.90").compareTo(new BigDecimal("1000"))` is correct and `new BigDecimal("100.00").compareTo(new BigDecimal("100"))` returns `0` — the inclusive-bound semantics are preserved, which is easy to get wrong if someone reaches for `equals` instead.
4. Going through `value.toString()` rather than `doubleValue()` is the right choice for `Float`: `new BigDecimal(Float.toString(0.1f))` gives exactly `0.1`, whereas `BigDecimal.valueOf(((Number) v).doubleValue())` on a float would surface the binary-widening artifact. Good instinct.
5. The failure messages still interpolate the **original** `value` and `minValue` / `maxValue`, not the normalized copies (`RangeValidationRule.java:86`, `RangeValidationRule.java:96`). That keeps user-facing messages byte-identical to previous releases. Also deliberate and correct — please keep it that way.

Runtime path — this is a per-record hot path, which matters for section 1.3:

```
Zeta/Flink/Spark task loop, once per record
  -> DataValidatorTransform.transformRow(SeaTunnelRow)   DataValidatorTransform.java:80
     -> for each FieldValidator                          DataValidatorTransform.java:91-105
        -> FieldValidator.validate(fieldValue, ctx, false)   FieldValidator.java:56-70
           -> RangeValidationRule.validate(...)          RangeValidationRule.java:57   <-- CHANGED
     -> ValidationResultHandler.processResults(...)      DataValidatorTransform.java:108
     -> errorHandleWay == FAIL      -> throw TransformCommonError.validationFailed   :117-120
        errorHandleWay == SKIP      -> return null                                   :121-122
        errorHandleWay.allowRouteToTable() -> tag row with error tableId             :123-138
```

Note there is **no try/catch around `rule.validate(...)`** anywhere in `FieldValidator.validate` (`FieldValidator.java:59-67`) or `DataValidatorTransform.transformRow` (`DataValidatorTransform.java:91-105`). Any unchecked exception thrown inside a rule escapes the transform entirely and fails the task, completely bypassing the configured `row_error_handle_way`. That is the mechanism behind both the original bug and Issue 1 below.

**Does the normal user path reach the changed logic?** Yes — this is the main per-record transform path for every `RANGE` rule on every engine, not an edge or recovery path. That cuts both ways: the fix is high value, and any new failure mode it introduces is also on the main path.

## 1.2 Compatibility Impact

**Partially incompatible** — one narrow but real behavioral regression, otherwise compatible.

Compatible cases (verified by walking the type combinations):

| value type | bound type | before | after |
| --- | --- | --- | --- |
| `Integer` / `Long` | `Integer` / `Long` | works | works, same result |
| `Double` | `Double` (bound written with `.`) | works | works, same result — **except NaN/Infinity, see Issue 1** |
| `BigDecimal` | `Integer` / `Long` / `Double` | **CCE** | works — this is the fix |
| `Integer` / `Long` | `Double`, e.g. `min_value = "0.0"` | **CCE** | works — bonus fix, not mentioned in the description |
| `String` | `String` | works | works (untouched, `value instanceof Number` is false) |
| `String` | `Integer` | **CCE** | still **CCE** — see Issue 2 |
| `Number` | `String` | **CCE** | still **CCE** — see Issue 2 |

Worth calling out that the `Integer` value + `Double` bound row is also fixed by this change, which is a nice side effect you can mention in the description.

No config option is added, renamed or re-defaulted; no serialization format changes; `RangeValidationRule` still Java-serializes and Jackson-deserializes identically (no new fields). Existing jobs whose rules currently work keep working, with the NaN/Infinity exception below. No `incompatible-changes.md` entry needed.

`transformRow` runs once per record. With this change, every record passing a `RANGE` rule on a numeric field now allocates up to three `String` objects (`value.toString()`, `min.toString()`, `max.toString()`) plus up to three `BigDecimal` objects, and performs `BigDecimal` comparisons instead of primitive ones.

That happens **unconditionally for all numeric values**, including the by-far most common case where the value and the bound are already the same type (`Integer` field with `Integer` bounds — the example in `docs/en/transforms/data-validator.md:146-148` and in `data_validator_valid.conf:69-73`), where no conversion is needed at all. It also happens when both bounds are `null`, in which case the conversion is pure waste.

`BigDecimal` construction from a `String` is not cheap — it parses character by character — so on a high-throughput job this is measurable young-gen pressure and a slower comparison on a path that previously did primitive `Integer.compareTo`. See Issue 3 for the concrete remedy.

No thread-safety, resource-leak or concurrency concern: the method remains stateless, and `min` / `max` are new local variables rather than mutations of the shared rule fields — that part is done correctly, and mutating `minValue` / `maxValue` in place would have been a real bug since `RangeValidationRule` instances are shared across all records.

### Issue 1

- **Location**: `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/validator/rule/RangeValidationRule.java:70`
- **Problem description**: `new BigDecimal(value.toString())` throws `java.lang.NumberFormatException` when `value` is a `Double` or `Float` holding `NaN`, `Infinity` or `-Infinity`, because `Double.toString(Double.NaN)` is `"NaN"` and `BigDecimal(String)` rejects it. This is a **regression**: previously, a `DOUBLE` field with bounds written using a decimal point (`min_value = "0.0"` -> `Double` via `parseComparable`, `DataValidatorTransformConfig.java:314-315`) compared via `Double.compareTo(Double)`, which handles `NaN` fine (it orders as greater than everything) and correctly reported the row as out of range. After this change the same job throws.
- **Potential risk**: This is more serious than the narrow trigger suggests, for two reasons. First, `NaN` / `Infinity` is exactly the class of dirty data a `DataValidator` `RANGE` rule exists to catch, so the rule now crashes on its target input. Second, as traced in 1.1, nothing catches unchecked exceptions between `RangeValidationRule.validate` and the task boundary — `FieldValidator.java:59-67` and `DataValidatorTransform.java:91-105` have no try/catch — so the `NumberFormatException` escapes the transform and fails the whole job, **bypassing `row_error_handle_way = SKIP` and `ROUTE_TO_TABLE`**. A user who explicitly configured "route bad rows to an error table" gets a dead job instead. Preconditions are narrow (float/double column + decimal-point bounds + non-finite value), but the blast radius when hit is total.
- **Best improvement**:
  - *Option A (minimal, recommended)*: only take the `BigDecimal` path when it is actually needed, and let same-type numeric comparison keep its existing semantics:
    ```java
    // Only normalize when the two operands are different concrete numeric types;
    // same-type values already compare correctly and non-finite floating point
    // values cannot be represented as BigDecimal.
    if (value instanceof Number && needsNormalization(value, min, max)) { ... }
    ```
    where `needsNormalization` returns false when the value is a non-finite `Double`/`Float` and false when `value.getClass()` already equals the bound's class. This fixes the regression and Issue 3 at the same time.
  - *Option B*: keep the current structure but guard the conversion explicitly:
    ```java
    if (value instanceof Number && isFinite((Number) value)) { ... }
    ```
    with a small private `isFinite` covering `Double.isNaN` / `Double.isInfinite` / `Float.isNaN` / `Float.isInfinite`. Simpler, but leaves the allocation cost of Issue 3 in place.
  - Separately, and independent of which option you pick, I would suggest a defensive `catch (RuntimeException e)` in `FieldValidator.validate` that converts an unexpected rule failure into `ValidationResult.failure(...)` so `row_error_handle_way` is honored. That is arguably out of scope for this PR — happy for it to be a follow-up issue — but it is the structural reason a single bad cast can kill a job.
  - Please add a test: `DOUBLE` field, `min_value = "0.0"`, `max_value = "100.0"`, value `Double.NaN`, asserting no exception is thrown and the row is handled by the configured error path.
- **Severity**: High

### Issue 2

- **Location**: `RangeValidationRule.java:69-77`
- **Problem description**: The normalization is one-directional and only covers `Number`-vs-`Number`. Two mixed combinations still throw the exact `ClassCastException` from #11804:
  1. Numeric value, non-numeric bound. `parseComparable` falls back to returning the raw `String` for anything it cannot parse (`DataValidatorTransformConfig.java:323-326`) — for example `min_value = "1,000"` (thousands separator) or a typo. `min` then stays a `String`, and `BigDecimal.compareTo(String)` throws.
  2. Non-numeric value, numeric bound. A `STRING` column with `min_value = 0` gives `String.compareTo(Integer)`. Likewise a `DATE` / `TIMESTAMP` column with string bounds gives `LocalDate.compareTo(String)`.
- **Potential risk**: Users hitting these variants get the same opaque `ClassCastException` and the same job kill, and will reasonably reopen #11804 believing the fix did not work. The underlying root cause — that `minValue` / `maxValue` are raw `Comparable` with no validation against the field's `SeaTunnelDataType` — is untouched.
- **Best improvement**: A complete fix belongs at config-parse time, where you have the information to fail fast with a clear message:
  - *Option A (preferred long term)*: validate in `DataValidatorTransform.initializeFieldValidators()`, where the field's `SeaTunnelDataType` is known, and throw a descriptive config error at job-construction time — "RANGE rule on field 'amount' (DECIMAL(10,2)) has non-numeric min_value 'abc'" — instead of a `ClassCastException` on record one.
  - *Option B (in scope for this PR)*: add a final defensive branch so that a genuinely incomparable pair returns `ValidationResult.failure("...")` rather than propagating a raw `ClassCastException`. Even a two-line guard would turn a job kill into a routed error row.
  - At minimum, please note in the PR description that mixed numeric/string bound combinations remain unfixed, so #11804 is not closed under a false assumption of completeness.
- **Severity**: Medium

# 2. Code Quality Assessment

## 2.1 Coding Standards

- Formatting is Spotless-clean and the import block ordering is correct.
- The explanatory comment at `RangeValidationRule.java:66-69` is genuinely good — it states the *why* (which two code paths disagree, and what the resulting exception is) rather than restating the code. This is exactly the standard the project wants for non-obvious compatibility logic.
- Comment/Javadoc completeness: no new public methods or fields are introduced, so nothing is missing. If you adopt Option A of Issue 1 and add a private `needsNormalization` helper, please give it a short Javadoc explaining the non-finite and same-type exclusions — it is precisely the kind of subtle predicate that gets "simplified" away later.
- Raw `Comparable` is used throughout (`Comparable min = minValue;`), producing unchecked-call warnings. That is pre-existing style in this class, so I would not change it here.

### Issue 4

- **Location**: `RangeValidationRule.java:63-77`
- **Problem description**: `Comparable comparableValue = (Comparable) value;` at line 63 is immediately and unconditionally overwritten at line 70 whenever `value instanceof Number`. The dead initial assignment makes the control flow harder to follow than it needs to be, and obscures the fact that there are really two distinct comparison modes (numeric and lexicographic).
- **Potential risk**: Readability only. But a future reader may not notice that the `String` path silently keeps raw `Comparable` semantics, which is the seam described in Issue 2.
- **Best improvement**: Extract the numeric branch into a small private method — `private ValidationResult validateNumeric(Number value, ...)` — and keep `validate` as a two-way dispatch. The two modes become explicit and each is independently testable.
- **Severity**: Low

`DataValidatorTransformTest.rangeRuleShouldAcceptDecimalFieldWithIntegerBounds` (`DataValidatorTransformTest.java:146-175`) is a good regression test: it is the precise scenario from #11804, and I verified it is not vacuous (see 1.1, point 2). The comment above it explaining the pre-fix failure is a nice touch.

**Test stability rating: Stable.**

Evidence, per the flaky-test anti-pattern checklist:
- No `Thread.sleep`, no `Awaitility`, no polling, no timing or ordering dependence.
- No shared static or instance state — a fresh `ReadonlyConfig`, `CatalogTable`, `DataValidatorTransform` and `SeaTunnelRow` are constructed inside the test method, so it is order-independent and parallel-safe alongside the three existing tests in the class.
- The assertion `assertEquals(new BigDecimal("99.90"), result.getField(0))` uses scale-sensitive `BigDecimal.equals`, which would normally be a flakiness smell — here it is safe and deterministic because `transformRow` returns `inputRow` unmodified (`DataValidatorTransform.java:145`), so the exact same `BigDecimal("99.90")` instance comes back. No floating-point delta concern.
- No container, network, filesystem or external resource dependency.

Coverage gaps (the first is the one I would like addressed):

### Issue 5

- **Location**: `seatunnel-transforms-v2/src/test/java/org/apache/seatunnel/transform/validator/DataValidatorTransformTest.java:146-175`
- **Problem description**: Only the "decimal inside the range, passes" case is asserted. There is no test that a decimal value **outside** the range is still correctly rejected. As written, a hypothetical future change that made `RANGE` skip decimals entirely would leave this test green while silently disabling validation for every `DECIMAL` column in production.
- **Potential risk**: The test proves the `ClassCastException` is gone, but not that range checking still *works* on decimals. That is the assertion that protects data quality.
- **Best improvement**: Add companion cases, all cheap:
  1. `BigDecimal("1500.00")` against `min_value = "0"` / `max_value = "1000"` -> expect the row routed/failed, i.e. assert the validation actually fires.
  2. Boundary equality: `BigDecimal("1000.00")` with `max_inclusive = true` -> pass; with `max_inclusive = false` -> fail. This locks the scale-insensitive `compareTo` behavior described in 1.1 point 3.
  3. `Long` / `Double` bounds against a `BigDecimal` field (`min_value = "0.0"`), covering the `Double` branch of `parseComparable`.
  4. The `Double.NaN` case from Issue 1.
  Optionally, `seatunnel-e2e/seatunnel-transforms-v2-e2e/seatunnel-transforms-v2-e2e-part-1/src/test/resources/data_validator_valid.conf:68-73` already contains a `RANGE` rule; adding a `DECIMAL` column there would give end-to-end coverage for very little effort.
- **Severity**: Medium

### Issue 3

- **Location**: `RangeValidationRule.java:69-77`
- **Problem description**: Bounds are re-normalized from `String` on **every record**, even though they are immutable configuration fixed at job-construction time, and the conversion runs even when the value and bound are already the same type or when both bounds are `null`.
- **Potential risk**: Avoidable allocation and CPU on a per-record hot path (up to 3 `String` + 3 `BigDecimal` per record per `RANGE` rule). On a high-throughput job with several range rules this becomes real young-gen churn and GC pressure, for zero benefit in the common `Integer`-vs-`Integer` case.
- **Best improvement**:
  - *Option A*: short-circuit first — `if (minValue == null && maxValue == null) { return ValidationResult.success(); }` before any conversion. One line, removes the pure-waste case.
  - *Option B*: normalize the bounds once instead of per record. Two viable shapes: (i) convert inside `DataValidatorTransformConfig.parseRangeRuleFromMap` (`DataValidatorTransformConfig.java:211-221`) so the rule receives already-`BigDecimal` bounds — but note this would change the `%s` rendering in the failure messages, so you would want to keep the original text separately; or (ii) add a lazily-initialized `transient` cache field inside `RangeValidationRule`. If you take (ii), the cache must be `transient` **and** `@JsonIgnore`-annotated: `ValidationRule extends Serializable` (`ValidationRule.java:43`) and the rule is Jackson-polymorphic via `@JsonTypeInfo` (`ValidationRule.java:31-41`), so an un-annotated extra field would leak into both the Java-serialized form shipped to workers and the JSON round-trip. `@Data` would also generate a public getter for it.
  - *Option C*: skip `BigDecimal` entirely for the fast paths, the way the codebase already does elsewhere — see `ConditionEvaluators.compareNumberValues` (`seatunnel-api/src/main/java/org/apache/seatunnel/api/configuration/util/ConditionEvaluators.java:202-217`), which uses `Long.compare` for integral types, `Double.compare` when either side is floating point, and only falls back to `BigDecimal` when one side actually is a `BigDecimal`. That is the same problem you are solving, already solved once in this repository, and it happens to sidestep the `NaN` regression in Issue 1 for free. It is package-private so you cannot call it directly, but it is a good template to mirror.
  - Given Option C also resolves Issue 1, it is my recommendation if you have the appetite; Option A + Issue 1 Option B is the minimal acceptable combination.
- **Severity**: Medium

## 2.3 Documentation Updates

No documentation change is required — this is a bug fix with no user-facing config surface change, and both `docs/en/transforms/data-validator.md` and `docs/zh/transforms/data-validator.md` remain accurate. So en/zh parity is not at risk.

One optional improvement: `docs/en/transforms/data-validator.md:77-83` describes `RANGE` as validating "that a numeric value is within a specified range" without saying which types are supported. Given the confusion this bug caused, a sentence stating that numeric bounds are compared numerically across `INT` / `BIGINT` / `FLOAT` / `DOUBLE` / `DECIMAL`, while non-numeric fields fall back to lexicographic `Comparable` ordering, would be genuinely useful. Fully optional, and if you do add it please mirror it into the `zh` file.

# 3. Architectural Soundness

This is a **targeted fix at the right layer for the reported symptom**, but it is a **partial** one rather than a complete root-cause fix. The true root cause is that `parseComparable` (`DataValidatorTransformConfig.java:307-327`) infers bound types from string syntax alone, with no knowledge of the target field's `SeaTunnelDataType`, and stores them as raw `Comparable`. Normalizing at comparison time papers over that at the point of use.

That is a defensible tradeoff for a hotfix — the comparison site is the one place where both operands are in hand, and fixing `parseComparable` properly would touch the config contract. I would accept the approach. What I would ask is that the PR description say so explicitly (see Issue 2), so the residual gap is recorded rather than assumed closed.

## 3.2 Maintainability

The comment carries the reasoning well, so the next reader will understand why `BigDecimal` appears here. The main maintainability concern is the asymmetry described in Issue 2 and Issue 4: the method now has two implicit modes with no structural signal separating them. The extraction suggested in Issue 4 would fix that cheaply.

## 3.3 Extensibility

No new abstraction or interface is introduced; `ValidationRule` and the `@JsonSubTypes` registry (`ValidationRule.java:31-41`) are untouched, so no other rule implementation is affected. If a future `SeaTunnelDataType`-aware bound-coercion step is added at parse time (the Issue 2 Option A direction), this normalization would become redundant and could simply be deleted — that is a good property for a hotfix to have.

- **Config contract**: `min_value` / `max_value` / `min_inclusive` / `max_inclusive` / `custom_message` semantics and parsing are unchanged. Existing job JSON/HOCON parses identically.
- **Serialization**: no field added or removed on `RangeValidationRule`, so both the Java-serialized form shipped to task workers and the Jackson polymorphic form are unchanged. A job upgraded in place will not fail to deserialize its rules. (This is worth preserving — please see the `transient` / `@JsonIgnore` caveat in Issue 3 Option B(ii) if you add a cache field.)
- **Checkpoint/state**: transforms hold no checkpointed state here; `transformRow` is stateless. Restore is unaffected.
- **Behavioral compatibility**: as tabulated in 1.2, every previously-working combination keeps working **except** non-finite floating point values, which is Issue 1 and is why I am treating it as a blocker rather than a nitpick.

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | `new BigDecimal(value.toString())` throws `NumberFormatException` on `Double`/`Float` `NaN` / `Infinity`; a regression, and it escapes the transform uncaught so `row_error_handle_way` is bypassed and the job dies | `RangeValidationRule.java:70`; escape path `FieldValidator.java:59-67`, `DataValidatorTransform.java:91-105` | High |
| 2 | Fix is one-directional: numeric value vs `String` bound, and `String`/date value vs numeric bound, still throw the same `ClassCastException` | `RangeValidationRule.java:69-77`; root cause `DataValidatorTransformConfig.java:307-327` | Medium |
| 3 | Immutable bounds re-converted to `BigDecimal` from `String` on every record, including when both operands are already the same type or both bounds are null | `RangeValidationRule.java:69-77` | Medium |
| 4 | Dead initial assignment to `comparableValue`; two comparison modes with no structural separation | `RangeValidationRule.java:63-77` | Low |
| 5 | Test asserts only the in-range decimal case; no assertion that out-of-range decimals are still rejected, no boundary/exclusive-bound case, no `Double`/`Long` bound case | `DataValidatorTransformTest.java:146-175` | Medium |

## CI status

At the time of writing, the `Build` check on head `8f9a903` is `queued`, so there is no result to diagnose yet. `labeler` and `Notify test workflow` are green. The branch is only 3 commits behind `dev`, so no rebase is needed.

Minor housekeeping, entirely optional: head commit `8f9a903` is an empty "trigger CI" commit, and the PR title uses `[Hotfix]` while the project's convention list is `Feature` / `Fix` / `Improve` / `Docs` / `Test` / `Chore` — your underlying commit `4bd717d` already uses the conventional `[Fix][Transform-V2][DataValidator] ...`, so renaming the PR title to match would be nice. Neither affects merging; maintainers squash-merge.

# 5. Merge Recommendation

   1. **Issue 1** — guard the `BigDecimal` conversion against non-finite `Double` / `Float` values. This is a regression on a previously-working configuration, and because nothing catches the exception between the rule and the task boundary it turns a routable bad row into a dead job — defeating the very purpose of `row_error_handle_way`. Please also add the `NaN` test case.

   1. Issue 5 — add the out-of-range decimal assertion and at least one boundary case, so the test proves validation still fires rather than only that the exception is gone.
   2. Issue 3 — at minimum the `minValue == null && maxValue == null` short-circuit; ideally mirror the `ConditionEvaluators.compareNumberValues` fast-path strategy, which resolves Issue 1 as a side effect.
   3. Issue 2 — state the remaining mixed-type gap in the PR description so #11804 is not closed as fully resolved, and consider a follow-up issue for `SeaTunnelDataType`-aware bound validation at config-parse time.
   4. Issue 4 — extract the numeric branch into its own private method.
   5. Optional: the `RANGE` supported-types sentence in `docs/en` + `docs/zh`, and the PR title prefix.

**Overall assessment.** The diagnosis is correct and precisely identified, the chosen fix layer is defensible, and the regression test is real rather than decorative — which is more than many first PRs manage. The single thing standing between this and a merge is that normalizing through `BigDecimal(String)` is not total over the `Number` domain: `NaN` and `Infinity` are `Number`s that `BigDecimal` cannot represent, and the current code assumes they cannot occur. Add that guard and I am happy to approve.

**Alternative implementation note.** If you would rather solve this once for the whole rule set instead of just `RANGE`, the higher-leverage change is to coerce `min_value` / `max_value` against the resolved field `SeaTunnelDataType` in `DataValidatorTransform.initializeFieldValidators()`, where the schema is already available. That would fix Issue 1 and Issue 2 together, eliminate Issue 3 entirely (bounds converted once at construction), and turn a whole family of runtime `ClassCastException`s into clear job-submission-time configuration errors. It is a larger change than a hotfix warrants, so I am not asking for it here — but it would make a strong follow-up PR if you are interested in continuing in this area.

davidzollo:

```

```

DanielLeens (changes requested):
Thanks for the warm welcome note above — nothing actionable in it for this review, so going straight into the re-review.

Full re-review of the current head, not a delta review — I re-read both changed files end to end and re-traced the whole comparison path again rather than only diffing against my last comment.

Reviewed head: `a6582920358397ffe39f868b23966a4880f97a6e`.
Delta since my last review (`8f9a9037cd2`): one commit, "Guard non-finite numbers in range rule and add tests," which adds an `isFinite(Number)` guard around the `BigDecimal` normalization, a `minValue == null && maxValue == null` short-circuit, and four new unit tests.

---

A `DataValidator` transform with a `RANGE` rule on a `DECIMAL` column died with `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.math.BigDecimal` (#11804), because `SeaTunnelRow` delivers `DECIMAL` field values as `BigDecimal` while `DataValidatorTransformConfig.parseComparable` (`DataValidatorTransformConfig.java:307-327`) only ever parses `min_value`/`max_value` into `Integer`, `Long`, `Double`, or `String` — never `BigDecimal` — and `Comparable.compareTo` between mismatched boxed-numeric types throws.

The fix normalizes numeric value and numeric bounds to `BigDecimal` before comparing. This round adds a guard so that step is skipped for non-finite `Double`/`Float` values (which `BigDecimal` cannot represent), plus a short-circuit when both bounds are absent, plus new regression tests for out-of-range decimals, an exclusive-bound boundary case, `Double`-typed bounds, and a non-finite value.

One-sentence summary: the core `BigDecimal`-normalization fix from the previous round is unchanged and still correct for its original scenario, this round's guard genuinely closes the `NumberFormatException` regression I flagged for the *same-type-bound* case, but it reopens the identical crash-the-job failure mode as a `ClassCastException` instead of a `NumberFormatException` whenever the non-finite value is compared against an `Integer`/`Long`-typed bound — which is the more common way users write bounds, and is not covered by the new test.

---

# 1. Code Change Review

## 1.1 Core Logic Analysis

Diff since my last review, in `RangeValidationRule.java`:

```java
// new: short-circuit when there is nothing to compare against
if (minValue == null && maxValue == null) {
    return ValidationResult.success();
}
...
// changed: guard added
if (value instanceof Number && isFinite((Number) value)) {
    comparableValue = new BigDecimal(value.toString());
    if (min instanceof Number) { min = new BigDecimal(min.toString()); }
    if (max instanceof Number) { max = new BigDecimal(max.toString()); }
}
...
private static boolean isFinite(Number value) {
    if (value instanceof Double) { return Double.isFinite((Double) value); }
    if (value instanceof Float) { return Float.isFinite((Float) value); }
    return true;
}
```

I traced every branch this guard creates, not just the one the new test exercises.

**When `value` is non-finite (`isFinite` returns `false`), the entire normalization block — including the conversion of `min`/`max` — is skipped.** `comparableValue` remains the raw `Double`/`Float` instance, and `min`/`max` remain whatever `parseComparable` produced for the configured bound strings: `Double` only if the bound string contains a `.` (e.g. `"0.0"`), otherwise `Integer` (or `Long` for values outside `int` range) — `parseComparable` never returns `Float` or `BigDecimal` (confirmed by reading `DataValidatorTransformConfig.java:307-327` in full).

The code then falls through to line 94: `int minComparison = comparableValue.compareTo(min);`, called through the raw `Comparable` type.

- If `min`/`max` are also `Double` (bound written with a decimal point, e.g. `"0.0"`), this dispatches to `Double.compareTo(Double)` — `Double.isNaN`/`isInfinite` are handled by `Double.compare`'s total ordering (non-finite sorts as documented), so this genuinely works. This is exactly what the new test `rangeRuleShouldRouteNonFiniteDoubleToErrorTableInsteadOfThrowing` (`DataValidatorTransformTest.java`) exercises, with `rangeRuleConfig("0.0", "100.0", true)`.
- **If `min`/`max` is `Integer` or `Long` (bound written *without* a decimal point, e.g. `"0"`/`"100"` — arguably the more natural way most users write an integer-looking bound regardless of the field's actual type), `comparableValue.compareTo(min)` calls `Double`'s (or `Float`'s) `compareTo(Object)` bridge method.** `Double implements Comparable<Double>`, so the compiler generates a synthetic bridge `public int compareTo(Object o) { return compareTo((Double) o); }` to satisfy the raw `Comparable` contract; that bridge unconditionally casts its argument to `Double`. Calling it with an `Integer` argument throws `ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Double` — the exact same exception class, just with the operand types reversed, as the one #11804 originally reported. This is standard JDK bridge-method behavior (not something I'm inferring from behavior I can't see — `Double`/`Float`/`Integer`/`Long`/`BigDecimal` all implement `Comparable<T>` for their own boxed type only, so every one of them generates this same kind of bridge method), so I'm confident in this without needing to execute it.

So the new guard's own comment — "Non-finite floating point values ... are left on the original Comparable path. There Double.compareTo / Float.compareTo handle them gracefully" — is only true when the bound happens to already be the same boxed type as the value. It is not true in general, and `RangeValidationRule` has no way to control what type a user's bound string parses to.

**Is this a regression versus `dev`, or a carryover of my own previously-raised Issue 1?** Neither exactly — it's a partial fix of Issue 1. Before this round, *every* non-finite value (regardless of bound type) hit `new BigDecimal(value.toString())` unconditionally and threw `NumberFormatException` — that was the full scope of my original Issue 1. This round trades that for: `NumberFormatException` is gone when the bound type matches the value type, but the identical failure mode (an uncaught exception escapes `RangeValidationRule.validate` all the way past the task boundary, since — I re-confirmed this is still true, see below — `FieldValidator.java` and `DataValidatorTransform.java` are untouched by this PR and still have no `try`/`catch` around rule evaluation) still occurs, now as a `ClassCastException`, whenever the bound is `Integer`/`Long`-typed. So my original Issue 1 is **partially, not fully, closed** — the fix narrows the trigger condition but does not eliminate the "NaN/Infinity value crashes the whole job instead of being routed as an out-of-range row" failure mode I flagged as the blocker.

I re-verified the "no try/catch" claim is still accurate at the current head: `FieldValidator.validate` (`FieldValidator.java:56-70`) and `DataValidatorTransform.transformRow` (`DataValidatorTransform.java:91-105`) are unchanged by this PR — confirmed via `git diff apache/dev...a6582920358 --stat`, which shows only `RangeValidationRule.java` and its test file touched.

**The `minValue == null && maxValue == null` short-circuit** is correct and matches my previous review's Issue 3 Option A recommendation exactly — a clean, safe, no-op-preserving addition that also incidentally resolves the pure-waste allocation case I flagged.

## 1.2 Compatibility Impact

Re-verified against the current head, extending the type-combination table from my last review with the new guard's effect:

| value type | bound type | before this PR series | after round 1 (`9c85f5692e50`-equivalent for this PR) | after this round |
| --- | --- | --- | --- | --- |
| `BigDecimal` | `Integer`/`Long`/`Double` | CCE | fixed | fixed (unchanged) |
| `Double`/`Float`, finite | `Integer`/`Long`/`Double` | CCE (if mismatched) | fixed | fixed (unchanged) |
| `Double`/`Float`, **NaN/Infinity** | `Double` (bound has a decimal point) | CCE | **NumberFormatException (regression)** | **fixed** — genuinely routes to the error table |
| `Double`/`Float`, **NaN/Infinity** | `Integer`/`Long` (bound has no decimal point) | CCE | NumberFormatException (regression) | **still crashes — now CCE again**, see Issue 1 (carryover) below |
| numeric value | `String` bound | CCE | CCE (untouched) | CCE (untouched) — Issue 2 (carryover), still open |

No config option, serialization field, or checkpoint/state interaction changed — same conclusion as my last review, still holds. `incompatible-changes.md` still not required.

The `minValue == null && maxValue == null` short-circuit removes one pure-waste case from my previous Issue 3, which is a genuine, if small, improvement. The larger allocation concern from Issue 3 — every record with a `RANGE` rule on numeric fields still re-parses `min`/`max` from `String` into a fresh `BigDecimal` on every single call, even in the overwhelmingly common case where the value and bounds are already the same concrete type (e.g. `Integer` field, `Integer` bounds) — is unchanged and still open. Not a blocker, same as last round.

Unchanged structurally: a rule that returns `ValidationResult.failure(...)` is handled correctly by `row_error_handle_way`; an uncaught exception still is not. See Issue 1 below for why this still matters for one bound-type combination.

---

### Issue 1 (carryover from my previous review — partially fixed, not closed)

- **Location:** `seatunnel-transforms-v2/src/main/java/org/apache/seatunnel/transform/validator/rule/RangeValidationRule.java:82-90, 117-125`; escape path unchanged at `FieldValidator.java:56-70`, `DataValidatorTransform.java:91-105`.
- **Problem description:** The new `isFinite` guard prevents the `BigDecimal(String)` `NumberFormatException` I flagged last round, but only when the configured bound happens to already be `Double`-typed (i.e., written with a decimal point, like `"0.0"`). When a `DOUBLE`/`FLOAT` field holds `NaN`/`Infinity`/`-Infinity` and the configured `min_value`/`max_value` is written *without* a decimal point (parses to `Integer`/`Long` via `DataValidatorTransformConfig.parseComparable`), `comparableValue.compareTo(min)` at `RangeValidationRule.java:94` dispatches to `Double`'s (or `Float`'s) synthetic `compareTo(Object)` bridge method, which unconditionally casts its argument to `Double`/`Float` and throws `ClassCastException` when given an `Integer`/`Long`. Since `FieldValidator.validate` and `DataValidatorTransform.transformRow` still have no `try`/`catch` around rule evaluation, this exception escapes the transform and kills the whole job — the exact failure mode `row_error_handle_way = ROUTE_TO_TABLE`/`SKIP` exists to prevent, and the exact failure mode I asked to be closed last round. The new comment above the guard ("Double.compareTo / Float.compareTo handle them gracefully") states this is resolved; it is only resolved for the same-type-bound sub-case.
- **Potential risk:** High, unchanged from last round's rationale — `NaN`/`Infinity` is exactly the class of dirty data a `RANGE` rule exists to catch, and an `Integer`/`Long`-typed bound (e.g. `min_value: 0`, `max_value: 100`, both written as whole numbers) is, if anything, a more natural authoring style than deliberately adding a trailing `.0` — so this is not a narrow corner case, it's close to the default case for anyone validating a `DOUBLE`/`FLOAT` column with round-number bounds.
- **Best improvement:** Extend `isFinite` (or add a second check) so the non-finite-value path also verifies the bound is the *same* boxed type as the value before taking the raw `Comparable` route — e.g., convert `Integer`/`Long` bounds to `Double` (safe: `Integer`/`Long` are always finite and exactly representable) whenever `value` is non-finite, so the comparison always resolves to `Double.compareTo(Double)`:
  ```java
  if (value instanceof Number && isFinite((Number) value)) {
      comparableValue = new BigDecimal(value.toString());
      if (min instanceof Number) { min = new BigDecimal(min.toString()); }
      if (max instanceof Number) { max = new BigDecimal(max.toString()); }
  } else if (value instanceof Double || value instanceof Float) {
      // value is non-finite; BigDecimal can't represent it, so normalize the
      // *bounds* to the value's own boxed type instead, so compareTo's bridge
      // method always sees a matching type.
      double d = ((Number) value).doubleValue();
      comparableValue = value instanceof Float ? (Float) (float) d : (Double) d;
      if (min instanceof Number) { min = ((Number) min).doubleValue(); }
      if (max instanceof Number) { max = ((Number) max).doubleValue(); }
  }
  ```
  (Or equivalent — the point is that whichever branch is taken, `comparableValue` and any numeric `min`/`max` must end up the same concrete boxed type before `compareTo` is called.) Please also extend `rangeRuleShouldRouteNonFiniteDoubleToErrorTableInsteadOfThrowing` (or add a sibling test) using `rangeRuleConfig("0", "100", true)` (no decimal point) with a `Double.NaN` value — that config is exactly what silently still throws today, and it's the one combination the current test suite does not touch.
- **Severity:** High

### Issue 2 (carryover, unchanged, still open)

- **Location:** `RangeValidationRule.java:82-90`; root cause `DataValidatorTransformConfig.java:307-327`.
- **Problem description:** Unchanged from my last review — a numeric value against a `String` bound (e.g. a mistyped `min_value` like `"1,000"`, which `parseComparable` falls back to returning as a raw `String`), or a non-numeric value (`String`/`Date`) against a numeric bound, still throws the original `ClassCastException`. Nothing in this round touches this.
- **Potential risk:** Same as before — users hitting this still get an opaque `ClassCastException` and a dead job, and may reasonably believe #11804 is not actually fixed.
- **Best improvement:** Same as last round — either validate bound/value type compatibility at config-parse time (preferred long-term), or add a defensive fallback in `validate()` that turns a genuinely incomparable pair into `ValidationResult.failure(...)` instead of a raw exception. At minimum, the PR description should note this gap remains, so #11804 is not closed under a false assumption of completeness.
- **Severity:** Medium

---

# 2. Code Quality Assessment

## 2.1 Coding Standards

Formatting is Spotless-clean. The new `isFinite` helper is small, correctly named, and the explanatory comment above the guard states intent — but, per Issue 1, the comment's factual claim ("handle them gracefully") is not accurate in the general case, which is worse than no comment because it will mislead the next reader into believing this is fully solved. I'd ask that the comment be corrected alongside whatever fix is chosen for Issue 1. `isFinite` itself has no Javadoc; given it's a `private static` one-liner used in exactly one place, I would not block on that, consistent with my "package-private/self-describing helper" standard from other reviews — but if Issue 1's fix adds branching logic here, a short Javadoc becomes worthwhile.

The dead initial assignment to `comparableValue` at line 68 (`Comparable comparableValue = (Comparable) value;`, unconditionally overwritten whenever the `if` at line 82 is taken) that I flagged as Issue 4 last round is unchanged — still Low, still open, unaddressed this round.

Four new tests were added, and I re-traced each rather than assuming from names alone:

- `rangeRuleShouldRouteDecimalAboveMaxToErrorTable` — a `BigDecimal(1500.00)` against `min=0`/`max=1000` bounds, asserts routing to the error table. This is exactly the "out of range decimal is still rejected" case I asked for (Issue 5, point 1). Deterministic, no timing dependency.
- `rangeRuleShouldRejectDecimalEqualToMaxWhenMaxNotInclusive` — boundary value equal to `max`, `max_inclusive=false`, asserts rejection. Covers half of Issue 5 point 2 (the exclusive-bound-rejects case); the complementary inclusive-bound-accepts-at-boundary case is not separately asserted, though it's implicit in the original `rangeRuleShouldAcceptDecimalFieldWithIntegerBounds` test's inclusive-range pass. Minor, non-blocking gap.
- `rangeRuleShouldCompareDecimalFieldAgainstDoubleBounds` — `Double`-typed bounds (`"0.0"`/`"1000.0"`) against a `BigDecimal` field, asserts pass. Covers Issue 5 point 3.
- `rangeRuleShouldRouteNonFiniteDoubleToErrorTableInsteadOfThrowing` — covers Issue 5 point 4, but only for the `Double`-typed-bound sub-case, per Issue 1 above; it does not exercise (and therefore did not catch) the `Integer`/`Long`-bound `ClassCastException` gap.

All four new tests are deterministic: no `Thread.sleep`, no `Awaitility`, no shared mutable state — each constructs its own `DataValidatorTransform`/`CatalogTable`/`SeaTunnelRow` locally, consistent with the existing test class's style. **Stability rating: Stable.** No test weakening anywhere — nothing removed, relaxed, or made more permissive.

This is a real, substantial improvement in coverage versus the previous round (my Issue 5 asked for exactly these four scenarios, and three of the four are now fully closed) — I want that credited clearly, even though Issue 1's residual gap means the suite doesn't yet prove the PR's own "NaN/Infinity no longer crashes" goal in general.

## 2.3 Documentation Updates

Unchanged from last round — no doc update required or made; `docs/en`/`docs/zh` remain accurate. My optional suggestion (documenting which types `RANGE` compares numerically) is still open but non-blocking.

---

# 3. Architectural Soundness

Still a targeted, defensible fix at the comparison site rather than the config-parse root cause, consistent with my last review's assessment of the overall approach. This round's guard is a reasonable *shape* of fix for the `NumberFormatException` regression, but as implemented it's incomplete — it treats "is the value finite" as sufficient, when the real invariant that must hold before calling `compareTo` is "do value and bound share a boxed type," which non-finite values can still violate against `Integer`/`Long` bounds.

## 3.2 Maintainability

The `isFinite` extraction is a clean, well-named addition. The comment inaccuracy flagged in 2.1 is a maintainability risk in the same way the pool-behavior comment was in the other PR I reviewed today: a future reader trusts it and doesn't re-derive the actual behavior.

## 3.3 Extensibility

Unchanged — no new abstraction, `ValidationRule`/`@JsonSubTypes` untouched.

Unchanged conclusion from last round: no config contract, serialization, or checkpoint/state change. `RangeValidationRule` still has no new fields, Java-serializes and Jackson-deserializes identically. An in-place upgrade is safe from a compatibility standpoint; the residual behavioral gap in Issue 1 affects *new* comparisons at runtime, not upgrade/restore.

---

# 4. Issue Summary

|---|---|---|---|---|
| 1 | `isFinite` guard only prevents the crash when the bound is also `Double`-typed; `Integer`/`Long`-typed bounds (the more natural authoring style) still throw `ClassCastException` for `NaN`/`Infinity` values, still uncaught, still kills the job | `RangeValidationRule.java:82-90, 117-125` | High | **Partially fixed** — narrowed from "always crashes" to "crashes unless the bound happens to match the value's boxed type" |
| 2 | Numeric value vs. `String` bound, or non-numeric value vs. numeric bound, still throws the original `ClassCastException` | `RangeValidationRule.java:82-90`; root cause `DataValidatorTransformConfig.java:307-327` | Medium | Open, unchanged from last round |
| Carried over | Dead initial assignment to `comparableValue`; two comparison modes with no structural separation | `RangeValidationRule.java:68-90` | Low | Open, unchanged |
| Carried over | Bounds re-converted to `BigDecimal` from `String` on every record even when types already match | `RangeValidationRule.java:82-90` | Medium | Partially addressed — the null/null short-circuit is fixed, the same-type fast-path is not |
| Prior Issue 5 | Test coverage gaps: out-of-range decimal, boundary case, `Double` bound, `NaN` case | `DataValidatorTransformTest.java` | Medium | **Mostly fixed** — 3 of 4 requested cases added; the `NaN` case only covers the `Double`-bound sub-case, which is exactly why it didn't catch Issue 1 |

---

# 5. Merge Recommendation

### Conclusion: Not yet ready to merge — changes requested

**1. Blockers — must be fixed**

- **Issue 1** — the guard added this round narrows but does not close the failure mode I asked to be fixed last round. Please make the non-finite-value path also normalize `Integer`/`Long` bounds to `Double` (or an equivalent fix) so `compareTo` never sees mismatched boxed types regardless of how the bound was written, and add a test using integer-literal bounds (e.g. `rangeRuleConfig("0", "100", true)`) with a `NaN`/`Infinity` value — that's precisely the gap the current test suite doesn't reach. I want to be clear this is not a new problem introduced by carelessness; it's a genuine, good-faith narrowing of the previous bug that just doesn't fully close it, and I should have specified the fix more precisely in my last review rather than accepting "Double.compareTo handles it" without also stating the type-matching precondition — apologies for not flagging that ambiguity the first time.

**2. Recommended fixes — non-blocking**

- Issue 2 — still open; at minimum note the remaining gap in the PR description.
- The `isFinite` code comment should be corrected once Issue 1's fix lands, since as written it currently overclaims.
- Issue 4 (dead assignment / extract numeric branch) — still open, cosmetic.
- The same-type fast path from my original Issue 3 is still not implemented — non-blocking, but worth a follow-up if this rule is on a hot path for high-throughput jobs.

**CI status.** The apache-side `Build` check for the current head is `CANCELLED`, not `FAILURE` — I want to correct that expectation explicitly rather than diagnose a failure that isn't actually present. I checked the contributor's fork (`BinTaoMa/seatunnel`, branch `fix-datavalidator-range-decimal`) directly: the only run against this exact head (`a6582920358397ffe39f868b23966a4880f97a6e`, run `31952328121`) also completed as `cancelled`, and there is no newer run superseding it. The previous head's fork run (`8f9a9037cd2`) did complete with `failure`, but that predates this round's fix and is not evidence about the current head. In short: there is currently no red or green CI signal for `a6582920358` — it simply has not finished running. I'd ask for a fresh CI run (e.g., an empty commit or re-triggering the workflow) so an actual result exists before merge, independent of the source-level Issue 1 above.

**Merge-state check.** `mergeStateStatus` is `BLOCKED` (reflecting my outstanding `CHANGES_REQUESTED`, not a merge conflict); `mergeable` is `MERGEABLE`.

**Overall assessment.** This round makes real, substantial progress: the `NumberFormatException` regression is genuinely fixed for the bound-type case the new test covers, the null/null short-circuit is a clean small win, and three of my four requested test-coverage gaps are closed with solid, non-vacuous tests. What keeps this from being ready is that the fix's own stated goal — "non-finite values are rejected as out-of-range instead of crashing the task" — only holds for one of the two realistic bound-type shapes, and the new test happens to only cover the shape where it already works. I traced the `compareTo` dispatch through the JDK's standard bridge-method behavior for `Double`/`Integer` to be confident this isn't a hypothetical: it's the same class of `ClassCastException` from the original issue, just with the operand types swapped, still reachable through an ordinary, arguably more common, bound-authoring style. Once Issue 1 is closed for both bound-type shapes (with a test that would have caught this), and a completed CI run exists for the actual head being merged, I'm glad to move to approval.

DanielLeens (approved):
Thanks for turning this around so fast — a fix, a corrected comment, and a targeted test landed within hours of my last review. Full re-review of the current head below, not a delta-only pass: I re-read `RangeValidationRule.java` end to end again and re-traced the comparison dispatch from scratch rather than trusting the commit message.

A `DataValidator` transform with a `RANGE` rule on a `DECIMAL` column died with `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.math.BigDecimal`, because `SeaTunnelRow` delivers `DECIMAL` values as `BigDecimal` while `DataValidatorTransformConfig.parseComparable` only ever parses `min_value`/`max_value` into `Integer`, `Long`, `Double`, or `String` — never `BigDecimal` — and `Comparable.compareTo` between mismatched boxed types throws.

The original fix normalized finite numeric values and bounds to `BigDecimal`. My previous round found that its non-finite (`NaN`/`Infinity`) guard only avoided the crash when the bound happened to already be `Double`-typed — an `Integer`/`Long`-typed bound (the more natural way to author a whole-number bound) still threw the identical-class `ClassCastException`, uncaught, still killing the job. This round's commit closes that gap by normalizing the non-finite value **and** any numeric bound to `Double` together, so `compareTo` always sees matching boxed types regardless of how the bound was written.

One-sentence summary: my blocking finding from the previous round is now genuinely closed for both bound-authoring styles, verified against the actual comparison dispatch and confirmed empirically in the fork's CI run rather than by reading the diff alone.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Diff since my last review, in `RangeValidationRule.java`:

```java
// Before:
if (value instanceof Number && isFinite((Number) value)) {
    comparableValue = new BigDecimal(value.toString());
    if (min instanceof Number) { min = new BigDecimal(min.toString()); }
    if (max instanceof Number) { max = new BigDecimal(max.toString()); }
}
// (else: non-finite values fell straight through to the raw Comparable path)
```

```java
// After:
if (value instanceof Number) {
    Number numberValue = (Number) value;
    if (isFinite(numberValue)) {
        comparableValue = new BigDecimal(numberValue.toString());
        if (min instanceof Number) { min = new BigDecimal(min.toString()); }
        if (max instanceof Number) { max = new BigDecimal(max.toString()); }
    } else {
        comparableValue = numberValue.doubleValue();
        if (min instanceof Number) { min = ((Number) min).doubleValue(); }
        if (max instanceof Number) { max = ((Number) max).doubleValue(); }
    }
}
```

I traced the new `else` branch against the exact scenario my last review flagged as still-broken: `value = Double.NaN`, `min_value = "0"`, `max_value = "100"` (no decimal point, so `parseComparable` parses both to `Integer`).

- `isFinite(NaN)` -> `false` -> the `else` branch runs.
- `comparableValue = numberValue.doubleValue()` -> primitive `double` `NaN`, autoboxed to `Double`.
- `min instanceof Number` -> `true` (it's `Integer(0)`) -> `min = ((Number) min).doubleValue()` -> `Double(0.0)`.
- `max` becomes `Double(100.0)` the same way.
- `comparableValue.compareTo(min)` now dispatches to `Double`'s `compareTo(Object)` bridge method with an actual `Double` argument — the cast inside the bridge succeeds, no `ClassCastException`. `Double.NaN` is defined by `Double.compareTo` to be greater than every other value including `POSITIVE_INFINITY`, so the row is correctly reported as out-of-range, not crashed.
- Same reasoning holds for `Double.POSITIVE_INFINITY`/`NEGATIVE_INFINITY` and for `Float` NaN/Infinity (widening `Float`->`double` is exact for these IEEE-754 special values, so no precision concern).

This is precisely the fix I suggested in my last review — implemented more simply than my own sketch: rather than special-casing `Float` to stay `Float`, everything non-finite converges on `Double`. That's a fine simplification — `Double` and `Float`'s non-finite values carry identical semantics for ordering purposes, and using a single target type removes a branch rather than adding one.

I did not stop at reading the code — I pulled the actual Windows unit-test job log from the contributor's fork run against this exact head and found the rule executing exactly as traced:

```
ERROR ...DataValidatorTransform - Validation failed for row: amount: Value NaN exceeds maximum 100
ERROR ...DataValidatorTransform - Validation failed for row: amount: Value Infinity exceeds maximum 100
```

Note the message reads "maximum 100" (the original `Integer` bound), not "100.0" — confirming the failure message still interpolates the un-normalized `minValue`/`maxValue` fields, not the local `min`/`max` copies, so user-facing text stays byte-identical to previous releases. This is real evidence of the fix working end-to-end, not just a static read of the diff.

**Does the fix now handle all numeric type combinations, not just the reported decimal case?** I re-walked every value/bound pairing the rule can encounter:

| value type | bound type | outcome |
| --- | --- | --- |
| `BigDecimal` (finite) | `Integer`/`Long`/`Double` | `BigDecimal` path, unchanged from round 1 — fixed |
| `Integer`/`Long`/`Double`/`Float` (finite) | any numeric | `BigDecimal` path, unchanged — fixed |
| `Double`/`Float`, non-finite | `Integer`/`Long` bound | new this round: both converge to `Double` — fixed |
| `Double`/`Float`, non-finite | `Double` bound | already fixed last round — unchanged, still fixed |
| numeric value | `String` bound (e.g. mistyped `"1,000"`) | untouched — still `ClassCastException`, carried-over, non-blocking |
| non-numeric value (`String`/date) | numeric bound | untouched — still `ClassCastException`, same carried-over issue |

The finite-value `BigDecimal` branch is byte-for-byte unchanged from the version I already assessed as correct, so there is no re-litigation needed there.

## 1.2 Compatibility Impact

Fully compatible with previously-working configurations, and the one narrow regression I flagged is now closed for both bound-authoring styles. No config option added, renamed, or re-defaulted; no serialization change — `RangeValidationRule` still has no new fields. No `incompatible-changes.md` entry needed.

No new concern introduced. The non-finite branch is on a cold path (only taken when the field value is actually `NaN`/`Infinity`), so the extra `doubleValue()` calls there are immaterial. The performance observation from my prior reviews — every record on the finite/common path still re-parses bounds from `String` into a fresh `BigDecimal`, with no same-type fast path — is unchanged by this commit and remains a non-blocking follow-up.

My prior blocker is fully resolved this round, with before/after evidence:

- **Before this round**: the `isFinite` guard skipped the `BigDecimal` conversion for non-finite values but left `min`/`max` un-normalized, so `comparableValue.compareTo(min)` dispatched to `Double`'s bridge `compareTo(Object)` with an `Integer`/`Long` argument whenever the bound lacked a decimal point -> `ClassCastException`, uncaught, job killed.
- **After this round**: the non-finite branch now also converts `min`/`max` to `Double` when they're `Number` instances, so the bridge method always receives a `Double` -> no exception, row correctly routed to the error table. Verified both by manual trace and by the fork's actual CI log output showing the error-routing behavior being honored for exactly the previously-crashing scenario.

The `isFinite` guard's explanatory comment was also corrected this round — it previously overclaimed correctness for a case that wasn't actually true in general. The new comment accurately states the two-branch strategy and why each is safe. Good that it was fixed alongside the logic rather than left behind.

**Carried over, still open, both already non-blocking:**

A numeric value against a `String` bound (mistyped `min_value` like `"1,000"`), or a non-numeric value against a numeric bound, still throws the original `ClassCastException`. Untouched by this commit. Recommend validating bound/value type compatibility at config-parse time (preferred long-term), or adding a defensive fallback that turns an incomparable pair into a validation failure rather than a crash. At minimum, note the gap in the PR description. Severity: Medium, non-blocking.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Spotless-clean. The corrected comment is now accurate and states the *why* for both branches rather than restating the code — this is exactly the standard I look for on non-obvious compatibility logic, and it's a genuine improvement over last round's comment.

One new test, and I traced it rather than trusting the name — it constructs `min_value`/`max_value` as integer literals (parsed to `Integer` via `parseComparable`), covers both `NaN` and `+Infinity`, and asserts the row is routed to the error table rather than the call throwing. This is exactly the combination the previous round's test suite didn't reach, and it's the one that would have caught the gap I raised. I confirmed via the fork's CI log that this test actually executed against the fix and passed with the expected "exceeds maximum" routing message.

Deterministic: no `Thread.sleep`, no shared mutable state, fresh config/table/transform/row per test, consistent with the existing suite's style. **Stability rating: Stable.**

Full suite result from the fork's actual run against this head (`unit-test`, windows-latest): all 10 tests in `DataValidatorTransformTest` pass, 0 failures, 0 errors. No test weakening anywhere across either round.

## 2.3 Documentation Updates

No documentation change required or made; unchanged from both prior rounds.

# 3. Architectural Soundness

Still a targeted fix at the comparison site rather than the config-parse root cause, which I've consistently assessed as a defensible tradeoff for a hotfix. This round's refinement is a clean generalization of the same idea — "make both operands the same boxed type before calling `compareTo`" — applied consistently to both the finite (`BigDecimal`) and non-finite (`Double`) cases.

## 3.2 Maintainability

Improved from last round: the corrected comment now matches the actual behavior.

## 3.3 Extensibility

Unchanged — no new abstraction, rule registry untouched, no other rule implementation affected.

Unchanged conclusion from both prior rounds: no config contract, serialization, or checkpoint/state change. `RangeValidationRule` has no new fields, Java-serializes and Jackson-deserializes identically. An in-place upgrade is safe.

# 4. Issue Summary

| --- | --- | --- | --- | --- |
| 1 | Numeric value vs. `String` bound, or non-numeric value vs. numeric bound, still throws `ClassCastException` | `RangeValidationRule.java` (comparison dispatch); root cause `DataValidatorTransformConfig.java` (`parseComparable`) | Medium | Open, non-blocking |

No new issues found in this round's changes. My prior High-severity blocker is fully resolved.

# 5. Merge Recommendation

**Blockers — none.** My blocking finding from the last review (non-finite `Double`/`Float` values still crashing the job when compared against `Integer`/`Long`-typed bounds) is fully closed. I verified this three ways: manual trace of the new branch showing both operands converge on `Double` before `compareTo` is ever called, confirmation that no remaining type mismatch is possible for any `Number`-vs-`Number` pairing, and pulling the real fork CI log for this exact head confirming the new test executed and produced the expected "routed to error table" outcome rather than a stack trace.

**Recommended fixes — non-blocking:**
- Issue 1: numeric-vs-`String` and non-numeric-vs-numeric bound mismatches remain unfixed; worth a sentence in the PR description or a follow-up issue.
- The same-type fast path / bound-caching idea from my original review is still open, worth a follow-up if this rule sits on a hot path for high-throughput jobs.

**CI status.** I checked the contributor's fork run for this exact head directly rather than relying on the apache-side pointer. One unit-test job showed an unrelated failure in `CoordinatorServicePipelineCleanupTest` with `IllegalStateException: Node failed to start!` — this matches a known environmental Hazelcast-on-Windows flake I've seen recur across unrelated PRs, not anything caused by this change; within that same job's log, `DataValidatorTransformTest` itself shows 10/10 passing. A separate IT job failed on a Postgres CDC Testcontainers timeout — also unrelated, since this PR touches only `seatunnel-transforms-v2`. I'd suggest re-running just those two flaky jobs before merge, but neither failure is a source-level concern raised by this diff.

**Overall assessment.** This is exactly the kind of fast, responsive hotfix iteration I'd want to see — the fix precisely targets the gap I raised, the comment that previously overclaimed correctness was fixed alongside the code rather than left stale, and the new test exercises the specific failure mode that the previous round's tests didn't reach. I'm glad to move to approval now that the fix is verified end-to-end; the remaining CI failures look like unrelated environmental flakes rather than a real blocker.

DanielLeens (approved):
Thanks for turning this around so fast across the last two rounds — a fix, a corrected comment, and a targeted test all landed within hours of each review. This round doesn't need a delta-only pass either: the new head only adds a merge commit from `dev`, so I re-verified the merged result end to end (production file, test file, and CI on the actual merge commit) rather than assuming the merge was a no-op.

A `DataValidator` transform with a `RANGE` rule on a `DECIMAL` column died with `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.math.BigDecimal` (#11804), because `SeaTunnelRow` delivers `DECIMAL` values as `BigDecimal` while `DataValidatorTransformConfig.parseComparable` only ever parses `min_value`/`max_value` into `Integer`, `Long`, `Double`, or `String` — never `BigDecimal` — and `Comparable.compareTo` between mismatched boxed types throws. A follow-up round closed a related gap: non-finite `Double`/`Float` values (`NaN`/`Infinity`) also threw `ClassCastException` when compared against `Integer`/`Long`-typed bounds, because the two operands still ended up as different boxed types.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Reviewed head: `653348504e597e5d8e23f139ef40520a95b8c3d7` ("Merge branch 'dev' into fix-datavalidator-range-decimal").

This commit is purely a merge of `dev` into the PR branch — it carries no new commits from the author. I confirmed this via `gh api repos/apache/seatunnel/pulls/11805/files`, which reports the cumulative `dev...head` diff as still exactly two files, byte-identical to what I approved last round:

```
seatunnel-transforms-v2/.../rule/RangeValidationRule.java   | +58 -4
seatunnel-transforms-v2/.../DataValidatorTransformTest.java | +137 -0
```

I diffed the patch content itself against the version I traced in my previous approval rather than trusting the file-count match alone. The comparison logic is unchanged:

```java
if (value instanceof Number) {
    Number numberValue = (Number) value;
    if (isFinite(numberValue)) {
        comparableValue = new BigDecimal(numberValue.toString());
        if (min instanceof Number) { min = new BigDecimal(min.toString()); }
        if (max instanceof Number) { max = new BigDecimal(max.toString()); }
    } else {
        comparableValue = numberValue.doubleValue();
        if (min instanceof Number) { min = ((Number) min).doubleValue(); }
        if (max instanceof Number) { max = ((Number) max).doubleValue(); }
    }
}
```

Walking every numeric type combination the rule can encounter, as I did last round — nothing has changed, so no re-litigation is needed on the logic itself:

| value type | bound type | outcome |
| --- | --- | --- |
| `BigDecimal` (finite) | `Integer`/`Long`/`Double` | `BigDecimal` path — correct |
| `Integer`/`Long`/`Double`/`Float` (finite) | any numeric | `BigDecimal` path — correct |
| `Double`/`Float`, non-finite | `Integer`/`Long` bound | both converge to `Double` — correct |
| `Double`/`Float`, non-finite | `Double` bound | `Double.compareTo(Double)` — correct |
| numeric value | `String` bound (e.g. mistyped `"1,000"`) | still `ClassCastException` — carried-over, non-blocking (see Issue 1) |
| non-numeric value (`String`/date) | numeric bound | still `ClassCastException` — same carried-over issue |

`FieldValidator.validate` and `DataValidatorTransform.transformRow` remain untouched by this PR (no `try`/`catch` around rule evaluation, as before) — the merge from `dev` did not touch either file, confirmed by the file list above.

## 1.2 Compatibility Impact

Unchanged from last round's conclusion: fully compatible with previously-working configurations, and the non-finite-value regression I raised across the two middle rounds is fixed for both bound-authoring styles. No config option added, renamed, or re-defaulted; no serialization change — `RangeValidationRule` still has no new fields. No `incompatible-changes.md` entry needed.

No change from last round. The non-finite branch stays on a cold path. The same-type fast-path / bound-caching idea from my original review remains open as a non-blocking follow-up if this rule sits on a hot path for high-throughput jobs.

No open issues at High or blocking severity. One carried-over, non-blocking item:

### Issue 1 (carryover, unchanged, still open)

- **Location**: `RangeValidationRule.java` (comparison dispatch); root cause `DataValidatorTransformConfig.java` (`parseComparable`).
- **Problem description**: A numeric value against a `String` bound (e.g. a mistyped `min_value` like `"1,000"`), or a non-numeric value (`String`/date) against a numeric bound, still throws the original `ClassCastException`. Untouched by any commit in this PR, including this round's merge.
- **Potential risk**: Users hitting this variant get an opaque `ClassCastException` and a dead job. Medium — narrower and less common than the decimal/non-finite cases this PR targets, since it requires either a malformed bound string or a field/bound type mismatch that has nothing to do with `DECIMAL` handling.
- **Best improvement**: Validate bound/value type compatibility at config-parse time (preferred long-term, in `DataValidatorTransform.initializeFieldValidators()` where the field's `SeaTunnelDataType` is known), or add a defensive fallback in `validate()` that turns a genuinely incomparable pair into `ValidationResult.failure(...)` instead of a raw exception. Worth a follow-up issue rather than blocking this hotfix.
- **Severity**: Medium

# 2. Code Quality Assessment

## 2.1 Coding Standards

Spotless-clean, unchanged from last round. The `isFinite` guard's comment is accurate and states the two-branch strategy and why each is safe.

Same five `RANGE`-specific tests as last round, unchanged: `rangeRuleShouldAcceptDecimalFieldWithIntegerBounds`, `rangeRuleShouldRouteDecimalAboveMaxToErrorTable`, `rangeRuleShouldRejectDecimalEqualToMaxWhenMaxNotInclusive`, `rangeRuleShouldCompareDecimalFieldAgainstDoubleBounds`, `rangeRuleShouldRouteNonFiniteDoubleToErrorTableInsteadOfThrowing`, `rangeRuleShouldRouteNonFiniteDoubleToErrorTableWithIntegerBounds`. Together they cover: the original reported decimal-vs-integer-bound scenario, out-of-range rejection (not just the pass case), an exclusive-bound boundary case, `Double`-typed bounds against a `BigDecimal` field, and both `NaN`/`+Infinity` against both `Double`-typed and `Integer`-typed bounds — which is exactly the combination that would have caught the gap from my second review round.

I did not just diff the test source — I pulled the actual `unit-test` job log from the fork's CI run against this exact merge head (`653348504e5`, run `32095808267`) and confirmed:

```
[INFO] Running org.apache.seatunnel.transform.validator.DataValidatorTransformTest
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.018 s
```

on all four `unit-test` job configurations (`ubuntu-latest`/`windows-latest` x Java 8/11) — all four completed with `conclusion: success`. **Stability rating: Stable.** No test weakening anywhere across any round: nothing removed, relaxed, or made more permissive.

## 2.3 Documentation Updates

No documentation change required or made; unchanged from all prior rounds.

# 3. Architectural Soundness

Unchanged assessment: a targeted, defensible fix at the comparison site rather than the config-parse root cause. It's a general type-safe comparison strategy for the numeric domain — "make both operands the same boxed type before calling `compareTo`" — applied consistently across finite (`BigDecimal`) and non-finite (`Double`) numeric cases, not a decimal-only patch. It does not extend to the numeric-vs-`String`/non-numeric-vs-numeric combinations in Issue 1, which remain a known, documented gap (the PR description's "Known limitation" section already states this).

## 3.2 Maintainability

Unchanged — the explanatory comment accurately matches the current behavior for both branches.

## 3.3 Extensibility

Unchanged — no new abstraction, `ValidationRule`/`@JsonSubTypes` registry untouched, no other rule implementation affected.

Unchanged conclusion from every prior round: no config contract, serialization, or checkpoint/state change. `RangeValidationRule` has no new fields, Java-serializes and Jackson-deserializes identically. An in-place upgrade is safe.

# 4. Issue Summary

| --- | --- | --- | --- | --- |
| 1 | Numeric value vs. `String` bound, or non-numeric value vs. numeric bound, still throws `ClassCastException` | `RangeValidationRule.java` (comparison dispatch); root cause `DataValidatorTransformConfig.java` (`parseComparable`) | Medium | Open, non-blocking, documented as a known limitation in the PR description |

No new issues found in this round. Every blocker I have raised across the three prior rounds (the original CCE, the `NumberFormatException` regression, and the partial `ClassCastException` fix for non-finite values against integer-typed bounds) is fully resolved.

# 5. Merge Recommendation

**Blockers — none.**

1. **Recommended fixes — non-blocking**
   1. Issue 1 — numeric-vs-`String` and non-numeric-vs-numeric bound mismatches remain unfixed; already noted as a known limitation in the PR description, which is sufficient for this hotfix's scope. A follow-up issue for config-parse-time type validation would close it properly.
   2. The same-type fast-path / bound-caching idea from my original review is still open, worth a follow-up if this rule sits on a hot path for high-throughput jobs.

**CI status.** The apache-side `Build` check is still `IN_PROGRESS` (it is a pointer to the fork's run, per usual). I checked the contributor's fork run for this exact merge head directly: run `32095808267` against `653348504e5`. All four `unit-test` jobs (the ones that execute `DataValidatorTransformTest`) and `transform-v2-it-part-1` — the jobs actually relevant to this change — completed with `conclusion: success`. One unrelated job, `all-connectors-it-6 (8, ubuntu-latest)`, failed on `CouchbaseIT` with a container-startup error; that is a known environmental flake on `ubuntu-latest` unrelated to this PR (this PR touches only `seatunnel-transforms-v2`, and Couchbase's own bootstrap issue is independently tracked). Three broad, unrelated IT jobs (`all-connectors-it-1` x2, `kudu-connector-it`) were still in progress at the time of this review and are not expected to touch anything this PR changed. I'd suggest letting the full run finish (or re-running just the Couchbase flake) before the actual merge click, but none of that is a source-level concern raised by this diff.

**Overall assessment.** This round required a genuine re-verification rather than a rubber-stamp, precisely because "just a merge commit" is exactly the kind of change that could silently reintroduce a conflict-resolution mistake — it didn't. The merged file content is byte-identical to the version I already traced and approved, the full regression test suite still passes 10/10 on the fork's CI for this exact head, and the one CI failure present is a well-documented, unrelated environmental flake. Every blocking issue raised across this PR's three prior review rounds is closed. Happy to approve.

PR #11817, opened 2026-08-15
[Improve][Connector-V2] Migrate email validation to declarative OptionRule

abolfazlmadanii:

Part of #11007.

This PR completes the Email sink declarative option rule by:

- declaring the SMTP port, attachment name, and field delimiter options;
- validating the complete `email_smtp_port` range (`1..65535`) with declarative `Conditions`;
- failing fast during option validation instead of failing later when the writer sends the email;
- adding focused tests for valid boundaries, defaults, explicit optional options, invalid ports, and the CLI-facing unknown-key validation path;
- documenting the SMTP port constraint in English and Chinese.

### Scope clarification

The SMTP port constraint is new fail-fast validation rather than a migration of an existing imperative check. Invalid port values previously reached the writer and failed later during email delivery. Maintainer confirmation that this validation should remain in this PR has been requested in the review discussion.

Yes. `email_smtp_port` values outside the valid TCP port range (`1..65535`) are now rejected during option validation. The default value and valid configurations remain unchanged.

- `./mvnw -pl seatunnel-connectors-v2/connector-email -DskipITs test`
- `./mvnw spotless:apply`

The connector-email module passed all 6 tests with no failures or errors, including 4 focused `EmailFactoryTest` cases and the 2 existing writer tests. The factory-test helper runs `validateUnknownKeys()` before normal option validation, directly protecting the option declarations used by the CLI validation path. The full `./mvnw test` command was also attempted earlier, but the reactor stopped in the unrelated `seatunnel-config-shade` module because shaded `ConfigFactory` classes were unavailable in its test classpath.

### Check list

* [x] English and Chinese Email connector documentation is updated.
* [x] No incompatible-change entry is required; valid configurations and defaults are unchanged.
* [x] Existing connector packaging, mapping, labels, and E2E wiring are unchanged.

### DIFF
--- a/docs/en/connectors/sink/Email.md
+++ b/docs/en/connectors/sink/Email.md
@@ -64,7 +64,7 @@ password. When set to `false`, the connector sends mail over plain SMTP without 
 
 ### email_smtp_port [int]
 
-SMTP server port. The value must be greater than `0`. The default `465` is the SMTPS port and is used together with
+SMTP server port. The value must be between `1` and `65535`, inclusive. The default `465` is the SMTPS port and is used together with
 `email_smtp_auth = true`. For plain SMTP without authentication, set the port that matches the
 server (for example `25` or `3025`).
 
--- a/docs/en/connectors/sink/Email.md
+++ b/docs/en/connectors/sink/Email.md
@@ -64,7 +64,7 @@ password. When set to `false`, the connector sends mail over plain SMTP without 
 
 ### email_smtp_port [int]
 
-SMTP server port. The default `465` is the SMTPS port and is used together with
+SMTP server port. The value must be greater than `0`. The default `465` is the SMTPS port and is used together with
 `email_smtp_auth = true`. For plain SMTP without authentication, set the port that matches the
 server (for example `25` or `3025`).
 
--- a/docs/zh/connectors/sink/Email.md
+++ b/docs/zh/connectors/sink/Email.md
@@ -62,7 +62,7 @@ import ChangeLog from '../changelog/connector-email.md';
 
 ### email_smtp_port [int]
 
-SMTP 服务器端口。默认值 `465` 为 SMTPS 端口，需与 `email_smtp_auth = true` 配合使用。如果使用不带认证的普通 SMTP，请填写与服务匹配的端口（例如 `25` 或 `3025`）。
+SMTP 服务器端口，取值必须大于 `0`。默认值 `465` 为 SMTPS 端口，需与 `email_smtp_auth = true` 配合使用。如果使用不带认证的普通 SMTP，请填写与服务匹配的端口（例如 `25` 或 `3025`）。
 
 ### email_authorization_code [string]
 
--- a/docs/zh/connectors/sink/Email.md
+++ b/docs/zh/connectors/sink/Email.md
@@ -62,7 +62,7 @@ import ChangeLog from '../changelog/connector-email.md';
 
 ### email_smtp_port [int]
 
-SMTP 服务器端口，取值必须大于 `0`。默认值 `465` 为 SMTPS 端口，需与 `email_smtp_auth = true` 配合使用。如果使用不带认证的普通 SMTP，请填写与服务匹配的端口（例如 `25` 或 `3025`）。
+SMTP 服务器端口，取值必须在 `1` 到 `65535` 之间（包含边界值）。默认值 `465` 为 SMTPS 端口，需与 `email_smtp_auth = true` 配合使用。如果使用不带认证的普通 SMTP，请填写与服务匹配的端口（例如 `25` 或 `3025`）。
 
 ### email_authorization_code [string]
 
--- a/seatunnel-connectors-v2/connector-email/src/main/java/org/apache/seatunnel/connectors/seatunnel/email/sink/EmailSinkFactory.java
+++ b/seatunnel-connectors-v2/connector-email/src/main/java/org/apache/seatunnel/connectors/seatunnel/email/sink/EmailSinkFactory.java
@@ -27,12 +27,16 @@ import org.apache.seatunnel.api.table.factory.TableSinkFactoryContext;
 
 import com.google.auto.service.AutoService;
 
+import static org.apache.seatunnel.api.configuration.util.Conditions.greaterThan;
+import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_ATTACHMENT_NAME;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_AUTHORIZATION_CODE;
+import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_FIELD_DELIMITER;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_FROM_ADDRESS;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_HOST;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_MESSAGE_CONTENT;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_MESSAGE_HEADLINE;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_SMTP_AUTH;
+import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_SMTP_PORT;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_TO_ADDRESS;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_TRANSPORT_PROTOCOL;
 
@@ -61,7 +65,11 @@ public class EmailSinkFactory implements TableSinkFactory {
                         EMAIL_AUTHORIZATION_CODE,
                         EMAIL_MESSAGE_HEADLINE,
                         EMAIL_MESSAGE_CONTENT)
-                .optional(SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA)
+                .optional(EMAIL_SMTP_PORT, greaterThan(EMAIL_SMTP_PORT, 0))
+                .optional(
+                        EMAIL_ATTACHMENT_NAME,
+                        EMAIL_FIELD_DELIMITER,
+                        SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA)
                 .build();
     }
 }
--- a/seatunnel-connectors-v2/connector-email/src/main/java/org/apache/seatunnel/connectors/seatunnel/email/sink/EmailSinkFactory.java
+++ b/seatunnel-connectors-v2/connector-email/src/main/java/org/apache/seatunnel/connectors/seatunnel/email/sink/EmailSinkFactory.java
@@ -28,6 +28,7 @@ import org.apache.seatunnel.api.table.factory.TableSinkFactoryContext;
 import com.google.auto.service.AutoService;
 
 import static org.apache.seatunnel.api.configuration.util.Conditions.greaterThan;
+import static org.apache.seatunnel.api.configuration.util.Conditions.lessOrEqual;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_ATTACHMENT_NAME;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_AUTHORIZATION_CODE;
 import static org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions.EMAIL_FIELD_DELIMITER;
@@ -65,7 +66,10 @@ public class EmailSinkFactory implements TableSinkFactory {
                         EMAIL_AUTHORIZATION_CODE,
                         EMAIL_MESSAGE_HEADLINE,
                         EMAIL_MESSAGE_CONTENT)
-                .optional(EMAIL_SMTP_PORT, greaterThan(EMAIL_SMTP_PORT, 0))
+                // Fail fast during option validation instead of when the writer sends the email.
+                .optional(
+                        EMAIL_SMTP_PORT,
+                        greaterThan(EMAIL_SMTP_PORT, 0).and(lessOrEqual(EMAIL_SMTP_PORT, 65535)))
                 .optional(
                         EMAIL_ATTACHMENT_NAME,
                         EMAIL_FIELD_DELIMITER,
--- a/seatunnel-connectors-v2/connector-email/src/test/java/org/apache/seatunnel/connectors/seatunnel/email/EmailFactoryTest.java
+++ b/seatunnel-connectors-v2/connector-email/src/test/java/org/apache/seatunnel/connectors/seatunnel/email/EmailFactoryTest.java
@@ -75,7 +75,9 @@ public class EmailFactoryTest {
     }
 
     private void validate(Map<String, Object> config) {
-        ConfigValidator.of(ReadonlyConfig.fromMap(config)).validate(optionRule);
+        ReadonlyConfig readonlyConfig = ReadonlyConfig.fromMap(config);
+        ConfigValidator.validateUnknownKeys(readonlyConfig, optionRule, "EmailSink");
+        ConfigValidator.of(readonlyConfig).validate(optionRule);
     }
 
     private Map<String, Object> requiredConfig() {
--- a/seatunnel-connectors-v2/connector-email/src/test/java/org/apache/seatunnel/connectors/seatunnel/email/EmailFactoryTest.java
+++ b/seatunnel-connectors-v2/connector-email/src/test/java/org/apache/seatunnel/connectors/seatunnel/email/EmailFactoryTest.java
@@ -36,26 +36,41 @@ public class EmailFactoryTest {
 
     @Test
     void testValidSmtpPorts() {
-        Map<String, Object> config = validConfig();
+        Map<String, Object> config = requiredConfig();
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 1);
+        Assertions.assertDoesNotThrow(() -> validate(config));
+
         config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 465);
         Assertions.assertDoesNotThrow(() -> validate(config));
 
-        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 1);
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 65535);
         Assertions.assertDoesNotThrow(() -> validate(config));
     }
 
     @Test
-    void testDefaultSmtpPort() {
-        Assertions.assertDoesNotThrow(() -> validate(validConfig()));
+    void testDefaultOptionalOptions() {
+        Assertions.assertDoesNotThrow(() -> validate(requiredConfig()));
     }
 
     @Test
-    void testNonPositiveSmtpPorts() {
-        Map<String, Object> config = validConfig();
+    void testExplicitOptionalOptions() {
+        Map<String, Object> config = requiredConfig();
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 465);
+        config.put(EmailSinkOptions.EMAIL_ATTACHMENT_NAME.key(), "report.csv");
+        config.put(EmailSinkOptions.EMAIL_FIELD_DELIMITER.key(), "|");
+        Assertions.assertDoesNotThrow(() -> validate(config));
+    }
+
+    @Test
+    void testInvalidSmtpPorts() {
+        Map<String, Object> config = requiredConfig();
         config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 0);
         Assertions.assertThrows(OptionValidationException.class, () -> validate(config));
 
         config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), -1);
+        Assertions.assertThrows(OptionValidationException.class, () -> validate(config));
+
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 65536);
         Assertions.assertThrows(OptionValidationException.class, () -> validate(config));
     }
 
@@ -63,7 +78,7 @@ public class EmailFactoryTest {
         ConfigValidator.of(ReadonlyConfig.fromMap(config)).validate(optionRule);
     }
 
-    private Map<String, Object> validConfig() {
+    private Map<String, Object> requiredConfig() {
         Map<String, Object> config = new HashMap<>();
         config.put(EmailSinkOptions.EMAIL_FROM_ADDRESS.key(), "sender@example.com");
         config.put(EmailSinkOptions.EMAIL_TO_ADDRESS.key(), "receiver@example.com");
@@ -73,8 +88,6 @@ public class EmailFactoryTest {
         config.put(EmailSinkOptions.EMAIL_AUTHORIZATION_CODE.key(), "code");
         config.put(EmailSinkOptions.EMAIL_MESSAGE_HEADLINE.key(), "subject");
         config.put(EmailSinkOptions.EMAIL_MESSAGE_CONTENT.key(), "content");
-        config.put(EmailSinkOptions.EMAIL_ATTACHMENT_NAME.key(), "report.csv");
-        config.put(EmailSinkOptions.EMAIL_FIELD_DELIMITER.key(), "|");
         return config;
     }
 }
--- a/seatunnel-connectors-v2/connector-email/src/test/java/org/apache/seatunnel/connectors/seatunnel/email/EmailFactoryTest.java
+++ b/seatunnel-connectors-v2/connector-email/src/test/java/org/apache/seatunnel/connectors/seatunnel/email/EmailFactoryTest.java
@@ -17,15 +17,64 @@
 
 package org.apache.seatunnel.connectors.seatunnel.email;
 
+import org.apache.seatunnel.api.configuration.ReadonlyConfig;
+import org.apache.seatunnel.api.configuration.util.ConfigValidator;
+import org.apache.seatunnel.api.configuration.util.OptionRule;
+import org.apache.seatunnel.api.configuration.util.OptionValidationException;
+import org.apache.seatunnel.connectors.seatunnel.email.config.EmailSinkOptions;
 import org.apache.seatunnel.connectors.seatunnel.email.sink.EmailSinkFactory;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class EmailFactoryTest {
 
+    private final OptionRule optionRule = new EmailSinkFactory().optionRule();
+
     @Test
-    void optionRule() {
-        Assertions.assertNotNull((new EmailSinkFactory()).optionRule());
+    void testValidSmtpPorts() {
+        Map<String, Object> config = validConfig();
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 465);
+        Assertions.assertDoesNotThrow(() -> validate(config));
+
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 1);
+        Assertions.assertDoesNotThrow(() -> validate(config));
+    }
+
+    @Test
+    void testDefaultSmtpPort() {
+        Assertions.assertDoesNotThrow(() -> validate(validConfig()));
+    }
+
+    @Test
+    void testNonPositiveSmtpPorts() {
+        Map<String, Object> config = validConfig();
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), 0);
+        Assertions.assertThrows(OptionValidationException.class, () -> validate(config));
+
+        config.put(EmailSinkOptions.EMAIL_SMTP_PORT.key(), -1);
+        Assertions.assertThrows(OptionValidationException.class, () -> validate(config));
+    }
+
+    private void validate(Map<String, Object> config) {
+        ConfigValidator.of(ReadonlyConfig.fromMap(config)).validate(optionRule);
+    }
+
+    private Map<String, Object> validConfig() {
+        Map<String, Object> config = new HashMap<>();
+        config.put(EmailSinkOptions.EMAIL_FROM_ADDRESS.key(), "sender@example.com");
+        config.put(EmailSinkOptions.EMAIL_TO_ADDRESS.key(), "receiver@example.com");
+        config.put(EmailSinkOptions.EMAIL_HOST.key(), "smtp.example.com");
+        config.put(EmailSinkOptions.EMAIL_TRANSPORT_PROTOCOL.key(), "smtp");
+        config.put(EmailSinkOptions.EMAIL_SMTP_AUTH.key(), true);
+        config.put(EmailSinkOptions.EMAIL_AUTHORIZATION_CODE.key(), "code");
+        config.put(EmailSinkOptions.EMAIL_MESSAGE_HEADLINE.key(), "subject");
+        config.put(EmailSinkOptions.EMAIL_MESSAGE_CONTENT.key(), "content");
+        config.put(EmailSinkOptions.EMAIL_ATTACHMENT_NAME.key(), "report.csv");
+        config.put(EmailSinkOptions.EMAIL_FIELD_DELIMITER.key(), "|");
+        return config;
     }
 }


abolfazlmadanii:
After reviewing the connector again, I think a more conservative approach would be to first confirm the intended scope with the maintainers, or limit this PR to registering the missing options in `OptionRule`. The Email connector does not currently have an imperative SMTP-port validation to migrate, so the `email_smtp_port > 0` rule adds new validation behavior. Would you prefer that this PR keep the new constraint, or only add the missing options to `OptionRule`?

DanielLeens (changes requested):
Thanks for your first contribution to Apache SeaTunnel, and for the careful investigation — the "unknown option key" root-cause you found is a real, pre-existing bug, and the fix for it is exactly right. I read through the full change against the connector's config/factory/writer source (not just the diff) and left a couple of things below that I'd like a maintainer and you to settle before this merges, but nothing here suggests a problem with your approach.

- User pain point: The Email sink's `OptionRule` (`EmailSinkFactory#optionRule()`) never declared `email_smtp_port`, `email_attachment_name`, or `email_field_delimiter`, even though `EmailSinkOptions` and `EmailSinkConfig` have defined and consumed these options for a long time (`seatunnel-connectors-v2/connector-email/.../config/EmailSinkOptions.java`, `EmailSinkConfig.java`). Because `ConfigValidator.validateUnknownKeys(...)` (wired into `seatunnel-core-starter/.../ConfigValidationUtils.java` and `SeaTunnelConfValidateCommand.java`, i.e. the `--check` / `--dry-run=static|connect` CLI paths) treats any key not present in the declared `OptionRule` as an "unknown option key" error, any user who explicitly set `email_smtp_port`, `email_attachment_name`, or `email_field_delimiter` in their job config — even to a perfectly valid value — would have had that config **rejected by `--check`/dry-run validation**, despite the same config running fine at real job submission time (verified below). Separately, there was no validation anywhere (declarative or imperative) that `email_smtp_port` must be a positive TCP port; an invalid value (e.g. `0` or `-1`) would silently pass all the way through job submission and only fail deep inside `EmailSinkWriter.close()` (`seatunnel-connectors-v2/connector-email/.../sink/EmailSinkWriter.java:100`) when `Transport.send()` is invoked, i.e. after all rows have already been buffered/written to the local attachment file.
- Fix approach: Register the three previously-undeclared options in `EmailSinkFactory#optionRule()` as `optional(...)`, and attach a new `Conditions.greaterThan(EMAIL_SMTP_PORT, 0)` value constraint so an invalid port is now rejected eagerly by `ConfigValidator.validate(rule)` at job-submission/config-parse time (`FactoryUtil.createAndPrepareSink` → `MultipleTableJobConfigParser.java:775`), before the job actually runs. Docs (`docs/en` and `docs/zh` `connectors/sink/Email.md`) are updated to state the `> 0` constraint, and four focused unit tests are added.
- One-sentence summary: This PR both fixes a real, pre-existing "unknown option key" false-rejection bug for three legitimate Email-sink options in the CLI's static/dry-run validation path, and — as an explicitly separate, previously-nonexistent behavior — adds a brand-new fail-fast `email_smtp_port > 0` constraint at job-submission time.

# 1. Code Change Review
## 1.1 Core Logic Analysis

**Important scope clarification (verified against source, not assumption):** despite the PR title "Migrate email validation to declarative OptionRule," there was **no old imperative/regex validation of `email_smtp_port` to migrate**. I confirmed this three ways:
1. `EmailSinkWriter.java` (pre- and post-PR, unchanged by this PR) never validates `config.getEmailSmtpPort()`; it is passed straight into `properties.setProperty("mail.smtp.port", config.getEmailSmtpPort().toString())` at `EmailSinkWriter.java:100`.
2. The base-branch `EmailSinkFactory.java` (`git show b1e21631ffdfd8cb33e6adca618b2d95224a9d66:...EmailSinkFactory.java`) never listed `EMAIL_SMTP_PORT` in `optionRule()` at all — only `.optional(SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA)`.
3. The PR author states this explicitly in their own PR comment (2026-08-15T08:36:04Z): *"The Email connector does not currently have an imperative SMTP-port validation to migrate, so the `email_smtp_port > 0` rule adds new validation behavior. Would you prefer that this PR keep the new constraint, or only add the missing options to `OptionRule`?"* — this question is still open/unanswered as of this review.

So there is no "old regex vs. new regex" to diff for a value-constraint — the correct comparison is "no validation" (before) vs. "`greaterThan(EMAIL_SMTP_PORT, 0)`" (after):

Before (`EmailSinkFactory.java`, base branch, `b1e21631f`):
```java
.optional(SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA)
```
`EMAIL_SMTP_PORT`, `EMAIL_ATTACHMENT_NAME`, `EMAIL_FIELD_DELIMITER` are absent — undeclared, unvalidated, but still readable via `ReadonlyConfig.get(...)` with their defaults (465 / `emailsink.csv` / `,` respectively, defined in `EmailSinkOptions.java`, itself untouched by this PR).

After (`EmailSinkFactory.java:68-72`, head `848e46a72`):
```java
.optional(EMAIL_SMTP_PORT, greaterThan(EMAIL_SMTP_PORT, 0))
.optional(
        EMAIL_ATTACHMENT_NAME,
        EMAIL_FIELD_DELIMITER,
        SinkConnectorCommonOptions.MULTI_TABLE_SINK_REPLICA)
```
I verified `Conditions.greaterThan` and `ConditionEvaluators.GREATER_THAN` (`seatunnel-api/.../util/Conditions.java`, `ConditionEvaluators.java:64-66`) perform a correct numeric comparison (`compareNumbers(v, c.getExpectValue()) > 0`), so `email_smtp_port = 0` and negative values are correctly rejected, and any positive integer (1..Integer.MAX_VALUE) is accepted. I also verified `OptionRule.Builder#optional(Option, Condition, Condition...)` (`OptionRule.java:346`) is a real, existing overload — this is valid, already-used framework API, not a fabricated call.

**Required-vs-optional flip risk (explicitly checked, none found):** all three fields are declared `.optional(...)`, matching their pre-existing `defaultValue(...)` in `EmailSinkOptions.java` (465 / `emailsink.csv` / `,`) and matching the docs table, which already marked them `no` (not required) before this PR. No field flipped from optional to required or vice versa.

**Timing-of-validation change (the real behavioral delta, verified via call chain):**
- `ConfigValidator.of(context.getOptions()).validate(factory.optionRule())` is invoked from `FactoryUtil.createAndPrepareSink` (`seatunnel-api/.../table/factory/FactoryUtil.java:252`), which is called from the Zeta engine's real job-submission parser `MultipleTableJobConfigParser.java:775` — i.e. this runs on the client/master side while parsing the job graph, before any Source/Transform/Sink task thread starts. So an invalid `email_smtp_port` now fails the job **before submission**, with a clear `OptionValidationException`, instead of failing **during** the job (after all rows are buffered to the local attachment file) inside `EmailSinkWriter.close()` with a generic `EmailConnectorException(SEND_EMAIL_FAILED, ...)` wrapping whatever `javax.mail`/socket exception a port of 0/-1 produces. This is a genuine fail-fast improvement and sits on the **normal execution path** for every job that uses the Email sink with an explicit or default port (not an edge/recovery-only path).
- Separately, `ConfigValidator.validateUnknownKeys(...)` is **only** wired into the offline `--check` / `--dry-run=static|connect` CLI command path (`seatunnel-core-starter/.../ConfigValidationUtils.java:validateSources/validateTransforms/validateSinks`, and `seatunnel-starter/.../SeaTunnelConfValidateCommand.java:validatePluginConfigs`) — I confirmed it is **not** called from `MultipleTableJobConfigParser.java` (grep found zero occurrences), so real (non-`--check`) job submissions were never rejected by the old "unknown key" gap; only the separate static-validation/dry-run tooling was affected. This PR fixes that dry-run-only false-rejection for these three keys as a side effect.

## 1.2 Compatibility Impact
**Partially incompatible** — one explicit, narrow behavior change, everything else is a strict compatibility fix.
- Strictly compatibility-improving: registering `EMAIL_ATTACHMENT_NAME` and `EMAIL_FIELD_DELIMITER` (no value constraint attached) fixes real jobs that previously failed `--check`/dry-run validation purely because these already-functional, already-documented keys were undeclared.
- Genuinely new, narrower behavior: any job config that explicitly sets `email_smtp_port` to `0` or a negative number will now fail at job-submission time with `OptionValidationException` (`SeaTunnelAPIErrorCode.OPTION_VALIDATION_FAILED`), where before it would have been accepted at submission and only failed later, mid-job, inside `EmailSinkWriter.close()`. Practically, no legitimate job would intentionally use a non-positive SMTP port, so the real-world blast radius is very small, but it is nonetheless a config that used to be accepted (lazily, and ultimately unusably) and is now rejected earlier and more clearly. Per `docs/en/introduction/concepts/incompatible-changes.md` conventions this is a borderline case; the PR body's "No incompatible-change entry is required" claim is reasonable given the practical impossibility of a working job with port ≤ 0, but this is exactly the open question the author raised to maintainers and that has not yet been answered.

No measurable performance impact: `ConfigValidator.validate(rule)` already runs once per sink at job-parse time regardless of this PR (it's the standard `FactoryUtil` path used by every connector); this PR only adds one more `Condition` object to `valueConstraints` and three more `Option` references to `optionalOptions`, both O(1) additions evaluated once, not per-record or per-checkpoint. No new I/O, no new object allocation in any hot path (`EmailSinkWriter.write()`/`close()` are untouched).

Issue 1: PR scope/intent is not yet confirmed by a maintainer, and the PR's own title overstates what changed ("migrate" implies a like-for-like move of existing validation, but the port check is brand-new behavior, not a migration).
- Location: PR title / description; `seatunnel-connectors-v2/connector-email/src/main/java/org/apache/seatunnel/connectors/seatunnel/email/sink/EmailSinkFactory.java:68`
- Problem description: The author's own PR comment (2026-08-15) explicitly asks maintainers whether the new `email_smtp_port > 0` constraint should be kept or whether the PR should be narrowed to just registering the missing options. No maintainer has answered yet.
- Potential risk: Merging without resolving this leaves an unreviewed product decision (should Email sink port validation exist at all, and should it be `>0` only or a full `1..65535` range) baked into the codebase under a misleading "migration" label.
- Best improvement: A maintainer should explicitly confirm (in the PR thread) whether the new constraint is in scope before merge; if kept, consider retitling to reflect "add missing OptionRule declarations + new SMTP port validation" for an accurate changelog/git-blame trail.
- Severity: Medium
- Raised by another reviewer: No (the author raised the underlying scope question themselves in a self-comment; no other reviewer/maintainer has weighed in yet)

Issue 2: The new constraint only checks a lower bound (`> 0`), not the valid TCP port upper bound.
- Location: `seatunnel-connectors-v2/connector-email/src/main/java/org/apache/seatunnel/connectors/seatunnel/email/sink/EmailSinkFactory.java:68`
- Problem description: `email_smtp_port` is validated with `greaterThan(EMAIL_SMTP_PORT, 0)` only. A value like `70000` (outside the valid 1-65535 TCP port range) still passes config validation and will only fail later at `Transport.send()` time, i.e. the exact class of "fails deep inside `EmailSinkWriter.close()`" problem this PR is trying to move earlier is only half-solved. The framework already supports chaining, as shown in `Conditions.java`'s own Javadoc example: `greaterOrEqual(PORT, 1).and(lessOrEqual(PORT, 65535))`.
- Potential risk: Low — out-of-range ports are a rare typo case, and the failure still happens (just not as early as it could).
- Best improvement: Use `greaterThan(EMAIL_SMTP_PORT, 0).and(lessOrEqual(EMAIL_SMTP_PORT, 65535))` (or `greaterOrEqual(..., 1)`) to fully validate the port range up front, matching the pattern the framework's own documentation recommends.
- Severity: Low

Issue 3: No inline code comment explains why the new value constraint was added or why only a lower bound was chosen.
- Location: `seatunnel-connectors-v2/connector-email/src/main/java/org/apache/seatunnel/connectors/seatunnel/email/sink/EmailSinkFactory.java:68-72`
- Problem description: `optionRule()` gained a new, behavior-changing value constraint with zero explanatory comment in the source; the rationale exists only in the PR description, which is not visible to a future reader of the code/git-blame.
- Potential risk: Future maintainers editing this method (e.g., to add the upper bound from Issue 2) have no in-code context for why the constraint exists or what edge case it guards against.
- Best improvement: Add a short comment above line 68, e.g. `// SMTP port must be positive; validated eagerly here instead of failing inside EmailSinkWriter.close() at send time.`
- Severity: Low

# 2. Code Quality Assessment
## 2.1 Coding Standards
The diff is small and idiomatic: correct use of existing `OptionRule.Builder#optional(Option, Condition...)` and `Conditions.greaterThan` framework APIs, static imports follow existing file conventions, no wildcard imports, no shaded-dependency violations. `EmailSinkOptions.java` (which defines the three options with their descriptions/defaults) is untouched by this PR, so no new field lacks a `withDescription(...)`. The one gap is the missing inline comment on the new constraint noted as Issue 3 above (Medium severity per the project's "does every new core method / non-trivial field have an explanatory comment" check — I am rating it Low here since it is a one-line addition to an existing declarative builder chain rather than a new method or field, but a comment would still meaningfully help).

`EmailFactoryTest.java` replaces the previous single smoke test (`Assertions.assertNotNull((new EmailSinkFactory()).optionRule())`) with four targeted tests:
- `testValidSmtpPorts` (lines 37-45): asserts 465 and 1 both validate without throwing.
- `testDefaultSmtpPort` (lines 47-50): asserts a config that omits `email_smtp_port` (relying on the default `465`) validates without throwing.
- `testNonPositiveSmtpPorts` (lines 52-60): asserts 0 and -1 both throw `OptionValidationException`.

Stability rating: **Stable**. No `Thread.sleep`, no ordering dependency between tests (each rebuilds its own `Map` via `validConfig()`), no shared mutable static state (`optionRule` field is effectively immutable after construction and stateless to call), no floating-point comparisons, no external I/O or network access — these are pure, synchronous, deterministic unit tests against `ConfigValidator`.

Coverage gap (non-blocking, see Issue 4 in the summary table below): `validConfig()` (lines 66-79) always explicitly sets `email_attachment_name` and `email_field_delimiter`. Since these two options flipped from "undeclared" to "declared-optional" in this exact PR, there is no test proving that a config which *omits* them (relying on their defaults, `emailsink.csv` and `,`) still validates successfully — which is the precise regression class this PR is meant to fix. Recommend adding such a case.

## 2.3 Documentation Updates
Both `docs/en/connectors/sink/Email.md:66` and `docs/zh/connectors/sink/Email.md:64` are updated with a one-line addition stating the port must be `> 0`, consistent with the code change. `email_attachment_name` and `email_field_delimiter` were already fully documented (including in the config table's "required" column, already correctly marked "no") before this PR, so no further doc changes were needed for those two options — verified by inspecting the full `Email.md` content, not just the diff.

# 3. Architectural Soundness
## 3.1 Elegance of the Solution (classify: Precise fix / Temporary workaround / Long-term solution)
**Long-term solution** for the "undeclared options" half of the fix (registering `EMAIL_ATTACHMENT_NAME`/`EMAIL_FIELD_DELIMITER` in `optionRule()` is exactly the correct, permanent fix for the dry-run false-rejection bug). The `email_smtp_port > 0` constraint is a reasonable, idiomatic use of the existing `Conditions`/`OptionRule` framework, but as discussed in Issue 2 it is an incomplete range check, and as discussed in Issue 1 its inclusion in this particular PR is still an open scope question from the author to the maintainers.

## 3.2 Maintainability
Good — the change reduces, rather than increases, future maintenance burden: it closes the gap between what `EmailSinkOptions`/`EmailSinkConfig` actually consume and what `optionRule()` declares, which is exactly the kind of drift that causes confusing "unknown option" errors for users and support burden for maintainers.

## 3.3 Extensibility
Good — using `Conditions.greaterThan(...)` (rather than a custom imperative `if` check) means a future PR can trivially extend this to a full port-range check (Issue 2) or add similar declarative constraints to other numeric Email-sink options without touching runtime connector code.

No checkpoint/savepoint, serialization, or SPI-contract concerns — `EmailSinkConfig`/`EmailSinkWriter` (the classes that actually execute and would need to deserialize any persisted state) are entirely untouched by this PR. The only compatibility surface is config-time validation, covered in section 1.2.

# 4. Issue Summary

|---|---|---|---|
| 1 | New `email_smtp_port > 0` constraint is genuinely new behavior, not a migration; author's own open question to maintainers about PR scope is unanswered | `EmailSinkFactory.java:68`; PR title/description | Medium |
| 2 | Port validation only checks lower bound (`> 0`), not the valid `1-65535` TCP port range | `EmailSinkFactory.java:68` | Low |
| 3 | No inline comment explaining the new validation constraint's rationale | `EmailSinkFactory.java:68-72` | Low |
| 4 | No test verifies that omitting `email_attachment_name`/`email_field_delimiter` (relying on defaults) still validates, despite that being exactly the regression class this PR fixes | `EmailFactoryTest.java:66-79` | Low |
| 5 | CI has not actually executed: "Build" check shows `ACTION_REQUIRED` (first-time contributor gate) and the fork (`abolfazlmadanii/seatunnel`) shows zero workflow runs for this head SHA — only the author's self-reported local `mvnw test` run exists as evidence | GitHub Actions status for `848e46a722b6` | Medium |

# 5. Merge Recommendation
### Conclusion: "Ready to merge after fixes"

   - (Issue 5, Medium) CI must actually run and pass. As of this review, `apache/seatunnel`'s "Build" check is `ACTION_REQUIRED` and the contributor's fork has zero workflow runs at head `848e46a722b6` — a maintainer must click "Approve and run workflows" so CI evidence exists beyond the author's self-reported local test run.
   - (Issue 1, Medium) A maintainer should explicitly answer the author's own open question in the PR thread — confirm whether the new `email_smtp_port > 0` constraint is in scope for this PR, or whether it should be split out / the PR retitled to accurately reflect "adds new SMTP port validation" rather than "migrate ... validation."

   - (Issue 2, Low) Add an upper-bound check (`lessOrEqual(EMAIL_SMTP_PORT, 65535)`) so the full valid port range is validated up front, matching the pattern already documented in `Conditions.java`'s own Javadoc.
   - (Issue 3, Low) Add a short comment above the new constraint explaining why it exists and why validation was moved earlier (fail-fast vs. late failure in `EmailSinkWriter.close()`).
   - (Issue 4, Low) Add a test case confirming that a config which omits `email_attachment_name`/`email_field_delimiter` (relying on their existing defaults) still validates successfully now that they are declared `.optional(...)`.

Overall assessment: This is a small, well-scoped, technically correct change. I independently verified — by reading `EmailSinkFactory.java`, `EmailSinkOptions.java`, `EmailSinkConfig.java`, `EmailSinkWriter.java`, `Conditions.java`, `ConditionEvaluators.java`, `OptionRule.java`, `ConfigValidator.java`, `FactoryUtil.java`, `MultipleTableJobConfigParser.java`, `ConfigValidationUtils.java`, and `SeaTunnelConfValidateCommand.java` at both the base and head commits — that the code is a correct, idiomatic use of the existing `OptionRule`/`Conditions` framework, that no field's required/optional status was flipped, and that the change genuinely improves fail-fast behavior for the normal job-submission path while also fixing a real pre-existing "unknown option key" false-rejection in the `--check`/dry-run CLI path. The one substantive concern is not a code defect but a scope/intent question the author raised to maintainers themselves and that remains unanswered; combined with CI never having actually executed for this first-time contributor, I recommend resolving both before merge, after which this is an easy approve.

abolfazlmadanii:
Thanks for the detailed review and for verifying the behavior against the full source.
I agree that the current title is not fully accurate, since the SMTP port constraint introduces new validation rather than migrating an existing check.
Since the scope question is still open, I’ll hold off on changing or extending the port constraint until a maintainer confirms whether it should be included in this PR. If it is kept, I can update the title and description accordingly and address the suggested range validation, comment, and additional default-value test.
The main CI workflow currently requires maintainer approval before it can run.
Thanks again for the thorough review.

goutamadwant:
@abolfazlmadanii The option registration fix looks correct. I reproduced the original path with ConfigValidator.validateUnknownKeys(): without EMAIL_ATTACHMENT_NAME and EMAIL_FIELD_DELIMITER in OptionRule it fails with both reported as unknown keys, and with this change it passes.

One test gap: EmailFactoryTest currently calls only ConfigValidator.validate(), which does not check unknown keys. The existing 5 tests still pass even if those two option declarations are removed.

Could we update the test helper to run validateUnknownKeys() before validate()? That would directly cover the CLI validation failure this PR fixes and prevent the declarations from being removed accidentally later.

DanielLeens:
Thanks both for the follow-up — this is exactly the kind of back-and-forth that makes a PR better.

@abolfazlmadanii, holding off on extending the port-range validation until a maintainer confirms scope is the right call, and I agree the title should describe what the diff actually does (adds a new SMTP-port constraint plus registers the two missing options) rather than "migrate," since "migrate" implies a pre-existing check just moved location, and for `email_smtp_port` there wasn't one. That's a small, easy fix once scope is settled — no need to block on it separately.

@goutamadwant, good catch on the test gap, and I can confirm it from the same angle: `EmailFactoryTest` only exercises `ConfigValidator.validate()`, which checks declared constraints (`Conditions`, ranges, etc.) but not "does the schema even know about this key" — that's `validateUnknownKeys()`'s job, and it's wired into the real CLI path (`ConfigValidationUtils`/`SeaTunnelConfValidateCommand`) but not into this test. So today the test suite would stay green even if `EMAIL_ATTACHMENT_NAME`/`EMAIL_FIELD_DELIMITER` were removed from `optionRule()` again in some future change — which is precisely the class of regression this PR is fixing. Adding a `validateUnknownKeys()` call (or a small helper that runs both) to `EmailFactoryTest` would turn that from "fixed by inspection" into "protected by a test," and I'd treat that as effectively closing my own default-value-test suggestion from the original review too, since a well-placed `validateUnknownKeys()` assertion covering all three options is more direct than testing each option's default individually.

Current status, to keep this from going in circles: the open items are (1) the scope question on the SMTP-port range/comment work — waiting on a maintainer, (2) the title/description update once scope is settled, and (3) the `validateUnknownKeys()` test-coverage gap @goutamadwant raised. Nothing here changes my read that the core option-registration fix is correct and the root cause you found is real — happy to take another look as soon as you've had a chance to act on whichever of these lands first.

abolfazlmadanii:
Thanks for the detailed review. I pushed commit `e9af64892` addressing the recommended changes:

- validate the complete SMTP port range (`1..65535`);
- add an inline comment explaining the fail-fast validation;
- cover omitted optional options and their existing defaults;
- cover explicitly configured optional options;
- update the English and Chinese documentation to match the full port range.

Local verification passed:

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-email -DskipITs test` — 6 tests, 0 failures, 0 errors

I have also updated the PR description to clarify that the SMTP port constraint is new fail-fast validation rather than a migration of an existing imperative check. Invalid port values previously failed later during email delivery. Could a maintainer please confirm whether keeping this validation in the scope of this PR is acceptable?

@DanielLeens, could you please take another look at the updated changes?

abolfazlmadanii:
@goutamadwant Thanks for catching this. I addressed the unknown-key coverage gap in commit `188bd0c10`.

The shared `EmailFactoryTest` validation helper now calls `ConfigValidator.validateUnknownKeys()` before `ConfigValidator.validate()`. Since `testExplicitOptionalOptions` explicitly configures `email_smtp_port`, `email_attachment_name`, and `email_field_delimiter`, removing any of those declarations from `OptionRule` will now fail the test and directly protect the CLI validation path that this PR fixes.

Verification passed:

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-email -DskipITs test` — 6 tests, 0 failures, 0 errors

@DanielLeens, this also closes the remaining test-coverage item from your status summary. Could you please take another look when convenient?

DanielLeens (commented):
Thanks for pushing both follow-up commits, @abolfazlmadanii — and thanks @goutamadwant for the sharp catch on the unknown-key test gap. I re-read the entire diff against the current head (`188bd0c10`), not just the delta since my last comment, and re-traced the full validation call chain again (`EmailSinkFactory` → `ConfigValidator` → `FactoryUtil`/`ConfigValidationUtils`/`SeaTunnelConfValidateCommand`) to make sure nothing regressed along the way.

**Result: all four technical issues from my previous review are now resolved.**
1. Port range — now the full `1..65535` via `greaterThan(EMAIL_SMTP_PORT, 0).and(lessOrEqual(EMAIL_SMTP_PORT, 65535))` (`EmailSinkFactory.java:70-71`), exactly matching the pattern `Conditions.java`'s own Javadoc recommends. Verified `Condition#and(Condition<?>)` and `Conditions.lessOrEqual` both exist and are used correctly — this is real, already-shipped framework API, not new surface.
2. Inline comment added above the constraint (`EmailSinkFactory.java:69`): "Fail fast during option validation instead of when the writer sends the email." — accurately explains the *why*.
3. Test coverage for omitted optional options now exists (`testDefaultOptionalOptions`) alongside a new explicit-optional-options case (`testExplicitOptionalOptions`) and an expanded invalid-port case that now also covers the new upper bound (`0`, `-1`, `65536`).
4. The unknown-key gap @goutamadwant flagged is closed: the shared `validate()` helper in `EmailFactoryTest.java:79-81` now runs `ConfigValidator.validateUnknownKeys(readonlyConfig, optionRule, "EmailSink")` before `ConfigValidator.of(readonlyConfig).validate(optionRule)`. I checked this isn't a one-off — `KafkaFactoryTest.java` already uses the identical `validateUnknownKeys` + `validate` pattern, so this brings `EmailFactoryTest` in line with an established repo convention rather than inventing a new one. `"EmailSink"` also correctly matches `EmailSinkFactory#factoryIdentifier()`, so the connector-name argument is right, not a placeholder.

I diffed `EmailSinkFactory.java` in full (not just the hunk) to confirm the combined `.optional(EMAIL_SMTP_PORT, ...).optional(EMAIL_ATTACHMENT_NAME, EMAIL_FIELD_DELIMITER, MULTI_TABLE_SINK_REPLICA)` builder chain is well-formed and that `EMAIL_SMTP_PORT`'s existing default (`465`) and the other two options' defaults (`emailsink.csv`, `,`) still fall inside/satisfy the new constraints — no default-value regression. `EmailSinkOptions.java`, `EmailSinkConfig.java`, and `EmailSinkWriter.java` remain untouched by this PR, so the runtime send path is unaffected; only the earlier, declarative validation gate changed.

**On the open scope question:** I want to be straightforward about what I can and can't resolve here. I'm not a project maintainer with merge authority, so I can't personally sign off on "should net-new `email_smtp_port` validation ship inside a PR titled 'migrate'" — that's the product/process call you correctly routed to a maintainer, and it's still unanswered as of this comment. What I *can* confirm: the PR body now has an explicit "Scope clarification" section stating this is new fail-fast validation, not a migration of an existing check, and the "Does this PR introduce any user-facing change?" checklist item is now accurately filled in ("Yes... rejected during option validation... default value and valid configurations remain unchanged") rather than left blank or marked "No." That's a real improvement in disclosure regardless of how the scope question is ultimately settled. The one loose end is that the PR *title* itself still reads "Migrate email validation to declarative OptionRule" — worth a quick rename to something like "Add SMTP port range validation and register missing Email sink options" once scope is confirmed, so the merged title/changelog doesn't overstate what changed. Non-blocking, easy to fix in the same pass as the scope answer.

Two things, now clearly disclosed as distinct in the PR body: (1) a genuine pre-existing bug — `EMAIL_SMTP_PORT`, `EMAIL_ATTACHMENT_NAME`, and `EMAIL_FIELD_DELIMITER` were consumed by `EmailSinkConfig`/`EmailSinkWriter` but never declared in `EmailSinkFactory#optionRule()`, so any job explicitly setting them failed CLI `--check`/dry-run's `validateUnknownKeys()` path even though the same config ran fine at real job submission; and (2) new, previously-nonexistent fail-fast validation that `email_smtp_port` must be a valid TCP port (`1..65535`), moving the failure from deep inside `EmailSinkWriter.close()`'s `Transport.send()` call (after rows are already buffered to the local attachment file) to job-submission time.

# 1. Code Change Review
## 1.1 Core Logic Analysis
`EmailSinkFactory.java:68-72` now declares all three previously-missing options as `.optional(...)`, with `EMAIL_SMTP_PORT` additionally carrying a value constraint validated via `ConfigValidator.of(...).validate(rule)`, which is invoked from `FactoryUtil.createAndPrepareSink` → `MultipleTableJobConfigParser.java:775` on the normal job-submission path (not an edge/recovery-only path) for every job that configures the Email sink. The range check (`> 0 AND <= 65535`) is now complete and matches the full valid TCP port space; no field's required/optional status changed; no default value changed. I re-verified `EmailSinkOptions.java` is untouched — `EMAIL_SMTP_PORT` still defaults to `465`, `EMAIL_ATTACHMENT_NAME` to `emailsink.csv`, `EMAIL_FIELD_DELIMITER` to `,` — all comfortably inside/compatible with the new constraints, so no existing job relying on defaults can regress.

## 1.2 Compatibility Impact
**Partially incompatible**, same classification as before but now properly disclosed: any job that explicitly configured `email_smtp_port` to `0`, a negative value, or a value `> 65535` will now be rejected at job-submission time with `OptionValidationException`, where previously it was accepted and only failed later, mid-job, inside the writer. No legitimate job could have depended on such a value working, so real-world blast radius stays small — and unlike the initial version of this PR, the body's "user-facing change" checklist item now says so explicitly rather than needing to be inferred. Everything else (`EMAIL_ATTACHMENT_NAME`, `EMAIL_FIELD_DELIMITER` registration, and the CLI-side `validateUnknownKeys` fix) is a strict compatibility improvement — it makes previously-rejected, perfectly valid configs pass again.

No change since the last review: the added `Condition` and two extra `Option` references are evaluated once per sink at job-parse time, O(1), no new I/O or per-record/per-checkpoint cost. `EmailSinkWriter.write()`/`close()` are untouched.

`OptionValidationException` thrown by `Conditions`/`ConfigValidator` carries the option name and constraint, and `validateUnknownKeys` (existing framework method, unchanged by this PR) reports the full list of unknown keys plus the declared-options set — both are actionable, specific failure messages rather than generic errors. The new inline comment on the constraint gives future readers the "why" without needing to dig into the PR history.

# 2. Code Quality Assessment
## 2.1 Coding Standards
Clean, idiomatic use of existing `OptionRule.Builder`/`Conditions` framework APIs; static imports follow file convention; no wildcards; no shaded-dependency issues. The `EmailFactoryTest` helper factoring (`validate()`/`requiredConfig()`) is a sensible reduction of duplication across the five test methods.

Five tests, all pure/synchronous/deterministic against `ConfigValidator` — no sleeps, no shared mutable state, no I/O:
- `testValidSmtpPorts` — boundaries `1`, `465`, `65535` all pass.
- `testDefaultOptionalOptions` — config that omits all three newly-declared optional options validates via defaults (closes the gap I flagged as Issue 4 last review).
- `testExplicitOptionalOptions` — explicitly setting all three passes.
- `testInvalidSmtpPorts` — `0`, `-1`, and now `65536` all throw `OptionValidationException` (upper bound now covered, matching the widened range check).
- The shared `validate()` helper runs `validateUnknownKeys()` before `validate()`, directly protecting against the exact regression class this PR fixes (declarations silently dropped from `optionRule()` in some future change).

**Stability rating: Stable.** Deterministic map-based config construction, no ordering dependency between tests, no external resources.

## 2.3 Documentation Updates
`docs/en/connectors/sink/Email.md:66` and `docs/zh/connectors/sink/Email.md:64` both updated to state the `1..65535` range (matches the widened code constraint, not just the original lower-bound-only version). Both language docs stay in sync.

# 3. Architectural Soundness
Long-term solution. Declarative `Conditions` chaining (`greaterThan(...).and(lessOrEqual(...))`) is the idiomatic, extensible way to express this constraint within the existing `OptionRule` framework — no imperative validation code added to the connector itself.

## 3.2 Maintainability
Good, and improved further this round: the test suite now actively guards against the exact "declaration silently removed" regression class that caused the original bug, via `validateUnknownKeys()`.

## 3.3 Extensibility
Good — same reasoning as before; a future numeric option can reuse this exact chaining pattern.

No checkpoint/savepoint, serialization, or SPI-contract concerns. `EmailSinkConfig`/`EmailSinkWriter` (the classes that execute and would need to deserialize persisted state) remain entirely untouched. Compatibility surface is limited to config-time validation, covered in 1.2.

# 4. Issue Summary

| Number | Issue | Location | Severity | Raised by another reviewer |
|---|---|---|---|---|
| 1 | PR title still says "migrate" though the SMTP-port check is net-new validation, not a migration; body now discloses this accurately but the title itself is unchanged | PR title | Low (was Medium; largely mitigated by the new "Scope clarification" section in the body) | No |
| 2 | Scope question (should new `email_smtp_port` validation ship in this PR) is still open pending maintainer confirmation | PR body "Scope clarification" section | Medium (process gate, not a code defect) | No |
| 3 | CI has not actually executed: `Build` check is `action_required` and the fork shows zero workflow runs at head `188bd0c10` — only self-reported local test output exists as evidence | `apache/seatunnel` check-runs for `188bd0c10`; `abolfazlmadanii/seatunnel` Actions (zero runs found) | Medium | No |

All four previously-open code-level issues (port upper bound, missing comment, default-value test gap, unknown-key test gap) are resolved as of `188bd0c10` and are not re-listed above.

# 5. Merge Recommendation
   - (Issue 2) A maintainer needs to explicitly confirm the `email_smtp_port` range-validation is in scope for this PR (or ask for it to be split out). This is the one substantive open item and it's a scope/process call, not something further code changes can resolve.
   - (Issue 3) CI needs to actually run. The `Build` check is `action_required` (first-time-contributor gate) and the fork has zero recorded workflow runs at this head SHA — a maintainer needs to approve running workflows so there's real CI evidence beyond local `mvnw test` output.
   - (Issue 1) Once scope is confirmed, consider retitling the PR to reflect what it actually does (adds SMTP-port range validation + registers two missing options), since "migrate" implies a like-for-like relocation that didn't happen here.

Technically, this is now a clean, well-tested, correctly-scoped-and-disclosed change — my compliments on how thoroughly the last round of feedback was addressed, including going out of your way to align with the existing `KafkaFactoryTest` convention for the unknown-key check. The only things standing between this and merge are a maintainer's scope sign-off and CI actually getting to run.

davidzollo:
Good job.

+1 if CI passes.
LGTM


To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration  https://github.com/apache/seatunnel/pull/11817/checks?check_run_id=95113893719

image

davidzollo:
image

davidzollo:

```

```

davidzollo (approved):
+1
LGTM

PR #9951, opened 2025-10-16
[Improve][Zeta] engine-server and seantunnel-ui support remote paginated queries

liucongjy:
### Purpose of this pull request

Objective: Enable pagination for querying finished and running tasks in the engine-ui interface.

Rationale: When the Seatunnel cluster handles numerous tasks，For example, when exceeding 2,000 tasks， fetching all tasks simultaneously can cause:

- Timeout errors from /running-jobs and /finished-jobs APIs
- Browser performance issues or crashes due to excessive data loading

The effect after implementing pagination is as follows:
image
3e77a213-4c1f-49ab-9932-d6c04f7d6ce3

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

- paginatiion test case: In class RestApiHttpsTest，testFinishedJobsApi() and testRunningJobsApi() test methods
- ui test case: In file seatunnel-engine/seatunnel-engine-ui/src/tests/jobs.spec.ts

### Check list

* [ ] If any new Jar binary package adding in your PR, please add License Notice according
  [New License Guide](https://github.com/apache/seatunnel/blob/dev/docs/en/contribution/new-license.md)
* [ ] If necessary, please update the documentation to describe the new feature. https://github.com/apache/seatunnel/tree/dev/docs
* [ ] If you are contributing the connector code, please check that the following files are updated:
  1. Update [plugin-mapping.properties](https://github.com/apache/seatunnel/blob/dev/plugin-mapping.properties) and add new connector information in it
  2. Update the pom file of [seatunnel-dist](https://github.com/apache/seatunnel/blob/dev/seatunnel-dist/pom.xml)
  3. Add ci label in [label-scope-conf](https://github.com/apache/seatunnel/blob/dev/.github/workflows/labeler/label-scope-conf.yml)
  4. Add e2e testcase in [seatunnel-e2e](https://github.com/apache/seatunnel/tree/dev/seatunnel-e2e/seatunnel-connector-v2-e2e/)
  5. Update connector [plugin_config](https://github.com/apache/seatunnel/blob/dev/config/plugin_config)

PR #11818, opened 2026-08-15
[Fix][Zeta] Add regression test for finished jobs endpoint

### LINKED ISSUE #9868: webui finished job is wrong
### Search before asking

### What happened

已完成作业请求的接口是错的，不知道在哪里能修改？
正确的是:http://localhost:8080/finished-jobs/finished
Image

### SeaTunnel Version

2.3.12

### SeaTunnel Config

```conf
standart config
```

### Running Command

```shell
no
```

### Error Exception

```log
no
```

_No response_

### Java or Scala Version

_No response_

### Screenshots

_No response_

### Code of Conduct

zhangshenghang:
## Purpose
Fixes #9868.

Keep the Finished Jobs page on the unfiltered `/finished-jobs` endpoint so the UI continues to show all terminal jobs, including FINISHED, FAILED, CANCELED, SAVEPOINT_DONE, and UNKNOWABLE jobs.

This PR adds a regression test to prevent the UI service from being changed to the filtered `/finished-jobs/finished` endpoint, which would hide non-FINISHED terminal jobs from the page.

## Testing
- `npx vitest run src/tests/jobs.spec.ts`
- `./mvnw spotless:apply`
- `npm run lint`
- `npm run build`
- `git diff --check`


seatunnel-engine/seatunnel-engine-ui/src/service/job/index.ts
```diff
@@ -20,7 +20,7 @@ import type { Job, JobPage } from './types'
 
 export const getRunningJobs = (page: number, rows: number) =>
   get<JobPage>('/running-jobs/summary', {page: page, rows: rows})
-export const getFinishedJobs = (page: number, rows: number) => get<JobPage>(`/finished-jobs`, {page: page, rows: rows})
+export const getFinishedJobs = (page: number, rows: number) => get<JobPage>(`/finished-jobs/finished`, {page: page, rows: rows})
```
goutamadwant: @zhangshenghang can you check this and if it makes sense - this changes the page behavior, not just the endpoint. /finished-jobs returns all terminal jobs, while /finished-jobs/finished returns only jobs with the FINISHED status. As a result, failed, canceled, and savepoint-completed jobs disappear from the UI, so users cannot inspect their errors or details.
I verified that /finished-jobs?page=1&rows=5 returns 200 on the current PR head. 
Please keep /finished-jobs here and reproduce the reported 500 against the current dev branch. If the 500 still occurs, it should be fixed server-side with a regression test for the record that triggers it.

DanielLeens (changes requested):
Thanks for picking up a long-standing UI complaint, @zhangshenghang — issue #9868 has been open since September 2025 and it deserves attention. The diff is admirably small, the trailing-newline cleanup in `service/job/index.ts` is a nice touch, and you added a test rather than shipping a bare one-liner, which I always appreciate.

Unfortunately, after tracing the endpoint end-to-end I have to recommend against merging this in its current form. My conclusion overlaps with what @goutamadwant already raised, and I want to add the missing piece: I believe I can show exactly what the original 500 was, and that it was already fixed on `dev` several months ago. Details below.

Reviewed at head `0868f6db3e27c1ca1d82ddaabeda04cb808eaff5` against base `dev` (`b1e21631ffdfd8cb33e6adca618b2d95224a9d66`). This is a static source review; I did not run the code.

**Stated problem.** Per the PR description, the Finished Jobs page requests `/finished-jobs`, "causing the server to interpret the query string as a job-state filter". The fix switches the UI to `/finished-jobs/finished`.

**Fix approach.** A single-character-class change in the UI service layer: `getFinishedJobs()` now targets `/finished-jobs/finished` instead of `/finished-jobs`, plus a vitest assertion that the new literal is passed to `get()`.

**One-sentence summary.** The PR changes which URL the bundled Web UI polls for the Finished Jobs table; it does not touch any server code, and the URL it moves to means something different from the URL it moves away from.

## Why the stated problem is not what is happening

The endpoint is served by `FinishedJobsServlet`, registered in `JettyService` as a prefix mapping:

- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/JettyService.java:227` — `context.addServlet(finishedJobsHolder, convertUrlToPath(REST_URL_FINISHED_JOBS));`
- `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/JettyService.java:276-278` — `convertUrlToPath(url)` returns `url + "/*"`, i.e. the mapping is `/finished-jobs/*`.

The servlet then derives the state filter from the path info:

```java
// seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/servlet/FinishedJobsServlet.java:45-53
String state = req.getPathInfo();
if (state != null && state.length() > 1) {
    state = state.substring(1);
} else {
    state = "";
}
writeJsonWithPagination(req, resp, jobInfoService.getJobsByStateJson(state));
```

`HttpServletRequest.getPathInfo()` returns the extra path information *after* the servlet path, and by the Servlet specification it never contains the query string (that is `getQueryString()`). For `GET /finished-jobs?page=1&rows=10` the servlet path is `/finished-jobs` and `getPathInfo()` is `null`, so `state` becomes `""`. The query string is consumed separately by `PageBaseServlet.getParameterMap(req)`. It is therefore not possible for `?page=1&rows=10` to be mistaken for a state filter.

This is also asserted by an existing test on `dev`:

```java
// seatunnel-engine/seatunnel-engine-server/src/test/java/org/apache/seatunnel/engine/server/rest/RestApiHttpsTest.java:167-176
restApiRequestHttp(
        "http://localhost:" + HTTP_PORT2 + "/finished-jobs?page=1&rows=" + pageSize,
        (code, content) -> {
            Assertions.assertEquals(200, code);
            ...
            Assertions.assertTrue(total == jobNum && data.size() == pageSize);
        });
```

`total == jobNum` — the endpoint the UI currently calls returns **all** finished jobs, with pagination, and returns 200. `RestApiHttpsTest.java:191` additionally covers the unpaginated `/finished-jobs` form. This matches @goutamadwant's manual observation that `/finished-jobs?page=1&rows=5` returns 200 on this PR's head.

## What issue #9868 most likely actually was

I think the reporter's diagnosis was an understandable misread of a real 500, and the real defect is elsewhere. The reporter was on 2.3.12. Comparing tag `2.3.12` with `dev`:

```java
// 2.3.12, JobInfoService.getJobsByStateJson
.filter(jobState -> { if (state.isEmpty()) { return true; }
                      return jobState.getJobStatus().name().equals(state.toUpperCase()); })
.sorted(Comparator.comparing(JobState::getFinishTime))   // <-- no null handling
```

`JobState.getFinishTime()` can legitimately be `null` for a record that reached the finished-job IMap without a finish timestamp. `Comparator.comparing` with a natural-ordering key comparator throws `NullPointerException` on such a record, which `RestHttpGetCommandProcessor.java:140` and the servlet path both surface as **HTTP 500**.

Crucially, the state filter runs **before** the sort. So if the offending record happened to be `CANCELED`, `FAILED` or `SAVEPOINT_DONE`, then `/finished-jobs/finished` filtered it out and the sort no longer threw — while bare `/finished-jobs` still 500'd. That is exactly the asymmetry the reporter saw, and it explains why they concluded the URL was wrong rather than the sort.

That root cause is already fixed on `dev` by **#11269 "[Fix][Zeta] Make finished jobs API tolerate missing history data"** (`54a176c5de`):

```java
// seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/service/JobInfoService.java:125,133-136
.filter(java.util.Objects::nonNull)
...
.sorted(Comparator.comparing(JobState::getFinishTime,
        Comparator.nullsLast(Comparator.reverseOrder())))
```

with a dedicated regression test, `JobInfoServiceNullSafetyTest.shouldListFinishedJobsWhenFinishTimeMetricsOrDagAreMissing`, which drives `getJobsByStateJson("")` — the empty-state path — with a `null` `finishTime` and asserts both jobs are returned in the right order. Separately, `#9951` (merged 2026-01-30) added the `{data,total}` pagination envelope that the current UI expects.

So my read is that **#9868 is already resolved on `dev`**, and the correct outcome for that issue is verification and closure (or a reproduction against `dev` if the reporter still sees it), not a client-side URL change. I could be wrong about the exact NPE — I have not reproduced it — but the 2.3.12-vs-`dev` diff and the targeted regression test added by #11269 make it the best-supported explanation, and it is testable.

## Broken vs correct response, concretely

Given three terminal jobs in the cluster:

`GET /finished-jobs?page=1&rows=10` — what the UI requests on `dev` today (correct):

```json
{"total": 3, "data": [
  {"jobId": "8884...01", "jobName": "etl_a", "jobStatus": "FINISHED",  "errorMsg": "",             "finishTime": "2026-08-15 10:03:11"},
  {"jobId": "8884...02", "jobName": "etl_b", "jobStatus": "FAILED",    "errorMsg": "Connection …", "finishTime": "2026-08-15 10:02:40"},
  {"jobId": "8884...03", "jobName": "etl_c", "jobStatus": "CANCELED",  "errorMsg": "",             "finishTime": "2026-08-15 10:01:05"}
]}
```

`GET /finished-jobs/finished?page=1&rows=10` — what this PR requests:

```json
{"total": 1, "data": [
  {"jobId": "8884...01", "jobName": "etl_a", "jobStatus": "FINISHED", "errorMsg": "", "finishTime": "2026-08-15 10:03:11"}
]}
```

On 2.3.12, the same first request would have returned HTTP 500 with an `NullPointerException` body whenever one of those records carried a `null` `finishTime`.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Terminal states.** `JobStatus` (`seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/job/JobStatus.java`) declares five `EndState.GLOBALLY` values: `FAILED` (:47), `SAVEPOINT_DONE` (:53), `CANCELED` (:60), `FINISHED` (:63), `UNKNOWABLE` (:66). All five land in `Constant.IMAP_FINISHED_JOB_STATE`.

**The filter.** `JobInfoService.getJobsByStateJson` (`JobInfoService.java:117`) treats the empty state as "match everything" and otherwise does an exact enum-name match:

```java
if (state.isEmpty()) { return true; }
return jobState.getJobStatus().name().equals(state.toUpperCase());
```

So `""` → 5 of 5 terminal states; `"finished"` → 1 of 5. This PR moves the UI from the first to the second. That is the central problem, and it is a behavior change, not an endpoint correction.

**Does the normal UI polling path reach this code?** Yes, and only this code. `views/jobs/finished-jobs.tsx:38-46` calls `JobsService.getFinishedJobs(page, pageSize)` on mount and re-arms a 5-second timer; `service/job/index.ts:23` is the sole implementation, and `finished-jobs.tsx:39` is its sole caller in the application. There is no other code path, so every Finished Jobs page load and every 5-second poll goes through the changed line.

**Design evidence that the page is meant to show all terminal states.** `finished-jobs.tsx:79-89` renders a dedicated **Status** column through `getColorFromStatus(row.jobStatus)`, and `utils/getTypeFromStatus.ts` assigns distinct colors to `CANCELED` (warning), `FAILED` (error), `SAVEPOINT_DONE` (info-muted) and `FINISHED` (info). A status column with a five-way color map is meaningless on a list that can only ever contain `FINISHED`. The documented REST contract agrees — `docs/en/engines/zeta/rest-api-v2.md:654` marks `state` as **optional**, precisely so the stateless form means "all terminal jobs".

---

**Issue 1 — Narrowing the endpoint drops four of five terminal job states from the UI**
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/service/job/index.ts:23`
- **Problem:** `/finished-jobs/finished` sets `state = "finished"` in `FinishedJobsServlet.java:45-51`, which makes `JobInfoService.java:117-126` return only `JobStatus.FINISHED` records. `FAILED`, `CANCELED`, `SAVEPOINT_DONE` and `UNKNOWABLE` jobs disappear from the Finished Jobs page.
- **Potential risk:** In a production cluster this is exactly the wrong half to hide. Operators use this page to find the job that just failed. The page also silently under-reports: the `total` used for the pager (`finished-jobs.tsx:41`) becomes the FINISHED-only count, so the UI actively tells the operator that fewer jobs completed than actually did. There is no compensating filter control in the UI to get them back.
- **Best improvement:** Revert `service/job/index.ts:23` to `/finished-jobs`. If per-state filtering is genuinely wanted, add it as a user-facing control on the page (a status dropdown that appends `/:state` only when a state is chosen), keeping the stateless call as the default.
- **Severity:** Critical
- **Raised by another reviewer:** Yes (@goutamadwant)

**Issue 2 — The stated root cause is not reproducible against the current code, and the real defect is already fixed on `dev`**
- **Location:** PR description; `seatunnel-engine/seatunnel-engine-ui/src/service/job/index.ts:23`
- **Problem:** The premise that "the server interprets the query string as a job-state filter" cannot hold: `FinishedJobsServlet.java:45` reads `req.getPathInfo()`, which by the Servlet specification excludes the query string, and `PageBaseServlet.java:45` reads `page`/`rows` from the parameter map. `RestApiHttpsTest.java:167-176` asserts `/finished-jobs?page=1&rows=N` returns 200 with `total == jobNum`. The 500 in #9868 is far better explained by the unguarded `Comparator.comparing(JobState::getFinishTime)` present in 2.3.12, fixed on `dev` by #11269 (`54a176c5de`) via `Comparator.nullsLast(Comparator.reverseOrder())` plus `.filter(Objects::nonNull)` at `JobInfoService.java:125,133-136`.
- **Potential risk:** Merging a fix built on an incorrect diagnosis leaves the true fix unattributed and closes #9868 for the wrong reason. If a residual server-side 500 does still exist, this change would mask it on the default page while leaving every direct REST consumer exposed.
- **Best improvement:** Reproduce #9868 against current `dev` (2.3.13-SNAPSHOT). If it no longer reproduces, close #9868 referencing #11269 and #9951 and withdraw this PR. If it still reproduces, capture the server-side stack trace and fix it in `JobInfoService`/`FinishedJobsServlet` with a regression test alongside `JobInfoServiceNullSafetyTest`.
- **Severity:** Critical
- **Raised by another reviewer:** Yes (@goutamadwant) — he asked for a reproduction against `dev`; the #11269 root-cause attribution above is new.

**Issue 3 — Failed and canceled jobs become unreachable through the Web UI entirely**
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/views/jobs/finished-jobs.tsx:50-52`, `.../views/jobs/detail.tsx:609`
- **Problem:** The job detail view is the only place the UI surfaces `job.errorMsg` (`detail.tsx:609`), and the only in-app navigation to it for a terminal job is the "View" button on the Finished Jobs row (`finished-jobs.tsx:50-52`, routing to `jobs/:jobId` per `router/routes.ts:40-41`). Running Jobs only lists non-terminal jobs. Once a `FAILED` job is filtered out of the list, there is no UI path left to its detail page or its error message.
- **Potential risk:** Operators lose in-product failure diagnosis and must fall back to `curl` or log files. This is a functional regression in the primary incident-response surface of the console.
- **Best improvement:** Covered by reverting Issue 1. If a state filter is added later, make the default include all terminal states, and consider surfacing `errorMsg` inline in the list for failed rows.
- **Severity:** High
- **Raised by another reviewer:** Yes (@goutamadwant) raised the symptom; the navigation dead-end analysis is new.

## 1.2 Compatibility Impact

**Explicit conclusion: no wire-format or REST contract break, but a user-visible behavior regression in the bundled UI.**

- **Response schema:** unchanged. Both `/finished-jobs` and `/finished-jobs/finished` are handled by the same `FinishedJobsServlet` and the same `PageBaseServlet.writeJsonWithPagination` (`PageBaseServlet.java:62-73`), producing `{"data": [...], "total": N}` when `page` is present and a bare array otherwise. The UI's `JobPage` type (`service/job/types.ts:98-101`) still matches. No field is added, removed or renamed.
- **Server API:** untouched. `RestConstant.REST_URL_FINISHED_JOBS`, the Jetty mapping, and the v1 `RestHttpGetCommandProcessor.java:140` route are all unmodified, so external REST consumers, the e2e suites (`ClusterSeaTunnelEngineContainer.java`, `ClassLoaderITBase.java:160`) and any third-party integration are unaffected.
- **What does change:** the semantics of the Finished Jobs page for every existing user upgrading to a build containing this commit. Someone who has used this console since 2.3.x will find their failed jobs silently gone with no setting to restore them and no release note. In a financial-grade operational context that is the kind of change that generates an incident ticket, not a bug fix.
- **Historical behavior:** as far back as tag `2.3.12` the UI called bare `/finished-jobs` (`2.3.12: service/job/index.ts` — `get<Job[]>('/finished-jobs')`), i.e. "all terminal states" has been the page's contract for its entire existence.

**No cross-user data exposure.** The engine REST layer has no per-user authorization model here; both URLs read the same cluster-wide `IMAP_FINISHED_JOB_STATE`. Narrowing the state filter neither adds nor removes an access-control boundary. Nothing to flag.

**Full scan and N+1, pre-existing.** `JobInfoService.java:117-136` calls `finishedJob.values()` — a full scan of the distributed `IMAP_FINISHED_JOB_STATE` — and then, per surviving record, performs `getFinishedJobMetricsJson(jobId)` (an `IMap.getOrDefault` on `IMAP_FINISHED_JOB_METRICS`) and `finishedJobDAGInfo.get(jobId)`. That is 2N additional IMap lookups per request. Pagination is applied *afterwards*, in the servlet (`PageBaseServlet.java:59-66`), so `?page=1&rows=10` still materializes JSON for every finished job in the cluster, every 5 seconds, for every open browser tab. The project already documents this (`docs/en/engines/zeta/rest-api-job-lifecycle.md:365`). This PR does not introduce it; because the state filter is applied before the `.map()`, narrowing to `FINISHED` incidentally *reduces* N. I mention it only so the incidental speedup is not mistaken for a justification — buying latency by hiding failed jobs is not a trade this project should take. The real fix is server-side pagination pushed below the metrics/DAG hydration, which would be a welcome follow-up PR.

**Pagination interaction — a genuine new failure mode.** See Issue 4.

---

**Issue 4 — Shrinking `total` can strand the pager on a page that now 400s, permanently freezing the polling loop**
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/views/jobs/finished-jobs.tsx:38-43`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/servlet/PageBaseServlet.java:77-78`
- **Problem:** `writeJsonWithPagination` throws `IllegalArgumentException("Page number exceeds total pages")` when `start > total`, which the REST layer maps to HTTP 400 (asserted at `RestApiHttpsTest.java:270-273`). The UI's poll loop is:
  ```ts
  const fetch = async () => {
    const res = await JobsService.getFinishedJobs(page.value, pageSize.value)
    jobs.value = res.data
    total.value = res.total
    timer = setTimeout(fetch, 5000)   // only reached on success
  }
  ```
  A rejected request skips the `setTimeout`, so the page stops refreshing for the rest of its lifetime with no error shown. This PR makes hitting that state substantially more likely: filtering to `FINISHED`-only can shrink `total` by an order of magnitude, so a user sitting on page 3 after an upgrade — or after a burst of jobs fails — lands on a page that no longer exists.
- **Potential risk:** A console that appears live but is silently frozen is worse than one that errors loudly, particularly during an incident when the operator is trusting it to update.
- **Best improvement:** Independently of this PR, wrap the `fetch` body in `try/finally` so the timer is always re-armed, surface the failure to the user, and clamp `page.value` when `total` shrinks below the current offset. Server-side, returning an empty `data` array with the correct `total` for an over-range page would be friendlier than a 400, though that is a contract change worth its own discussion.
- **Severity:** High

Nothing regresses here, and nothing improves. The changed line adds no error handling, and the UI layer above it has none: `service/service.ts` rejects on non-200 and `finished-jobs.tsx` has no `catch`, so a 400/500 from this endpoint surfaces only as an unhandled promise rejection in the browser console. The server side is in reasonable shape after #11269 — `getFinishedJobMetricsJson` and `getFinishedJobDAGInfo` (`JobInfoService.java:148-168`) both log a `WARN` and degrade gracefully. No sensitive data is logged. I would only note that the missing client-side `catch` (Issue 4) is the one gap worth closing.

# 2. Code Quality Assessment

## 2.1 Coding Standards

The diff is clean and idiomatic. Two small positives worth calling out: restoring the trailing newline at the end of `service/job/index.ts` is a genuine (if tiny) hygiene fix, and keeping the change to a single service-layer function rather than editing the component was the right layering instinct. No license-header, formatting or import concerns. One nit: the template literal on `service/job/index.ts:23` no longer interpolates anything and could be a plain string, matching the `'/running-jobs/summary'` style on the line above — pre-existing, not introduced here.

**Issue 5 — The new test is a tautology and asserts none of the behavior that matters**
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/tests/jobs.spec.ts:78-85`
- **Problem:** The test mocks `get` and asserts it was called with the literal `'/finished-jobs/finished'` — the same literal the PR just wrote into `service/job/index.ts:23`. It mirrors the implementation constant, so it passes for *any* URL as long as both sides are edited together, including a URL that returns the wrong jobs. It would have passed equally well before the change (with the old literal) and will pass after any future change. It proves the string is copied correctly; it proves nothing about the finished-jobs scenario that was reported broken.
- **Potential risk:** False confidence. A change-detector test in place of a behavior test is worse than no test, because it looks like coverage in review.
- **Best improvement:** Test the behavior, at the level where the behavior lives. Server side, add a case to `RestApiHttpsTest` (or a `JobInfoService` unit test next to `JobInfoServiceNullSafetyTest`) that seeds one job per terminal state — `FINISHED`, `FAILED`, `CANCELED`, `SAVEPOINT_DONE` — and asserts that `GET /finished-jobs?page=1&rows=10` returns all four while `GET /finished-jobs/FAILED` returns exactly one. That is the assertion that would have caught this regression, and it also pins the contract for future refactors. Client side, a component-level test that mounts `finished-jobs.tsx` against a mocked mixed-status payload and asserts a `FAILED` row renders would be the right complement.
- **Severity:** Medium

**Issue 6 — CI never executes the added test, so this PR ships with zero automated verification**
- **Location:** `.github/workflows/backend.yml:450-451` (`seatunnel-ui` job, `if: needs.changes.outputs.api == 'true'`) and `.github/workflows/backend.yml:230` (the `api` path globs)
- **Problem:** The `Build SeaTunnel UI` job is the only job that runs `npm run lint`, `npm run test:unit` and `npm run build` (`backend.yml:464-475`). It is gated on the `api` path filter, whose glob list at `backend.yml:230` does not include `seatunnel-engine/seatunnel-engine-ui/**`. On this PR's run, `Build SeaTunnel UI` is reported as **skipped**. The `unit-test` job that did run invokes Maven with `-Dskip.ui=true`, so it does not compensate. The net effect: a PR whose entire diff is TypeScript had none of its TypeScript linted, type-checked, built or tested by CI.
- **Potential risk:** UI-only PRs — including this one, and including any future revert — merge without ever being verified. A typo'd URL or a broken build would reach `dev` unnoticed. This is a coverage gap in the project's CI, not something you introduced, but it materially affects how much weight this PR's green-ish status carries.
- **Best improvement:** Worth a separate small PR adding `"seatunnel-engine/seatunnel-engine-ui/**"` to the `api_files` glob list at `backend.yml:230`, or better, giving the `seatunnel-ui` job its own dedicated path filter so UI changes trigger it without pulling in the full API connector matrix. Happy to help review that if you want to split it out.
- **Severity:** Medium

**Issue 7 — File-scoped `vi.mock` replaces the whole service module and silently drops two exports**
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/tests/jobs.spec.ts:30`
- **Problem:** `vi.mock('@/service/service', () => ({ get: vi.fn() }))` is hoisted to the top of the module and applies to every test in the file, replacing the real module with an object that has only `get`. The real `service.ts` also exports `post` and `axios`. Today nothing in this spec's import graph needs them, so it works — but the two existing component tests now run against a stubbed transport they do not expect, and any future test in this file (or any component newly imported into it) that touches `post`/`axios` will get `undefined` and fail with a confusing `is not a function`.
- **Potential risk:** Low today, but a latent trap for the next contributor, and it couples two unrelated tests to a global mock.
- **Best improvement:** Use `vi.mock('@/service/service', async (importOriginal) => ({ ...(await importOriginal<typeof import('@/service/service')>()), get: vi.fn() }))` to preserve the other exports, or scope the mocking to the single test with `vi.spyOn`.
- **Severity:** Low

## 2.3 Documentation Updates

No documentation change is required by this PR, because the response shape is unchanged. But the docs are worth a look because they contradict the PR's premise and are themselves slightly stale.

**Issue 8 — REST docs mark `state` as optional (contradicting this change) and still show the pre-pagination response shape**
- **Location:** `docs/en/engines/zeta/rest-api-v2.md:654-663` and `docs/zh/engines/zeta/rest-api-v2.md:631-640`
- **Problem:** Two things. First, the parameter table documents `state` as **optional** with values `FINISHED, CANCELED, FAILED, SAVEPOINT_DONE, UNKNOWABLE` — i.e. the documented contract is that omitting the state returns all of them, which is what the UI relies on and what this PR moves away from. Second, the `Responses` example still shows a bare JSON array even though `page`/`rows` are documented on the same endpoint and have returned `{"data": [...], "total": N}` since #9951. A reader following these docs would be surprised by the actual paginated payload — plausibly part of why #9868's reporter concluded the endpoint was "wrong".
- **Potential risk:** Users writing monitoring integrations against the documented shape will break on the real one. Reviewers checking this PR against the docs will find the docs supporting the current UI behavior, not the change.
- **Best improvement:** Update both `docs/en` and `docs/zh` v2 REST references to show the paginated envelope when `page` is supplied, alongside the bare-array form when it is not, and explicitly state that omitting `:state` returns all terminal states. Small, self-contained, and a nice standalone contribution.
- **Severity:** Low

# 3. Architectural Soundness

The instinct — fix it in one place, in the service layer, with a test — is right, and I want to encourage that. The problem is that the change is applied at the wrong layer for the defect. If the server were genuinely returning wrong data for `/finished-jobs`, patching the client to request a *different resource* would be treating the symptom; the endpoint would still be broken for every non-UI consumer, and the e2e suites that exercise `REST_URL_FINISHED_JOBS` would still be asserting against a broken path. A one-line client change is only the elegant fix when the client was genuinely asking the wrong question — and here, per `docs/en/engines/zeta/rest-api-v2.md:654` and `RestApiHttpsTest.java:167`, it was asking exactly the right one.

## 3.2 Maintainability

Neutral to slightly negative. The URL literal is now duplicated between `service/job/index.ts:23` and `tests/jobs.spec.ts:84` with no shared constant, so the two must be edited in lockstep and the test cannot catch a wrong edit (Issue 5). If per-state filtering is added later, hardcoding `finished` into the service function will have to be undone anyway in favor of a parameter.

## 3.3 Extensibility

The server side is already well-factored for this: `getJobsByStateJson(state)` (`JobInfoService.java:117`) plus the `/*` Jetty mapping gives you every state for free. The natural extension is a `getFinishedJobs(page, rows, state?)` in the UI service that appends `/${state}` only when a state is supplied, wired to a dropdown next to the existing Status column. That gets you the behavior this PR seems to want *and* keeps the default correct. If you would like to take that route, I think it would be a genuinely useful feature and I would be glad to review it.

The bundled UI has requested bare `/finished-jobs` since at least tag `2.3.12`, and the server has treated an absent state as "all terminal states" for just as long — the empty-state branch at `JobInfoService.java:120-122` is unchanged between `2.3.12` and `dev`, and is explicitly regression-tested by `JobInfoServiceNullSafetyTest.shouldListFinishedJobsWhenFinishTimeMetricsOrDagAreMissing`. There is no serialization, checkpoint, savepoint or state-storage surface involved, so there is no upgrade or recovery risk. The compatibility concern is purely behavioral: users upgrading across this commit lose visibility of failed jobs in the console with no migration note and no opt-out. A user-visible behavior change of this kind would at minimum need an entry in `docs/en/introduction/concepts/incompatible-changes.md` with migration guidance; as written it has neither.

# 4. Issue Summary

| 1 | Narrowing to `/finished-jobs/finished` drops 4 of 5 terminal states (`FAILED`, `CANCELED`, `SAVEPOINT_DONE`, `UNKNOWABLE`) from the UI | `seatunnel-engine-ui/src/service/job/index.ts:23` | Critical |
| 2 | Stated root cause is not reproducible; `getPathInfo()` excludes the query string, and the real 500 was fixed on `dev` by #11269 | PR description; `seatunnel-engine-ui/src/service/job/index.ts:23` | Critical |
| 3 | Failed/canceled jobs become unreachable in the UI — `errorMsg` is only shown on the detail page, reachable only from this list | `views/jobs/finished-jobs.tsx:50-52`; `views/jobs/detail.tsx:609` | High |
| 4 | Shrinking `total` can strand the pager on a page that 400s, permanently freezing the 5s poll loop with no error shown | `views/jobs/finished-jobs.tsx:38-43`; `rest/servlet/PageBaseServlet.java:77-78` | High |
| 5 | New test is a tautology mirroring the implementation constant; asserts no finished-job behavior | `seatunnel-engine-ui/src/tests/jobs.spec.ts:78-85` | Medium |
| 6 | CI skips `Build SeaTunnel UI` for UI-only PRs, so the added test, lint, type-check and build never run | `.github/workflows/backend.yml:450-451,230` | Medium |
| 7 | File-scoped `vi.mock` replaces the whole service module, dropping the `post` and `axios` exports | `seatunnel-engine-ui/src/tests/jobs.spec.ts:30` | Low |
| 8 | REST docs mark `state` as optional (contradicting this PR) and still show the pre-pagination response shape | `docs/en/.../rest-api-v2.md:654-663`; `docs/zh/.../rest-api-v2.md:631-640` | Low |

# 5. Merge Recommendation

**Conclusion: Request changes — do not merge in its current form.**

**Blockers (must be resolved before merge):**
1. **Issue 1** — revert `service/job/index.ts:23` to `/finished-jobs`. The Finished Jobs page must continue to show all terminal states.
2. **Issue 2** — reproduce #9868 against current `dev`. If it does not reproduce, close #9868 referencing #11269 (`54a176c5de`) and #9951 and withdraw this PR. If it does, fix it server-side in `JobInfoService`/`FinishedJobsServlet` with a regression test.
3. **Issue 3** — resolved automatically by reverting Issue 1; please confirm a `FAILED` job is still reachable end-to-end from the list to the detail page and its `errorMsg`.

**Recommended (valuable, but each better as its own PR):**
4. **Issue 4** — always re-arm the poll timer and clamp `page` when `total` shrinks; this is a real latent bug worth fixing regardless of the outcome here.
5. **Issue 5** — replace the tautological test with a multi-state assertion in `RestApiHttpsTest`/`JobInfoService` tests.
6. **Issue 6** — add `seatunnel-engine-ui/**` to the CI path filter so UI PRs are actually built and tested.
7. **Issues 7, 8** — mock hygiene and the stale REST docs.
8. Optional and genuinely nice: a status dropdown on the Finished Jobs page, defaulting to all states, using `/finished-jobs/:state`.

**CI status.** The `Build` check is failing, but neither failure is attributable to this PR — which is worth stating plainly so it does not muddy the discussion. On the fork run for head `0868f6db3e27c1ca1d82ddaabeda04cb808eaff5`:
- `unit-test (11, windows-latest)` — `MultiTableSinkWriterTest.testFailFastDoesNotUseTableRetry:288 Expected java.io.IOException to be thrown, but nothing was thrown.` in `seatunnel-api`. This is a timing-sensitive assertion that depends on the asynchronous `MultiTableWriterRunnable` having recorded its failure before `prepareCommit(1L)` is invoked; it is Windows-only here and is a known flake in that area.
- `all-connectors-it-1 (11, ubuntu-latest)` — `CouchbaseIT.startUp:99 » ContainerLaunch Container startup failed`, a Testcontainers infrastructure failure.
- The remaining `cancelled` jobs are fail-fast cascades from those two.

A diff consisting solely of two `.ts` files cannot influence either — and notably, the one job that *would* have exercised those files, `Build SeaTunnel UI`, was skipped (Issue 6). A rerun of the failed jobs should clear the red.

**Overall assessment.** I want to be clear that the underlying instinct here is good: #9868 is a real report that sat untriaged for months, and going after it with a minimal, tested change is exactly the right reflex. The gap is diagnostic rather than one of care — `getPathInfo()` genuinely looks like it might carry the query string until you check the Servlet contract, and the reporter's observation that `/finished-jobs/finished` "worked" is real and had a subtle cause (the state filter running before the NPE-throwing sort) that took some digging to pin down. The most valuable outcome of this PR may well be closing #9868 correctly with that explanation attached. If instead you find it still reproduces on `dev`, I would be very glad to review a server-side fix, and I would also happily review the CI path-filter fix in Issue 6 — that one benefits every future UI contributor. Thank you for the contribution, and please do ping me on the follow-up.

DanielLeens (changes requested):
Thanks for the update, @zhangshenghang — I re-reviewed the full diff at the new head. Good news: the new commit (`d5181829`, "Keep finished jobs endpoint unfiltered") reverts exactly what my previous round and @goutamadwant both flagged as the critical problem. I re-traced the whole thing end-to-end again rather than just diffing against my last review, so this comment supersedes my previous `Changes Requested`.

Reviewed at head `d5181829691a06042842be7e9273db0e9c19e1a8` against base `dev` (`f1a1a0abbe24bdac8cf23307995a78f778a3f467`). This is a static source review; I did not run the code.

**User pain point.** Issue #9868 reports the Finished Jobs page failing to load. The original version of this PR tried to fix that by pointing the UI at `/finished-jobs/finished` instead of `/finished-jobs`.

**Fix approach (current head).** The new commit reverses that: `getFinishedJobs()` now calls `/finished-jobs` again (byte-for-byte the same URL `dev` already uses), plus a genuinely new regression test that pins the service layer to that URL, plus a cosmetic quote-style/trailing-newline cleanup.

**One-sentence summary.** After this round's commit, the PR no longer changes what the Finished Jobs page requests — it adds a regression test that protects the currently-correct, currently-shipped behavior from being silently narrowed again.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**Precise description of the change.** Comparing `dev` and the current PR head directly:

```
diff --git a/seatunnel-engine/seatunnel-engine-ui/src/service/job/index.ts
-export const getFinishedJobs = (page, rows) => get<JobPage>(`/finished-jobs`, {page, rows})
+export const getFinishedJobs = (page, rows) => get<JobPage>('/finished-jobs', {page, rows})
```

That is the *entire* runtime-behavior diff between `dev` and this PR's head: a template literal became a plain string. There is no string interpolation in either form, so this is a no-op for the emitted URL. The trailing-newline fix is likewise cosmetic.

**Before / after across the whole PR lifetime:**
- Commit `0868f6d` (previous head, what I reviewed on 8/16): `/finished-jobs` → `/finished-jobs/finished`. This was the regression I flagged as Issue 1 (Critical).
- Commit `d518182` (current head): `/finished-jobs/finished` → `/finished-jobs`. This restores the original, correct URL.

**Key findings**
- The normal UI polling path (`finished-jobs.tsx:38-46`, 5s timer, sole caller of `JobsService.getFinishedJobs`) now resolves to the exact same server call as `dev`, so there is zero behavioral change for real users landing on this page.
- `JobInfoService.getJobsByStateJson` (`JobInfoService.java:117-126`) still receives an empty `state` for this call, so all five `EndState.GLOBALLY` values (`FINISHED`, `FAILED`, `CANCELED`, `SAVEPOINT_DONE`, `UNKNOWABLE`) are still returned — my Issue 1 and Issue 3 from the previous round are both resolved by this revert, not by any new server-side logic.
- This is best classified as a **defensive fix**: it does not touch the server-side path that #9868 actually complains about, and the diff no longer claims to. It converts what would otherwise be a silent, undiscussed URL regression into a pinned, tested contract.
- The new test in `jobs.spec.ts:78-85` now asserts the *correct* literal (`'/finished-jobs'`) instead of the incorrect one from the previous head. It is still a change-detector test (see Issue 5 below, downgraded from my last round), but the thing it detects a change *away from* is now the right baseline, so it has real regression value going forward.
- Blind spot that remains: this commit does not resolve the actual question behind #9868 — whether a 500 still reproduces against current `dev`. See Issue 2 below.

**In-depth correctness analysis.** Since `getFinishedJobs('/finished-jobs', ...)` is identical to `dev`, every downstream consequence I traced in my previous review (state-filter semantics, pagination envelope, `errorMsg`/detail-page reachability for `FAILED`/`CANCELED` jobs) is now unchanged from `dev`'s current, already-shipped behavior. There is no new state, branch, lifecycle, or race window introduced by this diff — it is exactly `dev`'s existing code path, which does not implicate checkpoint/recovery/serialization at all (this is a stateless REST read).

## 1.2 Compatibility Impact

**Explicit conclusion: Fully compatible.**

The service function now emits byte-identical URLs and payloads to what `dev` already sends. No API, config option, default value, protocol, or serialization format is touched, and — unlike the previous head I reviewed — there is no behavioral change for existing users either. Nothing needs to go into `incompatible-changes.md`.

No change from `dev`. The pre-existing full-scan-plus-N-IMap-lookups pattern in `JobInfoService.getJobsByStateJson` (unfiltered case) is untouched by this PR, as is the pre-existing pager-stranding risk I raised as Issue 4 last round (`finished-jobs.tsx:38-43` skips `setTimeout` on a rejected fetch). Since the URL is no longer narrowed, the specific trigger I described for Issue 4 ("filtering to FINISHED-only can shrink total by an order of magnitude") no longer applies to this PR, but the general bug is still live in `dev` independent of this PR and is still worth its own fix.

Unchanged from `dev`. No new exception handling, no new logging, nothing swallowed, nothing sensitive exposed.

### Issue 1 — PR title and description still describe the reverted (incorrect) behavior — **carried forward from source, newly relevant now that it will mislead the merge commit**
- **Location:** PR title (`[Fix][Zeta] Fix finished jobs UI endpoint`) and PR description (`## Purpose` section)
- **Problem description:** The PR description says "The page now requests `/finished-jobs/finished`, which explicitly selects all terminal jobs" — this describes the *previous*, now-reverted head, and it was already factually wrong even then (`/finished-jobs/finished` selects only `FINISHED`, not "all terminal jobs" — that's what plain `/finished-jobs` does). The commit message on the actual current head, `[Fix][Zeta] Keep finished jobs endpoint unfiltered`, correctly describes what the code now does, but the PR-level title/description were never updated to match.
- **Potential risk:** Apache SeaTunnel PRs are typically squash-merged using the PR title and description as the resulting commit message. If merged as-is, the permanent `dev` history would carry a commit that says it switched the endpoint to a filtered one, when the actual diff does the opposite (reverts to the unfiltered endpoint plus a regression test). That is exactly the kind of misleading commit message that makes future `git blame`/bisection on this file harder, and it directly contradicts the "why" a future maintainer would rely on.
- **Best improvement:** Update the PR title to match the final commit's intent (e.g. `[Fix][Zeta] Add regression test to keep finished jobs endpoint unfiltered`) and rewrite the description to state plainly: the endpoint is being kept as `/finished-jobs` (matching `dev`), a regression test now pins that contract, and #9868's underlying report should be tracked/verified separately (see Issue 2).
- **Severity:** Medium

### Issue 2 — #9868's actual root cause is still unverified against current `dev`
- **Location:** PR description ("Fixes #9868"); no code location, this is a process gap
- **Problem description:** Neither this round nor the previous one demonstrated that a 500 still reproduces on `dev`. My previous review's hypothesis — that the real defect was the unguarded `Comparator.comparing(JobState::getFinishTime)` NPE, already fixed by #11269 (`54a176c5de`) with a dedicated null-safety regression test — has not been confirmed or refuted by the author. @goutamadwant asked for the same reproduction in his review.
- **Potential risk:** Since this PR's current head makes no server-side change, merging it will not actually close #9868 if the reporter's environment still 500s. The issue would stay open with a merged "fix" attached to it that fixed nothing server-side, which is confusing for anyone triaging #9868 later.
- **Best improvement:** Before merge (or as a fast follow-up comment on the issue), reproduce #9868 against current `dev`/2.3.13-SNAPSHOT. If it no longer reproduces, close #9868 referencing #11269, and this PR becomes "add a regression test for the endpoint contract that #9868's reporter was confused about" — which is a fine, mergeable PR on its own merits. If it still reproduces, that is a separate server-side bug that needs its own fix and test in `JobInfoService`/`FinishedJobsServlet`.
- **Severity:** Medium
- **Raised by another reviewer:** Yes (@goutamadwant asked for the same reproduction; not resolved by this commit).

### Issue 3 — Pager can still strand the poll loop on an out-of-range page with no error surfaced (pre-existing in `dev`, unaffected by this PR)
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/views/jobs/finished-jobs.tsx:38-43`; `seatunnel-engine/seatunnel-engine-server/src/main/java/org/apache/seatunnel/engine/server/rest/servlet/PageBaseServlet.java:77-78`
- **Problem description:** `writeJsonWithPagination` throws `IllegalArgumentException` → HTTP 400 when `start > total` (`RestApiHttpsTest.java:270-273`). The UI's `fetch()` only re-arms its 5s `setTimeout` on success, so a rejected request permanently freezes the poll with no visible error. This is unchanged from `dev` and is no longer exacerbated by this PR (my previous Issue 4 concern about `FINISHED`-only shrinking `total` no longer applies once the URL is unfiltered again), but the underlying bug is still there for any operator who is on a stale page while jobs churn.
- **Potential risk:** A console that appears live but is silently frozen during an incident.
- **Best improvement:** Independent follow-up PR: wrap `fetch` in `try/finally` to always re-arm the timer, surface the failure, and clamp `page.value` when `total` shrinks below the current offset.
- **Severity:** Low (downgraded from High last round, since this PR no longer makes the trigger condition more likely)

### Issue 4 — Regression test still asserts a literal string rather than the underlying multi-state behavior
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/tests/jobs.spec.ts:78-85`
- **Problem description:** The new test mocks `get()` and asserts it is called with `'/finished-jobs'`. That is a correct assertion now (unlike the previous head, which asserted the wrong literal), and it does add real regression value — nobody can silently re-introduce the `/finished-jobs/finished` narrowing without this test failing. But it still does not exercise the thing that actually matters to users: that a mixed set of terminal-state jobs (`FINISHED`, `FAILED`, `CANCELED`, `SAVEPOINT_DONE`) all show up in the rendered list.
- **Potential risk:** Lower than before (the literal it pins is now correct), but a future contributor could still "fix" a real filtering bug in a way that keeps the URL literal at `/finished-jobs` while breaking the semantics upstream/downstream, and this test would not catch it.
- **Best improvement:** As a nice-to-have follow-up, add a case to `RestApiHttpsTest` (or a `JobInfoService` unit test next to `JobInfoServiceNullSafetyTest`) that seeds one job per terminal state and asserts `GET /finished-jobs` returns all four, closer to the server boundary where the actual filtering logic lives.
- **Severity:** Low (downgraded from Medium last round — the risk this test carries is now much smaller since it protects the correct default)

### Issue 5 — CI does not build/lint/test the UI module for UI-only PRs (pre-existing, unaffected by this PR)
- **Location:** `.github/workflows/backend.yml` — `seatunnel-ui` job gated on the `api` path filter, which does not include `seatunnel-engine/seatunnel-engine-ui/**`
- **Problem description:** Same gap I found last round. A PR whose entire diff is TypeScript still does not get its lint/type-check/unit-test/build job triggered by CI.
- **Potential risk:** This specific PR's new test can merge without CI ever having run it.
- **Best improvement:** Separate small PR adding `seatunnel-engine/seatunnel-engine-ui/**` to the relevant path filter.
- **Severity:** Low

### Issue 6 — File-scoped `vi.mock` still replaces the whole service module, dropping `post`/`axios` exports
- **Location:** `seatunnel-engine/seatunnel-engine-ui/src/tests/jobs.spec.ts:32`
- **Problem description:** Unchanged from last round. `vi.mock('@/service/service', () => ({ get: vi.fn() }))` is hoisted file-wide and drops the real module's other exports (`post`, `axios`). Not a problem for the current three tests, but a latent trap for the next contributor who touches this file.
- **Potential risk:** Low today; confusing `is not a function` failure for a future test in this file.
- **Best improvement:** `vi.mock('@/service/service', async (importOriginal) => ({ ...(await importOriginal()), get: vi.fn() }))`, or scope with `vi.spyOn` instead of a file-wide mock.
- **Severity:** Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

Clean, idiomatic, no license-header or formatting concerns. The quote-style change (template literal → plain string) actually improves consistency with the sibling `'/running-jobs/summary'` call on the line above.

Mandatory documentation check: no new core method, public API, or nontrivial field was added — `getFinishedJobs` already existed and its signature is unchanged, and the new test is a boilerplate-style unit test. No missing-comment issue to raise here.

**Coverage.** One new unit test was added; see Issue 4 above for its scope limitation. It does cover the specific regression this PR's own history demonstrates is real (someone can accidentally narrow this URL), which is the most relevant thing to protect against.

**Mandatory test-stability analysis (Section 5.10.2), since this PR modifies UT code:**
- **Assertion-risk check:** The new test (`jobs.spec.ts:78-85`) is fully synchronous — `getMock.mockClear()` then a direct call then a direct assertion. No `Thread.sleep`/timers, no async waits, no dependency on wall-clock time, container startup, or execution order. `beforeEach` re-creates a fresh Pinia instance per test, and `getMock.mockClear()` resets the mock's call history before use, so it is not order-dependent on the two preceding tests in the same file.
- **Flaky-test anti-pattern check:** None of the documented anti-patterns apply — no shared static state left dirty across tests, no floating-point/string normalization concerns, no hard-coded ports/hostnames, no reliance on real network I/O (the service module is mocked).
- **Stability rating: Stable.** Evidence: `seatunnel-engine/seatunnel-engine-ui/src/tests/jobs.spec.ts:78-85` (new test), `:32` (module mock, module-scoped but deterministic).

## 2.3 Documentation Updates

No `docs/en`/`docs/zh` update is required by this PR — the response shape and REST contract are unchanged from `dev`, and the docs already describe `state` as optional (which is what "keep it unfiltered" is consistent with). The stale pagination example I flagged in my previous review (`docs/en/engines/zeta/rest-api-v2.md:654-663`, `docs/zh/.../rest-api-v2.md:631-640`) is a pre-existing `dev` issue, unrelated to and unaffected by this diff — not this PR's responsibility to fix.

# 3. Architectural Soundness

Given that #9868's root cause was never confirmed as still present, "revert to the known-correct URL and add a regression test" is a sound, minimal, low-risk resolution for the code itself. It is a **defensive fix** rather than a root-cause fix — the actual claimed defect (whatever caused the reporter's original 500) is still unaddressed in this PR, but no incorrect fix is being shipped on top of it either.

## 3.2 Maintainability

Neutral to positive versus `dev`: the URL literal is still duplicated between `service/job/index.ts:23` and `tests/jobs.spec.ts:84` with no shared constant (so both would need to change in lockstep if per-state filtering is ever added), but that duplication now protects a correct default rather than an incorrect one.

## 3.3 Extensibility

Unchanged assessment from my previous review: the server side (`getJobsByStateJson(state)` + the `/*` Jetty mapping) already supports per-state filtering for free. A future status-dropdown feature defaulting to "all states" and appending `/:state` only when chosen remains the natural extension path, and I'd be glad to review that if it's picked up separately.

Fully compatible. The bundled UI now requests exactly what it has requested since at least tag `2.3.12`, and the server-side empty-state semantics are unchanged and still covered by `JobInfoServiceNullSafetyTest`. No serialization, checkpoint, savepoint, or state-storage surface is involved.

# 4. Issue Summary

| 1 | PR title/description still describe the reverted (incorrect) URL change; will mislead the squash-merge commit message | PR title/description | Medium |
| 2 | #9868's root cause still not reproduced/verified against current `dev` | PR description; issue #9868 | Medium |
| 3 | Pager can strand the poll loop on an out-of-range page with no error shown (pre-existing in `dev`, not caused by this PR) | `views/jobs/finished-jobs.tsx:38-43`; `rest/servlet/PageBaseServlet.java:77-78` | Low |
| 4 | New test asserts a literal URL string rather than actual multi-state list behavior | `seatunnel-engine-ui/src/tests/jobs.spec.ts:78-85` | Low |
| 5 | CI's `seatunnel-ui` job path filter still excludes `seatunnel-engine-ui/**`, so this PR's own test never ran in CI | `.github/workflows/backend.yml` | Low |
| 6 | File-scoped `vi.mock` drops `post`/`axios` exports from the real service module | `seatunnel-engine-ui/src/tests/jobs.spec.ts:32` | Low |

**Resolved since my previous review (no longer listed as numbered issues, kept here for traceability):**
- Previous Issue 1 (Critical) — narrowing to `/finished-jobs/finished` dropped 4 of 5 terminal states — **fixed**, reverted to `/finished-jobs`.
- Previous Issue 3 (High) — failed/canceled jobs unreachable in the UI — **fixed**, automatic consequence of the revert above.
- Previous Issue 8's docs-contradiction half (docs said `state` optional, PR disagreed) — **fixed**, PR now agrees with the docs again.

# 5. Merge Recommendation

**Blockers — must be fixed:**
1. **Issue 1** — please update the PR title and description before merge so the squash-merge commit accurately reflects what the diff does (keeps the endpoint unfiltered + adds a regression test), rather than describing the already-reverted filtered-URL change. This is a one-minute edit on GitHub, no code change needed.

**Recommended fixes — non-blocking:**
2. **Issue 2** — please confirm (in a comment on #9868, doesn't have to block this PR) whether the original 500 still reproduces on `dev`. If not, close #9868 referencing #11269; if it does, it needs a separate server-side fix.
3. **Issue 3** — the pager-stranding bug is real but pre-existing and independent of this PR; worth its own follow-up.
4. **Issue 4** — a state-seeded assertion at the `RestApiHttpsTest`/`JobInfoService` level would give this contract stronger protection, but the current test is a legitimate, correct regression guard as-is.
5. **Issues 5, 6** — CI path-filter gap and test-mock hygiene; neither is introduced by this PR.

**Overall assessment.** This is exactly the outcome I was hoping to see after my last review: the critical regression (narrowing the endpoint and hiding failed/canceled jobs) is gone, and in its place is a small, correct, fully backward-compatible regression test that guards against it happening again. I traced the full diff against `dev` line by line and confirmed there is no remaining runtime behavior difference at all — the only functional risk left is the PR's own title/description being out of sync with its final commit, which I'd like fixed before this lands in permanent history, but that's a metadata fix, not a code fix. Once Issue 1 is addressed, I'm happy to approve. Thanks for iterating on this, @zhangshenghang — and thanks again to @goutamadwant for catching the original regression early.

DanielLeens (approved):
Following up on my last review (`d5181829`, 2026-08-20): the only remaining blocker I raised was Issue 1 — the PR title/description still described the earlier, reverted `/finished-jobs/finished` change instead of what the final commit actually does, which would have produced a misleading squash-merge commit message.

I checked the PR timeline: `@zhangshenghang` renamed the title from `[Fix][Zeta] Fix finished jobs UI endpoint` to `[Fix][Zeta] Add regression test for finished jobs endpoint` at 2026-08-20T02:35:10Z (22 minutes after my review), and the current description now correctly states the endpoint stays on the unfiltered `/finished-jobs` (matching `dev`) with a regression test added to protect it — exactly the fix I asked for.

No code has changed since my last review (head is still `d5181829691a06042842be7e9273db0e9c19e1a8`, so my prior end-to-end trace of the diff — full compatibility, zero runtime behavior change from `dev`, stable new unit test — still holds). This was a metadata-only blocker (PR title/description, not code), and it's now resolved. All other items I listed (#9868 root-cause reproduction, the pre-existing pager-stranding bug, the UI CI path-filter gap, test-mock hygiene) were explicitly non-blocking follow-ups, not conditions for merge.

Approving.

davidzollo (approved):
+1 LGTM

PR #11820, opened 2026-08-15
[Fix][Zeta] LogUtil.getLogPath uses wrong variable in fileAppender br…

### LINKED ISSUE #11463: [Bug] [Zeta] LogUtil.getLogPath uses wrong variable in fileAppender branch (NPE / StringIndexOutOfBounds)

tagadearpit:
## Summary

This pull request fixes a copy-paste bug in `LogUtil.getLogPath()` when the selected Log4j2 reference is `fileAppender`. The method now derives the parent directory from `fileLogPath` rather than the unrelated `routingLogFilePath`, which may be null in a fileAppender-only configuration.

A regression test and fileAppender-only Log4j2 fixture reproduce the failure and verify the corrected directory result.

Closes #11463

## Verification

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-engine/seatunnel-engine-common -DskipITs test`
- `./mvnw -pl seatunnel-engine/seatunnel-engine-common -DskipITs -Dtest=LogUtilTest -Dsurefire.failIfNoSpecifiedTests=false test`

The repository-wide `./mvnw -q -DskipTests verify` was attempted with Java 17; it reached `seatunnel-engine-ui` and was blocked by its `npm install` step returning exit code 1 from the npm registry. This is unrelated to the Java change.

This is an internal bugfix with no user-facing configuration or behavior change, so no `docs/en` or `docs/zh` updates are needed.


goutamadwant:
@tagadearpit Reviewed at `71489bd1` commit. 

The fix is correct. The old `fileAppender` branch called `substring()` on `fileLogPath` but calculated the end index from `routingLogFilePath`. With only `fileAppender` selected, which is the repository's default logging configuration, this throws a `NullPointerException`. Using `fileLogPath` for both values fixes the actual cause.

I confirmed this by restoring the previous line locally, which made the new test fail with the reported `NullPointerException`. On the PR head, the full `seatunnel-engine-common` suite passes with 17 tests, and the focused test passes on JDK 8, 11, and 17. I did not find a code issue in this change.

Build is yet to be triggered and needs to pass. Thanks!

DanielLeens (commented):
Thanks for the contribution, @tagadearpit, and welcome to Apache SeaTunnel! This is a small, well-targeted patch that fixes a genuine latent defect in the Zeta engine, and the PR description is unusually clear about what was verified. I reviewed it at head `71489bd1ca5e7f97e11319ddf796dac177ffced4` against base `b1e21631ffdfd8cb33e6adca618b2d95224a9d66`, source-level only (I did not run the code). Because this touches `seatunnel-engine`, I applied the strictest review bar we use for engine changes.

**One-sentence summary:** `LogUtil.getLogPath()` computed the log *directory* for the `fileAppender` case by slicing `fileLogPath` at an index taken from a completely different string (`routingLogFilePath`), so any deployment whose two appenders do not resolve to the same directory got an exception or a silently wrong directory; this PR makes the `fileAppender` branch derive its cut index from its own path.

## The branch under repair

`LogUtil.getLogPath()` resolves the active Log4j2 configuration, extracts two file paths by reflection, then picks a branch based on which appender the root logger actually references:

- `getRoutingLogFilePath(config)` walks three component levels under the appender named `routingAppender` and returns the nested route appender's `fileName`, or **`null`** via `.orElse(null)` when no `routingAppender` component exists.
- `getFileLogPath(config)` returns the `fileName` of the appender named `fileAppender`, or `null` when absent.
- `logRef` is the first root-logger appender ref containing `routingAppender` or `fileAppender`.

Three outcomes follow:

| `logRef` | Behavior before this PR |
| --- | --- |
| `routingAppender` | `routingLogFilePath.substring(0, routingLogFilePath.lastIndexOf("/"))` — self-consistent, correct |
| `fileAppender` | `fileLogPath.substring(0, routingLogFilePath.lastIndexOf("/"))` — **string from one appender, index from the other** |
| neither | `IllegalArgumentException` — intentional |

The diff is a single line at `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:52`:

```java
-            return fileLogPath.substring(0, routingLogFilePath.lastIndexOf("/"));
+            return fileLogPath.substring(0, fileLogPath.lastIndexOf("/"));
```

I read the before/after blob directly rather than trusting the description: **the reported variable swap is real**, and the replacement is the self-consistent form that the sibling `routingAppender` branch already uses.

## Concrete before/after

**Case A — `fileAppender` only (no `routingAppender` component at all).** A very common simplification when operators strip the routing appender out of `log4j2.properties`:

```properties
rootLogger.appenderRef.file.ref = fileAppender
appender.file.name     = fileAppender
appender.file.fileName = /opt/seatunnel/logs/seatunnel.log
```

- Before: `routingLogFilePath` is `null` → `null.lastIndexOf("/")` → **`NullPointerException`**.
- After: `/opt/seatunnel/logs`.

**Case B — both appenders present but pointing at different directories.** This is the quiet one, because nothing throws:

```properties
rootLogger.appenderRef.file.ref            = fileAppender
appender.routing.route.job.appender.fileName = /data/st/job-${ctx:ST-JID}.log
appender.file.fileName                      = /opt/seatunnel/logs/seatunnel.log
```

`"/data/st/job-….log".lastIndexOf("/")` is `8`, so the old code evaluated `"/opt/seatunnel/logs/seatunnel.log".substring(0, 8)`:

- Before: `/opt/sea` — a directory that does not exist. No exception, no warning.
- After: `/opt/seatunnel/logs`.

**Case C — routing path longer than the whole file path**, e.g. routing under `/very/long/mounted/volume/seatunnel/job-logs/` while `appender.file.fileName = /logs/st.log`: the index exceeds `fileLogPath.length()` → **`StringIndexOutOfBoundsException`**. Matches the third mode described in #11463.

## Why this survived in production for so long

I checked every Log4j2 configuration shipped in the repository, and the answer is precise and worth recording, because it defines exactly who is affected:

| Shipped config | Root-logger ref | Branch taken | Old behavior |
| --- | --- | --- | --- |
| `config/log4j2.properties` (default server config) | `fileAppender` | buggy branch | **accidentally correct** |
| `config/log4j2_client.properties` | console only (`fileAppender` commented out) | `IllegalArgumentException` | unaffected |
| `deploy/kubernetes/seatunnel/conf/log4j2.properties` | console only | `IllegalArgumentException` | unaffected |
| `seatunnel-engine-server/src/test/resources/log4j2-test.properties` | `routingAppender` | routing branch | unaffected |
| `seatunnel-engine-client/src/test/resources/log4j2-test.properties` | console only | `IllegalArgumentException` | unaffected |
| e2e `job-log-file/log4j2.properties` | `routingAppender` | routing branch | unaffected |

The default `config/log4j2.properties` does take the buggy branch, but both appenders are built from the same `property.file_path`:

```properties
appender.routing.route.job.appender.fileName = ${file_path}/job-${ctx:ST-JID}.log
appender.file.fileName                       = ${file_path}/${file_name}.log
```

Both strings therefore have their last `/` at the **same index**, so `routingLogFilePath.lastIndexOf("/")` and `fileLogPath.lastIndexOf("/")` are numerically identical and the wrong variable produced the right answer. That is why this compiles, why the out-of-box cluster works, and why no existing test caught it: **no shipped or test configuration selects `fileAppender` without a same-directory `routingAppender`.** The defect manifests only when a user edits the logging config — which is a supported, documented thing to do.

## Real-world blast radius

I traced every caller of `getLogPath()`. It is not display-only; the returned string is used as a filesystem path:

1. `BaseLogService.getLogPath()` (`seatunnel-engine-server/.../rest/service/BaseLogService.java:46-52`) catches **only** `NoSuchFieldException | IllegalAccessException`. `NullPointerException` and `StringIndexOutOfBoundsException` are not caught, so Case A/C escape the `return null` fallback entirely and the `StringUtils.isBlank(logPath) → send404()` guard in `RestHttpGetCommandProcessor.getAllNodeLog` is never reached. They surface as **HTTP 500** through the `catch (Throwable)` in `RestHttpGetCommandProcessor.handle`, or as a servlet-level 500 from `CurrentNodeLogServlet.doGet` / `AllNodeLogServlet.doGet`, which have no such catch.
2. `LogService.allLogName()` → `FileUtils.listFile(getLogPath())`; `LogService.currentNodeLog()` likewise. In Case B the wrong directory yields `null` → an **empty log list**, so the REST/UI log browser reports "no logs" on a node that is logging normally.
3. `RestHttpGetCommandProcessor.prepareLogResponse` builds `new File(logPath, logName)` and canonicalizes it against `logPath` as the containment boundary. With a wrong `logPath` the requested file is outside the (wrong) boundary or simply absent → **HTTP 400** on every log download.
4. `TaskLogManagerService` (`.../telemetry/log/TaskLogManagerService.java:39-49`) is the one that concerns me most. `initClean()` caches `LogUtil.getLogPath()` and `clean(jobId)` then does `logDir.list((dir, name) -> name.contains(jobId))` followed by `Files.delete(...)`. Two consequences of the old code:
   - Case A: `initClean()` swallows the NPE at `debug` level, `path` stays `null`, `clean()` returns immediately — **per-job log retention silently stops working** and job logs accumulate without bound. A silent disk-growth failure on a long-lived cluster is exactly the class of bug that is expensive to diagnose.
   - Case B: if the truncated prefix happens to be a real directory (`/var/log/seatunnel/seatunnel.log` truncated to `/var/log` is not a contrived example), `clean()` would enumerate **that** directory and delete any file whose name contains the job id. The window is narrow, but it is a delete-in-the-wrong-directory path, and that alone justifies fixing this properly rather than treating it as cosmetic.

So: not a display-only bug. Log REST endpoints 500/400/return-empty, and the log-retention cleaner either stops working or, in a narrow configuration, operates on the wrong directory.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The fix is the **minimal correct swap**, not a mask over a deeper problem. The `fileAppender` branch's contract is "given the fileAppender's file path, return its parent directory"; deriving both the string and the index from `fileLogPath` is the only self-consistent way to express that, and it is textually symmetric with the `routingAppender` branch two lines above. No control flow, no signature, no exception contract changed.

I specifically checked whether the wrong variable was doing any useful work that the fix removes — for example whether some configuration deliberately relied on the routing directory. It was not: `routingLogFilePath` is now referenced only inside the `routingAppender` branch, which is where it belongs, and the value it used to contribute was an index into an unrelated string.

Two follow-on observations, neither of which blocks the change:

- With the fix, `routingLogFilePath` is computed on **every** call even when `logRef` is `fileAppender` and the value is unused. That was true before too (it was used, wrongly), so this is pre-existing, and `getRoutingLogFilePath` is a reflection-plus-stream walk over an in-memory component tree, so the cost is negligible on a REST call path. Not worth churning the diff for.
- The `logRef` selection uses `.filter(ref -> ref.contains(...))` but the branches use `.equals(...)`. A ref named e.g. `myFileAppender` would satisfy `contains` and then fall through to the `IllegalArgumentException`. Pre-existing, out of scope here, flagged only so it is on the record.

## 1.2 Compatibility Impact

**Explicit conclusion: this change is backward compatible and safe to ship.**

- No `Option` added, removed, renamed, or re-defaulted; no public API or SPI signature touched. `getLogPath()` keeps its `throws NoSuchFieldException, IllegalAccessException`.
- No serialization, checkpoint, savepoint, state, or wire format is involved. `LogUtil` is a stateless utility invoked from REST handlers and a telemetry cleaner; nothing persists its result.
- For the **default shipped configuration the returned string is byte-for-byte identical** before and after, per the index analysis above. Out-of-box clusters see no change whatsoever on upgrade.
- For the affected custom configurations, the returned value changes from "an exception" or "a directory that does not exist" to "the correct directory". On the question of whether external tooling could have scraped the old value and would now break: I considered it and the answer is no in any realistic sense — the old value in Case A/C was never observable (an exception), and in Case B it was a path that could not be listed or read, so nothing downstream could have built a working dependency on it. There is no user-visible contract being changed, only a broken one being repaired.
- No entry in `docs/en/introduction/concepts/incompatible-changes.md` is warranted.

The production change is one `substring` argument; there is no allocation, iteration, or I/O difference. Nothing in a hot path — `getLogPath()` runs per REST request and once per `initClean()`.

The one real side effect in this PR is on the **test** side, not production: `seatunnel-engine-common` previously had **no** Log4j2 test configuration at all, so its tests ran under Log4j2's `DefaultConfiguration` (console, ERROR). Adding `src/test/resources/log4j2-test.properties` with a root logger bound solely to a `File` appender at `INFO` re-routes the log output of every other test class in the module (`EnvironmentUtilTest`, `YamlSeaTunnelConfigParserTest`, `YamlSeaTunnelDomConfigProcessorTest`, `ExceptionUtilTest`, `CompletableFutureTest`) into a temp file that CI discards. See Issue 2.

I did confirm the containment boundary: `seatunnel-engine-common/pom.xml` does not attach a `test-jar`, so this test configuration cannot leak onto any other module's test classpath. Blast radius is exactly this one module.

I also checked neighboring branches for a sibling copy-paste defect, since that was the explicit failure mode here. The `routingAppender` branch is self-consistent and has **no** analogous swap. It does share a different latent fragility with the fixed branch — see Issue 3.

No logging is added or changed, which is right for a one-line correctness fix. The pre-existing `IllegalArgumentException` for the "neither appender" case is untouched, and `TaskLogManagerService.initClean()` still catches it specifically to avoid stack-trace noise in local/example runs.

Worth noting for reviewers: because `BaseLogService.getLogPath()` narrows its catch to the two reflection exceptions, this fix also removes the *only* realistic source of unchecked exceptions from that call site. After this patch, the `IllegalArgumentException` path is the only remaining failure mode, and it is handled at every caller. That is a real robustness improvement beyond the literal path correction, and it is worth one line in the PR description.

# 2. Code Quality Assessment

## 2.1 Coding Standards

- ASF license headers present and correct on both new files.
- Formatting is consistent with Spotless/AOSP; the properties file ends with a newline; no wildcard imports; the test uses a static import of `assertEquals`, matching repo convention.
- No `@DisplayName` (good — this project prefers plain method names).
- The test's `throws NoSuchFieldException, IllegalAccessException` is **required**, not leftover, because `LogUtil.getLogPath()` declares them. Correct as written.
- Commit title follows `[Fix][Zeta] …`. Note the title is truncated at the end (`… in fileAppender br…`); if you get a chance, please expand the PR title to the full sentence so it reads cleanly in the release notes.

Adding a regression test at all — with a fixture that genuinely reproduces the reported failure — is the right instinct, and it is more than many one-line fixes arrive with. I verified the test really does discriminate: under the old code, `log4j2-test.properties` defines no `routingAppender`, so `getRoutingLogFilePath` returns `null` and the old line throws `NullPointerException`. The test fails before and passes after. That is a true regression test for Case A.

On portability, I worked through the assertion `assertEquals(System.getProperty("java.io.tmpdir"), LogUtil.getLogPath())` across separator conventions, because the unit-test matrix in `.github/workflows/backend.yml` runs on **both `ubuntu-latest` and `windows-latest`** (Java 8 and 11), so this test will execute on Windows:

- Linux (`/tmp`, no trailing separator): `fileName` = `/tmp/seatunnel-logutil-test.log`, last `/` at 4, result `/tmp`. Passes.
- macOS (`/var/folders/…/T/`, trailing `/`): the config's literal `/` produces a doubled separator, the last `/` is the literal one, result reproduces the trailing slash. Passes.
- Windows (`C:\…\Temp\`, trailing `\`): `lastIndexOf("/")` finds the config's literal forward slash at exactly `tmpdir.length()`, so the result is `tmpdir` including its trailing backslash. Passes.

So I do **not** expect a cross-platform failure here. I want that on the record because it is the kind of thing that looks fragile and is easy to flag incorrectly.

What is genuinely missing is coverage breadth (Issue 1): only the NPE mode is covered. The two *silent* modes — the wrong-directory truncation of Case B and the `StringIndexOutOfBoundsException` of Case C — are precisely the ones that reach `TaskLogManagerService.clean()`, and neither is asserted. The structural reason is that all branch logic is welded to global Log4j2 state, so each additional case would need its own `LoggerContext`, which is not practical inside one surefire JVM. A tiny package-private seam solves that; see Issue 1.

Stability: the test depends on the process-wide `LoggerContext` and on the unchecked cast `(PropertiesConfiguration) context.getConfiguration()` inside `getLogConfiguration()`. If any test in this module ever reconfigures Log4j2, or if a future `log4j2.properties`/XML file wins the lookup, this fails with a `ClassCastException` rather than a readable assertion. Not a problem today (nothing in the module touches `LogManager`), but it is why I would rather see the branch logic tested through a pure seam.

## 2.3 Documentation Updates

No documentation change is required, and I agree with the author's assessment. `docs/en/engines/zeta/rest-api-v1.md` and `rest-api-v2.md` document the `/logs`, `/log`, and log-name endpoints but do not document *how* the log directory is derived, so there is no doc statement to correct. This restores the documented endpoint behavior rather than altering it. No `docs/zh` counterpart change either.

# 3. Architectural Soundness

Fixing the branch to be self-consistent is the correct, minimal expression of intent — I checked that a more "elegant" refactor (e.g. `new File(path).getParent()`, or a `Paths.get(...).getParent()`) would actually be a behavior change: `getParent()` normalizes separators and strips trailing slashes, which would alter the returned string on Windows and for trailing-slash inputs, and would change the currently-identical output for the default config. Keeping `lastIndexOf("/")` is the right call for a bugfix PR. Good judgment on scope.

## 3.2 Maintainability

The one structural weakness — and it is what allowed the original copy-paste to hide — is that `getLogPath()` fuses three responsibilities: reflectively reading Log4j2 internals, selecting the appender ref, and doing pure string arithmetic. The pure part is trivially testable; the rest is not. Splitting them (Issue 1) would both close the coverage gap and make a future recurrence of this exact class of bug impossible to miss.

## 3.3 Extensibility

Unchanged. Adding a third appender kind would still mean another `else if` with its own path variable. Given how rarely this evolves, that is acceptable; I am not asking for a strategy abstraction here.

No upgrade, restore, or historical-state path is involved. `LogUtil` holds no state, participates in no checkpoint or savepoint, and is not serialized. A cluster upgraded from any prior version with an unmodified `config/log4j2.properties` resolves the identical directory before and after. A cluster with a customized `fileAppender`-only configuration stops erroring — which is the intended repair, not a regression. **No historical-version compatibility risk.**

---

## Issue 1

- **Location:** `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java:24-29` (and the structure of `LogUtil.java:34-57`)
- **Problem description:** The regression test covers only the `NullPointerException` mode (no `routingAppender` configured). The other two modes reported in #11463 — a *silently wrong* truncated directory when both appenders resolve to different-length directories, and `StringIndexOutOfBoundsException` when the routing path's last-separator index exceeds `fileLogPath.length()` — have no assertion. Because all branch logic is coupled to the process-wide `LoggerContext`, additional cases cannot practically be added as-is: each would need its own Log4j2 configuration in the same surefire JVM.
- **Potential risk:** The silent-truncation mode is the dangerous one (it feeds `TaskLogManagerService.clean()`, which deletes files under the resolved directory) and it is exactly the mode with zero coverage. A future refactor of this method could reintroduce cross-variable indexing and the suite would stay green.
- **Best improvement:** Extract the pure logic behind a package-private seam and test all branches directly, keeping the existing config-driven test as the one end-to-end case:

  ```java
  // package-private for test: pure branch logic, no Log4j2 state involved
  static String resolveLogDir(String logRef, String routingLogFilePath, String fileLogPath) { ... }
  ```

  Then assert: (a) `fileAppender` with `routingLogFilePath == null` returns the fileAppender's parent; (b) `fileAppender` with a *shorter* routing directory still returns the fileAppender's parent (this is the case that fails under the old code without throwing); (c) `fileAppender` with a routing path longer than `fileLogPath` returns correctly instead of throwing; (d) `routingAppender` is unaffected; (e) an unknown ref still raises `IllegalArgumentException`. Case (b) is the assertion I would most like to see, since it is the one that fails *silently* today.
- **Severity:** Medium

## Issue 2

- **Location:** `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-test.properties:18-25`
- **Problem description:** This module had no Log4j2 test configuration before, so its tests ran under `DefaultConfiguration` (console). The new file binds the root logger to a single `File` appender at `INFO`, so **all** log output from the module's other five test classes now goes to a temp file instead of the CI console.
- **Potential risk:** When an unrelated test in `seatunnel-engine-common` fails in CI, its diagnostic output is no longer in the job log — it is in a temp file that the runner discards. That is a small but real reduction in CI debuggability, and it is an invisible side effect of a PR that is nominally about one line in `LogUtil`.
- **Best improvement:** Add console appenders alongside the file appender and reference all three from the root logger, mirroring `seatunnel-engine-server/src/test/resources/log4j2-test.properties`. I confirmed this is safe for the new test: `getLogPath()` filters root-logger refs with `ref.contains("routingAppender") || ref.contains("fileAppender")`, so `consoleStdoutAppender` / `consoleStderrAppender` refs are ignored and `logRef` still resolves to `fileAppender`.
- **Severity:** Low

## Issue 3

- **Location:** `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:50` and `:52`
- **Problem description:** Both branches call `substring(0, path.lastIndexOf("/"))` with no guard for `lastIndexOf` returning `-1`, and neither recognizes the Windows separator `\`. A configuration with a relative `appender.file.fileName = seatunnel.log`, or a Windows-style absolute `C:\seatunnel\logs\seatunnel.log`, yields `-1` and throws `StringIndexOutOfBoundsException` — which, as established in section 1, is **not** caught by `BaseLogService.getLogPath()` and reaches the REST layer as a 500.
- **Potential risk:** Same class of failure this PR is fixing, reached by a different route. This is strictly pre-existing and equally present in the `routingAppender` branch, so it is **not** a defect in this change — I raise it because the fix touches this exact expression and because the natural next question after "was there a sibling copy-paste bug nearby?" is "does the sibling share the same fragility?" (it does; it just does not have the swapped variable).
- **Best improvement:** Optional and entirely at the author's discretion. If you would like to fold it in, a two-line guard — resolve the index as `Math.max(path.lastIndexOf('/'), path.lastIndexOf('\\'))` and throw the existing `IllegalArgumentException` with the offending path when it is `-1` — turns an opaque 500 into an actionable configuration error. If you would rather keep this PR to the single-line fix, that is a perfectly good decision and this can be a follow-up issue; please just say which you prefer.
- **Severity:** Low

## Issue 4

- **Location:** `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-test.properties:22`
- **Problem description:** The fixture writes to a fixed filename `${sys:java.io.tmpdir}/seatunnel-logutil-test.log` that is never cleaned up, and the same name is used by every matrix leg (Java 8/11 x ubuntu/windows).
- **Potential risk:** Minor. Legs run on separate runners so collision is unlikely, but a leftover file accumulates on developer machines and two concurrent local builds of this module would share the handle.
- **Best improvement:** Nit only — either accept it, or disambiguate the name (for example by appending `${sys:user.name}`) so concurrent local runs cannot contend.
- **Severity:** Nit

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | Regression test covers only the NPE mode; the silent wrong-directory and `StringIndexOutOfBoundsException` modes are untested, and the branch logic has no testable seam | `seatunnel-engine-common/src/test/java/.../LogUtilTest.java:24-29` | Medium |
| 2 | New module-wide `log4j2-test.properties` redirects all other tests' output from console to a discarded temp file, reducing CI diagnosability | `seatunnel-engine-common/src/test/resources/log4j2-test.properties:18-25` | Low |
| 3 | Pre-existing (both branches): no `-1` guard on `lastIndexOf("/")` and no Windows `\` handling, producing an uncaught `StringIndexOutOfBoundsException` → HTTP 500 | `seatunnel-engine-common/src/main/java/.../LogUtil.java:50,52` | Low |
| 4 | Test fixture writes a fixed, never-cleaned filename in the shared temp directory | `seatunnel-engine-common/src/test/resources/log4j2-test.properties:22` | Nit |

# 5. Merge Recommendation

**Conclusion: Not yet ready to merge — the one-line production fix is correct, minimal, and backward compatible, but this PR has never had a successful CI run (GitHub Actions are not enabled on the author's fork), and the regression test covers only one of the three failure modes described in #11463.**

**Blockers**

1. **CI has not run.** The `Build` check on this head is `action_required`, not a test failure: *"Unable to detect the workflow run for testing the changes in your PR."* (https://github.com/apache/seatunnel/runs/95056497612). The author's fork currently reports zero Actions runs, which matches item 1 of that message. Please enable GitHub Actions on your fork (Actions tab → enable workflows), then push an empty commit to trigger a run. Nothing can be merged in this project without a green build, so this is the first thing to clear. **No source-level blocker exists in this change** — this gate is purely about getting CI to execute.

**Recommended (non-blocking)**

2. Issue 1 — add the wrong-directory case (Case B in my analysis). I would consider this close to required in spirit rather than in process, because it is the only silent failure mode and the one with a delete-path consequence. The package-private seam makes it a few lines.
3. Issue 2 — restore console output for the module's other tests.
4. Issues 3 and 4 — author's discretion; a follow-up issue is a fine home for Issue 3.

**Overall assessment**

This is a good first contribution: the diagnosis in #11463 is accurate, the fix is the correct minimal repair rather than a workaround, the scope is disciplined (the author explicitly resisted a broader refactor), and the compatibility reasoning in the PR description holds up under scrutiny. I independently confirmed the two claims that mattered most — that the swapped variable is real at `LogUtil.java:52`, and that the default shipped `config/log4j2.properties` masks it because both appenders resolve to the same `${file_path}` directory. That second point is the reason this defect went unnoticed, and it is also why the change is safe to ship: out-of-box deployments see a byte-identical result, while operators who customized their logging configuration get working `/logs`, `/log`, and log-name endpoints and a functioning per-job log cleaner instead of a 500 or an empty list.

Once CI is green and Issue 1 is addressed, I would be happy to approve. Thanks again for picking up a real bug and writing a reproducing test for it — that is exactly the pattern we want to see.

davidzollo:
Good job.

To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration  https://github.com/apache/seatunnel/pull/11820/checks?check_run_id=95056497612 

image

davidzollo:

```

```

tagadearpit:
Thanks @DanielLeens, @goutamadwant, and @davidzollo for the careful review and for independently confirming the correctness of the variable swap.

The CI gate is understood. The fork’s Build check is currently `action_required` because workflows have not run there yet. I will enable GitHub Actions for `tagadearpit/seatunnel` and trigger a fresh run before requesting merge.

Regarding the additional coverage, I agree that the silent wrong-directory case is the most valuable follow-up. This PR intentionally stays limited to the one-line production correction and the fileAppender-only regression that reproduces the reported NPE. A small package-private path-resolution seam with focused cases for the null, silent-truncation, overlong-index, routing-appender, and unknown-reference paths would make that broader coverage reliable; I will keep that as a follow-up unless maintainers prefer it in this PR.

I also appreciate the note about the test fixture’s module-wide logging effect. I will preserve console diagnostics when expanding the fixture, and the fixed temporary filename and separator/`-1` handling concerns are noted as separate cleanup items rather than folded into this narrowly scoped bugfix.

No documentation update is needed for this internal correction. Thanks again for the positive review and for pointing out the remaining verification and coverage work.

DanielLeens (commented):
Thanks for the ping, @davidzollo, and thanks @tagadearpit for the thoughtful reply. I re-reviewed the current head from scratch (not just the incremental diff) since a new commit landed after my last pass.

**What changed since my last review:** commit `a840431c` is a merge of `dev` into `fix/logutil-fileappender-path` (picking up ~10 unrelated upstream commits) with no conflicts. I independently diffed the three files under review between the old head `71489bd1` and the current head `a840431c` (`git diff 71489bd1..a840431c -- LogUtil.java LogUtilTest.java log4j2-test.properties`) and confirmed the result is **empty** — the production fix, the test, and the fixture are byte-identical to what I reviewed before. I also re-read `LogUtil.java` in full at the new head and re-walked the caller chain (`BaseLogService`, `TaskLogManagerService`, `RestHttpGetCommandProcessor`, `AllNodeLogServlet`, `CurrentNodeLogServlet`, `AllLogNameServlet`, `LogService`) to independently re-verify my prior findings rather than assume they still hold. All of it checks out unchanged; nothing below is new except the CI status update and the note on the discussion.

`LogUtil.getLogPath()`'s `fileAppender` branch computed the log directory by slicing `fileLogPath` (the right string) at an index taken from `routingLogFilePath` (the wrong appender's path). `LogUtil.java:51` now derives both the string and the index from `fileLogPath`, matching the pattern already used one line above in the `routingAppender` branch. Confirmed identical at the current head — see diff above.

# 1. Code Change Review

## 1.1 Core Logic Analysis
Unchanged from my prior review: this is the minimal, self-consistent fix. `routingLogFilePath` is still computed unconditionally (pre-existing, cheap, out of scope) and the `.contains(...)`/`.equals(...)` mismatch in `logRef` selection is still present (pre-existing, out of scope, noted for the record only).

## 1.2 Compatibility Impact
Unchanged conclusion: backward compatible. No `Option`, API, SPI, serialization, checkpoint/savepoint, or wire-format surface touched. Default shipped `config/log4j2.properties` produces a byte-identical result before and after because both appenders share `${file_path}`.

Unchanged: one `substring` argument, no hot-path impact. The `log4j2-test.properties` fixture still re-routes `seatunnel-engine-common`'s other test classes' console output to a discarded temp file (Issue 2, still open, still non-blocking).

Unchanged: no logging touched, `IllegalArgumentException` path unaffected.

**CI status (re-verified against the current head, correcting the framing I was given for this pass):** the "Build" check is `COMPLETED`/`ACTION_REQUIRED`, not a test *failure* — same state as at my last review. I checked `repos/tagadearpit/seatunnel/actions/runs` directly and it returns zero workflow runs, confirming GitHub Actions is still not enabled on the contributor's fork. `@tagadearpit` said above ("I will enable GitHub Actions for `tagadearpit/seatunnel` and trigger a fresh run before requesting merge") that this is in progress — as of this review it has not happened yet. This is environmental/configuration, not caused by this PR's diff, and it remains the sole hard blocker.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Unchanged: license headers, formatting, no `@DisplayName`, all correct.

Unchanged from my prior analysis: the regression test correctly reproduces and fixes the `NullPointerException` mode (Case A). The silent wrong-directory truncation (Case B — the one that reaches `TaskLogManagerService.clean()`'s delete path) and the `StringIndexOutOfBoundsException` mode (Case C) remain uncovered. `@tagadearpit`'s reply proposes exactly the package-private seam I suggested (null / silent-truncation / overlong-index / routing-appender / unknown-ref cases) as a deliberate follow-up rather than folding it into this narrowly-scoped bugfix. That is a reasonable scoping call for a first PR — I flagged Issue 1 as Medium/recommended, not blocking, in my original review, and I'm holding that line: it's a legitimate "ship the minimal correct fix now, harden coverage next" sequencing, provided the follow-up actually happens. I'd ask that a tracking issue be opened (or #11463 be kept open) so the wrong-directory case doesn't get lost once this merges.

## 2.3 Documentation Updates
Unchanged: no doc update needed, and the author's own assessment (no doc statement existed to correct) still holds under my re-check of `rest-api-v1.md`/`rest-api-v2.md`.

# 3. Architectural Soundness

Unchanged: keeping `lastIndexOf("/")` instead of switching to `File.getParent()`/`Paths.get(...).getParent()` was the right call to avoid a silent behavior change on Windows/trailing-separator inputs. Good scope discipline, reaffirmed by the author's reply explicitly declining to fold in the separator/`-1` guard (Issue 3) as a "separate cleanup item."

## 3.2 Maintainability
Unchanged: `getLogPath()` still fuses reflection, ref-selection, and string arithmetic in one method, which is exactly what let the original copy-paste hide. The proposed seam (Issue 1) is the right long-term fix for this.

## 3.3 Extensibility
Unchanged: acceptable as-is for two appender kinds.

Unchanged: `LogUtil` is stateless, not part of any checkpoint/savepoint/serialization path. No historical-version risk.

# 4. Issue Summary

| # | Issue | Location | Severity | New this round? |
| --- | --- | --- | --- | --- |
| 1 | Regression test covers only the NPE mode; silent wrong-directory (Case B) and `StringIndexOutOfBoundsException` (Case C) modes remain untested; author has committed to a follow-up seam rather than folding it into this PR | `LogUtilTest.java:24-29` | Medium | Carryover — status updated (author response noted, no code change) |
| 2 | Module-wide `log4j2-test.properties` redirects other test classes' console output to a discarded temp file | `log4j2-test.properties:18-25` | Low | Carryover — unchanged |
| 3 | Pre-existing (both branches): no `-1` guard on `lastIndexOf("/")`, no Windows `\` handling | `LogUtil.java:50,51` | Low | Carryover — unchanged, author declined to fold in |
| 4 | Fixed, never-cleaned test fixture filename in shared temp dir | `log4j2-test.properties:22` | Nit | Carryover — unchanged |

No new issues found on this full re-review of the current head.

# 5. Merge Recommendation

### Conclusion: Not yet ready to merge — the source-level fix remains correct and backward compatible, and the merge-from-dev commit introduced no functional change, but CI still cannot run because GitHub Actions is not enabled on `tagadearpit/seatunnel`.

**Blockers**
1. CI still has not executed. `Build` is `ACTION_REQUIRED` and `repos/tagadearpit/seatunnel/actions/runs` returns zero runs — Actions has not yet been enabled on the fork, despite the intent stated in the 2026-08-16T14:59:50Z comment. This remains the only blocker; there is no source-level blocker in the reviewed diff.

**Recommended (non-blocking)**
2. Issue 1 — please open a tracking issue (or keep #11463 open) for the wrong-directory/overlong-index coverage gap so it isn't forgotten after merge; I'd still like to see it land reasonably soon given it's the one silent, delete-adjacent failure mode.
3. Issue 2 — restore console appenders in the new `log4j2-test.properties` alongside the file appender.
4. Issues 3 and 4 — author's discretion, fine as standalone follow-ups.

**Overall assessment:** unchanged from my prior review — this is a good, disciplined first contribution with an accurate diagnosis and a minimal correct fix. The only thing standing between this and merge is getting the fork's Actions enabled so CI can actually run; once that's green I'm ready to approve.

DanielLeens (commented):
Thanks for pushing the trigger commit, @tagadearpit. I re-reviewed the current head from scratch (not just the incremental diff), since a new commit landed after my last pass.

**What changed since my last review:** commit `fe1a71c6eb` is an empty "chore" commit whose only purpose is to trigger a build now that GitHub Actions has been enabled on `tagadearpit/seatunnel`. I diffed the three files under review between the previously reviewed head `a840431c` and the current head `fe1a71c6eb` and confirmed it is **empty** — the production fix, the test, and the fixture are byte-identical to what I already reviewed twice. So this round is really a CI-status update, not a code re-review, but per process I re-walked `LogUtil.java` end-to-end again anyway (`getLogPath`, `getRoutingLogFilePath`, `getFileLogPath`, and the caller chain through `BaseLogService` / `TaskLogManagerService` / the REST log servlets) to independently re-confirm my prior findings still hold. They do.

The important update: CI has now actually run for the first time (previously it was `ACTION_REQUIRED` with zero runs on the fork). It came back **red**, so I dug into it rather than taking that at face value.

`LogUtil.getLogPath()`'s `fileAppender` branch computed the log directory by slicing `fileLogPath` (the right string) at an index taken from `routingLogFilePath` (the wrong appender's path). In a `fileAppender`-only configuration — which is the shipped default — `routingLogFilePath` is `null`, so `routingLogFilePath.lastIndexOf("/")` throws `NullPointerException`. `LogUtil.java:52` now derives both the string and the index from `fileLogPath`, matching the pattern already used one line above in the `routingAppender` branch (`LogUtil.java:50`).

# 1. Code Change Review

## 1.1 Core Logic Analysis
Unchanged from my prior review: this is the minimal, self-consistent fix. `routingLogFilePath` is still computed unconditionally even when only `fileAppender` is active (pre-existing, cheap, out of scope), and the `.contains(...)` filter vs `.equals(...)` comparison mismatch in the `logRef` selection at `LogUtil.java:46-51` is still present (pre-existing, out of scope, noted for the record only).

## 1.2 Compatibility Impact
Unchanged conclusion: backward compatible. No `Option`, API, SPI, serialization, checkpoint/savepoint, or wire-format surface touched. The shipped `config/log4j2.properties` produces a byte-identical result before and after, since both appenders share `${file_path}`.

Unchanged: one `substring` argument change, no hot-path impact. The `log4j2-test.properties` fixture still re-routes other test classes' console output in the module to a discarded temp file (Issue 2 below, still open, still non-blocking).

Unchanged: no logging touched, the `IllegalArgumentException` fallback path is unaffected.

**CI status (re-verified against the current head):** the "Build" check is `COMPLETED`/`FAILURE`, a real change from `ACTION_REQUIRED`. I pulled the job list for the fork's run (`tagadearpit/seatunnel` run `32067465721`) — 87 jobs, exactly one failure: `unit-test (8, windows-latest)`. Pulling that job's log, the failure is:

```
java.lang.IllegalStateException: Node failed to start!
  at org.apache.seatunnel.engine.server.resourcemanager.ResourceManagerTest.before(ResourceManagerTest.java:55)
  at ...AbstractSeaTunnelServerTest.before:70
Tests run: 338, Failures: 0, Errors: 1, Skipped: 20 (elapsed 311.262s on that one test)
```

This is `seatunnel-engine-server`'s `ResourceManagerTest` failing to bring up a Hazelcast node during `@BeforeEach` on `windows-latest` — nothing in this PR's diff touches `seatunnel-engine-server` at all (the PR only touches `seatunnel-engine-common`), and a Hazelcast-node-startup timeout on `windows-latest` is a known, previously-observed environmental flake in this repo unrelated to any specific PR's content. The PR is also fully caught up with `dev` (`git rev-list --left-right --count origin/dev...HEAD` → `0  4`, i.e. 0 behind), so a branch sync would not help here — this needs a rerun of the failing job, not a code or branch change. Every other check (Dead links, Code style, License header, Sanity check, Dependency licenses, all the `*-it-*` integration matrices, and `unit-test` on `ubuntu-latest`/other JDKs) is green.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Unchanged: license headers, formatting, no `@DisplayName`, all correct.

Unchanged from my prior analysis: the regression test (`LogUtilTest.java:26-30`) correctly reproduces and fixes the `NullPointerException` mode (Case A). The silent wrong-directory truncation (Case B — the one that reaches `TaskLogManagerService.clean()`'s delete path) and the `StringIndexOutOfBoundsException` mode (Case C) remain uncovered. As discussed in the prior round, `@tagadearpit` has agreed this is a deliberate follow-up rather than being folded into this narrowly-scoped bugfix, which I still consider a reasonable scoping call for a first PR. I'd still like to see a tracking issue opened (or #11463 kept open) so it isn't lost after merge.

Stability rating for the new/changed test code itself: **Stable**. `LogUtilTest.java` is a single deterministic assertion against `System.getProperty("java.io.tmpdir")` with no sleeps, no shared mutable state, and no timing dependency; `log4j2-test.properties` is a static fixture. The Windows CI failure above is in unrelated, pre-existing test code (`ResourceManagerTest`), not in anything this PR added or changed.

## 2.3 Documentation Updates
Unchanged: no doc update needed; this is an internal bugfix with no user-facing config or behavior change.

# 3. Architectural Soundness

Unchanged: keeping `lastIndexOf("/")` instead of switching to `File.getParent()`/`Paths.get(...).getParent()` avoids a silent behavior change on Windows/trailing-separator inputs. Good scope discipline.

## 3.2 Maintainability
Unchanged: `getLogPath()` still fuses reflection, ref-selection, and string arithmetic in one method, which is exactly what let the original copy-paste bug hide. A package-private path-resolution seam (as discussed with the author) remains the right long-term fix, but is reasonably deferred.

## 3.3 Extensibility
Unchanged: acceptable as-is for two appender kinds.

Unchanged: `LogUtil` is stateless and not part of any checkpoint/savepoint/serialization path. No historical-version risk.

# 4. Issue Summary

| # | Issue | Location | Severity | New this round? |
| --- | --- | --- | --- | --- |
| 1 | Regression test covers only the NPE mode; silent wrong-directory (Case B) and `StringIndexOutOfBoundsException` (Case C) modes remain untested; author has committed to a follow-up seam | `LogUtilTest.java:26-30` | Medium | Carryover — unchanged |
| 2 | Module-wide `log4j2-test.properties` redirects other test classes' console output to a discarded temp file | `log4j2-test.properties:18-25` | Low | Carryover — unchanged |
| 3 | Pre-existing (both branches): no `-1` guard on `lastIndexOf("/")`, no Windows `\` handling | `LogUtil.java:50,52` | Low | Carryover — unchanged, author declined to fold in |
| 4 | Fixed, never-cleaned test fixture filename in shared temp dir | `log4j2-test.properties:22` | Nit | Carryover — unchanged |
| 5 | CI `unit-test (8, windows-latest)` job failed with `IllegalStateException: Node failed to start!` in `ResourceManagerTest`, unrelated to this PR's `seatunnel-engine-common`-only diff | fork run `32067465721`, job `unit-test (8, windows-latest)` | Blocking (environmental, not source) | New this round — first real CI run |

No new source-level issues found on this full re-review of the current head.

# 5. Merge Recommendation

**Blockers**
1. The only failing CI job (`unit-test (8, windows-latest)`) failed on a pre-existing, unrelated `ResourceManagerTest` Hazelcast-node-startup timeout in `seatunnel-engine-server`, a module this PR does not touch. The branch is already fully caught up with `dev` (0 commits behind), so this is not a sync issue — it needs the failing job re-run. Please re-run the failed job (or the workflow) on your fork; if it stays red on a second attempt with the same `ResourceManagerTest` failure, flag it here and we'll treat it as a separate, pre-existing flake rather than something this PR needs to fix.

**Recommended (non-blocking)**
2. Issue 1 — please open a tracking issue (or keep #11463 open) for the wrong-directory/overlong-index coverage gap so it isn't forgotten after merge.
3. Issue 2 — restore console appenders in the new `log4j2-test.properties` alongside the file appender.
4. Issues 3 and 4 — author's discretion, fine as standalone follow-ups.

**Overall assessment:** unchanged from my prior reviews — this is a good, disciplined first contribution with an accurate diagnosis and a minimal, correct, backward-compatible fix. There is no source-level blocker anywhere in this diff. The one thing standing between this and merge is getting a clean CI run, and the current red is a known class of environmental Windows/Hazelcast flake, not something introduced by this change. Once the job comes back green (or is confirmed as a pre-existing flake unrelated to this diff), I'm ready to approve.

DanielLeens (commented):
Thanks, @tagadearpit. A new commit landed after my last pass, so per process I re-reviewed the current head from scratch rather than just the incremental diff.

**What changed since my last review:** commit `94004ec0a3` is another "Merge branch 'dev' into fix/logutil-fileappender-path" (5 commits pulled in, none touching this PR's files: the Error Data Bypass / DLQ feature, Databend/IoTDBv2 OptionRule validation, the Couchbase container-retry fix, a state-store refactor in `seatunnel-engine-server`, and an unrelated UI test). I confirmed this the reliable way — `gh api repos/apache/seatunnel/compare/fe1a71c6eb...94004ec0a3` — rather than trusting a local clone: none of `LogUtil.java`, `LogUtilTest.java`, or `log4j2-test.properties` appear in that file list. I also pulled the three blobs directly from the current head's tree (`git/trees/94004ec0a3...?recursive=1` → `git/blobs/<sha>`) and read them end-to-end: byte-identical to what I reviewed at `fe1a71c6eb` and `a840431c` before that. I independently re-walked `LogUtil.java` and the caller chain (`BaseLogService.getLogPath()` at line 48-49, `TaskLogManagerService.initClean()`/`clean()` at lines 36-55) in the base tree to re-confirm my prior findings rather than assume they still hold. They do — nothing below is new except the CI status.

`LogUtil.getLogPath()`'s `fileAppender` branch computed the log directory by slicing `fileLogPath` (the correct string) at an index taken from `routingLogFilePath` (a different appender's path). In a `fileAppender`-only configuration — the shipped default once the routing appender is stripped out — `routingLogFilePath` is `null`, so `routingLogFilePath.lastIndexOf("/")` threw `NullPointerException`; with both appenders present but pointing at different directories, the wrong index silently truncated `fileLogPath` into a directory that doesn't exist. The fix makes both the string and the index come from `fileLogPath`, matching the pattern already used one line above in the `routingAppender` branch.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Confirmed at the current head (`seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:50-52`):

```java
    return fileLogPath.substring(0, fileLogPath.lastIndexOf("/"));   // was: routingLogFilePath.lastIndexOf("/")
}
```

Before the fix, the `fileAppender` branch read `fileLogPath.substring(0, routingLogFilePath.lastIndexOf("/"))` — the source string and the cut index came from two unrelated appenders. After the fix, both come from `fileLogPath`, which is the only self-consistent way to express "parent directory of the fileAppender's own file." This is the minimal correct repair, not a mask over something deeper: I checked whether `routingLogFilePath` was doing any useful work the fix removes, and it was not — it is now used only inside the `routingAppender` branch, where it belongs.

Tracing the result forward: `getLogPath()` → `BaseLogService.getLogPath()`/`LogService.allLogName()`/`LogService.currentNodeLog()` treat the return value as a filesystem directory (`FileUtils.listFile(...)`, `new File(logPath, logName)` with `logPath` as the containment boundary in `RestHttpGetCommandProcessor.prepareLogResponse`), and `TaskLogManagerService.clean(jobId)` lists and deletes files under it. Before the fix, a `fileAppender`-only config hit an uncaught `NullPointerException` here (Case A); a two-appender config with differently-named directories got a directory that does not exist (Case B, silent). After the fix both resolve to the real log directory. Two pre-existing, out-of-scope observations carried over unchanged from my prior passes: `routingLogFilePath` is still computed unconditionally even when unused (cheap, no functional effect), and `logRef` selection uses `.filter(ref -> ref.contains(...))` but the branches compare with `.equals(...)`, so a hypothetical ref like `myFileAppender` would fall through to the `IllegalArgumentException` — pre-existing, not touched by this diff.

I re-checked for a sibling copy-paste of the *same* swapped-variable mistake elsewhere in the module (`TaskLogManagerService.java`, `BaseLogService.java`) — there is none; both simply call `LogUtil.getLogPath()` and handle its result/exceptions as described below.

## 1.2 Compatibility Impact

Unchanged conclusion: backward compatible. No `Option`, public API, or SPI signature touched; `getLogPath()` keeps its `throws NoSuchFieldException, IllegalAccessException`. Nothing here participates in serialization, checkpoint/savepoint, or any persisted state — `LogUtil` is a stateless utility invoked per REST request and once per `initClean()`. For the shipped `config/log4j2.properties`, both appenders resolve under the same `${file_path}`, so their last `/` indices are numerically identical and the returned string is byte-for-byte unchanged before and after this fix — out-of-box clusters see zero behavior change on upgrade. Custom configurations that previously threw or returned an unusable path now get the correct directory; there is no way external tooling could have built a working dependency on the old broken value (an exception is not observable state, and an unlistable/nonexistent directory can't back a working integration). No `docs/en/introduction/concepts/incompatible-changes.md` entry is warranted, and none was added.

The production change is a one-token `substring` argument swap — no allocation, iteration, or I/O difference, and nothing on a hot path. The one real side effect remains on the test side: the new `log4j2-test.properties` in `seatunnel-engine-common` binds the module's root logger solely to a `File` appender, so the other test classes in the module now log to a discarded temp file instead of the CI console (Issue 2 below, unchanged, non-blocking, and confirmed still contained to this module since it has no attached `test-jar`).

Unchanged: no logging added or touched, and the `IllegalArgumentException` fallback for "neither appender configured" is untouched. As noted in my first review, this fix also removes the only realistic source of an unchecked exception reaching `BaseLogService.getLogPath()`'s narrow `catch (NoSuchFieldException | IllegalAccessException e)` — after this patch, `IllegalArgumentException` is the only remaining failure mode there, and every caller already handles it.

**CI status (re-verified live at the current head, `94004ec0a3`):** the apache-side "Build" check is `QUEUED` — it is a pointer to the fork's own Actions run, not an independent execution. I checked the fork directly: `repos/tagadearpit/seatunnel/actions/runs` shows run `32115988617` for this exact head SHA. It has been sitting in `queued`/`conclusion: null` since `2026-08-18T08:21:47Z`; as of this review (`2026-08-18T10:37:25Z`) that is over two hours with no job having started, which is well outside normal GitHub Actions scheduling latency and looks like the run is stuck rather than merely waiting its turn. The prior run (`32067465721`, for head `fe1a71c6eb`) came back red on exactly one job, `unit-test (8, windows-latest)`, with `ResourceManagerTest` failing to bring up a Hazelcast node (`IllegalStateException: Node failed to start!`) — a module (`seatunnel-engine-server`) this PR does not touch, and a Windows/Hazelcast startup flake I have independently seen recur across unrelated PRs in this repo. Every other job on that run (including `unit-test` on `ubuntu-latest` and other JDKs, and all `*-it-*` matrices) was green. I can't yet report a conclusion for the new run because it has not executed — see Issue 5.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Unchanged: ASF license headers present and correct, formatting consistent with Spotless/AOSP, no wildcard imports, no `@DisplayName`. Note carried over from my first review: the PR title is still truncated (`… in fileAppender br…`) — worth expanding to a full sentence before merge so it reads cleanly in release notes, but this is cosmetic and non-blocking.

Unchanged: `LogUtilTest.returnsTheDirectoryForAFileAppenderOnlyConfiguration()` is a genuine regression test for Case A (I confirmed again that it fails with `NullPointerException` under the pre-fix line and passes after). It remains a single deterministic assertion against `System.getProperty("java.io.tmpdir")` with no timing dependency, and I already worked through Linux/macOS/Windows separator behavior in a prior pass — no cross-platform fragility expected. What's still missing is coverage for the two *silent* failure modes (Case B's wrong-directory truncation, which is the one that reaches `TaskLogManagerService.clean()`'s delete path, and Case C's `StringIndexOutOfBoundsException`) — see Issue 1, carried over as Medium/non-blocking per the author's stated intent to address it as a follow-up seam.

## 2.3 Documentation Updates

Unchanged: no doc update needed. `rest-api-v1.md`/`rest-api-v2.md` document the log endpoints but not how the directory is derived internally, so there's no statement to correct.

# 3. Architectural Soundness

Unchanged: keeping `lastIndexOf("/")` instead of `File.getParent()`/`Paths.get(...).getParent()` avoids a silent behavior change on Windows or trailing-separator inputs. Correct scope discipline for a bugfix PR.

## 3.2 Maintainability
Unchanged: `getLogPath()` still fuses reflection, appender-ref selection, and string arithmetic in one method — the structural reason the original copy-paste hid. A package-private seam (Issue 1) remains the right long-term fix, reasonably deferred by the author.

## 3.3 Extensibility
Unchanged: acceptable as-is for two appender kinds; no abstraction is warranted for this scope.

Unchanged: `LogUtil` is stateless, not part of any checkpoint/savepoint/serialization path, and not referenced by any upgrade logic. A cluster upgraded with an unmodified `config/log4j2.properties` resolves the identical directory before and after; a cluster with a customized `fileAppender`-only configuration stops erroring, which is the intended repair. No historical-version compatibility risk.

---

## Issue 1
- **Location:** `seatunnel-engine/seatunnel-engine-common/src/test/java/org/apache/seatunnel/engine/common/utils/LogUtilTest.java`
- **Problem description:** The regression test covers only the `NullPointerException` mode (Case A). The silent wrong-directory truncation (Case B — the one that feeds `TaskLogManagerService.clean()`'s delete path) and `StringIndexOutOfBoundsException` (Case C) remain untested.
- **Potential risk:** Case B is the dangerous one precisely because nothing throws; a future refactor of this method could reintroduce cross-variable indexing and the suite would stay green.
- **Best improvement:** Extract the pure branch logic behind a package-private seam (`static String resolveLogDir(String logRef, String routingLogFilePath, String fileLogPath)`) and assert all five cases directly, independent of the process-wide `LoggerContext`.
- **Severity:** Medium

## Issue 2
- **Location:** `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-test.properties`
- **Problem description:** The new module-wide fixture binds the root logger to a single `File` appender, so all other test classes' console output in `seatunnel-engine-common` is redirected to a discarded temp file.
- **Potential risk:** Reduced CI debuggability for unrelated test failures in this module.
- **Best improvement:** Add console appenders alongside the file appender, mirroring `seatunnel-engine-server/src/test/resources/log4j2-test.properties`; confirmed safe since `getLogPath()`'s ref filter ignores console refs.
- **Severity:** Low

## Issue 3
- **Location:** `seatunnel-engine/seatunnel-engine-common/src/main/java/org/apache/seatunnel/engine/common/utils/LogUtil.java:50,52`
- **Problem description:** Pre-existing in both branches: no `-1` guard on `lastIndexOf("/")` and no Windows `\` handling.
- **Potential risk:** A relative or Windows-style `fileName` yields `-1` → uncaught `StringIndexOutOfBoundsException` → HTTP 500. Same class of bug as this PR fixes, reached a different way; not introduced or worsened by this diff.
- **Best improvement:** Optional — resolve the index as `Math.max(path.lastIndexOf('/'), path.lastIndexOf('\\'))` and throw the existing `IllegalArgumentException` when it's `-1`. Fine as a follow-up.
- **Severity:** Low

## Issue 4
- **Location:** `seatunnel-engine/seatunnel-engine-common/src/test/resources/log4j2-test.properties`
- **Problem description:** Fixed, never-cleaned filename (`seatunnel-logutil-test.log`) in the shared temp directory.
- **Potential risk:** Minor — leftover file accumulation; unlikely CI collision since matrix legs run on separate runners.
- **Best improvement:** Nit only; optionally disambiguate the name.
- **Severity:** Nit

## Issue 5
- **Location:** Fork run `32115988617` (head `94004ec0a3`), apache-side "Build" check
- **Problem description:** CI for the current head has not completed. The fork's run has been `queued` for over two hours (since `08:21:47Z`, still queued at `10:37:25Z` when I checked) with no job started — that duration is atypical for normal scheduling and suggests the run is stuck (e.g. a runner/concurrency issue on the fork) rather than just waiting. The prior run (`32067465721`, previous head) came back with exactly one failing job, `unit-test (8, windows-latest)` / `ResourceManagerTest`, in a module this PR doesn't touch and matching a known environmental Windows/Hazelcast flake.
- **Potential risk:** None to this diff's correctness — this is purely a "has the gate actually run" question, not a source-level concern. But an indefinitely stuck queue means the PR can sit unmergeable even though the code is fine.
- **Best improvement:** @tagadearpit, could you check the Actions tab on `tagadearpit/seatunnel` for this run and, if it's genuinely stuck (not just slow), cancel and re-trigger it? If it comes back red only on the same unrelated `ResourceManagerTest`/Windows job, that confirms the flake reading from my last review rather than anything new.
- **Severity:** Blocking (CI-only, not source)

# 4. Issue Summary

| --- | --- | --- | --- |
| 1 | Regression test covers only the NPE mode; silent wrong-directory (Case B) and out-of-bounds (Case C) modes remain untested | `LogUtilTest.java` | Medium |
| 2 | Module-wide `log4j2-test.properties` redirects other tests' console output to a discarded temp file | `log4j2-test.properties` | Low |
| 3 | Pre-existing (both branches): no `-1` guard, no Windows `\` handling | `LogUtil.java:50,52` | Low |
| 4 | Fixed, never-cleaned test fixture filename | `log4j2-test.properties` | Nit |
| 5 | CI for the current head has not completed — stuck `queued` for 2+ hours | fork run `32115988617` | Blocking (CI-only) |

# 5. Merge Recommendation

### Conclusion: Not yet ready to merge — no source-level blocker exists anywhere in this diff; the sole open item is that CI for the current head has been queued for over two hours without a job starting.

**Blockers**
1. The fork's CI run for head `94004ec0a3` (run `32115988617`) has been `queued` since `08:21:47Z` and was still `queued`/`conclusion: null` when I last checked at `10:37:25Z` — over two hours with no job execution, which looks stuck rather than merely pending. Nothing in the diff changed since the last run that came back green on every job except one known-unrelated Windows flake, so I expect this run to land the same way once it actually runs, but I'm not going to approve against a run that hasn't executed. @tagadearpit, please check the Actions tab on your fork and cancel/re-trigger if it's genuinely stuck.

**Recommended (non-blocking)**
2. Issue 1 — please open a tracking issue (or keep #11463 open) for the wrong-directory/overlong-index coverage gap.
3. Issue 2 — restore console appenders in the new fixture.
4. Issues 3 and 4 — author's discretion.

**Overall assessment:** unchanged from my prior three reviews — this remains a good, disciplined first contribution with an accurate diagnosis (confirmed again by direct blob comparison at the current head) and a minimal, correct, backward-compatible fix. The two intervening commits since my last review are both no-op merges from `dev` that don't touch any file under review, verified via `compare` API rather than a local clone. Once the currently-queued run for this head completes green (or comes back red only on the same pre-existing `ResourceManagerTest`/Windows flake noted last time), I'm ready to approve.

davidzollo (approved):
+1
LGTM

PR #11477, opened 2026-07-17
[Improve][Connector-V2] Add declarative Console sink delay validation

asrajawat:
### Purpose of this pull request

Part of #11007.

Adds a declarative `OptionRule` constraint for the Console sink so `log.print.delay.ms` must be non-negative and invalid values are rejected during configuration validation. It also adds focused unit coverage and aligns the English and Chinese connector documentation.

### Does this PR introduce _any_ user-facing change?

Yes. A negative `log.print.delay.ms` value was previously accepted and effectively treated as no delay. It is now rejected during option validation with an `OptionValidationException`. Zero and positive values retain their existing behavior.

### How was this patch tested?

- `./mvnw spotless:apply`
- `./mvnw -pl seatunnel-connectors-v2/connector-console -am -DskipIT=true -Dskip.spotless=true verify`
  - Console tests: 7 passed, 0 failed
- `git diff --check`

The repository-wide `./mvnw -q -DskipTests verify` progressed beyond the focused connector but could not complete locally because the corporate TLS proxy prevented Maven from validating the certificate chain for the external Mulesoft repository used by the Druid connector. The focused Console reactor completed successfully.

### Check list

* [x] No new Jar binary package is added.
* [x] English and Chinese Console connector documentation are updated.
* [x] No incompatible API or SPI change is introduced.
* [x] No new connector registration or distribution change is required.

PR #11821, opened 2026-08-15
[Improve][Connector-V2][AmazonDynamoDB] Migrate max retries validation to OptionRule

goutamadwant:

Part of #11007.

This PR migrates the existing AmazonDynamoDB sink `max_retries` validation to the declarative `OptionRule` path.

- Require `max_retries` to be non-negative in `AmazonDynamoDBSinkFactory`.
- Remove the duplicate imperative check from `AmazonDynamoDBConfig`.
- Keep `max_retries` optional so configurations that use the default remain valid.
- Document the accepted range in the English and Chinese connector docs.
- Add focused coverage for defaults, valid and invalid values, and unknown options.

Yes. A negative `max_retries` value is now rejected during option validation instead of later while the connector configuration is created. The accepted values and default behavior remain unchanged.

- The effective default value is `10`.
- Zero and positive values are accepted.
- A negative value is rejected with an error that identifies `max_retries`.
- Unknown options are rejected.

Before the production change, the negative-value regression failed because factory validation accepted `max_retries = -1`.

Verified the complete AmazonDynamoDB connector module with Java 11:

```shell
./mvnw -pl seatunnel-connectors-v2/connector-amazondynamodb package
```

Result: 7 tests passed. Spotless, compilation, tests, and packaging completed successfully.

### Check list

* [x] English and Chinese documentation describe the accepted range.
* [x] No incompatible-change entry is required because valid configurations keep the same behavior.
* [x] No connector registration, distribution, label, or E2E updates are required because this changes validation only.


DanielLeens (changes requested):
Thanks for picking this one up, @goutamadwant. This is exactly the shape of change the `#11007` umbrella is asking for, and the implementation follows the accepted precedent (`ConsoleSinkFactory`, #11477) almost line for line. I read the change against `dev` at `b1e21631ff` and PR head `a9b409663c`, traced the validation path end to end, and checked the released baseline. Findings below; most are small, and one is a project-policy blocker that takes two lines to fix.

**One-sentence summary:** It moves the AmazonDynamoDB sink's `max_retries >= 0` check out of an imperative `if/throw` buried in the connector config constructor and into the declarative `OptionRule`, so that an invalid value is rejected by the framework's option validator instead of by hand-written code inside `AmazonDynamoDBConfig`.

**The approach:**

1. `AmazonDynamoDBSinkFactory.optionRule()` moves `MAX_RETRIES` out of the plain `.optional(...)` group and re-declares it as `.optional(MAX_RETRIES, Conditions.greaterOrEqual(MAX_RETRIES, 0))`.
2. The runtime `if (this.maxRetries < 0) throw new IllegalArgumentException(...)` is deleted from `AmazonDynamoDBConfig`.
3. A new `AmazonDynamoDBSinkFactoryTest` covers default / zero / positive / negative values.

**Simple example.** With this config block:

```hocon
sink {
  AmazonDynamoDB {
    url = "http://127.0.0.1:8000"
    region = "us-east-1"
    access_key_id = "dummy-key"
    secret_access_key = "dummy-secret"
    table = "sink_table"
    max_retries = -1
  }
}
```

* **Before:** `seatunnel.sh --dry-run static` passed cleanly. The job only blew up later at submission, inside `AmazonDynamoDBConfig`'s constructor, as `IllegalArgumentException: max_retries must be a non-negative integer, but got: -1`, wrapped by `FactoryException: Unable to create a sink for identifier 'AmazonDynamoDB'`.
* **After:** the dry-run itself fails with an `OptionValidationException` naming `max_retries` and the `'max_retries' >= 0` constraint, before any factory instance is constructed.

**Is the claimed bug real? Yes — verified in source.** The dry-run entry point `seatunnel-core/seatunnel-starter/src/main/java/org/apache/seatunnel/core/starter/seatunnel/command/SeaTunnelConfValidateCommand.java:233-235` calls `ConfigValidator.of(readonlyConfig).validate(factory.optionRule())`, `validateOptionTypes(...)` and `ConfigValidator.validateUnknownKeys(...)`, and **never** constructs the factory's runtime objects (the class javadoc is explicit: static dry-run "does not create source/sink runtime instances"). So an imperative check living inside `AmazonDynamoDBConfig` was structurally invisible to `--dry-run static`. Moving it into `OptionRule` genuinely closes that hole. This is not a cosmetic refactor.

# 1. Code Change Review

## 1.1 Core Logic Analysis

I traced the three validation surfaces the task touches.

**(a) Does the new constraint actually fire?** Yes. `OptionRule.Builder.optional(Option, Condition, Condition...)` (`seatunnel-api/.../util/OptionRule.java`) appends the option to `optionalOptions` and the condition to `valueConstraints`. `ConfigValidator.collectErrors` iterates `rule.getValueConstraints()` and evaluates each one via `ConditionEvaluators.evaluate`, which for `GREATER_OR_EQUAL` resolves the value with `config.get(condition.getOption())` and compares numerically. `MAX_RETRIES` is `Option<Integer>` and the literal `0` binds to `Integer`, so `Conditions.greaterOrEqual(MAX_RETRIES, 0)` type-checks and evaluates correctly.

**(b) Does the rule still build?** Yes. `verifyOptionOptionsDuplicate(MAX_RETRIES, OPTIONS_OPTION_WITH_CONSTRAINT)` checks `optionalOptions.contains(option)` and scans `requiredOptions`. Because the author correctly *removed* `MAX_RETRIES` from the first `.optional(...)` group rather than leaving it in both, there is no duplicate and no `OptionValidationException` at builder time. This is the exact trap the framework guards against, and the PR avoids it. Good.

**(c) When is the constraint skipped?** `ConfigValidator.isConstraintApplicable` returns `true` immediately only for options that are absolutely required; `MAX_RETRIES` is optional, so it falls through to `anyOrSegmentFullyPresent`, which requires `hasOption(MAX_RETRIES)`, i.e. `ReadonlyConfig.getOptional(...).isPresent()`. And `getOptional` returns `Optional.empty()` when the key is absent — it does **not** fall back to the default (`ReadonlyConfig.java:103-124`; the default is only applied by `get()` at line 65-67). Net effect: **when `max_retries` is not configured, the constraint is not evaluated at all.** That is harmless here (the default is `10`), but it has a direct consequence for the tests — see Issue 3.

**(d) Was a runtime guard silently dropped for other construction paths?** Partially, and I want to be precise about it rather than hand-wave:

* **Sink factory path — fully covered.** `FactoryUtil.createAndPrepareSink` calls `ConfigValidator.of(context.getOptions()).validate(tableSinkFactory.optionRule())` at `FactoryUtil.java:252`, *before* `tableSinkFactory.createSink(context)` at line 260. The rule now fires strictly earlier than the deleted check did. No regression.
* **Legacy fallback sink path — not applicable.** `AmazonDynamoDBSinkFactory.createSink` returns a lambda without dereferencing the context, so `isFallback`'s probe (`sinkFactory.createSink(null)`) does not throw and the deprecated `sink.prepare(Config)` branch is never taken for this connector.
* **Source factory path — a real, narrow gap.** `AmazonDynamoDBSourceFactory.createSource` also builds `new AmazonDynamoDBConfig(context.getOptions())`, and `AmazonDynamoDBConfig` unconditionally reads `AmazonDynamoDBSinkOptions.MAX_RETRIES` at line 66. The *source* `optionRule()` declares neither `MAX_RETRIES` nor the constraint, so the deleted guard was the only thing rejecting `max_retries = -1` in a source block. See Issue 6 — impact is negligible, but it should be acknowledged rather than assumed away.
* **Direct/Lombok construction** (`@AllArgsConstructor`) never had the guard in the first place, so nothing changes there.

**(e) Worst case if a negative value ever did reach runtime.** `DynamoDbSinkClient.flushWithRetry` loops `while (!pendingRequests.isEmpty() && retryCount <= maxRetries)`. With `maxRetries = -1` the loop body never executes, `pendingRequests` stays non-empty, and the method throws `RuntimeException("Failed to write N items to table T after -1 retries")`. So the failure mode is **loud, not silent** — no data would be dropped without an error. That materially lowers the risk of the gap in (d).

## 1.2 Compatibility Impact

**Explicit conclusion: no existing valid configuration changes behavior, and there is zero historical-version risk.**

* **Accepted domain is byte-for-byte identical.** The removed check rejected `maxRetries < 0`; the new constraint requires `>= 0`. Same set. No previously-accepted value is now rejected, and no previously-rejected value is now accepted *on the sink path*.
* **Default unchanged.** `MAX_RETRIES` remains `Options.key("max_retries").intType().defaultValue(10)`. The PR does not touch `AmazonDynamoDBSinkOptions`.
* **Released-version surface is empty.** I checked the latest release tag: `git show 2.3.13:.../AmazonDynamoDBSinkOptions.java` contains no `MAX_RETRIES` (nor `RETRY_BASE_DELAY_MS` / `RETRY_MAX_DELAY_MS`). This whole retry feature is unreleased `dev`-only work. No shipped user config can possibly be affected, and no `docs/en/introduction/concepts/incompatible-changes.md` entry is warranted. The PR's checklist claim on this point is correct.
* **No E2E or test resource regressions.** `git grep max_retries` across `seatunnel-e2e/**` returns hits only for `connector-edge-socket`, `connector-jdbc` and `connector-starrocks`; `connector-amazondynamodb-e2e/src/test/resources/amazondynamodbIT_source_to_sink.conf` does not set the option. The only remaining reference to the old message string in the tree is the line being deleted, so the removal is clean with no orphaned assertions.
* **One genuine behavior delta:** the exception *type* and *message* change (`IllegalArgumentException` → `OptionValidationException`, and the message no longer echoes the offending value). Both are still wrapped by `FactoryException` on the submission path, so no caller-visible type contract breaks. See Issue 4.

No performance concern. `optionRule()` is invoked once per plugin per job submission, and one extra `Condition` node in `valueConstraints` costs an `EnumMap` lookup plus a `BigDecimal`-backed numeric compare — nanoseconds, on a path that already does classloader construction and SPI scanning. Removing the constructor check removes one comparison from a once-per-sink code path. Nothing enters the record hot path.

One benign side effect worth noting: `OptionRule.getOptionalOptions()` ordering changes, because `MAX_RETRIES` now appends after `RETRY_BASE_DELAY_MS` / `RETRY_MAX_DELAY_MS` instead of sitting between `MULTI_TABLE_SINK_REPLICA` and them. This list feeds the web UI option listing and `OptionRule.equals`. I found no test or golden file in-tree that asserts option order for this connector (`AmazonDynamoDBSourceFactoryTest` only asserts non-nullness), so this is informational, not an issue.

The connector adds no logging, which is correct — validation errors belong to the framework, and `ConfigValidator` already aggregates them into a numbered `Option validation failed (N errors):` report. No sensitive values are logged; `access_key_id` / `secret_access_key` are untouched by this change.

The one substantive regression is diagnosability: `formatError(constraint.getOption().key(), TYPE_VALUE, constraint.toString())` produces a message built from `Condition.toString()`, i.e. `'max_retries' >= 0`. The **actual offending value is not in the message**, whereas the deleted check said `but got: -1`. For a single-option constraint this is usually recoverable by looking at the config, but it is a real step backwards for a user staring at a large multi-sink job. See Issue 4.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Formatting is Spotless-clean (AOSP), imports are explicit with no wildcards, and the new file carries the correct ASF license header. The `Conditions` import is a plain (non-static) import, which matches `ConsoleSinkFactory`; other migrated connectors such as `FlussSourceFactory` use static imports. Both styles exist in `dev`, so this is not a defect.

The chosen call shape — a dedicated `.optional(OPTION, Conditions.greaterOrEqual(OPTION, 0))` alongside the untouched bulk `.optional(...)` group — is *identical* to the merged `connector-console` migration (#11477) and to the guidance in the umbrella issue's "A. Numeric range" section. Consistency with precedent here is a positive.

On comment completeness: the production diff introduces no new method or field, so the "new core methods/fields need comments" rule has nothing to bite on in `main`. The new **test** class, however, has no class-level Javadoc explaining which production behavior it guards, and neither do the helper methods — see Issue 7. That is a low-severity gap, but it is the kind of context that pays off when someone later wonders why `testDefaultMaxRetries` exists.

Net coverage clearly increases: before this PR the `max_retries` guard had **no** test at all, so nothing here weakens historical test strictness. That said, the suite does not yet fully exercise what it claims to.

* `testNegativeMaxRetriesFails` is the load-bearing test and it does hit the new `valueConstraints` path. Good.
* `testDefaultMaxRetries` passes *vacuously* — with `max_retries` absent, `isConstraintApplicable` short-circuits and the constraint is never evaluated (see 1.1(c)). It would pass identically if the condition were `greaterOrEqual(MAX_RETRIES, 999)`. It still has value as a guard against someone accidentally promoting `MAX_RETRIES` to `required`, but that intent should be stated. See Issue 3.
* Nothing exercises `ConfigValidator.validateUnknownKeys(...)`, which is the *other* half of the dry-run gate that `SeaTunnelConfValidateCommand` runs. `connector-kafka`'s `KafkaFactoryTest` shows the established pattern for this. Because `collectDeclaredKeys` walks `rule.getValueConstraints()` via `collectConditionKeys` to harvest keys, restructuring an option into the constraint form is precisely the kind of edit that can accidentally drop a key from the declared set. See Issue 5.

**Flaky-pattern check: clean.** The new tests are pure in-memory `HashMap` → `ReadonlyConfig` → `ConfigValidator` calls. No sleeps, no timeouts, no ordering dependence between test methods, no shared mutable static state (`sinkRule` is rebuilt per test in `@BeforeEach`), no containers, no network, no filesystem, no clock or randomness. Nothing that can flake in CI.

## 2.3 Documentation Updates

`docs/en/connectors/sink/AmazonDynamoDB.md` and `docs/zh/connectors/sink/AmazonDynamoDB.md` both document `max_retries` (table row at line 36, prose section at line 76) but **neither states that the value must be non-negative** — before or after this PR. The PR checklist argues documentation is not required "because the accepted range and default remain unchanged," which is factually true, yet the same PR answers *Yes* to "does this introduce a user-facing change." Given the whole point of `#11007` is to make constraints discoverable, and given a negative value now hard-fails at dry-run with a message that omits the offending value, the constraint should be written down. See Issue 1 — this is the one blocker.

# 3. Architectural Soundness

**This is a precise fix, not a temporary workaround, and it is the correct long-term direction.** The umbrella issue's section E lists what should *stay* imperative — external system reachability, parser/semantic validation, runtime context checks. A static numeric range on a scalar option is none of those; it is the canonical case for declarative migration. The change moves a constraint from a place where only the runtime can see it to a place where the CLI validator, the runtime, and (via `OptionRule` metadata) the web UI and REST option listing can all see it. That is a strict architectural improvement.

The one place the solution stops short of "complete" is scope: it declares the *lower* half of `max_retries`' valid domain while the consuming code has a genuine upper-bound hazard (Issue 2), and it leaves the sibling options in the same `OptionRule` un-migrated (Issue 8).

## 3.2 Maintainability

Positive. The constraint now lives next to the option declaration instead of ~40 lines away in a different package, and it is expressed in the same vocabulary as every other migrated connector, so a reader who knows one knows all. Deleting the imperative branch also removes a duplicate source of truth — previously a future contributor could have tightened one and not the other.

## 3.3 Extensibility

Adding further constraints to this connector is now a one-line append per option, and `Conditions` chaining (`.and(...)`) plus `lessOrEqualField` are already available for the cross-field cases this connector needs (`retry_base_delay_ms` vs `retry_max_delay_ms`). The PR leaves the file in a strictly better state to extend.

**No risk.** As established in 1.2, `max_retries` and the entire retry-backoff option family do not exist in `2.3.13`, the latest release tag — they are unreleased `dev` work. There is therefore no shipped configuration, no persisted job definition, and no upgrade path that can observe this change. Checkpoint/savepoint state is unaffected (this option participates in no serialized state; `AmazonDynamoDBConfig` is only carried into the writer). Nothing in the change touches SPI signatures, `plugin-mapping.properties`, factory identifiers, or `@AutoService` registration.

# 4. Issue Summary

|---|---|---|---|
| 1 | `max_retries` non-negative constraint not documented in `docs/en` or `docs/zh` | `docs/en/connectors/sink/AmazonDynamoDB.md:36,76`; `docs/zh/connectors/sink/AmazonDynamoDB.md:36,76` | Medium (Blocker by project doc policy) |
| 2 | No upper bound declared; large `max_retries` overflows the backoff shift and crashes on `Thread.sleep(negative)` | `.../sink/AmazonDynamoDBSinkFactory.java:57`; `.../sink/DynamoDbSinkClient.java` (`flushWithRetry`) | Medium |
| 3 | `testDefaultMaxRetries` passes vacuously — constraint is skipped when the option is absent | `.../AmazonDynamoDBSinkFactoryTest.java:49-52` | Low |
| 4 | Error message no longer reports the offending value (regression vs. removed `IllegalArgumentException`) | `.../sink/AmazonDynamoDBSinkFactory.java:57` (framework: `OptionUtil.formatError`) | Low |
| 5 | No test exercises `ConfigValidator.validateUnknownKeys()` for the restructured rule | `.../AmazonDynamoDBSinkFactoryTest.java:85-87` | Low |
| 6 | Removed guard also covered the source construction path, which declares no equivalent constraint | `.../config/AmazonDynamoDBConfig.java:66`; `.../source/AmazonDynamoDBSourceFactory.java` (`optionRule`) | Low |
| 7 | New test class and helpers lack Javadoc stating the production behavior they guard | `.../AmazonDynamoDBSinkFactoryTest.java:40-87` | Low |
| 8 | Umbrella lists this connector as "Both" and unclaimed; sibling sink options and both source options remain unmigrated | `#11007` tracking table; `.../sink/AmazonDynamoDBSinkFactory.java:50-58`; `.../source/AmazonDynamoDBSourceFactory.java` | Low (process/scope) |

---

### Issue 1 — `max_retries` constraint is not documented in either language

* **Location:** `docs/en/connectors/sink/AmazonDynamoDB.md:36` (options table) and `:76` (`### max_retries [int]`); `docs/zh/connectors/sink/AmazonDynamoDB.md:36` and `:76`.
* **Problem description:** After this PR a negative `max_retries` is a hard validation failure at `--dry-run static`, but neither the English nor the Chinese doc states any constraint. The English prose reads only "The maximum number of retries when DynamoDB returns unprocessed items from a batch write request." The PR checklist asserts documentation is not required because the accepted range is unchanged, yet the same PR answers *Yes* to "does this PR introduce any user-facing change" — and the project rule is that any user-visible change updates both `docs/en` and `docs/zh`, with config constraints matching the code exactly.
* **Potential risk:** A user hits an `OptionValidationException` whose message (`'max_retries' >= 0`) does not include their actual value (see Issue 4) and finds nothing in the documentation confirming the rule. It also undercuts the umbrella's stated goal of making constraints discoverable rather than tribal knowledge.
* **Best improvement:** Add one sentence to each `### max_retries [int]` section, e.g. EN: "Must be greater than or equal to `0`. A value of `0` disables retries. The job fails validation if a negative value is configured." plus the equivalent Chinese sentence. Optionally append the same note to the options-table description cell at line 36 in both files.
* **Severity:** Medium (treated as a merge blocker under the project's documentation policy; the fix is two lines).
* **Raised by another reviewer:** No.

### Issue 2 — No upper bound, and the consuming backoff overflows at high retry counts

* **Location:** `seatunnel-connectors-v2/connector-amazondynamodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/amazondynamodb/sink/AmazonDynamoDBSinkFactory.java:57`; consuming code in `.../sink/DynamoDbSinkClient.java`, `flushWithRetry`.
* **Problem description:** This PR is precisely the change that declares `max_retries`' valid domain, but it declares only the lower half. `flushWithRetry` computes `long delay = Math.min(baseDelayMs * (1L << retryCount), maxDelayMs);`. With the default `retry_base_delay_ms = 100`, `100 * (1L << 57)` exceeds `Long.MAX_VALUE` and wraps negative; `Math.min(negative, 5000)` selects the negative value, the jitter term `(long)(delay * Math.random() * 0.5)` is also negative, and `Thread.sleep(delay)` then throws `IllegalArgumentException: timeout value is negative`. At `retryCount >= 63` the shift itself is degenerate (Java masks the shift distance to 6 bits, and `1L << 63` is `Long.MIN_VALUE`). The overflow threshold moves lower as `retry_base_delay_ms` grows — at `retry_base_delay_ms = 60000` it starts around `retryCount >= 47`.
* **Potential risk:** Because the per-attempt delay is capped at `retry_max_delay_ms` (default 5s), reaching 57 consecutive retries takes roughly five minutes of sustained throttling — entirely reachable for a user who sets `max_retries = 100` on a hot-partitioned table. The sink then dies with an opaque `IllegalArgumentException` from `Thread.sleep` instead of the intended "failed after N retries" error, and the operator has no signal pointing at their configuration. Note this is a **pre-existing** latent defect, not one introduced here; I raise it because this PR is the natural place to close it, and because `max_retries` is still unreleased (absent from `2.3.13`), so tightening the range now costs nothing in compatibility — whereas after a release it would be an incompatible change.
* **Best improvement:** Chain an upper bound at line 57, e.g. `.optional(MAX_RETRIES, Conditions.greaterOrEqual(MAX_RETRIES, 0).and(Conditions.lessOrEqual(MAX_RETRIES, 30)))`, picking a cap that is comfortably below the overflow threshold for any plausible `retry_base_delay_ms`. If you would rather keep this PR strictly scoped, the alternative is to clamp the shift in `flushWithRetry` (e.g. `1L << Math.min(retryCount, 30)`) in a separate fix PR — either is fine, but please pick one and say which, rather than leaving the hazard undeclared on both sides.
* **Severity:** Medium.
* **Raised by another reviewer:** No.

### Issue 3 — `testDefaultMaxRetries` passes for the wrong reason

* **Location:** `seatunnel-connectors-v2/connector-amazondynamodb/src/test/java/org/apache/seatunnel/connectors/seatunnel/amazondynamodb/AmazonDynamoDBSinkFactoryTest.java:49-52`.
* **Problem description:** The test asserts that a config without `max_retries` validates cleanly, and the PR body presents it as covering "the default `max_retries` value is accepted." In fact the default is never evaluated: `ConfigValidator.isConstraintApplicable` falls through to `anyOrSegmentFullyPresent`, which calls `hasOption(MAX_RETRIES)`; `ReadonlyConfig.getOptional` returns `Optional.empty()` for an absent key rather than the default, so the constraint is skipped entirely. The test would pass unchanged even if the condition were `greaterOrEqual(MAX_RETRIES, 999)`.
* **Potential risk:** Future readers trust it as coverage of the default-value path and skip writing the real check. If someone later swaps the operator or the literal, this test stays green.
* **Best improvement:** Keep the test — it is a genuine guard against `MAX_RETRIES` being accidentally promoted to `required` — but add a Javadoc line stating what it actually proves ("an absent `max_retries` must not fail validation; the value constraint is skipped when the option is not configured"). Separately, add an explicit case pinning the effective default, e.g. `Assertions.assertEquals(10, ReadonlyConfig.fromMap(requiredOptions()).get(MAX_RETRIES));`, so a silent default change is caught.
* **Severity:** Low.
* **Raised by another reviewer:** No.

### Issue 4 — Validation error no longer reports the offending value

* **Location:** `seatunnel-connectors-v2/connector-amazondynamodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/amazondynamodb/sink/AmazonDynamoDBSinkFactory.java:57` (message produced by `OptionUtil.formatError` + `Condition.toString()`).
* **Problem description:** The deleted check said `max_retries must be a non-negative integer, but got: -1`. The framework message is assembled from the constraint's own rendering — `option: max_retries / type: value / constraint: 'max_retries' >= 0` — and contains the rule but not the configured value.
* **Potential risk:** Diagnosability regression, mildest in a single-sink job and worst in a large multi-sink pipeline where the operator must hunt through the config to find which block carries the bad value. This is a framework-level characteristic shared by every migrated connector (including the merged `connector-console` precedent), so it is not this PR's fault — but the migration does make it user-visible for DynamoDB, and it compounds Issue 1.
* **Best improvement:** No change required in this PR. Please assert on the constraint text as well as the key in `testNegativeMaxRetriesFails` (see Issue 5), and consider raising a framework follow-up under `#11007` to have `ConfigValidator` include the actual value in `TYPE_VALUE` errors — that would improve every migrated connector at once.
* **Severity:** Low.
* **Raised by another reviewer:** No.

### Issue 5 — No coverage of `validateUnknownKeys()` and a weak negative assertion

* **Location:** `seatunnel-connectors-v2/connector-amazondynamodb/src/test/java/org/apache/seatunnel/connectors/seatunnel/amazondynamodb/AmazonDynamoDBSinkFactoryTest.java:85-87` (`validate` helper) and `:60-67` (`testNegativeMaxRetriesFails`).
* **Problem description:** Two related gaps. (a) The helper only calls `ConfigValidator.of(...).validate(sinkRule)`. `SeaTunnelConfValidateCommand` runs `validate`, `validateOptionTypes` **and** `ConfigValidator.validateUnknownKeys(...)`; the last one derives its declared-key set through `collectDeclaredKeys` → `collectConditionKeys`, which walks `rule.getValueConstraints()`. Restructuring an option into constraint form is exactly the edit that could drop a key from that set, and nothing here would catch it. (b) `Assertions.assertTrue(exception.getMessage().contains(MAX_RETRIES.key()))` only proves the string `max_retries` appears somewhere; it would also pass for a missing-required-option error or an unknown-key error that happens to list declared options.
* **Potential risk:** A future refactor of `OptionRule` or `ConfigValidator` could make `max_retries` an "unknown key" or shift the failure category, and this suite stays green while real configs break at dry-run.
* **Best improvement:** Follow the `KafkaFactoryTest` pattern — add a test calling `ConfigValidator.validateUnknownKeys(ReadonlyConfig.fromMap(requiredOptionsWithMaxRetries()), sinkRule, "AmazonDynamoDB")` and asserting it does not throw. Tighten the negative assertion to also require the value-constraint marker, e.g. `Assertions.assertTrue(exception.getMessage().contains(">= 0"))`.
* **Severity:** Low.
* **Raised by another reviewer:** No.

### Issue 6 — The removed guard also covered the source construction path

* **Location:** `seatunnel-connectors-v2/connector-amazondynamodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/amazondynamodb/config/AmazonDynamoDBConfig.java:66`; `.../source/AmazonDynamoDBSourceFactory.java` (`optionRule` / `createSource`).
* **Problem description:** `AmazonDynamoDBConfig` is shared: `AmazonDynamoDBSourceFactory.createSource` also constructs it, and the constructor unconditionally reads `AmazonDynamoDBSinkOptions.MAX_RETRIES`. The source's `optionRule()` declares neither the option nor the constraint, so after this PR a source block containing `max_retries = -1` is no longer rejected by the deleted guard, and nothing else replaces it on that path.
* **Potential risk:** Low in practice, and I want to be honest about how low: the source never reads `maxRetries`, and `ConfigValidator.validateUnknownKeys` would flag `max_retries` on a source block as an unknown key during `--dry-run static` anyway. Even in the sink case, a `-1` reaching `flushWithRetry` produces a loud `RuntimeException`, never silent data loss. So this is a completeness observation, not a live defect.
* **Best improvement:** No fix required for merge. Please just call this out in the PR description so reviewers of the eventual source-side migration know the shared `AmazonDynamoDBConfig` constructor no longer carries any guard. If you extend this PR to the source half (Issue 8), the natural resolution is for the source rule to declare its own constraints and for `AmazonDynamoDBConfig` to stay purely declarative.
* **Severity:** Low.
* **Raised by another reviewer:** No.

### Issue 7 — New test class and helpers lack explanatory Javadoc

* **Location:** `seatunnel-connectors-v2/connector-amazondynamodb/src/test/java/org/apache/seatunnel/connectors/seatunnel/amazondynamodb/AmazonDynamoDBSinkFactoryTest.java:40-87`.
* **Problem description:** The class has no type-level Javadoc naming the production behavior it protects (that `AmazonDynamoDBSinkFactory.optionRule()` rejects a negative `max_retries` at option-validation time, replacing the imperative check formerly in `AmazonDynamoDBConfig`). The helpers `validate`, `validateMaxRetries` and `requiredOptions` are also undocumented. The existing `AmazonDynamoDBSourceFactoryTest` at least carries a `/** Method: optionRule() */` marker.
* **Potential risk:** A future contributor cannot tell from the file why these cases exist or what regression they guard, which is exactly how vacuous tests like the one in Issue 3 survive refactors.
* **Best improvement:** Add a short class-level Javadoc naming the guarded behavior and referencing `#11007`, plus one-line Javadoc on `validate` explaining that it mirrors the dry-run validation entry point.
* **Severity:** Low.
* **Raised by another reviewer:** No.

### Issue 8 — Scope and tracking: connector is listed as "Both" and still unclaimed

* **Location:** `#11007` tracking table (row `Both | connector-amazondynamodb | | Todo |`); `.../sink/AmazonDynamoDBSinkFactory.java:50-58`; `.../source/AmazonDynamoDBSourceFactory.java` (`optionRule`).
* **Problem description:** The umbrella lists this connector under **Both**, with no contributor and status `Todo`. This PR covers one option on the sink half only. Several adjacent constraints are documented behavior today but still unenforced: `batch_size` — the docs state twice that DynamoDB accepts at most 25 write requests per batch call and that values above 25 must not be used, yet nothing rejects `batch_size = 500`, and `batch_size <= 0` silently disables size-triggered flushing in `DynamoDbSinkClient.write`; `retry_base_delay_ms` / `retry_max_delay_ms` have no lower bound and no ordering constraint, though `Conditions.lessOrEqualField` exists for exactly this; on the source side `parallel_scan_threads` becomes `totalSegments` in `AmazonDynamoDBSourceSplitEnumerator` (a value of `0` or less yields zero splits and a silently empty read), and `scan_item_limit` becomes the DynamoDB `Limit`, which the AWS API requires to be `>= 1`.
* **Potential risk:** Two contributors could claim the same connector and duplicate work; and a reader of the tracking table may see this connector marked complete when only a fraction of its imperative/undeclared validation has been migrated.
* **Best improvement:** Please comment on `#11007` to claim `connector-amazondynamodb` and get the table row updated to `In review` with this PR number. Then either extend this PR to the remaining sink options and the source rule, or state explicitly in the description that this is step 1 of N so the row is not closed prematurely. `parallel_scan_threads` (silently empty reads) is the highest-value remaining one.
* **Severity:** Low (process/scope).
* **Raised by another reviewer:** No.

# 5. Merge Recommendation

**Blockers (must fix)**

1. **Issue 1** — Add the `max_retries >= 0` constraint to both `docs/en/connectors/sink/AmazonDynamoDB.md` and `docs/zh/connectors/sink/AmazonDynamoDB.md`. The PR answers *Yes* to "user-facing change," and the project rule is that user-visible changes update both language docs with constraints matching the code exactly. This is a two-line fix and the only thing standing between this PR and a merge.
2. **CI must be green.** At the time of review the `Build` check on head `a9b409663c` is `queued` (fork run `goutamadwant/seatunnel` run 31902995638, also `queued`); the apache-side `Build` status is a pointer and reads `pending`. No failure to diagnose yet — this is simply not-yet-run, and the result must be confirmed before merge rather than assumed.

**Recommended (non-blocking)**

* **Issue 2** — Declare an upper bound on `max_retries`, or fix the `1L << retryCount` overflow in `DynamoDbSinkClient.flushWithRetry`. Worth doing *now* while the option is still unreleased; after a release, adding a cap becomes an incompatible change.
* **Issue 3** — Document why `testDefaultMaxRetries` passes, and add an explicit assertion on the effective default value.
* **Issue 5** — Add a `validateUnknownKeys` test (see `KafkaFactoryTest`) and tighten the negative assertion to include `>= 0`.
* **Issue 7** — Add class-level Javadoc to the new test naming the guarded production behavior.
* **Issue 6** — Note in the PR description that the shared `AmazonDynamoDBConfig` constructor now carries no guard for the source path.
* **Issue 8** — Claim the connector on `#11007` and state whether this PR is the complete migration or step 1.

**Overall assessment**

This is a well-executed, correctly-scoped migration and the analysis behind it holds up. I independently confirmed the claimed bug is real rather than theoretical: `SeaTunnelConfValidateCommand` never instantiates factory runtime objects, so a check living inside `AmazonDynamoDBConfig` was structurally invisible to `--dry-run static`, and moving it to `OptionRule` genuinely closes that hole. The implementation avoids the two traps I look for in this migration pattern — it removes `MAX_RETRIES` from the bulk `.optional(...)` group rather than duplicating it (which would have thrown at builder time), and it preserves the exact accepted domain rather than quietly widening or narrowing it. Compatibility risk is nil: I verified against tag `2.3.13` that neither `max_retries` nor the retry-backoff options exist in any release, so no shipped configuration can observe this. The tests are deterministic with no flake surface, and net coverage strictly increases since this guard previously had none.

The gaps are all at the edges rather than in the core. One is a documentation obligation the PR waives on a technicality that does not quite hold. The most interesting is Issue 2 — while tracing what a bad `max_retries` actually does downstream, the backoff shift turns out to overflow at high retry counts and crash on a negative `Thread.sleep`; that is pre-existing and not your bug, but this PR is the cheapest possible moment to close it, and I would encourage you to take it while the option is still unreleased.

Fix the docs, confirm CI green, and this is good to go. Thanks again for taking on an umbrella item — this kind of steady, well-tested migration work is genuinely valuable to the project, and the fact that the change matches the `connector-console` precedent without being told to says good things about how carefully you read the existing work first.

goutamadwant:
@DanielLeens Addressed the remaining validation coverage. The English and Chinese docs now state that max_retries must be non-negative, the tests pin the effective default to 10, and unknown options are covered. The full AmazonDynamoDB module passes with 7 tests.

davidzollo (commented):
Posted in error under the wrong GitHub account due to a local tooling mistake (token env var did not propagate to this call). Content removed; the correct re-review has been posted separately from @DanielLeens. Apologies for the noise.

DanielLeens (commented):
Thanks for the quick follow-up, @goutamadwant — I re-read the full diff at the new head (`d2d43b4a4e`, merged from `dev` on top of `6e2c0b2acc`), not just the delta, and re-traced the retry/backoff config path end to end rather than trusting my own prior notes. Both blockers from the last round are now fixed, one previously-recommended item picked up an extra layer of coverage, and I found one place where my own earlier review overstated a risk — flagging that here for the record rather than letting it stand uncorrected.

Same as before: it moves the AmazonDynamoDB sink's `max_retries >= 0` check out of an imperative `if/throw` in `AmazonDynamoDBConfig`'s constructor and into the declarative `OptionRule` on `AmazonDynamoDBSinkFactory`, so an invalid value is caught by `--dry-run static` / `ConfigValidator` instead of surfacing later as a wrapped `FactoryException` at submission time. That analysis held up on re-verification and nothing in the new commits changes it — see 1.1 below for the specific re-checks.

# 1. Code Change Review

## 1.1 Core Logic Analysis

The production diff is unchanged from the previous head: `AmazonDynamoDBSinkFactory.java:52-59` still removes `MAX_RETRIES` from the bulk `.optional(...)` group and re-declares it via `.optional(MAX_RETRIES, Conditions.greaterOrEqual(MAX_RETRIES, 0))`, and the imperative check at the old `AmazonDynamoDBConfig.java:66-70` is still deleted. I re-verified the constraint still fires as intended by reading `OptionRule.java` and `ConfigValidator.java` fresh (not reusing my prior notes verbatim):

* `OptionRule.Builder.optional(Option, Condition, Condition...)` (`OptionRule.java:346-356`) adds the option to `optionalOptions` at line 352 **and** the condition to `valueConstraints` at line 353 — two independent lists, both populated in one call.
* `ConfigValidator.collectAndValidate` walks `rule.getValueConstraints()` and, when `!validate(constraint)`, raises `formatError(constraint.getOption().key(), TYPE_VALUE, constraint.toString())` (`ConfigValidator.java:230-244`).
* `Condition.conditionToString` renders `GREATER_OR_EQUAL` as `'max_retries' >= 0` (`Condition.java:233-246`), so the final exception message contains the literal substring `>= 0`. This confirms the new test's `assertTrue(exception.getMessage().contains(">= 0"))` assertion is checking something real, not a guess.

**Self-correction on my own prior Issue 5:** I previously wrote that "restructuring an option into constraint form is exactly the edit that could drop a key from the declared set" used by `ConfigValidator.validateUnknownKeys`. Re-reading `ConfigValidator.collectDeclaredKeys` (`ConfigValidator.java:127-143`) now, line 131 calls `collectKeys(keys, rule.getOptionalOptions())` *before* it ever looks at `valueConstraints` — and, per the `OptionRule.Builder.optional(Option, Condition...)` code above, `MAX_RETRIES` was always added to `optionalOptions` regardless of which `.optional(...)` overload declares it. So the "declared key could be silently dropped" risk I described was overstated: this specific restructuring could never have caused `max_retries` to be treated as unknown. The new `testMaxRetriesIsADeclaredOption` test (see 2.2) is still good defense-in-depth against a *different* future mistake (e.g. someone moving the option into a `conditional(...)` branch that isn't reached), so I'm not withdrawing the recommendation to add it — I'm just correcting the reasoning I gave for it.

## 1.2 Compatibility Impact

**Fully compatible — unchanged from the last review, and nothing in the two new commits touches this.** Accepted domain is still byte-for-byte identical to the deleted check (`>= 0` in both). `MAX_RETRIES` still defaults to `10` (`AmazonDynamoDBSinkOptions` untouched). The retry-backoff option family is still absent from the `2.3.13` release tag, so there is no shipped configuration this can affect and no `docs/en/introduction/concepts/incompatible-changes.md` entry is needed. `git diff` for this round touches only `docs/en`, `docs/zh`, and the test file — zero production-code lines changed since my last review.

No change from last round: one extra `Condition` node evaluated once per sink per job submission, off the record hot path. Nothing new to flag.

Unchanged: validation errors are still framework-aggregated, no sensitive values touched. The diagnosability gap I flagged before (Issue 4 — the thrown message states the rule `'max_retries' >= 0` but not the user's actual configured value) is still present in production code and was not in scope for this follow-up; the new test compensates for it at the test level by asserting on the constraint text, which is the mitigation I suggested last time.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Still Spotless-clean, ASF header present on the test file, no wildcard imports. No new production code, so nothing new to check there.

The test file was substantially reworked between the two rounds (commit `6e2c0b2acc`, "Complete validation coverage"). Line-by-line against the current file:

* **Old blocker-adjacent gap now closed — default value pinned.** `testDefaultMaxRetries` (vacuous, per my last review) is replaced by `testOmittedMaxRetriesUsesDefault`, which keeps the "no throw" assertion but adds `Assertions.assertEquals(10, ReadonlyConfig.fromMap(config).get(MAX_RETRIES))` (`AmazonDynamoDBSinkFactoryTest.java:139-144`). This now genuinely exercises the default rather than passing regardless of what the constraint says. **Issue 3 from my last review: resolved.**
* **`validateUnknownKeys` coverage added.** `testMaxRetriesIsADeclaredOption` (`:162-171`) calls `ConfigValidator.validateUnknownKeys(ReadonlyConfig.fromMap(config), sinkRule, "AmazonDynamoDB")` and asserts it does not throw. As corrected in 1.1, the specific risk this was meant to catch wasn't actually live, but the test is still a legitimate regression guard and matches the `KafkaFactoryTest` pattern I pointed to. **Issue 5(a): resolved (test added as recommended).**
* **Negative-value assertion tightened.** `testNegativeMaxRetriesFails` now asserts both `exception.getMessage().contains(MAX_RETRIES.key())` and `exception.getMessage().contains(">= 0")` (`:153-160`), closing the "would also pass for an unrelated error mentioning the key" gap I raised. **Issue 5(b): resolved.**
* **Class-level Javadoc added, helper methods still bare.** The class now carries `/** Tests declarative sink validation exposed by {@link AmazonDynamoDBSinkFactory}. */` (`:128`). That's the class-level half of Issue 7. The three private helpers (`validateMaxRetries`, `requiredOptions`, `validate`, `:173-191`) still have no Javadoc explaining what each represents (e.g. that `validate` mirrors the dry-run entry point). **Issue 7: partially resolved** — worth a follow-up one-liner each, but not blocking.

**Stability rating: Stable.** Same as last round — pure in-memory `HashMap` → `ReadonlyConfig` → `ConfigValidator`/`OptionRule` calls, `sinkRule` rebuilt per test in `@BeforeEach`, no I/O, containers, sleeps, clocks, or shared mutable state. Nothing here can flake in CI, and the new tests (`testNonNegativeMaxRetries`, `testMaxRetriesIsADeclaredOption`) follow the same pattern. CI on this exact head (`d2d43b4a4e`) shows `Build` = SUCCESS (completed 2026-08-16T07:13:57Z, run 31925251758), and the author's follow-up comment states the full module passes 7 tests — consistent with 4 tests in this file plus the pre-existing 3 in `AmazonDynamoDBSourceFactoryTest`.

## 2.3 Documentation Updates

**Issue 1 from my last review: resolved.** Both `docs/en/connectors/sink/AmazonDynamoDB.md` and `docs/zh/connectors/sink/AmazonDynamoDB.md` now state the constraint in the options table (`... Must be at least 0.` / `... 必须大于等于 0。`) and in the `### max_retries [int]` prose section (`A value of 0 disables retries. The value must not be negative.` / `设置为 0 表示不重试，该值不能为负数。`). Wording is consistent between languages and matches the actual `>= 0` behavior exactly — no drift between doc and code.

# 3. Architectural Soundness

Unchanged assessment: this remains the correct migration shape per the `#11007` umbrella and the `connector-console` (#11477) precedent, and it still stops short of "complete" only at the edges (upper bound, source-path parity) rather than in the core mechanism.

## 3.2 Maintainability

Improved slightly versus last round: the test class now documents its own purpose at the class level, and the default-value and unknown-key assertions make the suite's intent legible without having to reverse-engineer it from `ConfigValidator` internals.

## 3.3 Extensibility

Unchanged: `Conditions.and(...)` / `lessOrEqualField` remain available and unused for the `retry_base_delay_ms` vs `retry_max_delay_ms` ordering constraint and for an upper bound on `max_retries`, should this or a follow-up PR pick them up.

**No risk**, re-confirmed: the retry-backoff option family, including `max_retries`, remains absent from the `2.3.13` release tag. No shipped config, no checkpoint/savepoint state, no SPI/registration surface is touched by either commit in this round.

# 4. Issue Summary

| # | Issue | Location | Status | Severity |
|---|---|---|---|---|
| 1 | `max_retries` constraint undocumented | `docs/en`/`docs/zh` `AmazonDynamoDB.md` | **Resolved** | was Medium (blocker) |
| 2 | No upper bound; backoff shift overflows and crashes `Thread.sleep` at high retry counts | `.../sink/DynamoDbSinkClient.java:165` (`flushWithRetry`); `.../sink/AmazonDynamoDBSinkFactory.java:59` | **Open (carryover)** | Medium |
| 3 | Default-value test was vacuous | `AmazonDynamoDBSinkFactoryTest.java` | **Resolved** | was Low |
| 4 | Validation error omits the offending value | `AmazonDynamoDBSinkFactory.java:59` (framework: `OptionUtil.formatError`) | **Open (carryover, informational)** | Low |
| 5 | No `validateUnknownKeys` coverage; weak negative assertion | `AmazonDynamoDBSinkFactoryTest.java` | **Resolved** | was Low |
| 6 | Source construction path (`AmazonDynamoDBConfig`) still has no equivalent constraint | `.../config/AmazonDynamoDBConfig.java:66`; `.../source/AmazonDynamoDBSourceFactory.java` | **Open (carryover, informational)** | Low |
| 7 | Test class/helpers lack Javadoc | `AmazonDynamoDBSinkFactoryTest.java:173-191` | **Partially resolved** (class Javadoc added; helpers still bare) | Low |
| 8 | Umbrella `#11007` connector unclaimed | `#11007` issue comments | **Resolved** — claimed 2026-08-15T18:57:07Z, ~44 min before this PR was opened | was Low (process) |
| 9 | My own prior Issue 5 overstated the "dropped declared key" risk | `ConfigValidator.java:127-143`, `OptionRule.java:346-356` | Self-correction, no action needed | N/A |

---

### Issue 2 — No upper bound; backoff shift still overflows (carryover, unchanged)

* **Location:** `seatunnel-connectors-v2/connector-amazondynamodb/src/main/java/org/apache/seatunnel/connectors/seatunnel/amazondynamodb/sink/DynamoDbSinkClient.java:165` (`long delay = Math.min(baseDelayMs * (1L << retryCount), maxDelayMs);`); `.../sink/AmazonDynamoDBSinkFactory.java:59`.
* **Problem description:** Unchanged from my last review — `max_retries` only gets a lower bound (`>= 0`). At high retry counts the `1L << retryCount` shift overflows `long`, `Math.min` selects the resulting negative value, and `Thread.sleep(delay)` at line 180 throws `IllegalArgumentException: timeout value is negative` instead of the intended "failed after N retries" error.
* **Potential risk:** Same as before — reachable with e.g. `max_retries = 100` on a hot-partitioned table; the failure mode becomes an opaque `Thread.sleep` crash rather than the clear `RuntimeException` the retry loop is designed to produce.
* **Best improvement:** Same recommendation as last round — chain `Conditions.lessOrEqual(MAX_RETRIES, 30)` (or similar) onto the existing constraint at line 59, or clamp the shift in `flushWithRetry`. Since `max_retries` is still unreleased, this remains free to fix now and becomes an incompatible change once released.
* **Severity:** Medium.
* **Raised by another reviewer:** No.
* **Status:** Open — not addressed in this round; not a merge blocker (consistent with my prior review, which also treated it as non-blocking).

### Issue 6 — Source construction path still uncovered (carryover, unchanged)

* **Location:** `.../config/AmazonDynamoDBConfig.java:66`; `.../source/AmazonDynamoDBSourceFactory.java` (`optionRule`).
* **Problem description:** Unchanged — `AmazonDynamoDBSourceFactory.optionRule()` still declares neither `MAX_RETRIES` nor a constraint on it, even though the shared `AmazonDynamoDBConfig` constructor (used by both source and sink) reads it.
* **Potential risk:** Same as before — low in practice (`validateUnknownKeys` already flags `max_retries` as unknown on a source-only block; a `-1` reaching `flushWithRetry` fails loudly, not silently).
* **Best improvement:** No fix required for merge; note it in the PR description for whoever migrates the source half.
* **Severity:** Low.
* **Raised by another reviewer:** No.
* **Status:** Open, informational.

# 5. Merge Recommendation

Both items I flagged as blockers last round are resolved on independent re-verification: the docs now state the `>= 0` constraint in both languages with matching wording, and CI on the current head (`d2d43b4a4e`) is green (`Build` = SUCCESS, completed 2026-08-16T07:13:57Z). Three of the four non-blocking recommendations were also picked up (default-value pinning, `validateUnknownKeys` coverage, tightened negative-message assertion), and the test class gained class-level Javadoc. No maintainer review has been posted yet on this PR, and `mergeStateStatus` is `BLOCKED` only because branch protection requires a human review (`reviewDecision: REVIEW_REQUIRED`) — `mergeable` is `MERGEABLE`, i.e. no real conflict with `dev`.

1. Blockers — must be fixed: none remaining.
   * **Issue 2** (Medium) — add an upper bound on `max_retries` or clamp the backoff shift in `flushWithRetry`; still free to do while the option is unreleased.
   * **Issue 7** (Low) — one-line Javadoc on `validateMaxRetries`/`requiredOptions`/`validate` explaining what each represents.
   * **Issue 4 / Issue 6** (Low, informational) — no action required for this PR; carried forward for awareness of anyone extending this migration to the source side or to `ConfigValidator`'s error formatting.

Good, thorough follow-up — you addressed the blocker and picked up nearly every non-blocking suggestion in one pass, including the parts that required actually reading how `ConfigValidator`/`OptionRule` behave rather than just satisfying the letter of the review. Nice work.

goutamadwant:
@DanielLeens if this looks good can you change your status to approve?
@nzw921rx let me know if this PR looks good or any suggestions. thanks!

nzw921rx (approved):
+1 LGTM

DanielLeens (approved):
Thanks for the ping, @goutamadwant — happy to formalize this.

To recap for the record: in my last re-review round (head `d2d43b4a4e`) both blockers I raised were confirmed resolved on independent re-verification — the `max_retries >= 0` constraint is now documented in both `docs/en` and `docs/zh` with matching wording, and CI (`Build`) was green on that head. Three of the four non-blocking suggestions were also picked up (default-value pinning via `testOmittedMaxRetriesUsesDefault`, `validateUnknownKeys` coverage, and a tightened negative-value message assertion), plus class-level Javadoc on the test class. Nothing in this round changes that assessment — no new commits have landed since, so I'm not re-doing the full walkthrough, just converting my "Ready to merge" conclusion into an actual approval.

Formally approving. The only remaining items are non-blocking follow-ups already on record (an upper bound / overflow guard on `max_retries` in `DynamoDbSinkClient.flushWithRetry`, and a couple of Javadoc nits) — worth a fast-follow, not a reason to hold this up.

@nzw921rx thanks for taking a look as well.

nzw921rx (approved):
+1 LGTM

Issue #10172, opened 2025-12-10
[Bug] [Connector-V2-File/HDFS] NoSuchMethodError for FsTracer.get with Hadoop 3.1.1 and Kerberos


rait-winter:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/seatunnel/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### Problem Description
When using SeaTunnel to write data from PostgreSQL to HDFS with Kerberos authentication enabled, the task fails to start and throws the exception `java.lang.NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get`. The Kerberos authentication itself is successful, but a Hadoop API incompatibility error occurs when initializing the HDFS client.

### Environment Information
*   **SeaTunnel Version**: 2.3.8
*   **Running Mode**: SeaTunnel Engine (Local)
*   **Java Version**: 1.8.0_291
*   **Source Database**: PostgreSQL
*   **Target HDFS Version**: Hadoop 3.1.1 (with Kerberos enabled)
*   **Client Hadoop Environment**: Not explicitly specified, suspected to have a version conflict with the built-in SeaTunnel version or the version from environment variables.

### Task Configuration File (`config/pg-to-hdfs.conf`)
```hocon
env {
  job.mode = "BATCH"
  execution.parallelism = 1
}

source {
  Jdbc {
    url = "jdbc:postgresql://157.3.xx.xx:5432/sqjw?gssEncMode=disable&sslmode=disable"
    driver = "org.postgresql.Driver"
    user = "root"
    password = "******"
    query = "SELECT * from tb limit 100"
    fetch_size = 1000
    result_table_name = "tb_data"
  }
}

sink {
  HdfsFile {
    path = "/db/hive/warehouse/YBSS/test-seatunnel"
    file_format_type = "orc"
    fs.defaultFS = "hdfs://hacluster"
    dfs.client.use.datanode.hostname = "true"
    hadoop.security.authentication = "kerberos"
    kerberos_principal = "Botest@HADOOP.COM"
    krb5_path = "/opt/krb5.conf"
    kerberos_keytab_path = "/opt/user.keytab"
# I also tried to configure using the hadoopConfig {} block, but the problem remains the same.  
 }
}

### SeaTunnel Version

2.3.8

### SeaTunnel Config

```conf
env {
  job.mode = "BATCH"
  execution.parallelism = 1
}

source {
  Jdbc {
[...]

PR #11823, opened 2026-08-16
[Improve][Connector-V2] Diagnose Hadoop client version mismatch on FileSystem init

RohanExploit:

Closes the diagnosability gap in #10172.

`HadoopFileSystemProxy` acquires a `FileSystem` in three places — the plain path in `initialize()`, the Kerberos path in `initializeWithKerberosLogin()`, and the proxy-user path in `initializeWithRemoteUserLogin()`. All three call `FileSystem.get(configuration)`, which constructs a `DFSClient` that reaches into Hadoop internals (e.g. `FsTracer`) whose method signatures have changed across Hadoop releases.

The Hadoop client jars are a `provided` dependency supplied by the deployment environment, not bundled here. When they don't match the Hadoop version SeaTunnel was built against (Hadoop 3.1.4, via `seatunnel-hadoop3-3.1.4-uber`), the failure surfaces as a bare `LinkageError` deep inside Hadoop's own code with nothing pointing at the real cause — in #10172 the user saw only `NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get`.

All three call sites now route through `getFileSystemWithDiagnostics()`, which rewrites the error into an `IOException` naming the likely cause and the Hadoop version this connector targets, keeping the original error as `getCause()`. The failure mechanism is not specific to any one authentication mode, so scoping the diagnostic to a single path would have left the other two opaque.

The catch covers the whole `LinkageError` family rather than `NoSuchMethodError`/`NoClassDefFoundError` alone: `NoSuchFieldError` and `IncompatibleClassChangeError` are equally plausible symptoms of the same jar mismatch. Errors outside that family — `OutOfMemoryError`, `StackOverflowError` — are not caught and propagate unchanged, as does an `IOException` raised by the filesystem itself.

The working path is untouched — this is diagnostic only, not a behavior change.

No change on success. On this specific failure the user sees a clear `IOException` explaining the likely version mismatch instead of a bare JVM linkage error. This now applies to all three initialization paths, not just Kerberos.

Unit tests in `HadoopFileSystemProxyTest`. All are deterministic — no I/O, no sleeps, no new dependencies, plain `@Test`.

Rewrite coverage, each asserting the original error is preserved as `getCause()` and the message names the Hadoop client version mismatch (they share an `assertRewrittenAsDiagnosticIOException(LinkageError)` helper):

- `testWrapClasspathMismatchRewritesNoSuchMethodErrorAsDiagnosticIOException`
- `testWrapClasspathMismatchRewritesNoClassDefFoundError`
- `testWrapClasspathMismatchRewritesNoSuchFieldError`
- `testWrapClasspathMismatchRewritesIncompatibleClassChangeError`

Boundaries of the rewrite:

- `testWrapClasspathMismatchDoesNotSwallowUnrelatedErrors` — an `OutOfMemoryError` propagates as the same instance, unwrapped. This is the regression test that makes widening the catch to `LinkageError` safe to review.
- `testWrapClasspathMismatchPropagatesIOExceptionUnchanged` — a plain `IOException` from the supplier comes back identical, with a null cause.
- `testWrapClasspathMismatchPassesThroughOnSuccess` — the success path is unchanged.

Verified locally with `spotless:check` and `mvn test -Dtest=HadoopFileSystemProxyTest` on the `connector-file-base` module.

---

Disclosure: this change was AI-assisted (Claude). I reviewed it and can speak to the reasoning.


DanielLeens (commented):

When SeaTunnel connects to a Kerberos-enabled HDFS cluster and the Hadoop client jars resolved at runtime (a `provided` dependency supplied by the deployment environment) don't match the Hadoop version SeaTunnel was built against (Hadoop 3.1.4, via `seatunnel-hadoop3-3.1.4-uber`), `FileSystem.get()` can fail deep inside Hadoop's own `DFSClient`/`FsTracer` construction with a raw `NoSuchMethodError` or `NoClassDefFoundError` — see the linked report, issue #10172, where a user only saw `NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get` with no indication of the real cause. This PR wraps the Kerberos `FileSystem.get()` call and rethrows those two link-time errors as an `IOException` that names the likely cause and the Hadoop version the connector targets, while leaving the success path completely untouched. One-sentence summary: it turns an opaque JVM linkage crash into an actionable diagnostic message, scoped to the exact failure originally reported.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**File**: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/hadoop/HadoopFileSystemProxy.java`

Before:
```java
this.fileSystem = FileSystem.get(configuration);
```
(inside the Kerberos login callback in `initializeWithKerberosLogin()`)

After:
```java
this.fileSystem = getFileSystemWithDiagnostics(configuration);
...
private static FileSystem getFileSystemWithDiagnostics(Configuration configuration)
        throws IOException {
    return wrapClasspathMismatch(() -> FileSystem.get(configuration));
}

static FileSystem wrapClasspathMismatch(FileSystemSupplier supplier) throws IOException {
    try {
        return supplier.get();
    } catch (NoSuchMethodError | NoClassDefFoundError e) {
        throw new IOException(
                "Failed to initialize Hadoop FileSystem, likely due to a Hadoop client "
                        + "version mismatch ... Hadoop 3.1.4 ...", e);
    }
}
```

**Key Findings**
- The wrapper is applied only to `initializeWithKerberosLogin()` (line 463); the identical `FileSystem.get(configuration)` call in the plain `initialize()` path (line 381) and in `initializeWithRemoteUserLogin()` (line 517) are left unwrapped — see Issue 1.
- The cause of the classpath mismatch (`DFSClient`/`FsTracer` construction inside `FileSystem.get()`) is not Kerberos-specific; it fires for any `hdfs://` filesystem, authenticated or not, so the fix's title/scope ("...on Kerberos FileSystem init") is narrower than the actual failure mechanism it targets.
- The hardcoded "Hadoop 3.1.4" claim in the message is accurate: I confirmed `connector-file-base/pom.xml` declares `seatunnel-hadoop3-3.1.4-uber` as the only Hadoop uber dependency for this module (provided/optional scope), so the message won't mislead users of this module.
- The original `Error` (not `Exception`) is preserved as the `IOException`'s cause via the two-arg constructor — no information is lost, nothing is swallowed.
- The catch clause is narrowly scoped to `NoSuchMethodError`/`NoClassDefFoundError`; it does not catch other `LinkageError` siblings (`NoSuchFieldError`, `IncompatibleClassChangeError`) that the same jar-mismatch scenario can also produce — see Issue 2 (minor).

**Runtime flow trace** (failure path only; success path is byte-for-byte unchanged):
```
HadoopFileSystemProxy(HadoopConf)                      [constructor, eager init]
  -> initialize()                              [@SneakyThrows]
    -> initializeWithKerberosLogin()           [throws IOException, InterruptedException]
      -> HadoopLoginFactory.loginWithKerberos(..., loginFunction)
        -> loginFunction.run(...) [lambda]
          -> getFileSystemWithDiagnostics(configuration)
            -> wrapClasspathMismatch(() -> FileSystem.get(configuration))
              -> FileSystem.get(configuration)
                -> DistributedFileSystem.initialize(...) -> new DFSClient(...) -> FsTracer.get(...)
                   [signature changed across Hadoop releases]
                -> NoSuchMethodError / NoClassDefFoundError thrown (unchecked Error)
              <- caught, rethrown as IOException(msg, cause=Error)
      <- IOException propagates out of loginFunction, through loginWithKerberos
    <- IOException propagates (declared) out of initializeWithKerberosLogin()
  <- @SneakyThrows lets the checked IOException escape initialize() without a throws clause
<- constructor throws; surfaces to the connector's source/sink open() call, failing task init
   with a message the user can act on instead of a bare Hadoop-internal stack trace.
```
This sits on connector initialization (task startup), a normal execution path when Kerberos is configured — but the specific branch that triggers the wrapping (a jar-version mismatch) is itself an environment/config error condition, not something exercised on every successful run. That matches the PR description ("no change on success").

## 1.2 Compatibility Impact

**Fully compatible.** No config option, default value, public API, or serialization format is touched. The only behavioral change is on an already-fatal failure path: what previously escaped as an unchecked `Error` (crashing task initialization with an opaque stack trace) now escapes as a checked `IOException` with the same `Error` preserved as `getCause()`. I checked whether SeaTunnel's Zeta engine treats `Error` and checked exceptions differently for job-failure/retry classification (e.g. `ExceptionUtil`, `TaskExecutionService`); the relevant catch sites use generic `catch (Throwable ...)` and don't special-case `LinkageError` for retry decisions (only `OutOfMemoryError` gets special dispatch, which is untouched here), so converting the Error to an IOException does not change how the job ultimately fails.

Negligible. The change adds one try/catch around an already-existing method call on the (rare) Kerberos-login initialization path — no allocation, no I/O, no locking, no retries added on the success path. No effect on CPU/memory/GC/networking/concurrency.

The original error is not swallowed (preserved as `getCause()`), no new logging is introduced (appropriate — the caller already logs/reports failures), and no sensitive information (credentials, tokens) is included in the new message. Formal issues below.

### Issue 1
- **Location**: `HadoopFileSystemProxy.java:381` (`initialize()`) and `HadoopFileSystemProxy.java:517` (`initializeWithRemoteUserLogin()`)
- **Problem description**: Both of these call sites do `this.fileSystem = FileSystem.get(configuration);` directly, without going through `getFileSystemWithDiagnostics()`. They construct the exact same `DFSClient`/`FsTracer` machinery as the Kerberos path and are subject to the exact same Hadoop-client-version-mismatch failure mode (`NoSuchMethodError`/`NoClassDefFoundError`). Only the Kerberos branch gets the diagnostic rewrite; a user connecting to a version-mismatched HDFS cluster without Kerberos (the plain path) or via a Hadoop proxy/remote user (the `enableRemoteUser()` path) still gets the original opaque JVM linkage error this PR set out to eliminate.
- **Potential risk**: The PR's stated goal — "closes the diagnosability gap in #10172" — is only partially achieved. Two of the three `FileSystem.get()` call sites in this class remain undiagnosed, so most non-Kerberos deployments hitting the same classpath problem get no benefit from this change.
- **Best improvement**:
  - Option A (minimal, consistent with the existing design): replace `fileSystem = FileSystem.get(configuration);` at line 381 with `fileSystem = getFileSystemWithDiagnostics(configuration);`, and similarly replace `this.fileSystem = FileSystem.get(configuration);` at line 517 with `this.fileSystem = getFileSystemWithDiagnostics(configuration);`. Both call sites already declare/propagate `IOException` (`initialize()` is `@SneakyThrows`, `initializeWithRemoteUserLogin()` is declared `throws Exception`), so this is a drop-in, zero-risk change reusing the helper that already exists.
  - Option B: keep the current Kerberos-only scope for this PR (since it directly closes the reported issue #10172) and file a fast follow-up for the other two call sites. Acceptable, but Option A is essentially free given the helper is already written and tested.
- **Severity**: High

### Issue 2
- **Location**: `HadoopFileSystemProxy.java:489-501` (`wrapClasspathMismatch`)
- **Problem description**: The catch clause is `catch (NoSuchMethodError | NoClassDefFoundError e)`. These are two of several `LinkageError` subtypes that a Hadoop client/runtime jar mismatch can produce; `NoSuchFieldError` and `IncompatibleClassChangeError` are equally plausible symptoms of the same root cause (e.g., a changed field or class hierarchy rather than a changed method signature) and would still escape as an undiagnosed `Error` today.
- **Potential risk**: Low — the two errors caught are the most common symptoms (and match the exact exception from #10172), so this is a coverage gap rather than a correctness bug.
- **Best improvement**: Widen the catch to `catch (LinkageError e)`, which covers `NoSuchMethodError`, `NoClassDefFoundError`, `NoSuchFieldError`, and `IncompatibleClassChangeError` uniformly while still excluding unrelated `Error`s (e.g. `OutOfMemoryError`, `StackOverflowError`) that shouldn't be attributed to a Hadoop version mismatch.
- **Severity**: Low

# 2. Code Quality Assessment

## 2.1 Coding Standards

Formatting and naming are consistent with the surrounding class (AOSP style, no wildcard imports). Both new methods (`getFileSystemWithDiagnostics`, `wrapClasspathMismatch`) are non-trivial (they implement the diagnostic rewrite this PR is about) but only carry a block `//` comment above them explaining the overall rationale, not a per-method Javadoc with `@param`/`@return`/`@throws`. This is a minor nit — the existing comment already explains the "why" clearly — but per this repo's own contribution guidance ("Add comments for important methods"), a short Javadoc on `wrapClasspathMismatch` (the package-visible, directly-tested entry point) would be a nice-to-have. Not a blocker. The new `FileSystemSupplier` functional interface is appropriately package-private and minimal.

Two unit tests are added to `HadoopFileSystemProxyTest`:
- `testWrapClasspathMismatchRewritesNoSuchMethodErrorAsDiagnosticIOException` — directly exercises the catch/rewrap logic with a synthetic `NoSuchMethodError`, asserts the original is preserved as `getCause()` and the message mentions "Hadoop client"/"version mismatch". Deterministic, no I/O, no timing dependency.
- `testWrapClasspathMismatchPassesThroughOnSuccess` — exercises the pass-through (non-error) branch against a real local filesystem (`file:///`), consistent with the existing tests in this file that also use `HadoopConf("file:///")`.

Both tests are synchronous, single-threaded, use no `Thread.sleep`, no shared static state, and don't depend on execution order — no flaky-test anti-patterns present. **Test stability rating: Stable.**

One coverage gap worth noting (non-blocking): neither test exercises `getFileSystemWithDiagnostics()` or the real `initializeWithKerberosLogin()` code path end-to-end (understandably, since that requires a live/mini-KDC, and the existing test suite in this file doesn't test the Kerberos path at all either). The tests validate the `wrapClasspathMismatch` helper in isolation, which is a reasonable and consistent choice given the existing test infrastructure, but it means the wiring between `initializeWithKerberosLogin()` and the new helper (i.e., that the real call site actually uses it) is only verified by code inspection, not by a test.

## 2.3 Documentation Updates

Not required. This is a diagnostics-only change to an internal error path; there is no new config option, no default-value change, and no public API/behavior change on the success path for `docs/en` or `docs/zh` to describe.

# 3. Architectural Soundness

Precise, minimal fix for the specific case it covers: a small `try/catch` wrapper reusing a `@FunctionalInterface` to keep the diagnostic logic in one testable, reusable place (`wrapClasspathMismatch`). Given Issue 1, the current PR is a **precise fix for the Kerberos branch, but a partial/incomplete fix for the diagnosability gap it claims to close** — the underlying mechanism (`wrapClasspathMismatch`) is already general enough to be reused for all three `FileSystem.get()` call sites in this class with a one-line change each.

## 3.2 Maintainability

Good. The helper is a small, self-contained static method with a clear single responsibility, and it's directly unit-tested rather than only reachable through the hard-to-test Kerberos path. Extending it to the other two call sites (Issue 1) would not add any new maintenance burden.

## 3.3 Extensibility

The `FileSystemSupplier` functional interface makes it trivial to wrap any `FileSystem.get(...)`-shaped call, including the other two call sites in this same file, or similar `FileSystem` acquisition points in sibling connectors (e.g., other file connectors built on this shared base) if they ever need the same diagnostic treatment.

No impact. No config, checkpoint/savepoint state, serialization format, or SPI/registration surface is touched. Existing saved jobs, historical task configs, and upgrade paths are unaffected — this only changes the text of an exception thrown on an already-fatal Kerberos-init failure.

# 4. Issue Summary

| 1 | Diagnostic wrapping only covers the Kerberos `FileSystem.get()` call; the plain-auth and remote-user paths hit the identical failure mode unwrapped | `HadoopFileSystemProxy.java:381`, `HadoopFileSystemProxy.java:517` | High |
| 2 | Catch clause covers only `NoSuchMethodError`/`NoClassDefFoundError`, not the full `LinkageError` family (`NoSuchFieldError`, `IncompatibleClassChangeError`) that the same jar mismatch can also produce | `HadoopFileSystemProxy.java:489-501` | Low |

# 5. Merge Recommendation

   - Issue 1 (High): Extend the diagnostic wrapping to the plain `initialize()` path (line 381) and `initializeWithRemoteUserLogin()` (line 517) by routing both through the already-written `getFileSystemWithDiagnostics()` helper. This is a two-line change with no new risk, and it's necessary for the PR to actually deliver on its stated goal of closing the diagnosability gap from #10172 rather than covering only one of three equivalent call sites.

   - Issue 2 (Low): Widen the catch clause to `LinkageError` for more complete coverage of classpath-mismatch symptoms.
   - Add a short Javadoc (`@param`, `@return`/`@throws`) on `wrapClasspathMismatch`, the directly-tested entry point, per this repo's convention of commenting non-trivial methods.

**Overall assessment**: This is a clean, well-scoped, low-risk diagnostic improvement with solid, stable unit tests and zero impact on the success path or on backward compatibility. The core idea (rewrap link-time classpath errors into an actionable `IOException`) is exactly the right fix for #10172, and the implementation is careful about preserving the original cause and not touching the happy path. The only real gap is scope: the fix is applied to one of three structurally-identical call sites in the same class, so as written it only helps Kerberos-authenticated users hitting this specific failure, even though the same helper trivially extends to the other two. I'd like to see Issue 1 addressed before merge — it's a small, safe change that makes the fix actually complete for the class of problem it targets. Thanks for tackling a real pain point here, and for including regression tests that fail before the change and pass after — that's exactly the right way to demonstrate the fix.

---

**CI note**: The current head commit's "Build" check on the apache/seatunnel side shows `action_required` with "Workflow run detection failed" — I confirmed this is because GitHub Actions has never run in your fork (`RohanExploit/seatunnel` has zero total workflow runs), not because of a failure in your code. Please enable GitHub Actions in your fork's Settings, and since your branch is currently 7 commits behind `dev`, it's also worth syncing/rebasing onto the latest `dev` before pushing again so CI runs against current `dev`. Once Actions is enabled and a new commit is pushed, the Build check should pick up your fork's real CI run.

davidzollo:
Good job.

To pass the automated CI checks for this PR, could you please enable CI workflows in your forked repository first? You can refer to the check run link below as a reference for the required CI configuration  https://github.com/apache/seatunnel/pull/11823/checks?check_run_id=95123508319

image

davidzollo:

```

```

RohanExploit:
@DanielLeens thanks for the detailed review.

Issue 1 was a real scope miss on my part, not a deliberate narrowing. The linkage failure happens while `FileSystem.get()` builds the `DFSClient` and `FsTracer`, which has nothing to do with Kerberos, so wrapping only that branch left the other two call sites just as opaque as before. I went with Option A. `initialize()` and `initializeWithRemoteUserLogin()` now both go through `getFileSystemWithDiagnostics()`. I checked the propagation instead of assuming it: `initialize()` is `@SneakyThrows` and `HadoopLoginFactory.LoginFunction#run` declares `throws Exception`, so both were a straight swap. Grepping for `FileSystem.get(configuration)` in the file now returns one hit, the call inside the helper.

Issue 2 is fixed by widening the catch to `LinkageError`. `NoSuchFieldError` and `IncompatibleClassChangeError` come out of the same jar mismatch, and `LinkageError` still leaves `OutOfMemoryError` and `StackOverflowError` alone. The `IOException` message body is unchanged.

Javadoc is on both `getFileSystemWithDiagnostics` and `wrapClasspathMismatch` now, replacing the old block comment. It says the failure is not specific to any one authentication mode and that all three call sites route through the helper, and it refers to `LinkageError` rather than `NoSuchMethodError`.

New tests:

* `testWrapClasspathMismatchRewritesNoClassDefFoundError`
* `testWrapClasspathMismatchRewritesNoSuchFieldError`
* `testWrapClasspathMismatchRewritesIncompatibleClassChangeError`
* `testWrapClasspathMismatchDoesNotSwallowUnrelatedErrors`, which throws an `OutOfMemoryError` and asserts the same instance comes back out unwrapped. That one is what makes the widening to `LinkageError` reviewable.
* `testWrapClasspathMismatchPropagatesIOExceptionUnchanged`, where a plain `IOException` from the supplier comes back identical with a null cause.

The four rewrite tests share a private `assertRewrittenAsDiagnosticIOException(LinkageError)` helper that asserts the original is the cause and the message names the Hadoop client version mismatch. The existing `NoSuchMethodError` test uses it too. All plain `@Test`, no I/O, no sleeps, no new dependencies. I sanity checked them by putting the old two type catch back: `NoSuchFieldError` and `IncompatibleClassChangeError` fail, everything else passes.

You were right that the title was narrower than the mechanism. Retitled it and rewrote the PR body around all three call sites.

RohanExploit:
@davidzollo thanks. GitHub Actions is enabled in my fork now, so CI runs on this push. I also rebased the branch onto current `dev`, and @DanielLeens review feedback is addressed in the same push. I'll join the #contributors channel on Slack.

DanielLeens (commented):
Thanks for the quick and thorough turnaround, @RohanExploit — I re-reviewed the full diff at the latest head against `dev` from scratch. Both items from my last pass check out exactly as you described: `initialize()` (line 381) and `initializeWithRemoteUserLogin()` (line 546) now route through `getFileSystemWithDiagnostics()` just like the Kerberos path, and the catch clause in `wrapClasspathMismatch` is widened to `LinkageError`. I grepped the file myself and confirmed there is now exactly one raw `FileSystem.get(configuration)` call left, inside the helper itself — the diagnostic wrapping genuinely covers all three initialization paths now, which is what actually closes the gap from #10172. The added Javadoc on both methods is also a nice touch, and the new tests for `NoClassDefFoundError`/`NoSuchFieldError`/`IncompatibleClassChangeError` plus the `OutOfMemoryError` and `IOException` pass-through regression tests are exactly the right shape to make the widened catch reviewable.

While re-tracing the full call chain for this round, I did find one nuance in the widened catch that's worth flagging — see Issue 1 below. It's a byproduct of my own suggestion last round to widen the catch to `LinkageError`, and on closer inspection that suggestion was itself slightly imprecise about how broad `LinkageError` actually is, so I want to be upfront about that rather than present it as something you should have caught.

**User pain point**: When the Hadoop client jars resolved at runtime (a `provided` dependency supplied by the deployment environment) don't match the Hadoop version SeaTunnel's file connectors are built against (Hadoop 3.1.4, via `seatunnel-hadoop3-3.1.4-uber`), `FileSystem.get()` can fail deep inside Hadoop's own `DFSClient`/`FsTracer` construction with a raw JVM linkage error. Issue #10172 shows exactly this: the user only saw `NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get`, with nothing pointing at the real cause (a classpath/version mismatch) or the connector's expected Hadoop version.

**Fix approach**: All three places in `HadoopFileSystemProxy` that call `FileSystem.get(configuration)` — the plain path (`initialize()`), the Kerberos path (`initializeWithKerberosLogin()`), and the proxy-user path (`initializeWithRemoteUserLogin()`) — now go through a shared `getFileSystemWithDiagnostics()` helper. It catches `LinkageError` around the `FileSystem.get()` call and rethrows it as an `IOException` that names the likely cause and the Hadoop version the connector targets, preserving the original `LinkageError` as `getCause()`. Nothing changes on the success path.

**One-sentence summary**: This turns an opaque, unguided JVM linkage crash into an actionable diagnostic message, consistently across all three ways this connector can acquire a Hadoop `FileSystem`.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**File**: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/hadoop/HadoopFileSystemProxy.java`

Before (all three sites, e.g. `initialize()` at the old line 381):
```java
fileSystem = FileSystem.get(configuration);
```

After:
```java
// initialize() – line 381
fileSystem = getFileSystemWithDiagnostics(configuration);

// initializeWithKerberosLogin() – line 463
this.fileSystem = getFileSystemWithDiagnostics(configuration);

// initializeWithRemoteUserLogin() – line 546
this.fileSystem = getFileSystemWithDiagnostics(configuration);

private static FileSystem getFileSystemWithDiagnostics(Configuration configuration)
        throws IOException {
    return wrapClasspathMismatch(() -> FileSystem.get(configuration));
}

static FileSystem wrapClasspathMismatch(FileSystemSupplier supplier) throws IOException {
    try {
        return supplier.get();
    } catch (LinkageError e) {
        throw new IOException(
                "Failed to initialize Hadoop FileSystem, likely due to a Hadoop client "
                        + "version mismatch between SeaTunnel and the Hadoop jars resolved "
                        + "at runtime ... This connector is built and tested against "
                        + "Hadoop 3.1.4 (see seatunnel-hadoop3-3.1.4-uber in the reactor pom) ...",
                e);
    }
}
```

**Key findings**
- All three `FileSystem.get(configuration)` call sites in this class are now wrapped; a grep for `FileSystem.get(` in the current head returns exactly one hit, inside `getFileSystemWithDiagnostics` itself. This is a real, complete fix for the diagnosability gap in #10172, not a partial one scoped to Kerberos only.
- The catch clause is now `catch (LinkageError e)`, which is a strict superset of the previous `NoSuchMethodError | NoClassDefFoundError`. See Issue 1 for a precision concern on how broad that superset actually is.
- The original error is never swallowed — it's preserved as `getCause()` on the new `IOException` in every rewrite path, verified both by the four dedicated rewrite tests and by direct code reading.
- `testWrapClasspathMismatchDoesNotSwallowUnrelatedErrors` (asserting `OutOfMemoryError` passes through as the same instance, unwrapped) is exactly the regression test needed to make the `LinkageError` widening safe to merge — it proves the catch doesn't over-reach into unrelated JVM-fatal errors like `OutOfMemoryError`/`StackOverflowError`.
- Both new methods now carry full Javadoc (`@param`/`@return`/`@throws`) rather than a block comment, addressing the "nice-to-have" from my previous round.

**Runtime flow trace** (failure path; success path is unchanged for all three modes):
```text
HadoopFileSystemProxy(HadoopConf)                         [constructor, eager init]
  -> initialize()                                  [@SneakyThrows]
    +-- enableKerberos() == true
    |     -> initializeWithKerberosLogin()
    |       -> HadoopLoginFactory.loginWithKerberos(..., loginFunction)
    |         -> loginFunction.run(...) -> getFileSystemWithDiagnostics(configuration)
    +-- enableRemoteUser() == true
    |     -> initializeWithRemoteUserLogin()
    |       -> HadoopLoginFactory.loginWithRemoteUser(..., loginFunction)
    |         -> loginFunction.run(...) -> getFileSystemWithDiagnostics(configuration)
    +-- neither (plain path)
          -> getFileSystemWithDiagnostics(configuration)   [directly, line 381]

  getFileSystemWithDiagnostics(configuration)
    -> wrapClasspathMismatch(() -> FileSystem.get(configuration))
      -> FileSystem.get(configuration)
        -> DistributedFileSystem.initialize(...) -> new DFSClient(...) -> FsTracer.get(...)
           [signature/shape changed across Hadoop releases]
        -> LinkageError thrown (unchecked)
      <- caught, rethrown as IOException(diagnostic message, cause=LinkageError)
  <- IOException propagates out through whichever of the three initialize*() paths was taken,
     out of initialize() via @SneakyThrows, and fails connector/task init with an actionable
     message instead of a bare Hadoop-internal stack trace — for all three auth modes now,
     not just Kerberos.
```
This sits on connector initialization (task startup) — a real, normal-use code path — but the specific branch that triggers the rewrite (a jar-version/classpath mismatch) is itself an environment/deployment error condition, consistent with "no change on success" in the PR description.

## 1.2 Compatibility Impact

**Fully compatible.** No config option, default value, public API, or serialization format is touched. The only behavioral change is on an already-fatal failure path: what previously escaped as an unchecked `Error` now escapes as a checked `IOException` with the same `Error` preserved as `getCause()`, and now this applies uniformly to all three initialization modes instead of only Kerberos. I re-checked how the Zeta engine's generic `catch (Throwable ...)` failure-classification sites treat this — they don't special-case `LinkageError` for retry/failure decisions (only `OutOfMemoryError` gets special dispatch, which remains untouched here), so converting the `Error` to an `IOException` doesn't change how the job ultimately fails, on any of the three paths.

Negligible, and unchanged from the previous round's assessment even though the wrapper is now applied at three call sites instead of one: it's a `try/catch` around an already-existing method call, evaluated once per `FileSystem` acquisition (i.e., once at connector/task initialization, not per-record). There is no active version probing, no extra reflection, no extra I/O — the wrapper is purely passive (it only does anything when `FileSystem.get()` itself throws). No measurable effect on CPU/memory/GC/networking/concurrency.

The original `LinkageError` is preserved as `getCause()` on every rewrite path (never swallowed), no new logging is introduced (appropriate — callers already log/report initialization failures), and no sensitive information (credentials, keytab paths, tokens) is included in the new message. One precision issue and two minor Copilot-flagged nits below.

### Issue 1
- **Location**: `HadoopFileSystemProxy.java:518-533` (`wrapClasspathMismatch`)
- **Problem description**: `LinkageError` is a broader family than the "classpath/version mismatch" symptoms this PR is diagnosing. Besides `NoSuchMethodError`/`NoClassDefFoundError`/`NoSuchFieldError`/`IncompatibleClassChangeError` (all genuine version-mismatch symptoms), `LinkageError` also covers `ExceptionInInitializerError` (a static initializer threw *any* `Throwable`, which could be entirely unrelated to Hadoop, e.g. a misconfigured logging framework or a bug in a static block), `UnsatisfiedLinkError` (a native library failed to load — a completely different fix path, e.g. installing `libhadoop.so` or setting `-Djava.library.path`), `ClassFormatError`/`VerifyError` (a corrupted or bytecode-incompatible class file), and `ClassCircularityError`. If `FileSystem.get()` ever throws one of these for a reason unrelated to a Hadoop client version mismatch, the user will see a message that confidently but incorrectly points them at "check that the Hadoop client jars on the classpath are compatible with 3.1.4" instead of the real problem.
- **Potential risk**: This is diagnostic-text-only — the job still fails exactly as it would have before (this doesn't change success/failure, doesn't lose data, and doesn't touch a recovery/checkpoint path), and the true cause is still fully recoverable via `getCause()` for anyone who reads past the top-level message. So I'd call the risk "misleading first-line diagnosis in a rarer sub-case," not a functional regression. I want to flag it explicitly rather than let it slide, though, since the whole point of this PR is diagnostic precision, and the message wording ("likely due to a Hadoop client version mismatch") reads as a confident, singular diagnosis rather than one of several possibilities.
- **Best improvement**:
  - Option A (lowest risk, smallest diff): soften the message to acknowledge it may be one of several classpath/linkage problems, e.g. "...this may indicate a Hadoop client version mismatch, or another classpath/linkage problem; see the cause below for the specific failure..." — keeps the broad `LinkageError` catch (simple, already well-tested) but stops overclaiming a single root cause.
  - Option B (more precise, more invasive): narrow the catch back to the error types that are actually version-mismatch symptoms — `NoSuchMethodError`, `NoClassDefFoundError`, `NoSuchFieldError`, `IncompatibleClassChangeError` (which already subsumes `AbstractMethodError`/`IllegalAccessError`/`InstantiationError`) — and let `ExceptionInInitializerError`, `UnsatisfiedLinkError`, `ClassFormatError`, `VerifyError`, and `ClassCircularityError` propagate unwrapped, since they point at genuinely different root causes.
  - I'd lean toward Option A since it's a one-line message tweak with no test churn, but either is reasonable — happy to discuss.
- **Severity**: Medium (diagnostic-only, no functional/compatibility impact, cause chain is fully preserved either way — does not block merge)

### Issue 2
- **Location**: `HadoopFileSystemProxy.java:527` (message literal inside `wrapClasspathMismatch`)
- **Problem description**: The Hadoop version this connector targets ("3.1.4") is hardcoded directly into the exception message string.
- **Potential risk**: Low — if `connector-file-base`'s Hadoop baseline ever changes, this string has to be found and updated by hand; it won't fail loudly if someone forgets.
- **Best improvement**: Centralize the target Hadoop version into a constant (or read it from the build, e.g. a Maven-injected property) and reference that constant from the message, so it can't drift out of sync with the actual `seatunnel-hadoop3-*-uber` dependency.
- **Severity**: Low
- **Raised by another reviewer**: Yes (@Copilot, inline comment on `HadoopFileSystemProxy.java:531` in the latest push) — I agree with this one; +1.

### Issue 3
- **Location**: `HadoopFileSystemProxy.java:521-531` (`wrapClasspathMismatch` catch body)
- **Problem description**: The rewritten `IOException` message doesn't include the caught `LinkageError`'s own class name or message text inline — only the generic diagnostic prose. The specific failure (e.g. `NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get`) is only visible via `getCause()`.
- **Potential risk**: Low — nothing is lost (the cause chain is intact and most log frameworks print it), but in environments where only the top-level exception message is captured (e.g. a truncated log line or a monitoring alert that shows `getMessage()` only), the specific failing symbol is not visible.
- **Best improvement**: Append `e.getClass().getSimpleName() + ": " + e.getMessage()` (guarding for a null message) to the diagnostic string, in addition to keeping it as `getCause()`.
- **Severity**: Low
- **Raised by another reviewer**: Yes (@Copilot, inline comment on `HadoopFileSystemProxy.java:530` in the latest push) — I agree this would help.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Formatting and naming are consistent with the surrounding class (AOSP style, no wildcard imports). Both `getFileSystemWithDiagnostics` and `wrapClasspathMismatch` now carry full Javadoc with `@param`/`@return`/`@throws`, which resolves the documentation nit from my previous round. The package-private `FileSystemSupplier` functional interface remains minimal and appropriately scoped.

`HadoopFileSystemProxyTest` now has 7 tests covering `wrapClasspathMismatch`:
- Four rewrite tests (`NoSuchMethodError`, `NoClassDefFoundError`, `NoSuchFieldError`, `IncompatibleClassChangeError`) sharing a private `assertRewrittenAsDiagnosticIOException(LinkageError)` helper, each asserting the original is preserved as `getCause()` and the message contains "Hadoop client"/"version mismatch". This directly covers the mismatch-detected path for every error type the widened catch now handles.
- `testWrapClasspathMismatchDoesNotSwallowUnrelatedErrors` — asserts an `OutOfMemoryError` passes through as the identical instance, unwrapped. This is the key boundary test that makes the `LinkageError` widening reviewable, and it's exactly the kind of regression test that would have caught an accidental over-broad catch (e.g. `catch (Throwable)`).
- `testWrapClasspathMismatchPropagatesIOExceptionUnchanged` — a plain `IOException` from the supplier passes through identical, with a null cause. Covers the no-mismatch-but-still-failed path.
- `testWrapClasspathMismatchPassesThroughOnSuccess` — exercises the genuine success path against a real local filesystem (`file:///`). Covers the no-mismatch path cleanly.

All seven are synchronous, single-threaded, deterministic (no `Thread.sleep`, no shared static state, no execution-order dependency, no network/container dependency). No flaky-test anti-patterns present. **Test stability rating: Stable.**

Remaining coverage gap (non-blocking, unchanged from last round): none of the tests exercise `initialize()`, `initializeWithKerberosLogin()`, or `initializeWithRemoteUserLogin()` end-to-end with a real classpath mismatch — understandably, since that would need a genuinely version-skewed Hadoop client on the test classpath (or a live/mini-KDC for the Kerberos branch specifically). That the three call sites now actually route through the helper is verified by direct code inspection (confirmed above — one remaining `FileSystem.get(` call site, inside the helper) rather than by an integration test, which is a reasonable trade-off given the existing test infrastructure in this file.

## 2.3 Documentation Updates

Not required. This remains a diagnostics-only change to an internal error path; there is no new config option, no default-value change, and no public API/behavior change on the success path for `docs/en` or `docs/zh` to describe.

# 3. Architectural Soundness

This is now a **precise, complete fix** for the diagnosability gap in #10172 — not a partial one. The `wrapClasspathMismatch`/`FileSystemSupplier` design is small, reusable, and directly unit-testable, and it's now actually applied everywhere the failure mode can occur in this class.

## 3.2 Maintainability

Good. The helper is a small, self-contained static method with a single clear responsibility, directly unit-tested rather than only reachable through the hard-to-test Kerberos/remote-user paths.

## 3.3 Extensibility

The `FileSystemSupplier` functional interface makes it trivial to reuse this pattern for any other `FileSystem.get(...)`-shaped acquisition point in sibling file connectors built on this shared base, should they ever need the same diagnostic treatment.

No impact. No config, checkpoint/savepoint state, serialization format, or SPI/registration surface is touched. Existing saved jobs, historical task configs, and upgrade paths are unaffected.

# 4. Issue Summary

| 1 | Widened `catch (LinkageError e)` also catches `ExceptionInInitializerError`, `UnsatisfiedLinkError`, `ClassFormatError`, `VerifyError`, `ClassCircularityError`, whose root causes are not always a Hadoop version mismatch | `HadoopFileSystemProxy.java:518-533` | Medium |
| 2 | Target Hadoop version ("3.1.4") is hardcoded in the message string instead of a shared constant | `HadoopFileSystemProxy.java:527` | Low |
| 3 | Diagnostic message doesn't inline the caught `LinkageError`'s concrete type/message (only via `getCause()`) | `HadoopFileSystemProxy.java:521-531` | Low |

*(Issues 1 and 2 from my previous review — Kerberos-only scoping, and the narrow two-error catch — are both confirmed fixed at this head and are not re-listed.)*

# 5. Merge Recommendation

   - None. Both blocking items from the previous round (partial call-site coverage, narrow catch) are verified fixed at the current head.

   - Issue 1 (Medium): Consider softening the message wording (Option A) or narrowing the catch to the specific version-mismatch symptom types (Option B) so the diagnostic doesn't overclaim a single root cause for the less common `LinkageError` subtypes. Doesn't block merge — this is a diagnostic-precision improvement on an already-fatal path, and the true cause remains fully recoverable via `getCause()` either way.
   - Issue 2 (Low): Centralize the "3.1.4" version string into a constant to avoid future drift.
   - Issue 3 (Low): Optionally inline the caught error's type/message into the diagnostic text for log environments that don't surface the full cause chain.

**Overall assessment**: This is now a clean, complete, low-risk diagnostic improvement with solid, stable unit tests and zero impact on the success path, configuration, APIs, or backward compatibility. The core fix (rewrap link-time classpath errors into an actionable `IOException`, preserving the original as cause) is exactly right for #10172, and it's now genuinely applied to all three `FileSystem` acquisition paths in this class rather than just one. The one substantive nuance I found this round — the `LinkageError` catch being broader than "version mismatch" — is a diagnostic-precision concern, not a functional or compatibility one, so I'm not treating it as a blocker. Thanks again for addressing both points from the last review so directly and for keeping the test coverage in lockstep with the code changes — this is a good example of how these diagnostic PRs should be built.

---

**CI note**: At the time of this review, the fork's `Build` workflow run for the current head (`ac2a1e71b1`) has all completed jobs green — `License header`, `Code style`, `Sanity check results`, `unit-test` on JDK 8 and 11 across both `ubuntu-latest` and `windows-latest`, and `updated-modules-integration-test-part-1` through `part-8` all passed — with one job (`connector-file-local-it`) still `in_progress`. No failures observed. The two `apache/seatunnel`-side checks currently showing `pending` (`Notify test workflow`, `labeler`) are the gating pointer checks, not the real build — the actual build results live on the fork run linked above. I'll leave this to finish; no action needed from you here.

DanielLeens (commented):
Thanks for the fast rebase, @RohanExploit. I re-reviewed the full diff at the current head against `dev` from scratch, and re-fetched my own previous review's commit (`ac2a1e71b`) to diff it directly against this one. The two are content-identical: `git diff ac2a1e71b1 5aad9a937` is empty. Same tree, same parent, same author timestamp — the only difference between the two commits is the committer timestamp and that this version drops the `Co-Authored-By: Claude Opus 5 (1M context)` trailer from the commit message (the PR description's AI-assistance disclosure is unchanged). No source, test, or doc bytes changed. So this pass is a fresh re-verification of the same code I approved last round, not a review of new logic — I'm not going to manufacture new findings where none exist, but I did re-trace the full call chain and re-check the open non-blocking items from last time rather than rubber-stamping.

**User pain point**: When the Hadoop client jars resolved at runtime (a `provided` dependency supplied by the deployment environment) don't match the Hadoop version SeaTunnel's file connectors are built against (Hadoop 3.1.4, via `seatunnel-hadoop3-3.1.4-uber`), `FileSystem.get()` can fail deep inside Hadoop's own `DFSClient`/`FsTracer` construction with a raw JVM linkage error. Issue #10172 shows exactly this — the user only saw `NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get`, with nothing pointing at the real cause.

**Fix approach**: All three `FileSystem.get(configuration)` call sites in `HadoopFileSystemProxy` — the plain path (`initialize()`), the Kerberos path (`initializeWithKerberosLogin()`), and the proxy-user path (`initializeWithRemoteUserLogin()`) — now route through a shared `getFileSystemWithDiagnostics()` helper that catches `LinkageError` around `FileSystem.get()` and rethrows it as an `IOException` naming the likely cause and the target Hadoop version, preserving the original `LinkageError` as `getCause()`. Nothing changes on the success path.

**One-sentence summary**: Turns an opaque JVM linkage crash into an actionable diagnostic message, consistently across all three ways this connector acquires a Hadoop `FileSystem`.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**File**: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/hadoop/HadoopFileSystemProxy.java`

Current head, all three call sites (confirmed via `grep -n "FileSystem.get(" HadoopFileSystemProxy.java` — the only remaining hit is inside the helper itself, line 494):

```java
// initialize() -- line 388
fileSystem = getFileSystemWithDiagnostics(configuration);

// initializeWithKerberosLogin() -- line 463 (inside the login callback)
this.fileSystem = getFileSystemWithDiagnostics(configuration);

// initializeWithRemoteUserLogin() -- line 546 (inside the login callback)
this.fileSystem = getFileSystemWithDiagnostics(configuration);
```

```java
// lines 490-495
private static FileSystem getFileSystemWithDiagnostics(Configuration configuration)
        throws IOException {
    return wrapClasspathMismatch(() -> FileSystem.get(configuration));
}

// lines 513-531
static FileSystem wrapClasspathMismatch(FileSystemSupplier supplier) throws IOException {
    try {
        return supplier.get();
    } catch (LinkageError e) {
        throw new IOException(
                "Failed to initialize Hadoop FileSystem, likely due to a Hadoop client "
                        + "version mismatch between SeaTunnel and the Hadoop jars resolved "
                        + "at runtime (the Hadoop dependency here is `provided`, so it comes "
                        + "from your deployment environment, not a bundled version). This "
                        + "connector is built and tested against Hadoop 3.1.4 (see "
                        + "seatunnel-hadoop3-3.1.4-uber in the reactor pom); check that the "
                        + "Hadoop client jars on the classpath are compatible with that "
                        + "version.",
                e);
    }
}
```

**Key findings**
- All three call sites are wrapped, verified by direct grep on the current head, matching what I confirmed last round.
- The original error is never swallowed — preserved as `getCause()` on every rewrite path, and this is directly asserted by `assertRewrittenAsDiagnosticIOException` in the test file for all four `LinkageError` subtypes exercised.
- I re-checked whether "Hadoop client jars ... compatible" is a real false-positive risk against *legitimately compatible* client/server version pairs, since that's specifically what I was asked to re-verify this round: `LinkageError` is a JVM class-loading/linking failure — it only fires when the classes actually present on the local classpath disagree with each other at the bytecode level (missing method/field/class, incompatible class hierarchy). It has nothing to do with the wire protocol between the Hadoop client and an HDFS NameNode/DataNode. A client that is wire-compatible with a differently-versioned server will never hit this catch, because no linking failure occurs in that scenario — the failure mode this PR targets is purely a local classpath-shape problem, not a client/server version-skew problem. So there's no plausible false-positive path where a healthy client/server pairing gets misdiagnosed as a "Hadoop client version mismatch" by this catch. The one caveat to that is Issue 1 below (carried over, still open at this head).
- Re-checked for sensitive-data leakage in the new message: it's static prose plus `getCause()` — no keytab path, principal, username, host, or credential is interpolated into the string. `hadoopConf.getKerberosPrincipal()`/`getRemoteUser()` are logged separately via the pre-existing `log.info(...)` success-path calls (lines 468, 549), which this PR does not touch and which only fire on success, not in the new exception path. No new leakage surface introduced.
- Confirmed `connector-file-base/pom.xml:52,135` still declares `seatunnel-hadoop3-3.1.4-uber` as the sole Hadoop uber dependency on current `dev`, so the hardcoded "3.1.4" claim in the message remains accurate.

**Runtime flow trace** (failure path; success path is unchanged for all three modes):
```text
HadoopFileSystemProxy(HadoopConf)                         [constructor, eager init]
  -> initialize()                                  [@SneakyThrows]
    +-- enableKerberos() == true
    |     -> initializeWithKerberosLogin()
    |       -> HadoopLoginFactory.loginWithKerberos(..., loginFunction)
    |         -> loginFunction.run(...) -> getFileSystemWithDiagnostics(configuration)
    +-- enableRemoteUser() == true
    |     -> initializeWithRemoteUserLogin()
    |       -> HadoopLoginFactory.loginWithRemoteUser(..., loginFunction)
    |         -> loginFunction.run(...) -> getFileSystemWithDiagnostics(configuration)
    +-- neither (plain path)
          -> getFileSystemWithDiagnostics(configuration)   [directly, line 388]

  getFileSystemWithDiagnostics(configuration)
    -> wrapClasspathMismatch(() -> FileSystem.get(configuration))
      -> FileSystem.get(configuration)
        -> DistributedFileSystem.initialize(...) -> new DFSClient(...) -> FsTracer.get(...)
           [signature/shape changed across Hadoop releases]
        -> LinkageError thrown (unchecked)
      <- caught, rethrown as IOException(diagnostic message, cause=LinkageError)
  <- IOException propagates out through whichever of the three initialize*() paths was taken,
     out of initialize() via @SneakyThrows, and fails connector/task init with an actionable
     message instead of a bare Hadoop-internal stack trace -- for all three auth modes.
```
This sits on connector initialization (task startup) — a real, normal-use code path — but the branch that triggers the rewrite (a jar-version/classpath mismatch) is itself an environment/deployment error condition, consistent with "no change on success."

## 1.2 Compatibility Impact

**Fully compatible.** No config option, default value, public API, or serialization format is touched. The only behavioral change is on an already-fatal failure path: an unchecked `Error` now escapes as a checked `IOException` with the original preserved as `getCause()`, uniformly across all three initialization modes. Re-confirmed the Zeta engine's generic `catch (Throwable ...)` failure-classification sites don't special-case `LinkageError` for retry/failure decisions (only `OutOfMemoryError` gets special dispatch, untouched here), so this doesn't change how the job ultimately fails.

Negligible, unchanged from last round: a passive `try/catch` around an already-existing method call, evaluated once per `FileSystem` acquisition at connector/task initialization, not per-record. No extra I/O, reflection, or version probing. No measurable CPU/memory/GC/networking/concurrency effect.

The original `LinkageError` is preserved as `getCause()` on every rewrite path, no new logging is introduced, and — re-verified this round specifically for leakage — no credentials, keytab paths, principals, hostnames, or tokens are included in the new message. One item carried forward from last round, still open at this head:

### Issue 1 (carried over from my 2026-08-19T10:38 review, unaddressed at this head)
- **Location**: `HadoopFileSystemProxy.java:513-531` (`wrapClasspathMismatch`)
- **Problem description**: `LinkageError` is broader than the "classpath/version mismatch" symptoms this PR is diagnosing. Besides `NoSuchMethodError`/`NoClassDefFoundError`/`NoSuchFieldError`/`IncompatibleClassChangeError` (genuine version-mismatch symptoms), `LinkageError` also covers `ExceptionInInitializerError` (a static initializer threw *any* `Throwable`, possibly unrelated to Hadoop), `UnsatisfiedLinkError` (a native library failed to load — a completely different fix, e.g. `libhadoop.so`/`java.library.path`), `ClassFormatError`/`VerifyError` (a corrupted or bytecode-incompatible class file), and `ClassCircularityError`. If `FileSystem.get()` ever throws one of these for a reason unrelated to a Hadoop client version mismatch, the message confidently but incorrectly points the user at "check that the Hadoop client jars on the classpath are compatible with 3.1.4."
- **Potential risk**: Diagnostic-text-only — the job still fails exactly as before (no change to success/failure, no data loss, no recovery-path impact), and the true cause remains fully recoverable via `getCause()`. So this is "misleading first-line diagnosis in a rarer sub-case," not a functional regression.
- **Best improvement**: Option A (lowest risk): soften the message to "...this may indicate a Hadoop client version mismatch, or another classpath/linkage problem; see the cause below..." — keeps the simple, already-tested `LinkageError` catch. Option B (more precise): narrow the catch to `NoSuchMethodError`, `NoClassDefFoundError`, `NoSuchFieldError`, `IncompatibleClassChangeError` and let the others propagate unwrapped.
- **Severity**: Medium (diagnostic-only, no functional/compatibility impact — does not block merge)
- **Raised by another reviewer**: No (this is my own finding, carried forward)

Two more Low items from last round are also still open and unaddressed — listed in the summary table below for completeness, not repeated in full since nothing about them changed.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Unchanged from last round: formatting/naming consistent with the surrounding class, both `getFileSystemWithDiagnostics` and `wrapClasspathMismatch` carry full Javadoc (`@param`/`@return`/`@throws`), and the package-private `FileSystemSupplier` functional interface is minimal and appropriately scoped.

Re-verified `HadoopFileSystemProxyTest` byte-for-byte against what I reviewed last round (confirmed via direct diff of the two commits — identical). Seven tests cover `wrapClasspathMismatch`: four rewrite tests (`NoSuchMethodError`, `NoClassDefFoundError`, `NoSuchFieldError`, `IncompatibleClassChangeError`) sharing `assertRewrittenAsDiagnosticIOException`, plus `testWrapClasspathMismatchDoesNotSwallowUnrelatedErrors` (OOM passes through as the same instance — the key boundary test making the `LinkageError` widening safe), `testWrapClasspathMismatchPropagatesIOExceptionUnchanged`, and `testWrapClasspathMismatchPassesThroughOnSuccess`. All synchronous, single-threaded, deterministic — no `Thread.sleep`, no shared static state, no execution-order dependency, no network/container dependency. **Test stability rating: Stable.**

CI at this exact head (`5aad9a937e7541d25c42cd9116a8414a462989b1`) is fully green on the fork's real run (`RohanExploit/seatunnel` run `32247289004`, started 2026-08-19T11:22:51Z): `unit-test` on JDK 8/11 × ubuntu/windows, `Code style`, `License header`, `Sanity check results`, `connector-file-local-it` and `connector-file-sftp-it` on JDK 8/11, and `updated-modules-integration-test-part-1` through `part-8` all `success`. The apache-side `Build` check is the usual gating pointer to that same fork run.

## 2.3 Documentation Updates

Not required — diagnostics-only change to an internal error path, no config/API/behavior change on the success path.

# 3. Architectural Soundness

Precise, complete fix for #10172: small, reusable, directly unit-tested (`wrapClasspathMismatch`/`FileSystemSupplier`), genuinely applied to all three `FileSystem` acquisition paths.

## 3.2 Maintainability

Good — small, self-contained static helper with one clear responsibility, unit-tested rather than only reachable through the hard-to-test Kerberos/remote-user paths.

## 3.3 Extensibility

`FileSystemSupplier` makes the pattern trivially reusable at any other `FileSystem.get(...)`-shaped acquisition point in sibling file connectors built on this shared base.

No impact. No config, checkpoint/savepoint state, serialization format, or SPI/registration surface touched. Existing saved jobs, historical task configs, and upgrade paths unaffected.

# 4. Issue Summary

| 1 | `catch (LinkageError e)` also catches `ExceptionInInitializerError`, `UnsatisfiedLinkError`, `ClassFormatError`, `VerifyError`, `ClassCircularityError`, whose root causes are not always a Hadoop version mismatch | `HadoopFileSystemProxy.java:513-531` | Medium |
| 2 | Target Hadoop version ("3.1.4") is hardcoded in the message string instead of a shared constant | `HadoopFileSystemProxy.java:527` | Low |
| 3 | Diagnostic message doesn't inline the caught `LinkageError`'s concrete type/message (only via `getCause()`) | `HadoopFileSystemProxy.java:521-528` | Low |

*(All three carried over unchanged from my 2026-08-19T10:38 review at `ac2a1e71b` — the code at this head is byte-identical to that commit, so none of them could have been addressed or regressed in between. The two blockers from my very first review — Kerberos-only scoping, and the narrow two-error catch — remain fixed, unaffected by this metadata-only amend.)*

# 5. Merge Recommendation

   - None.

   - Issue 1 (Medium): soften the message wording or narrow the catch to the specific version-mismatch symptom types, so rarer `LinkageError` subtypes don't get a confidently-wrong first-line diagnosis.
   - Issue 2 (Low): centralize "3.1.4" into a constant to avoid future drift from the actual `seatunnel-hadoop3-*-uber` dependency.
   - Issue 3 (Low): optionally inline the caught error's type/message for log environments that truncate the cause chain.

**Overall assessment**: This re-review confirms the code at the current head is functionally and textually identical to the version I approved last round — `git diff` between the two commits is empty, the only change is a commit-metadata amend. CI is fully green on the real fork run for this exact SHA. The diagnostic correctly targets a local JVM classpath-linking failure mode that cannot occur for a legitimately compatible Hadoop client/server pairing, so there's no false-positive risk against healthy deployments, and the message carries no credentials, paths, or other sensitive data. Nothing here changes my prior conclusion: this is a clean, low-risk, well-tested diagnostic improvement with zero impact on the success path, configuration, APIs, or backward compatibility. Thanks again for the quick turnaround on the earlier rounds — no further action needed from you unless you want to pick up one of the non-blocking suggestions above.

RohanExploit:
Thanks @DanielLeens, and no need to frame Issue 1 as your miss. You were right that the narrow two type catch was a coverage gap, and you are right now that `LinkageError` is broader than "version mismatch". Both things are true, the second only shows up once the first is fixed. All three items are addressed at `a0f6014`.

Issue 1. I took Option A rather than Option B, and I want to explain why since you offered both. Narrowing back to the four mismatch types would let `UnsatisfiedLinkError` and friends escape unwrapped, which puts those users back in exactly the position #10172 describes, a bare linkage error out of Hadoop internals with nothing pointing anywhere. The problem was never that the catch was too wide, it was that the message asserted one cause with more confidence than the catch could justify. So the catch still covers the whole family and the message now says "may indicate", names the other realistic possibilities, and points at the cause. Combining that with Issue 3 does most of the work, because the concrete error is now in the message itself:

```
Failed to initialize Hadoop FileSystem (UnsatisfiedLinkError: no hadoop in java.library.path).
This may indicate a Hadoop client version mismatch between SeaTunnel and the Hadoop jars
resolved at runtime (...), or another classpath or linkage problem such as a missing native
library or an unreadable class file. This connector is built and tested against Hadoop 3.1.4
(see seatunnel-hadoop3-3.1.4-uber in the reactor pom); check that the Hadoop client jars on
the classpath are compatible with that version, and see the cause below for the specific failure.
```

A native library failure now reads as a native library failure while still getting the classpath framing, which neither option gave on its own. If you would still rather have the narrow catch I will switch it, it is a small change.

Issue 2. The version moved into `TARGET_HADOOP_VERSION`, and the uber artifact name is derived from it as `"seatunnel-hadoop3-" + TARGET_HADOOP_VERSION + "-uber"` so the two cannot drift apart. Grepping the file for `3.1.4` now returns exactly one hit, the constant.

Issue 3. Done via a small `describeLinkageError` helper that renders `Type: message`, falling back to just the type when the error carries no message, so the text never contains a stray "null".

Tests are up to 13, two new ones:

* `testWrapClasspathMismatchNamesTheConcreteLinkageFailure` throws an `UnsatisfiedLinkError` and asserts both the type and its message appear in the rewritten text. This is the one that covers the non mismatch subtypes you raised.
* `testWrapClasspathMismatchHandlesLinkageErrorWithoutMessage` throws a `NoClassDefFoundError` with no message and asserts the type is still named and the text does not contain "null".

The shared `assertRewrittenAsDiagnosticIOException` helper now also asserts the concrete type is named, so all four original rewrite tests cover the inlining too.

I checked the new tests actually bite rather than assuming it. Dropping the null guard from `describeLinkageError` fails `testWrapClasspathMismatchHandlesLinkageErrorWithoutMessage` with `expected: <false> but was: <true>`, then passes again once restored. `spotless:check` is clean and all 13 tests pass locally on JDK 11, with the one pre-existing Windows only skip.

DanielLeens (approved):
Thanks for the detailed writeup, @RohanExploit — and for walking through the Option A vs Option B tradeoff explicitly rather than just picking one. Your reasoning holds up: narrowing the catch back to the four mismatch types would put `UnsatisfiedLinkError` users right back in the position issue #10172 describes (a bare linkage error out of Hadoop internals, no diagnosis at all), so keeping the wide `LinkageError` catch and fixing the message's overconfidence instead is the right call. I re-reviewed the entire current head from scratch (not just the incremental diff) against `dev`, and I'm happy to confirm all three of my carried-over issues are now closed with real test coverage, not just message wording.

When the Hadoop client jars resolved at runtime (a `provided` dependency supplied by the deployment environment) don't match the Hadoop version SeaTunnel's file connectors are built against (3.1.4, via `seatunnel-hadoop3-3.1.4-uber`), `FileSystem.get()` can fail deep inside Hadoop's own `DFSClient`/`FsTracer` construction with a raw JVM linkage error — issue #10172 shows a user seeing only `NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get`, with nothing pointing at the real cause. This PR routes all three `FileSystem.get(configuration)` call sites in `HadoopFileSystemProxy` (plain, Kerberos, remote-user) through a shared `wrapClasspathMismatch()` helper that catches the whole `LinkageError` family and rethrows it as an `IOException` naming the concrete failure and the likely causes, preserving the original error as `getCause()`. One-sentence summary: turns an opaque JVM linkage crash into an actionable diagnostic message, without changing anything on the success path.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**File**: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/hadoop/HadoopFileSystemProxy.java`

What changed since my last review at `5aad9a937` (current head `a0f6014ca`, confirmed via `git diff 5aad9a937 a0f6014ca`):

```java
private static final String TARGET_HADOOP_VERSION = "3.1.4";
private static final String TARGET_HADOOP_UBER_ARTIFACT =
        "seatunnel-hadoop3-" + TARGET_HADOOP_VERSION + "-uber";
...
static FileSystem wrapClasspathMismatch(FileSystemSupplier supplier) throws IOException {
    try {
        return supplier.get();
    } catch (LinkageError e) {
        throw new IOException(
                "Failed to initialize Hadoop FileSystem ("
                        + describeLinkageError(e)
                        + "). This may indicate a Hadoop client version mismatch between "
                        + "SeaTunnel and the Hadoop jars resolved at runtime (...), or another "
                        + "classpath or linkage problem such as a missing native library or an "
                        + "unreadable class file. This connector is built and tested against "
                        + "Hadoop " + TARGET_HADOOP_VERSION + " (see " + TARGET_HADOOP_UBER_ARTIFACT
                        + " in the reactor pom); check that the Hadoop client jars on the "
                        + "classpath are compatible with that version, and see the cause below "
                        + "for the specific failure.",
                e);
    }
}

private static String describeLinkageError(LinkageError e) {
    String message = e.getMessage();
    return message == null
            ? e.getClass().getSimpleName()
            : e.getClass().getSimpleName() + ": " + message;
}
```

Verified against the three items I raised last round:

- **Issue 1 (Medium, message overconfidence)**: fixed. The message now says "This may indicate ... or another classpath or linkage problem such as a missing native library or an unreadable class file" instead of asserting a version mismatch as the sole cause. The catch itself is unchanged (still the whole `LinkageError` family), which I agree is the correct call for the reason given above — narrowing it would silently regress the `UnsatisfiedLinkError`/`ExceptionInInitializerError`/etc. cases back to zero diagnosis.
- **Issue 2 (Low, hardcoded version)**: fixed. `TARGET_HADOOP_VERSION` is now the single source of truth, and `TARGET_HADOOP_UBER_ARTIFACT` is derived from it (`"seatunnel-hadoop3-" + TARGET_HADOOP_VERSION + "-uber"`), so the two can't drift apart. Confirmed `connector-file-base/pom.xml` still declares `seatunnel-hadoop3-3.1.4-uber` as the sole Hadoop uber dependency, so "3.1.4" is still accurate.
- **Issue 3 (Low, concrete error not inlined)**: fixed via `describeLinkageError(e)`, which renders `Type: message`, falling back to just the simple class name when `getMessage()` is null so the text never contains a stray `"null"`.

**Runtime flow** (unchanged from before, still applies to all three `initialize*()` paths):
```text
initialize() / initializeWithKerberosLogin() / initializeWithRemoteUserLogin()
  -> getFileSystemWithDiagnostics(configuration)
    -> wrapClasspathMismatch(() -> FileSystem.get(configuration))
      -> FileSystem.get(configuration) -> DistributedFileSystem.initialize(...)
         -> new DFSClient(...) -> FsTracer.get(...)  [signature/shape changed across Hadoop releases]
      -> LinkageError thrown (unchecked)
    <- caught, rethrown as IOException("Failed to initialize Hadoop FileSystem (Type: msg). ...", cause=LinkageError)
```
This sits on connector/task initialization — a real, normal-use path — but the rewrite only fires on an already-fatal classpath/linkage failure, so there's still zero behavior change on the success path.

No credentials, keytab paths, principals, or hostnames are interpolated into the new message — it's static prose plus `describeLinkageError(e)`, and `e.getMessage()` on a `LinkageError` from Hadoop's own class-loading machinery doesn't carry user secrets.

## 1.2 Compatibility Impact

**Fully compatible.** No config option, default value, public API, or serialization format touched. Same as before: an already-fatal failure path now surfaces as a checked `IOException` with the original preserved as `getCause()`, and Zeta's generic failure-classification (`catch (Throwable ...)`) doesn't special-case `LinkageError`, so job-failure handling is unaffected.

Negligible. `describeLinkageError()` is a single null-check plus (at most) one string concatenation, executed only on the already-rare failure path, once per `FileSystem` acquisition at initialization — not per-record. No new I/O, reflection, or hot-path cost.

No new issues found this round. All three of my previously carried Medium/Low issues are addressed as described above. No new logging is introduced; `getCause()` still carries the original `LinkageError` end-to-end.

# 2. Code Quality Assessment

## 2.1 Coding Standards

Both new/changed methods (`wrapClasspathMismatch`, `describeLinkageError`) have full Javadoc explaining intent and edge cases (the null-message fallback is explicitly documented). Naming is clear and consistent with the surrounding class.

Test count is now 13 (up from 11 at my last review). Two new tests directly targeting the fixed issues:
- `testWrapClasspathMismatchNamesTheConcreteLinkageFailure` — throws `UnsatisfiedLinkError`, asserts both the type name and the original message text appear in the rewritten `IOException` message. This is the test that actually exercises the "non-mismatch `LinkageError` subtype" case I raised as Issue 1.
- `testWrapClasspathMismatchHandlesLinkageErrorWithoutMessage` — throws `NoClassDefFoundError` with no message, asserts the type is still named and the text does not contain `"null"`.

The shared `assertRewrittenAsDiagnosticIOException` helper now additionally asserts the concrete error type is named, so all four original rewrite tests (`NoSuchMethodError`, `NoClassDefFoundError`, `NoSuchFieldError`, `IncompatibleClassChangeError`) also cover the inlining behavior, not just the two new tests. The author's PR comment also describes deliberately breaking `describeLinkageError`'s null guard locally to confirm `testWrapClasspathMismatchHandlesLinkageErrorWithoutMessage` actually fails without it — that's exactly the kind of mutation-style sanity check I like to see, and it lines up with what the test asserts. All tests remain synchronous, single-threaded, deterministic, no `Thread.sleep`, no shared static state, no ordering dependence. **Test stability rating: Stable.**

CI at the current head (`a0f6014ca3953817c7e68be19bffcbac98666cc5`) is green: `Build` passed (16m22s), plus the notify/labeler workflows.

## 2.3 Documentation Updates

Not required — this is a diagnostics-only change to an internal error path, no config/API/behavior change on the success path.

# 3. Architectural Soundness

Precise, complete fix. The wide `LinkageError` catch plus a message that names the concrete failure and presents the version mismatch as a likely-not-certain cause is a better solution than either option I offered last round in isolation — it keeps full diagnostic coverage for issue #10172's original symptom while no longer overclaiming the cause for rarer `LinkageError` subtypes.

## 3.2 Maintainability

Good. `TARGET_HADOOP_VERSION` as a single source of truth removes a future drift risk, and `describeLinkageError` is a small, independently-testable, self-contained helper.

## 3.3 Extensibility

Unchanged from before — `FileSystemSupplier` remains trivially reusable at any other `FileSystem.get(...)`-shaped acquisition point in sibling file connectors.

No impact. No config, checkpoint/savepoint state, serialization format, or SPI/registration surface touched.

# 4. Issue Summary

| 1 | Message asserted version mismatch as the sole cause for the whole `LinkageError` family | `HadoopFileSystemProxy.java` (`wrapClasspathMismatch`) | Medium | **Fixed** at `a0f6014ca` |
| 2 | Target Hadoop version hardcoded as a string literal | `HadoopFileSystemProxy.java` | Low | **Fixed** at `a0f6014ca` |
| 3 | Diagnostic message didn't inline the caught error's concrete type/message | `HadoopFileSystemProxy.java` | Low | **Fixed** at `a0f6014ca` |

No new issues found in this re-review.

# 5. Merge Recommendation

   - None.

   - None remaining. All previously open items are closed with test coverage.

**Overall assessment**: This is exactly how a re-review round should go — every carried-over issue got a real fix backed by a new test that specifically exercises it, not just a wording tweak, and the author's explanation of why Option A (keep the wide catch, fix the message) beats Option B (narrow the catch) is correct and well-argued. CI is green at the current head. I have no further concerns; this is a clean, low-risk, well-tested diagnostic improvement with zero impact on the success path, configuration, or backward compatibility. Thanks for the thorough follow-through, @RohanExploit.

DanielLeens (approved):
Thanks for the ping to re-check this one. Before writing anything else: I re-fetched the current PR head directly and it is `1f91908ff1e8d42cd64193c2b0d5f887ea54eb84` — the exact same commit I already reviewed and approved on the last pass. `git diff` between that commit and the one I approved is empty; nothing was pushed to this branch since my approval. The "behind dev" count moving (dev has picked up ~27 unrelated commits since) is why this landed back in front of me, not any new commit on this PR. So this is a pure re-verification, not a review of new logic — I re-read the full current diff against `dev` from scratch anyway rather than rubber-stamping, and re-checked CI and merge status.

When the Hadoop client jars resolved at runtime (a `provided` dependency supplied by the deployment environment) don't match the Hadoop version SeaTunnel's file connectors are built against (3.1.4, via `seatunnel-hadoop3-3.1.4-uber`), `FileSystem.get()` can fail deep inside Hadoop's own `DFSClient`/`FsTracer` construction with a raw JVM linkage error — issue #10172 shows a user seeing only `NoSuchMethodError: org.apache.hadoop.fs.FsTracer.get`, with nothing pointing at the real cause. This PR routes all three `FileSystem.get(configuration)` call sites in `HadoopFileSystemProxy` (plain `initialize()`, Kerberos `initializeWithKerberosLogin()`, remote-user `initializeWithRemoteUserLogin()`) through a shared `wrapClasspathMismatch()` helper that catches the whole `LinkageError` family and rethrows it as an `IOException` naming the concrete failure and the likely causes, preserving the original error as `getCause()`. Success path is untouched.

# 1. Code Change Review

## 1.1 Core Logic Analysis

**File**: `seatunnel-connectors-v2/connector-file/connector-file-base/src/main/java/org/apache/seatunnel/connectors/seatunnel/file/hadoop/HadoopFileSystemProxy.java`

Confirmed diff against `dev` merge-base is exactly what I reviewed last round: two new constants (`TARGET_HADOOP_VERSION`, `TARGET_HADOOP_UBER_ARTIFACT`), three call sites switched from `FileSystem.get(configuration)` to `getFileSystemWithDiagnostics(configuration)`, and the new `getFileSystemWithDiagnostics` / `wrapClasspathMismatch` / `describeLinkageError` helpers plus the `FileSystemSupplier` functional interface. `HadoopFileSystemProxyTest.java` carries the matching regression tests.

Key findings (all previously verified, re-confirmed on this pass):
- All three `FileSystem.get()` call sites are covered — I grepped the file and there is exactly one raw call left, inside the helper itself.
- Catching the full `LinkageError` family (not just the four mismatch-shaped subtypes) is deliberate and correct: narrowing it would silently regress `UnsatisfiedLinkError` users back to an undiagnosed crash, which is exactly the gap #10172 reported. The message correctly hedges with "may indicate" / "likely" rather than asserting the cause with false confidence.
- `describeLinkageError` keeps the concrete failure type+message visible even in log setups that only capture `getMessage()` and drop the cause chain — a real, previously-flagged gap that's now closed with test coverage (`NoClassDefFoundError`/`NoSuchFieldError`/`IncompatibleClassChangeError` cases).
- The original `LinkageError` is preserved via `getCause()`, so nothing is lost for callers that do inspect the full chain.
- Non-`LinkageError` failures (`OutOfMemoryError`, the supplier's own `IOException`) propagate untouched — verified by dedicated pass-through regression tests.

Correctness: the success path is byte-for-byte unchanged (same `FileSystem.get()` call, same downstream `setWriteChecksum`/`userGroupInformation` handling); only the failure path is rewritten, and only for the specific `LinkageError` family that indicates a broken/mismatched classpath rather than a legitimate application-level `IOException`.

## 1.2 Compatibility Impact

No behavioral change for any currently-working job: the diagnostic path only triggers on `LinkageError`, which by definition never occurs on a healthy classpath. No config, SPI, or public API surface changed. Purely additive.

No hot-path cost — the wrapping is a single `try/catch` around a one-time `FileSystem.get()` call executed during connector initialization, not per-record. No new allocations on the success path.

The rewritten `IOException` names the concrete failure (via `describeLinkageError`) and the target Hadoop version, and chains the original error as `getCause()` so nothing is swallowed. This directly closes the original opaque-crash complaint in #10172.

# 2. Code Quality Assessment

## 2.1 Coding Standards
Consistent with the surrounding class; Javadoc on the new helpers explains the "why" (linkage errors, provided dependency, version drift) rather than restating the code.

`HadoopFileSystemProxyTest` covers the widened catch with dedicated cases for the various `LinkageError` subtypes plus explicit `OutOfMemoryError`/`IOException` pass-through tests. These are static unit tests around a pure helper method (no I/O, no timing, no external containers) — no flakiness risk.

## 2.3 Documentation Updates
No user-facing config or behavior changed that would require `docs/en`/`docs/zh` updates; this is an internal diagnostic-message improvement.

# 3. Architectural Soundness

Centralizing all three call sites through one helper avoids the three-way duplication that a narrower per-call-site fix would have produced.

## 3.2 Maintainability
`TARGET_HADOOP_VERSION`/`TARGET_HADOOP_UBER_ARTIFACT` are declared once, so the message can't drift from the actual reactor dependency the next time the Hadoop baseline changes.

## 3.3 Extensibility
The `FileSystemSupplier` functional interface plus `wrapClasspathMismatch` being `static`/package-visible make it trivial to reuse for other Hadoop-facing initialization if a similar diagnostic is ever needed elsewhere.

No serialization, checkpoint, or config format touched. Nothing here affects restoring or resuming previously-created jobs.

# 4. Issue Summary

| — | — | None found on this pass; all prior findings (narrow catch, missing concrete-error detail, stale hardcoded version) were resolved in earlier rounds and re-verified here | — | Copilot's two suggestions (hardcoded version, missing concrete failure detail) were both already addressed before my prior APPROVED |

# 5. Merge Recommendation

### Conclusion: Approve (reaffirmed)

1. Blockers: none.
2. Recommended fixes: none outstanding.

CI is green (`Build`, `Notify test workflow`, `labeler` all passing on the current head). `mergeStateStatus` is `BLOCKED` only because this needs an approving review from an account with write access — my own approval doesn't satisfy branch protection. No code-side blockers remain from my side; this is ready for a maintainer with merge rights to take the final look and merge.

PR #11826, opened 2026-08-16
[Fix][Connector-V2] Fix issue of out-of-order fields in cdc

corgy-w:

Fix the issue of out-of-order fields in CDC when `column.include.list` (or other column-subsetting) is configured.

In `MySqlSnapshotSplitReadTask`, the snapshot reader picked the column definition with `table.columns().get(i)`, which returns columns in the **table DDL order**. When Debezium's `column.include.list` filters/orders the columns, the actual column order in the query result set differs from the table order, causing field values to be written into the wrong positions.

The fix uses `columnArray.getColumns()[i]`, which reflects the actual order of the columns in the executed query (matching Debezium's own `JdbcConnection.rowToArray` behavior).

This is a re-work of the previously closed PR #9838, rebased on the latest `dev`.

No. It fixes incorrect field ordering when reading CDC data with `column.include.list`.

- Added E2E test `testMysqlCdcWithColumnIncludeList` in `AbstractMysqlCDCITBase`: configures `column.include.list` with 10 columns (excluding others) in a MySQL CDC source, verifies snapshot data and then stream-stage insert/update/delete data are correctly written to the sink in the right column order.
- New job config: `mysqlcdc_to_mysql_with_column_include_list.conf`.
- Verified locally: `spotless:apply` clean, cdc-mysql module installs and `connector-cdc-mysql-e2e` test code compiles (the E2E test itself will run in CI).

### Check list

### DIFF
--- a/seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/reader/fetch/scan/MySqlSnapshotSplitReadTask.java
+++ b/seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/source/reader/fetch/scan/MySqlSnapshotSplitReadTask.java
@@ -223,7 +223,7 @@ public class MySqlSnapshotSplitReadTask
                 rows++;
                 final Object[] row = new Object[columnArray.getGreatestColumnPosition()];
                 for (int i = 0; i < columnArray.getColumns().length; i++) {
-                    Column actualColumn = table.columns().get(i);
+                    Column actualColumn = columnArray.getColumns()[i];
                     row[columnArray.getColumns()[i].position() - 1] =
                             readField(rs, i + 1, actualColumn, table);
                 }
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/AbstractMysqlCDCITBase.java
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e/src/test/java/org/apache/seatunnel/connectors/seatunnel/cdc/mysql/AbstractMysqlCDCITBase.java
@@ -96,6 +96,8 @@ public abstract class AbstractMysqlCDCITBase extends TestSuiteBase implements Te
     private static final String SOURCE_TABLE_2_CUSTOM_PRIMARY_KEY =
             "mysql_cdc_e2e_source_table_2_custom_primary_key";
     private static final String SINK_TABLE = "mysql_cdc_e2e_sink_table";
+    private static final String SINK_TABLE_COLUMN_INCLUDE =
+            "mysql_cdc_e2e_sink_table_column_include";
 
     private static final String MULTI_DATABASE_A = "mysql_multi_cdc_db_a";
     private static final String MULTI_DATABASE_B = "mysql_multi_cdc_db_b";
@@ -1388,6 +1390,57 @@ public abstract class AbstractMysqlCDCITBase extends TestSuiteBase implements Te
                                                                         TIMER_FLUSH_SRC_TABLE_2)))));
     }
 
+    @TestTemplate
+    public void testMysqlCdcWithColumnIncludeList(TestContainer container) {
+        // Clear related content to ensure that multiple operations are not affected
+        clearTable(MYSQL_DATABASE, SOURCE_TABLE_1);
+        clearTable(MYSQL_DATABASE, SINK_TABLE_COLUMN_INCLUDE);
+
+        CompletableFuture.supplyAsync(
+                () -> {
+                    try {
+                        container.executeJob("/mysqlcdc_to_mysql_with_column_include_list.conf");
+                    } catch (Exception e) {
+                        log.error("Commit task exception :" + e.getMessage());
+                        throw new RuntimeException(e);
+                    }
+                    return null;
+                });
+        await().atMost(60000, TimeUnit.MILLISECONDS)
+                .untilAsserted(
+                        () -> {
+                            log.info(
+                                    query(
+                                                    getColumnIncludeQuerySQL(
+                                                            MYSQL_DATABASE,
+                                                            SINK_TABLE_COLUMN_INCLUDE))
+                                            .toString());
+                            Assertions.assertIterableEquals(
+                                    query(
+                                            getColumnIncludeSourceQuerySQL(
+                                                    MYSQL_DATABASE, SOURCE_TABLE_1)),
+                                    query(
+                                            getColumnIncludeQuerySQL(
+                                                    MYSQL_DATABASE, SINK_TABLE_COLUMN_INCLUDE)));
+                        });
+
+        // insert update delete
+        upsertDeleteSourceTableColumnInclude(MYSQL_DATABASE, SOURCE_TABLE_1);
+
+        // stream stage
+        await().atMost(60000, TimeUnit.MILLISECONDS)
+                .untilAsserted(
+                        () -> {
+                            Assertions.assertIterableEquals(
+                                    query(
+                                            getColumnIncludeSourceQuerySQL(
+                                                    MYSQL_DATABASE, SOURCE_TABLE_1)),
+                                    query(
+                                            getColumnIncludeQuerySQL(
+                                                    MYSQL_DATABASE, SINK_TABLE_COLUMN_INCLUDE)));
+                        });
+    }
+
     private Connection getJdbcConnection() throws SQLException {
         return DriverManager.getConnection(
                 MYSQL_CONTAINER.getJdbcUrl(),
@@ -1569,5 +1622,65 @@ public abstract class AbstractMysqlCDCITBase extends TestSuiteBase implements Te
 
     private String getQuerySQL(String database, String tableName) {
         return String.format(QUERY_SQL, database, tableName);
+    }
+
+    // Query SQL for column include list test (selecting only 10 specific columns)
+    private static final String COLUMN_INCLUDE_SOURCE_SQL_TEMPLATE =
+            "select id, cast(f_binary as char) as f_binary, cast(f_blob as char) as f_blob, cast(f_long_varbinary as char) as f_long_varbinary,"
+                    + " cast(f_varbinary as char) as f_varbinary, f_smallint, f_smallint_unsigned, f_mediumint, f_mediumint_unsigned, f_int from %s.%s";
+
+    private static final String COLUMN_INCLUDE_SINK_SQL_TEMPLATE =
+            "select id, cast(f_binary as char) as f_binary, cast(f_blob as char) as f_blob, cast(f_long_varbinary as char) as f_long_varbinary,"
+                    + " cast(f_varbinary as char) as f_varbinary, f_smallint, f_smallint_unsigned, f_mediumint, f_mediumint_unsigned, f_int from %s.%s";
+
+    private String getColumnIncludeSourceQuerySQL(String database, String tableName) {
+        return String.format(COLUMN_INCLUDE_SOURCE_SQL_TEMPLATE, database, tableName);
+    }
+
+    private String getColumnIncludeQuerySQL(String database, String tableName) {
+        return String.format(COLUMN_INCLUDE_SINK_SQL_TEMPLATE, database, tableName);
+    }
+
+    private void upsertDeleteSourceTableColumnInclude(String database, String tableName) {
+        executeSql(
+                "INSERT INTO "
+                        + database
+                        + "."
+                        + tableName
+                        + " ( id, f_binary, f_blob, f_long_varbinary, f_longblob, f_tinyblob, f_varbinary, f_smallint,\n"
+                        + "                                         f_smallint_unsigned, f_mediumint, f_mediumint_unsigned, f_int, f_int_unsigned, f_integer,\n"
+                        + "                                         f_integer_unsigned, f_bigint, f_bigint_unsigned, f_numeric, f_decimal, f_float, f_double,\n"
+                        + "                                         f_double_precision, f_longtext, f_mediumtext, f_text, f_tinytext, f_varchar, f_date, f_datetime,\n"
+                        + "                                         f_timestamp, f_bit1, f_bit64, f_char, f_enum, f_mediumblob, f_long_varchar, f_real, f_time,\n"
+                        + "                                         f_tinyint, f_tinyint_unsigned, f_json, f_year )\n"
+                        + "VALUES ( 5, 0x61626374000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000,\n"
+                        + "         0x68656C6C6F, 0x18000000789C0BC9C82C5600A244859CFCBC7485B2C4A2A4CCBCC4A24A00697308D4, NULL,\n"
+                        + "         0x74696E79626C6F62, 0x48656C6C6F20776F726C64, 12345, 54321, 123456, 654321, 1234567, 7654321, 1234567, 7654321,\n"
+                        + "         123456789, 987654321, 123, 789, 12.34, 56.78, 90.12, 'This is a long text field', 'This is a medium text field',\n"
+                        + "         'This is a text field', 'This is a tiny text field', 'This is a varchar field', '2022-04-27', '2022-04-27 14:30:00',\n"
+                        + "         '2023-04-27 11:08:40', 1, b'0101010101010101010101010101010101010101010101010101010101010101', 'C', 'enum2',\n"
+                        + "         0x1B000000789C0BC9C82C5600A24485DCD494CCD25C85A49CFC2485B4CCD49C140083FF099A, 'This is a long varchar field',\n"
+                        + "         12.345, '14:30:00', -128, 255, '{ \"key\": \"value\" }', 1992 )");
+        executeSql(
+                "INSERT INTO "
+                        + database
+                        + "."
+                        + tableName
+                        + " ( id, f_binary, f_blob, f_long_varbinary, f_longblob, f_tinyblob, f_varbinary, f_smallint,\n"
+                        + "                                         f_smallint_unsigned, f_mediumint, f_mediumint_unsigned, f_int, f_int_unsigned, f_integer,\n"
+                        + "                                         f_integer_unsigned, f_bigint, f_bigint_unsigned, f_numeric, f_decimal, f_float, f_double,\n"
+                        + "                                         f_double_precision, f_longtext, f_mediumtext, f_text, f_tinytext, f_varchar, f_date, f_datetime,\n"
+                        + "                                         f_timestamp, f_bit1, f_bit64, f_char, f_enum, f_mediumblob, f_long_varchar, f_real, f_time,\n"
+                        + "                                         f_tinyint, f_tinyint_unsigned, f_json, f_year )\n"
+                        + "VALUES ( 6, 0x61626374000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000,\n"
+                        + "         0x68656C6C6F, 0x18000000789C0BC9C82C5600A244859CFCBC7485B2C4A2A4CCBCC4A24A00697308D4, NULL,\n"
+                        + "         0x74696E79626C6F62, 0x48656C6C6F20776F726C64, 12345, 54321, 123456, 654321, 1234567, 7654321, 1234567, 7654321,\n"
+                        + "         123456789, 987654321, 123, 789, 12.34, 56.78, 90.12, 'This is a long text field', 'This is a medium text field',\n"
+                        + "         'This is a text field', 'This is a tiny text field', 'This is a varchar field', '2022-04-27', '2022-04-27 14:30:00',\n"
+                        + "         '2023-04-27 11:08:40', 1, b'0101010101010101010101010101010101010101010101010101010101010101', 'C', 'enum2',\n"
+                        + "         0x1B000000789C0BC9C82C5600A24485DCD494CCD25C85A49CFC2485B4CCD49C140083FF099A, 'This is a long varchar field',\n"
+                        + "         12.345, '14:30:00', -128, 255, '{ \"key\": \"value\" }', 1999 )");
+        executeSql("DELETE FROM " + database + "." + tableName + " where id = 2");
+        executeSql("UPDATE " + database + "." + tableName + " SET f_bigint = 10000 where id = 3");
     }
 }
--- a/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e/src/test/resources/ddl/mysql_cdc.sql
+++ b/seatunnel-e2e/seatunnel-connector-v2-e2e/connector-cdc-mysql-e2e/src/test/resources/ddl/mysql_cdc.sql
@@ -318,12 +318,63 @@ CREATE TABLE mysql_cdc_e2e_sink_table
   DEFAULT CHARSET = utf8mb4
   COLLATE = utf8mb4_0900_ai_ci;
 
+CREATE TABLE mysql_cdc_e2e_sink_table_column_include
+(
+    `id`                   int       NOT NULL AUTO_INCREMENT,
+    `f_binary`             binary(64)                     DEFAULT NULL,
+    `f_blob`               blob,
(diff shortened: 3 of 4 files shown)


DanielLeens (commented):

In `MySqlSnapshotSplitReadTask#createDataEventsForTable` (the SeaTunnel-forked, split-based MySQL CDC snapshot reader), the per-row column loop picked the `Column` used to interpret each JDBC `ResultSet` value with:

```java
Column actualColumn = table.columns().get(i);
```

`table.columns()` is backed by Debezium's `TableEditorImpl#sortedColumns`, which — despite its name — is a `LinkedHashMap` ordered by **insertion/modification history**, not necessarily by `Column.position()`. It is therefore not guaranteed to line up index-for-index with the physical column order actually returned by the `SELECT *` snapshot query, particularly once the table's column set has been filtered/rebuilt (e.g. via Debezium's `column.include.list`/`column.exclude.list`). When the two orders diverge, `readField()` receives the *wrong* `Column`'s `jdbcType()` for a given `ResultSet` field, so its TIME/DATE/TIMESTAMP/TINYINT/SMALLINT special-case branches fire (or fail to fire) based on the wrong type, corrupting the value that ultimately lands in the row.

The fix replaces this with:

```java
Column actualColumn = columnArray.getColumns()[i];
```

`columnArray` comes from `ColumnUtils.toArray(rs, table)`, which builds its array strictly from the `ResultSet`'s own metadata (matching each physical result column, by name, to its `Column`). This is index-safe by construction and is exactly the pattern Debezium's own generic `JdbcConnection#rowToArray()` uses (verified directly against the vendored `debezium-core:1.9.8.Final` sources — see Issue-adjacent note below). The value-placement expression, `row[columnArray.getColumns()[i].position() - 1]`, was already correct before this change and is untouched.

# 1. Code Change Review

## 1.1 Core Logic Analysis

I pulled `io/debezium/jdbc/JdbcConnection.java` and `io/debezium/util/ColumnUtils.java` from the exact `debezium-core-1.9.8.Final-sources.jar` this module depends on (per `connector-cdc/pom.xml`'s `<debezium.version>1.9.8.Final</debezium.version>`) to confirm this independently rather than take the PR description at face value:

```java
// io.debezium.jdbc.JdbcConnection#rowToArray (upstream, unmodified by SeaTunnel)
for (int i = 0; i < columnArray.getColumns().length; i++) {
    row[columnArray.getColumns()[i].position() - 1] = getColumnValue(rs, i + 1,
            columnArray.getColumns()[i], table, databaseSchema);
}
```

This confirms the PR's fix is not a guess — it aligns exactly with the correct, already-verified-in-production pattern Debezium itself uses for the identical operation elsewhere. `RelationalSnapshotChangeEventSource#createDataEventsForTable` (the generic upstream base class SeaTunnel's `MySqlSnapshotSplitReadTask` is a customized fork of, needed to keep the DBZ-3238 TIME/DATE workaround) also delegates straight to `rowToArray` and never had the `table.columns().get(i)` pattern at all — so this bug is specific to the SeaTunnel fork's manual re-implementation of the loop, not something inherited unmodified from upstream Debezium.

I also confirmed `TableEditorImpl#sortedColumns` is a `LinkedHashMap<String, Column>` (insertion-ordered), and that column repositioning (`reposition`/`moveColumnToAfterOther`) explicitly removes-and-reinserts entries to change iteration order — concretely demonstrating that `table.columns()`'s list order is a function of *history*, not of `position()`, which is the structural reason `table.columns().get(i)` can never be relied on as a positional proxy for "the column that produced JDBC field `i+1`".

Net effect of the bug (important nuance not spelled out in the PR description): the row's **target slot** was always correct even before this fix (`row[columnArray.getColumns()[i].position() - 1]` is unchanged by this diff) — what was wrong was the **type interpretation** used to read the value into that slot. So the practical symptom is "wrong/corrupted values" (e.g. a value read via the wrong TIME/DATE/TIMESTAMP blob-decoding branch, or an unnecessary/unmet TINYINT-as-Boolean coercion) rather than a strict permutation of field order. This still matches the user-visible complaint ("out-of-order/incorrect fields"), but I'd suggest tightening the PR description's wording so future readers aren't misled into thinking this is purely an index/order bug — it's a column-identity (hence type) misattribution bug that happens to manifest as field corruption.

## 1.2 Compatibility Impact

**Classification: Fully compatible.**

- The change only affects in-memory interpretation of freshly-read `ResultSet` values during the snapshot scan loop; it does not touch `SnapshotSplit`, `MySqlOffsetContext`, watermark dispatch, or any serialized/persisted state shape.
- For jobs resumed from a checkpoint/savepoint captured **before** this fix (mid-snapshot or otherwise), nothing about the checkpoint's split boundaries, high/low watermark, or offset encoding changes — on restart the remaining un-scanned rows are simply read with the corrected logic going forward. No restore-path incompatibility.
- Downstream consumers that index into the row positionally (rather than by name) are **helped**, not hurt, by this fix: the row's positional contract (`Column.position()`-indexed slots) is unchanged; only the correctness of the value written into an already-correct slot improves.
- One thing worth calling out explicitly in the PR description for operators: rows that were **already** written incorrectly to the sink prior to upgrading (jobs run with `column.include.list` under the old code) are not retroactively corrected by this fix — that's expected for a bug fix, but should be flagged so ops teams know to consider a re-sync/backfill if they were relying on `column.include.list`.

No measurable performance impact — same number of loop iterations, same JDBC calls, just a different (cheaper, already-computed) array lookup instead of a `List.get(i)` call. No new locks, threads, retries, or resource lifecycle changes. No idempotency or ordering-guarantee change at the pipeline level; the change is confined to a leaf value-conversion step.

Unchanged — the existing `SQLException` → `ConnectException` wrapper around the scan loop is untouched. No new logging introduced, none needed for a one-line internal correctness fix.

### Issue 1 — E2E test's snapshot-phase assertion does not exercise the code path this PR fixes

- **Location**: `seatunnel-e2e/.../AbstractMysqlCDCITBase.java`, new method `testMysqlCdcWithColumnIncludeList` (added block, roughly lines 1391–1440 of the diff) and its helper `upsertDeleteSourceTableColumnInclude` (added block, roughly lines 1626–1670 of the diff).
- **Problem description**: The test calls `clearTable(MYSQL_DATABASE, SOURCE_TABLE_1)` (a `TRUNCATE`) immediately before starting the CDC job, and does **not** insert any baseline rows before the job's snapshot phase runs. The first `await().untilAsserted(...)` block therefore compares an empty source query result against an empty sink query result — a trivially-true assertion that exercises zero rows through `MySqlSnapshotSplitReadTask#createDataEventsForTable`, the exact method this PR patches. Only afterwards does `upsertDeleteSourceTableColumnInclude` insert 2 rows (`id=5`, `id=6`) and issue a `DELETE ... where id = 2` / `UPDATE ... where id = 3` that are no-ops (those ids never existed after the truncate — this helper is a verbatim copy of the pre-existing `upsertDeleteSourceTable` pattern, which relies on baseline rows 1-4 that were never seeded here). Those 2 inserted rows are consumed via the **binlog/streaming** path (a separate, unmodified code path), not the snapshot path.
- **Potential risk**: As written, this test would pass identically whether or not the one-line fix in `MySqlSnapshotSplitReadTask.java` is present or reverted, since the snapshot phase never processes a non-empty result set. The PR's stated goal ("Added E2E test ... verifies snapshot data ... correctly written") is not actually met by the test as implemented, leaving the fix without regression protection for the scenario it targets.
- **Best improvement**: Seed baseline rows into `SOURCE_TABLE_1` (via `executeSql`, similar to the existing table's other tests) *before* `container.executeJob(...)` is invoked, so the first `await()` comparison is exercised against a non-trivial snapshot containing values for type-sensitive included columns (the test already selects `f_smallint`, which triggers `readField`'s special-case `Types.SMALLINT` branch — good column choice, just needs data present at snapshot time). Also either drop the dead `DELETE id=2` / `UPDATE id=3` no-ops or adjust the ids to match rows that actually exist, to avoid a misleading no-op assertion.
- **Severity**: High (blocks confidence that the fix is actually regression-tested, though it does not affect production code correctness, which I independently verified above via source cross-reference).

### Issue 2 — Same anti-pattern exists in `legacy.SnapshotReader`, unaddressed by this PR

- **Location**: `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/io/debezium/connector/mysql/legacy/SnapshotReader.java:733` — `Column actualColumn = table.columns().get(i);` inside a `SELECT *`-driven row-reading loop, structurally identical to the bug being fixed here.
- **Problem description**: This is the same root defect pattern (positional indexing into `table.columns()` used as a proxy for the JDBC result column at the same index) living in a different, older ("legacy") snapshot reader class in the same module.
- **Potential risk**: If this class is ever reachable (e.g. via a legacy/non-incremental snapshot mode), it would exhibit an analogous bug that this PR does not fix.
- **Best improvement**: I checked reachability with `git grep -n "new SnapshotReader("` and `"new MySqlSnapshotChangeEventSource("` across `connector-cdc-mysql/src/main` — both returned zero matches, so this class appears to be dead/vestigial code inherited from the Debezium fork and not wired into SeaTunnel's active (split-based) snapshot pipeline. Given that, I'm not blocking on this, but it's worth a quick explicit confirmation from the author/maintainers that it's genuinely unreachable, and a follow-up cleanup ticket (or a matching fix, if it turns out to be reachable via some configuration path I didn't trace) rather than leaving a known-bad pattern lying around for a future connector mode to pick up.
- **Severity**: Low (informational / follow-up, not a blocker for this PR).

# 2. Code Quality Assessment

## 2.1 Coding Standards

The production-code change is minimal, surgical, and correctly scoped — a single-line substitution with no unrelated refactors, consistent with the "smallest change that fixes the problem" principle. Import list, formatting, and license headers are unaffected/unchanged in the main-code file. The new E2E resource file (`mysqlcdc_to_mysql_with_column_include_list.conf`) carries the required ASF license header.

Test code was added (new `@TestTemplate` method + helpers + new `.conf` resource). Per Issue 1 above, the added test does not, as currently written, exercise the snapshot-phase logic this PR patches — it only demonstrates that the (already correct, unmodified) binlog/streaming ingestion path works for the `column.include.list` scenario. No existing test assertions were weakened, loosened, or removed — this is purely additive, so there is no regression in existing test strictness. No unit tests were added for `MySqlSnapshotSplitReadTask` itself (understandable given it requires a live Debezium/JDBC harness; the E2E path is the appropriate venue, it just needs the fix in Issue 1).

**Stability rating: Risk present** — not because the new test is flaky, but because, as constructed, it does not actually validate the regression it claims to guard against (see Issue 1). Once Issue 1 is addressed I'd expect this to move to Stable.

## 2.3 Documentation Updates

No `Option`/config surface is added (this reuses the existing generic `debezium { ... }` passthrough map already supported by the MySQL CDC source), so no `docs/en` / `docs/zh` changes are required, and this is not an incompatible change, so `incompatible-changes.md` doesn't need an entry. Consistent with the author's own (unchecked, correctly so) PR checklist.

# 3. Architectural Soundness

Good — the fix reuses data that was already being computed (`columnArray`) two lines away, rather than introducing new state or a new lookup structure. It's the minimal correct fix, not a rewrite.

## 3.2 Maintainability

The change reduces a subtle, easy-to-miss coupling between `table.columns()`'s (undocumented, history-dependent) ordering and JDBC result ordering — future maintainers reading this method no longer need to reason about whether those two orderings happen to coincide.

## 3.3 Extensibility

No impact — this doesn't add or remove any extension point.

As covered in 1.2: no impact on checkpoint/savepoint restore, no change to persisted state formats, no change to the row's positional contract used by downstream deserialization/sink logic. Safe for in-place upgrade of already-running jobs.

# 4. Issue Summary

| # | Issue | Severity | Blocking |
| 1 | New E2E test's snapshot-phase assertion runs against an empty table and doesn't exercise `MySqlSnapshotSplitReadTask`, the method being fixed | High | Yes |
| 2 | Same `table.columns().get(i)` anti-pattern exists in unreachable-appearing `legacy.SnapshotReader.java:733`, not addressed here | Low | No (follow-up) |

# 5. Merge Recommendation

1. Blockers — must be fixed
   - Issue 1: Strengthen `testMysqlCdcWithColumnIncludeList` so the snapshot-phase comparison actually processes non-trivial rows written before the job starts (so the test would fail against the pre-fix code and pass against the fix). Please also fix or remove the no-op `DELETE id=2` / `UPDATE id=3` statements copied from `upsertDeleteSourceTable`, since those ids don't exist in this test's data set.
   - Issue 2: Confirm `io.debezium.connector.mysql.legacy.SnapshotReader` is genuinely unreachable from SeaTunnel's active MySQL CDC pipeline (my `git grep` found no construction sites), and if so consider a follow-up cleanup/removal ticket rather than leaving the same defect pattern dormant in the codebase.
   - Consider tightening the PR description's root-cause explanation: the bug is a column-identity/type misattribution during value *interpretation*, not a value-*placement*/index bug — the row's target slot (`row[columnArray.getColumns()[i].position() - 1]`) was already correct before this change.
   - Consider a short callout in the PR description noting that rows already written incorrectly under the old code (existing `column.include.list` users) are not retroactively corrected by this fix, for operator awareness.

CI on both the Apache-side pointer check and the contributor's fork was still queued/pending at review time; I did not wait on it and reviewed on source merits per policy. Please ensure it goes green (and re-run once Issue 1's test change is pushed) before merge.

DanielLeens (commented):
Thanks for the follow-up push, @corgy-w. Since my last review, one new commit landed (`8c54da82`): it adds `CREATE TABLE mysql_cdc_e2e_sink_table_column_include` (plus the matching `truncate table` line) to `ddl/mysql_cdc.sql`, to fix a CI-stability race where the JDBC sink's lazily-created table (`generate_sink_sql=true`) wasn't always visible in time for the test's first `clearTable()` call on a fresh run. That's a legitimate fix for a real flake and I confirmed the corresponding fork CI run (`mysql-cdc-connector-it`, Java 8 and 11) now passes on this exact head. However, this commit does **not** touch the test's data-seeding logic, so **Issue 1 from my previous review is still open** — see below for why, with a fresh trace against the current head. Issue 2 is unchanged/still open as a non-blocking follow-up. I re-verified the core production fix from scratch against the current head rather than just carrying my prior conclusion forward.

In `MySqlSnapshotSplitReadTask#createDataEventsForTable` (`seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/.../source/reader/fetch/scan/MySqlSnapshotSplitReadTask.java:226`), the per-row column loop picked the `Column` used to interpret each JDBC `ResultSet` value with `table.columns().get(i)`. `table.columns()` is backed by Debezium's `TableEditorImpl#sortedColumns`, a `LinkedHashMap` ordered by insertion/modification history, not by `Column.position()` and not by the actual physical order of the executed `SELECT` result set. When Debezium's `column.include.list`/`column.exclude.list` filters/reorders the table's columns, that history-ordering diverges from the query's real column order, so `readField()` is handed the *wrong* `Column`'s `jdbcType()` for a given JDBC field index — its TIME/DATE/TIMESTAMP/TINYINT/SMALLINT special-case branches then fire (or fail to fire) based on the wrong type, corrupting the value written into that row slot.

The fix replaces that with `Column actualColumn = columnArray.getColumns()[i];`, where `columnArray` comes from `ColumnUtils.toArray(rs, table)` — built strictly from the `ResultSet`'s own metadata, so it's index-safe by construction. This exactly matches the pattern Debezium's own generic `JdbcConnection#rowToArray()` uses for the identical operation. I re-read the current head's copy of this method end-to-end (not just the diff hunk) and confirmed the fix is still exactly this one-line substitution, with the value-placement expression `row[columnArray.getColumns()[i].position() - 1]` untouched, same as before.

# 1. Code Change Review

## 1.1 Core Logic Analysis

Unchanged from my last review and still correct: this is not a guess, it matches Debezium's own verified-in-production pattern for the identical operation, and the bug is specific to this SeaTunnel fork's manual re-implementation of the row-reading loop (the generic upstream Debezium class this is forked from delegates straight to `rowToArray` and never had the buggy pattern). Net effect: the row's *target slot* was always correct even before this fix; what was wrong was the *type interpretation* used to read the value into that slot — a column-identity/type-misattribution bug that manifests as field corruption, not a strict index-permutation bug. Confirmed no changes to this file between my last review and the current head.

## 1.2 Compatibility Impact

**Classification: Fully compatible.** No change to `SnapshotSplit`, offset context, watermark dispatch, or any serialized/persisted state shape. Jobs resumed from a checkpoint/savepoint captured before this fix are unaffected — the fix only changes how a fresh scan interprets already-in-flight `ResultSet` rows going forward, nothing about split boundaries or restore. As before, rows already written incorrectly under the old code (pre-existing `column.include.list` users) are not retroactively corrected — worth a one-line callout in the PR description for operator awareness, but not a blocker.

No measurable impact — same loop, same JDBC calls, a cheaper already-computed array lookup instead of `List.get(i)`. No new locks, threads, retries, or resource-lifecycle changes.

Unchanged — the existing `SQLException` → `ConnectException` wrapper around the scan loop is untouched. No new logging needed for a one-line internal correctness fix.

### Issue 1 — E2E test's snapshot-phase assertion still does not exercise the code path this PR fixes (carried over, still open)

- **Location**: `seatunnel-e2e/.../AbstractMysqlCDCITBase.java`, `testMysqlCdcWithColumnIncludeList` (`:1393`) and `upsertDeleteSourceTableColumnInclude` (`:1626`).
- **Problem description**: The test still calls `clearTable(MYSQL_DATABASE, SOURCE_TABLE_1)` (a `TRUNCATE`) immediately before starting the CDC job, with no baseline rows inserted beforehand. The first `await().untilAsserted(...)` therefore still compares an empty source query against an empty sink query — a trivially-true assertion that runs zero rows through `MySqlSnapshotSplitReadTask#createDataEventsForTable`, the exact method this PR patches. `upsertDeleteSourceTableColumnInclude` still inserts its 2 rows (`id=5`,`id=6`) only afterward, consumed via the binlog/streaming path — a separate, unmodified code path. The `DELETE ... where id=2` / `UPDATE ... where id=3` no-ops (those ids don't exist post-truncate) are also still present, unchanged from the version I reviewed last time.
- **What changed in the interim, and why it doesn't resolve this**: The only new commit (`8c54da82`) adds a DDL `CREATE TABLE` for the sink table, purely to fix a "table not visible yet" CI race. It does not add any source-table seeding or otherwise touch the test's logic, so this is a genuinely orthogonal fix — good to have, but Issue 1 is not addressed by it.
- **Context for fairness** (not a mitigation, just useful framing): this exact clear-then-run-then-insert pattern is the norm across almost the entire `AbstractMysqlCDCITBase` class (e.g. `testMysqlCdcCheckDataE2e:141`, `testMysqlCdcCheckDataE2eWithHeartbeat:185`) — so the "snapshot phase always sees zero rows" gap is largely inherited from an established test-harness convention in this file, not invented by this PR. One test in the class, `testMysqlCdcCheckDataWithDisableExactlyonce:278`, deliberately *skips* clearing `SOURCE_TABLE_1`, relying on leftover data from prior test execution order to get non-trivial snapshot coverage — itself a fragile pattern I wouldn't want copied here. None of this changes my conclusion: this specific test's stated purpose ("verifies snapshot data ... correctly written") is not met as written, and since the entire point of this PR is a snapshot-phase correctness fix, that's exactly the coverage this test needs to actually provide.
- **Potential risk**: As written, this test passes identically whether the one-line fix in `MySqlSnapshotSplitReadTask.java` is present or reverted (confirmed: it's now proven to pass in fork CI on this head — `mysql-cdc-connector-it` on Java 8 and 11 — which is consistent with, not contradictory to, this being a trivial pass since the snapshot phase never sees real rows). The fix ships without regression protection for the scenario it targets.
- **Best improvement**: Seed baseline rows into `SOURCE_TABLE_1` (via `executeSql`) *before* `container.executeJob(...)` is invoked, so the first `await()` comparison exercises a non-trivial snapshot containing values for type-sensitive included columns (the test already selects `f_smallint`, which triggers `readField`'s `Types.SMALLINT` branch — good column choice, just needs data present at snapshot time). Also fix or drop the dead `DELETE id=2`/`UPDATE id=3` no-ops.
- **Severity**: High (blocks confidence that the fix is actually regression-tested; does not affect production-code correctness, which I independently re-verified above).
- **Raised by another reviewer**: No (carried over from my own prior review; still unresolved).

### Issue 2 — Same anti-pattern in `legacy.SnapshotReader`, still unaddressed (unchanged, non-blocking)

- **Location**: `seatunnel-connectors-v2/connector-cdc/connector-cdc-mysql/src/main/java/io/debezium/connector/mysql/legacy/SnapshotReader.java:733` — `Column actualColumn = table.columns().get(i);`, same root defect pattern, in an older "legacy" snapshot reader class in the same module.
- **Re-verified reachability on current head**: `git grep -n "new SnapshotReader("` and `"new MySqlSnapshotChangeEventSource("` under `connector-cdc-mysql/src/main` both still return zero matches — still appears dead/vestigial, not wired into SeaTunnel's active split-based snapshot pipeline.
- **Best improvement**: Not blocking. Worth an explicit confirmation from the author/maintainers that it's genuinely unreachable, and a follow-up cleanup ticket (or a matching fix if some configuration path does reach it) rather than leaving a known-bad pattern dormant for a future snapshot mode to inherit.
- **Severity**: Low (informational / follow-up).

# 2. Code Quality Assessment

## 2.1 Coding Standards

The production-code change remains a minimal, surgical, single-line substitution with no unrelated refactors. The new DDL addition carries no license-header concerns (SQL file, matches existing style in the same file). No standards issues in the new commit.

Per Issue 1, the added test still does not exercise the snapshot-phase logic this PR patches — it demonstrates the binlog/streaming ingestion path works for `column.include.list`, which is useful but not what's being fixed. No existing test assertions were weakened or removed; this remains purely additive. The new commit is itself a test-infrastructure stability fix (avoids a "sink table not yet created" race) and does not introduce new flakiness — I don't see hard sleeps, timing-dependent assertions, or shared mutable static state in either commit.

**Stability rating: Risk present** — same rating as before, for the same reason (Issue 1), now with the added confirmation that the test does pass in real fork CI, which is exactly the evidence that it's not exercising the snapshot path (a snapshot-path bug of this shape would corrupt a TIME/DATE/SMALLINT-typed value read at scan time, and this test, as written, cannot observe that failure mode since no such value is ever scanned).

## 2.3 Documentation Updates

No `Option`/config surface added; this reuses the existing generic `debezium { ... }` passthrough already supported by the MySQL CDC source. No `docs/en`/`docs/zh` changes required, and this is not an incompatible change so `incompatible-changes.md` doesn't need an entry.

# 3. Architectural Soundness

Good — reuses data already computed two lines away (`columnArray`) rather than introducing new state. Minimal correct fix, not a rewrite.

## 3.2 Maintainability

Removes a subtle, easy-to-miss coupling between `table.columns()`'s undocumented, history-dependent ordering and JDBC result ordering.

## 3.3 Extensibility

No impact — no extension points added or removed.

No impact on checkpoint/savepoint restore, no change to persisted state formats, no change to the row's positional contract used downstream. Safe for in-place upgrade of already-running jobs.

# 4. Issue Summary

| # | Issue | Severity | Blocking | Status |
| 1 | New E2E test's snapshot-phase assertion runs against an empty table and doesn't exercise `MySqlSnapshotSplitReadTask`, the method being fixed | High | Yes | Still open — new commit only fixed an unrelated DDL/table-creation CI race |
| 2 | Same `table.columns().get(i)` anti-pattern exists in unreachable-appearing `legacy.SnapshotReader.java:733`, not addressed here | Low | No (follow-up) | Unchanged |

# 5. Merge Recommendation

The production fix itself has been correct and stable across both rounds of review — my concern remains entirely about test coverage for the exact regression this PR claims to guard against.

1. Blockers — must be fixed
   - Issue 1: Seed baseline rows into `SOURCE_TABLE_1` before `container.executeJob(...)` so the first `await()` in `testMysqlCdcWithColumnIncludeList` actually processes non-trivial rows through the snapshot path (i.e., the test would fail against the pre-fix code and pass against the fix). Please also fix or remove the no-op `DELETE id=2`/`UPDATE id=3` statements, since those ids don't exist in this test's data set.
   - Issue 2: Confirm `io.debezium.connector.mysql.legacy.SnapshotReader` is genuinely unreachable from SeaTunnel's active MySQL CDC pipeline, and if so consider a follow-up cleanup/removal ticket.
   - Consider a short PR-description callout that rows already written incorrectly under the old code (existing `column.include.list` users) are not retroactively corrected by this fix, for operator awareness.

CI: `Build` (apache-side pointer) is green, and I independently checked the underlying fork run — `mysql-cdc-connector-it` passed on both Java 8 and Java 11 for the current head (`8c54da82`). The PR is 4 commits behind `dev`, but since CI is green on the current head and there's no merge conflict (`mergeable: MERGEABLE`; the `BLOCKED` merge status is purely from the missing required review, not a conflict), I don't think a `dev` sync is needed as a precondition here — it's fine to land once Issue 1 is addressed.

davidzollo (approved):
+1
LGTM

